PR #2, opened 2018-08-04
Dockerfile: verify linuxqtdeploy checksum, break long lines

hkjn:
This makes sure we have exactly the `linuxdeployqt-continuous-x86_64.AppImage` file that we expect, since we fetch it remotely during build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,11 @@ ADD Makefile /tmp/
 RUN make -C /tmp/ envinit
 
 # Needed for qt5. fuse is needed to run the linuxdeployqt appimage.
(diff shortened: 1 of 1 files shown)


benma:
Thanks. Would it make sense to commit the tool into our source tree (or clone the repo at a specific revision and build it)? Otherwise, in a couple of weeks when there is a new upstream release, it won't break `dockerinit`.

hkjn:
@benma and I discussed offline, but for the benefit of posterity: yup, ideally this dependency (`linuxqtdeploy`) would be part of the repo, so that a local checkout could be built even in the absence of an Internet connection. Second best would be to lock in a specific released version, so that the build specifies a known good release of the dependency to pull in (the integrity of which is verified by checksum, as in this change).

jvaltin:
I would go a step further and make the packaging an entirely different target. It is then possible to run it in parallel and fail only for the specific parts that failed.

PR #3, opened 2018-08-04
docker: Give hint in dockerdev.sh if godbb image is missing

hkjn:
Without this change, `./scripts/dockerdev.sh` (and `make dockerdev`) fails
with `Unable to find image 'godbb:latest' locally` on clean checkout (tested at 5d8f347).

After this change, we check for the `godbb` image, and if it does not exist, we build the `Dockerfile`
in the root of the repo and tag it with this name.


benma:
Thank you :dancing_men:

PR #10, opened 2018-08-06
frontend: disable 2FA button unless paired

x1ddos:
This change simply prevents enabling 2FA from the web UI
if no phone is paired with the device, as reported by the backend.

It remains to be seen and verified whether the 2FA can be enabled
directy by sending "lock" command to the device
or backend which propagates the command to device.

R: @thisconnect @benma

### DIFF
--- a/frontends/web/src/routes/device/settings/settings.jsx
+++ b/frontends/web/src/routes/device/settings/settings.jsx
@@ -156,7 +156,7 @@ export default class Settings extends Component {
                                     <DeviceLock
                                         deviceID={deviceID}
                                         onLock={() => this.setState({ lock: true })}
-                                        disabled={lock} />
+                                        disabled={lock || !paired} />
                                 </div>
 
                                 <hr />


thisconnect:
cc @KasparEtter

x1ddos:
I should've probably added to the commit message: if a device is locked (2FA is enabled) but the bitbox app has no pairing info/channel, nothing prevents user from re-pairing a phone, which sounds right to me. At least, looking at the pairing code.

x1ddos:
> I should've probably added to the commit message: if a device is locked (2FA is enabled) but the bitbox app has no pairing info/channel, nothing prevents user from re-pairing a phone, which sounds right to me. At least, looking at the pairing code.

The fact that one can re-pair a new phone even after enabling 2FA wasn't clear to me until I read the source code. This isn't mentioned even on https://shiftcrypto.ch/start. I think the guide really needs a topic on 2FA while on the device settings screen.

x1ddos:
Please, do merge this. I don't have write permissions.

> If you insert another BitBox, the desktop app would still say "paired".
That is another problem, indeed. Or rather, a feature request/improvement for the currently WAI pairing as it is. Will create a bug for this.

thisconnect:
> I don't have write permissions.
same

screen shot 2018-08-08 at 9 45 52 am

thisconnect:
@benma yes I forgot to accept, thank you!

PR #11, opened 2018-08-09
Enable security compile and link time features

jvaltin:
This change set enables golang compiler and linking security features.

### DIFF
--- a/frontends/qt/server/Makefile
+++ b/frontends/qt/server/Makefile
@@ -1,7 +1,7 @@
 SECURITYFLAGS = " -O2 -D_FORTIFY_SOURCE=2 "
 SECURITYFLAGS += " -fstack-protector-all -fPIC -fPIE"
 SECURITYFLAGS += " -fstack-check "
-EXTLDFLAGS:="-Wl,-z,now,-z,relro,-z,noexecstack,-z,noexecheap,-z,defs,-pie,-s,--verbose"
+EXTLDFLAGS:="-Wl,-z,now,-z,relro,-z,noexecstack,-z,noexecheap,-z,defs,-pie,-s,--verbose,-E"
 CXXFLAGS += ${SECURITYFLAGS}
 CFLAGS += ${SECURITYFLAGS}
 LFLAGS += ${EXTLDFLAGS}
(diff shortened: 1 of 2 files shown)


jvaltin:
This change set should make memory corruption exploitation more of a challenge.

Related reading: 
https://rain-1.github.io/golang-aslr.html
https://developers.redhat.com/blog/2018/03/21/compiler-and-linker-flags-gcc/

frontends/qt/server/Makefile
```diff
@@ -1,5 +1,17 @@
...
+CFLAGS += ${SECURITYFLAGS}
+LFLAGS += ${EXTLDFLAGS}
+export CXXFLAGS
+export CFLAGS
+export LFLAGS
+export CGO_LDFLAGS_ALLOW=".*"
+export CGO_CFLAGS_ALLOW=".*"
 build:
-	CGO_ENABLED=1 go build -ldflags="-s -w" -buildmode=c-shared -o libserver.so
+	CGO_ENABLED=1 go build -buildmode=c-shared -ldflags -extldflags=${EXTLDFLAGS} -o libserver.so
```
benma: This does not work on macOS unfortunately: `running clang failed: exit status 1: unknown option -z`.

Please fix by making separate targets for now. We can fix if for macOS later.

jvaltin: Could you share the version of OS X, clang, ld, and golang? If you edit the EXTLDFLAGS, what works?

benma:
macOS 10.13.6
```clang --version

Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
```

Go 1.10.2

`ld` has no option to show the version. 

Will debug more tomorrow.

jvaltin:
For OS X and for Windows, we will set different compiler flags. I'll push a branch with updates that cover these different build environments.

frontends/qt/server/server.go
```diff
@@ -35,6 +35,7 @@ typedef struct ConnectionData {
 } ConnectionData;
 #endif
 */
+// #cgo CFLAGS: -O2 -D_FORTIFY_SOURCE=2 -fstack-check -fstack-protector-all -fPIC -fPIE
```
benma: Same flags are in the server/Makefile. Are both needed?

jvaltin:
I have a new branch which merges #11 and #13 into one set of changes.

PR #12, opened 2018-08-09
add non-standard library path to postinst script

jvaltin:
Add the BitBox library path to the system library path.

### DIFF
--- a/frontends/qt/resources/deb-afterinstall.sh
+++ b/frontends/qt/resources/deb-afterinstall.sh
@@ -6,3 +6,6 @@ printf "KERNEL==\"hidraw*\", SUBSYSTEM==\"hidraw\", ATTRS{idVendor}==\"03eb\", A
 
 udevadm control --reload
 udevadm trigger
+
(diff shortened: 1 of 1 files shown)


benma:
Why? The libs are being loaded from there already, aren't they?

jvaltin:
Without those two entries, libserver.so and libQT5Web*.so are not located at run time on my test systems.

benma:
Why not? Did it work with the build scripts currently on master? The app is supposed to find all libs locally, without changing the ld search path.

jvaltin:
> Why not? Did it work with the build scripts currently on master? The app is supposed to find all libs locally, without changing the ld search path.

The debs from the last release work on my system. The debs I produced from the repo did not work on my system. Adding this to my build process allowed BitBox to run.

benma:
When I asked why, I asked for why it works everywhere but not on your system. The commit should fix the root of the issue, ideally without changing the host system, and if not possible, the commit message should have an adequate explanation for how your system is different than the rest.

Also feel free to close if it is low prio.

benma:
Closing due to inactivity. Please reopen if this is a priority.

PR #13, opened 2018-08-09
Enable security compile and link time features

jvaltin:
This has been tested for the qt-linux target on amd64 GNU/Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,17 @@
 SHELL:=/bin/bash
 WEBROOT:=`pwd`/frontends/web
 
+SECURITYFLAGS = "-O2 -D_FORTIFY_SOURCE=2 "
+SECURITYFLAGS += " -fstack-protector-all -fwrapv -fPIC -fPIE "
+SECURITYFLAGS += " -fstack-check "
+EXTLDFLAGS:="-Wl,-z,now,-z,relro,-z,noexecstack,-z,noexecheap,-z,defs,-pie,-s,--verbose"
+export CXXFLAGS += ${SECURITYFLAGS}
+export CFLAGS += ${SECURITYFLAGS}
+export LFLAGS += ${EXTLDFLAGS}
+export QMAKE_CXXFLAGS = ${SECURITYFLAGS}
+export QMAKE_FLAGS = ${SECURITYFLAGS}
+export QMAKE_LDFLAGS = ${EXTLDFLAGS}
+
 catch:
 	@echo "Choose a make target."
 envinit:
--- a/frontends/qt/Makefile
+++ b/frontends/qt/Makefile
@@ -1,3 +1,14 @@
+SECURITYFLAGS = " -O2 -D_FORTIFY_SOURCE=2 "
+SECURITYFLAGS += " -fstack-protector-all -fPIC -fPIE"
+SECURITYFLAGS += " -fstack-check "
+EXTLDFLAGS:="-Wl,-z,now,-z,relro,-z,noexecstack,-z,noexecheap,-z,defs,-pie,-s,--verbose"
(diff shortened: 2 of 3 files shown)


jvaltin:
As with #11, this change set should make memory corruption exploitation more of a challenge.

Related reading:
https://developers.redhat.com/blog/2018/03/21/compiler-and-linker-flags-gcc/

Makefile
```diff
@@ -15,6 +15,17 @@
 SHELL:=/bin/bash
 WEBROOT:=`pwd`/frontends/web
 
+SECURITYFLAGS = "-O2 -D_FORTIFY_SOURCE=2 "
```
benma: Please remove those flags from this Makefile, as this makefile has a lot of unrelated targets and it is difficult to see where those flags apply in the end.

The only locations are the .pro file for the C++ compilation, which you already modified, and the qt/server/Makefile where the Go lib is compiled.

Do those flags affect the server.go compilation? If so, pls move the relevant changes over there.

frontends/qt/Makefile
```diff
@@ -1,3 +1,14 @@
+SECURITYFLAGS = " -O2 -D_FORTIFY_SOURCE=2 "
```
benma: Same here: better to move the flags to the innermost scope so it's clear how and where they apply.

jvaltin: It's meant to apply to everything inside of the `frontends/qt/` directory generally.

Does this set of flags interfere in any specific build target in `frontends`?

Makefile
benma: I don't think so, I meant this more from a code clarity point of view. When it is here, and again in frontends/qt/Makefile, it is redundant (maybe the flags even appear twice in the compile/link calls? haven't checked). 

Just easier to read and understand if the changes are applied near the actual compilation (which would be the .pro file and server/Makefile), instead of far away.

x1ddos: It seems like the same flags are used across both this and #11 changes.
Maybe what you want is a `Makeflags.inc` or something and include it where they're actually used like @benma says?

Also, I would fold #11 and this #13 into a single change. They both do exact same thing, only in different locations. If there were a single pull request with a "makeflags.inc", included and used where the actual compile and link happens, it would probably be easier to grok what's changed and where, and most likely easier to rollback in case of issues. 

For instance, this makefile exports  `QMAKE_CXXFLAGS` and other vars but later overwritten in bitbox.pro. If I didn't know better, I'd modify these and assume it changes everywhere, leading to a possible hard bug to find.

jvaltin: Builders of the software may call our Makefiles from any directory and we cannot assume a Makefile will only be called from the top most Makefile.

One commonly accepted solution to this is to generate Makefiles and to have a single configuration system. My understanding is that autoconf and automake and other tools of that sort are not welcome in the build system.

An include is probably a fine solution.

benma: > Builders of the software may call our Makefiles from any directory and we cannot assume a Makefile will only be called from the top most Makefile.

I don't understand the issue. The .pro file is handled by qmake in any case, so any flags relevant for the C++ Qt app should be there.

libserver is gonna be built by qt/server/Makefile in any case, so any flags relevant to that should be there (or inside server.go, in case it doesn't work from the Makefile).

Explicit > implicit.

jvaltin: Agreed - both of those places are where we must add flags. I've just reworked `qt/server/Makefile` and I think I'll also rework the .pro file shortly.

My observation was that someone may build the `qt/server/Makefile` target by calling it from the top most level, or from inside `qt/server/` itself. We need to define things inside the bottom most Makefile, for the latter case. We can do this with an include, of course.

jvaltin:
I have a new branch which merges #11 and #13 into one set of changes.

PR #14, opened 2018-08-09
reduce attack surface in QT desktop application

jvaltin:
The first two changes force the use of the 2d software rendering engine:

	qputenv("QMLSCENE_DEVICE", "softwarecontext");
	qputenv("QT_QUICK_BACKEND", "software");

Disable just-in-time regex engine:

	qputenv("QT_ENABLE_REGEXP_JIT", "0");

Disable just-in-time javascript engine:

	qputenv("QV4_FORCE_INTERPRETER", "1");

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -54,6 +54,15 @@ int main(int argc, char *argv[])
 //     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", QByteArray("1"));
 // #endif // QT_VERSION
 
+// QT configuration parameters which change the attack surface for memory
+// corruption vulnerabilities
+#if QT_VERSION >= QT_VERSION_CHECK(5,8,0)
+    qputenv("QMLSCENE_DEVICE", "softwarecontext");
+    qputenv("QT_QUICK_BACKEND", "software");
+    qputenv("QT_ENABLE_REGEXP_JIT", "0");
+    qputenv("QV4_FORCE_INTERPRETER", "1");
+#endif
+
     QApplication a(argc, argv);
     a.setApplicationName(QString("BitBox Wallet"));
     a.setOrganizationDomain("shiftcrypto.ch");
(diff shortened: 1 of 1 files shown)


frontends/qt/main.cpp
```diff
@@ -54,6 +54,15 @@ int main(int argc, char *argv[])
 //     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", QByteArray("1"));
 // #endif // QT_VERSION
 
+// QT configuration parameters which change the attack surface for memory
+// corruption vulnerabilities
+#if QT_VERSION >= QT_VERSION_CHECK(5,8,0)
```
benma: Safer to produce an error if the version is < 5.8?

benma: If we want to enable those protections, then I guess it would be good to not accidentally skip them by using an old Qt version.

Is there a reason to support older versions?

jvaltin: It seems preferable to ship a newer version of QT when we build and support any specific release. What version should we target? QT 5.11?

jvaltin: > Is there a reason to support older versions?

On Debian Stable, 5.7 is the current QT library version. Ideally, a user should be able to build with the correct system libraries. This will also aid us in building smaller, easier to manage packages.

TheCharlatan: I would always target Long Term Support releases, which is currently Qt 5.9.x .

jvaltin:
@benma Regarding environment variables for QT debugging: http://doc.qt.io/qt-5/debug.html#environment-variables-recognized-by-qt

jvaltin:
@benma - is there anything else that needs to be done before we are able to merge this branch?

jvaltin:
If it doesn't cause adverse effects, I think it's reasonable that the QT library will behave as is documented. I'm mostly curious if it does cause any adverse effects.

It should also be possible to run the application in a debugging mode that shows the status of each component. According to the [QT blog](http://blog.qt.io/blog/2017/01/18/opengl-implementation-qt-quick-app-using-today/) setting an environment variable of `QSG_INFO=1` may be a relevant technique.

PR #16, opened 2018-08-12
backend: guard Device's internals against concurrent access

x1ddos:
`devices/bitbox.Device`'s channel, onEvent and closed fields are accessed concurrently and need to be guarded with a mutex.

I consider this an intermediate solution, built on top of existing code structure. I believe a more elegant solution exists but would require substantial refactoring.

`bitbox.Device` now has a new unexported field which holds the channel config dir to aid automated testing. A couple nice side effects of this:

- your real channel.json config in the HOME dir is not touched during
  tests, if present;
- tests shouldn't hammer prod relay endpoint anymore because
  the channel config dir is a newly created temp dir for each run,
  missing the channel.json file so dbb.listenForMobile isn't run.

Verified locally with `go test` and manually resetting a real bitbox device and going through the pairing flow.

R: @stephaniestroka @benma

### DIFF
--- a/util/config/file.go
+++ b/util/config/file.go
@@ -22,9 +22,10 @@ import (
 	"runtime"
 )
 
-// DirectoryPath returns the absolute path to the application's directory.
+// DirectoryPath returns the absolute path to the application's directory
+// in the user standard config location.
 func DirectoryPath() string {
-	switch goos := runtime.GOOS; goos {
+	switch runtime.GOOS {
 	case "darwin":
 		return filepath.Join(os.Getenv("HOME"), "Library", "Application Support", "bitbox")
 	case "windows":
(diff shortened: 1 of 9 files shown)


util/config/file.go
```diff
@@ -22,9 +22,10 @@ import (
 	"runtime"
 )
 
-// DirectoryPath returns the absolute path to the application's directory.
+// DirectoryPath returns the absolute path to the application's directory
+// in the user standard config location.
 func DirectoryPath() string {
-	switch goos := runtime.GOOS; goos {
+	switch runtime.GOOS {
```
benma: hmm, this is almost the same as https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/qt/server/server.go#L138 :thinking: 

Can we delete this and use backend.arguments.MainDirectoryPath() instead when creating the manager in https://github.com/digitalbitbox/bitbox-wallet-app/pull/16/files#diff-1190f94ab2ad6f04f64211d316e67a04R476 ?

x1ddos: I tried by there are too many `logging.Get()` calls.
So, I proposed the reverse: remove `getAppFolder` and use `config.AppDir` instead, in https://github.com/digitalbitbox/bitbox-wallet-app/pull/20

x1ddos:
PTAL. Some minor edits to the tests I didn't notice before.

x1ddos:
FTR review comments are in https://github.com/digitalbitbox/bitbox-wallet-app/pull/20

PR #17, opened 2018-08-12
frontend: add guide section about 2FA

x1ddos:
This section is available while on the device settings screen.
Took the text from https://shiftcrypto.ch/start and made minor adjustments.

This is missing German translation at the moment..

R: @thisconnect

### DIFF
--- a/frontends/web/src/i18n/en.json
+++ b/frontends/web/src/i18n/en.json
@@ -645,6 +645,10 @@
                 "text": [ "You can remove the SD card from the BitBox at any time as long as you are not in the process of creating or restoring a backup." ]
             },
             {
+                "title": "What is a hidden wallet?",
+                "text": [ "It is a second wallet on the same device protected by a different device password and recovery password, which you can use for plausible deniability. The same backup seed is used for both your normal and hidden wallet, so no additional backup is required." ]
+            },
+            {
                 "title": "How to securely pair with your phone",
                 "text": [ "After having downloaded our mobile app for either iOS or Android, you scan the displayed QR code, which sets up a secure channel between the mobile app and this application. Once scanned, follow the instructions in the mobile app." ]
             },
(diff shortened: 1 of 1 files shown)


frontends/web/src/i18n/en.json
```diff
@@ -648,6 +648,12 @@
                 "title": "How to securely pair with your phone",
                 "text": [ "After having downloaded our mobile app for either iOS or Android, you scan the displayed QR code, which sets up a secure channel between the mobile app and this application. Once scanned, follow the instructions in the mobile app." ]
             },
+            {   "title": "How does 2FA work",
```
KasparEtter: Move it below "What is a hidden wallet?" (since the Secrets are above Pairing).

Formulate it as a question (FAQ-like). Suggestion: "How does Two Factor Authorization (2FA) work?".

frontends/web/src/i18n/en.json
```diff
@@ -648,6 +648,12 @@
...
+                "text": [
+                    "When 2FA is enabled, possession of a mobile phone is required in order to spend coins. Under the hood, an encrypted single-use PIN is sent to the mobile app, decrypted there, and returned to the BitBox when pressing the Accept button. This communication with the device is done via the channel between a mobile phone and this desktop app established during pairing.",
```
KasparEtter: Suggestions:
– "possession of a mobile phone is required in order to spend coins" –> "all transactions have to approved on the paired mobile phone" (I don't like the "a mobile phone" because it's a very specific one.)
– "PIN" –> "number" (I have no idea why Douglas called this PIN (also in the code). It has nothing to do with a personal identification number. Nonce and challenge would both be more suitable.)
– "between a mobile phone" –> "between the mobile phone"

frontends/web/src/i18n/en.json
x1ddos: Actually, moved "What is a hidden wallet?" below "How can I disable the two factor authorization (2FA)?", so the whole list order is even more consistent imho:

- "How can I eject the BitBox?"
- "How can I eject the SD card?"
- "How to securely pair with your phone"
- "How does Two Factor Authorization (2FA) work?"
- "How can I disable the two factor authorization (2FA)?"
- "What is a hidden wallet?"

Otherwise, you'd have:

- "How to securely pair with your phone"
- "What is a hidden wallet?"
- "How does Two Factor Authorization (2FA) work?"
- "How can I disable the two factor authorization (2FA)?"

In other words, doesn't feel like "hidden wallet" belongs in between phone pairing and 2FA.

Re: title. Updated to your suggestion.

image

frontends/web/src/i18n/en.json
x1ddos: All done. Thanks!

x1ddos:
PTAL

x1ddos:
> The hidden wallet is above 2FA in the device settings, thus I would take this entry first in the device settings

Ah, good point. Missed it. Let me quickly do that.

x1ddos:
Done. PTAL.

image

KasparEtter:
Great you did a test and posted it here! 🙂 Shouldn't the question at the bottom be adapted to "How can I disable Two Factor Authorization (2FA)?" then? (Or alternatively, change your new entry to lowercase?)

x1ddos:
Yes, it should! :)
I guess I was focusing on the new stuff.
How about now:

image

PR #23, opened 2018-08-17
util/config: check the app config dir only once

x1ddos:
There's no reason why AppDir should do filesystem stat calls every time it is called. It is enough to do it once at the startup and reuse the value for the entire lifetime of the app process.

One of the side effects of this is that app dir can now be replaced with a temp dir in tests.

Also added a test for AppDir but it has any value only on linux-based platforms.

R: @benma

### DIFF
--- a/util/config/appdir_test.go
+++ b/util/config/appdir_test.go
@@ -1,0 +1,83 @@
+// Copyright 2018 Shift Devices AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"testing"
+
+	"github.com/digitalbitbox/bitbox-wallet-app/util/test"
+)
+
+// TestAppDir runs itself in a child process for each test case.
+// It sets up HOME and XDG_CONFIG_HOME env variables, passing the expected
+// result in BITBOX_TEST_APP_DIR.
+//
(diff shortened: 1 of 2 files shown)


util/config/appdir_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/digitalbitbox/bitbox-wallet-app/util/test"
+)
+
+// TestAppDir runs itself in a child process for each test case.
+// It sets up HOME and XDG_CONFIG_HOME env variables, passing the expected
+// result in BITBOX_TEST_APP_DIR.
+//
+// If a test in a child process fails, it exits with a non-0 code
+// which makes the actual TestAppDir test fail.
+func TestAppDir(t *testing.T) {
```
benma: In the Go projects I worked for, exported functions are tested from a separate package (`package config_test`).

Can be applied here I think.

x1ddos: You're right, in this particular case TestAppDir doesn't need anything unexported to check but it is very uncommon. Feels like the idea of a separate package was ported from C/C++ or Java.

You might've seen separate test packages where it would otherwise get in the way of some kind of static analysis or other advanced stuff like GC collection stats, memory contention, etc. during tests. But it is very rare and there have to be a very good reason for it.

Anyway, the official getting-started docs about writing tests are here:
https://golang.org/doc/code.html#Testing and other pages linked from there.

> Add a test to the stringutil package by creating the file $GOPATH/src/github.com/user/stringutil/reverse_test.go containing the following Go code.
>
> package stringutil
> import "testing"
> func TestReverse(t *testing.T) {

I don't mind separating it into `config_test` if you insist but doing so just because it's an exported function doesn't really justify it imho.

x1ddos: Oh, another reason for a separate package can be where you already have _test.go tests in the same package but for some very weird reason need to add more tests which would otherwise somehow shadow or conflict with the regular _test.go package tests.

But again, would be very strange and uncommon. :)

benma: I see. :man_shrugging: I don't insist :) I did like the idea of external package testing as you then are really testing the public package API.

x1ddos: Yeah, I totally get the point about public API, although tbh in my Go tesing experience it mostly got in the way whenever I had to deal with it. More inconveniences than anything useful in the long run.

For instance, in C++ you have friend functions and other constructs which can aid testing internal stuff. Go is much simpler... for better or worse :)

PR #26, opened 2018-08-21
Add step in full node connection, to restart the wallet

TheCharlatan:
Also reflect on the role of the node, if the default nodes are left in. Not sure, if all of this is actually true. @benma please review, and explain if it is not :P

### DIFF
--- a/frontends/web/src/i18n/en.json
+++ b/frontends/web/src/i18n/en.json
@@ -198,6 +198,7 @@
             "step1": "Step 1. Enter the endpoint.",
             "step2": "Step 2: Enter a certificate of the server's certificate chain. Alternatively, download the remote certificate and compare it visually.",
             "step3": "Step 3: Check the connection and add the server.",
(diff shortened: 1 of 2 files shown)


frontends/web/src/i18n/en.json
```diff
@@ -198,6 +198,7 @@
             "step1": "Step 1. Enter the endpoint.",
             "step2": "Step 2: Enter a certificate of the server's certificate chain. Alternatively, download the remote certificate and compare it visually.",
             "step3": "Step 3: Check the connection and add the server.",
+            "step4": "Step 4: Restart the wallet. If you do not remove the default servers, your own node will be added as a redundancy",
```
benma: Missing a `.` at the end :)

Otherwise correct!

TheCharlatan:
Force pushed to correct nit.

benma:
Thanks!

PR #29, opened 2018-08-24
ci with travis

jvaltin:


jvaltin:
This branch is not ideal and it is probably also a reasonable start.

When travis-ci is enabled (by @benma, I think) for this repository, it uses travis to build the dockerdev Docker file and then it builds a second docker file which simply runs any targets we wish to build and test. It takes around five minutes as it builds both images from scratch. It pulls shift's qt docker image, so that part is extremely fast.

At the moment, it correctly detects that the build is broken from a clean checkout.

Dockerfile.travis
```diff
@@ -0,0 +1,27 @@
...
+ENV GOPATH /opt/go
+ENV GOROOT /opt/go_dist/go
+ENV PATH $GOROOT/bin:$GOPATH/bin:$PATH
+ENV PATH /opt/qt5/bin:$PATH
+
+RUN mkdir -p /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/
+ADD . /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/
+
+RUN make -C /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/ init
+RUN make -C /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/ qt-linux
```
benma: move down one line? better to see `ci` failures early, and the slower build that breaks less often after.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  - docker
+
+addons:
+  apt:
+    packages:
+      - make
+
+script:
+- docker build --tag bitbox-wallet-dev .
+- mv Dockerfile Dockerfile.dev
```
benma: i think you can specify the dockerfile with `docker build -f Dockerfile.travis`, so you don't have to do this awkward file swap :)

Dockerfile.travis
```diff
@@ -0,0 +1,27 @@
...
+RUN make -C /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/ ci
+RUN make -C /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/ ci-fast
```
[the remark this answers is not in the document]
jvaltin: Agreed, it's removed now.

.travis.yml
jvaltin: Agreed.

PR #33, opened 2018-08-27
Add MacOS to Travis CI build

jvaltin:
This adds MacOS to the Travis CI build process. It preps the Travis machine with a Makefile target and then builds the QT target for Mac OS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,6 @@ osx-init:
 	export GOPATH=~/go/
 	export PATH=$PATH:~/go/bin
 	make init
-
 servewallet:
 	go install ./cmd/servewallet/... && servewallet
 servewallet-mainnet:
--- a/scripts/travis-ci.sh
+++ b/scripts/travis-ci.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+    set -e
+    docker build --tag bitbox-wallet-dev -f Dockerfile .
+    docker build --tag bitbox-wallet-ci -f Dockerfile.travis .
+    docker run -i bitbox-wallet-ci bash \
+        -c "make -C \$GOPATH/src/github.com/digitalbitbox/bitbox-wallet-app ci" &
+    while [ -e /proc/$! ]; do echo -n "."  && sleep 60; done
+    docker run --privileged -i bitbox-wallet-ci bash \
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -41,7 +41,15 @@ osx-init:
 	export GOPATH=~/go/
 	export PATH=$PATH:~/go/bin
 	make init
-
+osx-init-travis:
```
benma: No need to have this in the Makefile - can just inline in the travis-ci.sh script?

jvaltin: OK. Revising and pushing.

scripts/travis-ci.sh
```diff
@@ -0,0 +1,28 @@
...
+    docker run -i bitbox-wallet-ci bash \
+        -c "make -C \$GOPATH/src/github.com/digitalbitbox/bitbox-wallet-app ci" &
+    while [ -e /proc/$! ]; do echo -n "."  && sleep 60; done
+    docker run --privileged -i bitbox-wallet-ci bash \
+        -c "make -C \$GOPATH/src/github.com/digitalbitbox/bitbox-wallet-app qt-linux"
+    set +e
+fi
+
+if [ "$TRAVIS_OS_NAME" == "osx" ]; then
+    brew install go
```
benma: replace all of this with `make osx-init`?

jvaltin: I think the first step would break as brew is already installed on the system.

scripts/travis-ci.sh
```diff
@@ -0,0 +1,28 @@
...
+    brew install yarn
+    brew install qt
+    export PATH="/usr/local/opt/qt/bin:$PATH"
+    export LDFLAGS="-L/usr/local/opt/qt/lib"
+    export CPPFLAGS="-I/usr/local/opt/qt/include"
+    export GOPATH=~/go/
+    export PATH=$PATH:~/go/bin
+    mkdir -p /Users/travis/go/src/github.com/digitalbitbox/
```
benma: `mkdir -p $GOPATH/src/github.com/digitalbitbox/`?

scripts/travis-ci.sh
benma: Could make it not fail somehow, or just remove it (make a small note in the README that brew is required).

jvaltin:
This is ready for merge and should enable Travis CI for MacOS building. The first commit fails and the second fixes the check script to not fail the build.

PR #37, opened 2018-08-28
Commit vendor

benma:


x1ddos:
@benma consider using go modules instead: https://golang.org/doc/go1.11#modules

PR #48, opened 2018-09-03
util/config: wrap appFolder init in sync.Once

x1ddos:
I underestimated quirks of CGO where an init func may be called
only after a package var is being accessed.

This does not currently apply to the BB app because appFolder
is accessed only from a Go function, i.e. no direct access from
C/C++ code via CGO.

Yet I didn't feel comfortable enough to let it be.
So, the proposed change serializes initialization of the appFolder
using sync.Once, the first time AppDir is called.

R: @benma

### DIFF
--- a/util/config/appdir.go
+++ b/util/config/appdir.go
@@ -18,15 +18,23 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"sync"
 )
 
 // appFolder is what AppDir always returns. It is initialized only once
-// in the init func.
+// in the initAppFolder func.
 //
-// Also useful to replace with a temp dir in tests.
+// It is tempting to initialize appFolder in an init func but because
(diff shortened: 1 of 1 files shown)


benma:
> I underestimated quirks of CGO where an init func may be called only after a package var is being accessed.

Wow! How did you realize that? Is that a bug, or documented behavior?

x1ddos:
I had a bad feeling about it for 2 weeks while travelling, so wrote a couple sample programs. Something didn't feel right. Then I found golang/go issue 11768 which only added up to my suspicions. I don't believe it's documented anywhere. 

There's also unrelated golang/go issue 26714 and [searching for c-shared](https://github.com/golang/go/issues?page=1&q=is%3Aissue+is%3Aopen+%22c-shared%22&utf8=%E2%9C%93) shows a few others. Just saying, keep an eye on those, especially when using certain compile/build flags /cc @jvaltin.

PR #49, opened 2018-09-04
ci: fix --fast flag

toxeus:
When migrating to the `golangci-lint` tool the fast linting
was removed from the Makefile. However, in `ci.sh` there's
still some dead code dealing with the `--fast` flag.

Since `golangci-lint` does support a `--fast` flag I've
opted to resurrect the flag in your build scripts.

Also, the unnecessary `else` case has been removed.

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -14,13 +14,6 @@
 # limitations under the License.
 
 
-# use `ci.sh --fast` to only run fast checkers.
-if [ "$1" == "--fast" ]; then
-    fast="--fast"
-else
-    fast=""
-fi
-
 go build ./...
 go test ./...
 golangci-lint run


scripts/ci.sh
```diff
@@ -17,10 +17,8 @@
...
 if [ "$1" == "--fast" ]; then
     fast="--fast"
-else
-    fast=""
 fi
 
 go build ./...
 go test ./...
-golangci-lint run
+golangci-lint run "$fast"
```
benma: CI fails because an empty argument like `golangci-lint run ""` somehow let's it think you want to lint no files at all (`level=error msg="Running error: context loading failed: no go files to analyze"`).

I would be okay dropping everything related to `--fast` too, as even the normal linting is very fast (10-15 seconds instead of 10 minutes with gometalinter).

toxeus: So the reason why it fails only on Linux is that it runs only on the Linux job I guess. I agree that if this tool is executing so fast then the fast flag doesn't make a lot of sense.

benma:
Thank you :grin:

PR #53, opened 2018-09-05
ci: also run javascript linter

benma:


thisconnect:
in `frontend/web`:

it seems to miss eslint, make sure `npm install`(or `yarn install`) did run first

would be great to have `npm list` (or `yarn list`) after the install to see the module tree

benma:
AppVeyor only builds the windows package, but does not run the tests/linters.

Makefile
```diff
@@ -52,6 +52,7 @@ generate:
 webdev:
 	make -C frontends/web dev
 weblint:
+	yarn --cwd=${WEBROOT} install
```
x1ddos: but you already added this command to scripts/ci.sh before running "make weblint", plus it is also run in "make generate", which you need to execute before any webdev, right.

maybe don't add yarn install here at all?

thisconnect: Could `envinit` be the place to do `yarn install`?

benma: No, `envinit` installs stuff to the host system (not good, should be improved some day). `yarn install` is local to the repo.

benma: @x1ddos good catch. Was a leftover :scream:

PR #55, opened 2018-09-06
Wizard rerebased

thisconnect:
in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/54/files


thisconnect:
@x1ddos I remove the comment, the warning message will be refactored into its own component in a different PR

frontends/web/src/i18n/en.json
```diff
@@ -418,6 +448,16 @@
             "notOK": "Backup does NOT match current wallet."
         }
     },
+    "success": {
+        "create": {
+            "title": "Success",
+            "info": "you successfully created a new wallet."
```
x1ddos: Title case? This is a full sentence and it's displayed at the center as such.

frontends/web/src/i18n/en.json
```diff
@@ -418,6 +448,16 @@
...
+        },
+        "restore": {
+            "title": "Success",
+            "info": "you successfully restored a wallet from a backup."
```
x1ddos: Title case? This is a full sentence and it's displayed at the center as such.

KasparEtter:
> I'd suggest displaying this only for new devices. That is, as soon as I insert a device and the app recognises a new one, show the warning. Otherwise, don't bother.

Ideally, the backend recognizes that a device, which has not been seen before, is already initialized and triggers the warning only then…

douglasbakkum:
@KasparEtter I guess that is not possible for a reset device after a replug (since there is no device specific info in ping cmd or usb descriptor)? Without a replug, would be possible.

KasparEtter:
> I guess that is not possible for a reset device after a replug (since there is no device specific info in ping cmd or usb descriptor)

We do have the serial number and the warning could also only be displayed once the user unlocked the BitBox…

PR #62, opened 2018-09-07
backend: suppress the error in the update handler

KasparEtter:
It is also triggered when the user is offline.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -249,7 +249,7 @@ func (handlers *Handlers) postOpenHandler(r *http.Request) (interface{}, error) 
 }
 
 func (handlers *Handlers) getUpdateHandler(_ *http.Request) (interface{}, error) {
-	return backend.CheckForUpdate()
+	return backend.CheckForUpdateIgnoringErrors(), nil
 }
 
 func (handlers *Handlers) getVersionHandler(_ *http.Request) (interface{}, error) {
--- a/backend/update.go
+++ b/backend/update.go
@@ -19,6 +19,7 @@ import (
 	"net/http"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/semver"
 )
 
@@ -64,4 +65,14 @@ func CheckForUpdate() (*UpdateFile, error) {
 
 	updateFile.CurrentVersion = Version
 	return &updateFile, nil
+}
+
+// CheckForUpdateIgnoringErrors suppresses any errors that are triggered, for example, when offline.
+func CheckForUpdateIgnoringErrors() *UpdateFile {
+	updateFile, err := CheckForUpdate()
+	if err != nil {
+		logging.Get().WithGroup("update").WithError(err).Warn("Check for update failed.")
+		return nil
+	}
+	return updateFile
 }


backend/handlers/handlers.go
```diff
@@ -249,7 +249,12 @@ func (handlers *Handlers) postOpenHandler(r *http.Request) (interface{}, error)
 }
 
 func (handlers *Handlers) getUpdateHandler(_ *http.Request) (interface{}, error) {
-	return backend.CheckForUpdate()
+	updateFile, err := backend.CheckForUpdate()
```
benma: change `CheckForUpdate()`, and return `backend.CheckForUpdate()` as before? I see no good reason to diverge on the http api vs the implementation.

KasparEtter: That was intentional. Maybe someone one day needs to be able to distinguish between no update available or failed to check whether update available. Admittedly, one could say the same about the frontend/Rest API, however, any error there triggers https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/utils/request.js#L43. Since we want more specific error messages in the future, we will have to overhaul that part anyway.

KasparEtter: And if you have strong feelings about this, I would simply add a second method in `update.go` so that the current variant still exists.

x1ddos: How about a middle group? Something like:

```go
update, err := backend.CheckForUpdate()
if err == backendErrors.ErrUnreachableOrSomethingLikeThat {
    err = nil
}
return update, err
```

benma: >  Maybe someone one day

This is where I always say: then let's do it then :) Most of the times, the future changes and the planned 'maybe we can use it' never materializes. 

My opinion is still that the http/rpc API should mirror the package API 1:1 as closely as possible (don't see a good reason to deviate in that layer at all). 

> I would simply add a second method in update.go so that the current variant still exists.

Then you have more code than is needed, and people looking at it would wonder why there are two, but only one is really used, etc.

x1ddos: hey @KasparEtter is there a bug for this? I mean, what's the use case, that it shows an annoying error in the UI? If so, what does the error say?

I think I can take this issue and try and make the backed report an "unreachable" error which would then result in some kind of a silent indicator like an offline icon or a status message in the UI. Unless you want to.

Ignoring the error might lead to an unforeseen bug-ache where the error has nothing to do with network but we'll never find out until it's uncomfortably too late or hard to pin point.

KasparEtter: As noted above, we get the generic frontend error message as an alert. I think suppressing (and logging) the errors is the right approach (at least for now) as update checks are not critical. Regarding offline status, I'd prefer to handle each coin individually and display the coin as "broken" in the sidebar. (It could also be that not the user but rather the (ElectrumX) backend is offline especially when self-hosted and this can be different for every coin.)

KasparEtter:
Happy now, @benma?

backend/handlers/handlers.go
x1ddos: I'm not sure I follow. `CheckForUpdate` doesn't seem to have anything to do with coins:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/3526014868441336b8d52b7c271559ca3fbb712c/backend/update.go#L46

```go
// CheckForUpdate checks whether a newer version of this application has been released.
// It returns the retrieved update file if a newer version has been released and nil otherwise.
func CheckForUpdate() (*UpdateFile, error) {
```

what am I missing?

x1ddos: The error could easily return a specific error invariant for when network isn't available, caught by the UI and display an "offline" indicator next to "check for updates" section.

KasparEtter: > CheckForUpdate doesn't seem to have anything to do with coins

Yes, that was my point. In case an update is available (and only then), a blue bar is displayed at the top of the application. There is no "check for updates" section otherwise. I think the GUI does not need a "generally offline" indicator (at least not for now) but a "coin cannot be synchronized" indicator would be nice but has nothing to do with this PR.

KasparEtter:
And what is the easiest way to retrigger AppVeyor? Its failure seems to have nothing to do with this PR.

backend/handlers/handlers.go
x1ddos: I don't think I ever meant anything like "generally offline". 
What I meant is:

> display an "offline" indicator next to "check for updates" section

So, why not return an error invariant which clearly states a network issue, and display it as such in the UI?

KasparEtter: Yes, that's possible. Has low priority in my opinion but I could do it (if Marko also prefers it).

x1ddos: I think it's especially good if it's a low prio. It's not like you need to release this tomorrow. So might as well do it "the right way"

Like I said, if you don't have time, I can take this.

KasparEtter: There's often no obvious right way. I think I'd prefer it if I'm not bothered with doing a manual update check once I'm back online. Rather, the update check should maybe be repeated from time to time if the user leaves the app open. (I could add this feature to the load decorator but this is then a different low-prio issue.) (Actually supporting a reasonable offline mode is way more important.)

x1ddos: I'd suggest refactor this PR into either different "check update" strategy, i.e. what you said about recurrent update checks, or supporting a generic offline mode. Ideally both, imho.

If you submit this as it right now, it won't have any effect until released anyway. So, why not take this as an opportunity and make it better for the next release. There's still time, right.

x1ddos: I was just thinking how one would implement an offline/online check on the backend, and one thing popped in my head: have a custom [roundtripper](https://golang.org/pkg/net/http/#RoundTripper) which, when failed to fetch a resource due to timeout or name resolution, indicates that with a specific error message or even updates some global state. This could then be used across the whole app for online/offline mode. Not sure, just an idea.

Then of course on the frontend you'd also use [offline/online events](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events), but I'm not entirely sure it is propagated via qt app.

KasparEtter: > This could then be used across the whole app for online/offline mode.

As mentioned above, I'd prefer a different approach, which I'm happy to discuss orally with anyone willing to actually implement it.

And instead of perfecting the offline behavior of the update check, we should rather persist the already synchronized transactions in an encrypted form and don't run in synchronization issues when offline. (This was the reason why I had to replace the transaction list with a spinner when offline.) Otherwise, the app doesn't do anything useful anyway when offline…

x1ddos: Sounds good. Let's talk offline. I'll also think about a periodic check for updates meanwhile. It actually feels more important that one might realize, imho.

benma:
I retriggered it.

Yes, thanks!

PR #63, opened 2018-09-07
frontend: refactor backend connected state from app.jsx to own component

KasparEtter:


thisconnect:
what is the best way to test this?
kill `$ make servewallet` ?

KasparEtter:
Yes, that's how I do it. Since the Qt app has no WebSocket, it never triggers the disconnected event. That's why I also made the error text a bit more specific.

frontends/web/src/connected.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { App } from './app';
+
+interface Props {
+    connected?: boolean;
+}
+
+function ConnectedApp({ connected = true }: RenderableProps<Props>): JSX.Element {
+    return connected ? <App /> : (
+        <div className="app" style="padding: 40px">
+            The WebSocket closed. Please restart the backend and reload this page.
```
x1ddos: Is this for a dev-only version? Otherwise, I don't thing any user will understand what "websocket closed" and "restart the backedn" mean.

I'd leave it the way it was, "An error occurred. Please restart the application.".

KasparEtter: > Since the Qt app has no WebSocket, it never triggers the disconnected event. That's why I also made the error text a bit more specific.

x1ddos: Sure, but what about non-qt app? You don't have any today, but about tomorrow? Tomorrow you'll probably need to go back and change it back. So, why not leave it as is for now and maybe change it later when you really need to?

KasparEtter: At that point one should internationalize the string anyway…

x1ddos: Sure but it doesn't have to be in this PR. This one is titled "refactor backend connected state from app.jsx to own component", so why should you care and change the actual wording or add translation support.

Anyway, I'm not insisting. It was just a suggestion.

KasparEtter: Yes, I see this, thanks. I just thought it would improve the clarity of both the code and the error message when faced with it as a new developer. (I could also have done the former with a comment, I know.)

KasparEtter:
Can someone approve this PR (or do you all want me to change the message back)? Thanks!

PR #64, opened 2018-09-07
frontend: caps lock detection without event.getModifierState('CapsLock')

thisconnect:
for some reason event.getModifierState('CapsLock') does not work in Qt

I do not prefer this change if possible as the caps indication will be delayed to the next event instead of appearing immediately

### DIFF
--- a/frontends/web/src/components/password.jsx
+++ b/frontends/web/src/components/password.jsx
@@ -41,8 +41,7 @@ export class PasswordSingleInput extends Component {
     }
 
     componentWillMount() {
-        document.addEventListener('keydown', this.handleCheckCaps);
-        document.addEventListener('keyup', this.handleCheckCaps);
+        window.addEventListener('keydown', this.handleCheckCaps);
     }
 
     componentDidMount() {
@@ -52,8 +51,7 @@ export class PasswordSingleInput extends Component {
     }
 
     componentWillUnmount() {
-        document.removeEventListener('keydown', this.handleCheckCaps);
-        document.removeEventListener('keyup', this.handleCheckCaps);
+        window.removeEventListener('keydown', this.handleCheckCaps);
     }
 
     tryPaste = event => {
@@ -95,8 +93,10 @@ export class PasswordSingleInput extends Component {
     }
 
     handleCheckCaps = event => {
-        const capsLock = event.getModifierState && event.getModifierState('CapsLock');
-        this.setState({ capsLock });
+        const capsLock = hasCaps(event);
+        if (capsLock !== null) {
+            this.setState({ capsLock });
+        }
     }
 
     render({
@@ -129,8 +129,6 @@ export class PasswordSingleInput extends Component {
                     placeholder={placeholder}
                     onInput={this.handleFormChange}
                     onPaste={this.tryPaste}
-                    onKeyUp={this.handleCheckCaps}
-                    onKeyDown={this.handleCheckCaps}
                     getRef={ref => this.password = ref}
                     value={password}>
                     {warning}
@@ -164,8 +162,7 @@ export class PasswordRepeatInput extends Component {
     }
 
     componentWillMount() {
-        document.addEventListener('keydown', this.handleCheckCaps);
-        document.addEventListener('keyup', this.handleCheckCaps);
+        window.addEventListener('keydown', this.handleCheckCaps);
     }
 
     componentDidMount() {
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/password.jsx
```diff
@@ -309,3 +304,8 @@ function MatchesPattern({ regex, value = '', text }) {
         <p style="color: var(--color-error);">{text}</p>
     );
 }
+
+function hasCaps(event) {
+    const char = String.fromCharCode(event.which);
```
x1ddos: Isn't `.which` is a deprecated property and you should use something else like `.code`?
See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which and https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

frontends/web/src/components/password.jsx
```diff
@@ -309,3 +304,8 @@ function MatchesPattern({ regex, value = '', text }) {
...
+    return char.toUpperCase() === char && char.toLowerCase() !== char && !event.shiftKey;
```
x1ddos: Can it actually happen that `char.toUpperCase() === char` and `char.toLowerCase() === char` at the same time?

x1ddos: ehm.. sorry, never mind this.

thisconnect:
I wasn't aware that event.which is deprecated (who would have thought?). Squash force pushed.

The workaround only detects when the user presses case-sensitive keys, iow we can't show rightaway when capslock has be activated :(

x1ddos:
Let my try it on linux...

frontends/web/src/components/password.jsx
```diff
@@ -129,8 +129,7 @@ export class PasswordSingleInput extends Component {
                     placeholder={placeholder}
                     onInput={this.handleFormChange}
                     onPaste={this.tryPaste}
-                    onKeyUp={this.handleCheckCaps}
-                    onKeyDown={this.handleCheckCaps}
+                    onKeyPress={this.handleCheckCaps}
```
x1ddos: j/w how's this different from ` window.addEventListener('keydown', this.handleCheckCaps);` above in `componentWillMount`. and if it's the same, why are there two different listeners. shouldn't one be enough?

x1ddos: Same question for `PasswordRepeatInput` below.

frontends/web/src/components/password.jsx
```diff
@@ -309,3 +306,13 @@ function MatchesPattern({ regex, value = '', text }) {
         <p style="color: var(--color-error);">{text}</p>
     );
 }
+
+const excludeKeys = /^(Shift|Alt|Backspace|CapsLock|Tab)$/i;
+
+function hasCaps({ key }) {
+    if (key.length > 1 || key.toUpperCase() === key.toLowerCase() || excludeKeys.test(key)) {
+        return null;
```
benma: Return `false` instead of `null`? Then the caller can just `setState()` it without a null check.

frontends/web/src/components/password.jsx
thisconnect: One on window (or document.body) would be practically enough.  In theory somebody could use the component, listen for the same event and cancel it bubbling up. In this case we would not get any event on window (or document.body).

x1ddos: I see. So the solution is to double-listen, on both `window` and the input field? :) There *has* to be a better way!
Why wouldn't listening only on the input field not be enough? I mean, why not remove `window.addEventListener('keydown', this.handleCheckCaps)` altogether.

frontends/web/src/components/password.jsx
x1ddos: Seems like it's a tri-state return value?
Maybe `hasCaps` func needs a little doc comment, describing what it returns in which case.

frontends/web/src/components/password.jsx
thisconnect: > Why wouldn't listening only on the input field not be enough?

because if the element has no focus it would not catch a capslock change, ideally it should always catch a capslock change EVEN when the app does not have focus, but that is not possible with JavaScript only.

x1ddos: but we're not listening for the caps lock change anymore, are we? you even have a comment saying that. so, seems redundant.

thisconnect: we try to catch every KeyboardEvent and try to if capslock is active. This would be no problem if `getModifierState` would work correctly, but in Qt it always return `false`

- https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
- https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState

PR #65, opened 2018-09-09
frontend: add support for testing the UI

x1ddos:
The tests are run using [jest](https://jestjs.io) and [ts-jest](https://www.npmjs.com/package/ts-jest) preprocessor.

Because the app is based on [preact](https://preactjs.com), we use [preact-render-spy](https://www.npmjs.com/package/preact-render-spy) package instead of [enzyme](https://airbnb.io/enzyme/) to test app components rendering and their state.

To prove that the setup works, I randomly picked a couple simple modules and wrote a few tests for them.

```
$ make webtest
make -C frontends/web jstest
make[1]: Entering directory 'bitbox-wallet-app/frontends/web'
yarn run test
yarn run v1.9.4
$ jest
 PASS  test/components/message/message.test.tsx
 PASS  test/utils/equal.test.tsx

Test Suites: 2 passed, 2 total
Tests:       22 passed, 22 total
Snapshots:   1 passed, 1 total
Time:        1.067s
Ran all test suites.
Done in 2.33s.
make[1]: Leaving directory 'bitbox-wallet-app/frontends/web'
```

See changes in README.md for more details overall.

R: @thisconnect @originalblend337 @KasparEtter @benma


frontends/web/Makefile
```diff
@@ -16,6 +16,12 @@ dev:
 	yarn run dev
 lint:
 	yarn run lint
+jstest:
+	yarn run test
+jstest-watch:
+	yarn run test --watch
+jstest-cover:
+	yarn run test --coverage
```
thisconnect: one has to `cd frontends/web` to be able to use `make jstest-watch` or can you call from project root?

in this case I would not add it to Makefile but rather to package.json scripts, maybe similar to  https://github.com/thisconnect/bitbox-wallet-app/blob/frontend/tests/frontends/web/package.json#L10

x1ddos: Sure, but moving these to package.json would still require to `cd frontends/web`. The reason being that this is a specific task and you won't need it most of the time, unless you're a heavy web developer and spend most of your time in frontends/web anyway.

It just seemed to be more consistent with the existing commands: everything is run via `make <something>` so did that.

x1ddos: Don't you like explicit flags to understand what's actually going to happen? :)

frontends/web/jest.config.js
```diff
@@ -0,0 +1,227 @@
...
+
+    // An array of directory names to be searched recursively up from the requiring module's location
+    // moduleDirectories: [
+    //   "node_modules"
+    // ],
+
+    // An array of file extensions the project modules use.
+    moduleFileExtensions: [
+        'js',
+        'json',
```
KasparEtter: My understanding of what we do here is quite limited but why do we include `json`? And shouldn't you add `ts`?

frontends/web/package.json
```diff
@@ -16,14 +17,19 @@
     "build/*"
   ],
   "devDependencies": {
+    "@types/jest": "^23.3.2",
```
KasparEtter: Again not my area of expertise but I guess/thought we want to fix the versions of our dependencies by removing the `^`.

frontends/web/Makefile
thisconnect: I think it is fine and don't mind at all. 👍 

> Don't you like explicit flags to understand what's actually going to happen

to understand what will happen when calling `npm test`/`yarn test` there is only 1 thing you can do... read the package.json (rtfp is the new rtfm) :)

frontends/web/test/components/message/message.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+import { h } from 'preact';
+import { shallow, deep } from 'preact-render-spy';
+
+import Message from '../../../src/components/message/message';
+
+describe('components/message/message', () => {
+  it('should use type attr as CSS class', () => {
```
KasparEtter: Doesn't this indentation give you lint errors?

frontends/web/package.json
KasparEtter: And don't forget to also commit an updated yarn.lock then.

frontends/web/jest.config.js
x1ddos: `json` is for making i18n's translations available as modules during tests.
re: `.ts`. I don't believe we have any .ts files, so didn't add it. is the plan to also switch to plain typescript? if so, need to add it here. otherwise, I don't see why.

frontends/web/test/components/message/message.test.tsx
x1ddos: nope. `make weblint` didn't complain. I guess it's because it only runs for `frontends/web/src/`. I guess I can add `frontends/web/test/` to the lint as well?

frontends/web/jest.config.js
x1ddos: I have to admin though, this is a bit inconsistent with the extensions listed in `collectCoverageFrom` above.

x1ddos: Actually, you're right. We do use .ts files: the generated xxx.d.ts from CSS files.
Will add `ts` to this list.

x1ddos: done

frontends/web/test/components/message/message.test.tsx
x1ddos: done

frontends/web/package.json
x1ddos: done

frontends/web/test/components/message/message.test.tsx
x1ddos: I meant: my test files to use 4 spaces indent.

x1ddos:
PTAL

x1ddos:
> Thanks for the changes, I'm also happy now!

@KasparEtter feel free to merge :)

PR #66, opened 2018-09-10
frontend: disable uglify plugin and enable modern js build

thisconnect:
I keep `"checkJs": false` for the moment to see if the build succeeds


frontends/web/tsconfig.json
```diff
@@ -1,14 +1,16 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "es2017",
```
x1ddos: So, what's the difference... or rather, what are the advantages for bb app to switch the compiler target?

thisconnect: One hope was that the build has real `class`, or that babel does not need to transpile `Object.assign` back to es5. Chromium should have close to 100% ES6+ compatibility. Last big open issue afaik was native `import` and `import()`. Qt chromium is probably a bit behind would be nice to have more info about that.

Another reason to up target was related to typescript
TypeScript complains about methods not existing on strings and arrays. 

btw.
The huge step in JavaScript was ES5 -> ES6 (which was renamed to es2015). From then  TC39 only included small stuff, afaik Array.prototype.includes, and in ES2017 async functions (I had to look that one up)

x1ddos: Cool. Although, "hope" and "probably" are not very reassuring. I was hoping for a bit more affirmative statements. :)

x1ddos:
So, apart from CI passing, have you tried building the whole Qt app. Does it still work?

frontends/web/tsconfig.json
```diff
@@ -1,14 +1,16 @@
...
         "moduleResolution": "node",
         "sourceMap": true,
         "allowJs": true,
+        "checkJs": false,
         "jsx": "react",
         "jsxFactory": "h",
         "strictNullChecks": true,
-        "experimentalDecorators": true
+        "experimentalDecorators": true,
+        "removeComments": true,
```
KasparEtter: Did I add too many comments to my TypeScript files? 🙂 You added this just to remove the file size a bit? What is the disadvantage? Are the comments then also missing in the mapped source?

frontends/web/preact.config.js
```diff
@@ -30,8 +30,20 @@ export default function (config, env, helpers) {
...
 
-    if (env.production) {
-        helpers.getPluginsByName(config, 'UglifyJsPlugin')[0].plugin.options.sourceMap = false;
+    {
+        // disable SWPrecacheWebpackPlugin
+        const SWPrecacheWebpackPlugin = helpers.getPluginsByName(config, 'SWPrecacheWebpackPlugin')[0];
+        if (SWPrecacheWebpackPlugin) {
+            config.plugins.splice(SWPrecacheWebpackPlugin.index, 1);
+        }
+    }
```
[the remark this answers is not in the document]
KasparEtter: How would/could we have used it? I didn't really study https://www.npmjs.com/package/sw-precache-webpack-plugin.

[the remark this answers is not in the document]
thisconnect: we disabled it in the production build 3 month ago https://github.com/digitalbitbox/bitbox-wallet-app/commit/8ebc8dc866f13160d4ec1b821ab248b39851e21e

frontends/web/package.json
```diff
@@ -33,7 +33,6 @@
     "react-i18next": "7.9.0"
   },
   "browserslist": [
-    "iOS >= 3.2",
-    "Safari >= 3.2"
+    "last 2 Chrome versions"
```
KasparEtter: I have never seen this – which doesn't mean anything, of course. 😆

frontends/web/tsconfig.json
thisconnect: They are deliberately chosen, the build still does NOT use native `class`.

thisconnect:
> So, apart from CI passing, have you tried building the whole Qt app. Does it still work?

yes, but there is indeed an issue on the transaction page, thanks for making me test again :)

```
js: Uncaught (in promise) TypeError: Cannot read property 'CHF' of undefined
```

frontends/web/tsconfig.json
thisconnect: yes as we currently have no minifier I expected this to remove the comments, but there still seem to be many comments. I think I just remove this again. No the sourcemaps would still point to the source code with comments (OR ELSE!).

frontends/web/tsconfig.json
thisconnect: > Qt chromium is probably a bit behind would be nice to have more info about that.

I'd love more info about that and shall gather info about chromium used in the Qt webengine

PR #67, opened 2018-09-10
Coins

benma:


backend/coins/btc/account.go
```diff
@@ -580,7 +581,7 @@ func (account *Account) ConvertToLegacyAddress(scriptHashHex blockchain.ScriptHa
 	if address == nil {
 		return nil, errp.New("unknown address not found")
 	}
-	if account.Coin().Net() != &ltc.MainNetParams || address.Configuration.ScriptType() != signing.ScriptTypeP2WPKHP2SH {
+	if account.coin.Net() != &ltc.MainNetParams || address.Configuration.ScriptType() != signing.ScriptTypeP2WPKHP2SH {
```
x1ddos: so, in here you use `account.coin` but in `func (account *Account) String() string {` above you do `account.Coin().Name()` ... any difference in accessing the same thing with `account.coin` vs `account.Coin()`?

x1ddos: I guess none. It just feels strange that you use both in different methods.

benma: I usually treat `String()` as formatting for logs, debug output etc., that can change without breaking anything (easier to put it in a sprintf for `%s` without thinking which method to call).

 `Name()` can't change though without having consequences.

benma: The difference is that one is the interface, the other is the concrete thing. Here, `Net()` is specific to the `btc` implementation, and not available on `Coin()`, which returns the interface.

x1ddos: Ah ok. Makes sense.

PR #68, opened 2018-09-11
account: move json business to handlers

benma:
- It is used in the http layer and defines the http API, so the
  definition should be local
- Other coins shouldn't reimplement the same function - reuse for
  other coins using the account interface.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -15,7 +15,6 @@
 package btc
 
 import (
-	"encoding/json"
 	"fmt"
 	"path"
 	"sort"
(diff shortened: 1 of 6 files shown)


backend/coins/btc/account.go
```diff
@@ -52,6 +51,8 @@ type Interface interface {
 	Info() *Info
 	Code() string
 	Coin() coin.Coin
+	// Name returns a human readable long name.
+	Name() string
```
x1ddos: It would be nice to also add doc comment to the `Code` and explain how it's different from name. 

You already say that name is a "human readable long name" but what is code then, a "human readable short name" or a "machine-readable code or something else?

x1ddos: At least, to me as an external viewer code and name seem to be related but I'm not sure what the difference is.

benma: Thx. Done!

PR #69, opened 2018-09-11
coins: remove dead code related to rates

benma:
Rates are fetched on demand from the /coins/rates/ endpoint.


backend/coins/btc/coin.go
```diff
@@ -130,41 +129,16 @@ func (coin *Coin) Unit() string {
...
-
-func formatAsCurrency(amount float64) string {
-	formatted := strconv.FormatFloat(amount, 'f', 2, 64)
-	position := strings.Index(formatted, ".") - 3
-	for position > 0 {
-		formatted = formatted[:position] + "'" + formatted[position:]
-		position = position - 3
-	}
-	return formatted
+	formattedAmount := coin.FormatAmountAsJSON(amount)
```
x1ddos: Maybe move JSON stuff into handlers like you did in a previous change? That looked like a good call.

benma: Great idea. This made me realize that FormatAmount() was completely unused, too. 

Pushed a 2nd commit.

backend/coins/btc/coin.go
```diff
@@ -130,42 +129,8 @@ func (coin *Coin) Unit() string {
...
-				conversions[key] = formatAsCurrency(float * value)
-			}
-		}
-	}
-	return coinpkg.FormattedAmount{
-		Amount:      strconv.FormatFloat(float, 'f', -int(btcutil.AmountBTC+8), 64),
-		Unit:        coin.Unit(),
-		Conversions: conversions,
-	}
+	return strconv.FormatFloat(float, 'f', -int(btcutil.AmountBTC+8), 64)
```
x1ddos: I'd use something like `btc` because `float` is too close to the `float32` and `float64` types.

x1ddos: Well, maybe not `btc` because it's the package name. Maybe `a` is best but I know you won't like it.

PR #70, opened 2018-09-11
btc/account: use address interface

benma:
Another step towards a coin agnostic interface.


backend/coins/btc/account.go
```diff
@@ -541,19 +541,24 @@ func (account *Account) Transactions() []*transactions.TxInfo {
 }
 
 // GetUnusedReceiveAddresses returns a number of unused addresses.
-func (account *Account) GetUnusedReceiveAddresses() []*addresses.AccountAddress {
+func (account *Account) GetUnusedReceiveAddresses() []coin.Address {
 	account.synchronizer.WaitSynchronized()
 	defer account.RLock()()
 	account.log.Debug("Get unused receive address")
+	addresses := []coin.Address{}
```
x1ddos: You could also leverage zero value:

```go
var addresses []coin.Address
```

PR #72, opened 2018-09-21
Frontend mobile compatibility

originalblend337:
Will be rebasing and cleaning up commits


benma:
Please fix the eslint issues: https://travis-ci.org/digitalbitbox/bitbox-wallet-app/jobs/432342492

(run locally with `make weblint`)

KasparEtter:
I don't know whether you already saw this but the linting still fails:
```
$ eslint --cache --ignore-path ../../.gitignore '**/*.{js,jsx}'
/opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/frontends/web/src/components/container/Container.jsx
  27:59  error  Expected property shorthand  object-shorthand
/opt/go/src/github.com/digitalbitbox/bitbox-wallet-app/frontends/web/src/routes/device/manage-backups/manage-backups.jsx
  17:13  error  'Component' is defined but never used. Allowed unused vars must match /^h$/  no-unused-vars
  46:2   error  Missing semicolon                                                            semi
✖ 3 problems (3 errors, 0 warnings)
```

benma:
Thanks :tada:  I haven't taken a deep look yet, but I noticed two things:

- the blocks synced progress text is not visible if there are transactions.
- the balance header is not the same in the tx screen and the send screen (dimensions)

I don't know if it because I am used to it, but I prefer the old header with the big balance better. Kind of not digging the blue header. Is it possible to keep the bigger header for bigger screens, and collapse it like this only for mobile?

The Guide `?` being blue before was nice because the Guide is also blue. Also preferred how there was only one button to open/close the Guide that staid in one location (now you have to ways to close the Guide).

@thisconnect what do you think?

thisconnect:
> The Guide ? being blue before was nice because the Guide is also blue. Also preferred how there was only one button to open/close the Guide that staid in one location (now you have to ways to close the Guide).

@benma I have not tested and checked out the PR, but from what I read, here inputs
- use the full screen for navigation and the guide on smaller screens
- opening / closing should be located at the same place
- determine what the smallest usable window-width is before collapsing and enabling fullscreen guide/navigation
- use headers (dimensions) in send and tx screens

originalblend337:
@benma -- doing a quick rerun on the header

originalblend337:
screen shot 2018-09-26 at 18 39 34
screen shot 2018-09-26 at 18 40 12

originalblend337:
@thisconnect -- although I did make the sidebar wider with larger text on small screens, I didn't go all the way and make it full screen since I thought the sidebar doesn't have enough content such as account balance under each label. Let's make it full screen when we populate the sidebar more in the near future?

frontends/web/src/routes/account/send/send.jsx
```diff
@@ -279,6 +281,10 @@ export default class Send extends Component {
         return this.props.accounts.find(({ code }) => code === this.props.code);
     }
 
+    toggleCoinControl = () => {
+        this.setState({ activeCoinControl: !this.state.activeCoinControl });
```
thisconnect: please pass callback when depending on previous state

PR #76, opened 2018-09-29
btc: remove unused log argument

benma:


backend/coins/btc/handlers/handlers.go
```diff
@@ -224,7 +223,7 @@ func (input *sendTxInput) UnmarshalJSON(jsonBytes []byte) error {
 }
 
 func (handlers *Handlers) postAccountSendTx(r *http.Request) (interface{}, error) {
-	input := &sendTxInput{log: handlers.log}
+	input := &sendTxInput{}
```
x1ddos: would `var input sendTxInput` work? More idiomatic.

benma: Yes. Double pointer even on the next line :/. See #77

PR #78, opened 2018-09-30
Frontend disable "set recovery password" button if there is no micro SD card

thisconnect:


thisconnect:
need 2 different message depending on you create a new seed or restore from a backup

frontends/web/src/i18n/en.json
```diff
@@ -346,12 +346,12 @@
...
         "create": "Create Wallet",
         "creating": "Creating wallet",
-        "creating2": "Backing up wallet to micro SD Card",
+        "creating2": "Backing up wallet to micro SD card",
         "backup": "Or restore a Backup…",
         "backToCreate": "Or create a new wallet…",
         "error" : {
             "102": "The password must consist of at least 8 characters.",
-            "200": "Wallet creation requires an micro SD card for automatic backup."
+            "200": "You need to insert the micro SD card in your BitBox to create a wallet, so a backup can be created automatically."
```
benma: `You need to insert a micro SD card into your BitBox in order to create a wallet, so a backup can be created automatically.`.

frontends/web/src/routes/device/setup/seed-create-new.jsx
```diff
@@ -153,7 +153,10 @@ export default class SeedCreateNew extends Component {
                         onClick={goBack}>
                         {t('button.back')}
                     </Button>
-                    <Button primary onClick={this.handleStart}>
+                    <Button
+                        primary
+                        onClick={this.handleStart}
```
benma: If you can click the button before it is disabed, `checkSDcard()` is called and starts a new timer-loop, even though one is already running.

frontends/web/src/i18n/en.json
thisconnect: fixed and squashed into "improve missing sd card error message" commit

frontends/web/src/routes/device/setup/seed-create-new.jsx
thisconnect: fixed in 44c5acfa2d66bf64f9a605bde5c65c45e85c6436

PR #79, opened 2018-09-30
Amounts

benma:


backend/coins/coin/amount.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"math/big"
+	"strings"
+
+	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
+)
+
+// Amount represents an amount in the smallest coin unit (e.g. satoshi).
+type Amount struct {
+	*big.Int
```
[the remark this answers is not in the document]
benma: @x1ddos 

the reason i put it into the struct was to avoid awkward conversions. in the case of `type Amount big.Int`, converting `*big.Int` to `*Amount` seems impossible without a copy (or am i missing sth?):

```
func NewAmount(amount *big.Int) *Amount {
	tmp := Amount(*amount)
	return &tmp
}
```

In the latter case, I would need to clutter my code with `&amount.Int` to extract `*big.Int`.

>  For instance, when passing a Amount to a func, I would expect a value to be copied. So, whatever happens to a in a function would leave the caller's value in the original state.

Could be mitigated with passing `*Amount` instead?

.golangci.yml
```diff
@@ -77,3 +77,4 @@ issues:
     - G101
     # gas: Duplicated errcheck checks
     - G104
+  exclude-use-default: false
```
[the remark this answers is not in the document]
benma: Done.

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// NewAmountFromInt64 creates a new amount.
+func NewAmountFromInt64(amount int64) Amount {
+	return Amount{Int: big.NewInt(amount)}
+}
+
+// NewAmountFromString parses a user given coin amount, converting it from the default coin unit to
+// the the smallest unit.
+func NewAmountFromString(s string, unit *big.Int) (Amount, error) {
```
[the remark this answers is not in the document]
benma: Did the third variant with a private immutable member.

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,55 @@
...
+	// big.Rat parsing accepts rationals like "2/3". Exclude those, we only want decimals.
+	if strings.ContainsRune(s, '/') {
+		return Amount{}, errp.Newf("could not parse %r", s)
```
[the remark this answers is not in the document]
benma: Good catch! Fixed.

backend/coins/btc/transaction.go
```diff
@@ -23,33 +25,56 @@ import (
...
+// = 1e8). Returns an error if the amount is not positive.
+func (sendAmount *SendAmount) Amount(unit *big.Int) (coin.Amount, error) {
+	if sendAmount.sendAll {
+		panic("can only be called if SendAll is false")
+	}
+	amount, err := coin.NewAmountFromString(sendAmount.amount, unit)
+	if err != nil {
+		return coin.Amount{}, errp.WithStack(TxValidationError("invalid amount"))
+	}
+	if amount.Int().Cmp(big.NewInt(0)) <= 0 {
```
x1ddos: You could also extract this into a helper method to avoid exposing big int, something like:

```go
func (a *Amount) Negative() bool {
    return a.amount.Sign() == -1
}

func (a *Amount) Zero() bool {
  return a.amount.Sign() == 0
}
```

And then you'd use it like:

```go
if amount.Negative() || amount.Zero() {
  return ..., err
}
```

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Int64 returns the int64 representation of amount.
+// If x cannot be represented in an int64, the result is undefined.
+func (amount *Amount) Int64() int64 {
+	return amount.amount.Int64()
+}
+
+// Int returns a copy of the underlying big integer.
+func (amount *Amount) Int() *big.Int {
```
x1ddos: I would try and avoid exposing big int due to the same concerns - it's mutable.

x1ddos: ah sorry, it returns a copy. still, helper methods like above for "negative" and "zero" might be a more readable solition?

benma: I guess, but on the other hand, it's the stdlib, so people should be able to use it. Not sure if it's a good idea to wrap every underlying functionality I need instead of just giving access.

For example, in other coins, big.Int could be the basis for calculations, and wrapping all the arithmetic functions is overkill?

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,66 @@
...
+	rat.Mul(rat, new(big.Rat).SetInt(unit))
+	if rat.Denom().Cmp(big.NewInt(1)) != 0 {
+		return Amount{}, errp.Newf("invalid amount %q", s)
+	}
+	return Amount{amount: rat.Num()}, nil
+}
+
+// Int64 returns the int64 representation of amount.
+// If x cannot be represented in an int64, the result is undefined.
+func (amount *Amount) Int64() int64 {
```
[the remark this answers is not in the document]
benma: Something I wish the stdlib would do, too ;)

Will fix.

backend/coins/coin/amount.go
x1ddos: Yeah, that's a downside. I haven't checked but are you really doing some math with the amount? I thought this was just to parse user input and use the value as is, no arithmetic operations.

Either way, I'd suggest renaming the method to `BigInt`. You already have `Int64` which returns `int64` so first thing that comes in mind is `Int()` returns the `int` type unless I go and look up the signature.

backend/coins/btc/transaction.go
x1ddos: Also, you want to go without helpers, you could still simplify this with `amount.Sign() <= 0`, right?

https://golang.org/pkg/math/big/#Int.Sign

backend/coins/coin/amount.go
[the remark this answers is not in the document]
benma: Done (pushed new commit)

backend/coins/btc/transaction.go
benma: Right! Done.

backend/coins/btc/handlers/handlers.go
```diff
@@ -165,9 +173,9 @@ func (handlers *Handlers) getUTXOs(_ *http.Request) (interface{}, error) {
 func (handlers *Handlers) getAccountBalance(_ *http.Request) (interface{}, error) {
 	balance := handlers.account.Balance()
 	return map[string]interface{}{
-		"available":   handlers.formatAmountAsJSON(int64(balance.Available)),
-		"incoming":    handlers.formatAmountAsJSON(int64(balance.Incoming)),
-		"hasIncoming": balance.Incoming != 0,
+		"available":   handlers.formatAmountAsJSON(balance.Available),
+		"incoming":    handlers.formatAmountAsJSON(balance.Incoming),
+		"hasIncoming": balance.Incoming.Int().Cmp(big.NewInt(0)) != 0,
```
x1ddos: I guess you could use the same trick here: `balance.Incoming.BigInt().Sign() > 0`.
Also, when the comparison is `-1`, is it indented that `hasIncoming` is `true`? I mean, is it possible for `Incoming` to be negative? I guess not but if for some reason there's a regression bug and it is indeed negative, it'll show `hasIncoming` as `true`.

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,70 @@
...
+	if rat.Denom().Cmp(big.NewInt(1)) != 0 {
+		return Amount{}, errp.Newf("invalid amount %q", s)
+	}
+	return Amount{amount: rat.Num()}, nil
+}
+
+// Int64 returns the int64 representation of amount. If x cannot be represented in an int64, an
+// error is returned.
+func (amount Amount) Int64() (int64, error) {
+	if amount.amount.Cmp(big.NewInt(math.MaxInt64)) == 1 || amount.amount.Cmp(big.NewInt(math.MinInt64)) == -1 {
```
x1ddos: Wouldn't it be simpler to use `amount.IsInt64()`?
https://golang.org/pkg/math/big/#Int.IsInt64

> IsInt64 reports whether x can be represented as an int64.

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,70 @@
...
+		return 0, errp.Newf("%s overflows int64", amount.amount)
+	}
+	return amount.amount.Int64(), nil
+}
+
+// Int returns a copy of the underlying big integer.
+func (amount Amount) Int() *big.Int {
```
x1ddos: I'd still rename this to `BigInt()` but not a big deal.

backend/coins/coin/amount.go
```diff
@@ -0,0 +1,66 @@
...
+	if rat.Denom().Cmp(big.NewInt(1)) != 0 {
+		return Amount{}, errp.Newf("invalid amount %q", s)
+	}
+	return Amount{amount: rat.Num()}, nil
+}
+
+// Int64 returns the int64 representation of amount.
+// If x cannot be represented in an int64, the result is undefined.
+func (amount *Amount) Int64() int64 {
+	return amount.amount.Int64()
```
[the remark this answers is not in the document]
benma: Done.

backend/coins/coin/amount.go
benma: I use it when formatting the amount (could be a struct method too), and Ethereum will use big.Int internally, so arithmetics will be done there (i.e. adding/subtracting fees, etc.).

backend/coins/coin/amount.go
benma: Done.

benma:
@x1ddos thanks for the thorough review!

ferdzbautistaolazo:
Thanks a lot.

PR #81, opened 2018-10-01
coins: move SendAmount, TxValidationError out btc

benma:
Move out of the btc package, make available for all coins.


benma:
Ty.

PR #82, opened 2018-10-01
frontend/receive: null validation bug

benma:
Could not verify/convert the first address anymore.

### DIFF
--- a/frontends/web/src/routes/account/receive/receive.jsx
+++ b/frontends/web/src/routes/account/receive/receive.jsx
@@ -65,9 +65,9 @@ export default class Receive extends Component {
     }
 
     verifyAddress = () => {
-        this.setState({ verifying: true });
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/account/receive/receive.jsx
```diff
@@ -65,9 +65,9 @@ export default class Receive extends Component {
     }
 
     verifyAddress = () => {
-        this.setState({ verifying: true });
         const { receiveAddresses, activeIndex } = this.state;
         if (receiveAddresses && activeIndex) {
```
KasparEtter: Shouldn't we use the same fix here? `if (receiveAddresses !== null && activeIndex !== null)`

benma: I could swear I have done that :laughing: Good catch. Done.

PR #83, opened 2018-10-01
btc/handlers: uncouple from bitbox

benma:
Enforce separation.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -60,6 +60,8 @@ type Interface interface {
 	Close()
 	Transactions() []*transactions.TxInfo
 	Balance() *transactions.Balance
+	// Creates, signs and broadcasts a transaction. Returns keystore.ErrSigningAborted on user
+	// abort.
 	SendTx(string, coin.SendAmount, FeeTargetCode, map[wire.OutPoint]struct{}) error
 	FeeTargets() ([]*FeeTarget, FeeTargetCode)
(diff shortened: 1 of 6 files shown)


backend/coins/btc/account.go
```diff
@@ -60,6 +60,8 @@ type Interface interface {
 	Close()
 	Transactions() []*transactions.TxInfo
 	Balance() *transactions.Balance
+	// Creates, signs and broadcasts a transaction. Returns `keystore.ErrSigningAborted` on user
+	// abort.
 	SendTx(string, coin.SendAmount, FeeTargetCode, map[wire.OutPoint]struct{}) error
```
x1ddos: "SendTx creates, signs and ..." and without backticks "`" - they are ignored in go doc.

backend/keystore/keystores.go
```diff
@@ -39,7 +39,8 @@ type Keystores interface {
 	// keystores that have a secure output.
 	OutputAddress(*signing.Configuration, coin.Coin) error
 
-	// SignTransaction signs the given proposed transaction on all keystores.
+	// SignTransaction signs the given proposed transaction on all keystores. Returns
+	// `ErrSigningAborted` if the user aborts.
```
x1ddos: backticks "`" are ignored by go doc. It's not recommended to use in doc comments. The common opinion is that they make the comments less readable.

benma: Good to know! Done.

PR #88, opened 2018-10-02
frontend: add headers to waiting and unlock component / use existing …

originalblend337:
…subheader containers in setup views

### DIFF
--- a/frontends/web/src/routes/device/unlock.jsx
+++ b/frontends/web/src/routes/device/unlock.jsx
@@ -22,6 +22,7 @@ import { Button, Input } from '../../components/forms';
 import { Message } from '../../components/message/message';
 import { BitBox, Shift } from '../../components/icon/logo';
 import { Guide } from '../../components/guide/guide';
+import Header from '../../components/header/Header';
 import Footer from '../../components/footer/footer';
 import Spinner from '../../components/spinner/Spinner';
 import * as style from './device.css';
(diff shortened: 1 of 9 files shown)


frontends/web/src/routes/device/waiting.jsx
```diff
@@ -23,19 +23,21 @@ import { Shift, Alert } from '../../components/icon';
...
+import Header from '../../components/header/Header';
 import Footer from '../../components/footer/footer';
 import { debug } from '../../utils/env';
 import InnerHTMLHelper from '../../utils/innerHTML';
 import * as style from './device.css';
 
-export default function Waiting({ testing }) {
+export default function Waiting(props) {
+    const { testing } = props;
     const title = i18n.t('welcome.title');
```
benma: convert this into a proper Component class with `@translate()`, to make `t` available in the props, to be consistent with the rest. 

I suspect that just using `i18n.t` directly does not work: switching language likely won't rerender this to change the text.

frontends/web/src/routes/device/unlock.jsx
```diff
@@ -131,6 +132,7 @@ export default class Unlock extends Component {
         return (
             <div class="contentWithGuide">
                 <div className={style.container}>
+                    <Header title={<h2>Welcome</h2>} {...this.props} />
```
benma: use `t('welcome.title')`?

PR #89, opened 2018-10-02
Eth

benma:


backend/devices/bitbox/keystore.go
```diff
@@ -75,11 +77,18 @@ func (keystore *keystore) ExtendedPublicKey(
...
+		return keystore.SignBTCTransaction(specificProposedTx)
+	case *eth.TxProposal:
+		return keystore.SignETHTransaction(specificProposedTx)
+	default:
+		panic("unknown proposal type")
 	}
-	keystore.log.Info("Sign transaction")
+}
+
+func (keystore *keystore) SignBTCTransaction(btcProposedTx *btc.ProposedTransaction) error {
```
KasparEtter: Make method private. (And personal preference: Avoid forward reference in `SignTransaction`, i.e. move this method above `SignTransaction` – but we're not consistent anyway and I can live without such a change.)

backend/devices/bitbox/keystore.go
```diff
@@ -132,3 +141,31 @@ func (keystore *keystore) SignTransaction(proposedTx coin.ProposedTransaction) e
 	}
 	return nil
 }
+
+func (keystore *keystore) SignETHTransaction(txProposal *eth.TxProposal) error {
```
KasparEtter: Make method private. (And personal preference: Avoid forward reference in `SignTransaction`, i.e. move this method above `SignTransaction` – but we're not consistent anyway and I can live without such a change.)

backend/devices/bitbox/keystore.go
benma: This is not C :P Done :)

backend/devices/bitbox/keystore.go
benma: Done.

PR #91, opened 2018-10-02
bitbox: better wording around backup creation

benma:
1) Explain that the recovery password is for verification (not a new
encryption password)
2) If the verification fails, refresh the backup list anyway, and give
a proper response instead of 'failed'.


frontends/web/src/components/backups/create.jsx
```diff
@@ -74,6 +75,10 @@ export default class Create extends Component {
                 alert(data.errorMessage); // eslint-disable-line no-alert
             } else {
                 this.props.onCreate();
+                console.log(data);
```
KasparEtter: Is this log still necessary? And doesn't ESLint complain?

frontends/web/src/components/backups/create.jsx
```diff
@@ -104,6 +109,7 @@ export default class Create extends Component {
                                     placeholder={t('backup.create.name.placeholder')}
                                     onInput={this.handleFormChange}
                                     value={backupName} />
+                                <InnerHTMLHelper tagName="p" html={t('backup.create.info')} />
```
KasparEtter: Why do you use `InnerHTMLHelper` instead of just `<p>{t('backup.create.info')}</p>`?

frontends/web/src/components/backups/create.jsx
```diff
@@ -74,6 +75,10 @@ export default class Create extends Component {
                 alert(data.errorMessage); // eslint-disable-line no-alert
             } else {
                 this.props.onCreate();
+                console.log(data);
+                if (!data.verification) {
+                    alert(this.props.t('backup.create.verificationFailed'));
+                }
```
x1ddos: A bit unusual... I don't think the app uses native alerts anywhere else. Maybe change to <Message> or whatever is used normally, for consistency?

frontends/web/src/components/backups/create.jsx
benma: No, good catch.

CI passed, so I guess not ;)

frontends/web/src/components/backups/create.jsx
benma: Done. I copied the approach from other dialogs like HiddenWallet, but I guess the same should be fixed there at some point.

frontends/web/src/components/backups/create.jsx
benma: Uses it in lots of places.

benma: Afaik all dialogs and single-button actions use alerts, and the setup wizard uses messages.

PR #94, opened 2018-10-05
frontend/guide: i18n fallback

benma:


frontends/web/src/components/guide/guide.tsx
```diff
@@ -61,14 +62,30 @@ interface ProvidedProps {
 
 type Props = ProvidedProps & SharedProps & TranslateProp;
 
+// guide entries are fetched from the translation files. Manual fallback to is implemented in case an
+// array element is undefined in the current language (can be null if not translated in order, for
```
KasparEtter: So the idea is that, for example,
```
        "backups": [
            {
                "title": "What is a backup?",
                "text": [ "It is a copy of the seed on an micro SD card. The seed together with your recovery password generates your wallet." ]
            },
            {
                "title": "Can I encrypt the backup?",
```
could be "translated" to German with
```
        "backups": [
            null,
            {
                "title": "Wie kann ich ein Backup verschlüsseln?",
```
if I'm too lazy to also translate the first entry?

frontends/web/src/components/guide/guide.tsx
```diff
@@ -61,14 +62,30 @@ interface ProvidedProps {
...
+    const key = 'guide.' + screen;
+    // For now, fallback to the main fallback language, not cascading.
+    const entriesFallback = i18n.getFixedT(i18n.languages[i18n.languages.length - 1])(key, { defaultValue: [] });
+    const entriesCurrentLng = t(key, { defaultValue: [] });
+    return entriesFallback.map((entry, index) => {
+        if (index < entriesCurrentLng.length && entriesCurrentLng[index] !== null) {
+            return entriesCurrentLng[index];
+        }
+        return entry;
+    });
```
KasparEtter: JavaScript has no `IndexOutOfBoundsException` and thus the following would be conciser and easier to read for me (but I haven't tried it):
```
return entriesFallback.map((entry, index) => entriesCurrentLng[index] || entry);
```
On a different note, it would be more intuitive to me if we would fill missing entries in the current language and not overwrite the entries where available of the fallback language.

frontends/web/src/components/guide/guide.tsx
```diff
@@ -61,14 +62,30 @@ interface ProvidedProps {
...
 type Props = ProvidedProps & SharedProps & TranslateProp;
 
+// guide entries are fetched from the translation files. Manual fallback to is implemented in case an
+// array element is undefined in the current language (can be null if not translated in order, for
+// example).
+function getGuideEntries(t, screen): ImplicitEntry[] {
+    const key = 'guide.' + screen;
+    // For now, fallback to the main fallback language, not cascading.
+    const entriesFallback = i18n.getFixedT(i18n.languages[i18n.languages.length - 1])(key, { defaultValue: [] });
+    const entriesCurrentLng = t(key, { defaultValue: [] });
```
KasparEtter: I have quite a strong preference to call such variables `entriesCurrentLanguage` (or `entriesOfCurrentLanguage` or `currentLanguageEntries`). It's mentally just so much easier to read/parse. I first thought `lng` might refer to `length`. 😬

frontends/web/src/components/guide/guide.tsx
```diff
@@ -61,14 +62,30 @@ interface ProvidedProps {
 
 type Props = ProvidedProps & SharedProps & TranslateProp;
 
+// guide entries are fetched from the translation files. Manual fallback to is implemented in case an
+// array element is undefined in the current language (can be null if not translated in order, for
+// example).
+function getGuideEntries(t, screen): ImplicitEntry[] {
+    const key = 'guide.' + screen;
+    // For now, fallback to the main fallback language, not cascading.
+    const entriesFallback = i18n.getFixedT(i18n.languages[i18n.languages.length - 1])(key, { defaultValue: [] });
```
KasparEtter: For consistency, `entriesFallback` should be renamed to `entriesFallbackLng` respectively `entriesFallbackLanguage`. Code (including variables) should speak as much as they can for themselves. 🕺

frontends/web/src/components/guide/guide.tsx
benma: Correct. More importantly, not due too laziness, but just out of order translation can also produce `null` entries, which crashes the frontend.

Also, if something is missed, it's nicer to show a fallback instead of nothing, like in the rest of the framework like with `t('only.exists.in.english')`, which falls back if not present in another language.

benma:
> Sorry for being nitpicky. However, code is read more often than it is written… nerd_face

:+1: :+1:

frontends/web/src/components/guide/guide.tsx
KasparEtter: > just out of order translation can also produce null entries

How so? Only with explicit `null` entries, or not?

frontends/web/src/components/guide/guide.tsx
```diff
@@ -61,14 +62,25 @@ interface ProvidedProps {
 
 type Props = ProvidedProps & SharedProps & TranslateProp;
 
+// Guide entries are fetched from the translation files. Manual fallback to is implemented in case an
```
KasparEtter: > Manual fallback to …

still missing.

frontends/web/src/components/guide/guide.tsx
benma: No, it is how the translation tool works. If you translate the 3rd entry of a guide, but for some reason have not done the 2nd one yet, the json that comes out will have `null` in the 2nd place. 

Guide array elements are translated as units, if that was the confusion.

frontends/web/src/components/guide/guide.tsx
benma: Oops. Done!

PR #95, opened 2018-10-06
Dockerfile: update linuxqtdeploy checksum

Lederstrumpf:
Prerelease version of `linuxqtdeploy` was updated yesterday, so checksum verification during Docker image creation would fail.


benma:
Can you please redo this with `e5294433d97504a5081c354cdedfebe918dd50188c378d965f598576fcbbf5b4`? I think they pushed a new one today which fixes the issue about libcairo.

PR #98, opened 2018-10-09
Frontend confirm alert updates

KasparEtter:
I made a PR for being able to give feedback.


frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    componentDidMount() {
+        document.addEventListener('keydown', this.handleKeyDown);
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('keydown', this.handleKeyDown);
+    }
+
+    handleClose = e => {
```
KasparEtter: I wanted to activate `--noImplicitAny` (see https://www.typescriptlang.org/docs/handbook/compiler-options.html) but could not because of the `checkJS`. However, once I can, `e` would be marked as an error because it is implicitly `any`. I haven't found the optimal solution yet but so far I fixed it with https://github.com/digitalbitbox/bitbox-wallet-app/blob/f4309109de5cf8392d2aba990e1798b6c041ff9c/frontends/web/src/components/fiat/fiat.tsx#L33-L37 Because you don't need e here in this case, you can omit the `as` part here.

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+        this.setState({
+            active: false,
+            message: '',
+        });
+    }
+
+    handleKeyDown = e => {
```
KasparEtter: Same here.

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+        if (e.keyCode === 13 && this.state.active) this.setState({ active: false });
```
KasparEtter: It has nothing to do with TypeScript but I tend to always use curly braces even for one-liners by now.

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    alerted = message => {
```
KasparEtter: I guess TypeScript cannot infer that `message` has to be a string and thus `message` is probably implicitly any again. Just use: `alerted = (message: string) => {`

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+        }, () => {
+            this.button.base.focus();
+        });
+    }
+
+    public render({}, {
+        message,
+        active,
+    }) {
+        return active ? (
```
KasparEtter: It sucks that you have to do this (instead of `active && …`) but this should be fixed with my accepted PR to preact (it had the number 1219 but I do not want to add a backreference to this public PR there).

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+                            key={i}
+                            class={ i === 0 ? 'first' : '' }>
+                            {line}
+                        </p>
+                    ))
+                }
+                <div class="buttons flex flex-row flex-end">
+                    <Button
+                        primary
+                        ref={ref => this.button = ref}
```
KasparEtter: Ah, now I see why you defined the interface above (and I didn't know that it works like this). However, it is straightforward to declare "fields" in classes with TypeScript. See, for example: https://github.com/digitalbitbox/bitbox-wallet-app/blob/f4309109de5cf8392d2aba990e1798b6c041ff9c/frontends/web/src/decorators/load.tsx#L63

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+                        onClick={this.handleClose}>
+                        {i18n.t('button.ok')}
```
KasparEtter: I always feel a bit uneasy with using `i18n.t` directly like this since the `@translate` decorator not only injects `t` as a property but also makes sure that the component is rerendered when the language is changed. I had this discussion with Eni in the past: It might be that it's enough if the uppermost element uses the translate decorator/higher-order component for rerendering. However, without verifying such a behavior (which might break with a future update of Preact), I wouldn't rely on it (as both React and Preact often preserve children with their state for performance reasons).

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    </Button>
+                </div>
+            </Dialog>
+        ) : null;
+    }
+}
+
+export { alerted };
+export default Alert;
```
KasparEtter: GitHub doesn't seem to be happy that there is no newline at the end of the file. 🙂

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -15,24 +15,34 @@
...
+        super(props);
+        confirmation = this.confirmation;
+        this.state = {
+            active: false,
+            message: '',
+        };
     }
 
-    confirmOverride = (message, callback) => {
+    confirmation = (message, callback) => {
```
KasparEtter: Types of `message` and `callback` will be implicitly `any` (which is not ideal, see above).

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -42,27 +52,24 @@ export default class Confirm extends Component {
...
         });
     }
 
-    render({}, {
+    public render({}, {
         message,
-        callback,
         active,
     }) {
-        if (!active) return null;
```
KasparEtter: `if (!active) return null;` should also work with TypeScript, so changing it to `return active ? (` should not have been necessary (but I don't mind, I just wanted to make this clear).

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -42,27 +52,24 @@ export default class Confirm extends Component {
...
-                    <p>{message}</p>
-                    <div class={style.buttons}>
-                        <Button secondary onClick={() => this.respond(false)}>Cancel</Button>
-                        <Button primary onClick={() => this.respond(true)}>Confirm</Button>
-                    </div>
+        return active ? (
+            <Dialog>
+                <p class="first">{message}</p>
+                <div class="buttons flex flex-row flex-between">
+                    <Button secondary onClick={() => this.respond(false)}>Cancel</Button>
```
KasparEtter: Ideally, we would avoid creating new functions on every render (but I know that we still do it all over the place).

frontends/web/src/routes/settings/electrum.jsx
```diff
@@ -69,23 +71,22 @@ class ElectrumServer extends Component {
             if (data.success) {
                 this.setState({ electrumCert: data.pemCert });
             } else {
-                alert(data.errorMessage);
+                // @ts-ignore
+                alerted(data.errorMessage);
```
KasparEtter: I guess using just `alert` doesn't work? Maybe use `alertUser` instead? `alerted` feels strange. And what does TypeScript complain about here?

frontends/web/src/routes/settings/electrum.jsx
```diff
@@ -214,8 +215,8 @@ class ElectrumServers extends Component {
     }
 
     resetToDefault = () => {
-        // @ts-ignore (non-standard usage of confirm method with a second argument)
-        confirm('Do you want to remove all servers and install the default servers?', response => {
+        // @ts-ignore
```
KasparEtter: I guess `@ts-ignore` should no longer be necessary now?

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+var alerted: (message: string) => void;
+
+interface State {
+    message: string;
+    active: boolean;
+}
+
+interface Alert {
+    button: any;
```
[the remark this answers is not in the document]
KasparEtter: What you are looking for might be `VNode<any>` or `JSX.Element`.

KasparEtter:
PS: Since this is work-in-progress anyway, I didn't really look at `dialog.jsx`. I will, once you convert it to TypeScript as well! 😉

KasparEtter:
And some of the things I remarked were also caught by TSLint (on Travis):
```
$ tslint -e 'src/**/*.css.d.ts' 'src/**/*.ts?(x)'
ERROR: src/components/alert/Alert.tsx[22, 1]: Forbidden 'var' keyword, use 'let' or 'const' instead
ERROR: src/components/alert/Alert.tsx[43, 5]: The class method 'componentDidMount' must be marked either 'private', 'public', or 'protected'
ERROR: src/components/alert/Alert.tsx[47, 5]: The class method 'componentWillUnmount' must be marked either 'private', 'public', or 'protected'
ERROR: src/components/alert/Alert.tsx[51, 5]: The class property 'handleClose' must be marked either 'private', 'public', or 'protected'
ERROR: src/components/alert/Alert.tsx[58, 5]: The class property 'handleKeyDown' must be marked either 'private', 'public', or 'protected'
ERROR: src/components/alert/Alert.tsx[59, 9]: if statements must be braced
ERROR: src/components/alert/Alert.tsx[62, 5]: The class property 'alerted' must be marked either 'private', 'public', or 'protected'
ERROR: src/components/alert/Alert.tsx[100, 22]: file should end with a newline
ERROR: src/components/confirm/Confirm.tsx[21, 1]: Forbidden 'var' keyword, use 'let' or 'const' instead
ERROR: src/components/confirm/Confirm.tsx[42, 5]: The class property 'confirmation' must be marked either 'private', 'public', or 'protected'
ERROR: src/components/confirm/Confirm.tsx[50, 5]: The class property 'respond' must be marked either 'private', 'public', or 'protected'
```

frontends/web/src/components/alert/Alert.tsx
originalblend337: @KasparEtter -- yes, this slipped through, will be updated.

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+interface State {
+    active: boolean;
+    message?: string;
+}
+
+@translate()
+class Alert extends Component<Props, State> {
+    private button!: Component;
```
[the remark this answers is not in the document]
KasparEtter: Ok, I see what you are doing here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization. I'd be very careful with using "definite assignment assertion" – especially since the field is not actually initialized after the constructor. It requires that you reasoned through all assignments and usages of the variables. Please add a comment why your assumption (that `button` is only used initialized) holds (like `// Initialized after initial rendering.`).

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+    private callback!: () => void;
```
KasparEtter: Maybe add `// Assigned when the alert is opened and only called when the alert is closed again.`?

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    constructor(props) {
+        super(props);
+        alertUser = this.alerted;
+        this.state = {
+            active: false,
+        };
+    }
+
+    private setButtonRef = element => {
```
KasparEtter: After typing `button` with `JSX.Element`, please type `element` as `private setButtonRef = (element: JSX.Element) => {`. I want to activate --noImplicitAny (see https://www.typescriptlang.org/docs/handbook/compiler-options.html) as soon as we have transformed all JS to TS.

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+        this.button = element;
+    }
+
+    private handleClose = e => {
```
KasparEtter: Please type `e` as `Event`, no import necessary, see https://github.com/digitalbitbox/bitbox-wallet-app/blob/f4309109de5cf8392d2aba990e1798b6c041ff9c/frontends/web/src/components/fiat/fiat.tsx#L33-L37

KasparEtter: Or since you don't use `e` at all, maybe just write `() => {…}`? JS is definitely happy with this and so should TS if I didn't make any mistakes: https://www.typescriptlang.org/play/index.html#src=const%20f%20%3D%20()%20%3D%3E%20%7B%20%7D%3B%0D%0Aconst%20g%3A%20(e%3A%20Event)%20%3D%3E%20void%20%3D%20f%3B%0D%0A

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+        if (this.callback) {
```
KasparEtter: Since you defined `callback` as `callback!`, this check isn't necessary, right? Personally, I prefer this check to the `!` declaration because without this check you add an implicit (and undocumented) contract (https://en.wikipedia.org/wiki/Design_by_contract) that `handleClose` may only be called after `alerted`.

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+            this.callback();
+        }
+        this.setState({
+            active: false,
+            message: undefined,
+        });
+    }
```
KasparEtter: Should you also clear the `callback`? If you say it will be overwritten anyway, then the same logic applies to `message`, right?

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+        this.callback = callback;
+        this.setState({
+            message,
+            active: true,
+        }, () => {
+            this.button.base!.focus();
+        });
+    }
+
+    public render({ t }, { message, active }) {
```
KasparEtter: The only reason why you could call `t` as you do below (given your above declaration as ``) is because it's implicitly any. Fix it like `public render({ t }: RenderableProps<Props>, { message, active }: State) {`. (Respectively use `TranslateProp` once you removed `Props`.)

See, for example, https://github.com/digitalbitbox/bitbox-wallet-app/blob/7ad66bf24820de61a50153cc028c632c570d432f/frontends/web/src/components/update/update.tsx#L38

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+    active: boolean;
+    message?: string;
+}
+
+@translate()
+class Alert extends Component<Props, State> {
+    private button!: Component;
+    private callback!: () => void;
+
+    constructor(props) {
```
KasparEtter: I guess this should be typed as well: `props: Props` (or then directly `props: TranslateProp`).

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+        if (this.callback) {
+            this.callback();
+        }
+        this.setState({
+            active: false,
+            message: undefined,
+        });
+    }
+
+    private alerted = (message, callback) => {
```
KasparEtter: I like the rename to `alertUser` but please also rename this method. (Before you also used the same name for both, correct?)

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component, h } from 'preact';
+import { translate } from 'react-i18next';
```
KasparEtter: Ditto

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Button } from '../forms';
+
+let confirmation: (message: string, callback: () => void) => void;
+
+interface State {
+    message: string;
+    active: boolean;
+}
+
+interface Props {
```
KasparEtter: Ditto

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+    t: () => void;
+}
+
+@translate()
```
KasparEtter: Ditto

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+class Confirm extends Component<Props, State> {
+    private callback!: (input: boolean) => void;
+
+    constructor(props) {
```
KasparEtter: Ditto

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+        super(props);
+        confirmation = this.confirmation;
+        this.state = {
+            active: false,
+            message: '',
+        };
+    }
+
+    private confirmation = (message, callback) => {
```
KasparEtter: Type variables.

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+        this.setState({
+            active: true,
+            message,
+        });
+        this.callback = callback;
+    }
+
+    private respond = input => {
```
KasparEtter: Type variable.

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+        this.callback(input);
+        this.setState({
+            active: false,
+            message: '',
+        });
+    }
```
KasparEtter: Clear `callback`?

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    private decline = () => {
+        this.respond(false);
+    }
+
+    private accept = () => {
+        this.respond(true);
+    }
+
+    public render({ t }, { message, active }) {
```
KasparEtter: Type the parameters.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component, h } from 'preact';
+import * as style from './dialog.css';
+
+interface Props {
+    children: JSX.Element[] | JSX.Element;
```
KasparEtter: This is usually added by `RenderableProps`. Any reason to declare it explicitly?

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+    title?: string;
+    small?: boolean;
+    disableEscape?: boolean;
+    onClose: (e: any) => void;
```
KasparEtter: `Event` instead of `any`?

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+interface State {
+    active: boolean;
+    currentTab: number;
+}
+
+class Dialog extends Component<Props, State> {
+    private modalContent!: HTMLElement;
+    private focusableChildren!: NodeListOf<HTMLElement>;
+
+    constructor(props) {
```
KasparEtter: Type parameter.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    public componentDidMount() {
+        setTimeout(this.activate, 10);
+    }
+
+    public componentWillUnmount() {
+        this.deactivate();
+    }
+
+    private setModalContent = element => {
```
KasparEtter: Untyped parameter.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    private focusWithin = () => {
+        this.focusableChildren = this.modalContent.querySelectorAll('a, button, input, textarea');
+        for (const c of this.focusableChildren) {
+            c.classList.add('tabbable');
+        }
+        document.addEventListener('keydown', this.controlKeys);
+    }
+
+    private addTabIndex = currentTab => {
```
KasparEtter: Untyped parameter.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+        let next = currentTab + 1;
+        if (next >= this.focusableChildren.length) {
+            next = 0;
+        }
+        this.setState({ currentTab: next });
+    }
+
+    private subtractTabIndex = currentTab => {
```
KasparEtter: Untyped parameter.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+        let previous = currentTab - 1;
+        if (previous < 0) {
+            previous = this.focusableChildren.length - 1;
+        }
+        this.setState({ currentTab: previous });
+    }
+
+    private controlKeys = e => {
```
KasparEtter: Untyped parameter.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+        document.removeEventListener('keydown', this.controlKeys);
+    }
+
+    private activate = () => {
+        this.setState({ active: true }, () => {
+            this.focusWithin();
+        });
+    }
+
+    public render({ title, small, children, onClose }, { active, currentTab }) {
```
KasparEtter: Untyped parameters.

frontends/web/src/components/alert/Alert.tsx
originalblend337: I changed `callback` to be optional (`private callback?: () => void;`), which makes the check for `callback` in `handleClose` useful

originalblend337:
@KasparEtter -- can you put your eyes through it again?

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,94 @@
...
+        this.callback = callback;
+        this.setState({
+            active: true,
+            message,
+        }, () => {
+            this.button.base!.focus();
+        });
+    }
+
+    public render({ t }: TranslateProp, { message, active }: State) {
```
KasparEtter: Strictly speaking, the types should be as following in order to properly override the abstract (meaning non-implemented) render method in `Component` (https://github.com/developit/preact/blob/master/src/preact.d.ts#L101):
```
public render({ t }: RenderableProps<TranslateProp>, { message, active }: Readonly<State>) {
```
Since we destructure the state immediately, `Readonly` has no effect here other than correctly overriding the abstract render method, whereas `RenderableProps` gives you access to the `children`, which you don't need to define in your `Props` type (unless you want to type them explicitly).

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+interface State {
+    active: boolean;
+    currentTab: number;
+}
+
+class Dialog extends Component<Props, State> {
+    private modalContent!: HTMLElement;
+    private focusableChildren!: NodeListOf<HTMLElement>;
```
KasparEtter: I'm already learning TypeScript from you! 😮 🎉

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+        }
+        document.addEventListener('keydown', this.controlKeys);
+    }
+
+    private addTabIndex = (currentTab: number) => {
+        let next = currentTab + 1;
+        if (next >= this.focusableChildren.length) {
+            next = 0;
+        }
+        this.setState({ currentTab: next });
```
KasparEtter: You could shrink this method to one line (but I don't care if you want to do it this time):
```
this.setState({ currentTab: (currentTab + 1) % this.focusableChildren.length });
```

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private subtractTabIndex = (currentTab: number) => {
+        let previous = currentTab - 1;
+        if (previous < 0) {
+            previous = this.focusableChildren.length - 1;
+        }
+        this.setState({ currentTab: previous });
```
KasparEtter: This one as well:
```
this.setState({ currentTab: (currentTab - 1) % this.focusableChildren.length });
```

frontends/web/src/routes/account/receive/receive.jsx
```diff
@@ -50,10 +51,19 @@ export default class Receive extends Component {
...
-        document.addEventListener('keydown', this.handleKeyDown);
+        this.registerEvents();
     }
 
     componentWillUnmount() {
+        this.unregisterEvents();
+    }
+
+    registerEvents = () => {
+        console.log('poopy');
```
KasparEtter: That one was not meant to be committed, right? 😆

benma:
:tada: thanks @originalblend337  and @KasparEtter for the epic review!

PR #100, opened 2018-10-10
Rates

benma:
Rates were observed when coins were initialized, which is after
unlock. The first rates notification happens immediately when the
backend is initalized, so the frontend misses the first notification,
and waits a full minute for the first one arrives.

Also, for each coin, one notification was made, but there is a single
global observer/rates variable for all coins in the frontend, so it
does redundant work.

Now, the rates are removed from coins and are part of the backend, to
match how they are used in the frontend. Can move it back if we ever
do per-coin fetches, but that seems unlikely.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -116,7 +116,7 @@ func NewBackend(arguments *arguments.Arguments) *Backend {
 		coins:     map[string]coin.Coin{},
 		log:       log,
 	}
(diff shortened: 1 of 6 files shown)


benma:
> should not be a problem because the rates are fetched initially 

It seems this happens before the rates are ready? not sure.

PR #101, opened 2018-10-11
frontend/i18n: fix backups

benma:
Needs to take `t` from props, otherwise the labels won't be updated when the language is changed.


KasparEtter:
I know I'm too late but generally I'd prefer that we use functional components whenever possible (i.e. when no state is required).

benma:
Not sure how it applies. This component has no state?

KasparEtter:
You turned a function into a class for no good reason – other than not knowing how to apply a decorator to a function.

benma:
I think the bigger issue was that I did not see that a function is better than a class. Will learn more from you soon on the topic!

PR #104, opened 2018-10-11
frontend: hidden wallet error messages

benma:
Was cryptic coming from backend.

### DIFF
--- a/frontends/web/src/routes/device/settings/components/hiddenwallet.jsx
+++ b/frontends/web/src/routes/device/settings/components/hiddenwallet.jsx
@@ -73,14 +73,16 @@ export default class HiddenWallet extends Component {
         apiPost('devices/' + this.props.deviceID + '/set-hidden-password', {
             pin: this.state.pin,
             backupPassword: this.state.password,
-        }).catch(() => {}).then(({ success, didCreate, errorMessage }) => {
+        }).catch(() => {}).then(({ success, didCreate, errorMessage, code }) => {
             this.abort();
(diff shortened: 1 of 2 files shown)


benma:
@TheCharlatan please check and approve :)

frontends/web/src/routes/device/settings/components/hiddenwallet.jsx
```diff
@@ -127,6 +129,7 @@ export default class HiddenWallet extends Component {
                                     onValidPassword={this.setValidPIN} />
                                 <PasswordRepeatInput
                                     idPrefix="password"
+                                    pattern="^.{8,}$"
```
thisconnect: The pattern attribute is nice, but if it is always active you can't submit the form in this case you never see the error. A user would not be informed that the password is too short when the input field only gets red.

benma: Makes sense. Should I remove the pattern, or add a small text `min. 8 characters`?

thisconnect: yes please remove the attribute for now

thisconnect: We could add it later in another PR, after we are sure there was an error, but maybe have to allow submitting even if it is invalid.

benma:
@thisconnect done

thisconnect:
thanks

PR #132, opened 2018-10-18
Frontend wizard restore optimization

thisconnect:


frontends/web/src/routes/device/setup/components/tampered.jsx
```diff
@@ -0,0 +1,31 @@
...
+export function Tampered({
+    style = null
+}) {
+    return (
+        <Message type="warning" style={style}>
+            <Alert />
+            {i18n.t('deviceTampered')}
+        </Message>
+    );
+}
```
KasparEtter: I thought the deal was that new code is written in TypeScript! 😜

benma:
Seems to be based on old commits, e.g. before i18n/en.json moved to locales/en/app.json, and much more.

Please rebase first :)

frontends/web/src/routes/device/setup/components/tampered.jsx
thisconnect: no sorry and I am going to do another change before TS

thisconnect:
rebased

frontends/web/src/locales/en/app.json
```diff
@@ -588,11 +589,16 @@
   },
   "seedRestore": {
     "info": {
-      "button": "Continue",
-      "description": "To continue please insert the micro SD card in to the BitBox.",
-      "title": "How to restore a wallet from a backup"
+      "title": "Restore a wallet",
+      "description_1": "Insert the micro SD card in to the BitBox and click “Continue”",
```
benma: Please don't use `_` in keys, unless it is used as dynamic context. e.g. `t('my.key', { context: 'foo' })` resolves to `my.key_foo`.

thisconnect: tried to fix, but some are coming from the backend. i.e. `unlock.error.e109_normal`

benma: Yes, that is a valid use of context, see [here](https://github.com/digitalbitbox/bitbox-wallet-app/blob/c843d712207594a567f785ba083f2980fc5a6522/frontends/web/src/routes/device/unlock.jsx#L127). Thanks for doing the rest :)

benma:
(this is technically already merged, GH having trouble right now with consistency)

PR #136, opened 2018-10-19
Keyed entries

benma:


frontends/web/src/components/guide/guide.tsx
```diff
@@ -107,6 +74,6 @@ function Guide({ screen, shown, t, children }: RenderableProps<Props>): JSX.Elem
     );
 }
 
-const HOC = translate<ProvidedProps>()(share<SharedProps, ProvidedProps & TranslateProp>(store)(Guide));
+const HOC = translate<{}>()(share<SharedProps, TranslateProp>(store)(Guide));
```
KasparEtter: `translate` has a default generic type of `{}` and thus you should be able to skip all the generics of `translate`: https://github.com/digitalbitbox/bitbox-wallet-app/blob/e144c48cdee32739fd144ba4a2488d190c6fd083/frontends/web/src/decorators/translate.ts#L66

frontends/web/src/routes/device/waiting.jsx
```diff
@@ -54,7 +54,7 @@ export default class Waiting extends Component {
                     </div>
                 </div>
                 <Guide>
-                    <Entry entry={t('guide.waiting.welcome')} />
+                    <Entry entry={t('guide.waiting.welcome')} shown={true} />
```
KasparEtter: Yay! 🎉

PR #137, opened 2018-10-22
frontend: tab behavior fixes in dialog component

originalblend337:


benma:
Found one issue: click i.e. on `Change device password`. Nothing is focused. Hitting tab does not focus the first element, but the second, as if the first was previously focused.

Would be good to just focus the first element.

PR #144, opened 2018-10-27
frontend: fetch state in waiting instead of parent

benma:
Also remove unused deviceIDs prop in settings.


frontends/web/src/routes/device/waiting.jsx
```diff
@@ -30,8 +30,20 @@ import * as style from './device.css';
 
 @translate()
 export default class Waiting extends Component {
```
KasparEtter: This would have been an excellent chance to use `@load(…)`! 🙂

PR #153, opened 2018-10-30
backend: fix a typo in btc and eth account

KasparEtter:


backend/coins/eth/account.go
```diff
@@ -50,7 +50,7 @@ type Account struct {
 	log *logrus.Entry
 }
 
-// NewAccount creats a new Account.
+// NewAccount creates a new account.
```
benma: `Account` was uppercase to match the struct name, but this does look nicer ;)

KasparEtter: Yeah, the PR was mainly intended to fix "creates". When I made the PR I realized that I could lowercase account in the same run and force pushed an amendment.

KasparEtter:
What shall I do about the CI, @benma? 🤷‍♂️

PR #157, opened 2018-10-30
backend: rename initialSyncDone to initialized in accounts

KasparEtter:


benma:
Isn't that more confusing now? Initialized() sounds like it is true after Init(), but it is actually about finishing the first round of syncing with electrumx.

KasparEtter:
Difficult to see/analyze. I guess it's triggered with https://github.com/digitalbitbox/bitbox-wallet-app/blob/236db43ffcc2239f2aa8773068e6814618e42c53/backend/coins/btc/account.go#L253-L255? Maybe we could/should add a comment that the initialization is asynchronous and not finished after `Initialize()`?

benma:
> Maybe we could/should add a comment that the initialization is asynchronous and not finished after Initialize()?

Or leave the previous name, which I feel is more descriptive? :)

I'll leave it to you though.

PR #162, opened 2018-10-30
Remove unused code in frontend

KasparEtter:


frontends/web/src/components/backups/backup.jsx
```diff
@@ -18,10 +18,9 @@ import { Component, h } from 'preact';
 import { translate } from 'react-i18next';
 import { Radio } from '../forms';
 
-@translate()
+@translate() // Leaving this here to ensure that the component re-renders when the language changes.
```
benma: Good catch!

frontends/web/src/components/backups/restore.jsx
```diff
@@ -123,7 +123,6 @@ export default class Restore extends Component {
         selectedBackup,
         requireConfirmation,
     }, {
-        password,
```
benma: If not used in rendering, you can remove it from the state map and just store it directly (`this.password`)?

frontends/web/src/routes/account/send/utxos.jsx
```diff
@@ -71,7 +71,6 @@ export default class UTXOs extends Component {
         t,
         active,
     }, {
-        show,
```
benma: Remove it from the whole component.

show() and hide() also not used.

Also seems like a new bug: when coin control was hidden via `hide()`, `this.props.onChange(this.state.selectedUTXOs);` was called, so the send.jsx parent component erased the selected utxos. Probably not happening anymore. 

Pls fix all of that at once, or make an issue to do it later :)

frontends/web/src/routes/settings/electrum.jsx
```diff
@@ -95,10 +95,8 @@ class ElectrumServer extends Component {
 
     render({
         t,
-        server,
         onAdd,
         onRemove,
-        index,
```
benma: Also remove here? https://github.com/digitalbitbox/bitbox-wallet-app/blob/72dd16b41281a0bb02709d7f9aff3101b854487c/frontends/web/src/routes/settings/electrum.jsx#L274

PR #171, opened 2018-10-31
Nonce

benma:


backend/coins/eth/account.go
```diff
@@ -222,6 +224,23 @@ func (account *Account) update() error {
...
 
+	// Nonce to be used for the next tx, fetched from the ETH node. It might be out of date due to
+	// latency, which is addressed below by using the locally stored nonce.
+	nodeNonce, err := account.coin.client.PendingNonceAt(context.TODO(), account.address.Address)
+	if err != nil {
+		return err
+	}
+	account.nextNonce = nodeNonce
+
+	// Usually, our locally stored last nonce is used to compute the next noce. Edge case: if a user
```
KasparEtter: Typo: noce -> nonce

backend/coins/eth/account.go
```diff
@@ -222,6 +224,23 @@ func (account *Account) update() error {
 		return err
 	}
 
+	// Nonce to be used for the next tx, fetched from the ETH node. It might be out of date due to
+	// latency, which is addressed below by using the locally stored nonce.
```
KasparEtter: The latency can also be not yet confirmed, right? If yes: `Since only confirmed transactions are considered, the nonce can be out-of-date, which is addressed …`

backend/coins/eth/account.go
KasparEtter: Why `usually`? Unless the user makes several transactions right after one another, `localNonce > account.nextNonce` evaluates to false and thus usually not the locally stored nonce is used.

backend/coins/eth/account.go
```diff
@@ -222,6 +224,23 @@ func (account *Account) update() error {
...
+	nodeNonce, err := account.coin.client.PendingNonceAt(context.TODO(), account.address.Address)
+	if err != nil {
+		return err
+	}
+	account.nextNonce = nodeNonce
+
+	// Usually, our locally stored last nonce is used to compute the next noce. Edge case: if a user
+	// has pending tx but there is no database (restore, use different computer, ...), the nonce
+	// fetched from the node will do.
```
KasparEtter: I find the explanation of the edge case a bit weird. Shouldn't you explain why this method fails in the mentioned edged case? You could change the end of the sentence to `the nonce fetched from the node is out of date and produces invalid transactions, unfortunately`.

backend/coins/eth/account.go
benma: :heavy_check_mark:

backend/coins/eth/account.go
benma: No, pending ones are taken into account by using `PendingNonceAt` instead of `NonceAt`.

backend/coins/eth/account.go
benma: If you have no db, there is no local nonce, so the nodeNonce is used instead. 

I agree the comment is confusing. I just removed it :)

PR #175, opened 2018-11-01
frontend: remove unused state property

originalblend337:


benma:
Please reword the commit title, otherwise we'd have two commits with the same title.

PR #178, opened 2018-11-01
frontend: adds success info

thisconnect:


frontends/web/src/routes/device/setup/success.jsx
```diff
@@ -42,9 +42,20 @@ export default class Success extends Component {
...
+                        <h1 className={style.title} style="text-align: center;">
+                            {t(`success.${goal}.title`)}
+                        </h1>
+                        <div className={style.block}>
+                            <p style="text-align: center;">
+                                {t(`success.${goal}.summary`)}
+                            </p>
+                            { goal === 'create' ? (
+                                <ul class={style.summary}>
+                                    <li>{t(`success.${goal}.info1`)}</li>
```
benma: pls replace the $(goal) with 'create' here, as it can only be this, and it makes looking for keys easier.

frontends/web/src/locales/en/app.json
```diff
@@ -600,6 +600,19 @@
...
+    "create": {
+      "title": "Your BitBox is ready to go!",
+      "summary": "Here is a summary of what you have done",
+      "info1": "Your wallet has been securely backed up to the micro SD Card. Remove it and keep it safe.",
+      "info2": "You have created a secure device password that unlocks the BitBox.",
+      "info3": "You have created a secure recovery password for your wallet that unlocks your funds and restores your backups."
+    },
+    "restore": {
+      "title": "Your BitBox is ready to go!",
+      "summary": "You have successfully restored a wallet from your backup."
```
benma: rename the key also in the japanese translation, so it does not have to be retranslated.

frontends/web/src/locales/en/app.json
```diff
@@ -600,6 +600,19 @@
       "title": "How to restore a wallet from a backup"
     }
   },
+  "success": {
```
benma: Doesn't matter much, but why did you move the whole section up? It will be move down again when I pull from locize.

thisconnect: oh noes

frontends/web/src/locales/en/app.json
thisconnect: done

PR #180, opened 2018-11-01
ethereum: add data field to send and use RPC to estimate required gas

KasparEtter:
Please note that this PR is still in early alpha and has not been tested yet. If you already have feedback, though, then you can give that to me! 🤓


frontends/web/src/routes/account/send/send.jsx
```diff
@@ -457,8 +467,20 @@ export default class Send extends Component {
                                     />
                                     */}
                                 </div>
-                                <p class={style.feeDescription}>{feeTarget && t('send.feeTarget.description.' + feeTarget) || ''}</p>
+                                {feeTarget && <p class={style.feeDescription}>{t('send.feeTarget.description.' + feeTarget)}</p>}
                             </div>
+                            {
+                                (account.coinCode === 'ETH' || account.coinCode === 'TETH') && <div class="row">
```
benma: don't you need `... ? ... : ''` here, otherwise it would render `false`?

backend/coins/eth/account.go
```diff
@@ -285,19 +287,41 @@ type TxProposal struct {
...
+		To:       &address,
+		Gas:      0,
+		GasPrice: suggestedGasPrice,
+		Value:    value,
+		Data:     data,
+	}
+
+	gasLimit, err := account.coin.client.EstimateGas(context.TODO(), message)
+	if err != nil {
+		return nil, err
```
benma: this error happens easily if you add a contract address which fails to execute with the input. should return an error code for the frontend to show a good message.

backend/coins/eth/account.go
```diff
@@ -285,19 +287,41 @@ type TxProposal struct {
...
-	const gasLimit = 21000 // simple transaction gas cost
 
 	suggestedGasPrice, err := account.coin.client.SuggestGasPrice(context.TODO())
 	if err != nil {
 		return nil, err
 	}
-	fee := new(big.Int).Mul(big.NewInt(gasLimit), suggestedGasPrice)
 
-	var value *big.Int
+	parsedAmount, err := amount.Amount(big.NewInt(params.Ether))
```
benma: you do the same below in the else branch, which can be removed (also `value := parsedAmount.BigInt()`)

KasparEtter:
Please have a look again, @benma. And testing the data field is still open for next week.

PR #191, opened 2018-11-05
backend/eth: handle connection error

benma:
Set account to offline, reconnect automatically in next poll
iteration.

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -42,6 +42,8 @@ type Account struct {
 	getSigningConfiguration func() (*signing.Configuration, error)
 	signingConfiguration    *signing.Configuration
 	keystores               keystore.Keystores
+	offline                 bool
+	onEvent                 func(Event)
 
 	initialized bool
 	// enqueueUpdateCh is used to invoke an account update outside of the regular poll update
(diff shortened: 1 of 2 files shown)


backend/coins/eth/account.go
```diff
@@ -42,6 +42,8 @@ type Account struct {
 	getSigningConfiguration func() (*signing.Configuration, error)
 	signingConfiguration    *signing.Configuration
 	keystores               keystore.Keystores
+	offline                 bool
+	onEvent                 func(Event)
```
KasparEtter: You could have used the observable package.

backend/coins/eth/coin.go
```diff
@@ -59,7 +59,7 @@ func (coin *Coin) Initialize() {
 		client, err := ethclient.Dial(coin.nodeURL)
 		if err != nil {
 			// TODO: init conn lazily, feed error via EventStatusChanged
-			panic(err)
+			panic("LOL")
```
KasparEtter: Why does a "LOL" make it to the master?

backend/coins/eth/account.go
benma: Using the same as btc/account.go, so we can refactor both at once.

backend/coins/eth/coin.go
benma: :scream:

PR #197, opened 2018-11-05
frontend: fix route in manage backups

benma:
When the device is removed, we staid in the manage backup
screen. Should route to /.

### DIFF
--- a/frontends/web/src/app.jsx
+++ b/frontends/web/src/app.jsx
@@ -156,12 +156,12 @@ export class App extends Component {
                         <Settings
                             path="/settings" />
                         {/* Use with TypeScript: {Route<{ deviceID: string }>({ path: '/manage-backups/:deviceID', component: ManageBackups })} */}
+                        {/* ManageBackups and DeviceSwitch need a key to trigger (re-)mounting when devices change, to handle routing */}
                         <ManageBackups
                             path="/manage-backups/:deviceID"
-                            // showCreate={true} // Does not exist!
-                            // deviceIDs={deviceIDs} // Does not exist!
+                            key={devices}
+                            devices={devices}
                         />
-                        {/* DeviceSwitch needs a key to trigger ComponentDidMount when devices change, to handle routing */}
                         <DeviceSwitch
                             path="/device/:deviceID"
                             key={devices}
--- a/frontends/web/src/routes/device/manage-backups/manage-backups.jsx
+++ b/frontends/web/src/routes/device/manage-backups/manage-backups.jsx
@@ -15,6 +15,7 @@
  */
 
 import { Component, h } from 'preact';
+import { route } from 'preact-router';
 import { translate } from 'react-i18next';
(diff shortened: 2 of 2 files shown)


frontends/web/src/routes/device/manage-backups/manage-backups.jsx
```diff
@@ -25,7 +26,23 @@ import * as styles from './manage-backups.css';
 
 @translate()
 export default class ManageBackups extends Component {
+    getDevice = () => {
+        if (this.props.deviceID === null) {
+            return null;
```
KasparEtter: As far as I can tell, these lines do not seem to be necessary. `object[null]` returns `undefined` for me and thus the return statement below already achieves the same.

frontends/web/src/routes/device/manage-backups/manage-backups.jsx
```diff
@@ -25,7 +26,23 @@ import * as styles from './manage-backups.css';
...
+    }
+
+    componentWillMount() {
+        if (this.getDevice() === null) {
+            route('/', true);
+        }
+    }
+
     render({ t, deviceID }, { }) {
+        if (this.getDevice() === null) {
```
KasparEtter: This seems quite defensive to me. Why do you think it's necessary / a good idea?

frontends/web/src/app.jsx
```diff
@@ -156,12 +156,12 @@ export class App extends Component {
                         <Settings
                             path="/settings" />
                         {/* Use with TypeScript: {Route<{ deviceID: string }>({ path: '/manage-backups/:deviceID', component: ManageBackups })} */}
+                        {/* ManageBackups and DeviceSwitch need a key to trigger ComponentDidMount when devices change, to handle routing */}
```
KasparEtter: Strictly speaking, you use `componentWillMount ` in `ManageBackups`, not `componentDidMount`, thus `to trigger re-mounting` would be more correct.

frontends/web/src/routes/device/manage-backups/manage-backups.jsx
benma: If it is rendered while the device is not plugged in, the device backend handlers are called and return an error (because already uninitialized).

frontends/web/src/routes/device/manage-backups/manage-backups.jsx
benma: I prefer `null` over `undefined`, it is an explicit value for 'does not exist'.

I side-stepped the issue though by returning a boolean (don't actually need the device anywhere).

frontends/web/src/routes/device/manage-backups/manage-backups.jsx
KasparEtter: But if `componentWillMount` `routes` the app away, will the component still be rendered? If no, when will there be "no device" in the `render` method? But nevermind…

PR #198, opened 2018-11-05
Ethereum send-all improvements

KasparEtter:


backend/coins/eth/account.go
```diff
@@ -320,43 +320,42 @@ func (account *Account) newTx(
...
+		return nil, errp.WithStack(coin.ErrInvalidData)
 	}
 
 	fee := new(big.Int).Mul(new(big.Int).SetUint64(gasLimit), suggestedGasPrice)
 
 	if amount.SendAll() {
+		// Set the value correctly and check that the fee is smaller than or equal to the balance.
 		value = new(big.Int).Sub(account.balance.BigInt(), fee)
-		if value.Sign() <= 0 {
+		if value.Sign() < 0 {
```
benma: `<=` was with the assumption of not having zero value transfers. Since the `allowZero` PR, this fix makes sense :+1:

PR #200, opened 2018-11-06
backend: fix variable scope issues

benma:
scopelint found a number of instances where the code is not robust
with regards to variable reference reuse in a loop (where the loop
body might run in a go-routine).


backend/coins/btc/maketx/maketx.go
```diff
@@ -166,6 +166,7 @@ func NewTx(
 
 		inputs := make([]*wire.TxIn, len(selectedOutPoints))
 		for i, outPoint := range selectedOutPoints {
+			outPoint := outPoint
```
benma: fyi: this works, just because wire.NewTxIn dereferences the outpoint, making a copy.

KasparEtter:
Great, I just did the same: 
https://github.com/digitalbitbox/bitbox-wallet-app/pull/198/commits/b5881cd51a1a3bf1a5f3eaff7d33ba22a22ea850

To be honest, I prefer mine with the comments. 🤓

PR #201, opened 2018-11-06
backend and frontend: add conversions into the formatted amount again

KasparEtter:
Otherwise, the fee has no conversion in the send screen.
All fiat conversions in the confirmation dialog also depended on it.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -101,6 +101,9 @@ type Backend struct {
 	log *logrus.Entry
 }
 
+// ExposedRatesUpdater leaks the rates updater to allow the coin handler to convert to fiat.
+var ExposedRatesUpdater *RatesUpdater
+
 // NewBackend creates a new backend with the given arguments.
 func NewBackend(arguments *arguments.Arguments) *Backend {
 	log := logging.Get().WithGroup("backend")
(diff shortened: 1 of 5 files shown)


backend/backend.go
```diff
@@ -101,6 +101,9 @@ type Backend struct {
 	log *logrus.Entry
 }
 
+// ExposedRatesUpdater leaks the rates updater to allow the coin handler to convert to fiat.
+var ExposedRatesUpdater *RatesUpdater
```
benma: :see_no_evil: 

Please no global mutable vars as shortcuts. They come back to bite us every time.

KasparEtter: Well, I don't need to call a REST endpoint from within Go (obviously, as the code compiles and works without it).

benma: I meant calling it from the frontend in the Send component. A suggestion to save all the extra backend code. Do you see any downside in that?

KasparEtter: In my opinion, the send component and its state (https://github.com/digitalbitbox/bitbox-wallet-app/blob/bb8403bc33f6785e17609effe4f8fd662a21e3a4/frontends/web/src/routes/account/send/send.jsx#L45-L68) is already quite a mess and I do not want to introduce four new state variables and API calls for this, but feel free to do so and close this PR then.

benma: No, that's a good point.

If we keep a global, e.g. looking at rates updater as a singleton, what about making a singleton instance in `rates.go`, and remove it from the Backend struct as well?

Somehow I feel uncomfortable with exposing it this way, as the account does not function properly anymore without a backend being instantiated. Kind of invisible side-effects that can cause maintenance problems.

PR #202, opened 2018-11-07
backend: move add account functionality into own method

KasparEtter:


KasparEtter:
@benma: Please review the second commit as well. I will rebase on CI fix as soon as it is available.

KasparEtter:
@benma: Yet another commit in this PR to review. Thanks!

backend/handlers/handlers.go
```diff
@@ -279,7 +279,6 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
 	}
 	// The following parameters only work for watch-only singlesig accounts at the moment.
 	jsonCoin := jsonBody["coin"]
```
benma: would name this `"coinCode"` for consistency instead of `"coin"`.

frontends/web/src/routes/account/add/addaccount.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { Button, Input, Select } from '../../../components/forms';
+import { Entry } from '../../../components/guide/entry';
+import { Guide } from '../../../components/guide/guide';
+import Header from '../../../components/header/Header';
+import { translate, TranslateProps } from '../../../decorators/translate';
+import { apiPost } from '../../../utils/request';
+
+function submit() {
+    const body = {
+        coin: (document.getElementById('coin') as HTMLSelectElement).value,
```
benma: I would either:

- use [controlled form elements](https://reactjs.org/docs/forms.html) for consistency (it is done that way with every other form in our frontend, I believe)
- or use refs to access the value instead of `id`s (IDs can collide and are hard to track, more easy to break silently, etc.) like described here: https://reactjs.org/docs/uncontrolled-components.html

thisconnect: > use controlled form elements for consistency (it is done that way with every other form in our frontend, I believe)

👍

KasparEtter: Fully agreed. I just hacked the Xpub-Import-MVP together yesterday.

I already collected a lot of outstanding tasks in https://github.com/shiftdevices/godbb/issues/618 and created issues for each of them now.

I wanted to leave form improvements to our frontend experts, which was for me part of `improve the add account route`. I made this issue now explicit (https://github.com/shiftdevices/godbb/issues/634) and will do it myself because I wanted to try https://preactjs.com/guide/linked-state for quite some time. 🤓

Conclusion: I merge this PR now and leave all further improvements, with which I fully agree, to future PRs.

PR #203, opened 2018-11-07
deps: add tool to fix version of dev tools

benma:
`go get ...` always gets master, which is not reliable/robust.

Start by fixing the linter version, so the CI does not break when that
updates in a backwards incompatible manner.


KasparEtter:
CI done. Please merge! 🙂

PR #206, opened 2018-11-08
account/add: use controlled form elements, add linkstate as a dependency

KasparEtter:


KasparEtter:
More info on the new dependency:
- https://preactjs.com/guide/linked-state
- https://github.com/developit/linkstate

PR #209, opened 2018-11-08
account/export: allow to export all transactions of an account in CSV

KasparEtter:


backend/coins/btc/handlers/handlers.go
```diff
@@ -195,6 +200,76 @@ func (handlers *Handlers) getAccountTransactions(_ *http.Request) (interface{},
...
+		"Fee",
+		"Address",
+		"Transaction ID",
+	})
+	if err != nil {
+		return nil, errp.WithStack(err)
+	}
+
+	for _, transaction := range handlers.account.Transactions() {
+		transactionType := map[coin.TxType]string{
```
benma: just `string(transaction.Type())` might work

KasparEtter: I just "blindly" copied your code a few lines higher: https://github.com/digitalbitbox/bitbox-wallet-app/blob/70d12a3f0eb671a6eabadb96ec8e920ddce1599d/backend/coins/btc/handlers/handlers.go#L171-L175

I think I leave it for now because I changed the label to past tense.

PR #213, opened 2018-11-09
 backend: change when the accounts status changed event is triggered

KasparEtter:
Attention: This might be one of those pull requests that break random other stuff.


KasparEtter:
Merged because I need this now. Please still check the changes, @benma.

benma:
I did. ack!

PR #214, opened 2018-11-12
frontend: fix hamburger not responding in manage device and waiting

originalblend337:


frontends/web/src/routes/device/deviceswitch.tsx
```diff
@@ -43,9 +43,9 @@ class DeviceSwitch extends Component<Props, {}> {
         }
         switch (devices[deviceID]) {
         case 'bitbox':
-            return <Device deviceID={deviceID} />;
```
benma: I don't understand this fix. The props passed to DeviceSwitch are:

```
<DeviceSwitch
                            path="/device/:deviceID"
                            key={devices}
                            devices={devices} />
```

Which one is relevant, and why?

Isn't the hamburger in sidebar.jsx?

benma:
fyi appveyor again breaking. Since it's not been very reliable, I removed it as a requirement for merging.

frontends/web/src/routes/device/deviceswitch.tsx
originalblend337: currently in `app.jsx` the `Container` component passes an extra prop of `toggleSidebar` to its child components, which is the function called in the `Header` component of each view.

originalblend337: Right, which is why the fix is to pass props to these components so the `Header` components within would have access to them. But I do agree there is room for refactor here.

I think this branch could be handled as-is since this would fix the bug included in the milestone, and at some point (next milestone?) will revisit this and implement a cleaner version of this.

benma: Ok. I tested the branch, but it didn't work for me. Maybe because the props are not passed through as I mentioned in the previous comment?

Or am I missing something?

PR #217, opened 2018-11-14
Frontend add account updates

originalblend337:


KasparEtter:
Thanks, @originalblend337! 😍

PR #218, opened 2018-11-15
backend/account/add: add coinCode to accountCode

benma:
The signing configuration alone can be ambiguous.


benma:
Same format as the db name:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/5049cb45fd0e7a1f61cef248cc27353a4cc6690c/backend/coins/btc/account.go#L214

(which should also just use the account code, but later)

PR #219, opened 2018-11-15
Ratesupdater

benma:


backend/backend.go
```diff
@@ -118,10 +114,8 @@ func NewBackend(arguments *arguments.Arguments) *Backend {
 		accounts:  []btc.Interface{},
 		log:       log,
 	}
-	ratesUpdater := NewRatesUpdater()
-	ratesUpdater.Observe(func(event observable.Event) { backend.events <- event })
-	backend.ratesUpdater = ratesUpdater
-	ExposedRatesUpdater = ratesUpdater
+	backend.ratesUpdater = GetRatesUpdaterInstance()
```
KasparEtter: Why do we still store it as a field in the backend if it's a singleton? Just subscribing to the events would be enough, wouldn't it?

benma: Yeah, didn't want to change the time when the rates are first initialized. Should look more closely to see if it would work as well.

KasparEtter: You would still initialize it here (because you need to subscribe to the events). You just don't need to persist a reference to a singleton…

benma: You're right. Will fix.

PR #220, opened 2018-11-15
appveyor: base image updated to Qt 5.11.2

benma:


appveyor.yml
```diff
@@ -2,7 +2,7 @@ image: Visual Studio 2017
 
 environment:
   matrix:
-    - QT: C:\Qt\5.11.1\msvc2017_64
+    - QT: C:\Qt\5.11.2\msvc2017_64
```
benma, line 5: why not 5.15.2 which is what we actually currently use? it's also in the README

PR #221, opened 2018-11-15
frontend: error handling in add account

benma:


frontends/web/src/routes/account/add/addaccount.tsx
```diff
@@ -50,7 +51,18 @@ class AddAccount extends Component<TranslateProps, State> {
             accountName: this.state.accountName,
             extendedPublicKey: this.state.extendedPublicKey,
         };
-        apiPost('account-add', body).then(accountCode => route('/account/' + accountCode));
+        interface ResponseData {
+            success: boolean;
+            errorCode?: 'xpubInvalid' | 'xpubWrongNet';
+            accountCode?: string;
+        }
```
KasparEtter: If you want to improve/exercise your TypeScript skills, you could write a [user-defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) to distinguish between the successful and the unsuccessful format. 😉 But I agree it's almost a bit overkill in this case.

backend/handlers/handlers.go
```diff
@@ -294,15 +295,26 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
 	keypath := signing.NewEmptyAbsoluteKeypath()
 	extendedPublicKey, err := hdkeychain.NewKeyFromString(jsonExtendedPublicKey)
 	if err != nil {
-		return nil, err
+		return map[string]interface{}{"success": false, "errorCode": "xpubInvalid"}, nil
+	}
+	if extendedPublicKey.IsPrivate() {
+		return map[string]interface{}{"success": false, "errorCode": "xpubInvalid"}, nil
```
KasparEtter: Since it's quick to change and very important information for the user, you could/should introduce an additional error code `xpubPrivate` with a corresponding message to warn the user that he/she should keep the xpriv confidential. (And after warning the user, we could still import the corresponding xpub as watch-only… 🤷‍♂️)

backend/handlers/handlers.go
```diff
@@ -294,15 +295,26 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
...
+	}
+	if btcCoin, ok := coin.(*btc.Coin); ok {
```
KasparEtter: Maybe fail if not a btc.Coin? 🤷‍♂️

KasparEtter:
More important is that this branch is merged until tomorrow 10 am. 🙂

frontends/web/src/routes/account/add/addaccount.tsx
benma: TIL. Thanks.

benma:
> More important is that this branch is merged until tomorrow 10 am. slightly_smiling_face
In that case I merged now :) Can address your points later.

PR #227, opened 2018-11-19
Accountcode

benma:


frontends/web/src/routes/account/account.jsx
```diff
@@ -123,8 +124,12 @@ export default class Account extends Component {
                 connected: !status.includes('offlineMode'),
                 determiningStatus: false,
             };
-            if (!status.initialized && !status.includes('accountDisabled')) {
+            if (!state.initialized && !status.includes('accountDisabled')) {
```
KasparEtter: Was this again a bug that TypeScript would have prevented (if used correctly)? 😮

benma: absolutely!

PR #229, opened 2018-11-20
Frontend fix guide toggler

originalblend337:


frontends/web/src/components/header/Header.css
```diff
@@ -76,54 +76,6 @@
...
-    width: 18px;
-    height: 18px;
-    line-height: 18px;
-    text-align: center;
-    transform: rotate(-45deg);
-    transition: all ease 0.1s;
-    user-select: none;
-}
-
-.guideWrapper:hover .guideToggler span {
```
KasparEtter: Just looking at the code (without checking it out), didn't the hovering get lost? 🤔 Please ignore if not. 🙂

originalblend337: @KasparEtter -- yeah, playing around with it for a while and first thought the animation didn't really make the usability better so I removed it, but I put it back now with some adjustments.

PR #235, opened 2018-11-25
Update

benma:


KasparEtter:
Was this the reason why the app crashed on macOS when started for the first time? (And now fixed with `go build -ldflags='-s' -x`?)

benma:
pls check by compiling without it. Don't see the connection, but would be good to know.

PR #238, opened 2018-11-26
frontend: wizard steps inside header components

originalblend337:


benma:
Please rebase, because of a conflict.

PR #244, opened 2018-11-27
frontend: disable tamper warning

benma:
Too confusing. Re-enable only when a pre-seeded device is inserted
that has not been seen before.


benma:
> can you just remove it completely?
Done

PR #249, opened 2018-12-06
restore: fix submit button

benma:
Should be updated when the password changes.

### DIFF
--- a/frontends/web/src/components/backups/restore.tsx
+++ b/frontends/web/src/components/backups/restore.tsx
@@ -14,37 +14,53 @@
  * limitations under the License.
  */
 
-import { Component, h } from 'preact';
+import { Component, h, RenderableProps } from 'preact';
 import { route } from 'preact-router';
-import { translate } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/backups/restore.tsx
```diff
@@ -14,37 +14,53 @@
...
+}
+
+type Props = RestoreProps & TranslateProps;
+
+interface State {
+    isConfirming: boolean;
+    activeDialog: boolean;
+    isLoading: boolean;
+    understand: boolean;
+    password: string | null;
```
KasparEtter: Using `undefined` instead of `null` (and then declaring it just as `password?: string;`) would be the TypeScript way to do this in my opinion but I don't mind how it is now.

benma: TIL! Thx, fixed.

thisconnect: `password?: string;` 👍

PR #253, opened 2018-12-13
Frontend app qr codes

originalblend337:


benma:
Thanks @originalblend337 . Please keep jsx->tsx conversions in separate commits so it's easier to see the diff related to the feature.

PR #255, opened 2018-12-18
Frontend RTL

thisconnect:


frontends/web/src/components/transactions/transaction.css
```diff
@@ -152,6 +169,13 @@
     transition: all 0.1s ease;
 }
 
+[lang="ar"] .toggleContainer,
```
benma: why not `.rtl .toggleContainer`?

thisconnect: doesn't work in CSS Modules, but maybe could be enabled somehow?

benma: Then please use css modules :) e.g. set `className={isRTL() ? style.rtl : null}` in an appropriate parent element, then you can use `.rtl .toggleContainer` etc. in the css modules.

KasparEtter: If `.rtl` is defined globally, you can probably use https://github.com/css-modules/css-modules#exceptions.

thisconnect: Picking this up soon™ `:global(.rtl) .toggleContainer` should work, thanks for the hints.

thisconnect: Looking also into https://developer.mozilla.org/en-US/docs/Web/CSS/:dir

thisconnect:
Sorry for keeping this open so long, will revisit later

PR #257, opened 2018-12-18
frontend: add Shift logo to settings screen

thisconnect:


benma:
Would be nice to have the same footer alignment and footer horizontal bar as in Manage Device (switching between them shows the difference).

Approved, so pls fix that if it makes sense, otherwise good to merge :)

PR #269, opened 2019-01-04
frontend: convert security-information and device jsx => tsx

originalblend337:
looked into the wizard and currently it seems to make more sense to change the back button text to 'cancel' or 'abort' at the moment. If/when we support going back a step (i.e. going back after setting a device password) the wizard process can be refactored.


benma:
> Will need to rename the button via locize?

`button.back` is used in more locations. Could use `button.abort` instead, or `dialog.cancel`.

PR #271, opened 2019-01-04
frontend: convert app, container to tsx

benma:


frontends/web/src/components/container/container.tsx
```diff
@@ -14,19 +14,25 @@
  * limitations under the License.
  */
 
-import { h, cloneElement } from 'preact';
+import { cloneElement, h, RenderableProps } from 'preact';
 import { Router } from 'preact-router';
 
+interface Props {
+    children: JSX.Element[];
```
originalblend337: I think the `children` prop is included within `RenderableProps`, so declaring this property here would be redundant.

frontends/web/src/app.tsx
```diff
@@ -14,50 +14,63 @@
...
+    public state = {
+        accounts: [],
         accountsInitialized: false,
         deviceIDs: [],
         devices: {},
         activeSidebar: false,
-    }
+    };
+
+    private unsubscribe?: () => void = undefined;
```
originalblend337: if we know that `unsubscribe` will definitely be defined (in this case `componentDidMount`?), you can declare it as `private unsubscribe!`. Although this works as is so if typescript yells at you for this I would leave it 👍

frontends/web/src/components/container/container.tsx
benma: TIL!

But when I remove it I get:

```
✖ ERROR [at-loader] ./src/components/container/container.tsx:32:15
    TS2533: Object is possibly 'null' or 'undefined'.
✖ ERROR [at-loader] ./src/components/container/container.tsx:32:24
    TS2339: Property 'map' does not exist on type 'string | number | object | VNode<any> | ComponentChild[]'.
  Property 'map' does not exist on type 'string'.
```

Any idea why?

benma: @KasparEtter let me know that RenderableProps children are optional and can be also other objects, according to `https://github.com/developit/preact/blob/87a4ebe99dc068eaeb8503644c60ffe8ad735771/src/preact.d.ts#L57`. So overriding it here and being explicit here is correct.

PR #277, opened 2019-01-08
frontend: add translation key to language switcher button

originalblend337:
I added a new translation key to the language button. As mentioned in the issue, this button will be 'Other languages' in English, and 'English' for other languages.


benma:
If it is those two fixed strings, they don't need to be in the locize files (needlessly putting the same 'English' into every language). Can just be hardcoded in the source, right?

originalblend337:
Thought this would be a safer solution in the case we want to update language. I’m ok with either, will go and hardcode it.

PR #290, opened 2019-01-17
Watch eth address

Tomasvrba:


backend/coins/btc/account.go
```diff
@@ -299,6 +299,7 @@ func (account *Account) Info() *accounts.Info {
 			account.signingConfiguration.ScriptType(),
 			account.signingConfiguration.AbsoluteKeypath(),
 			xpubs,
+			"",
```
benma: should be `account.signingConfiguration.Address()` (it is doing a copy of the signing configuration)

backend/coins/eth/account.go
```diff
@@ -156,9 +156,21 @@ func (account *Account) Initialize() error {
 	account.db = db
 	account.log.Debugf("Opened the database '%s' to persist the transactions.", dbName)
 
-	account.address = Address{
-		Address: crypto.PubkeyToAddress(*account.signingConfiguration.PublicKeys()[0].ToECDSA()),
+	addressBased := account.signingConfiguration.IsAddressBased()
+
+	if addressBased {
```
benma: just do `if account.signingConfiguration.IsAddressBased() {`? no need for a variable which is only used once.

backend/handlers/handlers.go
```diff
@@ -339,6 +343,25 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
...
 	keypath := signing.NewEmptyAbsoluteKeypath()
+
+	if jsonAddress != "" {
+		if !common.IsHexAddress(jsonAddress) {
+			return nil, errp.WithStack(errors.ErrInvalidAddress)
+		}
+
+		if jsonCoinCode == "teth" || jsonCoinCode == "eth" {
+			configuration := signing.NewAddressConfiguration(scriptType, keypath, jsonAddress)
+			getSigningConfiguration := func() (*signing.Configuration, error) {
```
benma: from here to line 358 is all duplicated code from, same lines appear below.

The only thing different is the configuration itself. So it makes sense to declare  `var configuration *signing.Configuration` above, and simply assign it here and below.

backend/handlers/handlers.go
```diff
@@ -339,6 +343,25 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
 		return nil, err
 	}
 	keypath := signing.NewEmptyAbsoluteKeypath()
+
+	if jsonAddress != "" {
+		if !common.IsHexAddress(jsonAddress) {
```
benma: eth address check should happen in the `if jsonCoinCode == "teth" || jsonCoinCode == "eth" { ` branch

backend/handlers/handlers.go
```diff
@@ -339,6 +343,25 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
...
+			return nil, errp.WithStack(errors.ErrInvalidAddress)
+		}
+
+		if jsonCoinCode == "teth" || jsonCoinCode == "eth" {
+			configuration := signing.NewAddressConfiguration(scriptType, keypath, jsonAddress)
+			getSigningConfiguration := func() (*signing.Configuration, error) {
+				return configuration, nil
+			}
```
benma: `} else { <return error> }`

External input should always be validated.

backend/signing/configuration.go
```diff
@@ -43,10 +44,11 @@ func NewConfiguration(
 	scriptType ScriptType,
 	absoluteKeypath AbsoluteKeypath,
 	extendedPublicKeys []*hdkeychain.ExtendedKey,
+	address string,
 	signingThreshold int,
 ) *Configuration {
-	if len(extendedPublicKeys) == 0 {
-		panic("A configuration has to contain at least one extended public key.")
+	if len(extendedPublicKeys) == 0 && len(address) == 0 || len(extendedPublicKeys) > 0 && len(address) > 0 {
```
benma: `address == ""` is simpler / more readable imho

backend/signing/configuration.go
```diff
@@ -68,7 +71,17 @@ func NewSinglesigConfiguration(
...
+}
+
+// NewAddressConfiguration creates a new account address configuration
+func NewAddressConfiguration(
+	scriptType ScriptType,
+	absoluteKeypath AbsoluteKeypath,
+	address string,
+) *Configuration {
+	return NewConfiguration(
+		scriptType, absoluteKeypath, []*hdkeychain.ExtendedKey{}, address, 1)
```
benma: maybe `0` instead (default value), as `signingThreshold` is not used in this configuration

backend/signing/configuration.go
```diff
@@ -89,6 +102,22 @@ func (configuration *Configuration) ExtendedPublicKeys() []*hdkeychain.ExtendedK
...
 }
 
+// Address returns the configuration's address
+func (configuration *Configuration) Address() string {
+	return configuration.address
+}
+
+// IsAddressBased returns whether configuration is address based or not
+func (configuration *Configuration) IsAddressBased() bool {
+	if len(configuration.extendedPublicKeys) > 0 {
```
benma: Just `return configuration.address != ""` suffices, as the `New*` constructors don't produce the cases you test here.

frontends/web/src/locales/en/app.json
```diff
@@ -28,6 +28,7 @@
       "xpubWrongNet": "Extended Public Key not valid for the selected account."
     },
     "extendedPublicKey": "Extended Public Key",
+    "ethAddress": "Ethereum Address",
```
benma: Maybe just `"address": "Address"`, so the text/translations can be reused for future Bitcoin/Litecoin/... addresses. The dropdown already makes it clear that it's Ethereum.

backend/signing/configuration.go
```diff
@@ -34,6 +34,7 @@ type Configuration struct {
 	absoluteKeypath    AbsoluteKeypath
```
benma: Update the Configuration docstring. 

Would also be good to describe the fields explicitly and when are defined/undefined, so the person who has to refactor it has an easier time ;)

frontends/web/src/routes/account/add/addaccount.tsx
```diff
@@ -169,11 +184,11 @@ class AddAccount extends Component<Props, State> {
                             </div>
                             <div class="row">
                                 <Input
-                                    label={t('addAccount.extendedPublicKey')}
-                                    onInput={linkState(this, 'extendedPublicKey')}
-                                    value={extendedPublicKey}
+                                    label={coinAndAccountCode === 'teth' || coinAndAccountCode === 'eth' ? t('addAccount.ethAddress') : t('addAccount.extendedPublicKey')}
```
benma: need to update `ethAddress`, it's `address` now. Same below.

backend/handlers/handlers.go
```diff
@@ -339,22 +342,35 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
...
-	if btcCoin, ok := coin.(*btc.Coin); ok {
-		expectedNet := &chaincfg.Params{
-			HDPublicKeyID: btc.XPubVersionForScriptType(btcCoin, scriptType),
+
+	var configuration *signing.Configuration
+
+	if jsonAddress != "" {
+		if jsonCoinCode == "teth" || jsonCoinCode == "eth" {
+			if !common.IsHexAddress(jsonAddress) {
+				return nil, errp.WithStack(errors.ErrInvalidAddress)
```
benma: Smeonean invalid address? This leads to a crash in the frontend in webdev mode, and is silently ignored in the Qt app. See the examples below of how to send an error to the frontend, e.g.

```Go
 			return map[string]interface{}{"success": false, "errorCode": "invalidAddress"}, nil
```

(need to also add a i18n key in app/en.json for `invalidAddress`).

backend/handlers/handlers.go
```diff
@@ -369,6 +385,7 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
 		}, nil
 	}
 	return map[string]interface{}{"success": true, "accountCode": accountCode}, nil
+
```
benma: remove empty line :)

backend/handlers/handlers.go
```diff
@@ -339,22 +342,35 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
...
-		return map[string]interface{}{"success": false, "errorCode": "xpubInvalid"}, nil
-	}
-	if extendedPublicKey.IsPrivate() {
-		return map[string]interface{}{"success": false, "errorCode": "xprivEntered"}, nil
-	}
-	if btcCoin, ok := coin.(*btc.Coin); ok {
-		expectedNet := &chaincfg.Params{
-			HDPublicKeyID: btc.XPubVersionForScriptType(btcCoin, scriptType),
+
+	var configuration *signing.Configuration
```
benma: Nice :+1:

benma:
please run `make locize-fix` to format the i18n file. The CI check that should catch this was buggy, fixed here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/292

Sorry about that!

PR #291, opened 2019-01-17
backend: add desktop notifications for new transactions

benma:
Since notifications work the same way for all accounts (remember
already noified new txs), they all take a Notifier instance, with the
db/notification logic coming from the backend. For simplicity, one
bbolt db is used for all accounts.

The frontend is notified via a backend even of new txs, which in turns
calls the backend endpoint to create a user desktop notification. The
roundtrip is done so that the frontend can localize the text.


benma:
Needs testing on macOS and Windows7 and 10.

PR #299, opened 2019-01-22
Frontend coincontrol feedback

originalblend337:
Will need to add a guide entry separately if we want to elaborate on how to use coin control.

### DIFF
--- a/frontends/web/src/decorators/translate.ts
+++ b/frontends/web/src/decorators/translate.ts
@@ -32,14 +32,14 @@ interface InterpolationValues {
 /**
  * This interface describes the default value to return if no translation was found.
  */
-// interface DefaultValue {
-//     defaultValue: any;
-// }
+interface DefaultValue {
+    defaultValue: any;
+}
 
 /**
  * This interface models the various options of the translate function.
  */
-type TranslateOptions = InterpolationValues;
+type TranslateOptions = InterpolationValues | DefaultValue;
 
 /**
  * This type describes the translate function of 'react-i18next'.
--- a/frontends/web/src/decorators/translate.ts
+++ b/frontends/web/src/decorators/translate.ts
@@ -39,7 +39,7 @@ interface DefaultValue {
 /**
  * This interface models the various options of the translate function.
  */
-type TranslateOptions = InterpolationValues | DefaultValue;
+type TranslateOptions = InterpolationValues | DefaultValue | string;
 
 /**
  * This type describes the translate function of 'react-i18next'.
@@ -58,7 +58,7 @@ interface HOCOptions {
     withRef?: boolean;
 }
 
-type Namespaces = string | string[] | null;
+type Namespaces = string | string[];
 
 type NamespacesFunction<ProvidedProps extends ObjectButNotFunction> = (props: ProvidedProps) => Namespaces;
 
--- a/frontends/web/src/decorators/translate.ts
+++ b/frontends/web/src/decorators/translate.ts
@@ -32,14 +32,14 @@ interface InterpolationValues {
 /**
  * This interface describes the default value to return if no translation was found.
  */
-interface DefaultValue {
-    defaultValue: any;
-}
+// interface DefaultValue {
+//     defaultValue: any;
+// }
 
 /**
  * This interface models the various options of the translate function.
  */
-type TranslateOptions = InterpolationValues | DefaultValue;
+type TranslateOptions = InterpolationValues;
 
 /**
  * This type describes the translate function of 'react-i18next'.
@@ -54,7 +54,11 @@ export interface TranslateProps {
     t: Translate;
 }
 
-type Namespaces = string | string[];
+interface HOCOptions {
+    withRef?: boolean;
+}
+
+type Namespaces = string | string[] | null;
 
 type NamespacesFunction<ProvidedProps extends ObjectButNotFunction> = (props: ProvidedProps) => Namespaces;
 
@@ -64,11 +68,12 @@ type NamespaceOptions<ProvidedProps extends ObjectButNotFunction> = Namespaces |
  * This function provides type safety for the 'react-i18next' translate decorator.
  */
 export function translate<ProvidedProps extends ObjectButNotFunction = {}>(
-    options?: NamespaceOptions<ProvidedProps>,
+    namespace?: NamespaceOptions<ProvidedProps>,
+    options?: HOCOptions,
 ) {
     return function decorator(
         WrappedComponent: ComponentFactory<TranslateProps & ProvidedProps>,
     ): ComponentFactory<ProvidedProps> {
-        return originalTranslate(options)(WrappedComponent);
+        return originalTranslate(namespace, options)(WrappedComponent);
     };
 }
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -657,9 +657,9 @@
     },
     "button": "Sign and Send",
     "coincontrol": {
-      "title": "Send from output",
       "address": "Address",
-      "outpoint": "Outpoint"
+      "outpoint": "Outpoint",
+      "title": "Send from Output"
     },
     "confirm": {
       "selected-coins": "Selected Coins",
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -657,6 +657,7 @@
     },
     "button": "Sign and Send",
     "coincontrol": {
+      "title": "Send from output",
       "address": "Address",
       "outpoint": "Outpoint"
     },
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Component, h } from 'preact';
+import { Component, h, RenderableProps } from 'preact';
 import { route } from 'preact-router';
 import reject from '../../../assets/icons/cancel.svg';
 import approve from '../../../assets/icons/checked.svg';
@@ -34,30 +34,30 @@ import { apiWebsocket } from '../../../utils/websocket';
 import { isBitcoinBased } from '../utils';
 import FeeTargets from './feetargets';
 import * as style from './send.css';
-import { SelectedUTXOProps, UTXOs } from './utxos';
+import { SelectedUTXO, UTXOs } from './utxos';
 
 interface SendProps {
-    accounts: AccountProps[];
+    accounts: Account[];
     code?: string;
     deviceIDs: string[];
 }
 
-interface AccountProps {
+interface Account {
     code: string;
     coinCode: string;
 }
 
-interface ProposedAmountProps {
-    amount: string | null;
-    unit: string | null;
-    conversions: ConversionsProps | null;
+interface ProposedAmount {
+    amount: string;
+    unit: string;
+    conversions: Conversions;
 }
 
-interface ConversionsProps {
-    [key: string]: string | null;
+interface Conversions {
+    [key: string]: string;
 }
 
-interface SignProgressProps {
+interface SignProgress {
     steps: number;
     step: number;
 }
@@ -65,12 +65,12 @@ interface SignProgressProps {
 type Props = SendProps & TranslateProps;
 
 interface State {
-    account: AccountProps | null | undefined;
+    account: Account | null | undefined;
     balance: BalanceInterface | null;
-    proposedFee: ProposedAmountProps | null;
-    proposedTotal: ProposedAmountProps | null;
+    proposedFee: ProposedAmount | null;
+    proposedTotal: ProposedAmount | null;
     recipientAddress: string | null;
-    proposedAmount: ProposedAmountProps | null;
+    proposedAmount: ProposedAmount | null;
     valid: boolean;
     amount: string | null;
     data: string | null;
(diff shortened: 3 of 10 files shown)


frontends/web/src/decorators/translate.ts
```diff
@@ -32,14 +32,14 @@ interface InterpolationValues {
 /**
  * This interface describes the default value to return if no translation was found.
  */
-// interface DefaultValue {
-//     defaultValue: any;
-// }
+interface DefaultValue {
```
benma: What does this commit do, why is this needed?

benma:
CI: 

> i18n files malformatted. Fix with: make locize-fix

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -16,33 +16,93 @@
...
+    deviceIDs: string[];
+}
+
+interface AccountProps {
+    code: string;
+    coinCode: string;
+}
+
+interface ProposedAmountProps {
+    amount: string | null;
```
KasparEtter: I would quite strongly prefer to use `amount?: string` (i.e. with the implicit `undefined`). In case you agree and just tried to model how the component was implemented before using TypeScript in this commit, then it's ok for me to leave this for a future commit.

KasparEtter: And on a different note: Shouldn't the logic be that *if* an `Amount` is provided, then the amount (and unit) field has to be set? It seems to me that you are making too many things optional here.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -16,33 +16,93 @@
...
+    unit: string | null;
+    conversions: ConversionsProps | null;
+}
+
+interface ConversionsProps {
```
KasparEtter: Remove `Props`.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -16,33 +16,93 @@
...
+    [key: string]: string | null;
```
KasparEtter: Do we really want to allow the object to be `{ "value": null }`? All values are already optional in the sense that you can simply not list them at all. Moreover, if I understand this correctly (and a more specific type would really help in this regard), you could use the following type for the `key`: https://github.com/digitalbitbox/bitbox-wallet-app/blob/06536915b08a3added0d7d3106e03bedc2a22225/frontends/web/src/components/rates/rates.tsx#L32

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -16,33 +16,93 @@
...
+}
+
+interface SignProgressProps {
```
KasparEtter: No `Props`. And do I interpret this correctly that `steps` is `totalSteps` and `step` is `currentStep`? Again, I guess you just modelled the existing implementation so leaving this for a separate commit/PR is fine.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -16,33 +16,93 @@
...
+    addressError: string | null;
+    amountError: string | null;
+    dataError: string | null;
+    paired: boolean | null;
+    noMobileChannelError: boolean;
+    signProgress: SignProgressProps | null;
+    signConfirm: boolean | null;
+    coinControl: boolean;
+    activeCoinControl: boolean;
+}
```
KasparEtter: I did not study (yet) whether the right fields are optional but as above I would prefer to make them really optional (with the `?`) so that you don't have to list them all if you declare the object.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -16,33 +16,93 @@
...
-export default class Send extends Component {
+class Send extends Component<Props, State> {
+    private utxos!: Element;
+    private selectedUTXOs!: SelectedUTXOProps;
+    private unsubscribe!: () => void;
 
     constructor(props) {
         super(props);
-
+        this.selectedUTXOs = {};
```
KasparEtter: I guess you could just declare it this way which would also make the `!` unnecessary: `private selectedUTXOs: SelectedUTXOProps = {};`.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -65,24 +125,24 @@ export default class Send extends Component {
...
             activeCoinControl: false,
+            data: null,
         };
-        this.selectedUTXOs = [];
     }
 
-    coinSupportsCoinControl = () => {
-        const account = this.getAccount();
+    private coinSupportsCoinControl = () => {
+        const { account } = this.state;
```
KasparEtter: Is this preferable to `const account = this.state.account;`? And since you use the account twice at most, I feel that the local const is no longer really necessary.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -154,13 +213,15 @@ export default class Send extends Component {
                     data: null,
                 });
                 if (this.utxos) {
+                    // @ts-ignore
```
KasparEtter: I would prefer if you cast the `getWrappedInstance()` (Which I guess is necessary? Or what is its type?). After you made `clear()` public, shouldn't the following (or at least something like it) work?

`(this.utxos.getWrappedInstance() as UTXOs).clear();`

This would have the advantage that TypeScript would help us detect if the method changed (its name or arguments) or is no longer there.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -298,89 +364,101 @@ export default class Send extends Component {
...
-        activeCoinControl,
-    }) {
-        const account = this.getAccount();
-        if (!account) return null;
+    private setUTXOsRef = (ref: Element) => {
+        this.utxos = ref;
+    }
+
+    public render(
+        { t, code }: Props,
```
KasparEtter: I think `RenderableProps<Props>` is always a good idea.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -298,89 +364,101 @@ export default class Send extends Component {
...
+                <br />
                 {t('send.signprogress.label')}: {signProgress.step}/{signProgress.steps}
             </span>
         ) : null;
         return (
             <div class="contentWithGuide">
                 <div class="container">
                     <Status type="warning">
-                        {paired === false && t('warning.sendPairing')}
+                        {paired === false ? t('warning.sendPairing') : null}
```
KasparEtter: As mentioned before, upgrading Preact should make this change unnecessary.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -298,89 +364,101 @@ export default class Send extends Component {
...
                     </Status>
                     <Header title={<h2>{t('send.title')}</h2>}>
-                        <Balance
-                            balance={balance} />
+                    <Balance balance={balance} />
```
KasparEtter: If the `Balance` is in the `Header`, it should be indented, shouldn't it?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -392,13 +470,14 @@ export default class Send extends Component {
                     <div class="innerContainer scrollableContainer">
                         <div class="content padded">
                             {
-                                coinControl && (
+                                coinControl ? (
```
KasparEtter: Dito regarding Preact.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -410,11 +489,13 @@ export default class Send extends Component {
                                     value={recipientAddress}
                                     autofocus
                                 />
-                                { debug && (
-                                    <span id="sendToSelf" className={style.action} onClick={this.sendToSelf}>
-                                        Send to self
-                                    </span>
-                                ) }
+                                {
+                                    debug ? (
```
KasparEtter: Dito regarding Preact.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -466,19 +547,24 @@ export default class Send extends Component {
                                     />
                                     */}
                                 </div>
-                                {feeTarget && <p class={style.feeDescription}>{t('send.feeTarget.description.' + feeTarget)}</p>}
+                                {
+                                    feeTarget ? (
```
KasparEtter: Dito regarding Preact.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -466,19 +547,24 @@ export default class Send extends Component {
...
-                                    <Input
-                                        label={t('send.data.label')}
-                                        placeholder={t('send.data.placeholder')}
-                                        id="data"
-                                        error={dataError}
-                                        onInput={this.handleFormChange}
-                                        value={data}
-                                    />
-                                </div>
+                                (account.coinCode === 'eth' || account.coinCode === 'teth' || account.coinCode === 'reth') ? (
```
KasparEtter: Dito regarding Preact.

KasparEtter: Also a few more occurrences of this below which I will no longer comment.

frontends/web/src/locales/en/app.json
```diff
@@ -654,6 +654,7 @@
     },
     "button": "Sign and Send",
     "coincontrol": {
+      "title": "Send from output",
```
KasparEtter: I haven't studied where this is used but if it says `title` it should probably be in title-case (with a capital `O` in output).

frontends/web/src/decorators/translate.ts
KasparEtter: Ah, then please ignore (parts of) my comments before.

frontends/web/src/routes/device/deviceswitch.tsx
```diff
@@ -24,6 +24,7 @@ export interface Devices {
 }
 
 interface Props {
+    key?: any;
```
originalblend337: @KasparEtter @benma  -- after updating preact to the latest, TS has been complaining that it cannot assign `devices` to the property `key` (occurs in `app.tsx`), do you know a better work around for this than defining it as `any`?

KasparEtter: Can you post the exact error message here? Or otherwise tell me what I need to look for resp. change if I check out your branch?

originalblend337: First error I get is
```
✖ ERROR [at-loader] ./src/app.tsx:181:29
    TS2326: Types of property 'key' are incompatible.
  Type 'Devices' is not assignable to type 'string | number | undefined'.
    Type 'Devices' is not assignable to type 'number'.
✖ ERROR [at-loader] ./src/app.tsx:186:29
    TS2326: Types of property 'key' are incompatible.
  Type 'Devices' is not assignable to type 'string | number | undefined'.
    Type 'Devices' is not assignable to type 'number'.
```

and when I try to define key to be `Devices`, which is
```
interface Devices {
    [deviceID: string]: string;
}
```
I get
```
✖ ERROR [at-loader] ./src/app.tsx:181:29
    TS2326: Types of property 'key' are incompatible.
  Type 'Devices' is not assignable to type '(Devices & string) | (Devices & number) | undefined'.
    Type 'Devices' is not assignable to type 'Devices & number'.
      Type 'Devices' is not assignable to type 'number'.
✖ ERROR [at-loader] ./src/app.tsx:186:29
    TS2326: Types of property 'key' are incompatible.
  Type 'Devices' is not assignable to type '(Devices & string) | (Devices & number) | undefined'.
    Type 'Devices' is not assignable to type 'Devices & number'.
      Type 'Devices' is not assignable to type 'number'.
```

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -65,26 +124,26 @@ export default class Send extends Component {
...
         if (this.props.deviceIDs.length > 0) {
-            apiGet('devices/' + this.props.deviceIDs[0] + '/has-mobile-channel').then(mobileChannel => {
+            apiGet('devices/' + this.props.deviceIDs[0] + '/has-mobile-channel').then((mobileChannel: boolean) => {
                 apiGet('devices/' + this.props.deviceIDs[0] + '/info').then(({ pairing }) => {
-                    const account = this.getAccount();
+                    const account = this.state.account;
+                    const paired = mobileChannel && pairing;
                     const noMobileChannelError = pairing && !mobileChannel && account && isBitcoinBased(account.coinCode);
-                    this.setState({ paired: mobileChannel && pairing, noMobileChannelError });
+                    this.setState(prevState => ({ ...prevState, paired, noMobileChannelError }));
```
KasparEtter: Why do we need the `prevState`? Why not just `this.setState({ paired, noMobileChannelError });`?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -14,35 +14,94 @@
...
+import { debug } from '../../../utils/env';
+import { apiGet, apiPost } from '../../../utils/request';
+import { apiWebsocket } from '../../../utils/websocket';
+import { isBitcoinBased } from '../utils';
 import FeeTargets from './feetargets';
-import UTXOs from './utxos';
-import approve from '../../../assets/icons/checked.svg';
-import reject from '../../../assets/icons/cancel.svg';
 import * as style from './send.css';
+import { Props as UTXOsProps, SelectedUTXO, UTXOs } from './utxos';
```
KasparEtter: Why do you import `Props` (instead of `UTXOsProps `)? It is declared as `export type Props = UTXOsProps & TranslateProps` and we don't need the `t` here.

PR #300, opened 2019-01-22
add account: warn instead of abort if xpub does not match the net

KasparEtter:


frontends/web/src/locales/en/app.json
```diff
@@ -27,13 +27,15 @@
...
-      "xpubWrongNet": "Extended Public Key not valid for the selected account."
+      "xpubInvalid": "Extended Public Key malformatted."
     },
     "extendedPublicKey": "Extended Public Key",
     "scriptType": "Script Type",
     "submit": "Add Account",
-    "title": "Add Account"
+    "title": "Add Account",
+    "warning": {
+      "xpubWrongNet": "The extended public key does not match the selected account but, since not all apps export extended public keys correctly, we created the account anyway."
```
benma: Better wording:

`"Warning: the Extended Public Key format indicates that it might not valid for the selected account type. It you have trouble with the account, try the other account types."`

PR #306, opened 2019-01-24
Makefile: use MAKE variable

toxeus:
Otherwise, flags such as `-j` are not propagated which would benefit the compilation of the Qt part.

See:
https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html


toxeus:
A minor thing I've noticed while playing around with your dev environment. I'm assuming that the CI covers most of the Makefiles :angel:

benma:
TIL! Thanks!

PR #309, opened 2019-01-30
frontend: update change pin button text to be titlecase

originalblend337:


benma:
Note [i18n.md](https://github.com/digitalbitbox/bitbox-wallet-app/blob/53fa1cca6c0526eaf678b1752ca0ca22d19633e8/docs/i18n.md):

> Changing values of any language (reference or otherwise) must be done inside of locize.

I did that quickly, so after a locize pull, it would result in the same commit ;) In the future, it should go via locize first, to not lose the changes.

PR #318, opened 2019-02-07
bitbox-api: add version to device info API call

Tomasvrba:


backend/devices/bitbox02/device.go
```diff
@@ -209,24 +215,28 @@ func (device *Device) SetDeviceName(deviceName string) error {
...
 			DeviceInfo: &messages.DeviceInfoRequest{},
 		},
 	}
 
+	deviceInfo := &DeviceInfo{}
+
 	response, err := device.query(request)
 	if err != nil {
-		return "", err
+		return deviceInfo, err
```
benma: `return nil, err` (if err != nil, all other return values are to be ignored anyway).

then you can define the above `deviceInfo` further down.

backend/devices/bitbox02/device.go
```diff
@@ -209,24 +215,28 @@ func (device *Device) SetDeviceName(deviceName string) error {
...
 	}
 
-	deviceInfo, ok := response.Response.(*messages.Response_DeviceInfo)
+	deviceInfoResponse, ok := response.Response.(*messages.Response_DeviceInfo)
 	if !ok {
-		return "", errp.New("Failed to retrieve device name")
+		return deviceInfo, errp.New("Failed to retrieve device info")
```
benma: same

frontends/web/src/routes/device/settings/components/deviceinfo.jsx
```diff
@@ -36,7 +36,8 @@ export default class DeviceInfo extends Component {
         apiGet(this.props.apiPrefix + '/get-info').then(info => {
             this.setState({
                 active: true,
-                deviceInfo: info,
+                deviceName: info.name,
```
benma: if you keep `deviceInfo`, then you have access to all its contents easily.

backend/devices/bitbox02/device.go
```diff
@@ -218,15 +224,18 @@ func (device *Device) DeviceInfo() (string, error) {
...
+		return nil, err
 	}
 
-	deviceInfo, ok := response.Response.(*messages.Response_DeviceInfo)
+	deviceInfoResponse, ok := response.Response.(*messages.Response_DeviceInfo)
 	if !ok {
-		return "", errp.New("Failed to retrieve device name")
+		return nil, errp.New("Failed to retrieve device info")
 	}
+	deviceInfo := &DeviceInfo{}
```
benma: not important, just  fyi: you can init directly like `&DeviceInfo{ Name: ..., Version: ... }`.

frontends/web/src/routes/device/settings/components/deviceinfo.jsx
```diff
@@ -28,15 +28,15 @@ export default class DeviceInfo extends Component {
         super(props);
         this.state = {
             active: false,
-            deviceInfo: 0,
+            deviceInfo: '',
```
benma: better to not mix string type with object type. init like deviceInfo: undefined.

PR #323, opened 2019-02-14
bitbox02: initialization worfklow

benma:


frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -34,43 +39,117 @@ interface Props {
...
-            </div>
-        );
+    public render({ deviceID }: RenderableProps<Props>, { hash, deviceVerified, status, settingPassword, creatingBackup }: State) {
+        switch (status) {
+            case 'unpaired':
+                return (
+                    <Dialog onClose={() => this.channelVerify(false)}>
+                        <p>Verify your BitBox</p>
+                        <p>{hash}</p>
+                        <Button primary onClick={() => this.channelVerify(true)} disabled={!deviceVerified}>Correct</Button>
```
Tomasvrba: I was going to put this on the firmware PR, but that one is merged now :)
In `static void _confirm(component_t* component)`, in pairing.c, the screen is popped from the device before user can press confirm on the app.
When confirming on app, user can't see the code on the device anymore. The app could potentially change the code the user is confirming with no way to compare anymore.

PR #325, opened 2019-02-19
Sys time api

Tomasvrba:


frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -15,14 +15,14 @@
...
 import { Component, h, RenderableProps } from 'preact';
-import DeviceInfo from '../../../routes/device/settings/components/deviceinfo';
 import RandomNumber from '../../../routes/device/settings/components/randomnumber';
-import SetDeviceName from '../../../routes/device/settings/components/setdevicename';
 import { apiGet, apiPost } from '../../../utils/request';
 import { apiWebsocket } from '../../../utils/websocket';
 import { alertUser } from '../../alert/Alert';
 import { Dialog } from '../../dialog/dialog';
 import { Button } from '../../forms';
+import DeviceInfo from './deviceinfo';
```
benma: nice!

backend/devices/bitbox02/device.go
```diff
@@ -338,9 +339,18 @@ func (device *Device) CreateBackup() error {
 	if device.status != StatusSeeded {
 		return errp.New("invalid status")
 	}
+
+	timestamp := time.Now().Unix()
```
benma:
```
now := time.Now()
_, offset := now.Zone()
```
and inline the other vars? i'm a fan of not having single-use vars (and the naming questions fall away).

backend/devices/bitbox02/device.go
```diff
@@ -338,9 +339,17 @@ func (device *Device) CreateBackup() error {
 	if device.status != StatusSeeded {
 		return errp.New("invalid status")
 	}
+
+	now := time.Now()
+	timestamp := now.Unix()
```
benma: nit: same comment as before, no need for a var that is used only once. feel free to keep or change ;)

PR #326, opened 2019-02-20
frontend/pairing: display confirmed device hash in app

Tomasvrba:


frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -113,6 +114,15 @@ class BitBox02 extends Component<Props, {}> {
                     <Dialog onClose={() => this.channelVerify(false)}>
                         <p>Verify your BitBox</p>
                         <p>{hash}</p>
+
+                        { !deviceVerified ?
+                            <p>First, verify the hash on your device</p>
+                        :
+                        <div className={'flex flex-1 flex-row flex-items-center spaced'}>
+                            <p>Device Hash: {hash}</p>
```
benma: Unsure what this does. The point was that the hash is displayed on both sides, and after it's confirmed on the device, the hash is not shown anymore on the screen (because it is not visible anymore on the device).

Here the hash is shown twice after confirm on the device?

After device confirmation, the app should ask the user if they confirmed or rejected on the device, so the app confirmation works without showing the hash after the device confirmation/rejection.

Tomasvrba:
Closing this PR, @originalblend337 will implement something along these lines in the UI. 
Eventually, we might want to make a API calls to get the actual hash confirmed on the device

PR #327, opened 2019-02-22
account: add support for watch-only btc/ltc addresses

Tomasvrba:


backend/coins/btc/addresses/addressonly.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2018 Shift Devices AG
```
benma: rename file to singleaddress.go - filenames usually match the struct name

backend/coins/btc/addresses/addressonly.go
```diff
@@ -0,0 +1,96 @@
...
+	if len(addresses.address) == 0 {
+		addresses.log.Panic("Concurrency error: Address not synced correctly")
+	}
+	return addresses.address
+}
+
+// addAddress stores the address.
+func (addresses *SingleAddress) addAddress() *AccountAddress {
+	addresses.log.Debug("Add new address to chain")
+	address := NewAccountAddress(
```
benma: all of this can be done in account.Init(), and passed to NewSingleAddress(). Then you don't have to pass the signing configuration to NewSingleAddress(), as it's not used otherwise.

backend/handlers/handlers.go
```diff
@@ -379,6 +382,21 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
 			}
 			configuration = signing.NewAddressConfiguration(scriptType, keypath, jsonAddress)
 		}
+		if jsonCoinCode == "btc" || jsonCoinCode == "ltc" {
+			_, err := btcutil.DecodeAddress(jsonAddress, &chaincfg.MainNetParams)
```
benma: you want &ltc.MainNetParams for litecoin, see https://github.com/digitalbitbox/bitbox-wallet-app/blob/83e9f2daf065b60196f80cbec56e143dd00f904f/backend/backend.go#L384

could make a var `params` and switch on jsonCoinCode to assign it to the correct one.

eventually it should be by just calling sth like `backend.Coin(jsonCoinCode).DecodeAddress(...)`

backend/coins/btc/addresses/address.go
```diff
@@ -53,71 +53,81 @@ type AccountAddress struct {
 
 // NewAccountAddress creates a new account address.
 func NewAccountAddress(
+	singleAddress btcutil.Address,
```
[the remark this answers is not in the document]
Tomasvrba: Yes, but account `accountConfiguration` has the address as string, not `btcutil.Address` so I'd have to decode it here, instead of in account.go. 
Thought we previously talked about doing it this way, no problem changing it though if it makes more sense to decode it here instead

backend/coins/btc/addresses/addressonly.go
Tomasvrba: I'm not exactly sure what you mean by "all of this", github doesn't really tell me which lines it's referring to. The addAddress() function? 
My goal was to separate things into a separate file from account.go
I'll probably have to take a look at this with you :)

backend/handlers/handlers.go
Tomasvrba: Made the switch, marked the eventual feature in a TODO

PR #329, opened 2019-02-25
frontend/version: move the version to a separate component and use a HOC

KasparEtter:
This was just a TypeScript and decorator example for @Tomasvrba.


Tomasvrba:
Great example :)

PR #335, opened 2019-03-05
Account summary

Tomasvrba:


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+interface Response {
+    accounts: AccountAndBalanceInterface[];
+    totals: Totals;
+}
+
+type Props = TranslateProps & AccountSummaryProps;
+
+class AccountsSummary extends Component<Props, State> {
+    private groupByCoin(accounts: AccountAndBalanceInterface[], coinCode: string) {
+        return accounts.reduce((accumulator, currentValue) => {
```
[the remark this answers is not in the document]
KasparEtter: Once you do this, I will understand what `key` actually is.

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+            if (!accumulator[key]) {
+                accumulator[key] = [];
+            }
+            accumulator[key].push(currentValue);
+            return accumulator;
+        }, {});
+    }
+
+    public componentDidMount() {
+        this.setState({ exported: ''} );
```
KasparEtter: Maybe you copied this from somewhere else but shouldn't we rather initialize the state in the constructor?

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+    private export = () => {
+        apiPost(`export-account-summary`).then(exported => {
+            this.setState({ exported });
+        });
+    }
+
+    public render(
+        { t, data }: RenderableProps<Props>, { exported }: State,
+    ) {
+            const groupedAccounts = this.groupByCoin(data.accounts, 'coinCode');
```
KasparEtter: As far as I can tell, `groupByCoin` is used only here. Is there a reason why you parameterized `'coinCode'` with `coinCode: string`?

KasparEtter: I'll verify once I checked out this PR but I think `groupedAccounts` might be untyped here.

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -0,0 +1,115 @@
...
+            const coins = Object.keys(groupedAccounts);
+            return (
+                <div className="contentWithGuide">
```
KasparEtter: Do we also use this class if there is no guide?

frontends/web/src/routes/account/summary/balancerow.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return (
+        <tr>
+            <td>{name}</td>
+            <td>{balance.amount}</td>
+            <td><FiatConversion amount={balance}/></td>
+        </tr>
+    );
+};
+
+const HOC = translate<ProvidedProps>()(BalanceRow);
```
KasparEtter: You don't need the `t`, which makes this component even simpler.

frontends/web/src/routes/account/summary/balancerow.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { translate, TranslateProps } from '../../../decorators/translate';
+
+interface ProvidedProps {
+    name: string;
+    balance: Amount;
+}
+
+type Props = ProvidedProps & TranslateProps;
+
+const BalanceRow = ({ name, balance }: RenderableProps<Props>) => {
```
KasparEtter: Is there a reason why you don't write `function BalanceRow({ name, balance }: RenderableProps<Props>) {`? I don't care too much about this but we seem to be somewhat inconsistent with this when compared to, for example, `update.tsx`.

frontends/web/src/routes/account/summary/balancestable.tsx
```diff
@@ -0,0 +1,75 @@
...
+    total: Amount;
+}
+
+type Props = ProvidedProps & TranslateProps;
+
+class BalancesTable extends Component<Props> {
+    public render(
+        { t, coinCode, accounts, total }: RenderableProps<Props>,
+    ) {
+        const rows = accounts.map(account => {
```
KasparEtter: Why don't you map the accounts inline where used?

frontends/web/src/routes/account/summary/balancestable.tsx
```diff
@@ -0,0 +1,75 @@
...
+            return <BalanceRow name={account.name} balance={account.balance.available}/>;
```
KasparEtter: … and use the expression directly without the additional `return` statement.

frontends/web/src/routes/account/summary/balancestable.tsx
```diff
@@ -0,0 +1,75 @@
...
+        });
+        return (
+            <div>
+                <div className={style.coinheader}>
+                    <div className="subHeader" class="row">
+                        <div class="flex flex-1 flex-row flex-between flex-items-center flex-center" height={64}>
+                            <Logo className={style.coinlogo} coinCode={coinCode} alt={coinCode} active={true} ></Logo>
```
KasparEtter: Close `Logo` directly with `/>`?

frontends/web/src/routes/account/summary/balancestable.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    <tfoot>
+                        <tr>
+                            <th>{t('accountSummary.total')}</th>
+                            <th>{total.amount}</th>
+                            <th><FiatConversion amount={total} /></th>
+                        </tr>
+                    </tfoot>
+                </table>
+            </div>
+            );
```
KasparEtter: The indentation is wrong here, isn't it?

frontends/web/src/routes/account/summary/initializeall.tsx
```diff
@@ -0,0 +1,106 @@
...
+        this.props.accounts.map((account: AccountInterface) => {
+            initialized[account.code] = false;
+        });
+        this.state = { initialized };
+      }
+
+    private unsubscribe!: () => void;
+
+    public componentDidMount() {
+        this.props.accounts.map((account: AccountInterface) => {
```
KasparEtter: Shouldn't those three lines (without the API subscription) also be executed on `componentDidUpdate()`?

frontends/web/src/routes/account/summary/initializeall.tsx
```diff
@@ -0,0 +1,106 @@
...
+            this.onStatusChanged(account.code);
+        });
+        this.unsubscribe = apiWebsocket(this.onEvent);
+    }
+
+    private allInitialized() {
+        return Object.keys(this.state.initialized).every(key => this.state.initialized[key]) ? true : false;
```
KasparEtter: `return booleanExpression ? true : false;` 🤔

frontends/web/src/routes/account/summary/initializeall.tsx
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private onEvent = data => {
```
KasparEtter: `data` is implicitly `any`, which we would have to fixe before activating `noImplicitAny`.

frontends/web/src/routes/account/summary/initializeall.tsx
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    private onStatusChanged(code) {
+        apiGet(`account/${code}/status`).then(status => {
+            const accountSynced = status.includes('accountSynced');
+            if (!accountSynced && !status.includes('accountDisabled')) {
+                apiPost(`account/${code}/init`);
+            }
+            this.setState(state => {
```
KasparEtter: I think we discussed this: An alternative would be to "flatten" the state so that you don't have to overwrite the whole `initialized` but could simply call `this.setState({ [code]: accountSynced })`. I'm not sure whether I would favor such an approach, though, because if the state grows again at some point, then we might run into subtle name conflicts if we don't keep the account codes separate.

backend/handlers/handlers.go
```diff
@@ -660,3 +667,115 @@ func (handlers *Handlers) apiMiddleware(devMode bool, h func(*http.Request) (int
...
+		"Unit",
+	})
+	if err != nil {
+		return nil, errp.WithStack(err)
+	}
+
+	for _, account := range handlers.backend.Accounts() {
+		coin := account.Coin().Code()
+		accountName := account.Name()
+		balance := account.Balance().Available().BigInt().String()
```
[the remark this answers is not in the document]
Tomasvrba: Good point :)

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -60,6 +61,14 @@ function Sidebar(
                 <div className="sidebarLogoContainer">
                     <BitBoxInverted className="sidebarLogo" />
                 </div>
+                <div className="sideBarItem">
+                    <Link activeClassName="sidebar-active" class="settings" href={`/account-summary`} title="Account Summary">
+                            <div className="single">
+                                <img draggable={false} className="sidebar_settings" src={info} alt={t('sidebar.addAccount')} />
```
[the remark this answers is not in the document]
Tomasvrba: It already does. It lights up when you hover over it if you check it out in the app

frontends/web/src/routes/account/summary/initializeall.tsx
Tomasvrba: Good catch, that was a pretty dumb line :D

PR #342, opened 2019-03-08
bitbox-api: add blocking calls to insert/remove sd card

Tomasvrba:


backend/devices/bitbox02/device.go
```diff
@@ -564,3 +564,23 @@ func (device *Device) BTCSign(
 		}
 	}
 }
+
+// InsertRemoveSDCard sends a command to the device to insert of remove the sd card based on the workflow state
+func (device *Device) InsertRemoveSDCard(action messages.InsertRemoveSDCardRequest_SDCardAction) (bool, error) {
```
benma: same: remove `bool` return

Tomasvrba: Thanks for the reminder. Fixed.

PR #348, opened 2019-03-12
Qr

benma:


thisconnect:
looks like it is nodejs v6.x on travis osx

https://travis-ci.org/digitalbitbox/bitbox-wallet-app/jobs/505613993#L131

btw I suggest to use the LTS nodejs versions 10.x 
(dont use 9.x, 11.x unless you develop a node module then use 11.x)

benma:
Good tip, thanks @thisconnect . Trying 10.13 now.

PR #354, opened 2019-03-14
account-summary: show account sync progress m/n synced

Tomasvrba:


frontends/web/src/routes/account/summary/initializeall.tsx
```diff
@@ -91,13 +91,15 @@ class InitializeAllAccounts extends Component<Props, State> {
...
-    public render({ t }: RenderableProps<Props>) {
+    public render({ t }: RenderableProps<Props>, { initialized }: State) {
         if (this.allInitialized()) {
             return <AccountsSummary />;
         }
         return (
             <div>
-                <Spinner text={t('accountSummary.synchronizing')}/>
+                <Spinner text={t('accountSummary.synchronizing') +
+                Object.values(initialized).filter(status => status === true).length +
```
KasparEtter: `filter(status => status === true)` -> `filter(status => status)`?

frontends/web/src/routes/account/summary/initializeall.tsx
```diff
@@ -91,13 +91,15 @@ class InitializeAllAccounts extends Component<Props, State> {
...
+                `/${this.props.accounts.length}`}/>
```
KasparEtter: Rather use `render({ t, accounts }: RenderableProps<Props>, …` to be consistent.

frontends/web/src/locales/en/app.json
```diff
@@ -24,7 +24,7 @@
     "fiatBalance": "Fiat Balance",
     "name": "Account Name",
     "noAccount": "There are no accounts to show.",
-    "synchronizing": "Synchronizing all accounts...",
+    "synchronizing": "Synchronizing accounts with the blockchain: ",
```
KasparEtter: I forgot whether we are allowed to change values here. Only if they have not yet been translated? And maybe "with their [respective] blockchain"? Or shall we sacrifice accuracy for easier understanding?

Tomasvrba: I agree that "with their respective blockchains" would be nice to be technically correct, but I think we should go for simplicity in this case. Also easier to translate.

Tomasvrba: If we get a lot of hate for it from grammar nazis then we can change it :fearful:

Tomasvrba: Changed back here and made the change in locize

KasparEtter: Great, thanks!

PR #355, opened 2019-03-15
Info.plist improvements for macOS

KasparEtter:
Solved the camera access problem with the own build.

### DIFF
--- a/frontends/qt/resources/MacOS/Info.plist
+++ b/frontends/qt/resources/MacOS/Info.plist
@@ -40,5 +40,8 @@
 
 	<key>LSApplicationCategoryType</key>
 	<string>public.app-category.finance</string>
+
+	<key>NSCameraUsageDescription</key>
+	<string>The camera is used to scan QR codes of receive addresses.</string>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


frontends/qt/resources/MacOS/Info.plist
```diff
@@ -40,5 +40,8 @@
 
 	<key>LSApplicationCategoryType</key>
 	<string>public.app-category.finance</string>
+
+	<key>NSCameraUsageDescription</key>
```
benma: Does this just show the message, or does it actually enable the use of the camera (e.g. doesn't work if missing)?

Would be great to add this info to the commit message or as a comment here if possible.

benma:
```
The app’s Info.plist must contain an NSCameraUsageDescription key with a string value explaining to the user how the app uses this data.
```
https://useyourloaf.com/blog/privacy-settings-in-ios-10/

KasparEtter: Yes, this solved the issue I had quite reliably. I amended the commit message. ✅

PR #359, opened 2019-03-19
Minor fixes

Tomasvrba:


backend/handlers/handlers.go
```diff
@@ -765,11 +770,38 @@ func (handlers *Handlers) postExportAccountSummary(_ *http.Request) (interface{}
...
+			address = signingConfiguration.Address()
+			scriptType = ""
+		} else {
+			accountType = "xpub"
+			switch specificAccount := account.(type) {
+			case *btc.Account:
+				for _, xpub := range specificAccount.Info().SigningConfiguration.ExtendedPublicKeys() {
+					xpubs += xpub.String() + "; "
+				}
+				scriptType = string(specificAccount.Info().SigningConfiguration.ScriptType())
```
benma: not needed, it's already `""`.

backend/handlers/handlers.go
```diff
@@ -765,11 +770,38 @@ func (handlers *Handlers) postExportAccountSummary(_ *http.Request) (interface{}
...
+			case *eth.Account:
+				xpubs = specificAccount.Info().SigningConfiguration.ExtendedPublicKeys()[0].String()
+				address = specificAccount.EthPubKeyToAddress(specificAccount.Info().SigningConfiguration).Hex()
+				scriptType = ""
```
benma: same

backend/coins/btc/handlers/handlers.go
```diff
@@ -481,11 +482,16 @@ func (handlers *Handlers) postVerifyAddress(r *http.Request) (interface{}, error
...
 func (handlers *Handlers) getCanVerifyExtendedPublicKey(_ *http.Request) (interface{}, error) {
-	btcAccount, ok := handlers.account.(*btc.Account)
-	if !ok {
-		return nil, errp.New("An account must be BTC based to support xpub verification")
+	switch specificAccount := handlers.account.(type) {
+	case *btc.Account:
+		return specificAccount.CanVerifyExtendedPublicKey(), nil
+	case *eth.Account:
+		// TODO: implement secure verification for ETH addresses
+		return nil, nil
```
benma: doesn't this produce an error in the info screen in eth (`null.includes(index)`?).

In any case, for non btc accounts you can just return an empty slice. (no panic in default)

backend/handlers/handlers.go
```diff
@@ -751,6 +752,10 @@ func (handlers *Handlers) postExportAccountSummary(_ *http.Request) (interface{}
 		"Name",
 		"Balance",
 		"Unit",
+		"Type",
```
KasparEtter: I think I would prefer it without that column.

backend/handlers/handlers.go
```diff
@@ -765,11 +770,38 @@ func (handlers *Handlers) postExportAccountSummary(_ *http.Request) (interface{}
...
 		unit := account.Coin().SmallestUnit()
+		var accountType string
+		var xpubs string
+		var address string
+		var scriptType string
+		if signingConfiguration := account.Info().SigningConfiguration; signingConfiguration.IsAddressBased() {
+			accountType = "address"
+			address = signingConfiguration.Address()
+			scriptType = ""
+		} else {
```
KasparEtter: In case we have both the address and the xpub, shouldn't we add both? Thus instead of `else` have an `if` the xpub is set.

backend/handlers/handlers.go
KasparEtter: You mean the `string(…)` part is not needed?

backend/signing/configuration.go
```diff
@@ -47,9 +47,6 @@ func NewConfiguration(
 	address string,
 	signingThreshold int,
 ) *Configuration {
-	if len(extendedPublicKeys) == 0 && len(address) == 0 || len(extendedPublicKeys) > 0 && len(address) > 0 {
```
KasparEtter: Shouldn't we leave a check that at least one or the other is provided?

backend/coins/eth/account.go
```diff
@@ -119,7 +119,7 @@ func (account *Account) Info() *accounts.Info {
 			account.signingConfiguration.ScriptType(),
 			account.signingConfiguration.AbsoluteKeypath(),
 			account.signingConfiguration.ExtendedPublicKeys(),
-			account.signingConfiguration.Address(),
+			account.address.String(),
```
KasparEtter: I thought we wanted to add the address when creating the signing configuration for ETH?

backend/handlers/handlers.go
benma: i think i commented on a commit, and when looking a all commits, the comment is on the wrong line...

the comment applies for where `scriptType = ""`

backend/coins/eth/account.go
```diff
@@ -113,7 +113,16 @@ func NewAccount(
 
 // Info implements accounts.Interface.
 func (account *Account) Info() *accounts.Info {
-	return nil
+	return &accounts.Info{
+		// TODO: Refactor Signing Configuration so we don't need coin specific dummies
+		SigningConfiguration: signing.NewConfiguration(
```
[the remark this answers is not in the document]
benma: I see now, to add account.address.String()

backend/coins/eth/account.go
benma: Might be the best after all :/

backend/coins/eth/account.go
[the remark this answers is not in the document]
Tomasvrba: Yes, but since we are leaning to including the address in the configuration, I will change it here :)

backend/handlers/handlers.go
Tomasvrba: Yes, I had it so both were added, hmm, not sure what happened there :)

backend/coins/eth/account.go
Tomasvrba: @KasparEtter Would that now mean though we would have to switch on the account type in the keystores.go where the signing configuration is created for the accounts?
`return signing.NewConfiguration(`
        `scriptType, absoluteKeypath, extendedPublicKeys, "", signingThreshold), nil`
Since now we would have to create a configuration with an address for ethereum, but without otherwise

backend/handlers/handlers.go
Tomasvrba: Having the address in the signing configuration will change how this is done anyway

backend/coins/btc/handlers/handlers.go
Tomasvrba: Doesn't produce an error since `null.includes(index)` is never reached when there is an address and there always has to be either an address or an xpub. 
But you're right anyway, returning an empty slice instead is better for consistency.

backend/handlers/handlers.go
Tomasvrba: Resolved by making an account AddressBased only if it has an address AND no xpub on configuration.go 
`return configuration.address != "" && len(configuration.ExtendedPublicKeys()) == 0`

KasparEtter:
If you want, you can force push again and fix the typo in your last commit message. 😉

Tomasvrba:
:man_facepalming:

backend/coins/btc/handlers/handlers.go
```diff
@@ -481,11 +482,15 @@ func (handlers *Handlers) postVerifyAddress(r *http.Request) (interface{}, error
...
 
 func (handlers *Handlers) getCanVerifyExtendedPublicKey(_ *http.Request) (interface{}, error) {
-	btcAccount, ok := handlers.account.(*btc.Account)
-	if !ok {
-		return nil, errp.New("An account must be BTC based to support xpub verification")
+	switch specificAccount := handlers.account.(type) {
+	case *btc.Account:
+		return specificAccount.CanVerifyExtendedPublicKey(), nil
+	case *eth.Account:
+		// TODO: implement secure verification for ETH addresses
```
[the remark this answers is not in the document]
Tomasvrba: I suppose since the 'i' button now also displays addresses, not only xpubs, the function should be renamed to something like CanVerifyExtendedPublicKeyOrAddress XD
The change was necessary since eth accounts now also have the 'i' button, but it displays an address, not an xpub

KasparEtter:
Thanks, @Tomasvrba! @benma please merge if you are happy now as well. 🙂

backend/coins/btc/handlers/handlers.go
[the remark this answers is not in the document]
benma: Already have CanVerifyAddress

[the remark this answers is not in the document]
Tomasvrba: Yes, but not a can-verify-address endpoint. 
In that case we would need to distinguish in the frontend in the constructor of signingconfiguration.tsx whether an address or xpub will need to be verified and either call the appropriate endpoint or redo the current /can-verify-extended-public-key endpoint to be more general like I mentioned above.
What are your thoughts about that right now?

PR #363, opened 2019-03-20
backend/accountSummary: exclude accounts with errors

benma:


backend/coins/btc/account.go
```diff
@@ -434,6 +436,9 @@ outer:
 
 // Balance implements the interface.
 func (account *Account) Balance() *accounts.Balance {
+	if account.fatalError {
+		return accounts.NewBalance(coin.NewAmountFromInt64(0), coin.NewAmountFromInt64(0))
```
KasparEtter: I'm not a fan of "lying" to callers. I would prefer "proper" design by contract.

backend/coins/btc/account.go
```diff
@@ -549,6 +554,9 @@ func (account *Account) subscribeAddress(
 
 // Transactions wraps transaction.Transactions.Transactions()
 func (account *Account) Transactions() []accounts.Transaction {
+	if account.fatalError {
+		return []accounts.Transaction{}
```
KasparEtter: Dito. Do we call/need this? Or could we just panic (as part of the contract)?

backend/coins/btc/account.go
benma: done

backend/coins/btc/account.go
benma: done

PR #364, opened 2019-03-20
frontend: make locize-pull

benma:


KasparEtter:
My impression is that we need a locize-push first because some strings are lost with this PR.

benma:
Good catch.

PR #370, opened 2019-03-22
Forceverification

originalblend337:


frontends/web/src/routes/account/receive/receive.jsx
```diff
@@ -142,36 +161,77 @@ export default class Receive extends Component {
...
+                            </a>
                         )
                     }
                     <p class={style.label}>{t('receive.label')} { receiveAddresses.length > 1 ? `(${activeIndex + 1}/${receiveAddresses.length})` : ''}</p>
                     {
                         receiveAddresses.length > 1 && (
-                            <Button
-                                transparent
+                            <a
+                                href="!#"
```
benma: what is `!#`? (I've only seen `#` before)

frontends/web/src/components/copy/Copy.css
```diff
@@ -1,49 +1,52 @@
 .button {
```
benma: why did you change the style and position of the copy button? tbh I liked the previous version much better, it looked more polished.

frontends/web/src/routes/account/receive/receive.jsx
originalblend337: typo.. lol although there is such thing as a hashbang (#!), where it means something different. fixed.

frontends/web/src/components/copy/Copy.css
originalblend337: I'm still not certain that one looks better over another, but since other inputs such as the address field in send has the qr code inside the field, thought it may be good to have the same style across the app.

benma:
Nice work!

PR #378, opened 2019-04-01
backend/config: update infura API keys for clients with old URLs

Tomasvrba:


backend/config/config.go
```diff
@@ -217,6 +218,7 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) *Config
 		accountsConfig:         newDefaultAccountsonfig(),
 	}
 	config.load()
+	config.appConfig, _ = config.checkAPIKeys(config.appConfig)
```
benma: never ignore errors (rare exception when handling it is impossible or you know for a fact the error can't happen, but then it should be turned into a panic)

Tomasvrba: Good catch, fixed.

backend/config/config.go
```diff
@@ -217,7 +221,12 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) *Config
 		accountsConfig:         newDefaultAccountsonfig(),
 	}
 	config.load()
-	return config
+	var err error
+	config.appConfig, err = config.checkAPIKeys(config.appConfig)
```
benma: Code reuse: use `config.SetAppConfig()` which sets and saves `config.appConfig.`

Naming: `checkAPIKeys` does not check, but modify the config. please use `updateInfuraAPIKeys()` or similar.

Tomasvrba: Well, it checks and then modifies if the check fails :)
But you're right, `update` works better

Tomasvrba: Updated.

backend/config/config.go
```diff
@@ -217,7 +221,12 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) *Config
 		accountsConfig:         newDefaultAccountsonfig(),
 	}
 	config.load()
-	return config
+	appConfig, err := config.updateInfuraAPIKeys(config.appConfig)
+	if err != nil {
+		return nil, errp.WithStack(err)
+	}
+	config.SetAppConfig(appConfig)
```
benma: handle error

backend/config/config.go
```diff
@@ -270,3 +279,15 @@ func (config *Config) save(filename string, conf interface{}) error {
...
 	return errp.WithStack(ioutil.WriteFile(filename, jsonBytes, 0644))
 }
+
+func (config *Config) updateInfuraAPIKeys(appConfig AppConfig) (AppConfig, error) {
+	for _, ethConfig := range []*ethCoinConfig{&appConfig.Backend.ETH, &appConfig.Backend.TETH, &appConfig.Backend.RETH} {
+		if strings.HasSuffix(ethConfig.NodeURL, "infura.io/") {
+			ethConfig.NodeURL += infuraAPIKey
+		}
+	}
+	if err := config.save(config.appConfigFilename, appConfig); err != nil {
```
benma: not needed, saving is already done in SetAppConfig()

PR #380, opened 2019-04-01
api: fix info handlers to distinguish between v1 and v2

Tomasvrba:


frontends/web/src/routes/account/account.tsx
```diff
@@ -84,12 +86,14 @@ class Account extends Component<Props, State> {
         exported: '',
         accountInfo: undefined,
         fatalError: false,
+        deviceIDs: Object.keys(this.props.devices),
```
benma: this probably doesn't work when the devices prop is changed.

you can just add a `private deviceIDs = () => { return Object.keys(this.props.devices); }` function and call it whenever needed.

Tomasvrba:
@benma updated and tested with `servewallett-multisig`, updates are working when the props change including plugging in a v1 and a v2 together :)

frontends/web/src/routes/account/account.tsx
```diff
@@ -84,12 +86,14 @@ class Account extends Component<Props, State> {
...
         accountInfo: undefined,
         fatalError: false,
+        deviceIDs: [],
     };
 
     private unsubscribe!: () => void;
 
     public componentDidMount() {
         this.unsubscribe = apiWebsocket(this.onEvent);
+        this.deviceIDs();
```
benma: doesn't do anything?

frontends/web/src/routes/account/account.tsx
```diff
@@ -117,14 +121,26 @@ class Account extends Component<Props, State> {
         if (this.props.code !== prevProps.code) {
             this.onStatusChanged();
         }
-        if (this.props.deviceIDs.length !== prevProps.deviceIDs.length) {
+        if (this.state.deviceIDs.length !== this.deviceIDs().length) {
+            this.setState({ deviceIDs: this.deviceIDs() });
```
benma: again, why is this a state var? (state vars are for stuff that impacts rendering, `render()` is triggered). it is less code and less fragile to just call `this.deviceIDs()` where needed (or pass it to checkSDCards as an argument).

frontends/web/src/routes/account/account.tsx
Tomasvrba: Whoops, meant to set the state there :)

frontends/web/src/routes/account/account.tsx
Tomasvrba: fixed

Tomasvrba:
@benma Changed so deviceIDs are always fetched from the devices on the props ;)

PR #387, opened 2019-04-17
Backups

Tomasvrba:


frontends/web/src/app.tsx
```diff
@@ -181,7 +181,7 @@ class App extends Component<Props, State> {
                         {/* Use with TypeScript: {Route<{ deviceID: string }>({ path: '/manage-backups/:deviceID', component: ManageBackups })} */}
                         {/* ManageBackups and DeviceSwitch need a key to trigger (re-)mounting when devices change, to handle routing */}
                         <ManageBackups
-                            path="/manage-backups/:deviceID"
+                            path="/manage-backups/:deviceID/:sdCardInserted?"
```
benma: why put sdCardinserted into the url? seems like the wrong place, especially since 

- you make api calls in BackupsV2 anyway. Why not make an api call there to check if the sd card is inserted?
- you turn the prop into a state and modify it after. That is rarely good, as then you have a useless prop lying around.
- you use ManageBackups for both v1 and v2, but the url param is only used for v2, which is inconsistent.

I'd just do it like in the v1 backups component and check if the sd card is inserted within.

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -349,9 +362,10 @@ class BitBox02 extends Component<Props, State> {
                                     </div>
                                     <div className={style.buttons}>
                                         <button
+                                            disabled={!sdCardInserted}
                                             className={[style.button, style.primary].join(' ')}
                                             onClick={this.restoreBackup}>
-                                            Name Device & Continue
+                                            {sdCardInserted ? 'Name Device & Continue' : 'Insert SD Card with Backup'}
```
benma: looks hard to reason about. Better to make a new `<Step ...>` at the beginning to enter the SD card, as it needs to be inserted in any case.

backend/devices/bitbox02/device.go
```diff
@@ -454,6 +454,24 @@ func (device *Device) ListBackups() ([]*Backup, error) {
 	return backups, nil
 }
 
+// CheckBackup checks the integrity of backups and if they are restorable
```
benma: the comment is wrong. The call checks whether any of the backups on the sd card is a backup of the current seed on the device .

frontends/web/src/app.tsx
Tomasvrba: The reason I'd done it this way is for UX purposes. In v1, the user has to click the "I have inserted the SD card" button after they have inserted it, and they can click this button even before inserting the SD card, and nothing happens when the button is pressed. I don't think that's great UX so my idea was to make it more seamless. 
This way there is no re-rendering (Preact still uses componentWillUpdate and other deprecated methods and doesn't seem to have getDerivedStateFromProps yet) and the updates get listed automatically when the user presses ok on the device. I agree that the props/state handling in manage backups is not ideal, but maybe a better solution would then be to separate manage backups into two components, one for v1 and one for v2, same as is currently done with backups.

backend/devices/bitbox02/device.go
Tomasvrba: done

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
Tomasvrba: Included in a higher step

benma:
Wizard: would be good to make `Please make sure you have a microSD card inserted in your BitBox.` bold, as it's easy to overlook but the most important part in the first step.

Tomasvrba:
Fixed restore backups list and bold text. 
Also made  the Insert SD Card dialog in the settings behave like in v1 for now until we implement a way to cancel the API call in the firmware and make the dialog properly blocking.
Relevant issues:
https://github.com/shiftdevices/godbb/issues/769
https://github.com/shiftdevices/firmware_v2/issues/537

PR #393, opened 2019-05-13
Add BitBoxBase Support for Demonstrations

TheCharlatan:
This is a barebones implementation to demonstrate connection with the BitBox base middleware.

MDNS scans for bitboxbase tcp service entries in the background. If a service entry is found, or an ip is specified in the frontend, TryMakeNewBase() is called in the backend, which creates a connection with the base, first making a get request, then spawning a websocket connection.

Once a connection to the middleware has been established the backend creates a new bitboxBase and bitboxBaseID, and finally emits an event that subsequently spawns a new base component in the frontend.

This connectBase component in the frontend then subscibes to the blockinfo endpoint of the middleware, which has been implemented for demo purposes. The backend is also capable of adding and removing connections to multiple bitbox base nodes, the frontend provides a 'delete' button for each
base component to showcase this. If the entry is removed in the frontend, mdns does not try to
re-register it again -> needs to be blacklisted as well. It is acceptable for now to keep this blacklist in RAM.
The BitBoxBaseIDs are sorted by time created before being passed to the frontend.

To test this pull request run the middleware in [here](https://github.com/shiftdevices/base-middleware) (requires bitcoind and lightningd to run), or connect to the same local network, where a [BitBox Base](https://github.com/digitalbitbox/bitbox-base) is configured with the middleware.


backend/base/mdns/detector.go
```diff
@@ -0,0 +1,147 @@
...
+func (detector *Detector) checkIfRemoved(bitboxBaseID string) bool {
+	//TODO implement this function, should check if the bitboxBase is reachable, if not properly unregister it.
+	return false
+}
+
+func (detector *Detector) listen() {
+	for {
+		//TODO This is where the mdns implementation should hook in, code is kept here to give an idea of the interfaces involved.
+
+		resolver, err := zeroconf.NewResolver(nil)
```
benma: move instance of this to the struct instead of creating a new one every iteration?

backend/base/mdns/detector.go
```diff
@@ -0,0 +1,147 @@
...
+	var err error
+	baseDevice, err = base.NewBitBoxBase(ip, bitboxBaseID)
+
+	if err != nil {
+		detector.log.WithError(err).Error("Failed to register Base")
+		return false, err
+	}
+	detector.baseDeviceInterface[bitboxBaseID] = baseDevice
+	if err := detector.onRegister(baseDevice); err != nil {
+		detector.log.WithError(err).Error("Failed to execute on-register")
```
benma: return the error

backend/base/updater/updater.go
```diff
@@ -0,0 +1,108 @@
+package updater
```
benma: pls add a docstring explaining what this package does

backend/base/updater/updater.go
```diff
@@ -0,0 +1,108 @@
...
+	//This should not be asynchronous, the frontend should not show something until the pairing initiates
+	simpleURL := url.URL{Scheme: "http", Host: ip, Path: "/"}
+	response, err := http.Get(simpleURL.String())
+	//defer response.Body.Close()
+	if err != nil {
+		updater.log.Println("No response from middleware", err)
+		updater.running = false
+		return err
+	}
+	if response.StatusCode == http.StatusOK {
```
benma: to save an indentation level and have consistent code flow, I prefer to handle the error first:

```
if response.StatusCode != http.StatusOK {
    return errp.New("...")
}
// continue with success handling
```

backend/base/updater/updater.go
```diff
@@ -0,0 +1,108 @@
...
+		_, err = regexp.MatchString("OK!", string(bodyBytes))
+		if err != nil {
+			updater.running = false
+			updater.log.Println("Unexpected Response Body Bytes")
+			return err
+		}
+	} else {
+		updater.log.Println("Received http status code from middleware other than 200")
+		updater.running = false
+		return err
```
benma: this `err` is unrelated to the http.StatusOK and is in fact `nil` instead of an error value.

backend/handlers/handlers.go
```diff
@@ -181,9 +189,13 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/coins/btc/headers/status", handlers.getHeadersStatus("btc")).Methods("GET")
 	getAPIRouter(apiRouter)("/certs/download", handlers.postCertsDownloadHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/certs/check", handlers.postCertsCheckHandler).Methods("POST")
+	getAPIRouter(apiRouter)("/connectbase", handlers.postConnectBaseHandler).Methods("POST")
```
benma: but connect and disconnect also unuder the bitboxBaseRouter, so it's `/bitboxbases/connect`?

frontends/web/src/components/base/connectBase.tsx
```diff
@@ -0,0 +1,103 @@
+/**
```
benma: by convention lowercase filenames `connectbase.tsx`

frontends/web/src/routes/base/base.tsx
```diff
@@ -0,0 +1,111 @@
+/**
```
benma: i prefer more explicit file/object names for the base, as `base` is a very generic and overloaded workd. `bitboxbase.tsx`, `BitBoxBase`, etc. will help.

backend/backend.go
```diff
@@ -581,11 +610,57 @@ func (backend *Backend) Events() <-chan interface{} {
...
+		Type: "bitboxbases",
+		Data: "registeredChanged",
+	}:
+	default:
+	}
+	return nil
+}
+
+// BitBoxBaseDeregister deregisters the device with the given ID from this backend.
+func (backend *Backend) BitBoxBaseDeregister(bitboxBaseID string) {
```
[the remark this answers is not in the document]
TheCharlatan: This is used in handlers/handlers.go to manually delete a registered bitboxbase.

backend/backend.go
```diff
@@ -581,11 +610,57 @@ func (backend *Backend) Events() <-chan interface{} {
...
+	if _, ok := backend.bitboxBases[bitboxBaseID]; ok {
+		backend.bitboxBases[bitboxBaseID].Close()
+		backend.onBitBoxBaseUninit(bitboxBaseID)
+		delete(backend.bitboxBases, bitboxBaseID)
+		backend.baseDetector.RemoveBase(bitboxBaseID)
```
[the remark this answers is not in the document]
TheCharlatan: When a manual removal is initiated, it is not in the power of the detector to do this.

backend/base/mdns/detector.go
TheCharlatan: It can't seem to resolve mdns connections anymore, when I create the resolver outside of the loop. I'm a bit puzzled why that is the case.

TheCharlatan:
Thanks for the review! I force-pushed to address all the comments, save for the ones where I left a reply.

backend/base/mdns/detector.go
TheCharlatan: I changed the code a bit here, there is now a separate function that starts a goroutine with a new context. The goroutine returns once <- context.Done() is triggered.

TheCharlatan:
I force-pushed again to implement the mdns lookup in a more robust way, rename some more instances, where I thought just 'base' is too ambiguous and added some small patches to make travis pass.

benma:
:man_dancing:

PR #400, opened 2019-05-21
backups: use backup names instead of IDs in App

Tomasvrba:
- Use backup names instead of IDs when listing backups
- Show name instead of ID of verified backup when checking backups
- Remove "Name Device" step when restoring from backup, since the name
of the backup gets restored as the device name anyway


Tomasvrba:
@benma At the moment, the check backup call returns both the ID and the name after I added the name. However, the ID is unused and the user doesn't see it anymore at this point. Should I just get rid of it from the API response and the protobuf message altogether?

Tomasvrba:
> @benma At the moment, the check backup call returns both the ID and the name after I added the name. However, the ID is unused and the user doesn't see it anymore at this point. Should I just get rid of it from the API response and the protobuf message altogether?

Based on other discussion the dialog now shows the actual matching BackupListItem to the backup ID it gets from the device. So the formatting is exactly the same in the dialog as in the list of backups itself and the user can see both the name and the ID of the matching backup

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -357,20 +357,15 @@ class BitBox02 extends Component<Props, State> {
...
-                                        <div className={style.inputGroup}>
-                                            <label className={style.label}>Wallet Name</label>
-                                            <input type="text" className={style.input} placeholder="Name of your BitBox" />
-                                        </div>
                                     </div>
                                     <div className={style.buttons}>
                                         <button
                                             className={[style.button, style.primary].join(' ')}
                                             onClick={this.restoreBackup}>
-                                            Name Device & Continue
```
benma: Great!

frontends/web/src/components/devices/bitbox02/checkbackup.tsx
```diff
@@ -31,21 +32,24 @@ type Props = CheckProps & TranslateProps;
 interface State {
     activeDialog: boolean;
     message: string;
+    foundBackup: Backup;
```
benma: `foundBackup?: Backup;` as it could be undefined/not found

frontends/web/src/components/devices/bitbox02/checkbackup.tsx
```diff
@@ -31,21 +32,24 @@ type Props = CheckProps & TranslateProps;
...
 }
 
 class Check extends Component<Props, State> {
     public state = {
         activeDialog: false,
         message: '',
+        foundBackup: { id: '', date: '', name: '' },
```
benma: `foundBackup: undefined` (showing an empty backup to the user does not help)

frontends/web/src/components/devices/bitbox02/checkbackup.tsx
Tomasvrba: Good points! Fixed

Tomasvrba:
Good points! Fixed

frontends/web/src/locales/en/app.json
```diff
@@ -64,7 +64,7 @@
         "placeholder": "Recovery password",
         "showLabel": "recovery password"
       },
-      "success": "Successfully verified backup: {{name}}",
+      "success": "Successfully verified backup:",
```
benma: reminder: you can't do this here - needs to be via locize (then `make locize-pull`) (or now that you did it here, do the same again in locize).

otherwise the net `make locize-pull` will revert this.

frontends/web/src/components/devices/bitbox02/checkbackup.tsx
```diff
@@ -31,21 +32,24 @@ type Props = CheckProps & TranslateProps;
...
 
-    private checkBackup = () => {
+    private checkBackup = backups => {
         apiPost('devices/bitbox02/' + this.props.deviceID + '/backups/check', {
             silent: false,
         }).then(({ backupID, success }) => {
             let message;
             if (success && backupID) {
-                message = this.props.t('backup.check.success', { name: backupID });
+                this.setState({foundBackup: backups.find(backup => backup.id === backupID)});
```
benma: if backup could not be found, you will be showing an empty dialog. probably better to show the notOK message if that happens, or even better `t('unknownError', { errorMessage: "not found" })`

frontends/web/src/locales/en/app.json
Tomasvrba: Right, thanks! Changed in locize

frontends/web/src/components/devices/bitbox02/checkbackup.tsx
Tomasvrba: The notOK message is already shown when a matching backup is not found, see the else statement on lines 53-4 after your comment: 
```
} else {
    message = this.props.t('backup.check.notOK');
}
```

benma: No I meant if the device returns a found backup ID, but we can't find it in the backups list. `foundBackup` can be undefined in the line 51.

Tomasvrba: Gotcha, now the `unknownError` makes more sense :)
I'll make a fix

PR #403, opened 2019-05-22
device-settings: persist device name entered during setup workflow

Tomasvrba:


frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -207,9 +209,28 @@ class BitBox02 extends Component<Props, State> {
         });
     }
 
+    private handleDeviceNameInput = (event: Event) => {
+        const target = (event.target as HTMLInputElement);
+        const value: string = target.value;
+        this.setState(prevState => ({
```
benma: what does this do over `this.setState({ deviceName: value })`?

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -207,9 +209,28 @@ class BitBox02 extends Component<Props, State> {
...
+            deviceName: value,
+        }));
+    }
+
+    private setDeviceName = () => {
+        apiPost('devices/bitbox02/' + this.props.deviceID + '/set-device-name', { name: this.state.deviceName })
+        .then(result => {
+            if (result.success) {
+                this.setPassword();
+            }
```
benma: `else { ? }` -> error handling ;)

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
Tomasvrba: I suppose in this case we don't really need it. Passing a function instead of an object to setState makes sure the state actually ends up being what you expect it to be, especially if multiple setStates are being called at the same and you don't know when the state actually gets changed. But since nothing else is happening to the state in this case, it's kind of an overkill... :)

Tomasvrba:
Fixed

PR #406, opened 2019-05-26
Accountswitching

benma:


frontends/web/src/routes/account/account.tsx
```diff
@@ -333,19 +350,19 @@ class Account extends Component<Props, State> {
...
                     )}
                     {this.isLegacy(account, accountInfo) && (
                         <Entry key="accountLegacyConvert" entry={t('guide.accountLegacyConvert')} />
                     )}
-                    {transactions.length > 0 && (
+                    {transactions !== undefined &&  transactions.length > 0 && (
                         <Entry key="accountTransactionAttributesGeneric" entry={t('guide.accountTransactionAttributesGeneric')} />
                     )}
-                    {transactions.length > 0 && isBitcoinBased(account.coinCode) && (
+                    {transactions !== undefined && transactions.length > 0 && isBitcoinBased(account.coinCode) && (
```
Tomasvrba: Could we just wrap these lines in a single expression so the same check isn't repeated four times but happens only once before rendering the TXs?

benma: you mean to put `transactions !== undefined && transactions.length > 0` into a var?

Tomasvrba: What I had in mind was more something like this: 
```
{ (transactions !== undefined && transactions.length > 0) 
                        ? (
                            <div>
                                <Entry key="accountTransactionLabel" entry={t('guide.accountTransactionLabel')} />
                                <Entry key="accountTransactionTime" entry={t('guide.accountTransactionTime')} />
                                <Entry key="accountTransactionAttributesGeneric" entry={t('guide.accountTransactionAttributesGeneric')} />
                                {isBitcoinBased(account.coinCode) && <Entry key="accountTransactionAttributesBTC" entry={t('guide.accountTransactionAttributesBTC')} /> }
                            </div>
                          )
                        : ''
                    }
```

benma: Not sure, it introduces more nesting, and also twice as there is a `accountLegacyConvert` entry in the middle. 

I prefer a flat hierarchy here.

Tomasvrba: ack

jonasschnelli:
Tested.
Seems to fix it.

PR #410, opened 2019-06-03
check-backup: make blocking, wait for user confirmation on device

Tomasvrba:
Send two check-backup requests, one silent and one not to receive
backup data from device and wait for user confirmation on device
screen.

Also fixes an issue when a successful backup was found and ID
returned, but could not be matched to a backup name in frontend


benma:
Well done!

PR #411, opened 2019-06-03
restore-backup: add set password Step to restore wizard

Tomasvrba:


Tomasvrba:
> Without looking at the bigger context of this PR, the way you pass `restoreSetPassword` to the child component looks fine to me.

Great, thanks. So typing the function with `restoreSetPassword?: () => void;` on the `BackupsProps` is an ok thing to do? That's mostly what I wasn't sure about.

PR #424, opened 2019-06-12
.dockerignore: Initial commit

hkjn:
This reduces the context that needs to be copied with `make dockerinit`
from ~1.21GB to ~0.35GB on my machine, which improves performance
when building Docker images.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+/frontends/qt/build


benma:
Why does this make an impact? I don't think our Dockerfile copies this folder, or the whole repo, so it shouldn't even be part of the image.

hkjn:
> Why does this make an impact? I don't think our Dockerfile copies this folder, or the whole repo, so it shouldn't even be part of the image.

It makes a difference since the entire build context (current working directory of the `docker build` command) is copied to the Docker daemon, so even if there's no `COPY` stanza referencing a certain directory, it still is copied to the context, just not included in the layers built.

For more, see:

* https://docs.docker.com/engine/reference/builder/
* https://docs.docker.com/engine/reference/builder/#dockerignore-file

benma:
Nice, thanks for the info. So maybe the whole directory could then be ignored instead of just the build directory?

Also happy to merge this as it is, lmk.

hkjn:
> Nice, thanks for the info. So maybe the whole directory could then be ignored instead of just the build directory?

Sure, let me know if you prefer I make any changes; I just looked for what the biggest directory was and was satisfied with the 70%+ improvement when excluding this specific one. :)

PR #431, opened 2019-06-13
Move devmode electrum config from backend to config

TheCharlatan:
This pull request moves the devmode certificates from the backend to the config. They are reachable by passing the DevMode argument when creating a new Config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,8 @@ servewallet-regtest:
 	go install ./cmd/servewallet/... && servewallet -regtest
 servewallet-multisig:
 	go install ./cmd/servewallet/... && servewallet -multisig
+servewallet-prodservers:
+	go install ./cmd/servewallet/... && servewallet -devservers=false
 buildweb:
 	node --version
 	rm -rf ${WEBROOT}/build
--- a/cmd/servewallet/main.go
+++ b/cmd/servewallet/main.go
@@ -72,6 +72,7 @@ func main() {
 	regtest := flag.Bool("regtest", false, "use regtest instead of testnet coins")
 	multisig := flag.Bool("multisig", false, "use the app in multisig mode")
 	devmode := flag.Bool("devmode", true, "switch to dev mode")
+	devservers := flag.Bool("devservers", true, "switch to dev servers")
 	flag.Parse()
 
 	logging.Set(&logging.Configuration{Output: "STDERR", Level: logrus.DebugLevel})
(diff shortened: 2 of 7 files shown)


backend/config/config.go
```diff
@@ -120,8 +120,106 @@ ffbrVM+I91v3R03Svv2Nte2xdbx1RmoI/y3tMyZL
...
+d1r8kFyCgpL9KASdahfyLBWCC7awlcOQP1QJA5QoO9u5Feq3lU0VnJF0YCZh8GOy
+py3n1TR6S59eT495BiKDjWnhdVchEa8zMGIW/wFW7EX/LyW2zX3hQsdfnmMWUPVr
+O3nOxjgSfRAfKWQ2Ny1APKcn6I83P5PFLhtO5I12
+-----END CERTIFICATE-----`
+
 // NewDefaultAppConfig returns the default app config.
-func NewDefaultAppConfig() AppConfig {
+func NewDefaultAppConfig(devMode bool) AppConfig {
+	if devMode {
+		return AppConfig{
```
benma: Since there is a lot of common settings which don't change in devmode, to avoid updating/maintaining both instances, it's easier to make one `appConfig := AppConfig{...}`, and modify if after in an `if devMode { /* modify appConfig here */ }` clause.

Makefile
```diff
@@ -44,6 +44,8 @@ servewallet-regtest:
 	go install ./cmd/servewallet/... && servewallet -regtest
 servewallet-multisig:
 	go install ./cmd/servewallet/... && servewallet -multisig
+servewallet-prod:
```
benma: maybe `-prodservers` to indicate it's only about this?

Makefile
```diff
@@ -44,6 +44,8 @@ servewallet-regtest:
...
+	go install ./cmd/servewallet/... && servewallet -devserver=false
```
benma: `-devservers`, as you use the plural everywhere else too

cmd/servewallet/main.go
```diff
@@ -69,6 +69,7 @@ func main() {
 	regtest := flag.Bool("regtest", false, "use regtest instead of testnet coins")
 	multisig := flag.Bool("multisig", false, "use the app in multisig mode")
 	devmode := flag.Bool("devmode", true, "switch to dev mode")
+	devserver := flag.Bool("devserver", true, "switch to dev servers")
```
benma: also plural here

Makefile
TheCharlatan: Yup.

Makefile
TheCharlatan: Fixed.

PR #433, opened 2019-06-14
Allow connection to bitbox-base electrs

TheCharlatan:
This commit adds support to connect with electrs running testnet or mainnet on the BitBox Base. A new postConnectElectrumHandler was created for this purpose. It takes a bitboxBaseID, from that gets
the connected BitBox Bases'IP and initiates a connection with the base.

The electrs port and network type (mainnet/testnet) are fetched with the GetEnv get request.

Once connected, other active backends will be unset, such that only bitcoin accounts will be available.

### DIFF
--- a/backend/bitboxbase/bitboxbase.go
+++ b/backend/bitboxbase/bitboxbase.go
@@ -15,9 +15,15 @@
 package bitboxbase
 
 import (
+	"encoding/json"
+	"strings"
 	"time"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/bitboxbase/updater"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/electrum"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/config"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
+	"github.com/sirupsen/logrus"
 )
 
 // Interface represents bitbox base.
@@ -38,26 +44,97 @@ type Interface interface {
 
 	// BlockInfo returns some blockchain information.
 	BlockInfo() string
+
+	// ConnectElectrum connects to the electrs server on the base and configures the backend accordingly
+	ConnectElectrum() error
 }
 
 //BitBoxBase provides the dictated bitboxbase api to communicate with the base
 type BitBoxBase struct {
 	bitboxBaseID    string //This is just the ip at the moment, but will be an actual unique string, once the noise pairing is implemented
 	registerTime    time.Time
+	address         string
 	closed          bool
 	updaterInstance *updater.Updater
+	electrsRPCPort  string
+	network         string
+	log             *logrus.Entry
+	config          *config.Config
 }
 
 //NewBitBoxBase creates a new bitboxBase instance
-func NewBitBoxBase(ip string, id string) (*BitBoxBase, error) {
+func NewBitBoxBase(address string, id string, config *config.Config) (*BitBoxBase, error) {
 	bitboxBase := &BitBoxBase{
+		log:             logging.Get().WithGroup("bitboxbase"),
 		bitboxBaseID:    id,
 		closed:          false,
-		updaterInstance: updater.NewUpdater(),
+		address:         strings.Split(address, ":")[0],
+		updaterInstance: updater.NewUpdater(address),
 		registerTime:    time.Now(),
+		config:          config,
 	}
-	err := bitboxBase.GetUpdaterInstance().Connect(ip, bitboxBase.bitboxBaseID)
+	err := bitboxBase.GetUpdaterInstance().Connect(address, bitboxBase.bitboxBaseID)
+	if err != nil {
+		return nil, err
+	}
+
+	bodyBytes, err := bitboxBase.GetUpdaterInstance().GetEnv()
+	if err != nil {
+		return nil, err
+	}
+	var envData map[string]interface{}
+	if err := json.Unmarshal(bodyBytes, &envData); err != nil {
+		bitboxBase.log.WithError(err).Error(" Failed to unmarshal GetEnv body bytes")
+		//bitboxBase.GetUpdaterInstance().Stop()
+		return bitboxBase, err
+	}
+	var ok bool
+	bitboxBase.electrsRPCPort, ok = envData["electrsRPCPort"].(string)
+	if !ok {
+		bitboxBase.log.Error(" Getenv did not return an electrsRPCPort string field")
+		return bitboxBase, err
+	}
+	bitboxBase.network, ok = envData["network"].(string)
+	if !ok {
+		bitboxBase.log.Error(" Getenv did not return a network string field")
+		return bitboxBase, err
+	}
 	return bitboxBase, err
+}
+
+// ConnectElectrum connects to the electrs server on the base and configures the backend accordingly
+func (base *BitBoxBase) ConnectElectrum() error {
+	electrumAddress := base.address + ":" + base.electrsRPCPort
+
+	electrumCert, err := electrum.DownloadCert(electrumAddress)
+	if err != nil {
+		base.log.WithField("ElectrumIP: ", electrumAddress).Error(err.Error())
+		return err
+	}
+
+	if err := electrum.CheckElectrumServer(
+		electrumAddress,
+		electrumCert,
+		base.log); err != nil {
+		base.log.WithField("ElectrumIP: ", electrumAddress).Error(err.Error())
+		return err
+	}
+
+	base.log.WithField("ElectrumAddress:", electrumAddress).Debug("Setting config to base electrum Server...")
+
+	// BaseBtcConfig sets the TBTC configs to the provided cert and ip.
+	if base.isTestnet() {
+		base.config.SetTBTCElectrumServers(electrumAddress, electrumCert)
+	} else {
+		base.config.SetBTCElectrumServers(electrumAddress, electrumCert)
+	}
+	// Disable Litecoin and Ethereum accounts - we do not want any more traffic hitting other servers
+	base.config.SetBtcOnly()
+
+	if err := base.config.SetAppConfig(base.config.AppConfig()); err != nil {
+		return err
+	}
+	return nil
 }
 
 //GetUpdaterInstance return ths current instance of the updater
@@ -78,6 +155,11 @@ func (base *BitBoxBase) Identifier() string {
 //GetRegisterTime implements a getter for the timestamp of when the bitboxBase was registered
 func (base *BitBoxBase) GetRegisterTime() time.Time {
 	return base.registerTime
+}
+
+//isTestnet returns a boolean that is true when connected to a base serving testnet and false otherwise
+func (base *BitBoxBase) isTestnet() bool {
+	return base.network == "testnet"
 }
 
 //Close implements a method to unset the bitboxBase
--- a/backend/bitboxbase/mdns/detector.go
+++ b/backend/bitboxbase/mdns/detector.go
@@ -21,7 +21,9 @@ import (
 	"time"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/bitboxbase"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/config"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
+
 	"github.com/grandcat/zeroconf"
 	"github.com/sirupsen/logrus"
 )
@@ -35,7 +37,8 @@ type Detector struct {
 
 	removedIDs map[string]bool //keeps track of the bases that were removed manually by the user.
 
-	log *logrus.Entry
+	log    *logrus.Entry
+	config *config.Config
 }
 
 // NewDetector creates a new detector. onRegister is called when a bitboxbase has been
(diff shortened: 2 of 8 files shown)


backend/bitboxbase/bitboxbase.go
```diff
@@ -38,25 +42,57 @@ type Interface interface {
...
+	if err != nil {
+		return nil, err
+	}
+	var envData map[string]interface{}
+	if err := json.Unmarshal(bodyBytes, &envData); err != nil {
+		bitboxBase.log.WithError(err).Error(" Failed to unmarshal GetEnv body bytes")
+		bitboxBase.GetUpdaterInstance().Stop()
+		return nil, err
+	}
+	bitboxBase.electrsRPCPort, _ = envData["electrsRPCPort"].(string)
```
benma: handle error (always for external input).

panics are reserved for programming errors (truly unexpected) or when it's unreasonably hard to do error handling.

backend/bitboxbase/bitboxbase.go
```diff
@@ -70,11 +106,26 @@ func (base *BitBoxBase) BlockInfo() string {
...
 func (base *BitBoxBase) Identifier() string {
 	return base.bitboxBaseID
 }
 
+// GetElectrsRPCPort implements a getter for the electrsport
+func (base *BitBoxBase) GetElectrsRPCPort() string {
+	return base.electrsRPCPort
+}
+
+//GetNetwork implements a getter for the network string (either mainnet or testnet)
```
benma: I suggest to do `GetCoinCode()` instead and return `"btc"`, `"tbtc"`, as this is what the rest of the app calls it.

You can keep the env api in the base as is, but translate it here to the local glossary.

backend/bitboxbase/mdns/detector.go
```diff
@@ -75,6 +77,14 @@ func (detector *Detector) TryMakeNewBase(ip string) (bool, error) {
 	}
 	var baseDevice bitboxbase.Interface
 	var err error
+
+	match := regexp.MustCompile(`[0-9]+(?:\.[0-9]+){3}(:[0-9]+)`).MatchString(ip)
```
benma: Thinking about it, why do you have IP everywhere instead of `address`, which could also be a hostname? I think it will be better to use `address` everywhere, as http.Get() etc. can also use that.

backend/bitboxbase/updater/updater.go
```diff
@@ -38,6 +38,10 @@ type Updater struct {
 	blockchainInfo string
 	log            *logrus.Entry
 	running        bool
+	ip             string
+	rootURL        url.URL
```
benma: the base of all the urls will be the same, and the endpoints are constants (the api is not dynamic), so it's enough to just keep the ip here and put the endpoints into constants.

backend/bitboxbase/updater/updater.go
```diff
@@ -84,19 +91,35 @@ func (updater *Updater) Connect(ip string, bitboxBaseID string) error {
...
-	go listenWebsocket(updater, ip, bitboxBaseID)
+	go listenWebsocket(updater, bitboxBaseID)
 	return nil
 }
 
+//GetEnv makes an http GET request to the base and on success returns some environment data from the base such as the electrs port
+func (updater *Updater) GetEnv() ([]byte, error) {
+	response, err := http.Get(updater.getEnvURL.String())
+	if err != nil {
+		updater.running = false
```
benma: call `updater.Stop()` so it is compatible with future changes.

Also, do you even need to stop? What if it is a transient error?

backend/bitboxbase/updater/updater.go
```diff
@@ -84,19 +91,35 @@ func (updater *Updater) Connect(ip string, bitboxBaseID string) error {
...
+		updater.log.WithError(err).Error("GetEnv request failed, unable to get information from BitBox Base")
+		return nil, err
+	}
+	bodyBytes, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		updater.running = false
```
benma: same

backend/config/config.go
```diff
@@ -228,6 +228,31 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) (*Config
...
 
+// SetBtcOnly sets non-bitcoin accounts in the config to false
+func (config *Config) SetBtcOnly() {
+	config.appConfig.Backend.LitecoinP2WPKHP2SHActive = false
+	config.appConfig.Backend.LitecoinP2WPKHActive = false
+	config.appConfig.Backend.EthereumActive = false
+}
+
+// BaseBtcConfig sets the TBTC configuration to the provided electrumIP and electrumCert
+func (config *Config) BaseBtcConfig(electrumIP, electrumCert string, network string) {
```
benma: To keep the separation of concerns, it's better to have methods here like `SetBTCElectrumServers`, `SetTBTCElectrumServers`, and do the logic of which to change how in the bitboxbase folder.

backend/handlers/handlers.go
```diff
@@ -191,6 +191,7 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/certs/check", handlers.postCertsCheckHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/bitboxbases/connectbase", handlers.postConnectBaseHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/bitboxbases/disconnectbase", handlers.postDisconnectBaseHandler).Methods("POST")
+	getAPIRouter(apiRouter)("/bitboxbases/connectElectrum", handlers.postConnectElectrumHandler).Methods("POST")
```
benma: consistency: `connect-electrum`.

Also, why move the endpoint into `bitboxbase/handlers`? It acts on a single base after all, right?

backend/handlers/handlers.go
```diff
@@ -686,6 +687,55 @@ func (handlers *Handlers) postDisconnectBaseHandler(r *http.Request) (interface{
...
+			"errorMessage": err.Error(),
+		}, nil
+	}
+
+	handlers.log.WithField("ElectrumIP:", electrumIP).Debug("Connecting to new electrum Server...")
+
+	config := handlers.backend.Config()
+	// BaseBtcConfig sets the TBTC configs to the provided cert and ip.
+	config.BaseBtcConfig(electrumIP, electrumCert, bitboxBase.GetNetwork())
+	// Disable Litecoin and Ethereum accounts - we do not want any more traffic hitting other servers
```
benma: I disagree. If I use the base for btc, it does not mean I only use btc ;) I'd remove this.

backend/handlers/handlers.go
```diff
@@ -686,6 +687,55 @@ func (handlers *Handlers) postDisconnectBaseHandler(r *http.Request) (interface{
 	return map[string]interface{}{"success": success}, nil
 }
 
+func (handlers *Handlers) postConnectElectrumHandler(r *http.Request) (interface{}, error) {
+	jsonBody := map[string]string{}
```
benma: When you move this to `bitboxbase/handlers`, don't forget to put the meat of the logic into `bitboxbase.go` and have the handler only do the http request parsing and response construction :smile:

frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -64,6 +64,18 @@ export class ConnectedBase extends Component<Props, State> {
 
     }
 
+    private connectElectrum = (event: Event) => {
+        event.preventDefault();
```
benma: Don't think this is needed.

frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -64,6 +64,18 @@ export class ConnectedBase extends Component<Props, State> {
...
+        apiPost('bitboxbases/connectElectrum', {
+            bitboxBaseID : this.props.bitboxBaseID,
+        }).then(data => {
+            const { success } = data;
```
benma: a small trick: could also do `.then(({ success }) => { ... }`  to save a line.

backend/bitboxbase/bitboxbase.go
```diff
@@ -38,25 +42,57 @@ type Interface interface {
...
 }
 
 //NewBitBoxBase creates a new bitboxBase instance
 func NewBitBoxBase(ip string, id string) (*BitBoxBase, error) {
 	bitboxBase := &BitBoxBase{
+		log:             logging.Get().WithGroup("bitboxbase"),
 		bitboxBaseID:    id,
 		closed:          false,
-		updaterInstance: updater.NewUpdater(),
+		ip:              strings.Split(ip, ":")[0],
```
benma: it is confusing that `ip` can mean with or without port depending on the context. Better stick to one way only.


elsewhere we use `address string` to be the endpoint including port.

speaking of: why

backend/handlers/handlers.go
TheCharlatan: I would not want my wallet app talking to any shift operated servers any more once I configure the base.

backend/bitboxbase/bitboxbase.go
TheCharlatan: Currently we still communicate with servers running on other ports on the BitBox Base. This will change once everything is hosted through a single port.

backend/handlers/handlers.go
Tomasvrba: But users don't really have the option unless we also provide them with a Litecoin and Ethereum node, or they would need to use a separate app

benma: I see the point, but at least there should be a message after it saying that those were disabled for that reason, and the user can re-enable them in the settings.

Stadicus: This seems like a difficult UX decision, and I think we should provide as much transparency for the user as possible. 

My preference is to display a message and ask whether to disable backends for other chains (default: no).

Reason: if I'm a Bitcoin Maximalist (non-toxic) I don't use the other chains anyway. If I have Ethereum / Litecoin enabled, I'd probably want to keep the wallet functionality.

TheCharlatan: Is it fine if we add proper notification flow in a separate pull request?

backend/bitboxbase/bitboxbase.go
```diff
@@ -38,28 +44,99 @@ type Interface interface {
...
+		return nil, err
+	}
+	var envData map[string]interface{}
+	if err := json.Unmarshal(bodyBytes, &envData); err != nil {
+		bitboxBase.log.WithError(err).Error(" Failed to unmarshal GetEnv body bytes")
+		//bitboxBase.GetUpdaterInstance().Stop()
+		return bitboxBase, err
+	}
+	var success bool
+	bitboxBase.electrsRPCPort, success = envData["electrsRPCPort"].(string)
```
benma: nit: common name here is `ok` (e.g. https://tour.golang.org/methods/15)

backend/bitboxbase/bitboxbase.go
```diff
@@ -38,28 +44,99 @@ type Interface interface {
...
+	if !success {
+		bitboxBase.log.Error(" Getenv did not return an electrsRPCPort field")
```
benma: btw, actually it might have, but not of type string.

backend/bitboxbase/mdns/detector.go
```diff
@@ -67,15 +72,16 @@ func (detector *Detector) TryMakeNewBase(ip string) (bool, error) {
...
+	bitboxBaseID := address
 	// Skip if already registered.
 	if _, ok := detector.baseDeviceInterface[bitboxBaseID]; ok {
 		return false, nil
 	}
 	var baseDevice bitboxbase.Interface
 	var err error
-	baseDevice, err = bitboxbase.NewBitBoxBase(ip, bitboxBaseID)
+
+	baseDevice, err = bitboxbase.NewBitBoxBase(address, bitboxBaseID, detector.config)
```
benma: nit: `:=` instead of `=`, then you can delete the two declaration lines above.

backend/config/config.go
```diff
@@ -228,6 +228,41 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) (*Config
...
+// SetBtcOnly sets non-bitcoin accounts in the config to false
+func (config *Config) SetBtcOnly() {
+	config.appConfig.Backend.LitecoinP2WPKHP2SHActive = false
+	config.appConfig.Backend.LitecoinP2WPKHActive = false
+	config.appConfig.Backend.EthereumActive = false
+}
+
+// SetBTCElectrumServers sets the BTC configuration to the provided electrumIP and electrumCert
+func (config *Config) SetBTCElectrumServers(electrumAddress, electrumCert string) {
+	var btcBaseCoinConfig = btcCoinConfig{
```
benma: var name invalid ;) you can also skip a variable here and directly assign the struct below.

backend/config/config.go
```diff
@@ -228,6 +228,41 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) (*Config
...
+				PEMCert: electrumCert,
+			},
+		},
+	}
+	config.appConfig.Backend.BTC = btcBaseCoinConfig
+}
+
+// SetTBTCElectrumServers sets the TBTC configuration to the provided electrumIP and electrumCert
+func (config *Config) SetTBTCElectrumServers(electrumAddress, electrumCert string) {
+	var btcBaseCoinConfig = btcCoinConfig{
```
benma: same

frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -51,19 +51,27 @@ export class ConnectedBase extends Component<Props, State> {
...
+        }).then(({ success }) => {
             if (!success) {
                 alertUser('Did not work');
             }
         });
 
     }
 
+    private connectElectrum = () => {
+        apiPost('bitboxbases/' + this.props.bitboxBaseID + '/connectelectrum', {
```
benma: from before: consistency: `connect-electrum`.

backend/handlers/handlers.go
benma: sure

benma:
:+1: please squash, then we can merge this beauty

TheCharlatan:
Squashed.

PR #448, opened 2019-07-02
Noise encrypted base

TheCharlatan:
This pull request implements noise encryption and authentication with the base. The 3 part noise 'XX' handshake is done over websocket. The keypair of the app and the public static key are saved to a file. If the public static key of either party does not exist in either party's configuration, the user is asked to verify the channel binding hash on the base by comparing its value with the app. For this an event notification is emitted, that allows the frontend to display the channel hash data.
The implementation lends heavily from the bitbox02/device.go noise handshake implementation. It might make sense to abstract the common functionality, like the writing into a config file, into a separate package.
The pull request can be tested against https://github.com/TheCharlatan/bitbox-base/tree/noiseEncryptedBase


frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -83,24 +83,18 @@ export class ConnectedBase extends Component<Props, State> {
...
                 <div class="row">
                     <div class="flex flex-1 flex-row flex-between flex-items-center spaced">
-                        <p>Block Number: {blocks}</p>
-                        <p>Difficulty: {difficulty}</p>
-                        <p>Incrementer: {i}</p>
+                        <p>Block Number: {blockInfo[blocks]}</p>
+                        <p>Difficulty: {blockInfo[difficulty]}</p>
                         <p>Device ID: {bitboxBaseID}</p>
-                        <p>Lightning Alias: {alias}</p>
+                        <p>Lightning Alias: {blockInfo[lightningAlias]}</p>
```
thisconnect: since `const lightningAlias = 'lightningAlias'` is always lightningAlias, you can just `{blockInfo.lightningAlias}` and remove `const lightningAlias`. Same for `blockInfo.blocks` and `blockInfo. difficulty `

frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -36,7 +36,7 @@ export class ConnectedBase extends Component<Props, State> {
     constructor(props) {
         super(props);
         this.state = {
-            blockInfo : '',
+            blockInfo : {},
```
thisconnect: keep in mind and empty string `''` is falsy, but and empty object is truthy, this matters in the render when you check `if (!blockInfo) {`

Screen Shot 2019-07-02 at 1 04 46 PM

TheCharlatan:
Force-pushed to fix some typos. No functional changes.

frontends/web/src/components/bitboxbase/connectbase.tsx
TheCharlatan: Typescript does not allow object referencing without giving the object a type first.I took this lazy way out, but should do it properly.

TheCharlatan:
Pushed a commit for @thisconnect's review

backend/bitboxbase/bitboxbase.go
```diff
@@ -143,8 +145,8 @@ func (base *BitBoxBase) GetUpdaterInstance() *updater.Updater {
 }
 
 //BlockInfo returns the received blockinfo packet from the updater
-func (base *BitBoxBase) BlockInfo() string {
-	return base.GetUpdaterInstance().BlockInfo()
+func (base *BitBoxBase) BlockInfo() interface{} {
```
benma: also rename to MiddlewareInfo then?

return type should ideally be a concrete type (struct, ...?) so it's clear what is expected. then it can also be validated.

backend/bitboxbase/handlers/handlers.go
```diff
@@ -61,7 +62,7 @@ func (handlers *Handlers) Uninit() {
 
 func (handlers *Handlers) getBlockInfoHandler(_ *http.Request) (interface{}, error) {
 	handlers.log.Debug("Block Info")
-	return handlers.base.BlockInfo(), nil
+	return jsonp.MustMarshal(handlers.base.BlockInfo()), nil
```
benma: MustMarshal panics, and panics are reserved for programming errors / impossible situations.

If base.BlockInfo() however validates the input and converts it to an appropriate Go type first (see above comment), then this is okay :)

backend/bitboxbase/updater/config.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"bytes"
+
+	"github.com/digitalbitbox/bitbox-wallet-app/util/config"
+	"github.com/flynn/noise"
+)
+
+const configFilename = "base.json"
+
+type noiseKeypair struct {
```
benma: the whole file should eventually be refactored/moved to util/, so it can be shared with the bitbox02 implementation. e.g. `util/noiseconfig`

backend/bitboxbase/updater/updater.go
```diff
@@ -25,45 +28,67 @@ import (
...
 	"github.com/digitalbitbox/bitbox-wallet-app/util/observable/action"
+	"github.com/flynn/noise"
 
 	"github.com/gorilla/websocket"
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	opICanHasHandShaek          = "h"
+	opICanHasPairinVerificashun = "v"
```
benma: :heart

backend/bitboxbase/updater/updater.go
```diff
@@ -25,45 +28,67 @@ import (
...
+	middlewareInfo      *BaseMiddlewareInfo
+	log                 *logrus.Entry
+	running             bool
+	address             string
+	bitboxBaseConfigDir string
+
+	baseNoiseStaticPubkey         []byte
+	channelHash                   string
+	channelHashAppVerified        bool
+	channelHashBitBoxBaseVerified bool
```
benma: sometimes it is named bitBoxBase, sometimes just base (a few lines above). would be nice to have the same naming everywhere.

backend/bitboxbase/updater/updater.go
```diff
@@ -25,45 +28,67 @@ import (
...
-	//TODO: This is where the initial rest call should go. The initial call initiates the noise authentication pairing.
-	//This should not be asynchronous, the frontend should not show something until the pairing initiates
 	response, err := http.Get("http://" + address + "/")
-	//defer response.Body.Close()
 	if err != nil {
 		updater.log.Println("No response from middleware", err)
 		updater.running = false
 		return err
 	}
+	//defer response.Body.Close()
```
benma: why is this commented?

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
 		return errors.New("updater: Unexpected Response Body Bytes")
 	}
 	updater.running = true
-	go listenWebsocket(updater, bitboxBaseID)
+	updater.log.Printf("connecting to base websocket")
+	client, _, err := websocket.DefaultDialer.Dial("ws://"+updater.address+"/ws", nil)
+	if err != nil {
+		updater.log.Printf("Websocket dial failed: %s", err.Error())
```
benma: return `errp.WithStack(err)` here.

the withstack wrapper is needed for all externally generated errors (stdlib or third party packages). Our own (`errp.New(...)`) automatically have a stack.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	}
+	err = updater.initializeNoise(client, bitboxBaseID)
+	if err != nil {
```
benma: for shortish lines, i like the inlined version:

`if err := ...; err != nil { ... }`

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+		Random:        rand.Reader,
+		Pattern:       noise.HandshakeXX,
+		StaticKeypair: *keypair,
+		Prologue:      []byte("Noise_XX_25519_ChaChaPoly_SHA256"),
+		Initiator:     true,
+	})
+	if err != nil {
+		return errors.New("failed to generate a new noise handshake state for the wallet app communication with the BitBox Base")
+	}
+	//Check the websocket connection
```
benma: comment wrong? next line is sending a msg, not checking a connection.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	err = client.WriteMessage(1, []byte(opICanHasHandShaek))
+	if err != nil {
+		return errors.New("unable to write BitBox Base Handshake request to websocket")
```
benma: use `errp.New()` everywhere, as that adds a stack trace.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	if string(responseBytes) != string(responseSuccess) {
+		return errors.New("no ACK received from BitBox Base as response to hanshake request")
+	}
+
+	// Do 3 part noise 'XX' handshake.
+	msg, _, _, err := handshake.WriteMessage(nil, nil)
+	if err != nil {
+		return errors.New("noise failed to write the first handshake message")
+	}
+	err = client.WriteMessage(websocket.TextMessage, msg)
```
benma: websocket.BinaryMessage, as binary keys are exchanged.

same below.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+		return errors.New("the websocket failed writing the third handshake message")
+	}
+
+	// Check if the user already authenticated the channel binding hash
+	updater.baseNoiseStaticPubkey = handshake.PeerStatic()
+	pairingVerificationRequiredByApp := !updater.configContainsBitBoxBaseStaticPubkey(
+		updater.baseNoiseStaticPubkey)
+	if pairingVerificationRequiredByApp {
+		msg = []byte(opICanHasPairinVerificashun)
+	}
```
benma: if clause should cover all the next lines about verifying the hash. otherwise you are re-sending the previous msg, which is garbage by now.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	err = client.WriteMessage(websocket.TextMessage, updater.sendCipher.Encrypt(nil, nil, msg))
```
benma: it's a great idea to encrypt those messages, but imho it is better to stick to the same protocol as bitbox02.

deviating from the bitbox02 protocol means it's harder to refactor noise into a package which both can use. i think it's better to stick to the same thing, as it's quite a lot of duplicated code that can be abstracted and reused.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	}
+	_, responseBytes, err = client.ReadMessage()
+	if err != nil {
+		return errors.New("the websocket failed writing the pairingVerificationRequiredByBitBoxBase message at the verification stage of the noise handshake")
+	}
+	responseBytes, err = updater.receiveCipher.Decrypt(nil, nil, responseBytes)
+	if err != nil {
+		return errors.New("noise failed decrypting the pairingVerificationRequredByBitBoxBase response bytes at the verification stage of the noise handshake")
+	}
+	pairingVerificationRequiredByBase := false
```
benma: those lines are the same as (a bit shorter):

`pairingVerificationRequiredByBase :=  string(responseBytes) == opICanHasPairinVerificashu`

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	if string(responseBytes) == opICanHasPairinVerificashun {
```
benma: this is a response to the app's `opICanHasPairinVerificashun` request? what if the app doesn't send the request but the base still wants/needs verification?

here too imho it would be worthwhile to use the exact same protocol as in bitbox02, as it covers all possibilities, but even more important, having the same protocol is easier to maintain and think about, document, etc.

backend/bitboxbase/updater/updater.go
```diff
@@ -83,7 +108,150 @@ func (updater *Updater) Connect(address string, bitboxBaseID string) error {
...
+	}
+	err = client.WriteMessage(websocket.TextMessage, updater.sendCipher.Encrypt(nil, nil, msg))
+	if err != nil {
+		return errors.New("the websocket failed writing the pairingVerificationRequiredByApp message at the verification stage of the noise handshake")
+	}
+	_, responseBytes, err = client.ReadMessage()
+	if err != nil {
+		return errors.New("the websocket failed writing the pairingVerificationRequiredByBitBoxBase message at the verification stage of the noise handshake")
+	}
+	responseBytes, err = updater.receiveCipher.Decrypt(nil, nil, responseBytes)
```
benma: there is quite a bit of code-noise with needing four calls to send and receive one message. it would be nice to put those into helper functions (maybe one to send/receive one message, and one to do the same but encrypted)

frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -36,7 +42,7 @@ export class ConnectedBase extends Component<Props, State> {
     constructor(props) {
         super(props);
         this.state = {
-            blockInfo : '',
+            blockInfo : {blocks: 0, difficulty: 0, lightningAlias: 'unknown'},
```
benma: space before `:` :package:

frontends/web/src/components/bitboxbase/connectbase.tsx
```diff
@@ -24,8 +24,14 @@ export interface ConnectedBaseProps {
...
 
+interface BlockInfoType {
+    blocks: number;
+    difficulty: number;
+    lightningAlias: string;
+}
+
 interface State {
-    blockInfo: string;
+    blockInfo: BlockInfoType;
```
benma: could also be `blockInfo?: BlockInfoType`, which is an alis for `BlockInfoType | undefined`, and then start with `undefined`. 

Then you know if you have a blockInfo or not do the workaround below with the dummy initial value.

backend/bitboxbase/updater/updater.go
TheCharlatan: I'd like to make at least one normally encrypted message as part of the handshake though, since it reveals whether you assigned the sendCipher and receiveCipher in the correct order on both sides of the handshake. I don't think that we can unify the two handshake functions soon in a manner that makes sense, since the bitbox02 requires some extra state to carry around and has a completely different communication protocol.

benma: > I'd like to make at least one normally encrypted message as part of the handshake though, since it reveals whether you assigned the sendCipher and receiveCipher in the correct order on both sides of the handshake. 

I see. You can achieve the same with unit tests or manual tests. Once the two request/respond message pairs have been exchanged, sendCipher and receiveCipher are set and good to go (it can't fail). I wouldn't deviate from the bitbox02 handshake protocol because of this, it.

> I don't think that we can unify the two handshake functions soon in a manner that makes sense, since the bitbox02 requires some extra state to carry around and has a completely different communication protocol.

The noise handshake protocol is (can be / should be) the same though, and is like 90% there already. The protocol inside (encrypted, domain specific) does not matter for this abstraction layer.

When refactored, on the initiator (app) side it could be as easy as `noisexx.Handshake(conn, confirmPairingHashCallback)` (with type `Handshake: func(io.ReadWriter, confirmPairingHashCallback string) bool)` or similar) in the beginning, taking care of all the details, used by both the base and bitbox02 parts.

backend/bitboxbase/updater/updater.go
TheCharlatan: This is the intention, either the base will receive garbage, or 'v'. I'd also like to do it strictly sequentially, where both sides have to wait for each other to answer, so I will change the order here.

backend/bitboxbase/updater/updater.go
TheCharlatan: Would something like `error.New("unable to write BitBox Base Handshake request to websocket: %s", errp.WithStack(err))` make sense?

backend/bitboxbase/updater/updater.go
TheCharlatan: Not really sure, if this is needed in the first place, but I'd like to send and receive some packets before I start the noise handshake, to make sure that the websocket is setup correctly.

backend/bitboxbase/bitboxbase.go
TheCharlatan: Can I do this in the next pull request, where I'll clean up the websocket code?

backend/bitboxbase/updater/updater.go
benma: sending garbage doesn't sound so good :) harder to reason about and potentially fragile, e.g. the garbage could start with 'v', etc.

Please do attempt to use the same handshake as bitbox02, a lot of thought has been put into it to make all edge cases work, and having one protocol means less effort, less audit, less reasoning about all cases, and so on.

backend/bitboxbase/updater/updater.go
benma: unfortunately not, as the 2nd arg would be turned into a string (without stack trace) before interpolation.

Just `errp.New` instead of `errors.New` is enough and will attach a stack trace.

backend/bitboxbase/updater/updater.go
benma: there is no reason to believe it is not set up correctly if the handshake passes. You would be effectively testing the noise package at runtime, while it is already unit-tested offline :smile:

backend/bitboxbase/updater/updater.go
TheCharlatan: Unit tests could be used on the desktop app side, but the desktop app has no guarantee that the other party has it correct. Obviously the other party can make unit tests as well, but it sounds preferable to me to have a round of communication that ensures that both parties have it setup correctly. If the setup happens to be incorrect on the base side and it tries to encrypt something with the `receiveCipher`, we'd only notice once we tried to decrypt something later on. From what I can gather from tls for example, having a round of test encryption is standard practice for a handshake.

benma: While I am not sure of the benefit, I guess it's fine to do it, but then it needs to be done correctly:

- send a random (long) challenge, which the base can echo back, and check it comes back correctly
- could do it as the first thing after the handshake (extend the handshake), which is safer and easier than trying to combine it with the previous steps

The current 'v' (or garbage) does not properly test if the encrypted channel works.

TheCharlatan: Sounds good to me, what challenge size would you go for? Would you then also be fine with encrypting the pairing verification request? Is there a reason why the BitBox sends `"\x01"` as the pairing request and not `"v"`?

TheCharlatan: Another key difference is how the `pairingVerificationRequiredByApp` is communicated. In short, in your case it is:
```
if pairingVerificationRequiredByDevice || pairingVerificationRequiredByApp {
    device.communication.SendFrame(opICanHasPairinVerificashun);
    reponse = device.communication.ReadFrame() 
    device.channelHashDeviceVerified = string(response) == responseSuccess
```
And in mine:
```
if pairingVerificationRequiredByApp {
    msg = []byte(opICanHasPairinVerificashun)
}
err = client.WriteMessage(websocket.BinaryMessage, msg) 
if pairingVerificationRequiredByBase || pairingVerificationRequiredByApp {
    _, responseBytes, err := client.ReadMessage()
    updater.channelHashBitBoxBaseVerified = string(responseBytes) == string(responseSuccess)
```
The main difference being that in my implementation the base listens for a request by the app even if no verification will be required, while in yours, the bitbox only receives a message if the pairing verification needs to be done.
This seems to mean that I should make the verification a separate api call for the base, such that it does not start listening indefinitely for a verification request from the wallet app that will never happen. This would require a special message in the base's api. Is it fine, if I add a TODO for now and fix this when we have a proper message protocol in place?

backend/bitboxbase/updater/updater.go
TheCharlatan: Would be nice to have some guidance on when to use it. For example with functions that only return an error, or function calls up to `n` characters in length.

backend/bitboxbase/updater/updater.go
TheCharlatan: On second though it seems weird to me that the app could completely skip over verification, even if the BitBox demands it, by just not sending the verification request.

benma: challenge: 16 or 32 bytes

> Would you then also be fine with encrypting the pairing verification request? 

As elaborated, I'd prefer to have the same protocol as bb02, so not encrypted.

>  Is there a reason why the BitBox sends "\x01" as the pairing request and not "v"?

Technically it is a response status code, not a request op code. Could have been "v" nonetheless, so no good reason ;) In bb02 the response here is basically a boolean `require_pairing_verification`.

 > On second though it seems weird to me that the app could completely skip over verification, even if the BitBox demands it, by just not sending the verification request.

The app cannot skip over it if the bitbox02 demands it. It blocks communication until it is done, if the bitbox requires it. See those lines:

https://github.com/digitalbitbox/bitbox02-firmware/blob/254e7e357c3303abf95afb8fa47ff695f062cabf/src/usb/noise.c#L244-L274

>  This would require a special message in the base's api. Is it fine, if I add a TODO for now and fix this when we have a proper message protocol in place?

The handhshake/verification protocol is kind of independent of the rest (e.g. you already send `v` instead of JSON, which is totally fine here). It's okay to do it later, but prefer to do the handshake right from the start.

TheCharlatan:
Thanks for the review! The handshake should mimic the existing implementation for the bitbox02 now and I added TODOs to the relevant places where type validation is required.

PR #452, opened 2019-07-05
Mobile sidebar

originalblend337:
main changes:
- sidebar to look the same in desktop and mobile
- swiping right from the edge toggles the sidebar
- sidebar now has headers (account and settings)

### DIFF
--- a/frontends/web/src/components/balance/balance.css
+++ b/frontends/web/src/components/balance/balance.css
@@ -92,18 +92,24 @@
 }
 
 @media (max-width: 560px) {
-    .balanceTable {
-        display: table;
-        margin-left: calc(32px + var(--spacing-half));
-    }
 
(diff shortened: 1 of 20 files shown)


benma:
Thanks, works pretty well!

Can't swipe left to collapse the sidebar again.

benma:
If you are already in the 'add  account' screen, clicking the add account `+` symbol does not close the sidebar

Tomasvrba:
Built and tested. Works well, nice!
Got just one change/addition for the last commit (https://github.com/digitalbitbox/bitbox-wallet-app/pull/452/commits/2c05fc87d98e84945068fab0a3477c2158f1a929)
The sidebar only closes when the active item is an account. Doesn't work for `Settings` and `Manage Device`
I would expect the same behavior

PR #457, opened 2019-07-12
Mdns

Tomasvrba:


backend/backend.go
```diff
@@ -150,20 +151,21 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
 		events:      make(chan interface{}, 1000),
 
-		devices:     map[string]device.Interface{},
-		bitboxBases: map[string]bitboxbase.Interface{},
-		keystores:   keystore.NewKeystores(),
-		coins:       map[string]coin.Coin{},
-		accounts:    []accounts.Interface{},
-		log:         log,
+		devices:             map[string]device.Interface{},
+		detectedBitBoxBases: make(map[string]string),
```
benma: would use the same map init style as the line above and below ;)

backend/backend.go
```diff
@@ -150,20 +151,21 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
+		log:                 log,
 	}
 	notifier, err := NewNotifier(filepath.Join(arguments.MainDirectoryPath(), "notifier.db"))
 	if err != nil {
 		return nil, err
 	}
 	backend.notifier = notifier
 
-	backend.baseManager = mdns.NewManager(backend.bitBoxBaseRegister, backend.BitBoxBaseDeregister, backend.config, backend.arguments.BitBoxBaseDirectoryPath())
+	backend.baseManager = mdns.NewManager(backend.BitBoxBaseDetect, backend.BitBoxBasesDetected, backend.bitBoxBaseRegister, backend.BitBoxBaseDeregister, backend.config, backend.arguments.BitBoxBaseDirectoryPath())
```
benma: feels strange to have a function callback here to give back the detected bases to the manager. The manager should already know about them, as it detected them in the first place?

Could go further and keep both the detected and registered bases only in the manager, and expose them with functions. Would mean only one variable in the backend struct `baseManager` instead of three, and probably less duplicated work (keeping track or bases).

(similar improvement maybe possible with devices as well, but that could be done later)

backend/backend.go
```diff
@@ -648,6 +650,17 @@ func (backend *Backend) BitBoxBasesRegistered() map[string]bitboxbase.Interface
 	return backend.bitboxBases
 }
 
+// BitBoxBasesDetected returns a map of IPs and Hostnames of detected Bases.
```
benma: "a map of hostnames to IPs"

backend/bitboxbase/mdns/manager.go
```diff
@@ -0,0 +1,183 @@
...
+		// Start mdns lookup
+		err := micromdns.Lookup(service, entries)
+		if err != nil {
+			manager.log.WithError(err).Error("mDNS lookup failed for service: ", service)
+		}
+
+		<-ctx.Done()
+
+		// Wait one second for go routine shutdown, necessary to correctly filter by "_bitboxbase._tcp"
+		time.Sleep(time.Second * 1)
```
benma: timers can be fragile. Look into sync.WaitGroup (can do wg.Wait() do wait for goroutines to end).

backend/bitboxbase/mdns/manager.go
```diff
@@ -0,0 +1,183 @@
...
+		entries := make(chan *micromdns.ServiceEntry)
+		go func() {
+			for entry := range entries {
+				host := entry.Host
+				resolvedHosts, err := net.LookupHost(host)
+				if err != nil {
+					manager.log.WithError(err).Error("Failed to resolve hostname: ", host)
+					continue
+				}
+				baseIPv4 := resolvedHosts[0] + ":" + strconv.Itoa(entry.Port)
```
benma: check that `resolvedHosts` is not empty, or this might panic.

frontends/web/src/routes/bitboxbase/bitboxbase.tsx
```diff
@@ -62,6 +68,16 @@ class BitBoxBase extends Component<Props> {
         });
     }
 
+    private connect = (ip: string) => {
+        apiPost('bitboxbases/connectbase', { ip })
+        .then(data => {
+            const { success } = data;
```
benma: extra line not needed if there if only use once? `if (!data.success)`

frontends/web/src/routes/bitboxbase/bitboxbase.tsx
```diff
@@ -70,35 +86,46 @@ class BitBoxBase extends Component<Props> {
...
+                    <Header title={<h2>{t('bitboxBase.title')}</h2>} />
                     <div class="innerContainer scrollableContainer">
                         <div class="content padded">
+                            {
+                                detectedBases ?
+                                (
+                                    <div>
+                                        <h3>{t('bitboxBase.detectedBases')}</h3>
+                                        {
+                                            Object.keys(detectedBases).map(hostname => (
```
benma: since you need both key and value, mapping over both at once with `Object.entries(([hostname, ip]) => ... )`  is more suitable

backend/backend.go
Tomasvrba: Agreed, I'll change it :)

Tomasvrba: Changed so detected bases are now kept in the manager and backend requests them with `backend.baseManager.GetDetectedBases()`

Not yet convinced about moving the registered bases to the manager. Right now the job of the manager is to run the mDNS scans and keep track of what it gets from those, not to actually initialize the Bases in the backend. 

This could definitely change, but for now I think it makes sense to register the bases with the backend.

Tomasvrba:
@benma addressed everything except the wait groups for now. Could not get the wait group to work properly, might have to ask you about that tomorrow.

frontends/web/src/routes/bitboxbase/bitboxbase.tsx
```diff
@@ -70,35 +85,46 @@ class BitBoxBase extends Component<Props> {
...
         ) {
             return (
             <div class="contentWithGuide">
                 <div class="container">
-                    <Header title={t('bitboxBase.title')} />
+                    <Header title={<h2>{t('bitboxBase.title')}</h2>} />
                     <div class="innerContainer scrollableContainer">
                         <div class="content padded">
+                            {
+                                detectedBases ?
```
TheCharlatan: Could you always display both? This makes it a bit annoying to test when I have the base middleware running locally and have some bases reachable with mdns at the same time.

Tomasvrba: Done.

PR #458, opened 2019-07-12
Add units tests to the backend handlers

TheCharlatan:
Adds some http request tests to the default handlers that are exposed by the backend. I'm not sure if these tests are actually useful, but it was a good exercise.

### DIFF
--- a/backend/handlers/handlers_test.go
+++ b/backend/handlers/handlers_test.go
@@ -15,27 +15,48 @@
 package handlers_test
 
 import (
+	"bytes"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"regexp"
 	"strings"
 	"testing"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/arguments"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/devices/usb"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/handlers"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/system"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/test"
 	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/require"
 )
+
+// testEnvironment implements backend.Environment
+type testEnvironment struct {
+}
+
+// NotifyUser implements backend.Environment
+func (testEnvironment) NotifyUser(text string) {
+}
+
+// DeviceInfos implements backend.Environment
+func (testEnvironment) DeviceInfos() []usb.DeviceInfo {
+	return usb.DeviceInfos()
+}
+
+// SystemOpen implements backend.Environment
+func (testEnvironment) SystemOpen(url string) error {
+	return system.Open(url)
+}
 
 // List all routes with `go test backend/handlers/handlers_test.go -v`.
 func TestListRoutes(t *testing.T) {
-	const skip = true
-	if skip {
-		t.Skip("manual listing of handlers")
-	}
 	connectionData := handlers.NewConnectionData(8082, "")
 	backend, err := backend.NewBackend(arguments.NewArguments(
 		test.TstTempDir("bitbox-wallet-listroutes-"), false, false, false, false, false),
-		nil,
+		testEnvironment{},
 	)
 	if err != nil {
 		fmt.Println(err)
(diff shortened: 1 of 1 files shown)


backend/handlers/handlers_test.go
```diff
@@ -15,27 +15,68 @@
...
 	"github.com/gorilla/mux"
 )
 
+// webdevEnvironment implements backend.Environment
+type webdevEnvironment struct {
+}
+
+// NotifyUser implements backend.Environment
+func (webdevEnvironment) NotifyUser(text string) {
+	log := logging.Get().WithGroup("servewallet")
```
benma: remove contents of this, you don't want actual notifications in tests ;)

backend/handlers/handlers_test.go
```diff
@@ -15,27 +15,68 @@
...
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/devices/usb"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/handlers"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/system"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/test"
 	"github.com/gorilla/mux"
 )
 
+// webdevEnvironment implements backend.Environment
+type webdevEnvironment struct {
```
benma: rename.

backend/handlers/handlers_test.go
```diff
@@ -57,19 +98,179 @@ func TestListRoutes(t *testing.T) {
...
 		return nil
 	})
 	if err != nil {
 		fmt.Println(err)
 	}
 }
+
+func requestTester(request *http.Request, t *testing.T) *httptest.ResponseRecorder {
+	connectionData := handlers.NewConnectionData(8082, "")
+	backendInstance, err := backend.NewBackend(arguments.NewArguments(
```
benma: if you think it makes it easier, you could also use bridgecommon (just merged https://github.com/digitalbitbox/bitbox-wallet-app/pull/451), which also handles making a backend and invoking backend calls.

backend/handlers/handlers_test.go
```diff
@@ -57,19 +98,179 @@ func TestListRoutes(t *testing.T) {
...
+		test.TstTempDir("test-tempdir-"), false, false, false, false, false),
+		webdevEnvironment{},
+	)
+	if err != nil {
```
benma: We have the `testify` dep, so those become one liners:

`require.NoError(t, err)`

backend/handlers/handlers_test.go
```diff
@@ -57,19 +98,179 @@ func TestListRoutes(t *testing.T) {
...
+	rr := httptest.NewRecorder()
+	handlers.Router.ServeHTTP(rr, request)
+	return rr
+}
+
+func checkResponse(t *testing.T, response *httptest.ResponseRecorder, expectedResponse string) {
+	if status := response.Code; status != http.StatusOK {
+		t.Errorf("Status code differs. Expected %d. Got %d", http.StatusOK, status)
+	}
+	if response.Body.String() != expectedResponse {
```
benma: same here, `require.Equal(response.Body.String(), expectedResponse)`

(and above and everywhere else ;))

backend/handlers/handlers_test.go
```diff
@@ -57,19 +98,179 @@ func TestListRoutes(t *testing.T) {
...
+		t.Errorf("Get Testing response body differs. Expected %q. Got %q", expectedResponse, response.Body.String())
+	}
+}
+
+func TestConfigDefaultHandler(t *testing.T) {
+	req, err := http.NewRequest("GET", "/api/config/default", nil)
+	if err != nil {
+		t.Errorf("An error occurred. %v", err)
+	}
+	rr := requestTester(req, t)
```
benma: Since you need a backend and common functionality in almost all tests, you can use a helper like this, but it might be better to use a testsuite, which has a struct with common vars, and a SetupTest() method which runs before every test. Similar to here:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/transactions/transactions_test.go#L118

backend/handlers/handlers_test.go
```diff
@@ -57,19 +98,179 @@ func TestListRoutes(t *testing.T) {
...
+	}
+
+	req, err = http.NewRequest("POST", "/api/config", bytes.NewBuffer(rr.Body.Bytes()))
+	if err != nil {
+		t.Errorf("An error occurred. %v", err)
+	}
+	rr = requestTester(req, t)
+	if status := rr.Code; status != http.StatusOK {
+		t.Errorf("Status code differs. Expected %d. Got %d", http.StatusOK, status)
+	}
```
benma: you wanna check if backend.NotifyUser() was actually called. For this you want to pass your own implementation of the Environment interface and mock it.

Tools like `mockery` help to autogenerate those, and testify's mocking helps to easily mock.

Example to generate:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/b5b67c42f42eb13d4f33925d2773b77ecbf4ac18/backend/coins/btc/headers/headers.go#L52
Example to use:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/transactions/transactions_test.go#L131

backend/handlers/handlers_test.go
```diff
@@ -57,19 +98,179 @@ func TestListRoutes(t *testing.T) {
...
+}
+
+func TestOpenHandler(t *testing.T) {
+	req, err := http.NewRequest("POST", "/api/open", bytes.NewBuffer([]byte("\"malformed\"")))
+	if err != nil {
+		t.Errorf("An error occurred. %v", err)
+	}
+
+	rr := requestTester(req, t)
+	checkResponse(t, rr, "{\"error\":\"Blocked /open with url: malformed\"}\n")
```
benma: this is not unit testing the handlers, but unit testing the backend, as the backend produces error.

What you want is to check that backend.SystemOpen() was actually called with the right argument, and mock a response. For this you want to pass your own implementation the backend interface and mock it.

Tools like `mockery` help to autogenerate those, and testify's mocking helps to easily mock.

Example to generate:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/b5b67c42f42eb13d4f33925d2773b77ecbf4ac18/backend/coins/btc/headers/headers.go#L52
Example to use:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/transactions/transactions_test.go#L131 

Same applies for other tests here.

It's not high prio to fix this, and it's also an effort, but in the spirit of unit testing, here you only want to test that the handlers do the correct input parsing, the correct backend call, and produce the correct output. The actual logic should be unit tested in backend_test.go

backend/handlers/handlers_test.go
```diff
@@ -15,27 +15,48 @@
...
+type testEnvironment struct {
+}
+
+// NotifyUser implements backend.Environment
+func (testEnvironment) NotifyUser(text string) {
+}
+
+// DeviceInfos implements backend.Environment
+func (testEnvironment) DeviceInfos() []usb.DeviceInfo {
+	return usb.DeviceInfos()
```
benma: same here and the one below, you don't want to actually call anything here. At most, you could check input and mock output, but in theory the handlers unit test would not exercise this code as it is part of backend.go, not handlers.go

backend/handlers/handlers_test.go
```diff
@@ -57,19 +78,129 @@ func TestListRoutes(t *testing.T) {
...
+	require.NoError(t, err)
+	rr := requestTester(req, t)
+	checkResponse(t, rr, "{\"error\":\"Blocked /open with url: malformed\"}\n")
+}
+
+func TestQRHandler(t *testing.T) {
+	req, err := http.NewRequest("GET", "/api/qr", nil)
+	require.NoError(t, err)
+	rr := requestTester(req, t)
+	checkResponse(t, rr, "{\"error\":\"no data to encode\"}\n")
```
benma: no happy test for QR? :P

backend/handlers/handlers_test.go
```diff
@@ -57,19 +78,129 @@ func TestListRoutes(t *testing.T) {
...
+	checkResponse(t, rr, "null\n")
+}
+
+func TestVersionHandler(t *testing.T) {
+	req, err := http.NewRequest("GET", "/api/version", nil)
+	require.NoError(t, err)
+	rr := requestTester(req, t)
+
+	require.Equal(t, rr.Code, http.StatusOK)
+	if !regexp.MustCompile(`([0-9]+\.[0-9]+\.[0-9]+)`).Match(rr.Body.Bytes()) {
```
benma: maybe simpler: `require.JSONEq(rr.Body.String(), jsonp.MustMarshal(backend.Version.String()))`

thisconnect:
@benma close?

benma:
> @benma close?

:+1:

PR #465, opened 2019-07-26
Handle Disconnects of the RPC Connection

TheCharlatan:
The backend now passes a callback to properly unregister through the manager and bitboxbase package to the rpcclient. In case the websocket connection fails, the callback is called and both read and write websocket loops are closed.
Fix the ordering of the close channel to make sure that no double closing, or writing into a closed websocket connection occurs.
The rpc methods now return proper errors and need to get handled based
on their purpose. Also print the errors in their proper logging format.

### DIFF
--- a/backend/bitboxbase/rpcclient/websocket.go
+++ b/backend/bitboxbase/rpcclient/websocket.go
@@ -24,7 +24,6 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/util/observable/action"
 	"github.com/flynn/noise"
 	"github.com/gorilla/websocket"
-	"github.com/sirupsen/logrus"
 )
 
 // runWebsocket sets up loops for sending/receiving, abstracting away the low level details about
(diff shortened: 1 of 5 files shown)


TheCharlatan:
@Tomasvrba this seems to break the frontend a bit, when mdns discovered nodes are connected, they register with their ip, but stay registered with their hostname in the mdns map, thus creating a double entry. We should discuss what exactly to use as an id for the base and need to cleanly incorporate both direct connections and mdns connections into the same disconnect and connect methods.

backend/bitboxbase/rpcclient/websocket.go
```diff
@@ -72,23 +74,31 @@ func (rpcClient *RPCClient) runWebsocket(client *websocket.Conn, writeChan <-cha
...
+					return
+
+				case message, ok := <-writeChan:
+					if !ok {
+						_ = client.WriteMessage(websocket.CloseMessage, []byte{})
+						return
+					}
+					err := client.WriteMessage(websocket.BinaryMessage, rpcClient.sendCipher.Encrypt(nil, nil, message))
+					if err != nil {
+						rpcClient.log.WithFields(logrus.Fields{"group": "rpcClient websocket", "error": err}).Error("websocket could not write message")
```
benma: "group" really needed? in any case, you can use WithError(err).WithFields(...) to not manually put the error there.

Tomasvrba:
> @Tomasvrba this seems to break the frontend a bit, when mdns discovered nodes are connected, they register with their ip, but stay registered with their hostname in the mdns map, thus creating a double entry. We should discuss what exactly to use as an id for the base and need to cleanly incorporate both direct connections and mdns connections into the same disconnect and connect methods.

Yes, the frontend should definitely be considered more of a mock-up for now. In any case, I would remove 
```
if _, ok := manager.detectedBases[address]; !ok {
		manager.detectedBases[address] = address
		manager.onDetect()
	}
```

from `TryMakeNewBase()` as `detectedBases` is meant to be used only by the mDNS scanner to keep track of which bases the manager can see on the network. 
Then we can separate how bases are displayed in the frontend, i.e. a connected base (if connected manually) can be displayed independently of being in `detectedBases` found by the scanner

benma:
Right, I also wondered about this! Good catch @Tomasvrba

PR #468, opened 2019-07-27
Scripts/Regtest: Remove obsolete electrumx env vars

TheCharlatan:
With the current version I get `electrumx.lib.env_base.EnvBase.Error: remove obsolete environment variables ['HOST', 'TCP_PORT', 'SSL_PORT', 'RPC_HOST', 'RPC_PORT']`

### DIFF
--- a/scripts/run_regtest.sh
+++ b/scripts/run_regtest.sh
@@ -60,14 +60,11 @@ docker run \
        -e DAEMON_URL="dbb:dbb@${DOCKER_IP}:10332" \
        -e COIN=BitcoinSegwit \
        -e NET=regtest \
-       -e RPC_PORT=10002 \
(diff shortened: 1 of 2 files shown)


scripts/run_regtest.sh
```diff
@@ -60,12 +60,7 @@ docker run \
        -e DAEMON_URL="dbb:dbb@${DOCKER_IP}:10332" \
        -e COIN=BitcoinSegwit \
        -e NET=regtest \
-       -e RPC_PORT=10002 \
```
benma: don't just delete, but replace with the new way of specifying those?

https://github.com/kyuupichan/electrumx/blob/master/docs/environment.rst#services

benma:
I changed the regtest script to use electrs instead in the meantime.

PR #469, opened 2019-07-28
bitboxbase: tunnel electrum through noise

benma:


benma:
Closing as out of date. Will keep the branch around just in case.

PR #472, opened 2019-08-06
 Frontend/Backend BitBoxBase: Initialization Wizard 

TheCharlatan:
This commit adds a basic initialization wizard to the frontend. It tracks the current position in the wizzard with a status string from the bitboxbase. Changes to the status are communicated to the frontend by the means of websocket notifications.

A connected base now gets its own sidebar entry. According to the wireframe designs, this should be changed in the future, where a view of a singular base replaces the bitboxbaseconnect view.

The wizard currently tracks the progress during the noise pairing with the base. It starts at 'unpaired' and moves to 'initialized' if successful, or 'pairingFailed' if not.

The transitions are triggered by events, either signifying that the noise channel hash state has changed, or that the status has changed.

The noise pairing verification had to be tweaked a bit in order for this commit to work. The BitBox Base middleware server now expects a message even if no verification is required. This makes sure that no rpc notification is emitted by the middleware while the wallet app is still listening for the verification response.


backend/bitboxbase/bitboxbase.go
```diff
@@ -126,19 +172,89 @@ func (base *BitBoxBase) ConnectElectrum() error {
...
+func (base *BitBoxBase) VerificationProgress() (rpcmessages.VerificationProgressResponse, error) {
+	if !base.active {
+		err := errp.New("Attempted a call to non-active base")
+		return rpcmessages.VerificationProgressResponse{}, err
+	}
+	return base.rpcClient.GetVerificationProgress()
+}
+
+// SyncOption returns true if the chosen sync option was executed successfully
+func (base *BitBoxBase) SyncOption(option string) (bool, error) {
```
benma: avoid stringly types. Make it a proper type with const values, then every option can be documented and there is less room for silent errors.

Similar to https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/accounts/event.go

backend/bitboxbase/handlers/handlers.go
```diff
@@ -77,3 +127,15 @@ func (handlers *Handlers) postConnectElectrumHandler(r *http.Request) (interface
...
+func (handlers *Handlers) postSyncOptionHandler(r *http.Request) (interface{}, error) {
+	payload := struct {
+		Option string `json:"option"`
+	}{}
+	if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
+		return nil, err
+	}
+
+	success, err := handlers.base.SyncOption(payload.Option)
+	return map[string]interface{}{"success": success}, err
```
benma: don't return `err`, as it prevents sending the real response to the app. See https://github.com/digitalbitbox/bitbox-wallet-app/blob/9b7eebccf89434cc7b20c64e83b4bdb41940d75a/backend/handlers/handlers.go#L727-L732

Note that in the compiled app, no response will be relayed at all in this case, not even the one you can see in the above lines.

Ideally, never return an error from handler functions. We need to decide at some point if we wanna get rid of the error response completely, or reserve it for unexpected errors that just make a log entry and are otherwise silently ignored.

Tomasvrba:
Rebased and updated in a new PR here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/500

PR #474, opened 2019-08-11
Add (Tor) Socks Proxy Support

TheCharlatan:
This is just a preliminary pull request. If tor is installed on the system, all electrum traffic, rates fetching, etherscan checks and update checks are routed over tor. Things left to do:
- [ ] Add configuration to the socksproxy
- [ ] Allow configuration to be tweaked from frontend (different ip and port)

@benma do you have a preference on how / where the configuration (to be precise the address of the socks5 proxy and wether it should be use) should be stored?


backend/proxy/proxy.go
```diff
@@ -0,0 +1,78 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		return dialer.Dial("tcp", address)
+	}
+	return net.Dial("tcp", address)
+}
+
+// GetHTTPClient returns a http client. Requests made with this client are proxied, if useProxy is true.
+func (socksProxy *SocksProxy) GetHTTPClient() *http.Client {
```
benma: the error branches should probably return the error

backend/backend.go
```diff
@@ -247,11 +251,11 @@ func (backend *Backend) CreateAndAddAccount(
 	switch specificCoin := coin.(type) {
 	case *btc.Coin:
 		account = btc.NewAccount(specificCoin, backend.arguments.CacheDirectoryPath(), code, name,
```
benma: account does not need socksProxy in general.

Better to pass it to `*\.NewCoin()`, as the coin instances perform the communication.

This will reduce the diff quite a bit.

benma:
> do you have a preference on how / where the configuration (to be precise the address of the socks5 proxy and wether it should be use) should be stored?

Yeah, this will fit nicely into config.Backend

backend/proxy/proxy.go
```diff
@@ -0,0 +1,78 @@
...
+func NewSocksProxy(useProxy bool, proxyAddress string) SocksProxy {
+	proxy := SocksProxy{
+		useProxy:     useProxy,
+		proxyAddress: proxyAddress,
+	}
+	return proxy
+}
+
+// GetTCPConnection returns a tcp connection. The connection is proxied, if useProxy is true.
+func (socksProxy *SocksProxy) GetTCPConnection(address string) (net.Conn, error) {
```
benma: A bit higher level:

`GetDialer(address string) (net.Dialer, error)`

then you don't need to do the dialing here.

backend/proxy/proxy.go
```diff
@@ -0,0 +1,78 @@
...
+	if socksProxy.useProxy {
+		// Create a proxy that uses Tor's SocksPort.
+		dialer, err := proxy.SOCKS5("tcp", socksProxy.proxyAddress, nil, nil)
+		if err != nil {
+			log.Fatal(err)
```
benma: use our logging infrastructure. `logging....Panic()`

backend/proxy/proxy.go
TheCharlatan: Ha, that's what I thought initially and was then surprised by the proxy dialer returning a dialer of its own type and not a net.Dialer. Only when connecting a generic net.Conn is returned. Using the proxy dialer directly would complicate the calling function.

backend/backend.go
TheCharlatan: I can easily move it there, but don't really think this will reduce the diff. Account just holds it in its struct until coin.Initialize() is called. My hunch is this will actually be more code with more repetition, since we call `NewCoin` way more than `NewAccount`.

TheCharlatan:
The socks proxy configuration is now stored in the backend. I added a checkbox to the frontend that configures the config to connect to the tor socks5 proxy. Once the app is restarted, all subsequent connections will run over tor.

backend/backend.go
benma: Ok the diff might not be smaller, but Coin is still the better/correct home for it ;) It is confusing to pass it through account just to pass it to coin.Initialize().

Also I slightly dislike argument to Initialize() functions, imho the instance having the pre-init data it needs at creation time is a bit nicer / concentrated.

backend/proxy/proxy.go
```diff
@@ -0,0 +1,93 @@
+// Copyright 2019 Shift Devices AG
```
benma: `util/` is I think the better location

Also the filename should be the packagename, e.g. `socksproxy/socksproxy.go`

backend/proxy/proxy.go
benma: I thought it was the same too, but it could work by returning `proxy.Dialer`, which is an interface which `net.Dialer` implements

TheCharlatan: That works, but then I cannot return an unproxied dialer.

benma: Why not? You can return `net.Dialer{}` which implements `proxy.Dialer` and is unproxied.

TheCharlatan: net.Dialer does not implement proxy.Dialer. net.DIaler's Dial has a pointer receiver, while ([i.e. an instance of net.Dialer](https://golang.org/pkg/net/#Dialer.Dial)), while proxy.Dialer has [a non-pointer receiver](https://godoc.org/golang.org/x/net/proxy#Dial).

benma: Alright, `&net.Dialer{}` implements it then :smile: 

```Go
package main

import (
	"fmt"
	"net"

	"golang.org/x/net/proxy"
)

func main() {
	var dialer proxy.Dialer = &net.Dialer{}
	_, err := dialer.Dial("tcp", "golang.org:80")
	fmt.Println(err)
}
```

TheCharlatan: I should stop reading documentation while im in some Ravine :P

backend/backend.go
```diff
@@ -162,10 +165,15 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 		return nil, err
 	}
 	backend.notifier = notifier
+	backend.socksProxy = socksproxy.NewSocksProxy(backend.config.AppConfig().Backend.UseProxy, backend.config.AppConfig().Backend.ProxyAddress)
+	backend.log.Println("Creating updater instance")
```
benma: `Println()` -> `Info()`, here and in more places. We do not append newlines anywhere else (I think this is automatically done), and `Info` is more descriptive.

backend/rates.go
```diff
@@ -44,26 +44,34 @@ var (
...
 // NewRatesUpdater returns a new rates updater.
-func NewRatesUpdater() *RatesUpdater {
-	updater := &RatesUpdater{
-		last: map[string]map[string]float64{},
-		log:  logging.Get().WithGroup("rates"),
-	}
-	go updater.start()
-	return updater
+func NewRatesUpdater(socksProxy socksproxy.SocksProxy) *RatesUpdater {
+	ratesUpdaterInstanceOnce.Do(func() {
```
benma: Please remove the global now that it is not a real global var anymore.

The other option would be to have one global per possible parameter input to this function, but that would be silly ;)

backend/backend.go
```diff
@@ -162,10 +165,15 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
+	_ = NewRatesUpdater(backend.socksProxy)
```
benma: If you remove the singleton, then this workaround/hack would not be necessary. Just keep the rates updater in the backend struct.

Could be a separate commit to make the rates not a global var, but a part of the Backend struct.

TheCharlatan:
Proxying the geth rpc client seems a bit more involved. I asked a few days ago on the geth gitter if someone did that already, but did not receive an answer. I will open an issue on goethereum to ask them to expose a dialer (or provide instructions for a workaround).

backend/backend.go
TheCharlatan: Already started on this, hope to have it finished by tonight.

TheCharlatan: Do you have an opinion on how to pass the rates to the coin/btc handlers for the coin conversion rates?

TheCharlatan:
There is an open issue for geth socks proxy support that is nearly as old as geth itself https://github.com/ethereum/go-ethereum/issues/1855 :/
The reasons given puzzle me a bit. Seems like they have not given this a lot of thought yet, even though karalabe has contributed to the tor<-> go bindings: https://github.com/cretz/bine

backend/backend.go
benma: Either pass it to the NewAccount() constructors, or pass to account.Initialize() as an argument.

accounts.Interface then will have a `RatesUpdader() *backend.RatesUpdater` method (or `rates.Updater` if moved to its own package).

benma:
> Proxying the geth rpc client seems a bit more involved. 

Does it not work by just using `rpc.DialHTTPWithClient()`?

TheCharlatan:
You are so right, the connection is proxied now using `DIalHTTPWithClient()`

backend/backend.go
TheCharlatan: I went with creating a new rates package and passing the `rateUpdater` struct to the `NewAccount` constructors.

TheCharlatan:
I have not tested running the proxy with a local node yet, but I have the feeling that won't work. Not sure how to handle this error case, maybe display something to the user about running local full nodes with a proxy.

benma:
If it does not work, then that does not sound like a good solution. You might want to have the proxy when accessing the rates and the remote nodes *and* run your full node for one of the coins.

But who knows, maybe the proxy is smart enough and it just works? Please test :)

backend/coins/btc/handlers/handlers.go
```diff
@@ -91,41 +89,11 @@ type FormattedAmount struct {
...
-	position := strings.Index(formatted, ".") - 3
-	for position > 0 {
-		formatted = formatted[:position] + "'" + formatted[position:]
-		position -= 3
-	}
-	return formatted
-}
-
-// Conversions handles fiat conversions
-func Conversions(amount coin.Amount, coin coin.Coin, isFee bool) map[string]string {
```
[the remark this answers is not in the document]
TheCharlatan: The Conversions function is used twice, once by the btc handlers and once by the backend handlers. Before, the backend handlers could just call the Conversion function in the btc handlers, because it used the rates global instance. So I can either do it like I do it in this patch, or define the Conversion function twice, once as part of the backend handlers and once for the btc handlers.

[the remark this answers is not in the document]
benma: Or: expose `backend.RatesUpdater()` and have `Conversions()` take the rates updater as an additional param.

`Conversions()` used to be local to account handlers, but with the account summary the backend handlers started making use of it. At that time, moving this function to its own file would have been good. I'd moveit into `backend/coins/coin/conversions.go`, as all params are from that package, add the rates updater as a param, and use it like this from everywhere.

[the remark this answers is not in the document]
TheCharlatan: Done, thanks for the description.

TheCharlatan:
At least for localhost, the proxy seems to need some configuring on the host before routing back to localhost. I'm not sure how to handle this properly. The two solutions that come to mind to me are either instructing the user to configure it and providing instructions for all platforms, or adding fine grained proxy options that allow skipping proxying of locally hosted services.

TheCharlatan:
Rebased.

TheCharlatan:
Seems like I'm tripping over all the globals in this one. Not quite sure why Travis complains though, since the test runs fine on my machine.

TheCharlatan:
Rebased.

TheCharlatan:
Rebased.

PR #488, opened 2019-08-19
Build issues

benma:


hardening.mk.inc
```diff
@@ -1,5 +1,4 @@
 CFORTIFY=-O2 -D_FORTIFY_SOURCE=2
-CASLR=-fPIC -fPIE
```
x1ddos: This wasn't even used and didn't realy make sense at all, mixing PIC with PIE.

x1ddos:
Looking forward to a new release.

hardening.mk.inc
benma: :+1:

benma:
> I was able to run the appimage. Would've preferred the PIE and unrelated changes in separate PRs though.

The PR has two commits, one for each build issue. Not sure how two PRs would have helped?

Thanks for testing!

x1ddos:
> The PR has two commits, one for each build issue. Not sure how two PRs would have helped?

Never mind. I think I'm too picky. It's just I find it useful to link to code reviews in other discussions. But this PR suddenly contains many other distracting changes and it makes it harder to follow because when I go to "Files changed" it shows a diff with all commits by default. We both know what is what but it might not be obvious to an occasional viewer. GitHub doesn't really matter though. Could've been an email thread with unrelated patches attached.

I shouldn't have mentioned libxcb troubles in that bug. :)

PR #491, opened 2019-08-20
Logos

benma:


thisconnect:
@benma wouldn't it make sense to use PNG directly instead of image/png;base64 inside svg?

benma:
Great catch @thisconnect ! I was just blindly converting to the same format.

Fixed now.

thisconnect:
thank you

PR #500, opened 2019-08-29
Base setup

Tomasvrba:
Previous description from @TheCharlatan:

"This commit adds a basic initialization wizard to the frontend. It tracks the current position in the wizzard with a status string from the bitboxbase. Changes to the status are communicated to the frontend by the means of websocket notifications.

A connected base now gets its own sidebar entry. According to the wireframe designs, this should be changed in the future, where a view of a singular base replaces the bitboxbaseconnect view.

The wizard currently tracks the progress during the noise pairing with the base. It starts at 'unpaired' and moves to 'initialized' if successful, or 'pairingFailed' if not.

The transitions are triggered by events, either signifying that the noise channel hash state has changed, or that the status has changed.

The noise pairing verification had to be tweaked a bit in order for this commit to work. The BitBox Base middleware server now expects a message even if no verification is required. This makes sure that no rpc notification is emitted by the middleware while the wallet app is still listening for the verification response."


backend/bitboxbase/bitboxbase.go
```diff
@@ -44,17 +48,49 @@ type Interface interface {
...
+	SyncOptionStatus(SyncOption) (bool, error)
 }
 
+// SyncOption is a user provided blockchain sync option during BBB initialization
+type SyncOption int
+
+// SyncOption iota has three options:
+// Accept pre-synchronized blockchain; delete the chainstate and reindex; resync bitcon from scratch with an IBD
+const (
+	PRESYNCED SyncOption = iota
```
benma: Call them `SyncOptionPresynced`, `SyncOptionReindex`, `SyncOptionInitianBlockDownload`

Need `SyncOption` prefix for namespacing.

Also pls put the documentation for each var right before the var in godoc style, like in `event.go`

backend/bitboxbase/bitboxbase.go
```diff
@@ -232,22 +243,22 @@ func (base *BitBoxBase) VerificationProgress() (rpcmessages.VerificationProgress
 	return base.rpcClient.GetVerificationProgress()
 }
 
-// SyncOption returns true if the chosen sync option was executed successfully
-func (base *BitBoxBase) SyncOption(option string) (bool, error) {
+// SyncOptionStatus returns true if the chosen sync option was executed successfully
+func (base *BitBoxBase) SyncOptionStatus(option SyncOption) (bool, error) {
```
benma: imho this is not a good name, as it is not fetching a status, but performing the sync action?

What about `SyncWithOption`?

Tomasvrba: Nice, sounds better

backend/bitboxbase/bitboxbase.go
Tomasvrba: ack, changed from iota to strings

PR #504, opened 2019-08-31
Improve Mobile Pairing Error Messages

TheCharlatan:
When the relay server is not reachable, the frontend just displays a timeout message, while the backend logs an unrelated error. This commit attempts to improve this by displaying a separate message in the frontend when messages are not relayed correctly and improves the logging in the backend.

### DIFF
--- a/backend/devices/bitbox/events.go
+++ b/backend/devices/bitbox/events.go
@@ -25,6 +25,12 @@ const (
 	// EventPairingTimedout is fired when the pairing timed out.
 	EventPairingTimedout event.Event = "pairingTimedout"
 
+	// EventPullFailed is fired when a message cannot be pulled through the pairing relay server.
+	EventPairingPullMessageFailed event.Event = "pairingPullMessageFailed"
+
+	// EventPairingScanningFailed is fired when the mobile does not respond with success after scanning the pairing code.
+	EventPairingScanningFailed event.Event = "pairingScanningFailed"
+
 	// EventPairingAborted is fired when the pairing aborted.
 	EventPairingAborted event.Event = "pairingAborted"
 
--- a/backend/devices/bitbox/pairing.go
+++ b/backend/devices/bitbox/pairing.go
@@ -46,13 +46,30 @@ func (device *Device) finishPairing(channel *relay.Channel) {
 	device.fireEvent(EventPairingSuccess, nil)
 }
 
+func (device *Device) handlePairingError(err error, message string) {
+	switch err.Error() {
+	case relay.PullFailedError:
+		device.log.Errorf("Failed to pull the mobile's %s.", message)
+		device.fireEvent(EventPairingPullMessageFailed, nil)
+	case relay.ResponseTimeoutError:
+		device.log.Errorf("Failed to wait for the mobile's %s.", message)
+		device.fireEvent(EventPairingTimedout, nil)
+	default:
+	}
+}
+
 // processPairing processes the pairing after the channel has been displayed as a QR code.
 func (device *Device) processPairing(channel *relay.Channel) {
-	if err := channel.WaitForScanningSuccess(time.Minute); err != nil {
-		device.log.WithError(err).Warning("Failed to wait for the scanning success.")
-		device.fireEvent(EventPairingTimedout, nil)
+	status, err := channel.WaitForScanningSuccess(10 * time.Second)
+	if err != nil {
+		device.handlePairingError(err, "scanning success message")
 		return
 	}
+	if status != "success" {
+		device.log.Error("Scanning unsuccessful")
+		device.fireEvent(EventPairingScanningFailed, nil)
+	}
+
 	deviceInfo, err := device.DeviceInfo()
 	if err != nil {
 		device.log.WithError(err).Error("Failed to check if device is locked or not")
@@ -67,8 +84,7 @@ func (device *Device) processPairing(channel *relay.Channel) {
 	device.fireEvent(EventPairingStarted, nil)
 	mobileECDHPKhash, err := channel.WaitForMobilePublicKeyHash(2 * time.Minute)
 	if err != nil {
-		device.log.WithError(err).Warning("Failed to wait for the mobile's public key hash.")
-		device.fireEvent(EventPairingTimedout, nil)
+		device.handlePairingError(err, "public key hash")
 		return
 	}
 	bitboxECDHPKhash, err := device.ecdhPKhash(mobileECDHPKhash)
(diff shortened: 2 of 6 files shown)


backend/devices/bitbox/events.go
```diff
@@ -25,6 +25,12 @@ const (
 	// EventPairingTimedout is fired when the pairing timed out.
 	EventPairingTimedout device.Event = "pairingTimedout"
 
+	// EventPullFailed is fired when a message cannot be pulled through the pairing relay server.
+	EventPairingPullMessageFailed device.Event = "pairingPullMessageFailed"
+
+	// EventPairingScanningFailed is fired when the mobile does not respond with success after scanning the pairing code.
+	EventPairingScanningFailed = "pairingScanningFailed"
```
benma: type ` device.Event` missing

backend/devices/bitbox/pairing.go
```diff
@@ -46,13 +46,30 @@ func (device *Device) finishPairing(channel *relay.Channel) {
 	device.fireEvent(EventPairingSuccess, nil)
 }
 
+func (device *Device) handlePairingError(err error, status string, message string) {
+	switch status {
+	case "pullFailed":
+		device.log.WithError(err).Warningf("Failed to pull the mobile's %s.", message)
```
benma: `Errorf` sounds more appropriate.

Maybe don't log the actual message, as try to avoid logging sensitive info

backend/devices/bitbox/relay/channel.go
```diff
@@ -171,14 +171,8 @@ func (channel *Channel) waitForValue(duration time.Duration, name string) (strin
...
 // Returns nil if the scanning success was retrieved from the relay server and an error otherwise.
-func (channel *Channel) WaitForScanningSuccess(duration time.Duration) error {
-	value, err := channel.waitForValue(duration, "id")
-	if err != nil {
-		return err
-	} else if value != "success" {
-		return errp.New("Scanning unsuccessful")
-	}
-	return nil
+func (channel *Channel) WaitForScanningSuccess(duration time.Duration) (string, error) {
```
benma: delete this function? it is too trivial of a wrapper now. The docstring is also outdated.

backend/devices/bitbox/relay/channel.go
```diff
@@ -148,13 +148,13 @@ func (channel *Channel) waitForValue(duration time.Duration, name string) (strin
 			unlock()
 			message, err := PullOldestMessage(relayServer(), channel)
 			if err != nil {
-				return "", err
+				return "pullFailed", err
```
benma: By convention, if a function returns an error, all other response values should be treated as undefined and not inspected.

In this case, it is a better fit to return `"", ErrPullFailed` or similar,  have the caller of the function check against this.

After all, you are not receiving the "pullFailed" value from the sever.

backend/devices/bitbox/relay/channel.go
TheCharlatan: Should I delete the other WaitFor* functions as well?

benma: Hmm good point. I would not, as they are nicely documented and seem to mirror the relay server api.

In that case, this one should also be kept then, but with updated documentation.

Or keep the function as it was, returning an error only, but have the caller just inspect the error to determine what action to take (ErrScanningFailed, ErrPullFailed, ...).

TheCharlatan: I prefer just updating the documentation then, since the other WaitFor* functions use a similar pattern.

TheCharlatan:
Anything left to be done here?

frontends/web/src/locales/en/app.json
```diff
@@ -668,9 +668,17 @@
       "text": "Something went wrong. Please start again.",
       "title": "Error"
     },
+    "pullFailed": {
+      "text": "Failed to pull a message from your mobile through the relay server. The relay server might be offline, please contact support.",
+      "title": "Pull Failed"
```
benma: remove title case :P

backend/devices/bitbox/relay/channel.go
```diff
@@ -148,13 +148,13 @@ func (channel *Channel) waitForValue(duration time.Duration, name string) (strin
 			unlock()
 			message, err := PullOldestMessage(relayServer(), channel)
 			if err != nil {
-				return "", err
+				return "", errp.New("pullFailed")
```
benma: normally you shouldn't match error text, as error text should be descriptive and should be able to change without breaking anything.

the usual solution is to return a global `var ErrPullFailed = errors.New("<descriptive error>")` and return this instead, and compare against this.

I'm ok skipping this fix however, up to you.

frontends/web/src/locales/en/app.json
TheCharlatan: lol, removed.

backend/devices/bitbox/relay/channel.go
TheCharlatan: Introducing a global seems like a bad solution as well. How about a function returning the error given const string arguments?

TheCharlatan: Pushed a commit to change to constant error values. Please review.

benma: seems like it, but it is not ;) `Err... error` global vars should be treated as immutable, and immutable globals are not evil as opposed to mutable ones. you would mark it as `const` if Go would let you, but it does not :(

anyway, merging!

benma: To elaborate; it's a normal Go pattern, also used in the stdlib, for example the list of io errors:

https://golang.org/pkg/io/#pkg-variables

TheCharlatan: Huh, so structs with no public members are counted as immutable?

benma: the struct is not mutable, but the global variable is.

```Go
somepackage.ErrShouldBeImmutable = errors.New("lol i changed you")
```

too bad Go does not let you mark more stuff as const.

benma: Example:

```Go

package main

import (
	"bytes"
	"errors"
	"fmt"
	"io"
)

func main() {
	io.EOF = errors.New("this is not the string you were looking for")
	buf := bytes.NewReader(nil)
	_, err := buf.ReadByte()
	fmt.Println(err)
}
```

```
go run main.go
this is not the string you were looking for
```

x1ddos: Alternatively, another way some advocate is to use interfaces. For instance: 
https://play.golang.org/p/uTa56pGLcBC

But these days probably the best is to use the new Is and As helpers from the stdlib errors package https://golang.org/pkg/errors/.
The blog post at https://blog.golang.org/go1.13-errors describes all the origins sentitel errors.

PR #505, opened 2019-09-03
Docs: Correct keypath description to 48'

TheCharlatan:


backend/keystore/keystore.go
```diff
@@ -28,7 +28,7 @@ var ErrSigningAborted = errors.New("signing aborted by user")
 // Keystore supports hardened key derivation according to BIP32 and signing of transactions.
 type Keystore interface {
 	// // Configuration returns the configuration of the keystore.
-	// // The keypath is m/44' for singlesig and m/46' for multisig.
+	// // The keypath is m/44' for singlesig and m/48' for multisig.
```
benma: could just as well delete the whole thing as it is commented? 44' is also not accurate for singlesig...

PR #506, opened 2019-09-05
Base setup RPCs and handlers  

0xB10C:
Together with https://github.com/digitalbitbox/bitbox-base/pull/176 this adds the RPCs and their handlers needed for the Base setup in the App.

### DIFF
--- a/backend/bitboxbase/bitboxbase.go
+++ b/backend/bitboxbase/bitboxbase.go
@@ -72,8 +72,11 @@ type Interface interface {
 	// If bitboxbase is not active, an error is returned.
 	Deregister() (bool, error)
 
-	// SyncOption returns true if the bitcoin resync rpc call executed successfully
-	SyncWithOption(SyncOption) (bool, error)
+	// ReindexBitcoin starts a bitcoin reindex on the base.
+	ReindexBitcoin() (bool, error)
+
+	// ResyncBitcoin starts a bitcoin resync on the base.
+	ResyncBitcoin() (bool, error)
 
 	// GetHostname returns the hostname of the BitBox Base
 	GetHostname() (string, error)
@@ -275,29 +278,42 @@ func (base *BitBoxBase) VerificationProgress() (rpcmessages.VerificationProgress
 	return base.rpcClient.GetVerificationProgress()
 }
 
-// SyncWithOption returns true if the chosen sync option was executed successfully
-func (base *BitBoxBase) SyncWithOption(option SyncOption) (bool, error) {
+// ReindexBitcoin returns true if the chosen sync option was executed successfully
+func (base *BitBoxBase) ReindexBitcoin() (bool, error) {
 	if !base.active {
 		err := errp.New("Attempted a call to non-active base")
 		return false, err
 	}
-	switch option {
-	case SyncOptionPresynced:
-		base.changeStatus(bitboxbasestatus.StatusInitialized)
-		return true, nil
-	case SyncOptionReindex:
-		base.log.Println("bitboxbase is making a ReindexBitcoin call")
-		replySuccess, err := base.rpcClient.ReindexBitcoin()
-		base.changeStatus(bitboxbasestatus.StatusInitialized)
-		return replySuccess.Success, err
-	case SyncOptionInitialBlockDownload:
-		base.log.Println("bitboxbase is making a ResyncBitcoin call")
-		replySuccess, err := base.rpcClient.ResyncBitcoin()
-		base.changeStatus(bitboxbasestatus.StatusInitialized)
-		return replySuccess.Success, err
-	default:
-		return true, nil
+	base.log.Println("bitboxbase is making a ReindexBitcoin call")
+	reply, err := base.rpcClient.ReindexBitcoin()
+	base.changeStatus(bitboxbasestatus.StatusInitialized)
+	if err != nil {
+		return false, err
 	}
+	if !reply.Success {
+		err := errp.Newf("ReindexBitcoin was not successful. Message: '%s'. Code: '%s'", reply.Message, reply.Code)
+		return false, err
+	}
+	return true, nil
+}
+
+// ResyncBitcoin returns true if the chosen sync option was executed successfully
+func (base *BitBoxBase) ResyncBitcoin() (bool, error) {
+	if !base.active {
+		err := errp.New("Attempted a call to non-active base")
+		return false, err
+	}
+	base.log.Println("bitboxbase is making a ResyncBitcoin call")
+	reply, err := base.rpcClient.ResyncBitcoin()
+	base.changeStatus(bitboxbasestatus.StatusInitialized)
+	if err != nil {
+		return false, err
+	}
+	if !reply.Success {
+		err := errp.Newf("ResyncBitcoin was not successful. Message: '%s'. Code: '%s'", reply.Message, reply.Code)
+		return false, err
+	}
+	return true, nil
 }
 
 // GetHostname returns the hostname of the bitboxbase
--- a/backend/bitboxbase/bitboxbase.go
+++ b/backend/bitboxbase/bitboxbase.go
@@ -88,6 +88,24 @@ type Interface interface {
 	// UserAuthenticate returns is the authentication with a username and password was successful
 	// TODO: this is a dummy
 	UserAuthenticate(string, string) (bool, error)
+
+	// MountFlashdrive checks for a flashdrive and then mounts it
+	MountFlashdrive() (bool, error)
+
+	// UnmountFlashdrive unmounts a mounted flashdrive
+	UnmountFlashdrive() (bool, error)
+
+	// BackupSysconfig backups the system config to the flashdrive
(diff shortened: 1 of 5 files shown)


backend/bitboxbase/bitboxbase.go
```diff
@@ -254,20 +286,205 @@ func (base *BitBoxBase) SyncWithOption(option SyncOption) (bool, error) {
 		base.changeStatus(bitboxbasestatus.StatusInitialized)
 		return true, nil
 	case SyncOptionReindex:
-		base.log.Println("bitboxbase is making a ReindexResyncOption call")
-		replySuccess, err := base.rpcClient.ResyncBitcoin(rpcmessages.Reindex)
+		base.log.Println("bitboxbase is making a ReindexBitcoin call")
+		replySuccess, err := base.rpcClient.ReindexBitcoin()
 		base.changeStatus(bitboxbasestatus.StatusInitialized)
 		return replySuccess.Success, err
```
0xB10C: This function should be split up into two to be consistent with the other ones in this file.

backend/bitboxbase/rpcclient/rpcclient.go
```diff
@@ -213,24 +211,154 @@ func (rpcClient *RPCClient) GetSampleInfo() (rpcmessages.SampleInfoResponse, err
...
 }
+
+// ReindexBitcoin makes a synchronous rpc call to the base and returns a ErrorResponse indicating if the called script was successfully executed.
+func (rpcClient *RPCClient) ReindexBitcoin() (rpcmessages.ErrorResponse, error) {
+	rpcClient.log.Println("Executing ReindexBitcoin rpc call")
+	var reply rpcmessages.ErrorResponse
+	err := rpcClient.client.Call("RPCServer.ReindexBitcoin", true /*dummy Arg */, &reply)
+	if err != nil {
+		rpcClient.log.WithError(err).Error("ReindexBitcoin RPC call failed")
+		return reply, err
```
Tomasvrba: This comment goes for all the calls in `rpcclient.go`
Since we are ignoring everything in the reply if there is an error, we probably don't need to return it and just do `return nil, err`?
Also, since this would be an error coming from the `RPCClient` and not from us on the Base, might be a good idea to wrap it with `return nil, errp.WithStack(err)` so we have a stack trace in bitboxbase.go and there we could do something like `log.WithField("error", err)` or `log.WithField("error", err).Panic(err)`

backend/bitboxbase/bitboxbase.go
Tomasvrba: Yes, let's split it up before merging

backend/bitboxbase/handlers/handlers.go
```diff
@@ -117,6 +138,130 @@ func (handlers *Handlers) getMiddlewareInfoHandler(r *http.Request) (interface{}
 	}, nil
 }
 
+func (handlers *Handlers) getGetHostnameHandler(r *http.Request) (interface{}, error) {
```
Tomasvrba: Since you only use the `r` when you're posting and decoding some data, you can ignore it in all the get requests `getGetHostnameHandler(_ *http.Request)`

backend/bitboxbase/bitboxbase.go
```diff
@@ -74,6 +74,38 @@ type Interface interface {
 
 	// SyncOption returns true if the bitcoin resync rpc call executed successfully
 	SyncWithOption(SyncOption) (bool, error)
+
+	// GetHostname returns the hostname of the BitBox Base
+	GetHostname() (string, error)
+
+	// SetHostname sets the hostname of the BitBox Base
+	SetHostname(string) (bool, error)
```
Tomasvrba: This applies to all the getters that return `(bool, error)`
With how we designed our error handling we can always assume success if there is no error so the success bool is redundant right? So we could only have an `error` return type everywhere and `return nil` when successful. 
Then in the handlers you don't have to assign two variables with `success, err := handlers.base...` and can just return `return map[string]interface{}{"success": true}, nil` instead of `return map[string]interface{}{"success": success}, nil`

backend/bitboxbase/handlers/handlers.go
Tomasvrba: Oh, also, we don't use `Get` in the namespace for getters in the app so you could just call it `HostnameHandler` and you'll have `getHostanameHandler` instead of `getGet...`

backend/bitboxbase/handlers/handlers.go
```diff
@@ -49,13 +59,24 @@ func NewHandlers(
 ) *Handlers {
 	handlers := &Handlers{log: log.WithField("bitboxbase", "base")}
 
+	handleFunc("/status", handlers.getStatusHandler).Methods("GET")
+	handleFunc("/gethostname", handlers.getGetHostnameHandler).Methods("GET")
+	handleFunc("/channel-hash", handlers.getChannelHashHandler).Methods("GET")
+	handleFunc("/mountflashdrive", handlers.getMountFlashdriveHandler).Methods("GET")
```
Tomasvrba: change relevant requests from GET to POST as discussed offline for any requests that modify anything on the Base

Tomasvrba:
Nice work! :100:

0xB10C:
Addressed feedback

backend/bitboxbase/handlers/handlers.go
```diff
@@ -117,7 +137,130 @@ func (handlers *Handlers) getMiddlewareInfoHandler(r *http.Request) (interface{}
...
+		Password string `json:"password"`
+	}{}
+	if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
+		return bbBaseError(err, handlers.log), nil
+	}
+	err := handlers.base.UserAuthenticate(payload.Username, payload.Password)
+	if err != nil {
+		return bbBaseError(err, handlers.log), nil
+	}
+	return map[string]interface{}{"success": true}, nil
```
TheCharlatan: Nit: Can you be consistent with inlining or indenting this line?

0xB10C:
Addressed feedback

backend/bitboxbase/bitboxbase.go
```diff
@@ -185,15 +220,15 @@ func (base *BitBoxBase) ConnectElectrum() error {
...
-		base.active = false
-		return true, nil
-	}
-	return false, errp.New("Attempted call to non-active base")
+func (base *BitBoxBase) Deregister() error {
+	if !base.active {
+		return errp.New("Attempted call to non-active base")
+	}
+	// let the frontend know that the base is disconnected
+	base.fireEvent("disconnect")
```
benma: `"disconnect"` is a magic string, but should be a `bitboxbasestatus.Event`. Please move it to a definition with docs in bitboxbase/status.go.

I wish Go would not automatically cast between type aliases :)

backend/bitboxbase/bitboxbase.go
```diff
@@ -228,44 +263,220 @@ func (base *BitBoxBase) RPCClient() *rpcclient.RPCClient {
...
+// ReindexBitcoin returns true if the chosen sync option was executed successfully
+func (base *BitBoxBase) ReindexBitcoin() error {
+	if !base.active {
+		return errp.New("Attempted a call to non-active base")
+	}
+	base.log.Println("bitboxbase is making a ReindexBitcoin call")
+	reply, err := base.rpcClient.ReindexBitcoin()
+	base.changeStatus(bitboxbasestatus.StatusInitialized)
+	if err != nil {
+		base.log.WithError(err)
```
benma: I don't think this produces a logline. Probably needs `.Error("...")`.

backend/bitboxbase/bitboxbase.go
```diff
@@ -228,44 +263,220 @@ func (base *BitBoxBase) RPCClient() *rpcclient.RPCClient {
...
+		// if the rpc client throws an error we don't want to pass that on to the frontend, because its probably junk
```
benma: The frontend needs to know there was an error anyway, and not pretend it worked.

It is correct to return an error and let the handler or the frontend deal with it.

backend/bitboxbase/bitboxbase.go
```diff
@@ -228,44 +263,220 @@ func (base *BitBoxBase) RPCClient() *rpcclient.RPCClient {
...
+		return nil
+	}
+	if !reply.Success {
+		return errp.Newf("ReindexBitcoin was not successful. Message: '%s'. Code: '%s'", reply.Message, reply.Code)
```
benma: if you slap an `Error() string` method on reply.Error, you can return that as a normal Go error, so the handler can inspect the error and relay the error code and message.

Like this, the frontend has no access to the error code.

backend/bitboxbase/bitboxbase.go
0xB10C: fyi: I just refactored the `if !base.active {}` part of `Deregister()`. This was in the code base already. 

But can do.

0xB10C:
Commits by @Tomasvrba and myself should address the review.

0xB10C:
Rebased onto master.

PR #507, opened 2019-09-06
frontend: upgrade preact to 8.5.2, fix TypeScript

KasparEtter:
As far as I can judge, the preact update exposed problems that we already had before but which TypeScript didn't detect because it could pick the wrong method declaration and basically opt out of type checking. I fixed the issues but
- @originalblend337: Please review and test the change in `app.tsx` to the guide toggler (the behavior should have been broken before and I haven't tested the effect of my fix).
- @benma: Please check and test the change in `send.tsx`. The problem was more or less that `boolean | undefined` could not be assigned to `boolean` so I made `noMobileChannelError` optional as well.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -91,7 +91,7 @@ interface State {
     amountError?: string;
     dataError?: string;
     paired?: boolean;
-    noMobileChannelError: boolean;
+    noMobileChannelError?: boolean;
     signProgress?: SignProgress;
     // show visual BitBox in dialog when instructed to sign.
     // can't be undefined because of the default touchConfirm param in the wait dialog.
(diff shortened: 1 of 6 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -91,7 +91,7 @@ interface State {
     amountError?: string;
     dataError?: string;
     paired?: boolean;
-    noMobileChannelError: boolean;
+    noMobileChannelError?: boolean;
```
benma: Why is it necessary to make it optional? Easier to keep like before, but fix the assignment to be `false` instead of `undefined`.

reason: to states for the var are sufficient, don't need three states where two of them mean the same.

KasparEtter: Feel free to do so. If I remember correctly, the `undefined` came from the `account` in https://github.com/digitalbitbox/bitbox-wallet-app/blob/16d6924663ce9088f55677aa8ab848c6f468925f/frontends/web/src/routes/account/send/send.tsx#L146

And for consistency, you could/should then also change this for `paired` and other fields. 🤓

benma: :man_shrugging: it's your PR.

PR #509, opened 2019-09-09
Ui

originalblend337:


frontends/web/src/components/transactions/transaction.jsx
```diff
@@ -71,108 +55,171 @@ export default class Transaction extends Component {
...
-                                    <span>{sDate}</span>
-                                </div>
-                                <div class={[style.address, style.multiline].join(' ')}>{addresses.join(', ')}</div>
-                            </div>
+            <div className={[style.container, collapsed ? style.collapsed : '', index === 0 ? style.first : ''].join(' ')}>
+                <div className={[parentStyle.columns, style.row].join(' ')}>
+                    <div className={parentStyle.columnGroup}>
+                        <div className={parentStyle.type}>
+                            {
+                                badge === 'In' ? (
```
benma: `badge` is a translated text! Use `type` instead (can be `receive`, `send` or `send_to_self`).

Looks like the badges have been broken in other languages than English for a while now :/

frontends/web/src/routes/account/receive/receive.css
```diff
@@ -91,6 +90,10 @@ pre.verifyAddress {
     color: var(--color-brightgray);
 }
 
+.arrow {
+
```
benma: delete unused?

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -136,12 +145,20 @@ class BitBox02 extends Component<Props, State> {
 
     private onChannelHashChanged = () => {
         apiGet(this.apiPrefix() + '/channel-hash').then(({ hash, deviceVerified }) => {
+            if (!this.state.deviceVerified && deviceVerified) {
+                this.channelVerify(true);
```
benma: Automatic confirm?! Has to be manual.

frontends/web/src/components/wait-dialog/wait-dialog.jsx
```diff
@@ -61,6 +61,7 @@ export default class WaitDialog extends Component {
         active,
     }) {
         const isActive = active ? style.active : '';
+        const titleText = title || 'Action needed on your BitBox02';
```
benma: Default title is not good. Now we have 'Action needed on your BitBox02' everywhere, for example the success message when sending a transaction using BitBox01.

frontends/web/src/components/devices/bitbox02/backups.tsx
```diff
@@ -101,67 +102,70 @@ class Backups extends Component<Props, State> {
...
+                                    <p>The following backups have been found. Please select which backup to restore.</p>
+                                    <div className={backupStyle.listContainer}>
+                                        {
+                                            backups.backups!.map(backup => (
+                                                <table className={style.table}>
+                                                    <BackupsListItem
+                                                        key={backup.id}
+                                                        disabled={restoring}
+                                                        backup={backup}
+                                                        selectedBackup={selectedBackup}
```
TheCharlatan: Do we still need to select backups? Check Backup go through the entire list of backups and checks if at least one of them matches the current wallet. Create Backup just creates a new one from the current wallet, so you do not need to select either.

benma: i think this might be reused with when you restore from backup, where you need to select one.

PR #512, opened 2019-09-10
Frontend: Small manage device bitbox02 text fixes

TheCharlatan:
Change the guide text for the bitbox02 reset confirmation. 
Explain that a BIP39 mnemonic passphrase can be used to 'encrypt' the wallet (not sure if encrypt is the correct terminology, but in my opinion we should encourage using it at least a bit).

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -134,7 +134,7 @@
   "bitbox02Settings": {
     "deviceName": {
       "input": "BitBox02 Name",
-      "title": "Set BitBox 02 name"
+      "title": "Set BitBox02 name"
     },
     "mnemonicPassphrase": {
       "description": "This allows you to enter a mnemonic passphrase after unlocking your BitBox02.\n\n<strong>NOTE:</strong> any mnemonic passphrase derives a new separate wallet.\nKeep your mnemonic passphrase <strong>secure</strong>.\n<strong>A loss of a mnemonic passphrase is NOT RECOVERABLE.</strong>",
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -134,7 +134,7 @@
   "bitbox02Settings": {
     "deviceName": {
       "input": "BitBox02 Name",
-      "title": "Set BitBox 02 name"
+      "title": "Set BitBox02 name"
```
benma: Changing values needs to be done inside locize, see https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/docs/i18n.md

I went in there and fixed this one. You can remove this change from this PR.

frontends/web/src/locales/en/app.json
```diff
@@ -718,6 +718,7 @@
     }
   },
   "reset": {
+    "bb02Understand": "I have a backup on my SD card or have the BIP39 mnemonic written down",
```
benma: I would do "I have backup" only, as that covers both. Non-experts won't be confused by the cryptic `BIP39`, while experts know what's up ;)

frontends/web/src/locales/en/app.json
```diff
@@ -441,7 +441,7 @@
         "title": "What is 'Check Backup'?"
       },
       "encrypt": {
-        "text": "No. Please keep the micro SD card safe, as it contains the unencrypted seed to recover your wallet.",
+        "text": "No. Please keep the micro SD card safe, as it contains the unencrypted seed to recover your wallet. If you wish to encrypt your seed you can enable a mnemonic passphrase in the ManageDevice's expert settings",
```
benma: same, i'll add it do locize directly. pls remove here

TheCharlatan:
Closing have been implemented within other pull requests.

PR #544, opened 2019-09-23
frontend: improve upgrade firmware screen UI

Tomasvrba:
- Improves the update firmware screen with our new toggle button.
- Makes the content box fixed size so it doesn't jump with the added checkbox div (which also moved the Advanced Settings link, making the user unable to open/close without moving cursor)
- Center and add margin to "Device oriented the wrong way"

Before:
Screencast2019-09-23131651

After:
Screencast2019-09-23131508

### DIFF
--- a/frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.tsx
+++ b/frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.tsx
@@ -42,7 +42,6 @@ interface State {
         upgradeSuccessful: boolean;
         rebootSeconds: number;
     };
-    advancedSettings: boolean;
 }
 
 class BitBox02Bootloader extends Component<Props, State> {
@@ -56,7 +55,6 @@ class BitBox02Bootloader extends Component<Props, State> {
                 upgradeSuccessful: false,
                 rebootSeconds: 0,
             },
-            advancedSettings: false,
         };
     }
 
@@ -108,7 +106,6 @@ class BitBox02Bootloader extends Component<Props, State> {
           erased,
         }: RenderableProps<Props>,
         { status,
-          advancedSettings,
         }: State,
     ) {
         let contents;
@@ -132,7 +129,7 @@ class BitBox02Bootloader extends Component<Props, State> {
             }
         } else {
             contents = (
-                <div className="box large">
+                <div className="box large" style="min-height: 390px">
                     <div className="buttons">
                         <Button
                             primary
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.tsx
```diff
@@ -156,13 +161,15 @@ class BitBox02Bootloader extends Component<Props, State> {
...
                     <hr/>
-                    <div>
-                        <a
-                            onClick={() => { this.setState({ advancedSettings: !advancedSettings }); }}
-                            style="text-decoration: underline; cursor: pointer;" >
-                            { advancedSettings ? '-' : '+' } Advanced Settings
-                        </a>
-                    </div>
+                    <div className="box slim divide">
+                        <div class="flex flex-row flex-between flex-center">
```
originalblend337: this should be `flex flex-row flex-between flex-items-center`

Tomasvrba: Thanks Jeff, fixed.

benma:
Hehe why is expanding/collapsing the new style toggler but the actual checkmark isn't? 

Not sure if the toggler applies well to expanding, as nothing is 'activated', not permanently at least.

@originalblend337 ?

Ok to merge anyway if you give the thumbs up.

frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.tsx
```diff
@@ -156,13 +161,15 @@ class BitBox02Bootloader extends Component<Props, State> {
...
-                        <a
-                            onClick={() => { this.setState({ advancedSettings: !advancedSettings }); }}
-                            style="text-decoration: underline; cursor: pointer;" >
-                            { advancedSettings ? '-' : '+' } Advanced Settings
-                        </a>
-                    </div>
+                    <div className="box slim divide">
+                        <div class="flex flex-row flex-between flex-items-center">
+                            <p className="m-none">{t('bb02Bootloader.advanced.label')}</p>
+                            <Toggle
```
benma: Why not `<Toggle .../>`?

Tomasvrba: Sure.
Because VS Code just auto-completes the closing tag... :)

Tomasvrba:
> Hehe why is expanding/collapsing the new style toggler but the actual checkmark isn't?
> 
> Not sure if the toggler applies well to expanding, as nothing is 'activated', not permanently at least.

Good point. Yeah, I agree that having an expand arrow or something and then the toggle checkbox for the for the fw hash would be nicer

Tomasvrba:
@benma and @originalblend337 
Latest fixup should address all issues mentioned above:

Screencast2019-09-23153644

benma:
The screencasts are awesome!

PR #554, opened 2019-09-24
User agreement

Tomasvrba:


frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -577,19 +577,42 @@ class BitBox02 extends Component<Props, State> {
                             {
                                 (!unlockOnly && appStatus === 'createWallet') && (
                                     <Step
+                                        width={700}
```
benma: mobile friendly?

@originalblend337 should take a look here as he dealt with card widths today.

PR #556, opened 2019-09-24
frontend: revert toast to bottom center and adjust for status height

Tomasvrba:


originalblend337:
If you can, can you make an adjustment so it can be positioned in its original position, or the center bottom? Going forward the top right might be a better position for generic messages

Tomasvrba:
> If you can, can you make an adjustment via adding a property so it can be positioned in its original position, or the center bottom? Going forward the top right might be a better position for generic messages

By original position you mean original at the bottom or the previous one upper right? 
Agreed that for generic messages upper right would work better. It just didn't feel right during the wizard when everything else is centered.

Since I'm not exactly sure what you have in mind, maybe you can take a look at it later?

I'd say it's not high-prio though :)

PR #571, opened 2019-09-25
Add higher res icons

### LINKED ISSUE #25: AppImage missing icons file
To package BitBox wallet app for Arch Linux AUR from the .AppImage, I extract the bitbox.desktop and bitbox(.png, .svg, etc) to include them in the current desktop environment. But the .AppImage does not include any image files.

`$ ./BitBox-4.1.0-x86_64.AppImage --appimage-extract`
`$ ls -la`
`drwx------  8 pizzaman users     4096 Aug 21 14:20 .`
`drwxr-xr-x  5 pizzaman users     4096 Aug 21 14:20 ..`
`lrwxrwxrwx  1 pizzaman users        6 Aug 21 14:20 AppRun -> BitBox`
`-rw-r--r--  1 pizzaman users   159474 Aug 21 14:20 assets.rcc`
`-rwxr-xr-x  1 pizzaman users    51268 Aug 21 14:20 BitBox`
`-rw-r--r--  1 pizzaman users      155 Aug 21 14:20 bitbox.desktop`
`-rw-r--r--  1 pizzaman users        0 Aug 21 14:20 BitBox-x86_64.AppImage`
`-rw-r--r--  1 pizzaman users        0 Aug 21 14:20 default.png`
`lrwxrwxrwx  1 pizzaman users       11 Aug 21 14:20 .DirIcon -> default.png`
`drwx------ 58 pizzaman users     4096 Aug 21 14:20 doc`
`drwx------  2 pizzaman users     4096 Aug 21 14:20 lib`
`drwx------  2 pizzaman users     4096 Aug 21 14:20 libexec`
`-rw-r--r--  1 pizzaman users 11645848 Aug 21 14:20 libserver.so`
`drwx------  9 pizzaman users     4096 Aug 21 14:20 plugins`
`-rw-r--r--  1 pizzaman users      145 Aug 21 14:20 qt.conf`
`drwx------  2 pizzaman users     4096 Aug 21 14:20 resources`
`drwx------  3 pizzaman users     4096 Aug 21 14:20 translations`

The only image file is default.png with 0 bytes.

ref. https://wiki.archlinux.org/index.php/desktop_entries#Icons

NickeZ:
* Added some more icons
* Embed icon in AppImage
* Use `QApplication::setWindowIcon()` so that icons work for uninstalled app as well (like AppImage)

Fixes #25


frontends/qt/resources/linux/Makefile
```diff
@@ -0,0 +1,9 @@
+SIZES=16x16 32x32 48x48 64x64 128x128
+
+TARGETS=$(addsuffix /apps/bitbox.png, $(addprefix usr/share/icons/hicolor/, ${SIZES}))
```
benma: magic! :100:

benma:
Is the file in /usr/share/pixmaps still needed?

NickeZ:
> Is the file in /usr/share/pixmaps still needed?

maybe older implementations of freedesktop desktop environments still use it? Works without it on my ubuntu.

frontends/qt/Makefile
```diff
@@ -36,7 +36,8 @@ linux:
 	cd build/linux && fpm --after-install ../../resources/deb-afterinstall.sh -s dir -t deb -n bitbox -v 4.13.0 -C ../tmp-deb/
 	cd build/linux && fpm --after-install ../../resources/deb-afterinstall.sh -s dir -t rpm -n bitbox -v 4.13.0 -C ../tmp-deb/
 	# create AppImage
-	sed 's+/usr/share/pixmaps/bitbox.png+default+' resources/linux/usr/share/applications/bitbox.desktop > build/linux-tmp/bitbox.desktop
+	cp resources/linux/usr/share/applications/bitbox.desktop build/linux-tmp
+	cp resources/linux/usr/share/icons/hicolor/128x128/apps/bitbox.png build/linux-tmp
```
benma: Tested and looks great for the AppImage.

The .deb/.rpm packages do not contain the bitbox.png though for the Qt window logo. You should move this line to before the `fpm` calls above.

benma:
> maybe older implementations of freedesktop desktop environments still use it? Works without it on my ubuntu.

Let's remove it then, I added it like this with the absolute path in the .desktop file without any special considerations.

frontends/qt/Makefile
benma: sorry not move, but have the same for `../tmp-deb`

frontends/qt/main.cpp
```diff
@@ -106,6 +106,7 @@ int main(int argc, char *argv[])
     a.setApplicationName(APPNAME);
     a.setOrganizationDomain("shiftcrypto.ch");
     a.setOrganizationName("Shift Cryptosecurity");
+    a.setWindowIcon(QIcon(QCoreApplication::applicationDirPath() + "/bitbox.png"));
```
benma: will have to test if this is necessary on Windows/macOS. In any case, like this it might be a problem on those platforms because this file does not exist (you'd need to copy the file in the Makefile also for macOS and windows).

NickeZ: Yeah I was afraid about that. Tried to compile it on windows, but, oh man, I will have to rewrite the whole build system before I can successfully do that :P. Could I bundle it into the assets file?

frontends/qt/Makefile
NickeZ: fixed

frontends/qt/resources/linux/Makefile
NickeZ: :)

frontends/qt/main.cpp
benma: Well if you have all the deps installed on windows, `make qt-windows` should work (see AppVeyor ;)). But yeah, Windows is the most painful to build.

assets file: that will probably work and is a good idea!

PR #580, opened 2019-10-03
frontend: ui issues

originalblend337:
- remove redundant key handlers from components where the dialog component already handles keydown events.
- update animation for revealing/hiding dialogs for smoother experience
- update hover colors for primary and secondary buttons to distinguish between hover / disabled state
- display scrollbars for scrollable content

### DIFF
--- a/frontends/web/src/components/devices/bitbox02/setdevicename.jsx
+++ b/frontends/web/src/components/devices/bitbox02/setdevicename.jsx
@@ -87,10 +87,8 @@ export class SetDeviceName extends Component {
                             <div className="columnsContainer half">
                                 <div className="columns half">
                                     <div className="column">
-                                        <Input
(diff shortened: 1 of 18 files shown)


frontends/web/src/components/devices/bitbox02/setdevicename.jsx
```diff
@@ -87,10 +87,8 @@ export class SetDeviceName extends Component {
                             <div className="columnsContainer half">
                                 <div className="columns half">
                                     <div className="column">
-                                        <Input
-                                            label="Current Device Name"
-                                            disabled={true}
-                                            value={name} />
+                                        <label>Current Device Name</label>
```
benma: please make a habit of putting new text in `t('...')`, as it's harder to find all the un-locized things after

PR #585, opened 2019-10-04
frontend: headers sync positionning fix

originalblend337:


frontends/web/src/components/headerssync/headerssync.jsx
```diff
@@ -42,11 +42,11 @@ export default class HeadersSync extends Component {
         t,
         status,
     }, {
-        show,
+        // show,
     }) {
-        if (!status || !show) {
-            return null;
-        }
+        // if (!status || !show) {
```
benma: uuh what's with this? :sweat_smile:

PR #586, opened 2019-10-04
Bbbase setup wizard

Tomasvrba:


Tomasvrba:
@benma and @originalblend337 since I'm not around next week, feel free to merge this whenever

PR #592, opened 2019-10-08
frontend/eth: warning to only receive only the current coin

benma:
In the receive tab, when verifying. Reason: we currently do not
support all erc-20 tokens. This guides the user to activate the token
they want in the settings.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -719,6 +719,10 @@
       "info": "Litecoin is transitioning from addresses starting with '3' to addresses starting with an 'M'.",
       "result": "Legacy format of {{address}}:\n{{legacyAddress}}"
     },
+    "onlyThisCoin": {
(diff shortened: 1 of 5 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -719,6 +719,10 @@
       "info": "Litecoin is transitioning from addresses starting with '3' to addresses starting with an 'M'.",
       "result": "Legacy format of {{address}}:\n{{legacyAddress}}"
     },
+    "onlyThisCoin": {
+      "description": "To receive other tokens, enable them in the settings.",
+      "warning": "Make sure to only receive {{accountName}} on this address."
```
jstrnbrg: I would maybe make this even more clear, like: "Make sure to only receive {{accountName}} on this address. If you deposit other tokens they might not be accessible until we add that token." ?  

I know they are accessible by exporting the seed to MEW but we should avoid anyone from having to do that..

benma: Adjusted the text now, pls check.

PR #606, opened 2019-10-16
GetServiceInfo RPC

0xB10C:
This commit adds the GetServiceInfo RPC and handler to the App.

The RPC provides the App with information about the services running on the BitBoxBase. The middleware notifies the App when the ServiceInfo change. The event is implement here as well, but not used yet.


Tomasvrba:
rebased on master before merging

PR #626, opened 2019-10-30
frontend: update green color and thicken tx arrows

Tomasvrba:


Tomasvrba:
Closing this PR as we will get rid of the arrow column.

PR #634, opened 2019-11-03
build: convert dependency management to Go modules

benma:
We vendor using `go mod vendor` followed by copying .h and .c files
because the karalable/usb and go-ethereum secp256k1 crypto modules
keep the cgo included files in subfolders, which do not get vendored
using `go mod vendor` normally. `make go-vendor` added to streamline
this until fixed.

We also vendor because `gomobile bind` has issues with Go modules (in
our case, the build hangs forever, but others have reported other
issues).

Lastly, I prefer vendor so that the project can be built offline as
well as if the upstream deps or cache proxies vanish.


benma:
I checked the build manually on linux, macOS and Win. Seems to work.

Makefile
```diff
@@ -18,13 +18,13 @@ WEBROOT:=`pwd`/frontends/web
...
-	go get golang.org/x/tools/cmd/goimports
-	go get -u github.com/jteeuwen/go-bindata/...
-	go get -u golang.org/x/mobile/cmd/gomobile
+	GO111MODULE=off ./scripts/go-get.sh v1.19.1 github.com/golangci/golangci-lint/cmd/golangci-lint
+	GO111MODULE=off go get -u github.com/stretchr/testify # needed for mockery
+	GO111MODULE=off go get -u github.com/vektra/mockery/cmd/mockery
+	GO111MODULE=off go get -u github.com/goware/modvendor
+	GO111MODULE=off go get golang.org/x/tools/cmd/goimports
+	GO111MODULE=off go get -u github.com/jteeuwen/go-bindata/...
+	GO111MODULE=off go get -u golang.org/x/mobile/cmd/gomobile
```
x1ddos: I believe `GO111MODULE` will be removed at some point. Maybe find a more stable solution?
From https://golang.org/cmd/go/#hdr-Module_support -

> ... Go 1.13 continues to respect a *temporary* environment variable, GO111MODULE, ...

README.md
```diff
@@ -113,9 +113,15 @@ Run `make servewallet` to compile the code and run `servewallet`. `servewallet`
...
+#### Go dependencies
 
-Run `dep ensure` to update dependencies.
+Go dependencies are managed by `go mod`, and vendored using `make go-vendor`. The deps are vendored
+so that
+
+- offline builds work
+- dependency diffs are easier to inspect
+- less reliance on remote systems
+- because `gomobile bind` does not support Go modules yet
```
x1ddos: Would be nice to link to their issue on github or elsewhere imho, if there is one.

x1ddos:
>  +9,902 −1,404,935

nice deletions counter :)

Makefile
benma: do you happen to know how to handle build/dev tools? where do they land if I `go get` them with go modules enabled?

README.md
benma: None seems to match our issue exactly, but this indicates that gomobile is not ready yet for go modules: https://github.com/golang/go/issues/27234

Makefile
x1ddos: Confusing, right! I've been searching for some official docs on it but haven't found. At the moment, I believe it still works as before: install into `$GOPATH/bin` where `GOPATH` defaults to `$HOME/go` if unset. Unlike `GO111MODULE`,`GOPATH` will stay for a much longer time. This works:

```sh
$ go version
go version go1.13.3 linux/amd64
$ go env GO111MODULE
on

$ mkdir /tmp/go && cd /tmp/go
$ go install golang.org/x/lint/golint
$ file bin/golint
bin/golint: ELF 64-bit LSB executable
```

benma: `go install` or `go get`? Never used `go install` directly on a remote package.

x1ddos: Ah, `go install` is just part of `go get`. From go help get:

> Get resolves and adds dependencies to the current development module
and then builds and installs them.
> With no package arguments and also without -u, 'go get' is not much more
than 'go install', and 'go get -d' not much more than 'go list'.

Essentially, for `main` packages, the "install" word above literally means `go install`. So, `go get -u golang.org/x/lint/golint` would've resulted in the same binary location. I just didn't want to re-fetch everything.

While re-reading the docs, it turns out that:

```sh
$ go help install
...
Executables are installed in the directory named by the GOBIN environment
variable, which defaults to $GOPATH/bin or $HOME/go/bin if the GOPATH
environment variable is not set.
```

So, I guess the `GO111MODULE=off` has no effect on where executables are installed.

Oh, and I've just noticed you're still on go1.13.1 according to https://github.com/digitalbitbox/bitbox-wallet-app/blob/579968c66846bb86cbc0d845502b8374780a34fc/scripts/docker_install.sh#L42

I'd recommend using the latest 1.13.4 and updating that dockerfile more frequently. There were some security bugfixes including crypto/x509 and crypto/ecdsa packages in 1.13.2 and .3 like this one: https://groups.google.com/d/topic/golang-announce/lVEm7llp0w0.

benma: > I'd recommend using the latest 1.13.4 and updating that dockerfile more frequently.

:+1:

benma: Done: https://github.com/digitalbitbox/bitbox-wallet-app/pull/644

PR #646, opened 2019-11-05
Base node management

Tomasvrba:


Tomasvrba:
Keeping this on the staging branch for now.

PR #654, opened 2019-11-08
base-settings: add basic Reboot rpc to the dashboard frontend

Tomasvrba:
Calls the RebootBase RPC after user confirmation.
Currently there is no automatic reconnect as the base gets
deregistered once the websocket connection shuts down. Needs
additional backend functionality to keep trying to reconnect after restart.

### DIFF
--- a/backend/bitboxbase/bitboxbase.go
+++ b/backend/bitboxbase/bitboxbase.go
@@ -608,6 +608,10 @@ func (base *BitBoxBase) RebootBase() error {
 	if !reply.Success {
 		return &reply
 	}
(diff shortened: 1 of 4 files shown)


TheCharlatan:
We should eventually use a configuration entry for connected bases. Typing in the ip every time would be annoying.

Tomasvrba:
> We should eventually use a configuration entry for connected bases. Typing in the ip every time would be annoying.

Yeah, and also only deregister in the backend maybe after a few hours and not immediately, and have automatic reconnect instead, when the Base comes back online

PR #655, opened 2019-11-08
Base updates

Tomasvrba:


backend/bitboxbase/handlers/handlers.go
```diff
@@ -462,3 +464,16 @@ func (handlers *Handlers) getServiceInfoHandler(_ *http.Request) (interface{}, e
 		"serviceInfo": baseInfo,
 	}, nil
 }
+
+func (handlers *Handlers) getUpdateInfoHanlder(_ *http.Request) (interface{}, error) {
```
TheCharlatan: Hanlder -> Handler

Tomasvrba: Good catch, darn typos. Sometimes better to copy-paste :D

PR #660, opened 2019-11-09
backend: add Close()

benma:
Call it after bridgecommon.Shutdown(). Otherwise running Serve() again
can lead to undesired behavior, like opening a database twice and
hanging.

This commit is a partial solution, as we need to close all resources.


x1ddos:
Not sure travis-ci actually picked up "go test -v" because https://travis-ci.org/digitalbitbox/bitbox-wallet-app/jobs/609596768 doesn't look like verbose test run.

PR #662, opened 2019-11-09
backend/coins/btc/headers: add Close()

benma:
Shutdown goroutine and close db.

### DIFF
--- a/backend/coins/btc/blockchain/mocks/mock.go
+++ b/backend/coins/btc/blockchain/mocks/mock.go
@@ -1,0 +1,125 @@
+// Copyright 2018 Shift Devices AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


backend/coins/btc/blockchain/mocks/mock.go
```diff
@@ -0,0 +1,112 @@
+package mocks
```
x1ddos: License headers and pkg doc comment?

backend/coins/btc/blockchain/mocks/mock.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	chainhash "github.com/btcsuite/btcd/chaincfg/chainhash"
+	wire "github.com/btcsuite/btcd/wire"
+	btcutil "github.com/btcsuite/btcutil"
+	blockchain "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/blockchain"
+)
+
+// BlockchainMock implements Interface for use in tests.
```
x1ddos: I'd say "implements blockchain.Interface". It was clear which interface as long as this BlockchainMock struct was in the same blockchain package, but it's split into "mocks" now.

backend/coins/btc/blockchain/mocks/mock.go
benma: Done

backend/coins/btc/blockchain/mocks/mock.go
benma: correct, done

PR #663, opened 2019-11-09
android: replace default app icon with bb app logo

x1ddos:
Took the icon from qt/resources/app_icon_source.png, resized to
192x192, 144x144, 96x96, 72x72, 48x48 and fed to optipng.

More recent Android versions use the [adaptive icon](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive) from anydpi-v26.
Had to delete it so that they all pick up the original PNG images.
The adaptive icon is nice but it is in SVG and would require to
somehow covert the PNG icon into SVG. Not easy and time consuming.
The original PNG icons feel sufficient to me for now.

A couple screenshots:

Screenshot_20191109-230850

Screenshot_20191109-230818


benma:
Awesome.

Please add a description in the commit message how exactly you added the icon to the android app. There are tons of xml files and different png files in different locations, it is unclear how it works.

Consider making a script that streamlines this, like:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/qt/resources/win/convert.sh
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/qt/resources/MacOS/convert_icon.sh
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/qt/resources/linux/Makefile

Otherwise the next guy to change the icon has to research it again.

x1ddos:
PTAL. Added the script.
Didn't create a make target because the script's output is nondeterministic which will make it run on each build. Besides the icon changes very infrequently.

benma:
> PTAL. Added the script.
> Didn't create a make target because the script's output is nondeterministic which will make it run on each build. Besides the icon changes very infrequently.

Awesome, thanks! Yeah it's meant mostly as "formalized" documentation, the other icon scripts are also meant to be run manually. I added them as it took me a while to figure out what needed to be done.

I will test this soon.

x1ddos:
SG. Will add it to the makefile then.

PR #667, opened 2019-11-09
Closing

benma:


benma:
@x1ddos this works for me (open, close, open) even though not everything is shut down yet.

x1ddos:
Nice. And I see the TestServeShutdownServe passes in this CI run, which now includes all those closing calls.

PR #674, opened 2019-11-11
frontend: display hamburger menu on spinner

originalblend337:
- display toggler in spinner
- add label to account info icon in accounts


benma:
Info button looks good. Next time please make separate PRs for separate things, but most certainly separate commits, so if it's reverted for one reason, the other feature does not get lost.

About the spinner: there is weird interaction with the guide. Have the guide open somewhere, then activate the spinner:

- in large window size, the guide is not shown
- in small window size. the guide is shown, graying the whole spinner area, including the burger icon
- you can close the guide by clicking anywhere on the spinner (which is not the case for the guide elsewhere, is it?), then the guide vanishes completely, no way to get it back

Imho ideally, the guide icon would also be visible in the spinner, and the opened/closed setting should be honored normally.

general question: on my mobile phone, the guide covers the full screen, which is nice. can't reproduce this on desktop by making the window smaller. is there a special mobile phone css rule, or is my phone screen just that small? :P

benma:
@originalblend337 any progress?

PR #675, opened 2019-11-11
Android rename

benma:


frontends/android/BitBoxApp/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">BitBoxApp</string>
```
x1ddos: Does this mean it'll show up as "BitBoxApp" in the android launcher, the home screen? It feels a bit like tautology: what isn't an app on android? :)

I get the idea of renaming the application ID but maybe leave the display name as is, just BitBox? I mean, you're not going to have another BitBoxSomething, right. Besides, you could always change this display name without consequences to the internal application ID, invisible to the users.

x1ddos:
Oh, forgot to mention another file: https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/.dockerignore

frontends/android/BitBoxApp/app/src/main/res/values/strings.xml
benma: I know, but this is the branding / official name of the app, even on mobile ;) `BitBox` is overloaded and with multiple apps in the playstore, this alone can be confusing too.

I'll discuss the naming question at Shift, in the meantime the name stays :smile:

benma:
Nice catches. Forced replaced the first commit fixing all of those. Thanks.

frontends/android/BitBoxApp/app/src/main/res/values/strings.xml
x1ddos: I get it. It's like WhatsApp :rofl:

PR #677, opened 2019-11-11
frontend: lazy load qrreader

thisconnect:
This should make the JavaScript main bundle.js smaller and postpone loading the zxing library until you navigate to the Send route.

Screen Shot 2019-11-11 at 11 47 25 AM

To create a chart use `cd frontends/web && yarn build --json` that will create a `stats.json`, you can drop that into https://chrisbateman.github.io/webpack-visualizer/

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -15,9 +15,9 @@
  * limitations under the License.
  */
 
-import { BrowserQRCodeReader } from '@zxing/library';
 import { Component, h, RenderableProps } from 'preact';
 import { route } from 'preact-router';
+import { BrowserQRCodeReader } from '@zxing/library';
 import * as accountApi from '../../../api/account';
 import { IAccount } from '../account';
 import reject from '../../../assets/icons/cancel.svg';
@@ -100,7 +100,7 @@ class Send extends Component<Props, State> {
     private utxos!: Component<UTXOsProps>;
     private selectedUTXOs: SelectedUTXO = {};
     private unsubscribe!: () => void;
-    private qrCodeReader: BrowserQRCodeReader = new BrowserQRCodeReader();
+    private qrCodeReader?: BrowserQRCodeReader;
 
     // pendingProposals keeps all requests that have been made
     // to /tx-proposal in case there are multiple parallel requests
@@ -174,11 +174,18 @@ class Send extends Component<Props, State> {
         this.registerEvents();
         const account = this.getAccount();
         if (account && !account.coinCode.startsWith('eth-erc20-') && account.coinCode !== 'eth') {
-            this.qrCodeReader
-                .getVideoInputDevices()
-                .then(videoInputDevices => {
-                    this.setState({ hasCamera: videoInputDevices.length > 0 });
-                });
+            import('../../../components/qrcode/qrreader')
+                .then(({ BrowserQRCodeReader }) => {
+                    if (!this.qrCodeReader) {
+                        this.qrCodeReader = new BrowserQRCodeReader();
+                    }
+                    this.qrCodeReader
+                        .getVideoInputDevices()
+                        .then(videoInputDevices => {
+                            this.setState({ hasCamera: videoInputDevices.length > 0 });
+                        });
+                })
+                .catch(console.error);
         }
     }
 
@@ -187,7 +194,9 @@ class Send extends Component<Props, State> {
         if (this.unsubscribe) {
             this.unsubscribe();
         }
-        this.qrCodeReader.reset();
+        if (this.qrCodeReader) {
+            this.qrCodeReader.reset();
+        }
     }
 
     private registerEvents = () => {
@@ -492,8 +501,10 @@ class Send extends Component<Props, State> {
 
     private toggleScanQR = () => {
         if (this.state.activeScanQR) {
-            // release camera; invokes the catch function below.
-            this.qrCodeReader.reset();
+            if (this.qrCodeReader) {
+                // release camera; invokes the catch function below.
+                this.qrCodeReader.reset();
+            }
             // should already be false, set by the catch function below. we do it again anyway, in
             // case it is not called consistently on each platform.
             this.setState({ activeScanQR: false });
@@ -503,12 +514,14 @@ class Send extends Component<Props, State> {
             activeScanQR: true,
             videoLoading: true,
         }, () => {
-            this.qrCodeReader
+            this.qrCodeReader && this.qrCodeReader
                 .decodeFromInputVideoDevice(undefined, 'video')
                 .then(result => {
                     this.setState({ activeScanQR: false });
                     this.parseQRResult(result.getText());
-                    this.qrCodeReader.reset(); // release camera
+                    if (this.qrCodeReader) {
+                        this.qrCodeReader.reset(); // release camera
+                    }
                 })
                 .catch(() => {
                     this.setState({ activeScanQR: false });
--- a/frontends/web/package.json
+++ b/frontends/web/package.json
@@ -28,7 +28,7 @@
     "typed-css-modules": "0.3.5"
   },
   "dependencies": {
-    "@zxing/library": "0.11.0",
+    "@zxing/library": "0.18.3",
     "i18next": "11.3.6",
     "i18next-locize-backend": "4.1.9",
     "lightweight-charts": "3.1.5",
--- a/frontends/web/src/components/qrcode/qrreader.ts
+++ b/frontends/web/src/components/qrcode/qrreader.ts
@@ -1,0 +1,21 @@
+/**
+ * Copyright 2021 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { BrowserQRCodeReader } from '@zxing/library';
+
+export {
+    BrowserQRCodeReader
+};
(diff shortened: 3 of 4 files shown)


thisconnect:
Unable to find image 'shiftcrypto/bitbox-wallet-app:3' locally
🤔 
https://travis-ci.org/digitalbitbox/bitbox-wallet-app/jobs/610285028#L503

benma:
> Unable to find image 'shiftcrypto/bitbox-wallet-app:3' locally

That is normal, travis does not cache things like this, it pulls the image every time.

CI failure comes from:

> ERROR: src/components/qrcode/qrreader.ts[21, 2]: file should end with a newline
> ERROR: src/components/qrcode/qrreader.ts[21, 2]: Missing semicolon

benma:
> This should make the JavaScript main bundle.js smaller

Hmm, the qr reader code still needs to be shipped, so if it's not part of bundle.js, it's in another js file that needs to be shipped, right?

Does the initial load speed change noticeably after this change?

thisconnect:
> ERROR: src/components/qrcode/qrreader.ts[21, 2]: file should end with a newline
> ERROR: src/components/qrcode/qrreader.ts[21, 2]: Missing semicolon

I didn't see that, also locally visual code didn't complain :(, Iwill fix it.

> Hmm, the qr reader code still needs to be shipped, so if it's not part of bundle.js, it's in another js file that needs to be shipped, right?

correct, but it is loaded when needed and not part of the initially loaded JS.

> Does the initial load speed change noticeably after this change?

Well that is just the first 🐉 next I'd like to look at bundled languages.

And I have not measured. Not sure what the best way is, maybe performance API https://developer.mozilla.org/en-US/docs/Web/API/Performance in the QT app... unless you oppose this approach or have a better idea.

thisconnect:
force pushed

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -104,7 +103,7 @@ class Send extends Component<Props, State> {
     private utxos!: Component<UTXOsProps>;
     private selectedUTXOs: SelectedUTXO = {};
     private unsubscribe!: () => void;
-    private qrCodeReader: BrowserQRCodeReader = new BrowserQRCodeReader();
+    private qrCodeReader: any;
```
thisconnect: @benma @KasparEtter FYI I didn't know how to keep the type here :(
but then this is only a POC of how to load bigger libraries

benma:
@thisconnect what do you think about this today? In what ways would lazy loading help the app? Is it measurable?

thisconnect:
> @thisconnect what do you think about this today? In what ways would lazy loading help the app? Is it measurable?

Users only need to ever have that library loaded if they scan a qr code.

It cannot be measured in `webdev` mode as the bundle is huge and I believe contians inline sourcemaps.

But it should be measurable by serving a production build via webbrowser. I can test this week.

Also I can change it to not use async/await anymore, only a simple `import()` promise. We may have to check that the bundled library  is available in the app when the frontend requires it. I believe there was a allow-list in go that provides the built frontend files. 

Also the library has never versions that we should consider upgrading to.

benma:
So the expectation is that the app loads faster if the qr scanner code is not loaded upfront, right?

How do you want to test it? Maybe some custom instrumentation is needed, e.g. a timer start in the go backend and a apiPost("measureTime") as soon the frontend is loaded and log the time it has taken?

thisconnect:
> So the expectation is that the app loads faster if the qr scanner code is not loaded upfront, right?

We can measure bundle size and loading speed in the frontend, with a production build and then serve web/build.

x1ddos:
Let's get https://github.com/digitalbitbox/bitbox-wallet-app/pull/1169 
in and then try this. But looks good!

x1ddos:
> > So the expectation is that the app loads faster if the qr scanner code is not loaded upfront, right?
> 
> We can measure bundle size and loading speed in the frontend, with a production build and then serve web/build.

Measruments sound very interesting and useful. Speed is not the only metric though. I would add memory consumption to it.

@thisconnect did you get a chance to measure something here?

thisconnect:
Rebased onto current upstream/master 

Quick note: the motivation for this is not only about performance, but also that we do not load code unless it is used.

## Performance

The improvement of the 

I made `make buildweb` for master and frontend-split-qrreader-chunk and moved those to `frontends/web/build-master` and `frontends/web/build-async-qrcode`.

To run the production frontend in the browser
- `npx serve frontends/web/build-master --cors -l 8080`
- `npx serve frontends/web/build-async-qrcode --cors -l 8080`

Tested with uptodate chromium, no addons:
- devtools / Lighthouse
- checked only "Performance" and "Desktop"
- click "view trace" when lighthouse is done
- check Summary Scripting, JS Heap, Evaluate Script, less better


### Master

master

### This branch

async-qrcode

Looks like only a small improvement, but it is recommended and if we start loading more asynchronously it will of corse add up.

Currently it it is about 30% of the main bundle

Screen Shot 2021-03-02 at 1 08 14 PM

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -174,11 +174,18 @@ class Send extends Component<Props, State> {
         this.registerEvents();
         const account = this.getAccount();
         if (account && !account.coinCode.startsWith('eth-erc20-') && account.coinCode !== 'eth') {
-            this.qrCodeReader
-                .getVideoInputDevices()
-                .then(videoInputDevices => {
-                    this.setState({ hasCamera: videoInputDevices.length > 0 });
-                });
+            import('../../../components/qrcode/qrreader')
```
x1ddos, line 177: I probably don't undestand something. I see that components/qrcode/qrreader imports `BrowserQRCodeReader` but this same type is already imported at the top of this file with `import { BrowserQRCodeReader } from '@zxing/library'`.

Doesn't that mean the `BrowserQRCodeReader` is already imported since the beginning when switching to send view? Then I guess I don't understand why this import here is needed.

x1ddos:
Tried "send coins" in a qt-linux and android builds. Both load the reader correctly :+1: 

I just want to understand how this works with two seemingly identical imports in two places.

benma:
> Quick note: the motivation for this is not only about performance, but also that we do not load code unless it is used.

Can you still elaborate what the goal is? Just loading less code per se has no benefits, either it improves performance or memory load or it achieves some other goal. 

If I read the performance comparison above, it seems like the initial load is much faster, which is great! I didn't expect such a win here. Is it noticeable?

frontends/web/src/routes/account/send/send.tsx
thisconnect: the BrowserQRCodeReader is already imported, is only for Typescript and only uses the type interface.

thisconnect:
Smaller bundle is recommended for performance or memory load.

Less code means smaller attack surface, but also loading less code from the start.

for example if there was an evil dependency in a library that would try to cover our UI with a similar looking UI that for example contained something that looks like moonpay.

frontends/web/src/routes/account/send/send.tsx
x1ddos: Oh I see. So, before this change the qr reader was loaded due to this line?

```js
    private qrCodeReader: BrowserQRCodeReader = new BrowserQRCodeReader();
```

thisconnect: correct

thisconnect:
> for example if there was an evil dependency in a library that would try to cover our UI with a similar looking UI that for example contained something that looks like moonpay.

of corse that example is a bit lame, because as soon as you once visited the send view the code would be loaded and executed. But still this would make it slightly harder for an evil dependency, we did discuss this in person, but I guess you forgot. :'(

frontends/web/src/routes/account/send/send.tsx
x1ddos: :+1:

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -503,12 +514,14 @@ class Send extends Component<Props, State> {
...
             videoLoading: true,
         }, () => {
-            this.qrCodeReader
+            this.qrCodeReader && this.qrCodeReader
                 .decodeFromInputVideoDevice(undefined, 'video')
                 .then(result => {
                     this.setState({ activeScanQR: false });
                     this.parseQRResult(result.getText());
-                    this.qrCodeReader.reset(); // release camera
+                    if (this.qrCodeReader) {
```
x1ddos, line 522: Didn't you already check this a couple lines above with `this.qrCodeReader && this.qrCodeReader.decode...`?

x1ddos (approved):
LGTM. I agree with @thisconnect's that things, especially larger, should be loaded when needed, given that it is no worse than before. Even if it doesn't result in huge size or performance benefits in Qt or Android, although it did feel slightly faster on Android subjectively speaking, I'd welcome this change just to improve webdev speed where it is noticeable I believe.

frontends/web/src/routes/account/send/send.tsx
thisconnect: Yes and if it was up for me that would be perfectly fine, because I know we are not overwriting `this.qrCodeReader` anywhere else. But Typescript insisted and I didn't want to shut him up, the issue is `.then(` is async and in the meantime something could in theory have set `this.qrCodeReader` to undefined again. I am not sure what to do in these situations and just do whatevery Typescript tells me to do 🙈  

Screen Shot 2021-03-05 at 2 28 04 PM

x1ddos:
Yeah it's ok. I guess it doesn't hurt. I was just checking.
Maybe add a comment for future readers with the same question?

if (this.qrCodeReader) { // make typescript happy

thisconnect:
Maybe add a comment for future readers with the same question?

we'd have `// make typescript happy` all over the place, unclear when to write it and when not, do you want that?

also technically typescript is right.

x1ddos:
ok then. never mind :)

thisconnect:
I have one more question, in the second commit 
https://github.com/digitalbitbox/bitbox-wallet-app/pull/677/commits/fd40bdbdbaf931e53f3a1c7b17259f36acef9e46 I updated @zxing/library" to newer better?, do you want the newer one or stick with the old one?

x1ddos:
I tried my builds with this newer version on qt-linux and android. I 
guess you did on macos. I don't think windows will be any different? but 
I can try if you have doubts.

x1ddos:
Ah, to answer the question: yes, I'd keep the newer version.

thisconnect:
> guess you did on macos

yes tested macOS app, but I think I am using a little newer Qt (the one from homebrew, `import()` seems to be introduced in Chrome 63, I hope that is fine with the bundled webview
https://caniuse.com/es6-module-dynamic-import

benma:
> But still this would make it slightly harder for an evil dependency, we did discuss this in person, but I guess you forgot. :'(

Indeed :O Shame on me! It is a good point!

PR #678, opened 2019-11-11
base-settings: add button functionality

Tomasvrba:
This PR implements functionality for these settings buttons:
- Enable/Disable Tor
- Change hostname
- Change password
- Shut down
- Create backup
- Enable ssh password authentication
- Set system password

Additionally:
- Add number of active lightning channels

### DIFF
--- a/backend/bitboxbase/bitboxbase.go
+++ b/backend/bitboxbase/bitboxbase.go
@@ -615,6 +615,11 @@ func (base *BitBoxBase) ShutdownBase() error {
 	if !reply.Success {
 		return &reply
 	}
+	// FIXME: Change backend node management to status: offline
+	err = base.Deregister()
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


Tomasvrba:
@TheCharlatan here is the PR for the Base which introduced the ssh password changes: https://github.com/digitalbitbox/bitbox-base/pull/237

backend/bitboxbase/bitboxbase.go
```diff
@@ -599,6 +599,11 @@ func (base *BitBoxBase) SetLoginPassword(password string) error {
 	if !reply.Success {
 		return &reply
 	}
+	// FIXME: Change backend node management to status: offline
+	err = base.Deregister()
```
TheCharlatan: Why are you deregistering here?

Tomasvrba: Yep, good catch, that shouldn't be there :) :100:

TheCharlatan: We should have just checked the code :P

TheCharlatan: What is the FIXME for?

Tomasvrba: That one also belongs to the `RebootBase()` method. 
I will remove it altogether here. 
Part of what I'm working on for the backend base management, auto-reconnects and different statuses based on what kind of restart/disconnect/shutdown happens.
This was a reminder to change the `Deregister()` to something else when restarting.

PR #679, opened 2019-11-11
android: prompt and force app exit on back button tap

x1ddos:
This change makes the app show an exit prompt and brutally kill itself, taking down all goroutines with it. The slower startup is very slightly noticeable but I've seen much worse tbh. Not too bad for an alpha quality version.

I didn't touch shutdown/close functions. Would prefer to deal with them separately.

Screenshot_20191112-003431

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -2,13 +2,16 @@ package ch.shiftcrypto.bitboxapp;
 
 import android.Manifest;
 import android.annotation.SuppressLint;
+import android.app.AlertDialog;
 import android.content.BroadcastReceiver;
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.hardware.usb.UsbManager;
+import android.os.Process;
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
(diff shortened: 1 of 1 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -263,5 +266,42 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
...
+    // Until such solution is implemented, forced app exit seems most appropriate.
+    @Override
+    public void onBackPressed() {
+        new AlertDialog.Builder(MainActivity.this)
+            .setTitle("Close BitBoxApp")
+            .setMessage("Do you really want to exit?")
+            .setPositiveButton(android.R.string.yes, new DialogInterface.OnClickListener() {
+                public void onClick(DialogInterface dialog, int which) {
+                    // See the following for details about task stacks:
+                    // https://developer.android.com/guide/components/activities/tasks-and-back-stack
```
benma: It's a lot of text, and `moveTaskToBack` is not mentioned. A quick note why this is needed before killing the process would be helpful. It seems surprising, I am guessing it is not needed?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -263,5 +266,42 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
                 break;
         }
     }
+
+    // The app cannot currently handle the back button action to allow users
```
benma: Splendid docs :heart_eyes:

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
x1ddos: You're right, it's essentially a noop today. Added it for future proof in case the app will have another activity in which case we'll most likely never remember to call moveTaskToBack before killing the process. If moveTaskToBack isn't called, the killProcess might actually kill a descendant proc so some remnants of the app would remain. At least, that's what I understood from reading the docs.

Will re-read again and add an explanation tonight.

benma:
I tested it and it works nicely! Let me when to merge.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
x1ddos: Turns out what I said earlier was incorrect. The `moveTaskToBack` isn't a noop. Without that call the system may attempt restarting the app because it can't tell whether it crashed or exited on purpose.

That's when in foreground. When in background, it doesn't really care. So, moving the process to the background is essential to avoid potential auto-restarts.

x1ddos:
PTAL. Updated the doc comments. Good to merge if you're happy.

x1ddos:
```
level=error msg="Running error: context loading failed: no go files to analyze"
Makefile:79: recipe for target 'ci' failed
make: *** [ci] Error 5
```

Not sure what's up with that. I'll rebase to the latest master, just in case.

x1ddos:
I actually have a different error locally. So, ironically can't even reproduce a run from docker container:

```
dockeruser@c61620f60c30:/opt/go/src/github.com/digitalbitbox/bitbox-wallet-app$ go version
go version go1.13.4 linux/amd64

dockeruser@c61620f60c30:/opt/go/src/github.com/digitalbitbox/bitbox-wallet-app$ golangci-lint run
ERRO Running error: context loading failed: failed to load program with go/packages:
could not determine GOARCH and Go compiler
```

x1ddos:
It seems the docker image is missing `/opt/go/pkg/mod` and it'll have to be with `chmod 777` to make  it writable because the user isn't root.

Once I did that, the run shows:

```
verifying github.com/btcsuite/btcd@v0.20.0-beta: checksum mismatch
        downloaded: h1:PamBMopnHxO2nEIsU89ibVVnqnXR2yFTgGNc+PdG68o=
        go.sum:     h1:DnZGUjFbRkpytojHWwy6nfUSA7vFrzWXDLpFNzt74ZA=
```

x1ddos:
Alternatively, `GO111MODULE=off golangci-ling run` also works but no idea whether it runs the checks in the same way on all the code base. Adding `--verbose` didn't help to find out.

benma:
Thanks a lot. I made an exception and pushed this without a CI rebuild, as no Go files have been touched, and I ran and tested the android build locally.

PR #684, opened 2019-11-15
ci: make golangci-lint great again

x1ddos:
This resolves the following error

    + golangci-lint run
    level=error msg="Running error: context loading failed:
    no go files to analyze"
    Makefile:79: recipe for target 'ci' failed

I believe this is because of a recent golangci-lint upgrade.
It seems to be running with the default go modules support now,
which in turn tries downloading all the package dependencies.

Unfortunately, one package dep was added to go.sum at the time
where package checksums were computed incorrectly in certain conditions.
More on that can be found in golang/go issue 29278.

This went unnoticed for quite some time because vendored dir is used
instead of go.mod. Since the upgrade of golangci-lint, it started
using go.mod.

This change updates go.mod checksum, even though it isn't used,
and makes golangci-lint use vendored dir as it should.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,9 @@ run:
   # include test files or not, default is true
   tests: true
 
+  # ignore go.mod and use vendor dir instead
+  modules-download-mode: vendor
+
 # output configuration options
 output:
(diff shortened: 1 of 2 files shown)


benma:
:heart:

PR #691, opened 2019-11-17
backend/bitboxbase: remove Self()

benma:
Maybe the whole interface should be removed if this function, breaking
the point of the interface, worked without issues.


TheCharlatan:
Thanks Marko, you are so right! This removes the point of the entire interface. If you don't want to go through the hassle of removing it, I have a patch ready now.

benma:
If you wanna get rid of the interface, feel free, and then I'll close this.

PR #698, opened 2019-11-20
ui updates to connectbase component

originalblend337:
- bigger items on larger screens
- sortable items feature
- layout for mobile screens


Stadicus:
Awesome layout! :-)

One thing I'd change is the small font size of "Connect manually" which could either be the same size as the rest, or a button.

And could we make the IP addresses a bit more gray, so that these are not so prominent?

PR #700, opened 2019-11-20
base-backend: refactor initial connection and noise pairing

Tomasvrba:
Because:
Previously, the '/connectbase' endpoint would not succeed until after
the noise pairing hash was confirmed by the user. This means that when
connecting to the Base, the user would not be routed to the setup
wizard until after the pairing had succeeded (which happens
automatically now, but in final implementation would mean the user
would never be routed to the wizard)

This commit:
Refactors the intial rpc client ping and websocket dial to a separate
function which succeeds on initial connection. Only after this is the
noise pairing initialized.

### DIFF
--- a/backend/bitboxbase/bitboxbase.go
+++ b/backend/bitboxbase/bitboxbase.go
@@ -39,6 +39,9 @@ type Interface interface {
 	// Identifier returns the bitboxBaseID.
 	Identifier() string
 
+	// ConnectRPCClient connects to the RPC client of the Base whose initial connection has already been established
+	ConnectRPCClient() error
+
 	// GetRPCClient returns the rpcClient so we can listen to its events.
 	RPCClient() *rpcclient.RPCClient
 
@@ -204,6 +207,14 @@ func NewBitBoxBase(address string,
 // Self returns the current bitbox base instance.
 func (base *BitBoxBase) Self() *BitBoxBase {
 	return base
+}
+
+// EstablishConnection establishes initial websocket connection with the middleware
+func (base *BitBoxBase) EstablishConnection() error {
+	if err := base.rpcClient.EstablishConnection(); err != nil {
+		return err
+	}
+	return nil
 }
 
 // ConnectRPCClient starts the connection with the remote bitbox base middleware
--- a/backend/bitboxbase/handlers/handlers.go
+++ b/backend/bitboxbase/handlers/handlers.go
@@ -27,6 +27,7 @@ import (
 
 // Base models the api of the base middleware
 type Base interface {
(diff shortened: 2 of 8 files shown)


TheCharlatan:
> but in final implementation would mean the user
> would never be routed to the wizard)
I don't understand what you mean with this. As soon as the user's pairing on the base is successful, he gets routed to the wizard. That's the current behaviour.

Tomasvrba:
> > but in final implementation would mean the user
> > would never be routed to the wizard)
> > I don't understand what you mean with this. As soon as the user's pairing on the base is successful, he gets routed to the wizard. That's the current behaviour.

Technically correct, I didn't explain that enough They will only get routed to the wizard after the pairing is confirmed and at that point they will get routed to the password page. Meaning they will never get routed to the verify pairing screen, not see the code, never be able to visually confirm it, and because of that never actually be routed to the wizard. Unless they of course just confirm it blindly on the Base without seeing it in the App :)

TheCharlatan:
> They will only get routed to the wizard after the pairing is confirmed and at that point they will get routed to the password page.

But that is the intended behaviour. Currently the following sequence of frontend events is implemented:
Base not Paired, No Password set, Setup Wizard not run through:
StatusUnpaired -> StatusPasswordNotSet -> StatusBitcoinPre -> StatusInitialized
Base Paired, Password set, Setup Wizard not run through:
StatusLocked ->StatusBitcoinPre -> StatusInitialized
Base Paired, no password set, Setup Wizard ran:
SttatusLocked -> StatusPasswordNotSet -> StatusInitialized

Which sequence of events are you missing?

Tomasvrba:
> > They will only get routed to the wizard after the pairing is confirmed and at that point they will get routed to the password page.
> 
> But that is the intended behaviour. Currently the following sequence of frontend events is implemented:
> Base not Paired, No Password set, Setup Wizard not run through:
> StatusUnpaired -> StatusPasswordNotSet -> StatusBitcoinPre -> StatusInitialized

The problem with this sequence is that the frontend never receives the `StatusUnpaired` Status to correctly route the user to the pairing screen, because the subscription to the `/event` endpoint can only happen after we know that we have successfully connected to the Base. However, currently we don't know that until the `/connectbase` call returns. But the pairing code verification is handled within this API call, so it never actually returns until after the user confirms the pairing. So the frontend doesn't know that the rpc client connection and websocket connection succeeded until after the pairing is confirmed. 

This PR unbundles that functionality so that the `rpcClient.Ping()` to the middleware and `websocket.Dial` happen as an initial connection, then the frontend knows it succeeded, the noise pairing can be initialized and the user routed to the pairing screen.

PR #709, opened 2019-11-26
Base pairing

Tomasvrba:
@KasparEtter could you please take a look at this commit: https://github.com/digitalbitbox/bitbox-wallet-app/commit/76545637be7633cbade7e4698d427f679bfe120b

specifically: https://github.com/Tomasvrba/bitbox-wallet-app/blob/e4696353cbc1b4b4d23313d381593eb6b5abef4d/frontends/web/src/routes/bitboxbase/bitboxbase.tsx#L94

I had to type the `registeredBases[baseID]` as `any` because the `RegisteredBases` interface includes a boolean:
```
interface RegisteredBases {
    [ID: string]: {
        paired?: boolean;
        status?: BaseStatus;
    };
}
```
This is because:
```
type T1 = number & string;  // number & string
type T2 = string & boolean;  // never
```

If there is no boolean field in the `RegisteredBases` interface, then `newState.registeredBases[baseID][key] = value;` works fine. But with the boolean, it's a `never` type. So I'm wondering if you have a clever way to do this without typing it as any :)

See https://github.com/microsoft/TypeScript/issues/31663 for context.

### DIFF
--- a/frontends/web/src/routes/bitboxbase/bitboxbase.tsx
+++ b/frontends/web/src/routes/bitboxbase/bitboxbase.tsx
@@ -26,6 +26,7 @@ import { PasswordRepeatInput } from '../../components/password';
 import { Spinner } from '../../components/spinner/Spinner';
 import { Step, Steps } from '../../components/steps';
 import * as stepStyle from '../../components/steps/steps.css';
+import { share } from '../../decorators/share';
 import { translate, TranslateProps } from '../../decorators/translate';
 import '../../style/animate.css';
 import { apiSubscribe } from '../../utils/event';
@@ -33,7 +34,7 @@ import { apiGet, apiPost } from '../../utils/request';
 import SimpleMarkup from '../../utils/simplemarkup';
 import { BaseSettings } from './basesettings';
 import * as style from './bitboxbase.css';
-import { BaseStatus, baseStore } from './bitboxbaseconnect';
+import { BaseStatus, baseStore, RegisteredBaseFields, SharedProps as SharedBaseProps } from './bitboxbaseconnect';
 
 export interface BitBoxBaseProps {
     bitboxBaseID: string | null;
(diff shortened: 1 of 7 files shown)


KasparEtter:
Solved the TypeScript issue with Tom directly. 🙂

Tomasvrba:
> Solved the TypeScript issue with Tom directly. slightly_smiling_face

Fixed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/709/commits/d67525489d14b0d5d1ecd2d9652c9e581b146dd0

TheCharlatan:
Did you solve the timeout issue yet?

Tomasvrba:
> Did you solve the timeout issue yet?

No, we haven't traced where exactly that happens yet. Well, there also seem to be two different timeout issues. 
Currently lower prio for beta, since confirmation within 30 seconds works fine :)

frontends/web/src/routes/bitboxbase/bitboxbase.tsx
```diff
@@ -74,8 +75,26 @@ export interface BaseUpdateInfo {
...
+    baseStore.setState({registeredBases: initialState});
+};
+
+export const updateSharedBaseState = <K extends keyof RegisteredBaseFields>(key: K, value: RegisteredBaseFields[K], baseID: string) => {
+    if (!(baseID in baseStore.state.registeredBases)) {
+        initSharedBaseState(baseID);
+    }
+    const newState = Object.assign({}, baseStore.state);
+    newState.registeredBases[baseID][key] = value;
+    baseStore.setState(newState);
```
thisconnect: In theory when you use current/previous state in setState you should pass a function to setState. The reason is that setState may be asynchronously and current/previous state might have changed in the meantime, so this probably won't matter.

https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

Tomasvrba: Indeed, except in this case it's not the setState from React that's being called, but our custom setState from here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/ed94890b7354c119cbde9111c79ac1720a65da13/frontends/web/src/decorators/store.ts#L56
which doesn't take a callback like setState in React and instead it calls `forceUpdate()` on each component. :man_shrugging: 
I'm open to suggestions :)

thisconnect: ah yeah I forgot, in that case please ignore my comment. sorry!!!!

forceUpdate :(

PR #711, opened 2019-11-26
Socksproxy

benma:


TheCharlatan:
Won't this make the app crash immediately if tor is not configured properly?

benma:
Not unless I missed something. `GetTCPProxyDialer()` had one error path via `proxy.SOCKS5()`, and that function only returns `nil` for error:

benma:
I do wonder why that function is like that. I highly dislike unused arguments/return values. Maybe it was more convenient for the stdlib to have a consistent signature for dialers? :man_shrugging:

benma:
With all the passing of socksproxy/dialers/http clients to enable Tor, I am almost tempted to use an evil global singleton :scream: @x1ddos talk me out of it :joy:

x1ddos:
Ok, I'll take a look at this whole socks business and report back here. :)

PR #712, opened 2019-11-26
rpc

benma:


benma:
Slowly getting rid of rpc.go and untangling the rpc mess.

PR #713, opened 2019-11-27
Jsonrpc

benma:


benma:
Best viewed commit by commit, as each builds on the previous and does one small step.

Overall goal: get rid of util/rpc and the interface contained there, and moving ServerInfo to backend/config where it originated.

backend/coins/btc/electrum/electrum.go
```diff
@@ -139,9 +123,15 @@ func NewElectrumConnection(servers []*rpc.ServerInfo, log *logrus.Entry, dialer
 	log = log.WithFields(logrus.Fields{"group": "electrum", "server-type": "electrumx", "servers": serverList})
 	log.Debug("Connecting to Electrum server")
 
-	backends := []rpc.Backend{}
+	backends := []*jsonrpc.Backend{}
```
x1ddos: `var []*jsonrpc.Backend` would make it slightly easier to read.
https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

backend/coins/btc/electrum/electrum.go
```diff
@@ -25,44 +25,28 @@ import (
...
-func (electrum *Electrum) ServerInfo() *rpc.ServerInfo {
-	return electrum.serverInfo
-}
-
-// EstablishConnection connects to a backend and returns an rpc client
+// establishConnection connects to a backend and returns an rpc client
 // or an error if the connection could not be established.
-func (electrum *Electrum) EstablishConnection() (io.ReadWriteCloser, error) {
+func establishConnection(
+	serverInfo *config.ServerInfo, dialer proxy.Dialer) (io.ReadWriteCloser, error) {
```
x1ddos: 80 or so characters isn't really Go's thing: https://github.com/golang/go/wiki/CodeReviewComments#line-length
I'd put it all on a single line.

benma: > There is no rigid line length limit in Go code, but avoid uncomfortably long lines.

In this project, it's not 80, but a "soft" rule of ~100. Imho not worth changing this instance as we have tons of line splits like this one. Should make it official with a style guideline, which will also include "no short variable names that have a bigger scope than a few lines" :laughing:

backend/coins/btc/electrum/electrum.go
benma: Done

backend/coins/btc/electrum/electrum.go
x1ddos: It just makes it harder to read when you look at 1000s lines of a consistent style Go code and suddenly see this.

x1ddos: (as in, 1000s lines in other random projects)

x1ddos: Could've also been a `dial(conf *config.ServerInfo, dialer proxy.Dialer) (io.ReadWriteCloser, error)` which you'll most likely object due to the naming. :)

benma: Why would I object to `dial`? Sounds like a very good name ;)

benma: > It just makes it harder to read when you look at 1000s lines of a consistent style Go code and suddenly see this.

Eh, I am really past any formatting discussions. I wish `gofmt` would leave *zero* room for discussion, and every language had a `gofmt` equivalent. After `gofmt`, the things that remain hardly matter for readability compared to everything else, like architecture, structure, docs, tests, and so on.

PR #724, opened 2019-12-04
Persist base config

Tomasvrba:


backend/bitboxbase/bbbconfig/bbbconfig.go
```diff
@@ -0,0 +1,229 @@
...
+const configFilename = "bitboxbase.json"
+
+// RegisteredBase has information about BitBoxBases registered with the backend
+type RegisteredBase struct {
+	BaseID   string `json:"baseID"`
+	Hostname string `json:"hostname"`
+}
+
+// BBBConfigurationInterface provides an interface to interact with the persisted BBBConfig
+type BBBConfigurationInterface interface {
```
TheCharlatan: I don't see what you need the interface for. Are you sure you actually need it?

Tomasvrba: Yeah, I suppose not since it's just implemented once in the same file. I'll remove it

PR #736, opened 2019-12-12
frontend: help (tooltip) feature

originalblend337:
- display a boxed question mark inline which will trigger a tooltip on click
- plans to add functionality to open a guide entry instead of a tooltip in the future

### DIFF
--- a/frontends/web/src/components/help/help.tsx
+++ b/frontends/web/src/components/help/help.tsx
@@ -1,0 +1,85 @@
+/**
+ * Copyright 2019 Shift Devices AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/help/help.tsx
```diff
@@ -0,0 +1,75 @@
...
+            });
+        }
+    }
+
+    public render(
+        { position, text }: RenderableProps<Props>,
+        { show }: State,
+    ) {
+        return (
+            <a className={style.container} onClick={this.showText}>?
```
Tomasvrba: Hmmm, generally I would expect tooltips to show up on `hover`, not on click. Isn't that the idea of a tooltip?

thisconnect: ideally both, and also on focus + tabindex

Tomasvrba:
Closing, currently not implementing tooltips

PR #748, opened 2019-12-17
Accountevents

benma:


benma:
The plan after is to get rid of AccountsStatus(), make backend observable, and use the subscribe decorator to get update the accounts in the frontend.

PR #751, opened 2019-12-18
Subscribe

benma:


frontends/web/src/app.tsx
```diff
@@ -249,5 +237,14 @@ class App extends Component<Props, State> {
...
 }
 
-const HOC = translate()(App);
+const subscribeHOC = subscribe<SubscribedProps, TranslateProps>(
+    {
+        accounts: 'accounts',
+        devices: 'devices/registered',
+    },
+    true,
+    false,
```
Tomasvrba: Technically I don't think you need the true and false here since they're the default values for those params?

benma: good point, i missed that

benma: Chose to not do it anyway to save some time, can't hurt to be explicit either.

PR #753, opened 2019-12-18
frontend: enforce state type

benma:
Otherwise a non-state var could sneak in.

Readonly because the state is supposed to be set via setState().

### DIFF
--- a/frontends/web/src/components/backups/restore.tsx
+++ b/frontends/web/src/components/backups/restore.tsx
@@ -45,7 +45,7 @@ interface State {
 }
 
 class Restore extends Component<Props, State> {
-    public state = {
+    public readonly state: State = {
         isConfirming: false,
         activeDialog: false,
         isLoading: false,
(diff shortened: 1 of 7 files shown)


benma:
I saw that but unfortunately this is a re-declaration. Without the readonly, you can just set the state directly anyway.

This is also the reason why `: State` is needed here, even though it already has the correct type in base component.

I think the really proper way is to not set the state like this, but set in in the constructor. Both work but the latter does not need the correct redeclaration.

Tomasvrba:
> I saw that but unfortunately this is a re-declaration. Without the readonly, you can just set the state directly anyway.
> 
> This is also the reason why `: State` is needed here, even though it already has the correct type in base component.
> 
> I think the really proper way is to not set the state like this, but set in in the constructor. Both work but the latter does not need the correct redeclaration.

Yep, you're right, I didn't realize just from looking at the diff it wasn't in the constructor.

PR #762, opened 2020-01-08
base: implement user-facing error messages based on error codes

Tomasvrba:
When an RPC to the BitBoxBase fails, the Base returns a predefined
error code from errorcodes.go, this commit maps these error codes
to user facing messages with the fallback value of
```
"UNEXPECTED_ERROR": "An unexpected error occurred. Please see
application log for additional details."
```
in the case the Base returns an unknown or empty error code when an
error occurs.


frontends/web/src/components/bitboxbase/confirmbaserpc.tsx
```diff
@@ -72,7 +72,7 @@ class ConfirmBaseRPC extends Component<Props, State> {
                 this.props.onSuccess && this.props.onSuccess();
                 this.setState({ success: true });
             } else {
-                this.setState({ success: false, failureMessage: response.message });
+                this.setState({ success: false, failureMessage: this.props.t(`bitboxBase.errors.${response.code}`, this.props.t('bitboxBase.errors.UNEXPECTED_ERROR')) });
```
benma: split into many lines?

frontends/web/src/utils/bbbaseError.ts
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { alertUser } from '../components/alert/Alert';
+import { Translate } from '../decorators/translate';
+
+export const bbBaseErrorMessage = (errorCode: string, t: Translate)  => {
```
benma: name sounds like it just returns the error message, but it actually does the alert. would be good to add `alert` somewhere in the name.

Tomasvrba:
Closing for now.

PR #784, opened 2020-02-17
qt/server: ignore -psn_xxxx flags on macOS

benma:
This is a blind attempt to fix the issue that the app does not start
on macOS the first time, somtimes.

The args are logged so we can inspect if the weird arg was really
passed or not.

### DIFF
--- a/frontends/qt/server/server.go
+++ b/frontends/qt/server/server.go
@@ -41,11 +41,15 @@ import "C"
 
 import (
 	"flag"
+	"os"
+	"runtime"
+	"strings"
 	"unsafe"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/bridgecommon"
 	btctypes "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/types"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/devices/usb"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/system"
 )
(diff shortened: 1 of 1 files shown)


frontends/qt/server/server.go
```diff
@@ -76,11 +80,23 @@ func serve(
...
 	// unrecognized flags
 	// _ = flag.Int("remote-debugging-port", 0, "")
 	testnet := flag.Bool("testnet", false, "activate testnets")
 
+	if runtime.GOOS == "darwin" {
+		// eat "-psn_xxxx" on Mac, which is passed when starting an app from Finder for the first time.
+		// See also: https://stackoverflow.com/questions/10242115/os-x-strange-psn-command-line-parameter-when-launched-from-finder
+		for _, arg := range os.Args[1:] {
+			if strings.HasPrefix(arg, "-psn_") {
+				_ = flag.Bool(arg[1:], false, "<ignored>")
```
benma: @x1ddos any idea how to ignore an argument more nicely?

benma:
Tested, this fixes the issue.

frontends/qt/server/server.go
x1ddos: Not really. This is kind of the "canonical" way to ignore flags with stdlib flag package, for better or worse... :)
A slightly nicer way would be to get rid of `_ =` assignment part. :-p

Also, I would probably do something like `strings.TrimLeft(arg, "-")` as a tiny bit more future proof in case they start using double dashes in the flags.

benma: Thx :pray:

PR #798, opened 2020-03-19
Eth

benma:


benma:
@Tomasvrba did you also test that erc20 tokens still work as expected? (it's the same underlying code)

PR #806, opened 2020-03-30
backend/rates: log failure

benma:


backend/rates/rates.go
```diff
@@ -81,11 +82,16 @@ func (updater *RateUpdater) update() {
 	}()
 
 	var rates map[string]map[string]float64
-	err = json.NewDecoder(response.Body).Decode(&rates)
+	responseBody, err := ioutil.ReadAll(response.Body)
```
x1ddos: You'll probably want to limit on max bytes. I'd imagine something like:

```go
const max = 16384
b, err := ioutil.ReadAll(io.LimitReader(response.Body, max+1))
// handle err
if len(b) > max {
  // report "rates response is too large" error
}
```

benma: Done

PR #810, opened 2020-04-23
frontends/qt: enable HiDPI auto-scaling on Windows

x1ddos:
This makes the app scale well with the rest of the display
when the scaling factor is greater than 100%, especially font sizes.

Only Windows because there has been unexpected behavior on other
platforms in the past, at least on Qt versions before 5.14.
But we are currently stuck at 5.11 due to various dependencies and
older platforms support. Will reconsider and possibly upgrade to 5.14
in about a year.

See more about Qt HiDPI here:
https://doc.qt.io/qt-5/highdpi.html

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -109,13 +109,12 @@ public:
 
 int main(int argc, char *argv[])
 {
-    // note: doesn't work as expected. Users with hidpi enabled should set the environment flag themselves
-    // turn on the DPI support**
-// #if QT_VERSION >= QT_VERSION_CHECK(5,6,0)
-//     QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
-// #else
-//     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", QByteArray("1"));
-// #endif // QT_VERSION
+// Enable auto HiDPI scaling on Windows only for now.
+// Historically, auto scaling did not work as expected on other platforms
+// before Qt v5.14 but we're at 5.11 due to older systems support.
+#if defined(_WIN32) && QT_VERSION >= QT_VERSION_CHECK(5,6,0)
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+#endif
 
 // QT configuration parameters which change the attack surface for memory
 // corruption vulnerabilities


frontends/qt/main.cpp
```diff
@@ -109,13 +109,12 @@ class WebEngineView : public QWebEngineView {
...
-    // turn on the DPI support**
-// #if QT_VERSION >= QT_VERSION_CHECK(5,6,0)
-//     QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
-// #else
-//     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", QByteArray("1"));
-// #endif // QT_VERSION
+// Enable auto HiDPI scaling on Windows only for now.
+// Historically, auto scaling did not work as expected on other platforms
+// before Qt v5.14 but we're at 5.11 due to older systems support.
+#if defined(_WIN32) && QT_VERSION >= QT_VERSION_CHECK(5,6,0)
```
benma, line 115: since we are at least on qt 5.11, the `QT_VERSION >= QT_VERSION_CHECK(5,6,0)` check could be skipped.

x1ddos: True but so could the others below.
How about I make a separate PR to remove all these conditions and update readme in places where it mentions Qt dependencies?

jonasschnelli:
In Bitcoin Core, we set this flag for all platforms. Has anyone tested this under Ubuntu (or say a HiDPI capable linux windows manager)?

benma:
> In Bitcoin Core, we set this flag for all platforms. Has anyone tested this under Ubuntu (or say a HiDPI capable linux windows manager)?

as the comment in the PR says, there at least used to be unexpected issues. @x1ddos mentioned this fix messed sth up on his linux.

we only ever got reports for issues on windows, never for linux/macOS, I believe.

x1ddos:
To add to @benma's comment: the issue on a hidpi linux - huge size - may have been fixed in Qt 5.14 but we can't upgrade to that Qt version yet because it requires ubuntu 18.04 at the minimum and we would loose ubunut 16.04 support which is still active for about a year from now.

jonasschnelli:
I see. So we link against the system qt libraries? I guess static builds (at least for the qt libs) are a complex beast?

benma:
>  I guess static builds (at least for the qt libs) are a complex beast?

Very much. In fact, static builds are not officially supported (QtWebEngine), and making it work might be next to impossible.

jonasschnelli:
Yeah. Sorry for rehashing the discussion. A PPA package of our app (that could include some dependent libraries) would maybe be nice? However, ... I'm starting to sidetrack this PR. Ignore me.

x1ddos:
> Yeah. Sorry for rehashing the discussion. A PPA package of our app (that could include some dependent libraries) would maybe be nice? However, ... I'm starting to sidetrack this PR. Ignore me.

Yeah, +1 to a PPA. It will definitely come sooner or later!

PR #811, opened 2020-04-28
macOS build: minor fixes

Stadicus:
When building the BitBoxApp on macOS High Sierra with current software versions, there were a few issues. This pull request 
* installs 'homebrew' as recommended on brew.sh (the previous command recommends this as well in a warning)
* replaces 'pwd' with the var REPOROOT so that make can be called from outside the working directory
* moves necessary environment variables to the top of Makefile and removes the 'export' commands (that were not doing anything)
* removes 'brew install qt' as there is no reliable way to install the older version Qt 5.11 inside the script
* updates the README with more specific information about dependencies and versions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SHELL:=/bin/bash
-WEBROOT:=`pwd`/frontends/web
+SHELL    := /bin/bash
+REPOROOT := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+WEBROOT  := $(REPOROOT)/frontends/web
+GOPATH   ?= $(HOME)/go
+PATH     := $(PATH):$(GOPATH)/bin
 
 catch:
 	@echo "Choose a make target."
@@ -26,17 +29,14 @@ envinit:
 	GO111MODULE=off go get -u github.com/jteeuwen/go-bindata/...
 	GO111MODULE=off go get -u golang.org/x/mobile/cmd/gomobile
 	gomobile init
-# This must be run in $GOPATH/src/github.com/digitalbitbox/bitbox-wallet-app
+# Initializiation on MacOS
+#  - run make from $GOPATH/src/github.com/digitalbitbox/bitbox-wallet-app
+#  - additional dependencies: Qt 5.11 & Xcode command line tools
+#  - add to $PATH: /usr/local/opt/go@1.13/bin
 osx-init:
-	/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+	/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 	brew install yarn
-	brew install go
-	brew install qt
-	export PATH="/usr/local/opt/qt/bin:$PATH"
-	export LDFLAGS="-L/usr/local/opt/qt/lib"
-	export CPPFLAGS="-I/usr/local/opt/qt/include"
-	export GOPATH=~/go/
-	export PATH=$PATH:~/go/bin
+	brew install go@1.13	
 	$(MAKE) envinit
 servewallet:
 	go install -mod=vendor ./cmd/servewallet/... && servewallet
--- a/README.md
+++ b/README.md
@@ -45,17 +45,23 @@ The below instructions assume a unix environment.
 
 ### Requirements
 
-- [Go](https://golang.org/doc/install) version 1.13.
-- [Yarn](https://yarnpkg.com/en/) - for managing the web UI deps.
-- [Qt5](https://www.qt.io)
-  - Install via https://www.qt.io/download, also install WebEngine, and put `qmake` and `rcc` into
-    your PATH.
+The following dependencies need to be installed:
 
-Make sure `$GOPATH` is set and `$GOPATH/bin` and `$GOROOT/bin` is in your `$PATH`.
+- [Go](https://golang.org/doc/install) version 1.13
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ WEBROOT:=`pwd`/frontends/web
 catch:
 	@echo "Choose a make target."
 envinit:
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOPATH)/bin v1.19.1
+	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b ${GOPATH}/bin v1.19.1
```
x1ddos, line 21: This is strange. Both should behave exactly the same:

> To substitute a variable’s value, write a dollar sign followed by the name of the variable in parentheses or braces: either ‘$(foo)’ or ‘${foo}’ is a valid reference to the variable foo

(from https://www.gnu.org/software/make/manual/html_node/Reference.html#Reference)

Or is it not GNU make? Either way, I think POSIX make should be the same. What's your `make --version`?

But we really should update to go1.14 and use the standard go install method: https://github.com/golangci/golangci-lint#go

Makefile
```diff
@@ -28,15 +28,15 @@ envinit:
...
+	/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 	brew install yarn
 	brew install go
 	brew install qt
 	export PATH="/usr/local/opt/qt/bin:$PATH"
 	export LDFLAGS="-L/usr/local/opt/qt/lib"
 	export CPPFLAGS="-I/usr/local/opt/qt/include"
 	export GOPATH=~/go/
-	export PATH=$PATH:~/go/bin
+	export PATH=${PATH}:~/go/bin
```
x1ddos, line 39: This should really be `$(PATH):$(GOPATH)` to make it consistent because in `envinit` target above is using `$(GOPATH)/bin` anyway.

But it isn't nice to overwrite user-set GOPATH. I would've expected it to be defined somewhere at the top of the makefile. For example:

```
$ cat Makefile
GOPATH ?= /default/gopath

dummy:
	@echo $(GOPATH)
```

Then, it works as expected:

```
$ unset GOPATH
$ make
/default/gopath
$ export GOPATH=/let/me/set/my/gopath
$ make
/let/me/set/my/gopath
$ make GOPATH=/another/one
/another/one
$ GOPATH=/another/one make
/another/one
```

x1ddos: More about env variables in make: https://www.gnu.org/software/make/manual/html_node/Environment.html

Makefile
Stadicus: That probably depends on the shell, and on macOS I get the following behaviour:

```
$ export FOO=bar
$ echo $FOO
bar
$ echo ${FOO}
bar
$ echo $(FOO)
-bash: FOO: command not found
```

x1ddos: Could you try this simple makefile:

```
$ cat Makefile
dummy:
	@echo ${SHELL}
```

And also, what does `make --version` say?

Stadicus: Ah, the errors I had might not be directly related, if GNU make handles variable substitution independently from the shell.

Stadicus: Fixed this issue by moving the variables PATH and GOPATH to the top of the `Makefile`

Makefile
```diff
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SHELL:=/bin/bash
-WEBROOT:=`pwd`/frontends/web
+SHELL    := /bin/bash
+REPOROOT := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```
Stadicus, line 16: Replaced `pwd` with the root path of the repository

Makefile
```diff
@@ -26,17 +29,14 @@ envinit:
...
+#  - add to $PATH: /usr/local/opt/go@1.13/bin
 osx-init:
-	/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+	/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 	brew install yarn
-	brew install go
-	brew install qt
-	export PATH="/usr/local/opt/qt/bin:$PATH"
-	export LDFLAGS="-L/usr/local/opt/qt/lib"
-	export CPPFLAGS="-I/usr/local/opt/qt/include"
```
Stadicus, line 37: These variables were not active and are dropped.

README.md
```diff
@@ -45,17 +45,23 @@ The below instructions assume a unix environment.
...
-  - Install via https://www.qt.io/download, also install WebEngine, and put `qmake` and `rcc` into
-    your PATH.
+The following dependencies need to be installed:
 
-Make sure `$GOPATH` is set and `$GOPATH/bin` and `$GOROOT/bin` is in your `$PATH`.
+- [Go](https://golang.org/doc/install) version 1.13
+- [Yarn](https://yarnpkg.com/en/) for managing the web UI deps
+- [Qt5](https://www.qt.io) version 5.11.3
+  - Installers for older versions can be found in [this Qt archive](https://download.qt.io/new_archive/qt/5.11/5.11.3)
+  - install Qt for your plattform, plus WebEngine.
```
x1ddos, line 54: Typo: "plattform", too many `t`s.

Also, I would say "Install Qt for your platform, including the WebEngine component". Otherwise, it makes me thing WebEngine is something extra, outside of Qt install.

Stadicus: Ack, fixed and force-pushed

x1ddos:
:+1: thanks!

PR #815, opened 2020-04-30
Mobile

benma:


frontends/qt/server/server.go
```diff
@@ -130,6 +130,9 @@ func serve(
 			},
 			DeviceInfosFunc: usb.DeviceInfos,
 			SystemOpenFunc:  system.Open,
+			UsingMobileData: func() bool {
+				return false
```
x1ddos, line 134: I suppose a user could also be on a tethered connection over a phone but there's not much we can do, right.

x1ddos:
Looks good but:

```
cmd/servewallet/main.go:119:20: cannot use webdevEnvironment literal (type webdevEnvironment) as type backend.Environment in argument to backend.NewBackend:

	webdevEnvironment does not implement backend.Environment (missing UsingMobileData method)

# github.com/digitalbitbox/bitbox-wallet-app/frontends/qt/server

frontends/qt/server/server.go:133:4: unknown field 'UsingMobileData' in struct literal of type bridgecommon.BackendEnvironment
```

benma:
> Looks good but:

yep fixed up CI now hopefully.

frontends/qt/server/server.go
benma: I guess not.

PR #816, opened 2020-04-30
Backup refresh

### LINKED ISSUE #814: After restoring wallet from mnemonics and creating backup, the process is not finished
When you set up a new wallet on BitBox02 with mnemonic seed, and then create a backup on microSD card the procedure is not finished properly.

When selecting create backup a BitBoxApp asks to unlock the device. After unlocking the device the process is finished there (device idle or waiting the App) but nothing happens in the BBApp. The orange button "Check backup" is inactive.

There should be a message saying something like: "Congrats! You just created a backup and you should double check it in the next steps." or show the orange button "Check backup" as active.

If you go back (Back button is active) and then again to "Manage backups" all is as expected and the backup is visible.

benma:
Closes #814

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -921,6 +921,7 @@ func (backend *Backend) Register(theDevice device.Interface) error {
 		backend.onDeviceUninit(theDevice.Identifier())
 		return err
 	}
+	theDevice.Observe(backend.Notify)
 
 	// Old-school
 	select {
--- a/backend/devices/bitbox02/device.go
+++ b/backend/devices/bitbox02/device.go
@@ -24,6 +24,8 @@ import (
 	keystoreInterface "github.com/digitalbitbox/bitbox-wallet-app/backend/keystore"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/signing"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/observable"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/observable/action"
 	bitbox02common "github.com/digitalbitbox/bitbox02-api-go/api/common"
 	"github.com/digitalbitbox/bitbox02-api-go/api/firmware"
 	"github.com/digitalbitbox/bitbox02-api-go/util/semver"
@@ -42,6 +44,8 @@ type Device struct {
 	mu       sync.RWMutex
 	onEvent  func(event.Event, interface{})
 	log      *logrus.Entry
+
+	observable.Implementation
 }
 
 // NewDevice creates a new instance of Device.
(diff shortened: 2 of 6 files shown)


backend/backend.go
```diff
@@ -921,6 +921,7 @@ func (backend *Backend) Register(theDevice device.Interface) error {
 		backend.onDeviceUninit(theDevice.Identifier())
 		return err
 	}
+	theDevice.Observe(backend.Notify)
```
benma, line 924: there are lots of lines like this in backend.go, where in theory we should store the result (which is the unsubscribe function), and unsubscribe from observing once the instance the device/base/whatever is not available anymore.

Would do this in a different PR at some point, but not very high prio, as an unplugged device does not send any events.

x1ddos: I guess you could call it a small memory leak?

benma: It is also a potential source of future bugs, if we add for example notifications that leak (e.g. by a long running goroutine that is not shut down or finishes after the device is unplugged). :shrug:

backend/devices/bitbox02/device.go
```diff
@@ -145,3 +149,14 @@ func (device *Device) Reset() error {
 	device.init()
 	return nil
 }
+
+func (device *Device) CreateBackup() error {
```
x1ddos, line 153: Wouldn't you also want this for `EraseBackup`?

backend/backend.go
x1ddos: I hope we have an issue for this so we won't forget!

backend/devices/bitbox02/device.go
benma: > `EraseBackup`

where is this magical function :joy: 

Kidding aside, bb02 does not support deleting backups. We deemed it most important to keep all backups so there are no accidentally deleted backups that turned out to have funds.

x1ddos: hehe. I just found it here with a grep: https://github.com/digitalbitbox/bitbox-wallet-app/blob/b6030d7203fbd6df2d8b2278d02666612186757a/backend/devices/bitbox/device.go#L936

But I didn't realize this is probably bitbox01?

x1ddos: So, basically all devices are in bitbox02-api-go except bb01. Maybe some cleanup in the future?

benma: not sure what you mean? bitbox02-api-go is bitbox02 only (firmware and bootloader).

`backend/devices/bitbox` is bitbox01 (implementing the `Device` interface)
`backend/devices/bitbox02*` is bitbox02 (using bitbox02-api-go to implement the `Device` interface`)

x1ddos: well, there's also bitboxbase. it's probably obvious to you that it's also bb02 but I kind of forgot.

no but my point was maybe move bb01 stuff in a separate repo? if it makes sense.

benma: Got no reason to move it to a different repo, it is only used here. Would be just needless overhead.

bitbox02-api-go was moved to a repo so it can be used in bitbox02-api-js.

benma: > well, there's also bitboxbase. it's probably obvious to you that it's also bb02 but I kind of forgot.

Right, didn't think of it. Maybe more accurately, bitbox02-api-go is for all firmware API coming out of bitbox02-api-firmware ;)

PR #818, opened 2020-05-04
Fixing select field positioning

jstrnbrg:


frontends/web/src/components/forms/select.css
```diff
@@ -1,5 +1,7 @@
 .select {
```
benma, line 1: do the fee targets when making a transaction still look ok?

thisconnect: tested and looks good, also tested on small screen. The Input component has the margin as well. 

It does not add extra margin after Fee Target, due to the grid.css. Last item in the grid has no margin
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/style/grid.css#L74

thisconnect:
tACK

PR #819, opened 2020-05-04
Rename transaction balance columns

jstrnbrg:


frontends/web/src/components/transactions/transactions.jsx
```diff
@@ -48,7 +47,7 @@ export default class Transactions extends Component {
                     <div className={style.address}>{t('transaction.details.address')}</div>
                     <div className={style.status}>{t('transaction.details.status')}</div>
                     <div className={style.fiat}>{t('transaction.details.fiat')}</div>
-                    <div className={style.currency}>{unit}</div>
+                    <div className={style.currency}>{t('transaction.details.balance')}</div>
```
benma, line 50: The column header now would say `Balance`, but that is wrong. The row does not show the account balance, but the transaction amount.

thisconnect: can you change the tx overview to use `Amount` and `Fiat amount`?

frontends/web/src/components/transactions/transactions.jsx
```diff
@@ -47,8 +46,8 @@ export default class Transactions extends Component {
                     <div className={style.date}>{t('transaction.details.date')}</div>
                     <div className={style.address}>{t('transaction.details.address')}</div>
                     <div className={style.status}>{t('transaction.details.status')}</div>
-                    <div className={style.fiat}>{t('transaction.details.fiat')}</div>
-                    <div className={style.currency}>{unit}</div>
+                    <div className={style.fiat}>{t('transaction.details.fiat')} {t('transaction.details.amount')}</div>
```
thisconnect, line 49: IMO don't combine the 2 translations, just use `{t('transaction.details.fiat')}` and update the i18n

benma: :+1: being smart about combining two i18n entries often just does not work in one of the languages due to different grammar, structure etc.

jstrnbrg: I see, gotcha 👍

jstrnbrg:
@thisconnect ok should be fixed now

PR #820, opened 2020-05-04
Mobile (UI)

benma:
Includes UI changes based off #815

### DIFF
--- a/backend/bridgecommon/bridgecommon.go
+++ b/backend/bridgecommon/bridgecommon.go
@@ -28,7 +28,7 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/arguments"
 	btctypes "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/types"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/devices/usb"
-	backendHandlers "github.com/digitalbitbox/bitbox-wallet-app/backend/handlers"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/handlers"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/config"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/jsonp"
@@ -44,9 +44,10 @@ type NativeCommunication interface {
 }
 
 var (
-	handlers      *backendHandlers.Handlers
-	communication NativeCommunication
-	token         string
+	backendInstance  *backend.Backend
+	handlersInstance *handlers.Handlers
+	communication    NativeCommunication
+	token            string
 
 	shutdown func()
 )
@@ -71,7 +72,7 @@ func (r *response) WriteHeader(int) {
 
 // BackendCall bridges GET/POST calls (serverless, directly calling the backend handlers).
 func BackendCall(queryID int, jsonQuery string) {
-	if handlers == nil {
+	if handlersInstance == nil {
 		return
 	}
 	query := map[string]string{}
@@ -93,7 +94,7 @@ func BackendCall(queryID int, jsonQuery string) {
 			panic(errp.WithStack(err))
 		}
 		request.Header.Set("Authorization", "Basic "+token)
-		handlers.Router.ServeHTTP(resp, request)
+		handlersInstance.Router.ServeHTTP(resp, request)
 		responseBytes := resp.Body.Bytes()
 		communication.Respond(queryID, string(responseBytes))
 	}()
@@ -155,7 +156,8 @@ func Serve(
 		WithField("version", backend.Version).
 		Info("environment")
 
-	backend, err := backend.NewBackend(
+	var err error
+	backendInstance, err = backend.NewBackend(
 		arguments.NewArguments(
 			config.AppDir(),
 			testnet,
@@ -173,19 +175,21 @@ func Serve(
 	quitChan := make(chan struct{})
 	shutdown = func() {
 		close(quitChan)
-		if err := backend.Close(); err != nil {
+		if err := backendInstance.Close(); err != nil {
 			log.WithError(err).Error("backend.Close failed")
 		}
+		handlersInstance = nil
+		backendInstance = nil
 		shutdown = nil
 	}
 
 	token = hex.EncodeToString(random.BytesOrPanic(16))
 
 	// the port is unused, as we bridge directly without a server.
-	handlers = backendHandlers.NewHandlers(backend,
-		backendHandlers.NewConnectionData(-1, token))
+	handlersInstance = handlers.NewHandlers(backendInstance,
+		handlers.NewConnectionData(-1, token))
 
-	events := handlers.Events()
+	events := handlersInstance.Events()
 	go func() {
 		for {
 			select {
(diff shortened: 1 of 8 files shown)


backend/bridgecommon/bridgecommon.go
```diff
@@ -93,12 +96,23 @@ func BackendCall(queryID int, jsonQuery string) {
...
-		handlers.Router.ServeHTTP(resp, request)
+		handlersInstance.Router.ServeHTTP(resp, request)
 		responseBytes := resp.Body.Bytes()
 		communication.Respond(queryID, string(responseBytes))
 	}()
 }
 
+// UsingMobileDataChanged should be called when the network connnection changed.
+func UsingMobileDataChanged() {
+	if backendInstance == nil {
```
x1ddos, line 107: Isn't this a concurrent access issue? Assuming it can be read here and set in `Serve` func below on line 173 at the same time.

If it isn't possible, would remove this nil check and assume it cannot be nil at the time `UsingMobileDataChanged` is called.

x1ddos:
> ... let me try it locally meanwhile.

:+1:

backend/bridgecommon/bridgecommon.go
benma: In Java world I don't check that this function is not called before Serve() is called. Same deal for BackendCall(). Looked easier to check it here instead. WDYT?

x1ddos: Yeah, I think it would look ugly to trace what's called first in Java at runtime.

My problem is with the `backendInstance == nil` check existence here. The check tells me it's possible for it to be `nil` but I don't think it's the case and so it confuses me. If it were possible though, it would be a race condition between here and in `Serve` func below on line 173 where it's set with `backendInstance, err = backend.NewBackend(...)`, in which case `backendInstance` should've been guarded with a mutex.

In the end, I don't think it is possible for `backendInstance` to be nil and so there's no need for `if backendInstance == nil` at all. The `backendInstance` is created in `Serve` which is called from `GoViewModel` in its constructor https://github.com/digitalbitbox/bitbox-wallet-app/blob/a860d64f/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java#L185. And the `GoViewModel` is created only once at the app startup or so it says in https://github.com/digitalbitbox/bitbox-wallet-app/blob/fd0c7fcd/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java#L102-L104

Anyway, my point is, `backendInstance` should either be guarded with a mutex or there's no need for `nil` check at all because it can't be. I believe it's the latter, not the former!

Looks the same for BackendCall to me but we can deal with it later.

benma: The function is called from the network broadcast receiver. It sounds hard to reason about it, but to me it sounds like that could happen at any time, even before Serve() is called. Maybe I am wrong. In any case, this is what I meant: I'd rather have the check here than try to reason about the Android SDK or our Java code.

I can add a mutex, though I don't understand how it helps? If this function is called before Serve() (race condition or not), it just doesn't do anything, but that isn't bad.

I'll add one anyway, I guess sometimes a mutex is just expected to not have to think about it.

x1ddos: Well, when a race condition is possible, a mutex prevents that which otherwise can cause a panic and crash the app. Here's a simple example. I put the output at the end: https://play.golang.org/p/ojnIQVN9GYZ

x1ddos: > The function is called from the network broadcast receiver. It sounds hard to reason about it, but to me it sounds like that could happen at any time, even before Serve()

The receiver is indeed created in the app constructor but it is registered to receive events only in `onResume` in https://github.com/digitalbitbox/bitbox-wallet-app/blob/fd0c7fcd/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java#L247 which is [always called after](https://developer.android.com/guide/components/activities/activity-lifecycle) `onCreate`, the place where `backendInstance` is eventually created.

x1ddos: OTOH if `backendInstance` access is moved elsewhere and used concurrently in some refactoring, we will most likely forget to put a mutex on it. So, I would add it now to be future safe.

benma: Makes sense of course, not sure what I was thinking ;)

Added a commit.

benma:
> I would put this in `shutdown = func() { ... }` above, to be close to what it actually guards: `backendInstance = nil`.

It does, as it's called in Shutdown(). I once read about the best practice to use locks in exported functions, not private ones.

For example, if the lock was in the private function `shutdown()`, and someone added the lock also to `Shutdown()`, there would be a deadlock.

Unresponsiveness: ooh that's bad. Do you know how to reproduce?

x1ddos:
> I once read about the best practice to use locks in exported functions, not private ones.

Locks should be close to what the guard, regardless of where the function is used. It's much easier to maintain. You shuffle functions around, factor out, export, unexport, move code around. There's only one sane constant: locks are close to what they guard.

I read this:

```go
	shutdown = func() {
		close(quitChan)
		if err := backendInstance.Close(); err != nil {
			log.WithError(err).Error("backend.Close failed")
		}
		handlersInstance = nil
		backendInstance = nil
		shutdown = nil
	}
```

and think, "oh, they must've missed the lock! let me add one to where it belongs..."

If the locks are in some exported function which just happened to call that `shutdown` today, I'll have to search for it. Who knows where that shutdown call moves tomorrow. That's a maintenance nightmare imho.

> For example, if the lock was in the private function shutdown(), and someone added the lock also to Shutdown(), there would be a deadlock.

Why would anyone add a lock to a random function which doesn't touch the guarded vars? Silly example but, would you add a filesystem lock when you allocate RAM? because you know, in end you might need to start swapping RAM to disk... :)

> Unresponsiveness: ooh that's bad. Do you know how to reproduce? Maybe the lock spans a backend call that takes a while and blocks other concurrent calls?

That was with `sync.Mutex`, without read-lock. Let me try it now...

benma:
> Why would anyone add a lock to a random function which doesn't touch the guarded vars?

You might touch the guarded var and add a lock, which would result in a deadlock.

> f the locks are in some exported function which just happened to call that shutdown today, I'll have to search for it. Who knows where that shutdown call moves tomorrow. That's a maintenance nightmare imho.

Unsure... the exported function must be well documented/specified in what vars it touches or what it does, so you can't randomly remove the shutdown call. 

Anyway, I am okay with moving it there, can't say I am speaking from experience here :sweat_smile:

x1ddos:
> Unsure... the exported function must be well documented/specified in what vars it touches or what it does, so you can't randomly remove the shutdown call.

But locks are internal implementation details. The exported funs are public for a reason - API consumers shouldn't need to care about what locks and what doesn't. Well, usually! Of course if we're talking about the public API of the sync.Mutex, then yes it describes what happens to it. :)

x1ddos:
> You might touch the guarded var and add a lock, which would result in a deadlock.

True! Can definitely happen. If you touch a var, you put its lock guard on top. It would be then caught with `go test -race`.

x1ddos:
> > Unresponsiveness: ooh that's bad. Do you know how to reproduce? Maybe the lock spans a backend call that takes a while and blocks other concurrent calls?
> 
> That was with `sync.Mutex`, without read-lock. Let me try it now...

All good :+1:. This was totally my sloppiness. Please, ignore.

x1ddos:
Oh, I didn't build this locally but I assumed you did. :)

PR #827, opened 2020-05-07
Update App name

jstrnbrg:
Found the old name quite irritating as the tab name ;)

### DIFF
--- a/frontends/web/package.json
+++ b/frontends/web/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "name": "preact2",
+  "name": "BitBoxApp",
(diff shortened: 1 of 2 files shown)


thisconnect:
there would be another one https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/package.json#L3

I don't think it is used anywhere

jstrnbrg:
@thisconnect not sure I understand, what should `preact2` be changed to?

x1ddos:
> @thisconnect not sure I understand, what should `preact2` be changed to?

Same thing, I guess: BitBoxApp :)

PR #833, opened 2020-05-13
backend/eth: hardcode gas limit for standard eth txs

benma:
Which is 21000. This saves a network request to estimate the gas
limit. It is more robust as the gas estimation api call has been
unstable lately.

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -564,10 +564,14 @@ func (account *Account) newTx(
 			}
 		}
 	}
-	gasLimit, err := account.coin.client.EstimateGas(context.TODO(), message)
-	if err != nil {
-		account.log.WithError(err).Error("Could not estimate the gas limit.")
-		return nil, errp.WithStack(errors.ErrInvalidData)
+	gasLimit := uint64(21000) // gas limit for standard ethereum transactions
+	if account.coin.erc20Token != nil {
+		n, err := account.coin.client.EstimateGas(context.TODO(), message)
+		if err != nil {
+			account.log.WithError(err).Error("Could not estimate the gas limit.")
+			return nil, errp.WithStack(errors.ErrInvalidData)
+		}
+		gasLimit = n
 	}
 
 	fee := new(big.Int).Mul(new(big.Int).SetUint64(gasLimit), suggestedGasPrice)


backend/coins/eth/account.go
```diff
@@ -564,10 +564,14 @@ func (account *Account) newTx(
...
 		}
 	}
-	gasLimit, err := account.coin.client.EstimateGas(context.TODO(), message)
-	if err != nil {
-		account.log.WithError(err).Error("Could not estimate the gas limit.")
-		return nil, errp.WithStack(errors.ErrInvalidData)
+	gasLimit := uint64(21000) // gas limit for standard ethereum transcations
+	if account.coin.erc20Token != nil {
+		var err error
+		gasLimit, err = account.coin.client.EstimateGas(context.TODO(), message)
```
x1ddos, line 570: Can leave as is but just a couple notes how I would personally see it more in idiomatic Go:

```go
gasLimit := big.NewInt(21000)
  ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
  defer cancel()
  n, err := account.coin.client.EstimateGas(ctx, message)
  if err != nil {
    return nil, ...
  }
  gasLimit.SetUint64(n)
}
fee := new(big.Int).Mul(gasLimit, suggestedGasPrice)
```

- don't use `gasLimit` for possibly undefined values, even though the function returns early with an error, I know; it pays back over long term during refactoring
- use a context with a deadline; you never know what's on the other side

But maybe it's for a separate PR. I think there are few places like this.

backend/coins/eth/account.go
```diff
@@ -564,10 +564,14 @@ func (account *Account) newTx(
 			}
 		}
 	}
-	gasLimit, err := account.coin.client.EstimateGas(context.TODO(), message)
-	if err != nil {
-		account.log.WithError(err).Error("Could not estimate the gas limit.")
-		return nil, errp.WithStack(errors.ErrInvalidData)
+	gasLimit := uint64(21000) // gas limit for standard ethereum transcations
```
x1ddos, line 567: typo: transactions

x1ddos: oh wow. you didn't need me for this. CI knows how to read!

```
ackend/coins/eth/account.go:567:63: `transcations` is a misspelling of `transactions` (misspell)

	gasLimit := uint64(21000) // gas limit for standard ethereum transcations
```

backend/coins/eth/account.go
benma: fixed first part

x1ddos: thanks!

PR #835, opened 2020-05-14
rewording location warning

jstrnbrg:


benma:
@jstrnbrg you need to change text inside locize, not in the json files directly. See also https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/docs/i18n.md

Pls ping me when done in locize :)

PR #841, opened 2020-06-02
backend: unstringlify coin codes

benma:
For clarity.

### DIFF
--- a/backend/coins/btc/coin.go
+++ b/backend/coins/btc/coin.go
@@ -1,4 +1,5 @@
 // Copyright 2018 Shift Devices AG
+// Copyright 2020 Shift Crypto AG
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -42,7 +43,7 @@ import (
 // Coin models a Bitcoin-related coin.
 type Coin struct {
 	initOnce              sync.Once
-	code                  string
+	code                  coin.Code
 	unit                  string
 	net                   *chaincfg.Params
 	dbFolder              string
@@ -59,7 +60,7 @@ type Coin struct {
 
 // NewCoin creates a new coin with the given parameters.
 func NewCoin(
-	code string,
+	code coin.Code,
 	unit string,
 	net *chaincfg.Params,
 	dbFolder string,
@@ -128,7 +129,7 @@ func (coin *Coin) Initialize() {
 }
 
 // Code implements coin.Coin.
-func (coin *Coin) Code() string {
+func (coin *Coin) Code() coin.Code {
 	return coin.code
 }
 
@@ -171,7 +172,7 @@ func (coin *Coin) Headers() *headers.Headers {
 }
 
 func (coin *Coin) String() string {
-	return coin.code
+	return string(coin.code)
 }
 
 // BlockExplorerTransactionURLPrefix implements coin.Coin.
--- a/backend/coins/coin/codes.go
+++ b/backend/coins/coin/codes.go
@@ -1,0 +1,50 @@
+// Copyright 2020 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package coin
+
+// Code represents a unique coin code. Usually the coin acronym in lowercase. ERC20-Tokens are also coins.
+type Code string
+
+const (
+	// CodeBTC is Bitcoin.
+	CodeBTC Code = "btc"
+	// CodeTBTC is Bitcoin Testnet.
+	CodeTBTC Code = "tbtc"
+	// CodeRBTC is Bitcoin Regtest.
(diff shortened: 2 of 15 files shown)


backend/devices/bitbox/relay/channel.go
```diff
@@ -267,13 +268,13 @@ func (channel *Channel) SendXpubEcho(xpubEcho string, typ string) error {
...
 	signingEcho string,
-	coin string,
+	coin coin.Code,
 	scriptType string,
 	transaction string,
 ) error {
 	return PushMessage(relayServer(), channel, map[string]string{
 		"echo":               signingEcho,
-		"coin":               coin,
+		"coin":               string(coin),
```
benma, line 277: maybe this should be an explicit map, even though key and value would be the same (in theory one should be able to change the coin codes and everything should work fine after, right?).

backend/devices/bitbox02/protobuf.go
```diff
@@ -41,8 +42,8 @@ var btcMsgOutputTypeMap = map[txscript.ScriptClass]messages.BTCOutputType{
 	txscript.WitnessV0ScriptHashTy: messages.BTCOutputType_P2WSH,
 }
 
-var ethMsgCoinMap = map[string]messages.ETHCoin{
-	"eth":                 messages.ETHCoin_ETH,
+var ethMsgCoinMap = map[coin.Code]messages.ETHCoin{
+	coin.CodeETH:          messages.ETHCoin_ETH,
 	"eth-erc20-usdt":      messages.ETHCoin_ETH,
```
benma, line 47: not sure how to deal with those, but prefer to deal with them later if at all.

backend/handlers/handlers.go
```diff
@@ -400,13 +400,13 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
...
 	// The following parameters only work for watch-only singlesig accounts at the moment.
-	jsonCoinCode := jsonBody["coinCode"]
+	jsonCoinCode := coin.Code(jsonBody["coinCode"])
 	jsonScriptType := jsonBody["scriptType"]
 	jsonAccountName := jsonBody["accountName"]
 	jsonExtendedPublicKey := jsonBody["extendedPublicKey"]
 	jsonAddress := jsonBody["address"]
 
-	coin, err := handlers.backend.Coin(jsonCoinCode)
+	theCoin, err := handlers.backend.Coin(jsonCoinCode)
```
benma, line 409: why did Go ever think it's a good idea to mix package namespace with variable namespace :(

backend/devices/bitbox/relay/channel.go
x1ddos: I would leave it as is. How frequent do the codes change? If you rename a coin code in Go but keep it unchanged in the frontend, I'd argue it might become confusing. So, if I needed to rename or add a new code, I'd prefer to modify both backend and the frontend.

backend/devices/bitbox02/protobuf.go
x1ddos: why not add them as say `coin.CodeERC20USDT`, `coin.CodeERC20Link` and so on?

backend/handlers/handlers.go
x1ddos: :smile: you used the `coinpkg` as a good workaround, I think. why not use it here as well?

x1ddos: also, an old blog post about choosing package names: https://blog.golang.org/package-names

backend/devices/bitbox02/protobuf.go
benma: cause conceptually we might add 500 of them, so i'd rather keep the codes in erc20.go (or move to some json file) and change this part instead in some other clever way that avoids tons of lines.

backend/coins/coin/codes.go
```diff
@@ -0,0 +1,50 @@
...
+	// CodeRETH is Ethereum Rinkeby
+	CodeRETH Code = "reth"
+	// CodeERC20TEST is an arbitrarily picked test ERC20 token on Ropsten.
+	CodeERC20TEST Code = "erc20Test"
+	// If you add coins, don't forget to update `testnetCoins` below.
+	// There are some more coin codes for the supported erc20 tokens in erc20.go.
+)
+
+// TestnetCoins is the subset of all coins which are available in testnet mode.
+var TestnetCoins = map[Code]struct{}{
```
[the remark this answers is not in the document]
benma: on the other hand, there is only one use of it, and there might be an arbitrarily large number of erc20 tokens that would have to be repeated here. imho pass for now, ok?

backend/coins/btc/coin.go
```diff
@@ -171,7 +172,7 @@ func (coin *Coin) Headers() *headers.Headers {
 }
 
 func (coin *Coin) String() string {
```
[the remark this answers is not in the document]
benma: It's not a string, this is `Coin` not `Code`.

backend/handlers/handlers.go
benma: done with `coinpkg`

backend/coins/btc/coin.go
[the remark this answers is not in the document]
x1ddos: oh, sorry. never mind :see_no_evil:

backend/coins/coin/codes.go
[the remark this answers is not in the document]
x1ddos: yeah, sure. sounds good. although even 100s of entries would be ok with a bool map because compared to the rest of memory usage it's a drop in the ocean.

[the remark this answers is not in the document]
benma: not worried about memory, more about maintaining multiple places with many repeated lines ;)

PR #843, opened 2020-06-02
backend/eth: fix typo when returning error

benma:
Returning nil here led to an app crash if the etherscan http call
failed.


benma:
> I guess the errcheck linter doesn't always catch it.

The errcheck linter just checks that the `err` var is used, which it is. In some cases you even want to return nil when handling an error, so a linter like this would have some false positives. But could be useful anyway ;)

PR #847, opened 2020-06-06
backend/eth: query gasLimit also for the ETH account

benma:
With a premature optimization, the gasLimit for the ETH account was
hardcoded to 21000. When sending regular ETH to a contract, the gas
limit can be higher, leading to rejected/cancelled transactions.

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -570,14 +570,10 @@ func (account *Account) newTx(
 			}
 		}
 	}
-	gasLimit := uint64(21000) // gas limit for standard ethereum transactions
-	if account.coin.erc20Token != nil {
-		n, err := account.coin.client.EstimateGas(context.TODO(), message)
-		if err != nil {
-			account.log.WithError(err).Error("Could not estimate the gas limit.")
-			return nil, errp.WithStack(errors.ErrInvalidData)
-		}
-		gasLimit = n
+	gasLimit, err := account.coin.client.EstimateGas(context.TODO(), message)
+	if err != nil {
+		account.log.WithError(err).Error("Could not estimate the gas limit.")
+		return nil, errp.WithStack(errors.ErrInvalidData)
 	}
 
 	fee := new(big.Int).Mul(new(big.Int).SetUint64(gasLimit), suggestedGasPrice)


x1ddos:
So, this is basically a revert of https://github.com/digitalbitbox/bitbox-wallet-app/pull/833?
What about the issue the original PR was solving?

> It is more robust as the gas estimation api call has been unstable lately.

backend/coins/eth/account.go
```diff
@@ -570,14 +570,10 @@ func (account *Account) newTx(
...
-		n, err := account.coin.client.EstimateGas(context.TODO(), message)
-		if err != nil {
-			account.log.WithError(err).Error("Could not estimate the gas limit.")
-			return nil, errp.WithStack(errors.ErrInvalidData)
-		}
-		gasLimit = n
+	gasLimit, err := account.coin.client.EstimateGas(context.TODO(), message)
+	if err != nil {
+		account.log.WithError(err).Error("Could not estimate the gas limit.")
+		return nil, errp.WithStack(errors.ErrInvalidData)
```
x1ddos, line 576: How about instead of returning the error it would do this:

```go
if err != nil {
    log error
    return nil, err
  }
  gasLimit = uint64(21000) // gas limit for standard ethereum
  // go ahead and try with the standard limit
}
```

I'm just trying to incorporate the idea from the original https://github.com/digitalbitbox/bitbox-wallet-app/pull/833.

benma: The original #833 was a quick (too quick) fix as gas limit estimation by etherscan was broken, for standard eth txs.

Since then, the proper fix has landed (https://github.com/digitalbitbox/bitbox-wallet-app/pull/836), making gas estimation work again for all cases.

Without a gas estimation, one can't just "try" with a lower limit, as that leads to actual funds lost in fees. (eth transactions are executed up to the limit, costing fees, anyway, and then just mined as "rejected" into the chain if more gas was required but ran out).

benma: I didn't revert the quick fix as I thought that saving a network query in the standard case made sense, but I failed to realize in that moment that the gas limit is not always 21000, even in the ETH account. So the optimization was buggy.

x1ddos: :+1: TIL
good to have these details in here.

x1ddos: > The original #833 was a quick (too quick) fix as gas limit estimation by etherscan was broken, for standard eth txs.
> 
> Since then, the proper fix has landed (#836), making gas estimation work again for all cases.

So, what you're saying is #833 should've been reverted in #836?

benma:
Looks like maybe the whole org was flagged by travis ci. Will run CI locally and merge without CI until Travis CI unlocks us again, hopefully soon.

x1ddos:
> Looks like maybe the whole org was flagged by travis ci. Will run CI locally and merge without CI until Travis CI unlocks us again, hopefully soon.

I'm almost done with github workflows. running it at this very moment: https://github.com/digitalbitbox/bitbox02-firmware/runs/745607910

x1ddos:
> > Looks like maybe the whole org was flagged by travis ci. Will run CI locally and merge without CI until Travis CI unlocks us again, hopefully soon.
> 
> I'm almost done with github workflows. running it at this very moment: https://github.com/digitalbitbox/bitbox02-firmware/runs/745607910

That was actually a different repo. Here's one for the app: https://github.com/digitalbitbox/bitbox-wallet-app/pull/848.

backend/coins/eth/account.go
benma: I am rather saying #833 should have never been merged, it never worked properly.

PR #848, opened 2020-06-06
ci: add github workflow

x1ddos:
Alternative to Travis CI.
Runs the same scripts except SSH for nightly builds.

### DIFF
--- a/frontends/qt/Makefile
+++ b/frontends/qt/Makefile
@@ -52,7 +52,7 @@ osx:
 	mv build/BitBox.app build/osx/
 	cp resources/MacOS/Info.plist build/osx/BitBox.app/Contents/
 	cp resources/MacOS/icon.icns build/osx/BitBox.app/Contents/Resources/
-	macdeployqt build/osx/BitBox.app/
+	macdeployqt build/osx/BitBox.app
 	cp server/libserver.so build/osx/BitBox.app/Contents/Frameworks
 	cp build/assets.rcc build/osx/BitBox.app/Contents/MacOS/
 	install_name_tool -change libserver.so @executable_path/../Frameworks/libserver.so build/osx/BitBox.app/Contents/MacOS/BitBox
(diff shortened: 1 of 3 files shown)


frontends/qt/Makefile
```diff
@@ -52,7 +52,7 @@ osx:
 	mv build/BitBox.app build/osx/
 	cp resources/MacOS/Info.plist build/osx/BitBox.app/Contents/
 	cp resources/MacOS/icon.icns build/osx/BitBox.app/Contents/Resources/
-	macdeployqt build/osx/BitBox.app/
```
x1ddos, line 55: The spurious trailing slash was causing macdeployqt fail on github macos runner.
Not sure where it was coming from but it doesn't seem to be required according to https://doc.qt.io/qt-5/macos-deployment.html.

benma:
Can you explain again what is going on with the vendoring?

Cause `make go-vendor` does not change anything when I run it locally.

Is Go 1.14 forcing a dep update?

If so, let's update Go properly in the Dockerfile?

benma:
> If so, let's update Go properly in the Dockerfile?

Tried doing that here (might be independent, but was about time anyway):

https://github.com/digitalbitbox/bitbox-wallet-app/pull/850

x1ddos:
I remember there was a change in Go 1.14 which was not backported to 1.13. The change made `go mod vendor` add all transitive deps where 1.13 would sometimes skip it. This results in 1.14 arriving at different deps versions where 1.13 just wouldn't care and skip. This is my understanding. It's buried somewhere in https://github.com/golang/go/issues/33848#issuecomment-536088704. Release notes https://golang.org/doc/go1.14#tools don't mention this inconsistency either.

> Cause make go-vendor does not change anything when I run it locally.

Yeah, I would expect exactly that on 1.13.

> Is Go 1.14 forcing a dep update?

Yes. It complains otherwise,

> If so, let's update Go properly in the Dockerfile?

Right, it's independent. I would expect updated deps to work with both 1.13 and 1.14.

x1ddos:
I'll wait for #850 and then rebase. Feels like deps sync belongs more to that than this PR.

x1ddos:
Ok. Removed the sync deps commit but forgot to sign.

PR #853, opened 2020-06-09
backend/account: rename Initialized() -> Synced()

benma:
Because that's what it really means. This is to disambiguate with the
Initialize() function and the `alreadyInitialized` var used inside,
which are unrelated.

### DIFF
--- a/backend/accounts/account.go
+++ b/backend/accounts/account.go
@@ -38,9 +38,10 @@ type Interface interface {
 	Coin() coin.Coin
 	// Name returns a human readable long name.
 	Name() string
-	// Initialize only starts the initialization, the account is not initialized right afterwards.
+	// Initialize only starts the synchronization, the account is not synced right afterwards.
 	Initialize() error
-	Initialized() bool
+	// Synced indicates whether the account has loaded and finished the initial sync.
+	Synced() bool
 	Offline() bool
 	FatalError() bool
 	Close()
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -88,7 +88,9 @@ type Account struct {
 
 	feeTargets []*FeeTarget
 
-	initialized bool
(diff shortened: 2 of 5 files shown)


backend/accounts/account.go
```diff
@@ -40,7 +40,8 @@ type Interface interface {
 	Name() string
 	// Initialize only starts the initialization, the account is not initialized right afterwards.
```
x1ddos, line 41: does the "initialization" also means "synchronization" here? if so, I'd add it to the comment:

> // Initialize starts account sync and returns immediately without waiting for completion

x1ddos:
Watch out for errors. I think other fields need to be updated:

```
# github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc_test [github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc.test]

backend/coins/btc/account_test.go:75:26: account.Initialized undefined (type *btc.Account has no field or method Initialized)

backend/coins/btc/account_test.go:77:25: account.Initialized undefined (type *btc.Account has no field or method Initialized)
```

backend/accounts/account.go
benma: Good catch. Changed the words.

benma:
> Watch out for errors. I think other fields need to be updated:
> 
> ```
> # github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc_test [github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc.test]
> 
> backend/coins/btc/account_test.go:75:26: account.Initialized undefined (type *btc.Account has no field or method Initialized)
> 
> backend/coins/btc/account_test.go:77:25: account.Initialized undefined (type *btc.Account has no field or method Initialized)
> ```

Yup, fixed.

x1ddos:
:+1:

PR #855, opened 2020-06-09
backend/keystore: don't pass full signing config to CanVerifyAddress

benma:
Only the multisig flag is needed (and only in the bitbox01).

Multisig in the app is a simple demo of p2sh-multisig with multiple
bitbox01, so it is technical debt in a sense.

When combining accounts, it is currently simpler to assume that a
device can verify the addresses of all supported types (see
SupportsAccount()), which is true for all supported devices and
account types. Multisig is the only exception, where the bitbox01 can
sign it, but the app does not implement verifying the addresses on the
2fa app.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -752,8 +752,7 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
 	if address == nil {
 		return false, errp.New("unknown address not found")
 	}
-	canVerifyAddress, _, err := account.Keystores().CanVerifyAddresses(
-		account.signingConfiguration.Multisig(), account.Coin())
+	canVerifyAddress, _, err := account.Keystores().CanVerifyAddresses(account.Coin())
 	if err != nil {
 		return false, err
 	}
@@ -768,7 +767,7 @@ func (account *Account) CanVerifyAddresses() (bool, bool, error) {
 	if account.signingConfiguration == nil {
 		return false, false, errp.New("account must be initialized")
 	}
-	return account.Keystores().CanVerifyAddresses(account.signingConfiguration.Multisig(), account.Coin())
+	return account.Keystores().CanVerifyAddresses(account.Coin())
 }
 
 // Keystores returns the keystores of the account.
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -752,7 +752,8 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
 	if address == nil {
 		return false, errp.New("unknown address not found")
 	}
-	canVerifyAddress, _, err := account.Keystores().CanVerifyAddresses(account.signingConfiguration, account.Coin())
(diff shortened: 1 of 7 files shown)


benma:
On the fence on whether we should just delete all the multisig stuff. It's been half baked and unused, the only merit in keeping it is for reference if we ever want to support multisig in the app.

x1ddos:
> the only merit in keeping it is for reference if we ever want to support multisig in the app.

we have git history for this, right? delete the code, make a bug and point to the last commit where the reference code still existed?

backend/keystore/keystore.go
```diff
@@ -53,8 +53,8 @@ type Keystore interface {
 	// This is typically done through a screen on the device or through a paired mobile phone.
 	// optional is true if the user can skip verification, and false if they should be incentivized
```
x1ddos, line 54: I wonder what "incentivized" means exactly. I'd keep it simpler: "optional is true if the user can skip verification." and that's it - it's either optional or non-optional.

backend/devices/bitbox02/keystore.go
```diff
@@ -74,7 +74,7 @@ func (keystore *keystore) SupportsAccount(
 }
 
 // CanVerifyAddress implements keystore.Keystore.
-func (keystore *keystore) CanVerifyAddress(configuration *signing.Configuration, coin coinpkg.Coin) (bool, bool, error) {
+func (keystore *keystore) CanVerifyAddress(multisig bool, coin coinpkg.Coin) (bool, bool, error) {
 	optional := false
```
x1ddos, line 78: `const optional = false` is better imho. here and in the other places.

backend/coins/btc/account.go
```diff
@@ -752,7 +752,8 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
 	if address == nil {
 		return false, errp.New("unknown address not found")
 	}
-	canVerifyAddress, _, err := account.Keystores().CanVerifyAddresses(account.signingConfiguration, account.Coin())
+	canVerifyAddress, _, err := account.Keystores().CanVerifyAddresses(
+		account.signingConfiguration.Multisig(), account.Coin())
```
x1ddos, line 756: How about this:

```go
!account.signingConfiguration.Multisig() && account.Keystores().CanVerifyAddresses(account.Coin())
```

benma: that would preclude e.g. the bitbox02 from verifying multisig addresses.

x1ddos: it just the multisig bool arg is either unused or always negated in various implementations of CanVerifyAddresse. feels a bit wasteful.

benma: Found a simpler solution: disable multisig for the bb01. If needed, the bb02 would be a better device for multisig anyway, where this arg would not be needed.

x1ddos: :+1: nice

PR #856, opened 2020-06-10
backend/coins/btc: simplify code

benma:
Directly pass the address chain, skip indirection via the
`addresses()` method.


benma:
> LGTM. I feel like we're on a good wave of code simplification. <3

all for the sake of easy to review commits later :)

PR #857, opened 2020-06-10
backend/coins/btc: simplify GetUnusedReceiveAddresses()

benma:
No need to special case address based accounts here, as the
addresschain already contains the relevant special case. The
`SingleAddress` address chain already only returns only one address.

Furthermore, the `[:receiveAddressesLimit]` had potential to panic if
there were too little addresses. It relied on the hidden assumption
that the gap limit is always at least 20 (`receiveAddressesLimit`).


backend/coins/btc/account.go
```diff
@@ -719,12 +719,13 @@ func (account *Account) GetUnusedReceiveAddresses() []accounts.Address {
 	defer account.RLock()()
 	account.log.Debug("Get unused receive address")
 	addresses := []accounts.Address{}
```
x1ddos, line 721: `var addresses []accounts.Address` usually reads nicer.

PR #861, opened 2020-06-11
backend/btc: export function to create a pubkey script hash

benma:
Code de-duplication, and it will be used from another package to get
the input configuration when creating a new tx.


backend/coins/btc/blockchain/blockchain.go
```diff
@@ -80,6 +80,11 @@ func (history TxHistory) Status() string {
 // ScriptHashHex is the hash of a pkScript in reverse hex format. Always 64 chars.
 type ScriptHashHex string
 
+// NewScriptHashHex creates the hash of an output's pubkeyScript.
```
x1ddos, line 83: Is it only an output, though? what about a receiving address.

benma: removed `output's`.

PR #862, opened 2020-06-12
backend/bitbox01: remove singlesig check

benma:
Since f9118872, multisig was disabled on the bb01, so this check does
nothing.

### DIFF
--- a/backend/devices/bitbox/device.go
+++ b/backend/devices/bitbox/device.go
@@ -1036,13 +1036,11 @@ func (dbb *Device) signBatch(
 
 		if txProposal.ChangeAddress != nil {
 			configuration := txProposal.ChangeAddress.Configuration
-			if configuration.Singlesig() {
-				publicKey := configuration.PublicKeys()[0]
-				command["sign"]["checkpub"] = []map[string]interface{}{{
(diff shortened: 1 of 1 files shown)


backend/devices/bitbox/device.go
```diff
@@ -1053,7 +1053,7 @@ func (dbb *Device) signBatch(
 	}
 
 	mobchan := dbb.mobileChannel()
-	if txProposal != nil && paired && txProposal.AccountConfiguration.Singlesig() && mobchan != nil {
+	if txProposal != nil && paired && mobchan != nil {
```
x1ddos, line 1056: What about the lines above?

```go
if txProposal.ChangeAddress != nil {
			configuration := txProposal.ChangeAddress.Configuration
			if configuration.Singlesig() {
				publicKey := configuration.PublicKeys()[0]
				command["sign"]["checkpub"] = []map[string]interface{}{{
					"pubkey":  hex.EncodeToString(publicKey.SerializeCompressed()),
					"keypath": configuration.AbsoluteKeypath().Encode(),
				}}
			}
		}
```

benma: good catch, fixed

PR #863, opened 2020-06-15
Integrate paybtc.au.com

thisconnect:
Add https://paybtc.au.com to the list of exchanges


x1ddos:
@thisconnect rebase and you should see CI green.

PR #866, opened 2020-06-15
ci: don't run envinit on macOS

x1ddos:
The CI on macOS requires no tools installed in envinit make target.
This speeds it up a bit and avoids failures when upstream moves tools
source code around because we install with "go get".

Ideally, to fix the latter, we would install tools at specific release
versions in GO111MODULE=on mode. While this would probably work for
most, gomobile doesn't support "go build -mod=vendor" and regardless,
Go modules seem to make this more complicated and confusing. Also,
see https://github.com/golang/go/issues/25922.

### DIFF
--- a/scripts/travis-ci.sh
+++ b/scripts/travis-ci.sh
@@ -36,7 +36,6 @@ if [ "$TRAVIS_OS_NAME" == "osx" ]; then
     # here and pop in the end.
     pushd ../ && cp -a bitbox-wallet-app $GOPATH/src/github.com/digitalbitbox/
     cd $GOPATH/src/github.com/digitalbitbox/bitbox-wallet-app/
-    make envinit
     make qt-osx
     popd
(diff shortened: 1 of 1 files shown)


x1ddos:
> we would install tools at specific release versions in GO111MODULE=on mode

But then they would get into go.mod and go.sum. I tried it last night, followed by `go mod tidy` to remove from go.mod after "go get" but it of course bumped other indirect import versions and go.mod didn't get back to a clean state.

Another alternative is similar to what's described in golang/go#25922 but it really sucks:

```sh
cd $(mktemp -d)
export GOBIN=$PWD/bin
go get ...
mv $GOBIN/* ...to where?
```

:see_no_evil:

x1ddos:
Plus, commands like `cd $(mktemp -d)` suddenly makes it platform dependent.

PR #867, opened 2020-06-16
BB02 v9.0.0 compat

benma:


x1ddos:
3m23s now compared to 5m31s in https://github.com/digitalbitbox/bitbox-wallet-app/runs/764817549. This is after removal of envinit from CI on macOS.

PR #868, opened 2020-06-16
backend: remove multisig from default account init

benma:
It was not functional anyway since we disabled multisig in the bb01 in
the app, and would need some love to re-enable for bb02, but that is
not necessary now.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -575,22 +575,19 @@ func (backend *Backend) initDefaultAccounts() {
 	if backend.keystores.Count() == 0 {
 		return
 	}
+	if backend.keystores.Count() > 1 {
+		// If needed, insert multisig account initialization here based on multiple connected
+		// keystores.
+		return
+	}
 	if backend.arguments.Testing() {
-		switch {
-		case backend.arguments.Multisig():
-			TBTC, _ := backend.Coin(coinpkg.CodeTBTC)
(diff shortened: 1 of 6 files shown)


backend/backend.go
```diff
@@ -575,22 +575,19 @@ func (backend *Backend) initDefaultAccounts() {
...
 		return
 	}
+	if backend.keystores.Count() > 1 {
+		// If needed, insert multisig account initialization here based on multiple connected
+		// keystores.
+		return
+	}
 	if backend.arguments.Testing() {
-		switch {
-		case backend.arguments.Multisig():
```
x1ddos, line 580: Might be confusing now if I provide the multisig arg in https://github.com/digitalbitbox/bitbox-wallet-app/blob/b8075014/backend/arguments/arguments.go#L52 and expect it to do something but it will be just ignored.

How about exiting with an error if multisig arg is provided or simply remove it from backend.arguments for now?

benma: Removed from arguments - returning fixed `false` for now. Otherwise the rabbit whole does not end and all of multisig will be removed, which is a *lot* :sweat_smile:

benma: It is not ignored btw, you can still unlock two instead of only one account - just no accounts show up that can make use of them :P

x1ddos: I think it's good enough for me :)

PR #869, opened 2020-06-16
backend/btc: bundle address chains and signing config in one struct

benma:
In preparation for unifying account types into one account.

I already make this a slice, but with only one element, so it behaves
the same as before.

When unifying e.g. Bitcoin Legacy, Segwit and Native Segwit accounts,
there would be one entry per such signing configuration.

Each accesss is marked with:

// TODO unified-accounts

with each following commit fixing one of those. This hopefully results
in smaller and easier to review commits.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -55,6 +55,12 @@ const (
 	maxGapLimit = 2000
 )
 
+type subaccount struct {
+	signingConfiguration *signing.Configuration
+	receiveAddresses     AddressChain
+	changeAddresses      AddressChain
+}
+
 // Account is a account whose addresses are derived from an xpub.
 type Account struct {
 	locker.Locker
@@ -69,14 +75,12 @@ type Account struct {
 	db                      transactions.DBInterface
 	forceGapLimits          *types.GapLimits
 	getSigningConfiguration func() (*signing.Configuration, error)
-	signingConfiguration    *signing.Configuration
 	keystores               *keystore.Keystores
 	getNotifier             func(*signing.Configuration) accounts.Notifier
 	notifier                accounts.Notifier
 	blockchain              blockchain.Interface
 
-	receiveAddresses AddressChain
-	changeAddresses  AddressChain
+	subaccounts []subaccount
 
 	transactions *transactions.Transactions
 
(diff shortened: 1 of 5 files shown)


x1ddos:
Makes sense. Once test fails though:

```
# github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/maketx_test [github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/maketx.test]
backend/coins/btc/maketx/maketx_test.go:96:21: too many arguments in call to maketx.NewTx
	have (*btc.Coin, *signing.Configuration, map[wire.OutPoint]maketx.UTXO, *wire.TxOut, btcutil.Amount, *addresses.AccountAddress, *logrus.Entry)
	want (coin.Coin, map[wire.OutPoint]maketx.UTXO, *wire.TxOut, btcutil.Amount, *addresses.AccountAddress, *logrus.Entry)
FAIL	github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/maketx [build failed]
```

x1ddos:
Why did appveyor pass? :sweat_smile:

benma:
> Why did appveyor pass? sweat_smile

Cause it only does the build for windows, no unit tests (those only run in linux; no need to run them on every platform).

backend/coins/btc/account.go
```diff
@@ -69,14 +75,12 @@ type Account struct {
...
 	getSigningConfiguration func() (*signing.Configuration, error)
-	signingConfiguration    *signing.Configuration
 	keystores               *keystore.Keystores
 	getNotifier             func(*signing.Configuration) accounts.Notifier
 	notifier                accounts.Notifier
 	blockchain              blockchain.Interface
 
-	receiveAddresses AddressChain
-	changeAddresses  AddressChain
+	subaccounts []subaccount
```
x1ddos, line 83: Would a map instead of a slice make things easier? Keyed by ScriptType.

benma: Looking at my branch with the rest of the commits, a slice seems to work fine. There is only one lookup by script type needed, but that is in the bb02 keystore, so at best I might convert the signing configs in the tx proposal to a map, but not here.

if we need lookups in the future, changing it to a map should be very simple (iterating over them works the same way).

x1ddos: :+1:

Issue #715, opened 2019-11-28
configuration hash can't rely on json.Marshal


benma:
Discussed with @x1ddos 

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/signing/configuration.go#L230-L233

`json.Marshal` is not guaranteed to create the same serialization always. The order of the keys could change, or whitespacing, or other things.

The Hash() should be replaced with a hand rolled hash accumulator over the fields.

Account database filenames depend on this hash, so the new impl. must either be compatible, or think about the consequence of creating new account databases (at the moment this merely means re-downloading the transactions, and get desktop notifications for the transactions again).

x1ddos:
In addition to that, [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt) explicitly defines JSON objects as unordered collections. It works today just because of an implementation detail of the specific package in use.

x1ddos:
Resolved by #1267 and https://github.com/digitalbitbox/bitbox-wallet-app/pull/1314.

PR #870, opened 2020-06-16
backend/signing/configuration: hash over slice of configs

benma:
Will be used as the new account db identifier when combining multiple
signing configs in one account (mixed inputs).

### DIFF
--- a/backend/signing/configuration.go
+++ b/backend/signing/configuration.go
@@ -1,4 +1,5 @@
 // Copyright 2018 Shift Devices AG
+// Copyright 2020 Shift Crypto AG
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -226,7 +227,7 @@ func (configuration *Configuration) UnmarshalJSON(bytes []byte) error {
 	return nil
 }
 
-// Hash returns a hash of the configuration.
+// Hash returns a hash of the configuration in hex format.
 func (configuration *Configuration) Hash() string {
 	hash := sha256.Sum256(jsonp.MustMarshal(configuration))
 	return hex.EncodeToString(hash[:])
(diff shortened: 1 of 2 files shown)


benma:
The hash of a single config is yet to be changed: https://github.com/digitalbitbox/bitbox-wallet-app/issues/715

backend/signing/configuration.go
```diff
@@ -240,3 +240,22 @@ func (configuration *Configuration) String() string {
...
 }
+
+// Configurations is a collection of configurations.
+type Configurations []*Configuration
+
+// Hash returns a hash of all configurations in hex format. It is defined as
+// `sha256(<32 bytes hash 1>|<32 bytes hash 2>|...)`.
+func (configs Configurations) Hash() string {
+	h := sha256.New()
+	for _, cfg := range configs {
```
x1ddos, line 251: This makes the resulting hash order-dependent: `[]Configurations{a, b}` and `[]Configurations{b, a}` is the same functionality during runtime, right? But the hash is different.

It just seems a bit fragile.

benma: Good point. I commented that the slice is unordered and now sort the individual hashes before hashing again. Added test to make sure the ordering does not matter.

backend/signing/configuration_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcutil/hdkeychain"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfigurationsHash(t *testing.T) {
+	xpub, err := hdkeychain.NewMaster(make([]byte, 32), &chaincfg.TestNet3Params)
+	require.NoError(t, err)
```
benma, line 27: i want a `.unwrap()` in Go that just panics :)

x1ddos: Make a proposal! :sweat_smile:

backend/signing/configuration_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcutil/hdkeychain"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfigurationsHash(t *testing.T) {
```
x1ddos, line 25: Nice. Could you also add a Benchmark test? I'm slightly worried about time complexity. Just to make sure it doesn't take too long to compute the hash.

https://golang.org/pkg/testing/#hdr-Benchmarks

benma: uuuh seriously? Sounds like premature optimization. The hash function is gonna be called once per account, with at most three configs in the list. There is no way you could ever notice this, right?

x1ddos: But there's no optimization. It's just a test.

x1ddos: I was just curious to see ops/sec numbers, that's all.

x1ddos: Don't you like benchmarks? They are awesome! :)

x1ddos: Ok, I can live without the benchmark, I guess, but would've been happier with it.

benma: I mean I do like them, but not spending time on one if there is no need to make it faster :)

PR #871, opened 2020-06-16
backend/config,frontend/settings: toggle Bitcoin, Litecoin

benma:
Not the individual subaccount types.

For combined accounts ("Bitcoin", "Litecoin"), it just toggles the
account.

For split accounts, now either all account types are shown (Bitcoin:
bech32, Bitcoin, Bitcoin Legacy if supported), or none.

This is an acceptable trade-off, and only affects users with the
bitbox01 or users that will manually opt-in to split accounts.

The alternative would be a very hairy and granular settings screen,
which defeats the purpose of trying to simplify it.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1184,12 +1184,7 @@
   },
   "settings": {
     "accounts": {
-      "bitcoinP2PKH": "Bitcoin Legacy",
-      "bitcoinP2WPKH": "Bitcoin: bech32",
-      "bitcoinP2WPKHP2SH": "Bitcoin",
       "ethereum": "Ethereum",
-      "litecoinP2WPKH": "Litecoin: bech32",
-      "litecoinP2WPKHP2SH": "Litecoin",
       "title": "Active accounts"
     },
     "electrum": {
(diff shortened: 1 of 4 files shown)


benma:
So here we disregard the old settings and introduce new ones, which means that BTC/LTC would be enabled in the next release, even if the user previously disabled them.

Not sure if it is worth it to properly migrate the config. Doing so seems like a bit of extra effort, as we have no infrastructure there to migrate configs :(

frontends/web/src/locales/en/app.json
```diff
@@ -1184,12 +1184,7 @@
   },
   "settings": {
     "accounts": {
-      "bitcoinP2PKH": "Bitcoin Legacy",
-      "bitcoinP2WPKH": "Bitcoin: bech32",
-      "bitcoinP2WPKHP2SH": "Bitcoin",
       "ethereum": "Ethereum",
-      "litecoinP2WPKH": "Litecoin: bech32",
-      "litecoinP2WPKHP2SH": "Litecoin",
```
x1ddos, line 1192: Shouldn't this be removed from all the other translations as well?

benma: If I remember correctly, removing needs to be done only in the reference lang (en), then locize-push will delete it for the other langs on locize.

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/docs/i18n.md seems to confirm. I'll try.

benma: Confirmed, next locize-pull will remove them.

x1ddos: thanks!

PR #872, opened 2020-06-17
backend/config: option to force split accounts

benma:
For the time being, we allow the user to go back to one account per
address/script type in the expert settings, like it was before unified
accounts.

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -99,6 +99,10 @@ type Backend struct {
 	LitecoinP2WPKHActive     bool `json:"litecoinP2WPKHActive"`
 	EthereumActive           bool `json:"ethereumActive"`
 
+	// Whether Bitcoin, Litecoin should be shown in multiple accounts - one per script type -
+	// instead of a combined account.
+	SplitAccounts bool `json:"splitAccounts"`
+
 	BTC  btcCoinConfig `json:"btc"`
 	TBTC btcCoinConfig `json:"tbtc"`
 	RBTC btcCoinConfig `json:"rbtc"`
@@ -190,6 +194,8 @@ func NewDefaultAppConfig() AppConfig {
 			LitecoinP2WPKHP2SHActive: true,
 			LitecoinP2WPKHActive:     true,
 			EthereumActive:           true,
+
+			SplitAccounts: false,
 
 			BTC: btcCoinConfig{
 				ElectrumServers: []*ServerInfo{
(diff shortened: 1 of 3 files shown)


backend/config/config.go
```diff
@@ -191,6 +195,8 @@ func NewDefaultAppConfig() AppConfig {
 			LitecoinP2WPKHActive:     true,
 			EthereumActive:           true,
 
+			SplitAccounts: false,
```
x1ddos, line 198: Nicely chosen. This should work for both a new app where no config exists yet and already installed apps.

frontends/web/src/routes/settings/settings.tsx
```diff
@@ -354,6 +354,21 @@ class Settings extends Component<Props, State> {
...
+                                                                    <Badge type="primary">BB02</Badge>
+                                                                    <Badge type="secondary" className="m-left-quarter">BB02-BTC</Badge>
+                                                                    <Badge type="generic" className="m-left-quarter">BTC</Badge>
+                                                                    <Badge type="generic" className="m-left-quarter">LTC</Badge>
+                                                                </p>
+                                                            </div>
+                                                            <Toggle
+                                                                id="splitAccounts"
+                                                                checked={config.backend.splitAccounts}
+                                                                onChange={this.handleToggleAccount} />
```
x1ddos, line 370: How do I find this handler, where is it?

backend/config/config.go
x1ddos: Actually, ignore this. It's only in my head.

frontends/web/src/routes/settings/settings.tsx
thisconnect: https://github.com/digitalbitbox/bitbox-wallet-app/blob/5813f002a37a06cf8ef710add60d263e0f1787cd/frontends/web/src/routes/settings/settings.tsx#L102

x1ddos: oh, it's reusing an existing function. sorry for my ignorance.

benma: Further up in the same class:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/routes/settings/settings.tsx#L102

Calling `POST/GET /config` in backend/handlers.go.

x1ddos: yeah, somehow I thought it had to be a new handler func. not sure why.

PR #873, opened 2020-06-17
frontend/transactions: fix buggy send/receive/self arrows

benma:
The arrows were switched on the translated text, which only works in
English. Of course, the type should be used directly to figure out
what it is.

The translation has been deleted too, unused.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1265,11 +1265,6 @@
     }
   },
   "transaction": {
-    "badge": {
-      "receive": "In",
-      "send": "Out",
-      "send_to_self": "Self"
-    },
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -1265,11 +1265,6 @@
     }
   },
   "transaction": {
-    "badge": {
-      "receive": "In",
-      "send": "Out",
-      "send_to_self": "Self"
-    },
```
thisconnect, line 1272: 👍

thisconnect:
Nice catch, I saw that too yesterday when rewriting transaction into ts 

https://github.com/digitalbitbox/bitbox-wallet-app/pull/874

PR #874, opened 2020-06-17
Convert transactions to typescript

thisconnect:


frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -14,22 +14,53 @@
  * limitations under the License.
  */
 
-import { Component, h } from 'preact';
-import { translate } from 'react-i18next';
-import { FiatConversion } from '../rates/rates';
+import { Component, h, RenderableProps } from 'preact';
+import { translate, TranslateProps } from '../../decorators/translate';
+import { Fee } from '../../routes/account/send/send';
```
thisconnect, line 19: @benma I am not sure about importing the interface from `../../routes/account/send/send'` I renamed it  in send, is that ok?

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -14,22 +14,53 @@
...
 import A from '../anchor/anchor';
-import * as style from './transaction.css';
-import * as parentStyle from './transactions.css';
 import { Dialog } from '../dialog/dialog';
 import { ProgressRing } from '../progressRing/progressRing';
+import { FiatConversion } from '../rates/rates';
+import { Amount } from '../rates/rates';
```
thisconnect, line 24: Also import an interface here, isn't there a naming convention to see that it is an interface, or would you duplicate the interface in this file?

frontends/web/src/components/transactions/transaction.jsx
```diff
@@ -208,42 +140,7 @@ export default class Transaction extends Component {
                         </div>
                         <div className={[parentStyle.action, parentStyle.showOnMedium].join(' ')}>
                             <a href="#" className={style.action} onClick={this.showDetails}>
-                                {
-                                    !transactionDialog ? (
```
benma, line 212: why is there two times code that looks almost identical?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -51,7 +51,7 @@ interface SendProps {
     deviceIDs: string[];
 }
 
-interface ProposedAmount {
+export interface Fee {
```
benma, line 54: `Fee` is not the right name, as it is used below also for non-fee values. What about `Amount`?

frontends/web/src/components/transactions/transaction.tsx
benma: importing interfaces is fine, but commented on the naming in the other comment.

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -14,22 +14,53 @@
...
+    amount: Amount;
+    fee: Fee;
+    feeRatePerKb: Fee;
+    gas: number;
+    vsize: number;
+    size: number;
+    weight: number;
+    numConfirmations: number;
+    numConfirmationsComplete: number;
+    time: string;
```
benma, line 47: this can be also null or undefined (not sure which one) on new tx that have not been verified yet (that's why it's accessed with `time ? ... : ...` below.

frontends/web/src/components/transactions/transaction.tsx
benma: i don't know about conventions, would also be interested. lmk if you find out :)

in typescript i don't even know exactly what the difference is between a type and an interface...

anyway, duplication no good :D

thisconnect: In other places there is a similar situation and then the name is `ComponentnameInterface`
https://github.com/digitalbitbox/bitbox-wallet-app/blob/38d58be846abea9f9d2264df86c3024b0b3c4a9f/frontends/web/src/routes/account/account.tsx#L35

I propose to change it to `AmountInterface` in this case.

thisconnect: random search result on the internet says:

> In TypeScript, type does not create a new name for instance. In TypeScript, an interface can create the new name that can be used everywhere. ... An interface can extend multiple interfaces and class as well. Type is mainly used when a union or tuple type needs to be used.

frontends/web/src/components/transactions/transaction.jsx
thisconnect: The reason for the 2 different "show details" button is the different positioning for the responsive design.  

BTW: I'll get rid of `<a href="#" ` and replace it by button or just add click handler on the arrow.

Screen Shot 2020-06-17 at 1 28 04 PM
Screen Shot 2020-06-17 at 1 36 31 PM

thisconnect: I'd like to leave it for now and merge this PR soon™

thisconnect:
PTAL

frontends/web/src/routes/account/account.tsx
```diff
@@ -72,7 +73,7 @@ class Account extends Component<Props, State> {
     public readonly state: State = {
         initialized: false,
         connected: false,
-        transactions: undefined,
+        transactions: undefined, // TBD: default to empty array: []
```
benma, line 76: comment might be wrong. empty list of transactions is different than not having the transactions list available at all.s 
maybe this doesn't happen though, good to check.

frontends/web/src/routes/account/account.tsx
```diff
@@ -342,8 +342,6 @@ class Account extends Component<Props, State> {
                                         handleExport={this.export}
                                         explorerURL={account.blockExplorerTxPrefix}
                                         transactions={transactions}
-                                        unit={balance!.available.unit}
```
benma, line 345: TS FTW :P

frontends/web/src/routes/account/account.tsx
thisconnect: nono you are correct, but I'd change it everywhere to always be an array and never undefined. I removed the comment and might make a PR for it another time.

frontends/web/src/routes/account/account.tsx
```diff
@@ -58,7 +59,7 @@ export interface AccountInfo {
 interface State {
     initialized: boolean;
     connected: boolean;
-    transactions?: any[]; // define once transaction.jsx is converted
+    transactions?: TransactionInterface[]; // TBD: change to required?
```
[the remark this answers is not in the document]
thisconnect: see other comment

PR #877, opened 2020-06-17
unified-accounts: resolve TODO related to gap limits

benma:
Before:

- Legacy gap limits higher than regular gap limits by default

- User can increase the limits with the `-gapLimitReceive` and
`-gapLimitChange` cli args. They apply to all accounts equally.

After:

The same applies to the subaccounts in a unified account.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -236,6 +236,10 @@ func (account *Account) defaultGapLimits(signingConfiguration *signing.Configura
 // gapLimits gets the gap limits as stored in the account configuration, and defaults to
 // `defaultGapLimits()` if there is no configuration or the configuration limits are smaller than
 // the default limits.
+//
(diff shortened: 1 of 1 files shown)


backend/coins/btc/account.go
```diff
@@ -367,17 +371,15 @@ func (account *Account) Initialize() error {
...
-	if err != nil {
-		return err
-	}
-
-	account.log.Infof("gap limits: receive=%d, change=%d", gapLimits.Receive, gapLimits.Change)
-
 	for _, signingConfiguration := range signingConfigurations {
 		signingConfiguration := signingConfiguration
 
+		gapLimits, err := account.gapLimits(signingConfiguration)
```
x1ddos, line 377: This is used only for the chained addresses. I'd move it in the ` else { ... }` below. Otherwise, it would seem like `gapLimits` is needed for single addresses too.

benma: done

PR #878, opened 2020-06-17
unified-accounts: resolve TODO, syncing addresses

benma:
Sync all subaccount addresses.


x1ddos:
Forgot to mention - double "unified-accounts unified-accounts" in commit msg.

PR #880, opened 2020-06-17
unified-accounts: resolve TODO, pick change

benma:
For now we always use native segwit for change.

If it does not match the input types, it might even be a small privacy
benefit, as change is usually assumed to be the output that matches
the inputs type.

### DIFF
--- a/backend/coins/btc/transaction.go
+++ b/backend/coins/btc/transaction.go
@@ -110,7 +110,7 @@ func (account *Account) newTx(
 			wireUTXO,
 			wire.NewTxOut(parsedAmountInt64, pkScript),
 			*feeTarget.feeRatePerKb,
-			// TODO unified-accounts
+			// Change address is of the first subaccount, always.
 			account.subaccounts[0].changeAddresses.GetUnused()[0],
 			account.log,
 		)
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -126,7 +126,11 @@ const (
 )
 
 // NewAccount creates a new account.
+//
 // forceGaplimits: if not nil, these limits will be used and persisted for future use.
+//
+// getSigningConfigurations: defines the script types used in this account. The first one is used as
+// a default for receive addresses, and always used for change.
 func NewAccount(
 	coin *Coin,
 	dbFolder string,


backend/coins/btc/transaction.go
```diff
@@ -110,7 +110,7 @@ func (account *Account) newTx(
 			wireUTXO,
 			wire.NewTxOut(parsedAmountInt64, pkScript),
 			*feeTarget.feeRatePerKb,
-			// TODO unified-accounts
+			// Change address is of the first subaccount, always.
 			account.subaccounts[0].changeAddresses.GetUnused()[0],
```
x1ddos, line 114: > For now we always use native segwit for change.

But how do you guarantee that here? I think this is where a map might've been more indicative:

```go
accounts.subaccount[signing.ScriptTypeP2WPKH].changeAddresses.GetUnused()[0]
```

At the moment, the order is hardcoded in https://github.com/digitalbitbox/bitbox-wallet-app/blob/c05f9fa9/backend/backend.go#L746-L748, right. But that could change.

benma: It could change, but then you'd be ignoring the function docstring:

```
// getSigningConfigurations: defines the script types used in this account. The first one is used as
// a default for receive addresses, and always used for change.
```

I see it is a bit fragile though. Options

1. keep this and the other PRs as is for now, turn it into a map after?
2. provide another arg to specify the default - makes it explicit
3. do nothing

would avoid option 4 which is to turn it into a map straight away and fix up all other PRs :innocent:

benma: actuaaaly:

"here" is the wrong place to define the default, as the account does not even know what kind of signing configs it may have. the caller (in this case `backend`) must decide the default somehow. in this case, an index into the slice vs a key into the map provided as a second arg sound mostly equivalent. in any case, option 1) above is dismissed based on this, right?

x1ddos: This would also be ok according to the docstring:

```go
backend.createAndAddBTCAccount(keystore, BTC,
	"btc", "Bitcoin",
	[]scriptTypeWithKeypath{
		newScriptTypeWithKeypath(signing.ScriptTypeP2PKH, "m/44'/0'/0'"),
		newScriptTypeWithKeypath(signing.ScriptTypeP2WPKH, "m/84'/0'/0'"),
		newScriptTypeWithKeypath(signing.ScriptTypeP2WPKHP2SH, "m/49'/0'/0'"),
	},
)
```

And yet the assumption: 

> For now we always use native segwit for change.
> ... it might even be a small privacy benefit

would no longer be valid.

Yeah, an index for which `accounts.subaccount` to pick would also work, I guess.

This is a separate unified-accounts branch anyway. I think this can wait until other PRs are merged, as long as we don't forget in the end.

benma: > And yet the assumption:
> 
> > For now we always use native segwit for change.
> > ... it might even be a small privacy benefit
> 
> would no longer be valid.

sure, but this is the commit message, and this commit does fulful it. But I agree it's brittle, will fix it.

benma: not sure how though, e.g. the index/key specified as default might not even make it to the account instead, as it might not be supported by the keystore. the current way is like a priority list with the first supported one being the default. 

let's discuss it offline and think of how to solve it in a separate PR, ok?

x1ddos: sounds good

PR #883, opened 2020-06-17
unified-accounts: resolve TODO, receive screen

benma:
The backend now sends unused addresses of each subaccount to the
frontend.

Frontend now only shows, as before, the addresses from the first
subaccount. Work needs to be done to allow the user to cycle through
the address types (see TODO).

### DIFF
--- a/backend/accounts/account.go
+++ b/backend/accounts/account.go
@@ -24,6 +24,9 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/signing"
 )
 
+// AddressList is a list of addresses.
+type AddressList []Address
+
 // Interface is the API of a Account.
 type Interface interface {
 	Info() *Info
@@ -53,7 +56,10 @@ type Interface interface {
 	FeeTargets() ([]FeeTarget, FeeTargetCode)
 	TxProposal(string, coin.SendAmount, FeeTargetCode, map[wire.OutPoint]struct{}, []byte) (
 		coin.Amount, coin.Amount, coin.Amount, error)
-	GetUnusedReceiveAddresses() []Address
+	// GetUnusedReceiveAddresses gets a list of list of receive addresses. The result can be one
+	// list of addresses, or if there are multiple types of addresses (e.g. `bc1...` vs `3...`), a
+	// list of lists.
+	GetUnusedReceiveAddresses() []AddressList
 	CanVerifyAddresses() (bool, bool, error)
 	VerifyAddress(addressID string) (bool, error)
 	Keystores() *keystore.Keystores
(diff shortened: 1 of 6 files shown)


backend/coins/btc/exchange.go
```diff
@@ -32,10 +32,10 @@ func (account *Account) SafelloBuySupported() bool {
 func (account *Account) SafelloBuy() *safello.Buy {
 	switch account.coin.Net().Net {
 	case chaincfg.MainNetParams.Net:
-		address := account.GetUnusedReceiveAddresses()[0]
+		address := account.GetUnusedReceiveAddresses()[0].Addresses[0]
```
x1ddos, line 35: Why are we still keeping Safello?

backend/accounts/account.go
```diff
@@ -24,6 +24,11 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/signing"
 )
 
+// Addresses is a list of addresses.
+type Addresses struct {
+	Addresses []Address
```
x1ddos, line 29: How about this:

```go
type AddressList []Address
GetUnusedReceiveAddresses() []AddressList

unused := make([]accounts.AddressList, len(account.subaccounts))
for i, sub := range account.subaccounts {
    for n, addr := range sub.receiveAddresses.GetUnused() {
        if n >= receiveAddressesLimit {
            // Limit receive addresses gap to avoid longer scans or "lost" funds during a recovery.
            break
        }
        unused[i] = append(unused[i], addr)
    }
}
```

x1ddos: It's just a bit hard to read Addresses[i].Addresses.Addresses.Addresses...

backend/coins/btc/exchange.go
benma: they promised to restore service by Q1 2020, so we made an issue to wait a bit and remove it if that didn't happen. obviously it didn't; i'll reach out quickly to see if there is any hope left.

backend/accounts/account.go
benma: I made it a struct so we can add more info to the struct if the frontend needs it, such as the name to show for the address type, or script type identifier, or sth like that.

premature?

x1ddos: IMHO a bit premature, yeah.

benma: :+1: done

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -167,7 +167,10 @@ class Receive extends Component<Props, State> {
         const forceVerification = secureOutput.hasSecureOutput && !secureOutput.optional;
         const enableCopy = !forceVerification;
 
-        let address = receiveAddresses[activeIndex].address;
+        // TODO: unified-accounts, let user cycle through address types
+        const currentAddresses = receiveAddresses[0];
```
x1ddos, line 171: I guess @thisconnect will take care of this.

thisconnect: 👍   yes, when it is merged into `staging-unified-accounts`

PR #884, opened 2020-06-19
frontend: remove t(settings.accounts.ethereum)

benma:
No point in translating a constant name.


benma:
image


:see_no_evil: @jstrnbrg

benma:
>  but looks like a hickup in yarn registry, I'd assume if you run CI again it will pass

Travis passed, good thing we have redundant CI now.

PR #886, opened 2020-06-19
backend/keystore: simplify VerifyExtendedPublicKey() signature

benma:
keypath is directly derived from configuratio and redundant.


benma:
> LGTM. Any idea why keypath arg was there in the first place? I dug through history a bit and it seem this was originally called `VerifyBTCPub`, added in [3f72de7](https://github.com/digitalbitbox/bitbox-wallet-app/commit/3f72de75).

No idea. The keypath was already useless in the commit you linked.

PR #888, opened 2020-06-23

benma:


x1ddos:
> LGTM

Well, assuming CI pass but why wouldn't they.

benma:
> > LGTM
> 
> Well, assuming CI pass but why wouldn't they.

no need to even mention that, CI passing is a given ;)

PR #891, opened 2020-06-24
backend/coins/btc: enable RBF for BTC transactions

benma:
Litecoin does not have RBF.

We opt-in to RBF to mitigate transactions that might get stuck in
periods of high fees (bull market), or if the user enters a fee that
is too low (when we add the feature to manually choose fees).

The app does not support fee bumping yet, but in emergencies, users
can use Electrum+BitBox02 and bump the fee there.


benma:
Can only be merged if https://github.com/digitalbitbox/bitbox02-firmware/pull/513 is also merged.

PR #894, opened 2020-06-26
frontend: suppress dialog animations when requested

x1ddos:
Popup animations prevent dialogs from showing up at all, at least on
Android devices where animations are disabled.

Being able to completely turn off animations is an accessibility
feature in most phones these days.

To reproduce the current issue on Android, go to Accessibility settings
and enable "Remove animations". Close the app completely and reopen.
Navigate to Settings, Connect your own full node, tap on "Check" for
any row in the Servers list. You'll see nothing and unable to scroll
up/down anymore. Expected: a dialog with the check status.

Trial and error has shown that Android WebView adheres to "Disable
Animations" accessibility setting and honors prefers-reduced-motion
@media CSS query, in which we suppress dialog animations.

See
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
for more details about the @media query.

### DIFF
--- a/frontends/web/src/components/dialog/dialog.css
+++ b/frontends/web/src/components/dialog/dialog.css
@@ -236,3 +236,16 @@
         padding: var(--space-half);
     }
 }
+
+/**
+ * Turn off animations when requested.
+ * On an Android phone, this is typically a "Remove animations" switch
+ * in Accessibility section.
+ */
+@media (prefers-reduced-motion) {
+    .overlay,
+    .modal {
+        animation: none;
+        opacity: 1;
+    }
+}
--- a/frontends/web/src/components/steps/steps.css
+++ b/frontends/web/src/components/steps/steps.css
@@ -406,3 +406,15 @@
         margin-bottom: var(--space-default) !important;
     }
 }
+
+/**
+ * Turn off animations when requested.
+ * On an Android phone, this is typically a "Remove animations" switch
+ * in Accessibility section.
+ */
+@media (prefers-reduced-motion) {
+    .step {
+        animation: none;
+        transition: none;
+    }
+}
--- a/frontends/web/src/utils/animation.ts
+++ b/frontends/web/src/utils/animation.ts
@@ -1,8 +1,29 @@
+import { debug } from './env';
+
+const mq = window.matchMedia('(prefers-reduced-motion: reduce)');
+let reduceMotion = mq.matches;
+try {
+    // experimental https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList
+    // @ts-ignore
+    mq.addListener(e => reduceMotion = e.matches);
+} catch (e) {
+    if (debug) {
+        // tslint:disable-next-line: no-console
+        console.log(e);
+    }
+}
+
 export function animate(
     el: Element,
     effect: string,
     callback?: () => void,
 ) {
+    if (reduceMotion) {
+        if (callback) {
+            callback();
+        }
+        return;
+    }
     el.addEventListener('animationend', onAnimationEnd, false);
     function onAnimationEnd() {
         el.classList.remove('animated', 'faster', effect);


thisconnect:
LGTM
I tested dc1918e6bea2b8faea90cac90ed6025c100476d9 
on electrum server "check" and bitboxbase "connect manually" popup/modal, there is still an "in" animation,
maybe these styles that you removed were from an older verison before animate.css?

👍

thisconnect:
I just noticed after moving back to master that the "in" animation does flicker with your change. :(

thisconnect:
could you try to overwrite the styles as in the example here https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
```css
@media (prefers-reduced-motion) {
  .overlay,
  .modal {
    animation-name: dissolve;
    opacity: 1;
  }
}
```

🤞 that this is supported…

x1ddos:
> could you try to overwrite the styles as in the example here https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
> 
> ```css
> @media (prefers-reduced-motion) {
>   .overlay,
>   .modal {
>     animation-name: dissolve;
>     opacity: 1;
>   }
> }
> ```
> 
> crossed_fingers that this is supported…

OMG this actually works!

x1ddos:
PTAL. Switched to `prefers-reduced-motion` CSS media query but, instead of using a different animation type, I turn it off completely. The media query is too coarse grained. If a user requests reduced animations, it isn't actually known how much a "reduced" is, so I figured safest is to turn them off.

x1ddos:
Actually, "suppress" is a better word here.

benma:
> flick

same here, definitely flickers ;)

thisconnect:
> > flick
> 
> same here, definitely flickers ;)

Where & how?

Does it flicker if you enable reduce montion?

This change should have no impact if you run in "normal" mode, and if the media query is not supported it will be just ignored.

The flickering is gone for me when testing fbb51f844a11c453d043d6a77ce3cb7efe0d87b3 BitBoxBase / Connect manually or Settings / Enable tor proxy (as this 2 has the nicest in+out animation)
 
Connect your own fullnode / Download remote certificate doesn't have the out animation but it was already like that on master

trying to test now on macOS with reduced motion enabled…

thisconnect:
Tested with reduced motion

- with `x1ddos:drop-dialog-animations` in the brower I can't close the dialogs
- with `master` the dialogs wont even open
- with last release (v4.19.1) real app NO issue!
- with last tagged v4.19.1 in the browser wont open

Looks like reduced motion is ignored in QT webengine on macOS, that is why it works and there are no issues :)

I would still like to PR a fix for it.
I suspect that the onanimation end event is not fired. I try to make a fix and open a PR into x1ddos:drop-dialog-animations

thisconnect:
@x1ddos I opened a PR that fixes this:

> with x1ddos:drop-dialog-animations in the brower I can't close the dialogs

x1ddos:
Could somebody try this on qt-macos?

x1ddos:
> Could somebody try this on qt-macos?

I mean now with https://github.com/digitalbitbox/bitbox-wallet-app/pull/894/commits/64b12ad55d3a89839d6cda4fc5d1b71c48959a02 from @thisconnect.

thisconnect:
> > Could somebody try this on qt-macos?
> 
> I mean now with [64b12ad](https://github.com/digitalbitbox/bitbox-wallet-app/commit/64b12ad55d3a89839d6cda4fc5d1b71c48959a02) from @thisconnect.

built of 64b12ad55d3a89839d6cda4fc5d1b71c48959a02 works fine, dialogs open and can be closed, the animation is always on even with reduced motion enabled, but this was expected as this is a newer API and the QT webengine doesn't support it.

LGTM

PR #895, opened 2020-06-29
frontend: make locize-pull

benma:


frontends/web/src/locales/ja/app.json
```diff
@@ -941,11 +941,6 @@
   },
   "settings": {
     "accounts": {
-      "bitcoinP2PKH": "Bitcoin Legacy",
-      "bitcoinP2WPKH": "Bitcoin: bech32",
-      "bitcoinP2WPKHP2SH": "Bitcoin",
```
thisconnect, line 946: Interesting the other languages had `"ethereum": "Ethereum",` here, I assume it never made it into japanese translation

PR #896, opened 2020-06-29
frontend/settings: move legacy, bech32, segwit guide entries

benma:
Since unified accounts. there are no settings mentioning those words.

We ove it to the accounts guide, but only in split accounts, showing
the relevant part.

We might delete the entries soon altogether, as there is also the
accounts description banner that says more or less the same. Keeping
it for now though to not lose all the translations, in case we need
it.

Litecoin currently does not show them, as the text has 'Bitcoin' all
over. Not gonna fix this either.

### DIFF
--- a/frontends/web/src/routes/account/info/signingconfiguration.tsx
+++ b/frontends/web/src/routes/account/info/signingconfiguration.tsx
@@ -29,8 +29,10 @@ interface ProvidedProps {
     signingConfigIndex: number;
 }
 
+export type ScriptType = 'p2pkh' | 'p2wpkh-p2sh' | 'p2wpkh';
+
 export interface SigningConfigurationInterface {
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/account/info/signingconfiguration.tsx
```diff
@@ -29,8 +29,10 @@ interface ProvidedProps {
     signingConfigIndex: number;
 }
 
+export type ScriptType = 'p2pkh' | 'p2wpkh-p2sh' | 'p2wpkh';
+
 export interface SigningConfigurationInterface {
-    scriptType: 'p2pkh' | 'p2wpkh-p2sh' | 'p2pkh';
```
thisconnect, line 33: `p2wpkh` was missing and `p2pkh` twice :|

PR #899, opened 2020-06-30
ci: speed bump by running builds in parallel

x1ddos:
This cuts overall run to about a half: 17min vs 9min on average.
And the test/lint job typically requires only ~5min.

Because Docker Hub may rate limit "docker pulls", use alternative image
stored in GitHub's docker.pkg.github.com registry.

https://docs.docker.com/docker-hub/download-rate-limit/ specifies that

> Docker will gradually impose download rate limits with an eventual
> limit of 300 downloads per six hours for anonymous users.

With the current CI image of 13 layers, it would limit us to 1 PR per
hour at the least. Add some typo corrections which trigger CI builds
and we may hit rate limiting quite fast.

Plus, it's always good to have a separate copy of the CI image for
times when DockerHub is inaccessible.

See ci.yml in .github/workflows for how to update this image in
the GitHub's registry.

On Travis CI however jobs are still run serially. I believe they have a
limitation of 1 "concurrent" job for free open-source projects. So, it
wouldn't help splitting CI tasks into separate jobs.

Also, some nits:
- removed sudo from Travis CI since it's a noop these days
- switch to bionic dist from trusty; the latter is a very old Ubuntu
  14.04.
- don't bother with "staging" branches; it's been unused for a long
  time now
- merge travis_ssh.sh content into upload_nightlies for easier
  maintenance

### DIFF
--- a/scripts/travis_upload_nightlies.sh
+++ b/scripts/travis_upload_nightlies.sh
@@ -2,6 +2,15 @@
 
 set -e
 
+# Don't upload anything unless it's a push commit to the master branch.
+if [ "$TRAVIS_PULL_REQUEST" != "false" ] || [ "$TRAVIS_BRANCH" != "master"]; then
+    exit 0
+fi
+
+openssl aes-256-cbc -K $encrypted_59febc5c238f_key -iv $encrypted_59febc5c238f_iv -in travis_rsa.enc -out travis_rsa -d
+chmod 600 travis_rsa
+mv travis_rsa ~/.ssh/id_rsa
+
 LINUX_BUILD=$TRAVIS_BUILD_DIR/frontends/qt/build/linux
 ANDROID_BUILD=$TRAVIS_BUILD_DIR/frontends/android/BitBoxApp/app/build/outputs/apk/debug
 UPLOADS=$TRAVIS_BUILD_DIR/uploads
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,23 +2,65 @@
 # https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions
 name: ci
 on: [push, pull_request]
+env:
+  # Use alternative image when running on GitHub workflows CI to avoid potential
+  # rate limiting when executing jobs in parallel: they can't cache docker images
+  # and always pull.
+  #
+  # To update this image, generate a personal token with write:packages scope
+  # on https://github.com/settings/tokens and authenticate yourself locally with
+  # "docker login https://docker.pkg.github.com -u <github-username>" using the
+  # newly generated token as password.
+  # Once logged in, tag an new image:
+  #   docker tag shiftcrypto/bitbox-wallet-app:VERSION \
+  #   docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:VERSION
+  # and push as usual:
+  #   docker push docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:VERSION
+  # Lastly, update the next line to use the newly pushed image version.
+  # See docs for more details:
+  # https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages
+  CI_IMAGE: docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:5
+  TRAVIS_BUILD_DIR: ${{github.workspace}}
 
 jobs:
-  linux-docker:
+  test-lint:
     runs-on: ubuntu-18.04
     steps:
       - name: Clone the repo
         uses: actions/checkout@v2
+      - name: Docker login
+        run: docker login -u ${{github.actor}} -p ${{github.token}} https://docker.pkg.github.com
       - name: Run CI script
-        run: ./scripts/travis-ci.sh
+        run: ./scripts/travis-ci.sh ci
         env:
           TRAVIS_OS_NAME: linux
-          TRAVIS_BUILD_DIR: ${{github.workspace}}
-      - name: Upload Android
+  android:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v2
+      - name: Docker login
+        run: docker login -u ${{github.actor}} -p ${{github.token}} https://docker.pkg.github.com
+      - name: Build Android
+        run: ./scripts/travis-ci.sh android
+        env:
+          TRAVIS_OS_NAME: linux
+      - name: Upload APK
         uses: actions/upload-artifact@v2
         with:
           path: frontends/android/BitBoxApp/app/build/outputs/apk/debug/app-debug.apk
           name: BitBoxApp-android-${{github.sha}}.apk
+  qt-linux:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v2
+      - name: Docker login
+        run: docker login -u ${{github.actor}} -p ${{github.token}} https://docker.pkg.github.com
+      - name: Build Qt-Linux
+        run: ./scripts/travis-ci.sh qt-linux
+        env:
+          TRAVIS_OS_NAME: linux
       - name: Upload AppImage
         uses: actions/upload-artifact@v2
         with:
@@ -39,11 +81,10 @@ jobs:
     steps:
       - name: Clone the repo
         uses: actions/checkout@v2
-      - name: Run CI script
-        run: ./scripts/travis-ci.sh
+      - name: Build macOS app
+        run: ./scripts/travis-ci.sh qt-osx
         env:
           TRAVIS_OS_NAME: osx
-          TRAVIS_BUILD_DIR: ${{github.workspace}}
       - name: Archive app
         run: >
           pushd ~/go/src/github.com/digitalbitbox/bitbox-wallet-app/frontends/qt/build/osx;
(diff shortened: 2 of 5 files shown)


x1ddos:
Actually, this now takes way longer on travis-ci. Maybe I should revert it and leave changes only for github ci.

benma:
> Upside is the
> pull runs faster. Downside: need to remember to update CI image in two
> places now.

How much faster? If only a little, maybe it's not worth it. If a lot, maybe Travis CI should pull the GH docker image instead?

benma:
> Actually, this now takes way longer on travis-ci. Maybe I should revert it and leave changes only for github ci.

How much slower (before vs after)? Imho if one of them is quick, that is enough. One green light is enough, the other is only redundancy.

x1ddos:
PTAL. Partially reverted travis-ci to run all builds in a single job script, same as before this PR.

x1ddos:
> > Actually, this now takes way longer on travis-ci. Maybe I should revert it and leave changes only for github ci.
> 
> How much slower (before vs after)? Imho if one of them is quick, that is enough. One green light is enough, the other is only redundancy.

Nah, it would've been to slow, ~10-15min each job. So, test/lint + linux + android + macos = ~40min.

x1ddos:
In comparison, check out github ci:

image

(travis is still crunching macOS build...)

x1ddos:
> How much faster? If only a little, maybe it's not worth it. If a lot, maybe Travis CI should pull the GH docker image instead?

Depends on which job. Slowest is qt-linux: 9min now, was ~15-17min. Fastest is test/lint.

x1ddos:
> > How much faster? If only a little, maybe it's not worth it. If a lot, maybe Travis CI should pull the GH docker image instead?
> 
> Depends on which job. Slowest is qt-linux: 9min now, was ~15-17min. Fastest is test/lint.

On Travis, pulling image from github wouldn't make it considerably faster. Running build jobs in parallele makes it actually faster. I'm just afraid one day dockerhub goes down and we're left without the docker image. So, seemed good to diversify.

benma:
> we're left without the docker image.

We have a Dockerfile, we can always build it again. 

> Depends on which job. Slowest is qt-linux: 9min now, was ~15-17min. Fastest are macOS (4min) and test/lint (5min).

I was asking about the docker pull specifically, not the total job, where you are now comparing one subjob to the total time before? 

Just trying to see if the GH registry is worth it. Maintaining to uploads seems like overhead that we might skip, unless the gain is considerable.

 >On Travis, pulling image from github wouldn't make it considerably faster

It's good enough if it isn't much slower. My goal is basically to not maintain two registries :P

x1ddos:
> We have a Dockerfile, we can always build it again.

```
$ curl -I https://download.qt.io/archive/qt/5.11/5.11.2/single/qt-everywhere-src-5.11.2.tar.xz
HTTP/1.1 404 Not Found
Date: Tue, 30 Jun 2020 13:49:15 GMT
Server: Apache
X-Prefix: 213.144.144.0/20
X-AS: 13030
Content-Type: text/html; charset=iso-8859-1
```

that's from https://github.com/digitalbitbox/bitbox-wallet-app/blob/d5f9512fa2f49ab58dbba295ab051b3863df374e/frontends/qt/docker-qt5base/Dockerfile#L84

> I was asking about the docker pull specifically, not the total job, where you are now comparing one subjob to the total time before?

Right. That's total including the docker pull. Only the pull is:

```
2020-06-30T12:34:27.1330183Z + docker pull docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:5
...
2020-06-30T12:37:17.5280279Z Status: Downloaded newer image for docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:5
```

So, 3min on github. Unfortunately, travis-ci doesn't provide timestamps but I suspect it would be 1-2min slower at the most.

> Maintaining two registries seems like overhead that we might skip, unless the gain is considerable.

Just like with two CIs - at least one will most likely work - 2 independent registries is better than one.

> It's good enough if it isn't much slower. My goal is basically to not maintain two registries :P

If you insist, I'll revert and come back asking about it next time hub.docker.com fails and brings down both CIs at the same time :sweat_smile:. Github was barely to no usable yesterday. I wonder who's next tomorrow.

x1ddos:
> If you insist, I'll revert and come back asking about it next time hub.docker.com fails and brings down both CIs at the same time sweat_smile. Github was barely to no usable yesterday. I wonder who's next tomorrow.

Although to be fair if github is unusable, no need for CI anyway.

benma:
> > We have a Dockerfile, we can always build it again.
> 
> ```
> $ curl -I https://download.qt.io/archive/qt/5.11/5.11.2/single/qt-everywhere-src-5.11.2.tar.xz
> HTTP/1.1 404 Not Found
> Date: Tue, 30 Jun 2020 13:49:15 GMT
> Server: Apache
> X-Prefix: 213.144.144.0/20
> X-AS: 13030
> Content-Type: text/html; charset=iso-8859-1
> ```
> 
> that's from
> 
> https://github.com/digitalbitbox/bitbox-wallet-app/blob/d5f9512fa2f49ab58dbba295ab051b3863df374e/frontends/qt/docker-qt5base/Dockerfile#L84
> 
> > I was asking about the docker pull specifically, not the total job, where you are now comparing one subjob to the total time before?
> 
> Right. That's total including the docker pull. Only the pull is:
> 
> ```
> 2020-06-30T12:34:27.1330183Z + docker pull docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:5
> ...
> 2020-06-30T12:37:17.5280279Z Status: Downloaded newer image for docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:5
> ```
> 
> So, 3min on github. Unfortunately, travis-ci doesn't provide timestamps but I suspect it would be 1-2min slower at the most.
> 
> > Maintaining two registries seems like overhead that we might skip, unless the gain is considerable.
> 
> Just like with two CIs - at least one will most likely work - 2 independent registries is better than one.
> 
> > It's good enough if it isn't much slower. My goal is basically to not maintain two registries :P
> 
> If you insist, I'll revert and come back asking about it next time hub.docker.com fails and brings down both CIs at the same time sweat_smile. Github was barely to no usable yesterday. I wonder who's next tomorrow.

Afaik there was not even a first time ;) We enabled GH ci after travis acted up multiple times. Let's do the same if and when dockerhub starts failing.

x1ddos:
Yeah, reactive vs proactive. Ok, I'll close for now.

benma:
Hmm okay, not sure why I was hesitant about a 2nd registry. I think maybe I was lazy to setup the authentication.. Feel free to overrule me anyway!

What about the parallel run? That was still worth it in any case, right?

Maybe the same can be applied in bitbox02-firmware, where CI is taking even longer?

x1ddos:
Without using github registry and pulling from hub.docker.com instead, I'm just afraid we'd start hammering dockerhub too much with parallel jobs. I'm already surprised docker hasn't rate-limited us already with all the pulls on each PR. :sweat_smile:

Reading https://docs.docker.com/docker-hub/download-rate-limit/

> Docker will gradually impose download rate limits with an eventual limit of 300 downloads per six hours for anonymous users.

Now, `podman inspect --format '{{.RootFS.Layers}}' shiftcrypto/bitbox-wallet-app:5`  shows 13 layers. That's 23 docker pulls every 6 hours. If we run test/lint + linux + android + macos in parallel per each PR, thats 4 docker pulls per PR or ~6 PRs every 6 hours, right. Add to that typos and "git push -f" which generate new CI builds and it's getting too close until docker starts returning 429 Too Many Requests and failing CI.

I'm sure both github and travis have some kind of a deal with docker for higher rate limits but I'm also certain everyone has their limits. 

Yeah, we could eventually add "docker login" to avoid "anonymous user rate limits" but it would be just postponing the inevitable because even being logged in won't allow infinite limits.

I guess I should've put all of this in the PR description! But tbh, I realized all this only later. 

> Maybe the same can be applied in bitbox02-firmware, where CI is taking even longer?

Definitely coming! Will try something soon.

x1ddos:
Ok, I reopened with a new description and added some docs, esp. on how to update the registry image.
PTAL

x1ddos:
> So, 3min on github. Unfortunately, travis-ci doesn't provide timestamps but I suspect it would be 1-2min slower at the most.

Now with timing, https://travis-ci.org/github/digitalbitbox/bitbox-wallet-app/jobs/703661865 shows 

```
real	4m30.585s
user 0m0.376s
sys	0m0.183s
```

So, docker pull on travis is indeed a bit slower, by ~1.5min.

x1ddos:
To sum up, if this is submitted:

- github: 9min (from 17min)
- travis: 22min (no change)
- appveyor: 7min (no change)

scripts/travis_upload_nightlies.sh
```diff
@@ -2,6 +2,15 @@
 
 set -e
 
+# Don't upload anything unless it's a push commit to the master branch.
```
benma, line 5: can be nice to test unmerged branches, or staging branches especially.

x1ddos: ah but you always have them available on github build artifacts too. pick any run here: https://github.com/digitalbitbox/bitbox-wallet-app/actions

and github even builds and upload in forks. like mine here: https://github.com/x1ddos/bitbox-wallet-app/actions/runs/153204239

this file, travis_upload_nightlies.sh is run on travis-ci only. or did you mean something else?

benma: True! All good.

PR #900, opened 2020-06-30
frontend: fix height in dialog for long addresses

thisconnect:
Long address can still break onto two lines, but this change
fixes the ugly tiny scrollbar by adding flexibleHeight prop.
The padding of the copy box is smaller and the width of the
dialog little wider so it should fit a few more letter in
one line.

### DIFF
--- a/frontends/web/src/components/copy/Copy.css
+++ b/frontends/web/src/components/copy/Copy.css
@@ -45,7 +45,7 @@
     font-weight: 400;
     min-width: 200px !important;
     width: 100%;
-    padding: var(--space-half) calc(var(--space-half) * 3) var(--space-half) var(--space-half);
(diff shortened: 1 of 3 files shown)


benma:
What is flexible height? Googling for it the first page seems irrelevant :O 

Did you try it in the Qt app?

thisconnect:
> Googling for it the first page seems irrelevant :O

It's not a standard, it just sets `height: auto` and then the logic in Copy.tsx can try to set the `row` attribute of the textarea element.

https://github.com/digitalbitbox/bitbox-wallet-app/blob/7dc542b3db737c2f4d2b1ef526d6293c1312e8f7/frontends/web/src/components/copy/Copy.tsx#L94

`flexibleHeight` property is used almost in all instances of the Copy component.

https://github.com/digitalbitbox/bitbox-wallet-app/search?q=flexibleHeight&unscoped_q=flexibleHeight

PR #903, opened 2020-07-01
qt/macOS: fix and document notarization

benma:
Notarization requires a hardened runtime. The app had two issues with
this:

- Gray screen at launch - fixed by whitelisting unsigned memory
execution in for QtWebEngine

- Couldn't open the HID device - fixed by
`allow-dyld-environment-variables`, by trial and error.

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -39,6 +39,21 @@ Build the QT frontend for MacOS:
 Build artifacts:
 * `frontends/qt/build/osx/BitBox.app`
 
+### Signing & Notarization
+
+Requires Xcode 10+ and macOS 10.13.6+.
+
+```
+$ cd frontends/qt/build/osx/
+$ # Sign with hardened runtime:
+$ codesign -f --deep --strict --timestamp -o runtime --entitlements ../../resources/MacOS/entitlements.plist -s CODESIGN_IDENTITY BitBox.app
+$ /usr/bin/ditto -c -k --keepParent BitBox.app BitBox.zip
+$ # Notarize
+$ xcrun altool --notarize-app --primary-bundle-id "ch.shiftcrypto.bitboxapp" --username "APPLE_ID" --password "PASSWORD" --file BitBox.zip
+$ # Check notarization status
+$  xcrun altool --notarization-info NOTARIZATION_ID --username "APPLE_ID" --password "PASSWORD"
+```
+
 ## Windows
 
 The build requires `mingw-w64`, `bash`, `make`, `Microsoft Visual Studio 2017`, `go 1.10`, `yarn`,
--- a/frontends/qt/resources/MacOS/entitlements.plist
+++ b/frontends/qt/resources/MacOS/entitlements.plist
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <!-- needed for QtWebEngine, otherwise there is a gray screen at launch -->
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <!-- needed for USB HID access, who knows why -->
+    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
+    <true/>
+
+    <!-- the QR code scanner works even without this. We add it just in case it might be enforced on some systems or in the future. -->
+    <key>com.apple.security.device.camera</key>
+    <true/>
+    <!-- the USB HID access works even without this. We add it just in case it might be enforced on some systems or in the future. -->
+    <key>com.apple.security.device.usb</key>
+    <true/>
+  </dict>
+</plist>


docs/BUILD.md
```diff
@@ -39,6 +39,21 @@ Build the QT frontend for MacOS:
...
 * `frontends/qt/build/osx/BitBox.app`
 
+### Signing & Notarization
+
+Requires Xcode 10+ and macOS 10.13.6+.
+
+```
+$ cd frontends/qt/build/osx/
+$ # Sign with hardened runtime:
+$ codesign -f --deep --timestamp -o runtime --entitlements ../../resources/MacOS/entitlements.xml -s CODESIGN_IDENTITY BitBox.app
```
x1ddos, line 49: should this read `entitlements.plist` instead of `entitlements.xml`?

frontends/qt/resources/MacOS/entitlements.plist
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <!-- needed for QtWebEngine, otherwise there is a gray screen at launch -->
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
```
x1ddos, line 6: :(

benma: :(

docs/BUILD.md
```diff
@@ -39,6 +39,21 @@ Build the QT frontend for MacOS:
...
 
+### Signing & Notarization
+
+Requires Xcode 10+ and macOS 10.13.6+.
+
+```
+$ cd frontends/qt/build/osx/
+$ # Sign with hardened runtime:
+$ codesign -f --deep --timestamp -o runtime --entitlements ../../resources/MacOS/entitlements.plist -s CODESIGN_IDENTITY BitBox.app
+$ zip -y -r BitBox.zip BitBox.app
```
Stadicus, line 50: This `zip` command results in an archive with a broken code signature, which is then rejected by the notarization process.

To preserve the valid code signature, the following command should be used.

```sh
/usr/bin/ditto -c -k --keepParent BitBox.app BitBox.zip
```

docs/BUILD.md
```diff
@@ -39,6 +39,21 @@ Build the QT frontend for MacOS:
...
 * `frontends/qt/build/osx/BitBox.app`
 
+### Signing & Notarization
+
+Requires Xcode 10+ and macOS 10.13.6+.
+
+```
+$ cd frontends/qt/build/osx/
+$ # Sign with hardened runtime:
+$ codesign -f --deep --timestamp -o runtime --entitlements ../../resources/MacOS/entitlements.plist -s CODESIGN_IDENTITY BitBox.app
```
Stadicus, line 49: Although that `codesign` command works, I would consider adding the `--strict` flag, which is recommended in the Apple notarization documentation.

It's worthwhile to note that using the `--timestamp` flag requires a good internet connectivity, due to external timestamp validation. If the connectivity is shaky, the code signing without `--strict` will still succeed without warning, but the result cannot be notarized. With the option `--strict`, the code signing will fail, which is preferrable in this case.

docs/BUILD.md
benma: Great! Updated the docs.

x1ddos: @Stadicus can you explain in details what breaks, why `ditto` and not xyz. Is this from Apple's official docs or elsewhere? Very curious.

x1ddos: Also, why `--keepParent` and `-k`?

Stadicus: This is the general consensus in compressing a file or folder exactly like MacOS Finder would do it internally. The most "official" doc I could find is on Stack Overflow: https://stackoverflow.com/questions/107903/how-to-create-a-zip-file-in-the-same-format-as-the-finders-compress-menu-item

This is not the *exact* emulation, however, as we don't use the `--sequesterRsrc` flag (see below), which causes an error and is not necessary in this context.

See the [ditto man page](https://ss64.com/osx/ditto.html) for additional detail.

```
ditto

Copy directory hierarchies, create and extract archives. Ditto can preserve ownership / permissions, 
resource forks and file / folder metadata. Ditto will automatically create the destination folder if it 
doesn’t yet exist, if the destination does exist and contains files, then ditto will merge them.

Syntax
     ditto -c [options] Source  Destination_archive

     -c            Create an archive at the destination path.  The default format is CPIO, unless
                   -k is given.  CPIO archives should be stored in files with names ending in .cpio.
                   Compressed CPIO archives should be stored in files with names ending in .cpgz.

     -k            Create or extract from a PKZip archive instead of the default CPIO.
                   PKZip archives should be stored in filenames ending in .zip.

     --keepParent  When creating an archive, embed the parent directory name src in dst_archive.

     --sequesterRsrc
                   When creating a PKZip archive, preserve resource forks and HFS meta-data in the
                   subdirectory __MACOSX.  PKZip extraction will automatically find these resources.
```

x1ddos: Yeah, I saw that man page. It's what/how vs why. Would signing work without preserving parent? I just don't personally like using flags for unknown or no reason, esp. if it works without too.

So, to clarity: you did try with `-y` flag to `zip ...` command and that didn't work during signing? I still don't get what "broken code signature" means.

benma: sorry merged prematurely before seeing this discussion. i guess we can adjust the doc in a new PR if it works with less flags.

benma: > So, to clarify: you did try with -y flag to zip ... command and that didn't work during signing? I still don't get what "broken code signature" means.

he did, that didn't work.

Stadicus: My goal was to do on the command line what @benma did manually up until now. That includes having the parent directory in the archive (as Finder does). Would possibly work without, but it's a very simple and transparent flag, and not using it would lead to a different archive that needs additional testing.

The "broken code signature" means that I sign the binary and verify it, then create the archive, and then notarization complains that the code signature is invalid. Other users online had the same issue, speculating about wrong UTF8 encoding, but I did not do a research study on that. In my view it's pretty simple: `zip` creates a archive with a no longer valid code signature, so we use the generally recommended tool to create an archive that preserves the code signature.

x1ddos: ok. thanks.

PR #907, opened 2020-07-02
ci: preserve symlinks while zipping macos app

x1ddos:
See build docs https://github.com/digitalbitbox/bitbox-wallet-app/blob/673708a/docs/BUILD.md#signing--notarization and #903.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
       - name: Archive app
         run: >
           pushd ~/go/src/github.com/digitalbitbox/bitbox-wallet-app/frontends/qt/build/osx;
-          zip -r ${{github.workspace}}/BitBoxApp-macos.zip BitBox.app;
+          ditto -c -k --keepParent BitBox.app ${{github.workspace}}/BitBoxApp-macos.zip;
           popd;
       - name: Upload app
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - name: Archive app
         run: >
           pushd ~/go/src/github.com/digitalbitbox/bitbox-wallet-app/frontends/qt/build/osx;
-          zip -r ${{github.workspace}}/BitBoxApp-macos.zip BitBox.app;
+          zip -y -r ${{github.workspace}}/BitBoxApp-macos.zip BitBox.app;
```
benma, line 91: Actually we should use `ditto`

https://github.com/digitalbitbox/bitbox-wallet-app/pull/903#discussion_r448808691

as `zip` seems to modify the app bundle in undesired ways. `ditto` behaves like the Finder when compressing.

x1ddos: Done.

x1ddos:
PTAL

x1ddos:
Also, this now reports smaller size indeed:

```
Finished uploading artifact BitBoxApp-macos-4dcc80d0b8a8ce8fedb20426ce5f5bd546cf6a24.zip. Reported size is 110714984 bytes
```

from https://github.com/digitalbitbox/bitbox-wallet-app/pull/907/checks?check_run_id=830161403

PR #909, opened 2020-07-02
ci: use correct date arguments when uploading nightly on macos

x1ddos:
The date command on macOS doesn't support the short -I or --iso flag:

    $  ./scripts/travis_upload_nightlies.sh
    date: illegal option -- I
    usage: date [-jnRu] [-d dst] [-r seconds] [-t west]
    [-v[+|-]val[ymwdHMS]] ...
    [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]

So, use the longer +format argument.

### DIFF
--- a/scripts/travis_upload_nightlies.sh
+++ b/scripts/travis_upload_nightlies.sh
@@ -41,7 +41,7 @@ fi
 
 if [ "$TRAVIS_OS_NAME" == "osx" ]; then
     cd $MACOS_BUILD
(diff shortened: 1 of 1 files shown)


x1ddos:
> Confused, your previous PR finished the upload script without an error shrug

PR yes, but not the commit push to master which is where that upload nightlies script runs:

image

image
(at the bottom of https://travis-ci.org/github/digitalbitbox/bitbox-wallet-app/jobs/704241548)

x1ddos:
The trouble with this is it runs only on push commits to master. So, harder to debug locally.

PR #911, opened 2020-07-03
Frontend: fixes account info cropped on linux

thisconnect:
Removed classes that were previously used to vertically center
the address. Unfortunatelly this issue was not reproducible in
the macOS app. The addresses do not have to be vertically centered
anymore so removing the flex classes is fine.


thisconnect:
this should go into the release if possible

PR #912, opened 2020-07-03
Frontend: update split settings

thisconnect:
- Increase width of expert setting column to 2/3
- Mention that split account is legacy behavior

Screen Shot 2020-07-03 at 10 48 49 AM


frontends/web/src/locales/en/app.json
```diff
@@ -1224,7 +1224,7 @@
         "title": "Connect your own full node"
       },
       "setProxyAddress": "Set proxy address",
-      "splitAccounts": "Separate accounts by address type",
+      "splitAccounts": "Separate accounts by address type (legacy behavior)",
```
benma, line 1227: please remember that i18n changes need to happen on locize, otherwise the changes will be lost. See https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/docs/i18n.md.

So you can remove this from the commit.

benma:
Do you have a screenshot in an Android-sized window?

thisconnect:
Update settings view to use full width on small screen

Before:
Screen Shot 2020-07-04 at 1 59 56 PM

After:
Screen Shot 2020-07-04 at 1 59 13 PM

PR #913, opened 2020-07-03
ci: enable caching to avoid intermittent build errors

x1ddos:
Sometimes yarn install command fails with a network error.
Gradle and Homebrew do too but less often.
This will hopefully make the builds more stable.

Most recent run failed due to network: https://github.com/digitalbitbox/bitbox-wallet-app/runs/835012963, but there are more if you go back in build history, on both travis and github.

To invalidate the cache on GitHub, it is sufficient to change the cache key. Reference docs:
https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy.
Travis has an option in their UI: https://docs.travis-ci.com/user/caching/#clearing-caches

Both Travis and GitHub cache node modules, gradle and homebrew.
Some rough cache size numbers at the time of this commit:

- node modules: 37Mb
- gradle: 380Mb
- homebrew: 3.4Gb on github but it's the whole setup, not just ours;
  github cache is capped at 5Gb
- travis-ci says total cache size ~455Mb; cap is unknown

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,11 @@ jobs:
         uses: actions/checkout@v2
       - name: Docker login
         run: docker login -u ${{github.actor}} -p ${{github.token}} https://docker.pkg.github.com
+      - name: Enable caching
+        uses: actions/cache@v2
+        with:
+          key: ${{runner.os}}-test-lint
+          path: frontends/web/node_modules
       - name: Run CI script
         run: ./scripts/travis-ci.sh ci
         env:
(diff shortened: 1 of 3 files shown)


benma:
Sorry can you elaborate again what intermittent CI errors have to do with caching? Can't one just re-start the CI build if it failed for no good reason?

x1ddos:
> Sorry can you elaborate again what intermittent CI errors have to do with caching? Can't one just re-start the CI build if it failed for no good reason?

Sure but then you have to babysit it, esp. when it's a commit push to master and you may be expecting a "nightly" build to show up. Here's a typical failure example:

image

With caching enabled, `yarn install` doesn't fetch from network anymore in that example, unless package.json changes or CI cache is evicted.

I don't see how this can impact CI builds in a negative way regardless. In fact, the builds are now ~1min shorter although it isn't the goal here.

PR #916, opened 2020-07-04
backend/rates: avoid excessive logging and drop go-spew

x1ddos:
I find it very annoying, seeing nothing except "exchange rates changed"
messages in the logs. Hard to look for anything else.

Apart from the usual "no news is good news" reasoning, if exchange rate
update fails, it should be surfaced to the app UI in some way because
noone reads the logs except us. But it's beyond this simple commit.

Turns out spew package was used only in that one place.
Hence, vendor/modules.txt removes '## explicit' but the package is
still there because it's used by some others.

### DIFF
--- a/backend/rates/rates.go
+++ b/backend/rates/rates.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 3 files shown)


x1ddos:
Damn github doesn't allow octopus merges in protected branches... but it's the same commit hashes :see_no_evil:.

PR #917, opened 2020-07-04
util/socksproxy: remove excessive logging

x1ddos:
The "using an unproxied ..." log line does not help in any way
but it keeps repeating every time something's using it, like the
exchange rate updates. It is actually any of these and more:

    backend/update.go:
    client, err := backend.socksProxy.GetHTTPClient()

    backend/devices/bitbox/relay/request.go:
    httpClient, err := request.channel.socksProxy.GetHTTPClient()

    backend/coins/eth/etherscan/etherscan.go:
    client, err := etherScan.socksProxy.GetHTTPClient()

    backend/rates/rates.go:
    client, err := updater.socksProxy.GetHTTPClient()

    backend/backend.go:
    httpClient, err := backend.socksProxy.GetHTTPClient()

When logged, it can't be attributed to the caller.
It's a separate line and logging is typically buffered; line ordering
should not be relied upon.

Similar reasoning with when the connection is proxied.
This should be logged once, at the app startup, or by the caller of
NewSocksProxy if really necessary. This commit opts for the former.

One other part is devices/bitbox/relay but this is bb01 and the proxy
is always unused.

As an example, here's how it looked like before this commit, with https://github.com/digitalbitbox/bitbox-wallet-app/pull/916 patched in:

image

### DIFF
--- a/util/socksproxy/socksproxy.go
+++ b/util/socksproxy/socksproxy.go
@@ -46,16 +46,14 @@ func NewSocksProxy(useProxy bool, proxyAddress string) SocksProxy {
 // GetTCPProxyDialer returns a tcp connection. The connection is proxied, if useProxy is true.
 func (socksProxy *SocksProxy) GetTCPProxyDialer() proxy.Dialer {
 	if socksProxy.useProxy {
-		socksProxy.log.WithField("address", socksProxy.proxyAddress).
-			Info("Using proxy connection")
 		// Create a proxy that uses Tor's SocksPort.
 		dialer, err := proxy.SOCKS5("tcp", socksProxy.proxyAddress, nil, nil)
 		if err != nil {
-			panic("Failed to create tcp connection over socks5 proxy")
+			// TODO: Remove this panic.
+			socksProxy.log.WithError(err).Panic("Failed to create SOCKS5 TCP dialer")
 		}
 		return dialer
 	}
-	socksProxy.log.Info("Using an unproxied tcp client")
 	return &net.Dialer{}
 }
 
@@ -63,8 +61,6 @@ func (socksProxy *SocksProxy) GetTCPProxyDialer() proxy.Dialer {
 func (socksProxy *SocksProxy) GetHTTPClient() (*http.Client, error) {
 	if socksProxy.useProxy {
 		// Create a transport that uses Tor Browser's SocksPort.
-		socksProxy.log.WithField("address", socksProxy.fullProxyAddress).
-			Info("Creating new socksProxy http client")
 		tbProxyURL, err := url.Parse(socksProxy.fullProxyAddress)
 		if err != nil {
 			socksProxy.log.WithError(err).Error("Failed to parse proxy URL")
@@ -86,6 +82,5 @@ func (socksProxy *SocksProxy) GetHTTPClient() (*http.Client, error) {
 		client := &http.Client{Transport: tbTransport}
 		return client, nil
 	}
-	socksProxy.log.Info("Using an unproxied http connection")
 	return &http.Client{}, nil
 }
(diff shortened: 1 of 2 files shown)


util/socksproxy/socksproxy.go
```diff
@@ -46,25 +46,22 @@ func NewSocksProxy(useProxy bool, proxyAddress string) SocksProxy {
 // GetTCPProxyDialer returns a tcp connection. The connection is proxied, if useProxy is true.
 func (socksProxy *SocksProxy) GetTCPProxyDialer() proxy.Dialer {
 	if socksProxy.useProxy {
-		socksProxy.log.WithField("address", socksProxy.proxyAddress).
-			Info("Using proxy connection")
 		// Create a proxy that uses Tor's SocksPort.
 		dialer, err := proxy.SOCKS5("tcp", socksProxy.proxyAddress, nil, nil)
 		if err != nil {
+			socksProxy.log.WithError(err).Error("Failed to create SOCKS5 TCP dialer")
```
benma, line 52: can do `.Panic()` instead of `.Error()` and remove the panic below.

x1ddos: I'd actually prefer to remove all panic from this package... and the others at some point. Not sure why it's even here: `proxyAddress` is user input. It can literally be anything. The app shouldn't crash because of bad user input or broken network.

Or you mean just for the time being, until I get to removing panic. :)

benma: > Or you mean just for the time being, until I get to removing panic. :)

This in any case, but:

 > Not sure why it's even here: proxyAddress is user input. It can literally be anything. The app shouldn't crash because of bad user input or broken network.

I dare you to make it crash it bad user input related to socks proxy :P See, this function only returns the dialer, and does not actually dial. Dialing itself will fail if the address is bad.

This is the full SOCKS5 implementation:


```Go
// Copyright 2011 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package proxy

import (
	"context"
	"net"

	"golang.org/x/net/internal/socks"
)

// SOCKS5 returns a Dialer that makes SOCKSv5 connections to the given
// address with an optional username and password.
// See RFC 1928 and RFC 1929.
func SOCKS5(network, address string, auth *Auth, forward Dialer) (Dialer, error) {
	d := socks.NewDialer(network, address)
	if forward != nil {
		if f, ok := forward.(ContextDialer); ok {
			d.ProxyDial = func(ctx context.Context, network string, address string) (net.Conn, error) {
				return f.DialContext(ctx, network, address)
			}
		} else {
			d.ProxyDial = func(ctx context.Context, network string, address string) (net.Conn, error) {
				return dialContext(ctx, forward, network, address)
			}
		}
	}
	if auth != nil {
		up := socks.UsernamePassword{
			Username: auth.User,
			Password: auth.Password,
		}
		d.AuthMethods = []socks.AuthMethod{
			socks.AuthMethodNotRequired,
			socks.AuthMethodUsernamePassword,
		}
		d.Authenticate = up.Authenticate
	}
	return d, nil
}

```

as you can see, `nil` is the only error ever returned.

Of course the implementation might change in this regard, but currently this panic can't happen.

x1ddos: > Of course the implementation might change in this regard (seems a bit unlikely though), but currently this panic can't happen.

"... And Other Hilarious Jokes You Can Tell Yourself."

Are you trolling me? :sweat_smile:
That's like going in a direction opposite to reliable software.

benma: You know my answer before I give it :D 

Of course I agree that error handling is better and more reliable than a panic. I have seen implementations and assumptions change under my feet as well.

I don't mind the panic in this particular case anyway :woman_shrugging: of course not against removing it anyway :)

x1ddos: > can do `.Panic()` instead of `.Error()` and remove the panic below.

Ok done. Will deal with panic later anyway. :)

PR #920, opened 2020-07-05
pick UI language from system locale

x1ddos:
I believe UI language detection was gone around commit 61ff0427. This PR restores the functionality but uses a different approach: query app native layer, without jibber-jabber and syscalls. Works on Linux, Windows and Android. No idea about macOS but I expect it to.

See commit messages for details.

@thisconnect please take a look at relevant frontend commits.

---

Before trying this manually, delete `userLanguage` field from your frontend app config. Here's how to find it:

- on linux, app config is usually `~/.config/bitbox/config.json` or whichever your `XDG_CONFIG_HOME` points to
- on macOS it's in `~/Library/Application Support/bitbox/config.json`
- on Windows, it's typically `C:\Users\<username>\Application Data\bitbox\config.json`
- on Android... probably impossible on unrooted phone: uninstalling the app seems to keep the config anyway 

## linux

The simplest is to overwrite `LANG` env variable. For example:

    LANG=fr BitBoxApp.AppImage

## macos

Try like linux but no idea really. I'm sure they have a GUI option. Don't have a macOS at the moment.

## windows 10

Settings => Time & Language => Language

Then "Add a preferred language", one that is [supported by the app](https://github.com/digitalbitbox/bitbox-wallet-app/tree/master/frontends/web/src/locales), and select it in the "Windows display language" dropdown.

Log out and back in for it to take effect.

## android 9

Settings => System => Languages & input => Languages

"+ Add a language" at the bottom if you haven't already and drag it to the top.


backend/config/config.go
```diff
@@ -127,8 +128,13 @@ func (backend Backend) CoinActive(code coin.Code) bool {
...
 // AppConfig holds the whole app configuration.
 type AppConfig struct {
-	Backend  Backend     `json:"backend"`
-	Frontend interface{} `json:"frontend"`
+	Backend Backend `json:"backend"`
+	// Frontend structure is owned by frontend/web/src/{i18n,config},
+	// hence empty interface values. However, one key is shared with
+	// the backend: userLanguage. When a stored app config is missing,
+	// the backend attempts to parse user preferred UI language and
+	// sends it to the frontend in a default config.
```
benma, line 136: What about graduating the user language to a backend config field?

The Backend configs are meant for fields that the backend will access. Frontend is for frontend-only config and opaque to the backend.

backend/config/config.go
```diff
@@ -170,8 +176,22 @@ ffbrVM+I91v3R03Svv2Nte2xdbx1RmoI/y3tMyZL
...
+	lang, err := language.Parse(localeHint)
+	// ValueError may be returned when the input is well-formed but the
+	// respective subtag is not recognized, in which case it's ok because
+	// the canonical string representation is still valid.
+	if _, ok := err.(language.ValueError); err != nil && !ok {
+		lang = language.English
+	}
 	return AppConfig{
+		Frontend: map[string]interface{}{
+			"userLanguage": lang.String(),
```
benma, line 193: The way I understand it, the system langauge is stored in the config on first run, and from then on, that one is used.

Why store the language in the config in the first place? If the user changes the system language, the app won't change the language, right?

backend/config/config_test.go
```diff
@@ -0,0 +1,84 @@
...
+		{"en-something", "en"},
+		{"de", "de"},
+		{"de-DE", "de-DE"},
+		{"de-CH", "de-CH"},
+	}
+	for _, test := range tt {
+		appconf := NewDefaultAppConfig(test.locale)
+		if v := appconf.Frontend["userLanguage"].(string); v != test.wantLang {
+			t.Errorf("%s: userLanguage=%q; want %s", test.locale, v, test.wantLang)
+		}
```
benma, line 39: since we are using testify everywhere else, you could save some lines and manual error msg formatting by using it, e.g. `require.Equal(t, test.wantLang, appconf.Frontend["userLanguage"].(string))`.

backend/config/config_test.go
```diff
@@ -0,0 +1,84 @@
...
+	if v := conf.appConfig.Frontend["userLanguage"].(string); v != "fr" {
+		t.Errorf("userLanguage = %q; want 'fr'", v)
+	}
+
+	// Manually set lang in app config.
+	confFile2, err := ioutil.TempFile("", "appconfig2")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(confFile2.Name()) //nolint:errcheck
```
benma, line 70: maybe the linter could be disabled for tests somehow

frontends/qt/server/server.go
```diff
@@ -131,7 +132,7 @@ func serve(
 			DeviceInfosFunc:     usb.DeviceInfos,
 			SystemOpenFunc:      system.Open,
 			UsingMobileDataFunc: func() bool { return false },
-			NativeLocaleFunc:    func() string { return "" },
+			NativeLocaleFunc:    func() string { return C.GoString(preferredLocale) },
```
benma, line 135: This is a closure, so `preferredLocale` from C++ might already be popped of the stack by the time this function is called?

backend/config/config_test.go
x1ddos: Here's my partially imagined experience with testify today. Took some notes:

```go
tt := []struct{ locale, wantLang string }{
  {"de DE", "de-DE1"},
  {"de-CH", "de-CH2"},
}
```

(wantLang is wrong to make tests fail on purpose)

- 10 extra lines of useless text per test case:

```
$ go test # with testify
--- FAIL: TestNewDefaultAppConfigLanguage (0.00s)
    config_test.go:39:
                Error Trace:    config_test.go:39
                Error:          Not equal:
                                expected: "en"
                                actual  : "de-DE1"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -en
                                +de-DE1
                Test:           TestNewDefaultAppConfigLanguage
```

That's exactly 100 extra lines of useless text in the worst case to dig through if I have 10 cases. ...yeah, I guess it can always get worse. Let's keep an open mind.
 
Although, all I really need, with regular testing pkg, is this:

```
--- FAIL: TestNewDefaultAppConfigLanguage (0.00s)
    config_test.go:39: de DE: userLanguage="en"; want de-DE1
    config_test.go:39: de-CH: userLanguage="de-CH"; want de-CH2
```

With testify, I see only values, no context. In larger tests it plays an even more important role.

Ok, I guess I can add another arg to require.Equal. Let's see:

```go
require.Equal(t, appconf.Frontend["userLanguage"].(string), test.wantLang, "userLanguage")
```

Apart from the line now being longer and I might not see it all on the screen so maybe opt for split on 2 lines instead, here's what I see:

```
--- FAIL: TestNewDefaultAppConfigLanguage (0.00s)
    config_test.go:39:
                Error Trace:    config_test.go:39
                Error:          Not equal:
                                expected: "de-DE1"
                                actual  : "en"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -de-DE1
                                +en
                Test:           TestNewDefaultAppConfigLanguage
                Messages:       userLanguage
```

A message! What does it mean? Who knows. I'll go read the test code instead of fixing the problem...

Maybe I'll end up writing a longer message like "if you see this, it means userLanguage output is incorrect" so next time someone sees it, they'll know what it means!

Hold on, expected vs actual. Which is which? go doc require Equal:

```go
func Equal(t TestingT, expected interface{}, actual ...
```

Aha! Need to swap the args. Ok, for the damned 10th time, I really hope I'll remember. It's the opposite of the convention which I see every day when reading normal testing pkg tests and always get confused. I hope Python people are not reading this.

Anyway, notice how regular tests above displayed two errors, not one, which brings me to:

- `require.Equal` is actually `t.Fatal` but I need `t.Error`.

Seeing all errored test cases often tells a pattern of what's wrong, unlike only the first failure and you start wondering, "but if the input were _this_, what would've been the output?". That's why I don't normally roll the dice between `t.Fatal` and `t.Error` and pick the one that fits best.

I guess I'll go and temporary comment out some lines to let the test keep going instead of having possibly already understood the problem...

For example, in TDD you start from nothing, add test cases, keep it running and implement until it's all green. Repeat. Basically imposible with testify because it fails on first error.

- Ok, maybe there's a `require.EqualButNotFatal`. Let's see:

```
$ go doc require | grep -E ^func | wc -l
105
```

That's how many require.Xxx funcs there are. I guess I'll go read them instead of writing a simple test...

Ok, I stopped reading. That's too many.

- Sure, I can always t.Run. This also improves on t.Fatal and the TDD problem:

```go
t.Run(test.locale, func(t *testing.T) {
  // test code here
})
```

```
--- FAIL: TestNewDefaultAppConfigLanguage (0.00s)
    --- FAIL: TestNewDefaultAppConfigLanguage/de_DE (0.00s)
        config_test.go:40:
                Error Trace:    config_test.go:40
                Error:          Not equal:
                                expected: "de-DE1"
                                actual  : "en"
```

Wait a second, actual "en" from input "de_DE"? That's not good at all.
Oh man, of course, did I forget to shadow the loop var? Maybe.

```go
test := test
```

... nope. Same result.

... after 1 hour:

Totally forgot that `t.Run(name, func)` replaces spaces with underscores in the name! So, the input was actually "de DE", with a space. Makes sense now why the actual output is "en".

Ok, I guess I'll have to add the input to my already long message `... output is incorrect; the input was %q", test.locale`.

So, what's `t.Run`'s test name? A confusing artifact of me being forced to use subtests. Otherwise, it gets even worse due to t.Fatal everywhere.

x1ddos: If you really insist, I'll replace with testify but I hope the above justifies why I resist.

backend/config/config.go
x1ddos: One case is this. If you pick a different language in the app, close it and open again next day, do you expect it to preserve your language of choice from the day before?

> If the user changes the system language, the app won't change the language, right?

Right. It reads from the config. Config always takes precedence.

Also, I wonder what @thisconnect thinks.

backend/config/config_test.go
x1ddos: I didn't realize even https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures mentions it.

frontends/qt/server/server.go
x1ddos: In C++ the `serve` is actually called from main so it's never popped from the stack until exit. But you're right, I guess I got lucky! - we could potentially move the serve elsewhere in the future and this might suddenly start crashing.

backend/config/config.go
benma: > If you pick a different language in the app, close it and open again next day, do you expect it to preserve your language of choice from the day before?

Yes I do. But if I never *explicitly* change the language, I would expect it to be my system language.

backend/config/config_test.go
benma: > Ok, maybe there's a `require.EqualButNotFatal.` Let's see: ... Repeat. Basically imposible with testify because it fails on first error.

You can use `assert.Equal` instead of `require.Equal` if you don't wish to abort.

> A message! What does it mean? Who knows. I'll go read the test code instead of fixing the problem...

That's normal? If a test fails, you open the test code first in any case. Never had an issue identifying the problem with testify.

> Wait a second, actual "en" from input "de_DE"? That's not good at all.

Sorry I don't think I follow what's happening here anymore ;) Explain over :beer: 

In any case, of course I don't resist ;) but I also think you dismiss it too early. Longer error output in the test can be good and bad, but mostly good. More concise test code is good imho. Let's convert this for fun together one day and discuss pros and cons.

Maybe something other than testify also does the job in a way you like better, but adding at least three lines of code and manual formatting of error mesages everytime I want to write one assertion is almost a no-go for me.

frontends/qt/server/server.go
benma: Ah true. I guess I was confused by:

`preferredLocale.toStdString().c_str()`

Which is kind of like a local var before the function call:

`auto cstr = preferredLocale.toStdString().c_str();`

which also lives until the end, but I kinda assumed that as a function argument, maybe it could be cleaned up earlier than when main() exits. On 2nd thought that sounds very wrong ;)

backend/config/config.go
x1ddos: Right. Very good point and it was like this before. I changed it here, I think. Will try to restore that functionality: never store userLanguage unless it's been manually changed.

backend/config/config_test.go
x1ddos: I guess the main difference is I prefer to not even read the test code and rely on the easy-to-read test output so I can go directly to the actual source code and fix the problem. Testify and other assert-like packages get very much in the way of that.

Source code: must be readable
Test code: don't care as much but its **output** must be easy readable 

You rely on the test source code and not so much on the test output. As in, for you the test output is more like a hint.

backend/config/config_test.go
x1ddos: done in https://github.com/digitalbitbox/bitbox-wallet-app/pull/922

backend/config/config.go
x1ddos: I thought about it The backend is always stored on disk though, and I wanted something runtime-only. So, added a new struct field called "volatile". Let's see how it goes.

backend/config/config.go
x1ddos: Done using a "volatile" struct field which is never stored.

frontends/qt/server/server.go
x1ddos: Done

x1ddos:
PTAL although I can't seem to be able to get rid if `frontend.userLanguage`. It's always saved at start up.

backend/config/config.go
benma: If the field is a pointer, it can be nil, which could mean "optional / not set / not defined", like is usual in Go<->Json, and still be in the Backend struct.

Adding a whole struct full of fields that are "optional" seems a bit off when you can have `null` for the optional fields instead. WDYT?

x1ddos: Sure, a nil field in the backend struct would also work but it's really a hack at that point. :)
I just don't like mixing things. To me, the backend config is for all background things, nothing to do with the native language. So, something in between clearly separated it.

x1ddos: Take a look at the new version. Tell me what you think.

x1ddos:
> I can't seem to be able to get rid if `frontend.userLanguage`. It's always saved at start up.

Yep, even on the master branch, this part seems like it's always called at startup: https://github.com/digitalbitbox/bitbox-wallet-app/blob/3d1e5efc/frontends/web/src/i18n/i18n.js#L107-L119

```js
i18n.on('languageChanged', (lng) => {
    apiGet('config').then((config = {}) => {
        if (config.frontend && config.frontend.userLanguage === lng) {
            return;
        }
        const newConf = Object.assign(config, {
            frontend: Object.assign({}, config.frontend, {
                userLanguage: lng
            })
        });
        apiPost('config', newConf);
    });
});
```

@thisconnect any idea why `languageChanged` might be triggered at start up?

backend/config/config.go
x1ddos: All trouble is because AppConfig is used for both saving to disk and communicating with the frontend.

x1ddos: After talking elsewhere, settled on a separate `GET /native-locale` endpoint instead of the hacky config fields stuff. Will change it.

x1ddos:
> @thisconnect any idea why `languageChanged` might be triggered at start up?

Never mind. I think it's ok that it's triggered. I'll just need to change this part:

```js
if (config.frontend && config.frontend.userLanguage === lng) {
```

backend/config/config.go
x1ddos: Done.

backend/config/config.go
x1ddos: Replaced "volatile" with `native-locale` backend API endpoint.

x1ddos:
PTAL all commits from the top. I managed to make it work with and without `frontend.userLanguage`. It is now always consistent with the system locale settings.

@thisconnect you might be interested in bf9cf41. It even has tests.

x1ddos:
@benma found one inconsistency. When system locale uses underscores like in `de_DE`, i18next fails to parse it correctly. This results in the UI displayed in English instead of German.

Created an issue in https://github.com/i18next/i18next/issues/1484.

x1ddos:
> @benma found one inconsistency. When system locale uses underscores like in `de_DE`, i18next fails to parse it correctly. This results in the UI displayed in English instead of German.
> 
> Created an issue in [i18next/i18next#1484](https://github.com/i18next/i18next/issues/1484).

PTAL. Added a naive workaround, replacing `_` with `-`. Works flowlessly on Android :)

PR #922, opened 2020-07-06
ci: upgrade golangci-lint to v1.27 and make it happy

x1ddos:
Additionally, exclude some annoying and useless linters, especially in test files.

Unfortunately, I went down the godot linter hole deep enough before it was too late. So, here it is. On the bright side, doc comments look more consistent now. Why a doc comment must start with the type name but not end with a dot.

This commit also adds a separate golint-ci job as a GitHub workflow action. This allows for nice annotations right in the Checks tab on a PR:

image

@benma if you're ok with this, could you rebuild and push the new image? I'm dockerless for quite some time now. I'd need to pull 10s of Gb otherwise... Instructions on how to update github's image copy are here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/0e0543f/.github/workflows/ci.yml#L6-L22.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,10 +19,22 @@ env:
   # Lastly, update the next line to use the newly pushed image version.
   # See docs for more details:
   # https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-docker-for-use-with-github-packages
-  CI_IMAGE: docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:5
+  #
+  # Keep this in sync with default in scripts/travis-ci.sh.
+  CI_IMAGE: docker.pkg.github.com/digitalbitbox/bitbox-wallet-app/ci-builder:6
   TRAVIS_BUILD_DIR: ${{github.workspace}}
 
 jobs:
+  golangci:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v2
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v1
+        with:
+          # Keep this in sync with the version in envinit make target.
+          version: v1.27
   test-lint:
     runs-on: ubuntu-18.04
     steps:
@@ -31,6 +43,8 @@ jobs:
       - name: Docker login
         run: docker login -u ${{github.actor}} -p ${{github.token}} https://docker.pkg.github.com
       - name: Run CI script
+        # The script also runs golang-ci but it's ok: doesn't take too long and may be useful
+        # to keep its linter errors in this log, too.
         run: ./scripts/travis-ci.sh ci
         env:
           TRAVIS_OS_NAME: linux
--- a/backend/coins/eth/db/db.go
+++ b/backend/coins/eth/db/db.go
@@ -91,9 +91,11 @@ func (tx *Tx) PutOutgoingTransaction(transaction *types.TransactionWithMetadata)
 
 type byNonce []*types.TransactionWithMetadata
 
-func (txs byNonce) Len() int           { return len(txs) }
-func (txs byNonce) Less(i, j int) bool { return txs[i].Transaction.Nonce() < txs[j].Transaction.Nonce() }
-func (txs byNonce) Swap(i, j int)      { txs[i], txs[j] = txs[j], txs[i] }
+func (txs byNonce) Len() int      { return len(txs) }
+func (txs byNonce) Swap(i, j int) { txs[i], txs[j] = txs[j], txs[i] }
+func (txs byNonce) Less(i, j int) bool {
+	return txs[i].Transaction.Nonce() < txs[j].Transaction.Nonce()
+}
 
 // OutgoingTransactions implements DBTxInterface.
 func (tx *Tx) OutgoingTransactions() ([]*types.TransactionWithMetadata, error) {
--- a/backend/devices/bitbox/device.go
+++ b/backend/devices/bitbox/device.go
@@ -1264,7 +1264,7 @@ func (dbb *Device) ecdhPK(mobileECDHPK string) (interface{}, error) {
 	return reply["ecdh"], nil
 }
 
-// ecdhChallenge forwards a ecdh challenge command to the Bitbox
+// ecdhChallenge forwards a ecdh challenge command to the Bitbox.
 func (dbb *Device) ecdhChallenge() error {
 	if dbb.bootloaderStatus != nil {
 		return errp.WithStack(errNoBootloader)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
-# This file contains all available configuration options
-# with their default values.
+# See reference docs for details:
+# https://golangci-lint.run/usage/configuration/
 
 # options for analysis running
 run:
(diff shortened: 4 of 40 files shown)


x1ddos:
"test-lint" check is failing as expected with:

```
+ golangci-lint run
level=error msg="Running error: no such linter \"gocognit\""
```

because it's still running on the old docker image. travis will also fail.

backend/coins/eth/db/db.go
```diff
@@ -91,9 +91,11 @@ func (tx *Tx) PutOutgoingTransaction(transaction *types.TransactionWithMetadata)
 
 type byNonce []*types.TransactionWithMetadata
 
-func (txs byNonce) Len() int           { return len(txs) }
-func (txs byNonce) Less(i, j int) bool { return txs[i].Transaction.Nonce() < txs[j].Transaction.Nonce() }
-func (txs byNonce) Swap(i, j int)      { txs[i], txs[j] = txs[j], txs[i] }
+func (txs byNonce) Len() int      { return len(txs) }
```
benma, line 94: what happened here? why dfferent order, and why reformatted?

backend/devices/bitbox/device.go
```diff
@@ -859,7 +859,7 @@ func (dbb *Device) Random(typ string) (string, error) {
 	dbb.log.WithField("random", rand).Debug("Generated random")
 	if len(rand) != 32 {
 		dbb.log.WithField("random-length", len(rand)).Error("Unexpected length: expected 32 bytes")
-		return "", fmt.Errorf("unexpected length, expected 32, got %d", len(rand))
+		return "", fmt.Errorf("unexpected length, expected 32, got %d", len(rand)) //nolint:goerr113
```
benma, line 862: what does the linter want here instead?

what about disabling the goerr113 linter, seems there is no way to do it right except when we get rid of `errp` or extend it to be 1.13 compatible? we also need to be careful to keep Go1.12 working for bitbox02-api-go until GopherJS supports newer.

.github/workflows/ci.yml
```diff
@@ -23,6 +23,16 @@ env:
   TRAVIS_BUILD_DIR: ${{github.workspace}}
 
 jobs:
+  golangci:
```
benma, line 26: how does this even work? there is no `run`.

.github/workflows/ci.yml
```diff
@@ -23,6 +23,16 @@ env:
...
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v2
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v1
+        with:
+          # Keep this in sync with the version in envinit make target.
```
benma, line 34: mention the same in the Makefile, to update here?

why not use docker here as well like with the rest, so then they are automatically in sync?

backend/coins/eth/db/db.go
x1ddos: that's more recent goimports, which newever golangci-lint uses, too.
`go get -u golang.org/x/tools/cmd/goimports` and try formatting. you should get the same result.

re: different order. it just looked nicer. I can revert but `Less` method would still be formated on multiline.

.github/workflows/ci.yml
x1ddos: welcome to the confusing world of github actions :sweat_smile: 

`golangci` is an arbitrary job name here but below, where it says "Run golangci-lint", there's no `:run` like in the others because it has `uses: golangci/golangci-lint-action@v1`. they are exclusive.

that tells github: go fetch from github.com/golangci/golangci-lint-action at that "version" and run their script.

.github/workflows/ci.yml
x1ddos: this doesn't run on docker. it's a separate thing, what github calls "workflow action". it's just a parameter for whatever script is supposed to run from github.com/golangci/golangci-lint-action, like a command line argument.

I added this step to have nice annotations right in pull requests, but it always requires a version to be specified.

made a reminder note in the Makefile as well, to keep it in sync with this one.

backend/devices/bitbox/device.go
x1ddos: oh, very good point about gopherjs. I completely forgot tbh.
reverted, disabled goerr113 and added a comment in golangci.yml.

which reminds me, this line should read go1.12: https://github.com/digitalbitbox/bitbox02-api-go/blob/8e5a24b5/.travis.yml#L12, and we probably need to add a build on go1.12 here too.

x1ddos:
> Pushd `shiftcrypto/bitbox-wallet-app:6` to dockerhub, pls update the ci scripts to use it.

oh, I was hoping you'd push it to github too:

> Instructions on how to update github's image copy are here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/0e0543f/.github/workflows/ci.yml#L6-L22.

x1ddos:
Either way PTAL. Changes after first review: https://github.com/digitalbitbox/bitbox-wallet-app/compare/07fb97f8e939f68dc4c30e78ab2f0e8b3365df8a..5409937ad73c1aed147a39d6d6a8cbb603518646

backend/devices/bitbox/device.go
benma: or better yet, actually do the js compilation as part of CI.

.github/workflows/ci.yml
benma: and they know it's a Go program by inspecting the source? :thinking:

backend/devices/bitbox/device.go
x1ddos: ah, js compilation already happens but in api-js repo: https://github.com/digitalbitbox/bitbox02-api-js/blob/d9b4718b/.github/workflows/ci.yml#L16-L20

but another good point indeed. best to fail right away in api-go if it's already known we won't be able to compile in api-js.

.github/workflows/ci.yml
x1ddos: well, `golangci-lint` by definition runs only against Go code, right? :)
also, they of course use our `.golangci.yml` config which tells it what to check and not to.

x1ddos: of course, if you put it against api-js repo, I doubt it'll have any useful effect :smile:

x1ddos:
Plus, one other advantage of having the golangci github action is only 16s is needed to find out if something's wrong with Go code:

image

instead of 4-5min.

benma:
> Either way PTAL. Changes after first review: https://github.com/digitalbitbox/bitbox-wallet-app/compare/07fb97f8e939f68dc4c30e78ab2f0e8b3365df8a..5409937ad73c1aed147a39d6d6a8cbb603518646

Wrong diff or did the previous version really accidentally revert a couple of commits (guide entry and Hebrew)?

Anyway, LGTM.

x1ddos:
> > Either way PTAL. Changes after first review: https://github.com/digitalbitbox/bitbox-wallet-app/compare/07fb97f8e939f68dc4c30e78ab2f0e8b3365df8a..5409937ad73c1aed147a39d6d6a8cbb603518646
> 
> Wrong diff or did the previous version really accidentally revert a couple of commits (guide entry and Hebrew)?
> 
> Anyway, LGTM.

Ah, that's because I rebased on top of most recent master which already had those other commits. Yeah, did I already mention gerrit? :sweat_smile:

x1ddos:
One character typo change in commit msg, wait for gree and then submit.

PR #924, opened 2020-07-07
frontend: add guide entry to receive about address formats

benma:
Only for unified accounts, where the user can actually choose the format.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -793,6 +793,10 @@
         "text": "As soon as you transact, a new address is automatically added to the list so there are always 20 addresses available which have never received any coins.",
         "title": "When do the addresses change?"
       },
+      "addressFormats": {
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -793,6 +793,10 @@
         "text": "As soon as you transact, a new address is automatically added to the list so there are always 20 addresses available which have never received any coins.",
         "title": "When do the addresses change?"
       },
+      "addressFormats": {
+        "text": "A compatible address is standard segwit whereas the modern address is native segwit (bech32). By default the BitBox02 uses the modern address. If you are having issues sending to this address, click \"change to compatible address\".",
```
thisconnect, line 797: Nit: I believe we write Segwit with capital S, most of the time, but there are a few occurences of segwit lowercase.

benma: should probably be lowercase everywhere else?

thisconnect: Segwit is a name like Bitcoin always capital first letter IMO

PR #927, opened 2020-07-08
frontend/account: show number of synced addresses in the spinner

benma:
On a slow network or on Tor, the initial sync can take a very long
time (minutes). After 10 or 20 seconds without any progress
indication, the user is left guessing if it will ever finish.

Unforuntately, the number of addresses that will be synced is not
known upfront, it could be 20 or 1000 addresses, and it is a rolling
window until there are 20 (or whatever the gap limit) unused addresses
in a row.

We could show `synced X out of Y` addresses, which both numbers
creeping up, but after some discussion decided that this does not help
more than just showing the number of addresses synced. The user still
does not know when it will finish, but at least sees that there is
progress.

The additional line is only shown for >1 synced
addresses. Single-address accounts like address-watchonly or Ethereum
do not show this progress line.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -86,7 +86,11 @@ type Account struct {
 
 	subaccounts []subaccount
 	// How many addresses were synced already during the initial sync. This value is emitted as an
-	// event when it changes.
+	// event when it changes. This counter can overshoot if an address is updated more than once
+	// during initial sync (e.g. if there is a tx touching it). This should be rare and have no bad
+	// consequence, as the counter is only used to display absolute progress in the frontend. If you
+	// need an accurate count of addresses synced, this should probably be turned into a map (set)
+	// instead.
 	syncedAddressesCount uint32
 
 	transactions *transactions.Transactions
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path"
 	"sort"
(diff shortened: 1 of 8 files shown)


backend/coins/btc/account.go
```diff
@@ -603,6 +610,17 @@ func (account *Account) Balance() (*accounts.Balance, error) {
 	return account.transactions.Balance(), nil
 }
 
+func (account *Account) incAndEmitSyncCounter() {
+	if !account.synced {
```
x1ddos, line 614: It seems like `synced` is updated in multiple places, some of which look like the they run in separate goroutines. If so, wouldn't it need to be guarded by a lock?

I see `Account` already embeds `locker.Locker` but it's not clear at all what it guards. Grepping its usage shows `Initialize`, `updateFeeTargets` methods and some closure funcs here and there. Can't tell for sure.

x1ddos: If `synced` really needs a lock, there are a few other places to update, so maybe in a separate commit/PR, but I wish the mutex embeded in the `Account` had some docs.

benma:
@x1ddos added one more commit with a better comment, PTAL.

backend/coins/btc/account.go
benma: Made an issue for it as it is a bit bigger than this PR.

x1ddos: :+1:

PR #928, opened 2020-07-08
Makefile: fix servewallet commands

benma:
Regression from 401bf795715738d5f47e95f4b060ac2a91a09ba9 where the
args were dropped.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,11 +41,11 @@ osx-init:
 servewallet:
 	go run -mod=vendor ./cmd/servewallet
 servewallet-mainnet:
-	go run -mod=vendor ./cmd/servewallet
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,13 +39,13 @@ osx-init:
 	brew install go@1.14
 	$(MAKE) envinit
 servewallet:
-	go run -mod=vendor ./cmd/servewallet
+	go run -mod=vendor ./cmd/servewallet/...
```
x1ddos, line 42: I don't believe you need triple dot. Does it not work for you without it?

benma: Right, fixed

benma:
> OMG so sorry I dropped the args see_no_evil

Sorry for the careless review ;)

PR #930, opened 2020-07-09
frontend: remove unused property showLogo in Spinner

benma:
Last use of showLogo was in 7f3f21b937e982d9c82e1967efbcf157f23a5ccf.


benma:
@thisconnect maybe you remember why you dropped the logo there, but imho by now it probably doesn't matter anymore, and we can just remove unused stuff.

thisconnect:
> @thisconnect maybe you remember why you dropped the logo there, but imho by now it probably doesn't matter anymore, and we can just remove unused stuff.

IIRC the reason was that you see a logo already and there is no added value is showing the logo twice.

PR #933, opened 2020-07-15
backend/bitbox01: restore longer timeout when starting the pairing

benma:
Regression from 53466beec694c0bfdfdd80c5ba4ae10ddba6ff52.

Was accidentally reduced to 10s (maybe a debug leftover).


jstrnbrg:
@benma the error message in the App is "The pairing timed out after two minutes. Start again if you still want to pair the mobile app." So either update the error message as well or change the timeout to 2 min in the code so that it is consistent

benma:
> @benma the error message in the App is "The pairing timed out after two minutes. Start again if you still want to pair the mobile app." So either update the error message as well or change the timeout to 2 min in the code so that it is consistent

good catch, increased to 2min. Easier than fixing all the translations.

PR #934, opened 2020-07-15
cmd/servewallet: strip charset from LANG in NativeLocale

x1ddos:
Some locale-involved formatting does not support charset appended
to a "code_territory" locale name and breaks the UI in
unpredictable ways when in servewallet dev mode.

In production, Qt ignores charset:
https://doc.qt.io/qt-5/qml-qtqml-qt.html#locale-method

Android doesn't seem to even mention charset:
https://developer.android.com/reference/java/util/Locale.html

### DIFF
--- a/cmd/servewallet/main.go
+++ b/cmd/servewallet/main.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strings"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/arguments"
@@ -82,10 +83,13 @@ func (webdevEnvironment) UsingMobileData() bool {
 // NativeLocale naively implements backend.Environment.
 // This version is unlikely to work on Windows.
 func (webdevEnvironment) NativeLocale() string {
-	if v := os.Getenv("LC_ALL"); v != "" {
-		return v
+	v := os.Getenv("LC_ALL")
+	if v == "" {
+		v = os.Getenv("LANG")
 	}
-	return os.Getenv("LANG")
+	// Strip charset from the LANG. It is unsupported by the i18next package
+	// used in the frontend.
+	return strings.Split(v, ".")[0]
 }
 
 func main() {


benma:
> It seems i18next package used in the frontend does not support charset
appended to a "code_territory" 

It actually does, `fr-FR.UTF-8` is accepted by i18next and the language is set correctly. The problem is that we use the same language string as-is in other contexts (date formatting, nothing to do with i18next).

x1ddos:
> The problem was that we use the same language string as-is in other contexts (date formatting, nothing to do with i18next).

Hm... then I didn't get the problem. Where is it used as-is?

x1ddos:
> > The problem was that we use the same language string as-is in other contexts (date formatting, nothing to do with i18next).
> 
> Hm... then I didn't get the problem. Where is it used as-is?

or rather, what breaks exactly?

x1ddos:
> > > The problem was that we use the same language string as-is in other contexts (date formatting, nothing to do with i18next).
> > 
> > 
> > Hm... then I didn't get the problem. Where is it used as-is?
> 
> or rather, what breaks exactly?

Anyway, updated the commit.

benma:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/08e3c28f0a17cfcef0c5bcc3bee73dd7c88aa809/frontends/web/src/components/transactions/transaction.tsx#L74

`toLocaleString()` doesn't work with the charset. The `i18n.language` is whatever we set it here:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/ef22a19f8cdbde4cc8476c7e0c62efa804a7f636/frontends/web/src/i18n/config.js#L41

That i18next is storing the language string as-is and does strip the charset is not necessarily wrong. i18next still behaves correctly.

x1ddos:
Nice. From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString:

> locales: A string with a BCP 47 language tag

So, I guess this should all work fine now. It was only the `LANG` variable on linux which may include the extra `.charset`.

PR #936, opened 2020-07-16
frontend/start: remove moreCoins guide entry

benma:


thisconnect:
LGTM what about the other languages, can you remove them in locize?

benma:
> LGTM what about the other languages, can you remove them in locize?

they are removed by locize and will be gone in the repo with the next locize pull.

PR #939, opened 2020-07-28
Frontend labels

thisconnect:


thisconnect:
I'll squash i18n in the last commit…

benma:
The 'save' console log entry appears only every 2nd time I click the save-icon. Clicking it once saves, then I edit again and click save again; nothing happens.

Also: Hitting `[Enter]` should save too imho.

benma:
Small bug: when I want to delete a note, I enter `backspace` to delete all chars, and at the last one when it is empty, the full note is restored.

I want to be able to save an empty note (no note).

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -223,6 +222,7 @@ class Send extends Component<Props, State> {
                     fiatAmount: undefined,
                     amount: undefined,
                     data: undefined,
+                    note: '',
```
benma, line 225: please add the real note here

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -279,6 +280,7 @@ class Send extends Component<Props, State> {
                     proposedFee: result.fee,
                     proposedAmount: result.amount,
                     proposedTotal: result.total,
+                    note: '',
```
benma, line 283: what's this for? this should not reset the state.

with this, the note is deleted if you modify the amount for example.

thisconnect:
> The 'save' console log entry appears only every 2nd time I click the save-icon. Clicking it once saves, then I edit again and click save again; nothing happens.

After saving the icon

> 
> Also: Hitting `[Enter]` should save too imho.

frontends/web/src/routes/account/send/send.tsx
thisconnect: This is the default

frontends/web/src/routes/account/send/send.tsx
thisconnect: fixed

benma:
Hmm why did you close it?

thisconnect:
> The 'save' console log entry appears only every 2nd time I click the save-icon. Clicking it once saves, then I edit again and click save again; nothing happens.
> 
> Also: Hitting `[Enter]` should save too imho.

fixed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/946/commits/c8bc611f8085b11a552fbe56e210fd14797932e9

thisconnect:
> Small bug: when I want to delete a note, I enter `backspace` to delete all chars, and at the last one when it is empty, the full note is restored.
> 
> I want to be able to save an empty note (no note).

fixed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/946/commits/1035a8884279e0da5cd4bdebdeff9acb86899b88

PR #944, opened 2020-08-03
Notes on the backend

benma:


backend/accounts/notes/notes.go
```diff
@@ -0,0 +1,104 @@
...
+// not exist, no error is returned.  Returns an error for other kinds of file read errors.
+func LoadNotes(filename string) (*Notes, error) {
+	notes := &Notes{
+		filename: filename,
+	}
+	data, err := read(filename)
+	if err != nil {
+		return nil, err
+	}
+	notes.data = data
```
x1ddos, line 82: Or you could create the struct right here, when everything's ready:

```go
return &Notes{data: data}, nil
```

benma: very good! fixed

NickeZ:
:amaze:

PR #945, opened 2020-08-03
backend/accounts: expose notes to frontend

benma:


backend/accounts/account.go
```diff
@@ -65,6 +66,10 @@ type Interface interface {
 	// Safello returns the infos needed to load the Safello Buy widget. panics() if Safello is not
 	// supported for this coin. Check support with `SafelloBuySupported()` before calling this.
 	SafelloBuy() *safello.Buy
+
+	Notes() *notes.Notes
+	// SetTxNote sets a tx note and refreshes the account.
+	SetTxNote(txID string, note string) error
```
x1ddos, line 72: Can also type `(txID, note string)`

PR #946, opened 2020-08-04
Frontend labels

thisconnect:
Not sure why GH closed https://github.com/digitalbitbox/bitbox-wallet-app/pull/939

### DIFF
--- a/frontends/web/src/components/transactions/transaction.tsx
+++ b/frontends/web/src/components/transactions/transaction.tsx
@@ -19,6 +19,7 @@ import { Component, h, RenderableProps } from 'preact';
 import { Input } from '../../components/forms';
 import { translate, TranslateProps } from '../../decorators/translate';
 import { AmountWithConversions } from '../../routes/account/send/send';
+import { apiPost } from '../../utils/request';
 import A from '../anchor/anchor';
 import { Dialog } from '../dialog/dialog';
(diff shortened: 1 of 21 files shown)


frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -127,35 +173,41 @@ class Transaction extends Component<Props, State> {
...
+                        { note ? (
+                            <div className={parentStyle.activity}>
+                                <span className={style.address}>
+                                    {note}
+                                </span>
+                            </div>
+                        ) : (
+                            <div className={parentStyle.activity}>
+                                <span className={style.label}>
+                                    {t(`transaction.tx.${type === 'receive' ? 'received' : 'sent'}`)}
```
benma, line 189: it is better to keep the full t(...) string always explicit, so it is easier to look for them later, or to automatically find unused ones etc. so better to do:

`type === 'receive' ? t('transaction.tx.received') : t('transaction.tx.sent')`.

thisconnect: You are right searching for `transaction.tx.received` is nice. I'll change it to:

`t(type === 'receive' ? 'transaction.tx.received' : 'transaction.tx.sent')`

thisconnect:
PTAL (Besides tests)

PR #947, opened 2020-08-05
Sendtx note

benma:


benma:
I accidentally squashed the storing of the tx note into the first commit.. Please don't make me undo it :D

backend/coins/btc/handlers/handlers.go
```diff
@@ -397,6 +400,7 @@ func (handlers *Handlers) postAccountTxProposal(r *http.Request) (interface{}, e
 		input.feeTargetCode,
 		input.selectedUTXOs,
 		input.data,
+		input.note,
```
x1ddos: I would've actually left `TxProposal` as is and call `SetTxNote` here in the handler:

```go
outputAmount, fee, total, err := handlers.account.TxProposal(addr, amount, fee, utxo, data)
if err != nil {
  return txProposalError(err)
}
err := handlers.account.SetTxNote(input.note)
if err != nil {
  return txProposalError(err)
}
return ...
```

It's just strange to see `TxProposal` taking a `note` where it doesn't really care about it for the purposes of a transaction proposal. Feels like `note` is part of a TX proposal.

backend/accounts/account.go
```diff
@@ -52,8 +62,7 @@ type Interface interface {
 	// SendTx signs and sends the active tx proposal, set by TxProposal. Errors if none available.
 	SendTx() error
 	FeeTargets() ([]FeeTarget, FeeTargetCode)
-	TxProposal(string, coin.SendAmount, FeeTargetCode, map[wire.OutPoint]struct{}, []byte, string) (
-		coin.Amount, coin.Amount, coin.Amount, error)
+	TxProposal(*TxProposalArgs) (coin.Amount, coin.Amount, coin.Amount, error)
```
x1ddos, line 65: This is a very nice change! Now I know what each one means when looking at the `TxProposal` here in the interface. Would be cool if returnd values were also named. Unclear which `coin.Amount` is which.

backend/coins/btc/handlers/handlers.go
benma: That breaks the rule that the handler *only* does http input parsing and output serializing, not doing any logic, except for calling the counterpart on the object it exposes to the frontend (Account in this case).

benma: Maybe TxProposal() can be renamed? In the end it just saves all the input data for use when SendTx() is called, which includes the note.

x1ddos: Can we abandon that rule? :)

benma: most definitely not lol. that would be mixing of concerns, which never ends well.

x1ddos: I see one concern of calculating a transaction and another one is attaching notes/labels to transactions. Very different and independent concepts. They need not know about each other.

```
backend/coins/btc/maketx/maketx.go
32:// TxProposal is the data needed for a new transaction to be able to display it and sign it.
33:type TxProposal struct {

88:// SetTxNote stores a note for a transaction. An empty note will result in the entry being deleted
91:func (notes *Notes) SetTxNote(txID string, note string) error {
```

Yet, because of a "rule", these two are mixed together. Even if `TxProposal` was named a `TxProposalWithNote`, it's still a mix. The code in question need not know about a text note when calculating a transaction.

PR #952, opened 2020-08-10
backend/btc/transactions: store tx first seen time and sort by it

benma:
Unconfirmed transactions would appear in an arbitrary order in the
app. This commit adds a time of first seen to each tx, and sorts by
that instead, so even unconfirmed transactions are ordered by age.

Same for transactions that confirmed in the same block, which are also
sorted in no particular order currently.

### DIFF
--- a/backend/coins/btc/db/transactionsdb/transactionsdb.go
+++ b/backend/coins/btc/db/transactionsdb/transactionsdb.go
@@ -123,11 +123,12 @@ func (tx *Tx) Commit() error {
 }
 
 type walletTransaction struct {
-	Tx              *wire.MsgTx
-	Height          int
-	Addresses       map[string]bool `json:"addresses"`
-	Verified        *bool
-	HeaderTimestamp *time.Time `json:"ts"`
+	Tx               *wire.MsgTx
+	Height           int
+	Addresses        map[string]bool `json:"addresses"`
+	Verified         *bool
+	HeaderTimestamp  *time.Time `json:"ts"`
+	CreatedTimestamp *time.Time `json:"created"`
 }
 
 func newWalletTransaction() *walletTransaction {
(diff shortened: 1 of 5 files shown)


backend/coins/btc/db/transactionsdb/transactionsdb.go
```diff
@@ -155,24 +156,30 @@ func writeJSON(bucket *bbolt.Bucket, key []byte, value interface{}) error {
...
+		walletTx.CreatedTimestamp = &now
+	}
 	f(walletTx)
 	return writeJSON(tx.bucketTransactions, key, walletTx)
 }
 
 // TxInfo implements transactions.DBTxInterface.
-func (tx *Tx) TxInfo(txHash chainhash.Hash) (*wire.MsgTx, []string, int, *time.Time, error) {
+func (tx *Tx) TxInfo(txHash chainhash.Hash) (
+	*wire.MsgTx, []string, int, *time.Time, *time.Time, error) {
```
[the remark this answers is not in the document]
x1ddos: Better yet, would it make sense to turn the returned value into a `TxInfo` struct? I would then rename the method from `TxInfo` to just `Info`.

backend/coins/btc/db/transactionsdb/transactionsdb_test.go
```diff
@@ -146,9 +149,12 @@ func TestTxQuick(t *testing.T) {
 				require.NoError(t, tx.MarkTxVerified(txHash, expectedHeaderTimestamp))
 				delete(allUnverifiedTxHashes, txHash)
 				require.True(t, checkTxHashes())
-				_, _, _, headerTimestamp, err := tx.TxInfo(txHash)
+				_, _, _, headerTimestamp, createdTimestamp, err := tx.TxInfo(txHash)
 				require.NoError(t, err)
 				require.Equal(t, expectedHeaderTimestamp.String(), headerTimestamp.String())
+				now := time.Now()
+				require.NotNil(t, createdTimestamp)
```
x1ddos, line 156: For existing app installs, this will be `nil`, I guess. What happens then, does everything work?

benma: Yeah, this PR has all instances of accessing this field. They all check for nil.

backend/coins/btc/transactions/transactions.go
```diff
@@ -394,11 +394,16 @@ func (transactions *Transactions) Balance() *accounts.Balance {
...
-// height. Special case for unconfirmed transactions (height <=0), which come last.
+// height. Special case for unconfirmed transactions (height <=0), which come last. If the height
+// is the same for two txs, they are sorted by the created (first seen) time instead.
 type byHeight []*TxInfo
 
 func (s byHeight) Len() int { return len(s) }
 func (s byHeight) Less(i, j int) bool {
+	// Secondary sort by the time we've first seen the tx in the app.
+	if s[i].Height == s[j].Height && s[i].createdTimestamp != nil && s[j].createdTimestamp != nil {
+		return s[i].createdTimestamp.Before(*s[j].createdTimestamp)
```
x1ddos, line 405: Wouldn't you want to sort by `createdTimestamp` even if `s[x].Height <=0 `?

backend/coins/btc/db/transactionsdb/transactionsdb_test.go
x1ddos: Ah sorry! I somehow missed byHeight's sort.Interface.

backend/coins/btc/db/transactionsdb/transactionsdb.go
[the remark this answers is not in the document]
benma: Done (kept the method name though, since Go has no method overloading like Rust ;))

backend/coins/btc/transactions/transactions.go
benma: It is, but as a secondary sort. Primary sort is still the block height.

backend/coins/btc/db/transactionsdb/transactionsdb.go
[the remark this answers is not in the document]
x1ddos: True, although Go does have [gorename](https://godoc.org/golang.org/x/tools/cmd/gorename) tool.

PR #953, opened 2020-08-10
frontend: don't persist dismissal of bb02 promo code for bb01 users

benma:


frontends/web/src/components/status/status.tsx
```diff
@@ -89,15 +91,15 @@ export default class Status extends Component<Props, State> {
...
+        if (!show || (childrenList && childrenList.length === 1 && !childrenList[0])) {
             return null;
         }
         return (
             <div className={[style.container, style[type], className ? className : ''].join(' ')}>
                 <div className={style.status}>
                     {children}
                     {
-                        dismissable && (
+                        dismissable !== undefined && (
```
x1ddos, line 102: I'm not an expert, but can it also be `null`?

benma: No, that would have to be explicitly typed as `... | null`. the `?: ...` type means something or `undefined`.

PR #956, opened 2020-08-10
frontend/send: remove weird whitespace

benma:


benma:
Seems to be a [non-breaking whitespace](https://en.wikipedia.org/wiki/Non-breaking_space). @thisconnect maybe check your keyboard or editor config :joy:

PR #957, opened 2020-08-11
backend/accounts: separate tx proposal from note proposal

benma:
When making a tx in the send screen, a tx proposal is created when
form fields are changed, to store the tx proposal to be signed and
sent, and also to compute the tx fees that the frontend should
display. This endpoint can have a long delay, e.g. in Ethereum it does
network requests.

The note was also part of the tx proposal, but that lead to a needless
delay and a display of "calculating fees..." when typing a note.

This commit makes a separate tx note proposal, as it can always be
processed instantly.

### DIFF
--- a/backend/accounts/account.go
+++ b/backend/accounts/account.go
@@ -59,7 +59,8 @@ type Interface interface {
 	Notifier() Notifier
 	Transactions() ([]Transaction, error)
 	Balance() (*Balance, error)
-	// SendTx signs and sends the active tx proposal, set by TxProposal. Errors if none available.
+	// SendTx signs and sends the active tx proposal, set by TxProposal. Errors if none
+	// available. The note, if set by ProposeTxNote(), is persisted for the transaction.
 	SendTx() error
 	FeeTargets() ([]FeeTarget, FeeTargetCode)
 	TxProposal(*TxProposalArgs) (coin.Amount, coin.Amount, coin.Amount, error)
@@ -77,6 +78,9 @@ type Interface interface {
 	SafelloBuy() *safello.Buy
 
 	Notes() *notes.Notes
+	// ProposeTxnote stores a note. The note is is persisted in the notes database upon calling
+	// SendTx(). This function must be called before `SendTx()`.
+	ProposeTxNote(string)
 	// SetTxNote sets a tx note and refreshes the account.
 	SetTxNote(txID string, note string) error
 }
--- a/backend/accounts/baseaccount.go
+++ b/backend/accounts/baseaccount.go
@@ -17,6 +17,7 @@ package accounts
 import (
 	"fmt"
 	"path"
+	"sync"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/notes"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/synchronizer"
@@ -58,6 +59,9 @@ type BaseAccount struct {
 	offline bool
 
 	notes *notes.Notes
+
+	proposedTxNote   string
+	proposedTxNoteMu sync.Mutex
 }
 
 // NewBaseAccount creates a new Account instance.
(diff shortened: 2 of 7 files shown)


benma:
If the overall structure is okay, I will start writing unit tests for BaseAccount.

backend/accounts/baseaccount.go
```diff
@@ -133,6 +137,24 @@ func (account *BaseAccount) Notes() *notes.Notes {
...
+func (account *BaseAccount) ProposeTxNote(note string) {
+	account.proposedTxNoteMu.Lock()
+	defer account.proposedTxNoteMu.Unlock()
+
+	account.proposedTxNote = note
+}
+
+// GetAndClearProposedTxNote returns the note previously set using ProposeTxNote(). If none was set,
+// the empty string is returned. The proposed note is cleared by calling this function.
+func (account *BaseAccount) GetAndClearProposedTxNote() (string, error) {
```
x1ddos, line 150: Just string, no error? Simpler. No downsides, or so it seems.

backend/accounts/account.go
```diff
@@ -77,6 +78,9 @@ type Interface interface {
 	SafelloBuy() *safello.Buy
 
 	Notes() *notes.Notes
+	// ProposeTxnote stores a note. The note is is persisted in the notes database upon calling
+	// SendTx(). This function must be called before `SendTx()`.
+	ProposeTxNote(string)
```
x1ddos, line 83: It's interesting that this method is in the interface but `GetAndClearProposedTxNote` isn't. Maybe `ProposeTxNote` shouldn't be here either? Just feels a bit inconsistent.

Also, what if the the frontend simply makes to calls, send tx and then set note? You wouldn't even need to store a tx proposal note on the backend at all then.

x1ddos: "what if the frontend simply makes **two** calls"

x1ddos:
SourceForge really doesn't want to fix their TLS :hankey: 

```
$ curl -I 'https://astuteinternet.dl.sourceforge.net/project/nsis/NSIS%203/3.05/nsis-3.05-setup.e><
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.

$ showcert astuteinternet.dl.sourceforge.net -noout -dates
notBefore=May 10 03:04:22 2020 GMT
notAfter=Aug  8 03:04:22 2020 GMT
```

backend/accounts/baseaccount.go
benma: Ah yes, leftover from a previous version. fixed

backend/accounts/account.go
benma: It's in the interface if it is accessed by handlers.go, the only place where the interface is used.

> Also, what if the the frontend simply makes to calls, send tx and then set note? You wouldn't even need to store a tx proposal note on the backend at all then.

The frontend does not have the txID required to do that. And it seemed more robust this way, then the frontend cannot fuck up.

backend/accounts/baseaccount.go
```diff
@@ -133,6 +137,24 @@ func (account *BaseAccount) Notes() *notes.Notes {
...
+}
+
+// GetAndClearProposedTxNote returns the note previously set using ProposeTxNote(). If none was set,
+// the empty string is returned. The proposed note is cleared by calling this function.
+func (account *BaseAccount) GetAndClearProposedTxNote() string {
+	account.proposedTxNoteMu.Lock()
+	defer account.proposedTxNoteMu.Unlock()
+	note := account.proposedTxNote
+	account.proposedTxNote = ""
+	return note
```
x1ddos, line 155: This would also have worked:

```go
account.proposedTxNoteMu.Lock()
defer func() {
  account.propostedTxNote = ""
  account.proposedTxNoteMu.Unlock()
}()
return account.proposedTxNote
```

Here's an example: https://play.golang.org/p/D1gxxJ0PleD

benma: nice

x1ddos:
> SourceForge really doesn't want to fix their TLS :hankey: 

Turns out somebody has already reported it to them for exact same issue: https://sourceforge.net/p/forge/site-support/21088/

PR #958, opened 2020-08-11
Frontend long notes

thisconnect:
- support long multiline notes in tx overview
- better focus for editing long notes
- limit note to 256 chars

### DIFF
--- a/frontends/web/src/components/transactions/transaction.tsx
+++ b/frontends/web/src/components/transactions/transaction.tsx
@@ -258,6 +258,7 @@ class Transaction extends Component<Props, State> {
                                     transparent
                                     placeholder={t('note.input.placeholder')}
                                     value={newNote}
+                                    maxLength={256}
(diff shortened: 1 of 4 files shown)


frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -107,6 +107,7 @@ class Transaction extends Component<Props, State> {
                 note: this.state.newNote,
             });
         }
+        this.focusEdit();
         this.setState(
             ({ editMode }) => ({ editMode: !editMode }),
             this.focusEdit,
```
benma, line 113: what does this do actually?

thisconnect: tries to focus the input element, there is 2 issues (also see commit message) if the text is really long or sometimes the this.input isnt set or rerendered. But it still seems to not be able to focus when you reopen the same tx and edit again :(

Issue #938, opened 2020-07-25
App crashes while headphones plugged in on mobile.


TLausZ:
My headphones are plugged in my USB-C port. I open the app and it crashes silently after half of a second.

App: Android-4.20.1
Android V10 on Pixel 3
Kernel version: 
4.9.210-g28c696160049-ab6386370
#0 Fri Apr 10 23:00:22 UTC 2020

benma:
Interesting :thinking: 

What kind of headphones are they? Brand/model.

x1ddos:
Reproduced on Moto g7play Android 9, latest app v4.40.1. Here's what I see in adb logcat:

```
07-25 15:12:55.072  8608  8608 D ch.shiftcrypto.bitboxapp: XYZ: lifecycle: onCreate
07-25 15:12:55.370  1924  1967 I ActivityManager: Start proc 8629:com.android.chrome:webview_service/u0a139 for service com.android.chrome/org.chromium.android_webview.services.VariationsSeedServer
07-25 15:12:55.449  1924  1967 I ActivityManager: Start proc 8645:com.android.chrome:sandboxed_process0/u0i51 for webview_service ch.shiftcrypto.bitboxapp/org.chromium.content.app.SandboxedProcessService0
07-25 15:12:56.070  8608  8682 I GoLog   : Logging configuration taken from '/data/user/0/ch.shiftcrypto.bitboxapp/files/logging.json'.
07-25 15:12:56.071  8608  8682 I GoLog   : Logging into '/data/user/0/ch.shiftcrypto.bitboxapp/files/log.txt' from 'debug'.
07-25 15:12:56.071  8608  8608 I GoLog   : time="2020-07-25T13:12:56Z" level=info msg="--------------- Started application --------------" group=server
07-25 15:12:56.071  8608  8608 I GoLog   : <nil>
07-25 15:12:56.072  8608  8608 I GoLog   : time="2020-07-25T13:12:56Z" level=info msg=environment goarch=arm goos=android group=server version=4.20.1
07-25 15:12:56.072  8608  8608 I GoLog   : <nil>
[...]

PR #959, opened 2020-08-11
android: make sure device list is filtered by vendor/product ID

x1ddos:
Before this change, any device plugged into a USB-C would be accepted
as a BitBox02 device. This results in app crash if the the device isn't
actually a BitBox02 because Go's USB manager expects only valid devices
on the backend.

This commit simply ensures that the Android native code skips unknown
devices when looking for a BitBox02 in the list.

Closes https://github.com/digitalbitbox/bitbox-wallet-app/issues/938

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
@@ -119,8 +119,17 @@ public class GoViewModel extends AndroidViewModel {
             Iterator<UsbDevice> deviceIterator = deviceList.values().iterator();
             while (deviceIterator.hasNext()){
                 UsbDevice device = deviceIterator.next();
-                this.device = new GoDeviceInfo(device);
-                break; // only one device supported for now
+                // One other instance where we filter vendor/product IDs is in
+                // @xml/device_filter resource, which is used for USB_DEVICE_ATTACHED
+                // intent to launch the app when a device is plugged and the app is still
+                // closed. This filter, on the other hand, makes sure we feed only valid
+                // devices to the Go backend once the app is launched or opened.
+                //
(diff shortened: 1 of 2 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
```diff
@@ -119,8 +119,16 @@ public void updateDeviceList() {
             Iterator<UsbDevice> deviceIterator = deviceList.values().iterator();
             while (deviceIterator.hasNext()){
                 UsbDevice device = deviceIterator.next();
-                this.device = new GoDeviceInfo(device);
-                break; // only one device supported for now
+                // BitBox02 Vendor ID: 0x03eb, Product ID: 0x2403.
```
benma, line 122: So the device_filter.xml does not make sure that the deviceList only has devices with right vendor/product id? wtf?

anyway, make a comment in device_filter.xml that any changes have to be also done here?

x1ddos: device_filter.xml is to launch the app when it's still closed and you plug it the device first.

This one, updateDeviceList, is when the app is already launched and opened, and you plug in a device after. Well, it also runs when device_filter.xml is trigger but in this case there's no issue becaues device_filter.xml already filters by the known vendor/product ID.

Makes sense?

benma:
utACK (you did test it right? :D)

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
x1ddos: Slightly updated the comment. Hope it makes it better.

x1ddos:
> utACK (you did test it right? :D)

Yes I did... but it depends on what you mean by "tested" :)

PR #962, opened 2020-08-12
backend/accounts: add BaseAccount tests

benma:
As promised before @x1ddos

### DIFF
--- a/backend/accounts/baseaccount_test.go
+++ b/backend/accounts/baseaccount_test.go
@@ -1,0 +1,103 @@
+// Copyright 2020 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


backend/accounts/baseaccount_test.go
```diff
@@ -0,0 +1,103 @@
...
+		done()
+		require.Equal(t, EventSyncDone, checkEvent())
+
+		account.ResetSynced()
+		require.False(t, account.Synced())
+		account.ResetSynced()
+		require.False(t, account.Synced())
+	})
+
+	t.Run("offline", func(t *testing.T) {
```
x1ddos, line 79: Doesn't seem like "offline", "synchronizer" and "config" belong to TestNotes test or do they?

benma: Ah, forgot to update the test name. This function is like a test suite with a common setup part, like explained here:

https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

x1ddos: :+1:

PR #964, opened 2020-08-13
backend/accounts: export note as well in account export

benma:


benma:
> LGTM. No tests? :)

okay, coming ;)

PR #965, opened 2020-08-13
 backend/accounts: move export csv from handler to BaseAccount

benma:
The `moq` tool is used to generate a struct from an interface, much how we wrote them by hand previously.


backend/accounts/baseaccount.go
```diff
@@ -65,9 +67,10 @@ type BaseAccount struct {
 }
 
 // NewBaseAccount creates a new Account instance.
-func NewBaseAccount(config *AccountConfig, log *logrus.Entry) *BaseAccount {
+func NewBaseAccount(config *AccountConfig, coin coin.Coin, log *logrus.Entry) *BaseAccount {
```
x1ddos, line 75: So, can I provide a config with say `eth-erc20-usdt` code and a btc coin? If so, maybe it would make sense to enforce only compatible combinations? This is where a NewXxx function actualle comes in handy.

Makefile
```diff
@@ -25,6 +25,7 @@ envinit:
 	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOPATH)/bin v1.27.0
 	GO111MODULE=off go get -u github.com/stretchr/testify # needed for mockery
 	GO111MODULE=off go get -u github.com/vektra/mockery/...
+	GO111MODULE=off go get -u github.com/matryer/moq
```
x1ddos, line 28: What about the mockery just above moq?

mockery:

> mockery provides the ability to easily generate mocks for golang interfaces

moq:

> Interface mocking tool for go generate.

I thought one would be more than enough?

backend/accounts/baseaccount_test.go
```diff
@@ -46,7 +47,9 @@ func TestBaseAccount(t *testing.T) {
 		GetSigningConfigurations: nil,
 		GetNotifier:              nil,
 	}
-	account := NewBaseAccount(cfg, logging.Get().WithGroup("baseaccount_test"))
+
+	coin := &mocks.CoinMock{}
```
x1ddos, line 51: This would've worked as well but without any new tool dependency:

```go
type testCoin struct {
  code Code
  coinUnit, feeUnit, smallestUnit string
  coinDecimals, unitDecimals int
  urlPrefix string
}

func (tc *testCoin) Code() Code { return tc.code }
func (tc *testCoin) BlockExplorerTransactionURLPrefix() string { return tc.urlPrefix }
func (tc *testCoin) SmallestUnit() string { return tc.smallestUnit }
...
func (tc *testCoin) Initialize() {}
func (tc *testCoin) Close() error { return nil }
```

Write once and use everywhere in tests.
But I know you like mocks so won't insist.

backend/accounts/account.go
```diff
@@ -83,6 +85,9 @@ type Interface interface {
 	ProposeTxNote(string)
 	// SetTxNote sets a tx note and refreshes the account.
 	SetTxNote(txID string, note string) error
+
+	// ExportCSV exports the given transaction in CSV format (comma-separated).
+	ExportCSV(transactions []Transaction, w io.Writer) error
```
x1ddos, line 90: Go typically takes output in the first argument. Would've been more consistent with stdlib:

```go
ExportCSV(io.Writer, []Transaction)
```

backend/accounts/baseaccount.go
```diff
@@ -173,3 +178,62 @@ func (account *BaseAccount) SetTxNote(txID string, note string) error {
...
+			})
+			if err != nil {
+				return errp.WithStack(err)
+			}
+			// a multitx is output in one row per receive address. Show the tx fee only in the
+			// first row.
+			feeString = ""
+		}
+	}
+	return nil
```
x1ddos, line 238: Better check for flush error. It may actually fail. I'd flush here at the very end:

```go
writer.Flush()
return writer.Error()
```

> Error reports any error that has occurred during a previous Write or Flush. 

https://golang.org/pkg/encoding/csv/#Writer.Error

backend/coins/btc/handlers/handlers.go
```diff
@@ -222,65 +221,12 @@ func (handlers *Handlers) postExportTransactions(_ *http.Request) (interface{},
 		}
```
x1ddos, line 221: Wouldn't you want to return this error from the handler? If file.Close fails, the data might not have been written to the file. I'd move this out of defer, to the very end after ExportCSV.

Makefile
benma: mockery makes a different kind of mock struct, specifically to be used with testify mock.

moq creates them like we use to write them by hand: one struct field function per interface method.

i haven't settled yet on which is better, currently i prefer the latter ;)

backend/accounts/baseaccount_test.go
benma: what you wrote is basically what the new tool generates. 

i can also just not add it to the Makefile and use it locally instead of writing it by hand, would almost come out to the same result :P

backend/accounts/baseaccount.go
benma: A code is only an identifier, anything is valid. The caller of this function decides what to identify the accounts by.

The opposite would also be possible, to enumerate all possibilities here, and move them out from the backend. Not sure if that would be helpful, sounds less modular.

backend/accounts/baseaccount_test.go
x1ddos: yes, similar functionality except ~40 SLOC, no dependencies and easy to read VS ~400 SLOC, one extra tool and 10 sync.RWMutex. compare the above 1 line `Coin` method with the generated version:

```go
func (mock *CoinMock) Code() coin.Code {
	if mock.CodeFunc == nil {
		panic("CoinMock.CodeFunc: method is nil but Coin.Code was just called")
	}
	callInfo := struct {
	}{}
	lockCoinMockCode.Lock()
	mock.calls.Code = append(mock.calls.Code, callInfo)
	lockCoinMockCode.Unlock()
	return mock.CodeFunc()
}
``` 

yeah, you can say you'll never need to look at the generated source. but when a test fails, you might actually need to. plus, these giant generated methods may also introduce bugs. not today but when you or someone else `go get -u moq` next time and fetch a shiny new moq version to generate a brand new mock.

backend/accounts/baseaccount.go
x1ddos: Ah ok. I thought account code had some more strict meaning. Maybe rename it to just that in another PR, `Identifier`? to avoid any doubts.

backend/accounts/baseaccount_test.go
benma: on the flip side, if i know what the tool does, i don't have to read the code; but i'll have to read custom mock structs every time...

but sure, i can get rid of the dependency, and manually delete the mutexes, and end up with the code i would have written by hand. For example like done here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/blockchain/mocks/mock.go

x1ddos: > if i know what the tool does, i don't have to read the code; but i'll have to read custom mock structs every time...

well, I have to read any code every time when it isn't clear to me, including dubious function signatures like `func f() (string, string, string)` :)

but yeah, sure. I see it. good points. 

> For example like done here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/blockchain/mocks/mock.go

that one actually looks much nicer. but apart from mutexes, one other difference is it doesn't panic if a stub function doesn't exist. the moq's mock, on the other hand, does.

anyway, like I said I don't mind either way. in the end of the day, it's just my personal preference.

x1ddos: ok, after talking offline I like moq a bit more now :)

backend/accounts/account.go
benma: Done

backend/accounts/baseaccount.go
benma: Done

backend/coins/btc/handlers/handlers.go
benma: Done.

Should keep in mind though that errors returned like this from handlers are not handled by the frontend in the Qt build. It's a long-standing design issue that needs to be resolved some day.

Currently, when a frontend must handle errors, we return an error object like `{ "success": false, "code" ... }`.

backend/accounts/baseaccount.go
```diff
@@ -165,3 +178,63 @@ func (account *BaseAccount) SetTxNote(txID string, note string) error {
 	account.config.OnEvent(EventStatusChanged)
 	return nil
 }
+
+// ExportCSV implements accounts.Account.
+func (account *BaseAccount) ExportCSV(w io.Writer, transactions []Transaction) error {
+	writer := csv.NewWriter(w)
+	defer writer.Flush()
```
x1ddos, line 185: A leftover flush. It is now at the very end.

benma: uuh good catch, that made the Error() thing useless.

backend/accounts/transaction_mock.go
```diff
@@ -0,0 +1,395 @@
+// Code generated by moq; DO NOT EDIT.
```
benma, line 1: @x1ddos i committed this by accident :( it's not used anywhere. will delete soon

x1ddos: ah, wasn't sure.
sounds good! no worries. :+1:

PR #967, opened 2020-08-16
backend/accounts: get rid of the Transaction interface

benma:
It represents merely data, not any computations, so a data struct
suffices. Less interfaces make for easier to read code, and no mocking
needed in tests.


benma:
@x1ddos lmk if the dif is too big (it's quite big!). 

The `eth` account diff especially are maybe best compared with both versions side by side, the diff in GH looks more confusing than it actually is.

I will also review it once more to check that all the function-to-structfield moves are correct. A manual test of btc, eth and erc20 account looked good to me.

benma:
> I will also review it once more 

Done, still lgtm.

backend/accounts/transaction.go
```diff
@@ -54,40 +55,51 @@ type AddressAndAmount struct {
...
-// Transaction models a transaction with common transaction info.
-type Transaction interface {
+// TransactionData holds transaction data to be shown to the user. It is as coin-agnostic as
+// possible, but contains some fields that are only used by certain coins.
+type TransactionData struct {
 	// Fee is nil for a receiving tx. The fee is only displayed (and relevant) when sending funds
 	// from the wallet.
-	Fee() *coin.Amount
-
+	Fee *coin.Amount
```
x1ddos, line 63: I'd remove pointer. It's like a `type Point struct {x, y int}` or `time.Time`. 

```go
// Amount represents an amount in the smallest coin unit (e.g. satoshi).
type Amount struct {
	n *big.Int
}
```

So, we have pointer to `Amount` which has only one field, pointer to `big.Int`. If `Fee` is not `nil` but `Fee.n` is, that's kind of a weird and invalid state. Maybe a bit simplified version would be:

```go
type Amount big.Int
Fee *coin.Amount
```

then a Fee pointer field makes a bit more sense imho. Probably not in this PR though.

backend/accounts/transaction.go
```diff
@@ -54,40 +55,51 @@ type AddressAndAmount struct {
...
 	// Time of confirmation. nil for unconfirmed tx or when the headers are not synced yet.
-	Timestamp() *time.Time
-
+	Timestamp *time.Time
```
x1ddos, line 65: Could also be `Timestamp time.Time`, i.e. no pointer:

```go
// Is unconfirmed or not sync'ed yet?
Timestamp.IsZero()
```

backend/accounts/transaction.go
```diff
@@ -54,40 +55,51 @@ type AddressAndAmount struct {
...
-
+	Amount coin.Amount
 	// Addresses money was sent to / received on.
-	Addresses() []AddressAndAmount
+	Addresses []AddressAndAmount
+
+	// --- Fields only used by BTC follow:
+
+	// FeeRatePerKb is the fee rate of the tx (fee / tx size).
+	FeeRatePerKb *btcutil.Amount
```
x1ddos, line 92:
```go
$ go doc github.com/btcsuite/btcutil Amount

package btcutil // import "github.com/btcsuite/btcutil"

type Amount int64
    Amount represents the base bitcoin monetary unit (colloquially referred to
    as a `Satoshi'). A single Amount is equal to 1e-8 of a bitcoin.
```

Sounds very similar to our own `Amount`. Would be nice to use the same type and convert to `btcutil.Amount` only when needed. This is basically the `Fee` but specific to BTC, right.

x1ddos: I would remove the pointer.

backend/accounts/transaction.go
```diff
@@ -54,40 +55,51 @@ type AddressAndAmount struct {
...
+	VSize int64
+	// Size is the serialized tx size in bytes.
+	Size int64
+	// Weight is the tx weight.
+	Weight           int64
+	CreatedTimestamp *time.Time
+
+	// --- Fields only used for ETH follow
+
+	Gas uint64
```
x1ddos, line 104: I guess this is equivalent to BTC's `FeeRatePerKb`, right? So, `Fee` field is coin-independent which then translates into either `FeeRatePerKb` or `Gas` but somehow `Gas` is not a pointer and `FeeRatePerKb` is. Feels a bit inconsistent.

backend/coins/btc/transactions/transactions_test.go
```diff
@@ -184,7 +184,6 @@ func (s *transactionsSuite) TestUpdateAddressHistorySingleTxReceive() {
 	)
 	transactions := s.transactions.Transactions(func(blockchainpkg.ScriptHashHex) bool { return false })
 	require.Len(s.T(), transactions, 1)
-	require.Equal(s.T(), tx1, transactions[0].Tx)
```
x1ddos, line 187: So, how do you ensure `transactions[0]` is what's expected? Because the `Height`, balance and utxo match? Just checking.

backend/coins/btc/transactions/transactions_test.go
```diff
@@ -184,7 +184,6 @@ func (s *transactionsSuite) TestUpdateAddressHistorySingleTxReceive() {
...
 	require.Equal(s.T(), expectedHeight, transactions[0].Height)
```
x1ddos, line 187: `const expectedHeight = 10` above would make it slightly more reassuring it doesn't change during the test. github doesn't let me comment on unchanged lines :)

backend/coins/eth/types/types.go
```diff
@@ -85,138 +78,87 @@ func (txh *TransactionWithMetadata) UnmarshalJSON(input []byte) error {
...
 
-// InternalID implements accounts.Transaction.
-func (txh *TransactionWithMetadata) InternalID() string {
-	return txh.TxID()
-}
+	if erc20Token != nil {
+		// ERC20 transfer.
+		data := txh.Transaction.Data()
+		amount = coin.NewAmount(new(big.Int).SetBytes(data[len(data)-32:]))
+		address = common.BytesToAddress(data[4+32-common.AddressLength : 4+32]).Hex()
```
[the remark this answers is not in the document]
benma: We can alias the import name.

backend/coins/btc/handlers/handlers.go
```diff
@@ -153,48 +151,46 @@ func (handlers *Handlers) getAccountTransactions(_ *http.Request) (interface{},
...
-			Note:      handlers.account.Notes().TxNote(txInfo.InternalID()),
+			Note:      handlers.account.Notes().TxNote(txInfo.InternalID),
 		}
-		switch specificInfo := txInfo.(type) {
-		case *transactions.TxInfo:
-			txInfoJSON.VSize = specificInfo.VSize
-			txInfoJSON.Size = specificInfo.Size
-			txInfoJSON.Weight = specificInfo.Weight
-			feeRatePerKb := specificInfo.FeeRatePerKb()
+		switch handlers.account.Coin().(type) {
```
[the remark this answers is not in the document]
benma: Coin().Code needs a lot more entries in the enum: btc, tbtc, ltc, tltc, regtest, eth, teth, reth.

I want to only switch on the type of the coin, not all instances of it.

backend/coins/btc/handlers/handlers.go
```diff
@@ -153,48 +151,46 @@ func (handlers *Handlers) getAccountTransactions(_ *http.Request) (interface{},
...
+			txInfoJSON.VSize = txInfo.VSize
+			txInfoJSON.Size = txInfo.Size
+			txInfoJSON.Weight = txInfo.Weight
+			feeRatePerKb := txInfo.FeeRatePerKb
 			if feeRatePerKb != nil {
 				txInfoJSON.FeeRatePerKb = handlers.formatBTCAmountAsJSON(*feeRatePerKb, true)
 			}
-		case types.EthereumTransaction:
-			txInfoJSON.Gas = specificInfo.Gas()
+		case *eth.Coin:
```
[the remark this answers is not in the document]
benma: yeah, btc/handlers is actually an account handler. btc came first, when i added ETH i was trying to think of how to have a handlers for common stuff and then specific endpoints in the right package, but never got around to it.

maybe we can just move it to accounts/ where it is a better fit.

backend/accounts/transaction.go
benma: It's a pointer because it can be missing / nil.

The Amount field element should never be nil. 

That is a problem of Go: using pointers to save space vs. indicating that it is not present. They interfere with each other :/

backend/coins/eth/account.go
```diff
@@ -282,18 +282,17 @@ func (account *Account) outgoingTransactions(allTxs []accounts.Transaction) (
...
 	allTxHashes := map[string]struct{}{}
 	for _, tx := range allTxs {
-		allTxHashes[tx.TxID()] = struct{}{}
+		allTxHashes[tx.TxID] = struct{}{}
 	}
 
-	transactions := []accounts.Transaction{}
+	transactions := []*ethtypes.TransactionWithMetadata{}
 	for _, tx := range outgoingTransactions {
 		tx := tx
```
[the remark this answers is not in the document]
benma: True. A bit hard to tell sometimes :O

backend/accounts/transaction.go
benma: Will check in a new PR if I can improve it.

PR #968, opened 2020-08-17
Frontend: set focus on note input depending on editMode

thisconnect:
When the user opens the tx detail and the note is empty
it renders in edit mode. When the user clicks save with
an empty note the input was still editable.
This change sets autofocus and readonly only based on
the editMode, so that the save behavior is consistent.


benma:
@Stadicus please test this PR

Stadicus:
Works exactly as expected, LGTM

PR #970, opened 2020-08-17
frontend/account: show account info only for btc coins

benma:
In ETH/ERC-20, the account info screen does not show anything over the
receive screen.

Stripping down the account info page is out of scope for this commit.


thisconnect:
Didn't see that :)

PR #973, opened 2020-08-18
backend/accounts: test ExportCSV

benma:


backend/accounts/baseaccount_test.go
```diff
@@ -103,4 +109,44 @@ func TestBaseAccount(t *testing.T) {
...
+		const header = "Time,Type,Amount,Unit,Fee,Address,Transaction ID,Note\n"
+
+		require.Equal(t, header, export(nil))
+
+		require.NoError(t, account.notes.SetTxNote("some-internal-tx-id", "some note, with a comma"))
+		fee := coin.NewAmountFromInt64(101)
+		require.Equal(t,
+			header+
+				",sent,123,satoshi,101,some-address,some-tx-id,\"some note, with a comma\"\n"+
+				",sent_to_yourself,456,satoshi,,another-address,some-tx-id,\"some note, with a comma\"\n",
```
x1ddos, line 129: You can also use the backticks for slightly nicer reading:

```
`,sent_to_self,456,satoshi,,another-address,some-tx-id,"some note, with a comma"\n`
```

although not sure about `\n`. might not be interpreted correctly but worth a try.

Altertnatively,

```go
const body = `,sent,123,...
,sent_to_self,456,...
`
```

(no need for `\n`)

backend/accounts/baseaccount_test.go
```diff
@@ -103,4 +109,44 @@ func TestBaseAccount(t *testing.T) {
...
+
+		const header = "Time,Type,Amount,Unit,Fee,Address,Transaction ID,Note\n"
+
+		require.Equal(t, header, export(nil))
+
+		require.NoError(t, account.notes.SetTxNote("some-internal-tx-id", "some note, with a comma"))
+		fee := coin.NewAmountFromInt64(101)
+		require.Equal(t,
+			header+
+				",sent,123,satoshi,101,some-address,some-tx-id,\"some note, with a comma\"\n"+
```
x1ddos, line 128: would've been nice to also test the timestamp.

backend/accounts/baseaccount_test.go
benma: :+1: done

backend/accounts/baseaccount_test.go
benma: Done

PR #975, opened 2020-08-19
backend/btc: close db last

benma:
The db was closed before the Transactions instance was closed. This
could lead to a database access to a closed database in the
Transactions instance.


x1ddos (approved):
LGTM. Nice catch!
I checked and it seems no such issue with ETH accounts.

PR #976, opened 2020-08-19
backend/rates: too much logging

benma:
The log is a debug leftover, and does not need to be logged every
minute.

### DIFF
--- a/backend/rates/rates.go
+++ b/backend/rates/rates.go
@@ -73,10 +73,9 @@ func (updater *RateUpdater) update() {
 		strings.Join(coins, ","),
 		strings.Join(fiats, ","),
 	)
-	updater.log.Infof("Rates url: %v", url)
 	response, err := client.Get(url)
 	if err != nil {
-		updater.log.Printf("Error getting rates: %v\n", err)
+		updater.log.WithError(err).WithField("url", url).Error("Error getting rates")
 		updater.last = nil
 		return
 	}


backend/rates/rates.go
```diff
@@ -73,10 +73,10 @@ func (updater *RateUpdater) update() {
 		strings.Join(coins, ","),
 		strings.Join(fiats, ","),
 	)
-	updater.log.Infof("Rates url: %v", url)
 	response, err := client.Get(url)
 	if err != nil {
-		updater.log.Printf("Error getting rates: %v\n", err)
+		updater.log.Errorf("Error getting rates: %v\n", err)
+		updater.log.Infof("Rates url: %v", url)
```
x1ddos, line 79: I guess it's ok to log it separately but the idea is to log the URL only when an error happens. So, they kind of logically tied together. I would use a single log statement:

```go
updater.log.Errorf("Error getting rates from %q: %v", url, err)
```

backend/rates/rates.go
```diff
@@ -73,10 +73,10 @@ func (updater *RateUpdater) update() {
 		strings.Join(coins, ","),
 		strings.Join(fiats, ","),
 	)
-	updater.log.Infof("Rates url: %v", url)
 	response, err := client.Get(url)
 	if err != nil {
-		updater.log.Printf("Error getting rates: %v\n", err)
+		updater.log.Errorf("Error getting rates: %v\n", err)
```
x1ddos, line 78: No need for `\n`. Log formatting should always append one if missing.

x1ddos (approved):
Thank you!
LGTM

PR #977, opened 2020-08-19
config: add option to enable expert fee settings

benma:
This will enable setting custom fees as well as seeing the sat/vB fee
rate estimations for the four fee targets.

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -97,6 +97,10 @@ type Backend struct {
 	LitecoinActive bool `json:"litecoinActive"`
 	EthereumActive bool `json:"ethereumActive"`
 
+	// If enabled, the user will see sat/B fees in the fee target priorities, and be able to set a
+	// custom fee in sat/B.
+	ExpertFee bool `json:"expertFee"`
+
 	// Whether Bitcoin, Litecoin should be shown in multiple accounts - one per script type -
 	// instead of a combined account.
 	SplitAccounts bool `json:"splitAccounts"`
@@ -183,6 +187,8 @@ func NewDefaultAppConfig() AppConfig {
 			BitcoinActive:  true,
 			LitecoinActive: true,
 			EthereumActive: true,
+
+			ExpertFee: false,
 
 			SplitAccounts: false,
 
--- a/frontends/web/src/routes/settings/settings.tsx
+++ b/frontends/web/src/routes/settings/settings.tsx
@@ -105,6 +105,15 @@ class Settings extends Component<Props, State> {
             });
     }
 
+    private handleToggleExpertFee = (event: Event) => {
+        const target = (event.target as HTMLInputElement);
+        setConfig({
+            backend: {
+                [target.id]: target.checked,
+            },
+        }).then(config => this.setState({ config }));
+    }
+
     private handleToggleCoinControl = (event: Event) => {
         const target = (event.target as HTMLInputElement);
         setConfig({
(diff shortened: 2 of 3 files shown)


frontends/web/src/routes/settings/settings.tsx
```diff
@@ -362,6 +371,19 @@ class Settings extends Component<Props, State> {
                                                                 checked={config.backend.splitAccounts}
                                                                 onChange={this.handleToggleAccount} />
                                                         </div>
+                                                        <div className={style.currency}>
```
x1ddos, line 374: Funny that frontend uses a style name `currency` for expert settings.

frontends/web/src/routes/settings/settings.tsx
```diff
@@ -105,6 +105,15 @@ class Settings extends Component<Props, State> {
             });
     }
 
+    private handleToggleExpertFee = (event: Event) => {
+        const target = (event.target as HTMLInputElement);
+        setConfig({
+            backend: {
+                [target.id]: target.checked,
```
x1ddos, line 112: Seems like this is generic enough that the other `handleToggleCoinControl` and `handleToggleEthereum` can use it.

backend/config/config.go
```diff
@@ -97,6 +97,10 @@ type Backend struct {
 	LitecoinActive bool `json:"litecoinActive"`
 	EthereumActive bool `json:"ethereumActive"`
 
+	// If enabled, the user will see sat/B fees in the fee target priorities, and be able to set a
+	// custom fee in sat/B.
+	ExpertFee bool `json:"expertFee"`
```
x1ddos, line 102: Shouldn't it go to the `Frontend` part of the config? Unless the idea is to enforce it on the backend and refuse custom fee if this is unset. But maybe I'm missing something.

frontends/web/src/routes/settings/settings.tsx
x1ddos: @thisconnect ^^

frontends/web/src/routes/settings/settings.tsx
benma: It is funny, isn't it :joy:

frontends/web/src/routes/settings/settings.tsx
benma: one reinitializes accounts. the other doesn't. Not sure if refactoring this is worth it, but I defer ;)

backend/config/config.go
benma: Yeah, I add the custom fee target in the backend based on this, and forward the `expert` field in the same endpoint; see https://github.com/digitalbitbox/bitbox-wallet-app/pull/978.

I also had a version where I did everything in the frontend. Also possible. Not sure which is better, but I think I prefer Backend, so that more business logic is in the backend instead of in the frontend.

x1ddos: Ok but one day we'll need to talk about the opaque `Frontend` vs frontend settings in the `Backend` :)

x1ddos (approved):
LGTM unless @thisconnect has suggestions. Looks very similar to the other setting toggles though.

backend/config/config.go
benma: sure. but what frontend settings in the backend? Afaik all fields in the Backend config are used by the backend.

benma:
Merged. @thisconnect please add PRs to the staging branch for things that need to be improved.

backend/config/config.go
x1ddos: Like this one in https://github.com/digitalbitbox/bitbox-wallet-app/blob/dfde5c96d162b96cfd6295593a7f03035478388f/frontends/web/src/i18n/config.js#L31-L35

So, the frontend could also be like:

```js
apiGet('config').then(({ frontend }) => { frontend.expertFee }
```

benma: Still confused,`userLanguage` is only used in the frontend, `expertFee` is used in the backend. Let's take it offline.

Maybe it makes sense to structure the Frontend anyway instead of putting `interface{}`.

thisconnect:
> Merged. @thisconnect please add PRs to the staging branch for things that need to be improved.

https://github.com/digitalbitbox/bitbox-wallet-app/pull/979

but golang lint isn't happy :( https://github.com/digitalbitbox/bitbox-wallet-app/runs/1006638515

PR #978, opened 2020-08-19
Custom transaction fee support

benma:


benma:
https://github.com/digitalbitbox/bitbox-wallet-app/pull/977 is included here (this builds on top of that).

backend/accounts/feetarget.go
```diff
@@ -22,6 +22,7 @@ import (
 // Coin specific methods are implemented in corresponding coins.
 type FeeTarget interface {
 	Code() FeeTargetCode
+	FeeRateInfo() string
```
x1ddos, line 25: "info" to me usually means some structured data. something like `FmtFeeRate` or `FormattedFeeRate` with a comment `// xxx returns a formatted fee rate to display in the UI` might have been closer to what it does.

I mean here and below in `jsonFeeTarget`.

backend/coins/btc/handlers/handlers.go
```diff
@@ -348,18 +348,31 @@ func (handlers *Handlers) postAccountTxProposal(r *http.Request) (interface{}, e
...
 
+type jsonFeeTarget struct {
+	Code        accounts.FeeTargetCode `json:"code"`
+	FeeRateInfo string                 `json:"feeRateInfo"`
+}
+
 func (handlers *Handlers) getAccountFeeTargets(_ *http.Request) (interface{}, error) {
 	feeTargets, defaultFeeTarget := handlers.account.FeeTargets()
-	result := []map[string]interface{}{}
+	result := []jsonFeeTarget{}
```
x1ddos, line 358: :pray:

backend/coins/btc/handlers/handlers.go
```diff
@@ -348,18 +348,31 @@ func (handlers *Handlers) postAccountTxProposal(r *http.Request) (interface{}, e
 	}, nil
 }
 
+type jsonFeeTarget struct {
```
x1ddos, line 351: You don't have to define it globally. More recent Go versions allow for a type to be defined within a function.

frontends/web/src/routes/account/send/feetargets.tsx
```diff
@@ -38,10 +38,12 @@ export type Props = FeeTargetsProps & TranslateProps;
 
 interface FeeTarget {
     code: Code;
+    feeRateInfo: string;
 }
 interface State {
     feeTargets: FeeTarget[] | null;
     feeTarget?: string | null;
+    expert?: boolean;
```
x1ddos, line 46: Here's where I think `config.frontend.expertFee` would've made it slightly easier. Backend doesn't really care what the UI shows. If the backend receives `custom` fee target, it computes custom. Otherwise, one of the standard. And the UI shows what it needs to show.

backend/coins/btc/handlers/handlers.go
```diff
@@ -348,18 +348,31 @@ func (handlers *Handlers) postAccountTxProposal(r *http.Request) (interface{}, e
...
+	if expert {
+		result = append(result, jsonFeeTarget{
+			Code:        accounts.FeeTargetCodeCustom,
+			FeeRateInfo: "",
+		})
 	}
 	return map[string]interface{}{
 		"feeTargets":       result,
 		"defaultFeeTarget": defaultFeeTarget,
+		"expert":           expert,
```
x1ddos, line 375: Here's where I think `config.frontend.expertFee` would've made it slightly easier. Backend doesn't really care what the UI shows. If the backend receives `custom` fee target, it computes custom. Otherwise, one of the standard. And the UI shows what it needs to show.

Otherwise, backend need to propagate this `expert` for not much of a use, only for UI.

backend/coins/btc/handlers/handlers.go
```diff
@@ -348,18 +348,31 @@ func (handlers *Handlers) postAccountTxProposal(r *http.Request) (interface{}, e
...
-			map[string]interface{}{
-				"code": feeTarget.Code(),
-			})
+		result = append(result, jsonFeeTarget{
+			Code:        feeTarget.Code(),
+			FeeRateInfo: feeTarget.FeeRateInfo(),
+		})
+	}
+	expert := handlers.account.Config().ExpertFee
+	if expert {
```
x1ddos, line 366: Why not always append it and let the frontend deside whether to show it or not? Seems like the backend here directly dictates the frontend how to show controls in the UI.

x1ddos (approved):
LGTM but maybe @thisconnect has some JS/TS suggestions.

backend/coins/btc/handlers/handlers.go
benma: Wow this changes everything! When did that happen? Pretty sure that was not possible a year ago.

backend/coins/btc/handlers/handlers.go
benma: Okay, I'll move the setting to the frontend

x1ddos:
Looking at this example in https://golang.org/doc/go1.8#language, since 
at least 1.8 but I'm too lazy to dig through the release notes now :)

backend/coins/btc/handlers/handlers.go
benma: Done

backend/coins/btc/handlers/handlers.go
benma: Done, PTAL

backend/accounts/feetarget.go
benma: Done

x1ddos:
nice! does make the backend part a bit cleaner, no? slightly more work on the frontend but that's how it's supposed to be here imho.

x1ddos:
LGTM

x1ddos:
golangci failing for some weird reason. investigating...

x1ddos:
> golangci failing for some weird reason. investigating...

so, ignore it for now. test-lint also includes golangci.

benma:
golangci passes locally, so not a blocker.

benma:
webtest failing, gotta learn how to deal with that :nerd_face:

PR #981, opened 2020-08-20
backend/btc: too low fee error handling 

benma:
Builds on https://github.com/digitalbitbox/bitbox-wallet-app/pull/978/ - will rebase once the other one is done.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"sync"
 	"sync/atomic"
 
 	"github.com/btcsuite/btcd/chaincfg"
@@ -89,6 +90,9 @@ type Account struct {
 	activeTxProposalLock locker.Locker
 
 	feeTargets []*FeeTarget
+	// Access this only via getMinRelayFeeRate(). sat/kB.
+	minRelayFeeRate   *btcutil.Amount
+	minRelayFeeRateMu sync.Mutex
 
 	// true when initialized (Initialize() was called).
 	initialized bool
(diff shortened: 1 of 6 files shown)


benma:
Rebased

backend/coins/btc/account.go
```diff
@@ -234,6 +236,31 @@ func (account *Account) gapLimits(
...
 }
 
+// getMinRelayFeeRate fetches the min relay fee from the server and returns it. The value is cached
+// so that subsequent calls are instant. This is important as this function can be called many times
+// in succession when validating tx proposals.
+func (account *Account) getMinRelayFeeRate() btcutil.Amount {
+	unlock := account.RLock()
+	cached := account.minRelayFeeRate
+	unlock()
+	if cached != nil {
```
x1ddos, line 246: Seems like this is cached for the whole lifetime of the app. If I keep it running for hours, wouldn't it be pretty much a stale fee amount?

benma: It is the min relay fee, it is a bitcoin core server config that does not change at runtime, so it is stable.

Different servers could have different min relay fees, which is why I reset the value on a re-connect.

x1ddos: :+1: I misunderstood

backend/coins/btc/account.go
```diff
@@ -234,6 +236,31 @@ func (account *Account) gapLimits(
...
+		defer account.Lock()()
+		account.minRelayFeeRate = &feeRatePerKb
+		close(doneCh)
+	}, func(error) {})
+	<-doneCh
+
+	unlock = account.RLock()
+	defer unlock()
+	account.log.Infof("min relay fee rate: %s", *account.minRelayFeeRate)
+	return *account.minRelayFeeRate
```
x1ddos, line 261: Too many lock/unlocks imho. What is returned here might not match `account.coin.Blockchain().RelayFee(...)` result, in theory. In practice, like you said it's stable so doesn't matter much. It just leaves this feeling of not being entirely consistent.

If the next bitcoin core server version somehow makes the fee more dynamic, this might become an issue, in addition to the cache above.

I would always simply unlock at the method return. So, only one lock/unlock. The callers of this menthod always block and wait for the result anyway. But it would require a separate sync.Mutex though. Wouldn't want to lock the whole account.

x1ddos (approved):
LGTM although doesn't give me a great feeling of future proof. But I guess it's ok for now.

x1ddos:
Interesting, yet another issue with CI today:

```
mkdir %APPVEYOR_BUILD_FOLDER%
cp %GOPATH%\src\github.com\digitalbitbox\bitbox-wallet-app\frontends\qt\BitBox-installer.exe %APPVEYOR_BUILD_FOLDER%\BitBox-installer.exe
Discovering tests...OK
Collecting artifacts...
Found artifact 'BitBox-installer.exe' matching 'BitBox-installer.exe' path
Uploading artifacts...
Maximum allowed artifact storage size of 50000 Mb will be exceeded.
```

backend/coins/btc/account.go
benma: Totally right! Simplified it with a separate mutex. PTAL.

x1ddos: Much nicer! :+1: :pray:

x1ddos:
LGTM

PR #982, opened 2020-08-20
ci: delete troublesome golangci github action

x1ddos:
TL;DIR: we're stuck with go1.14 + linter v1.27 for now. The longer explanation below is for potential gotchas when and if we'll ever upgrade to go1.15 or a more recent golangci-lint.

golangci-lint always uses the latest version of Go, which has recently released go1.15. Any linter version starting with the currently in use 1.27 and up to the latest 1.30 fails on go1.15 with panics. Any linter version past 1.27 with go1.14 panics, too.

Unfortunately, golangci-lint always overwrites the Go version when run in a github action in https://github.com/golangci/golangci-lint-action/blob/v1.2.2/src/install.ts#L26. I made a working PoC which allows to specify a Go version in https://github.com/x1ddos/golangci-lint-action/commit/6f8897f but wouldn't be able to send them a pull request because they always work in the master branch. Their master branch is way ahead and [doesn't support 1.27 linter version anymore](https://github.com/golangci/golangci-lint-action/blob/1d3f25bff95b27bd81c4a177e585f69a0296368d/src/version.ts#L69-L75). So, my PoC is useless: we wouldn't want to maintain a fork of golangci-lint-action.

Note that golangci-lint is alwasy run in the docker container, too, in test-lint action anyway. What's removed here was a very quick separate CI step to identify potential pull request issues within seconds.

Alternatives are more time consuming and not worth it at the moment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,16 +25,6 @@ env:
   TRAVIS_BUILD_DIR: ${{github.workspace}}
 
 jobs:
-  golangci:
-    runs-on: ubuntu-18.04
-    steps:
-      - name: Clone the repo
-        uses: actions/checkout@v2
-      - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v1
-        with:
-          # Keep this in sync with the version in envinit make target.
-          version: v1.27
   test-lint:
     runs-on: ubuntu-18.04
     steps:


x1ddos:
@benma I think I know why appveyor's failing: we have probably exhaused 50Gb of artifact storage space limit.

```
Found artifact 'BitBox-installer.exe' matching 'BitBox-installer.exe' path
Uploading artifacts...
Maximum allowed artifact storage size of 50000 Mb will be exceeded.
```

But the account is on your name:

image

so I guess you'll have to login and remove some oldest manually.

x1ddos:
> Strange that one cannot specify the Go version used? Correct dependency versions are important for correct CI :S

Created an issue at https://github.com/golangci/golangci-lint-action/issues/75 but not sure it'll help us short term anyway. Best case is can always revert this commit but I wouldn't hold my breath.

> maybe we can change the appveyor script to not push an artefact on a normal PR, but just when merged to master or branches called `staging-*`? We barely ever (or never) use the builds random PRs.

I was going to suggest the same thing. Also for github. Will send you a separate PR.

PR #985, opened 2020-08-26
Staging frontend send customfees

thisconnect:


benma (approved):
Works! There is an issue with using `number` instead of `string` for the type, for example when adding `1.0` the `0` is instant-stripped, which is unusual. 

This needs a backend change though, so I can do it in another PR.

Can the input be limited to at most 3, or even 1 1 satdecimal places? `0.001 / vB` is the smallest denomination possible, as it is `1 sat/vkB`. If you can, I'd just limit it to one decimal place (0.1), I doubt anyone needs more granularity.

I also saw that some block explorers show the fee rate slightly differently in sat/vB (sub-satoshi difference), I will look into it, to see who is right :P

thisconnect:
> Can the input be limited to at most 3, or even 1 1 satdecimal places? `0.001 / vB` is the smallest denomination possible, as it is `1 sat/vkB`. If you can, I'd just limit it to one decimal place (0.1), I doubt anyone needs more granularity.

can do but then the +/- arrows use that for increasing/decreasing the value as well.

benma:
> > Can the input be limited to at most 3, or even 1 1 satdecimal places? `0.001 / vB` is the smallest denomination possible, as it is `1 sat/vkB`. If you can, I'd just limit it to one decimal place (0.1), I doubt anyone needs more granularity.
> 
> can do but then the +/- arrows use that for increasing/decreasing the value as well.

nevermind then ;)

PR #986, opened 2020-08-31
frontend: make locize-pull

benma:


frontends/web/src/locales/en/app.json
```diff
@@ -1172,7 +1172,6 @@
       "placeholder": "Enter hexadecimal data"
     },
     "error": {
-      "feeTooLow": "fee too low",
```
thisconnect, line 1175: why is feeTooLow gone again?

benma: hmm good question, i guess we pushed locize from master while this was in the staging branch :/ will fix

benma: First: https://github.com/digitalbitbox/bitbox-wallet-app/pull/987

benma:
Will re-open on `master` later.

PR #992, opened 2020-09-14
frontend/feeTargets: save the estimation translations

benma:
c38dd0fc63704d290f1b552c67a1022f108ff8bd changed the translation keys,
which means all translations are lost.

Since it is only a text modification, it is best to keep the keys, so
the previous translations are marked for review instead of being
deleted.

The context field can be be translated with `_context` keys,
e.g. `_ltc`. If the translation is missing, it falls back to the key
without context.


x1ddos (approved):
Nice. Thanks!
LGTM

PR #993, opened 2020-09-14
frontend/send: no autoformatting of custom fee rate

benma:
The `Number()` conversion automatically cut off e.g. a zero, so it was
impossible to enter `1.01` without it being truncated back to `1` when
entering `1.0`.

Also, a default rate of `1` might not be good.

Solution: treat the fee rate as a string, like the amount, and let the
backend convert/validate.

### DIFF
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


backend/coins/btc/handlers/handlers.go
```diff
@@ -288,7 +289,11 @@ func (input *sendTxInput) UnmarshalJSON(jsonBytes []byte) error {
 	if err != nil {
 		return errp.WithMessage(err, "Failed to retrieve fee target code")
 	}
-	input.FeePerKb = btcutil.Amount(jsonBody.FeePerByte * 1000)
+	feePerKb, err := strconv.ParseFloat(jsonBody.FeePerByte, 64)
+	if err != nil {
+		return err
```
x1ddos, line 294: A better option is probably to set `feePerKb` to 0 and continue. Otherwise, the app's state on the UI becomes inconsistent.

x1ddos (approved):
LGTM

PR #995, opened 2020-09-15
frontend: focus custom fee input and rearange fee info

thisconnect:
The input focus stopped working after last rearrangement.
This fix delays calling input.focus as somewhat the input
is not ready at render to be focused.

This change also swaps network fee info and conditional
estimated confirmation time, as the fee is always displayed
and it is less flickering.

Screen Shot 2020-09-16 at 8 03 52 AM

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -748,10 +748,10 @@ class Send extends Component<Props, State> {
                                 <div className={style.confirmItem}>
                                     <label>{t('send.amount.label')}</label>
                                     <p>
-                                        <span>{proposedAmount && proposedAmount.amount || 'N/A'} {proposedAmount && proposedAmount.unit || 'N/A'}</span>
+                                        <span>{proposedAmount && proposedAmount.amount || 'N/A'} <small>{proposedAmount && proposedAmount.unit || 'N/A'}</small></span>
                                         {
                                             proposedAmount && proposedAmount.conversions && (
-                                                <span> <span className="text-gray">/</span> {proposedAmount.conversions[fiatUnit]} {fiatUnit}</span>
+                                                <span> <span className="text-gray">/</span> {proposedAmount.conversions[fiatUnit]} <small>{fiatUnit}</small></span>
                                             )
                                         }
                                     </p>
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -765,12 +765,16 @@ class Send extends Component<Props, State> {
...
-                                            )
-                                        }
+                                        <span key="amount">{proposedFee && proposedFee.amount || 'N/A'} <small>{proposedFee && proposedFee.unit || 'N/A'}</small></span>
+                                        {proposedFee && proposedFee.conversions && (
+                                            <span key="conversation">
+                                                <span className="text-gray"> / </span>
+                                                {proposedFee.conversions[fiatUnit]} <small>{fiatUnit}</small>
+                                            </span>
+                                        )}
+                                        {feePerByte ? (
```
benma, line 775: you probably need to check that the fee target is set to custom as well?

otherwise, if you go to custom, set a fee per byte, then go back to a predefined target like 'low', then feePerByte might still be set to the previous value and shown in the dialog?

thisconnect: Right, now it always resets the feePerByte to '' when changing the feetarget
https://github.com/digitalbitbox/bitbox-wallet-app/pull/995/commits/e5adf72002409307ce96ee18b0fb3e6e1009ab0b ack?

benma: ack. need to test again with a device though.

thisconnect: I'm looking into passing the sat/vB info up to the parent component (sent.tsx) to display it in the confirm dialog…

thisconnect: > I'm looking into passing the sat/vB info up to the parent component (sent.tsx) to display it in the confirm dialog…

dropped, for now it will only show if custom fee was selected

thisconnect:
Added dropdown icon so it is clear that it is a dropdown:

Screen Shot 2020-09-17 at 9 02 45 AM

thisconnect:
PTAL

benma (approved):
tACK

PR #996, opened 2020-09-15
frontend/web/preact: remove unused typings-for-css-modules-loader

benma:
There is a complicated config in preact.config.js, which seems to try
to generate .d.ts files from .css files so typescript can type-check
references to style classes.

This seems to not have been in effect since `typed-css-modules` was
added and run in the npm postinstall script as `npx tcm src
--camelCase dashes`, which also generates all .d.ts files.

postinstall is not the right place for it though, it should run before
every build. It seems to have worked in practice as `make buildweb`
runs `yarn install` as a first step, so this postinstall script is
invoked.


frontends/web/package.json
```diff
@@ -3,12 +3,12 @@
   "name": "BitBoxApp",
   "version": "0.0.0",
   "scripts": {
-    "postinstall": "npx tcm src --camelCase dashes",
-    "build": "preact build --no-prerender --service-worker=false",
-    "dev": "preact watch",
-    "lint": "eslint --max-warnings 0 --cache --ignore-path ../../.gitignore '**/*.{js,jsx}'",
-    "postlint": "tslint -e 'src/**/*.css.d.ts' 'src/**/*.ts?(x)'",
-    "test": "jest",
+    "gen-ts-css": "npx tcm src --camelCase dashes",
```
thisconnect, line 6: no need for using `npx` here, as `tcm` will  be installed with `typed-css-modules` and is "like" globally available from within package.json as it is in `node_modules/.bin/tcm -> ../typed-css-modules/lib/cli.js`

https://github.com/Quramy/typed-css-modules/blob/3b9f9e55993b28f9946b2573053bf30a240d9731/package.json#L16

just `"tcm src --camelCase dashes"` is enough

thisconnect (changes requested):

frontends/web/package.json
benma: nice, fixed (also for `npx browserlist`).

thisconnect: nice

thisconnect (approved):
LGTM

PR #999, opened 2020-09-16
frontend: preact to 3.0.1 and be explicit about the typescript version

benma:
The preact-cli-plugin-typescript does not work anymore and is also not
present in the official preact typescript template:

https://github.com/preactjs-templates/typescript/blob/master/template/

Typescript was a transitive dep, but was stuck on an old version. By
adding it as an explicit dep, we can define the version. We need a
newest version as a dependency we want to add requires
it (`lightweight-charts`).


thisconnect:
Just fyi, preact-cli still has the 3.x version and will use that to build
```
 % npm list typescript
BitBoxApp@0.0.0 bitbox-wallet-app/frontends/web
├─┬ preact-cli@3.0.1
│ └── typescript@3.9.7 
└── typescript@4.0.2 
```

also sw.js (serviceworker JavaScript) seems to be back
Screen Shot 2020-09-16 at 12 05 33 PM

the flags have changed in preact-cli@3.x could you build with the following arguments:
`preact build --no-sw --no-esm --no-inline-css --no-prerender --verbose`
(unfortunatelly the changelog just started so there is no useful info https://github.com/preactjs/preact-cli/blob/master/packages/cli/CHANGELOG.md)

see all arguments `npx preact-cli --help` and `npx preact-cli build --help`

```
% npx preact-cli build --help

  Description
    Create a production build

  Usage
    $ preact build [src] [options]

  Options
    --src              Specify source directory  (default src)
    --dest             Specify output directory  (default build)
    --cwd              A directory to use instead of $PWD  (default .)
    --sw               Generate and attach a Service Worker  (default true)
    --json             Generate build stats for bundle analysis
    --template         Path to custom HTML template
    --preload          Adds preload tags to the document its assets  (default false)
    --refresh          Enables experimental preact-refresh functionality  (default false)
    --analyze          Launch interactive Analyzer to inspect production bundle(s)
    --prerenderUrls    Path to pre-rendered routes config  (default prerender-urls.json)
    -c, --config       Path to custom CLI config  (default preact.config.js)
    --babelConfig      Specify the babel config file  (default .babelrc)
    --esm              Builds ES-2015 bundles for your code.  (default true)
    --brotli           Builds brotli compressed bundles of javascript.  (default false)
    --inline-css       Adds critical css to the prerendered markup.  (default true)
    -v, --verbose      Verbose output
    -h, --help         Displays this message
```

The help doesn't really help with disabling see https://github.com/preactjs/preact-cli/issues/798

see
https://github.com/digitalbitbox/bitbox-wallet-app/compare/master...thisconnect:frontend-preactX#diff-168d3f2097b89b7f7f551a423acc4554R7

benma:
> ```
>  % npm list typescript
> BitBoxApp@0.0.0 bitbox-wallet-app/frontends/web
> ├─┬ preact-cli@3.0.1
> │ └── typescript@3.9.7 
> └── typescript@4.0.2 
> ```

Very good catch! One is a devdep, the other a normal dep, but this reveals that I made a mistake. I dropped the typescript dep, everything still works fine.

> 
> also sw.js (serviceworker JavaScript) seems to be back
> Screen Shot 2020-09-16 at 12 05 33 PM
> 
> the flags have changed in [preact-cli@3.x](mailto:preact-cli@3.x) could you build with the following arguments:
> `preact build --no-sw --no-esm --no-inline-css --no-prerender --verbose`
>

thisconnect (approved):
ok lightweight-charts also use typescript@3.9.x 

LGTM

PR #1004, opened 2020-09-17
backend/accounts: add type for OrderedTransactions

benma:
Refactor btc and eth to not repeat the same work of sorting.

In a later commit, the balance at each tx will be computed there as
well, so it can be used in the portfolio overview.


backend/coins/btc/account_test.go
```diff
@@ -90,7 +90,7 @@ func TestAccount(t *testing.T) {
 
 	transactions, err := account.Transactions()
 	require.NoError(t, err)
-	require.Equal(t, []*accounts.TransactionData{}, transactions)
+	require.Equal(t, accounts.OrderedTransactions{}, transactions)
```
x1ddos, line 93: Would be nice to have a test with 2 entries where it verifies the sorting was done, in correct order.

x1ddos (approved):
LGTM

PR #1005, opened 2020-09-22
Balance timeseries

benma:


backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+func (txs OrderedTransactions) findLastIndexBeforeTime(
+	start int, target time.Time) (*int, error) {
+	currentIndex := start
+	for {
+		tx := txs[currentIndex]
+
+		if tx.Height == 0 {
+			// We reached the unconfirmed txs. We ignore those. The current index is the last tx
+			// before the target time.
+			return &currentIndex, nil
```
x1ddos, line 202: If this is the very first loop iteration, shouldn't it return `nil` instead of &currentIndex?

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+	return json.Marshal(map[string]interface{}{
+		"time":  c.Time.Unix(),
+		"value": int64,
+	})
+}
+
+// findLastIndexBeforeTime returns the index of the last transaction with a time before `target`,
+// starting at `start`.
+// Unconfirmed txs are ignored.
+// If there is no such transaction before `target`, nil is returned.
```
x1ddos, line 190: Why not return -1 like it happens in many other similar "findXxx" functions in Go's stdlib or many other languages?

backend/accounts/transaction.go
x1ddos: later below you do `if currentIndex == start { return nil, nil }`. this is a similar case.

x1ddos: in other words, this is the case where all transactions are unconfirmed.

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+			}
+			// We reached the first tx beyond target. Return the previous index, which is the last
+			// tx before target.
+			currentIndex++
+			return &currentIndex, nil
+		}
+		if currentIndex == 0 {
+			return &currentIndex, nil
+		}
+		currentIndex--
```
x1ddos, line 221: Could use binary search to reduce this loop from O(N) to O(log N) and the loop in `Timeseries` from `O(M*N)` to `O(M*log N)` by extension for those who have many transactions.
https://golang.org/pkg/sort/#Search

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+				// No tx before target, return nil.
+				return nil, nil
+			}
+			// We reached the first tx beyond target. Return the previous index, which is the last
+			// tx before target.
+			currentIndex++
+			return &currentIndex, nil
+		}
+		if currentIndex == 0 {
+			return &currentIndex, nil
```
x1ddos, line 219: what if it's still the case that `txTime < target`? then according to the description, it should `return nil, nil`.

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+	}
+	currentTime := start
+	if currentTime.IsZero() {
+		return nil, nil
+	}
+
+	result := []TimeseriesEntry{}
+	currentIndex := len(txs) - 1
+	for {
+		nextIndex, err := txs.findLastIndexBeforeTime(currentIndex, currentTime)
```
x1ddos, line 240: I think it might be possible to remove `findLastIndexBeforeTime` and compute everything in one loop here, in some potential future improvements.

backend/handlers/handlers.go
```diff
@@ -786,10 +786,11 @@ func (handlers *Handlers) formatAmountAsJSON(amount coin.Amount, coinInstance co
...
 func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error) {
 	type accountJSON struct {
-		CoinCode    coinpkg.Code           `json:"coinCode"`
-		AccountCode string                 `json:"accountCode"`
-		Name        string                 `json:"name"`
-		Balance     map[string]interface{} `json:"balance"`
+		CoinCode    coinpkg.Code               `json:"coinCode"`
+		AccountCode string                     `json:"accountCode"`
+		Name        string                     `json:"name"`
+		Balance     map[string]interface{}     `json:"balance"`
```
x1ddos, line 792: Would've been much nicer type-wise if `Balance` was a struct.

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+// starting at `start`.
+// Unconfirmed txs are ignored.
+// If there is no such transaction before `target`, nil is returned.
+// Note: the transactions are ordered from newest to oldest, so the returned index will be smaller
+// or equal to `start` (advancing backwards).
+func (txs OrderedTransactions) findLastIndexBeforeTime(
+	start int, target time.Time) (*int, error) {
+	currentIndex := start
+	for {
+		tx := txs[currentIndex]
```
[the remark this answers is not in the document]
benma: why is this important? it is a private function only called with valid args. out of bounds index is a programming error and should panic?

backend/accounts/errors/errors.go
```diff
@@ -14,6 +14,10 @@
 
 package errors
 
+import (
+	errs "errors"
```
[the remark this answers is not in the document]
x1ddos: s/page/package :see_no_evil:

backend/accounts/transaction.go
benma: It is already O(N) (n=len(txs)). The index strictly advances through the slice.

backend/accounts/transaction.go
benma: I tried but it was definitely harder for me to write than this ;)

backend/accounts/transaction.go
```diff
@@ -84,6 +88,11 @@ type TransactionData struct {
 	Type TxType
 	// Amount is always >0 and is the amount received or sent (not including the fee).
 	Amount coin.Amount
+	// Balance is balance of the account at the time of this transaction. It is the sum of all
+	// transactions up to this point.
+	// This value is only valid pas part of `OrderedTransactions`.
```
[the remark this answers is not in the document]
benma: done

backend/accounts/errors/errors.go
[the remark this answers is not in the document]
benma: done

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+
+// TimeseriesEntry contains the balance of the account at the given time.
+type TimeseriesEntry struct {
+	Time  time.Time
+	Value coin.Amount
+}
+
+// MarshalJSON serializes the entry as JSON.
+func (c *TimeseriesEntry) MarshalJSON() ([]byte, error) {
+	int64, err := c.Value.Int64()
```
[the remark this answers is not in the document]
benma: done

backend/accounts/transaction.go
benma: done

backend/accounts/transaction.go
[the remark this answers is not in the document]
x1ddos: it's just a very common pracice. it's a separate function with an error return type. it takes some arguments and should check if they're actually valid.

> private function only called with valid args

today, in this pull request, yes. what about tomorrow or what if the function which calls this function actually has a bug? it'll crash the app. not a good UX.

backend/accounts/transaction.go
benma: if txtime < target, this is the last tx before target.

Say target is one year from now, and your latest tx (index = 0) is today. Then `0` is the last tx before target.

backend/accounts/transaction.go
x1ddos: so, you're saying O(N) is better than O(log N)?

backend/accounts/transaction.go
benma: Great catch. Fixed.

Will think about how to add more comprehensive tests later.

backend/accounts/transaction.go
```diff
@@ -133,5 +142,122 @@ type OrderedTransactions []*TransactionData
...
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(map[string]interface{}{
```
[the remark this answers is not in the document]
benma: Done.

benma:
Addressed all. PTAL

backend/accounts/transaction.go
x1ddos: > if txtime < target, this is the last tx before target.

but there's nothing else "before" the target. it's the last element.

consider this: t1, t2, t3, and target is `t1 - 1`. if `currentIndex == 0`, so we're at t1, the currentIndex still doesn't point to an item with timestamp < target. in other words: no such transaction.

or am I missing something?

x1ddos:
> Addressed all. PTAL

what about these:

- https://github.com/digitalbitbox/bitbox-wallet-app/pull/1005#discussion_r492680660
- https://github.com/digitalbitbox/bitbox-wallet-app/pull/1005#discussion_r492671974
- https://github.com/digitalbitbox/bitbox-wallet-app/pull/1005#discussion_r492670456

backend/accounts/transaction.go
benma: Ah, haha brain was off it seems ;) 

You are right, but I would skip it for simpler code. Binary search here will only help if you have lots of transactions within one interval, but then also probably only measurable if it is quite a large number.

x1ddos: But try it! I think it might actually simplify code. You need to provide only the lookup-at-index function:

```go
func Search(n int, f func(int) bool) int
```

x1ddos: It was designed for cases exactly like this one:

> Search uses binary search to find and return the smallest index i in [0, n) at which f(i) is true

backend/accounts/transaction.go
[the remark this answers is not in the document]
benma: Done.

backend/accounts/transaction.go
benma: Fixed edge case where target==txTime, changing from `target.Before()` to `!target.After()`.

Still thinking about the rest.

backend/accounts/transaction.go
benma: I added it in a new commit, please check. Nice idea!

backend/accounts/transaction.go
```diff
@@ -133,5 +142,94 @@ type OrderedTransactions []*TransactionData
...
+
+	result := []TimeseriesEntry{}
+	for {
+		// Find the latest tx before `currentTime`.
+		nextIndex := sort.Search(len(txs), func(idx int) bool {
+			tx := txs[idx]
+			if tx.Height == 0 {
+				return false
+			}
+			return txs[idx].Timestamp.Before(currentTime)
```
x1ddos, line 215: or use the `tx` var since it's there anyway? `return tx.Timestamp.Before(currentTime)`

x1ddos: oh, I think you'll actually want `return tx.Timestamp.Before(currenTime) || tx.Timestamp.Equal(currentTime)`. if there's a transaction which falls exactly on the `currentTime`, you'll want to pick its balance, not the one before, right.

But that condition is harder to read. A better equivalent would be: `return !tx.Timestamp.After(currentTime)` or `return !currentTime.Before(tx.Timestamp)`.

backend/accounts/transaction_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestOrderedTransactions(t *testing.T) {
+	tt := func(t time.Time) *time.Time { return &t }
+	fee := coin.NewAmountFromInt64(1)
+	txs := []*TransactionData{
+		{
+			Timestamp: tt(time.Date(2020, 9, 15, 12, 0, 0, 0, time.UTC)),
+			Height:    15,
```
x1ddos, line 31: Would be nice to sneak in an entry with 0 height somewhere in between.

x1ddos (approved):
Yeah, much nicer imho! Thanks for sticking with my comments 'till the end.
LGTM

backend/accounts/transaction.go
x1ddos: actually, maybe `tx.Timestamp.Before(currenTime) || tx.Timestamp.Equal(currentTime)` is easer to read in the end.

backend/accounts/transaction_test.go
benma: done

backend/accounts/transaction.go
benma: done

PR #1007, opened 2020-09-22
android: separate appID and app display name for debug builds

benma:
I don't want to uninstall my main app to test a new build. This will
install a separate app if build in debug mode.


x1ddos (approved):
Nice and simple. LGTM.

PR #1009, opened 2020-09-23
backend/rates: an MVP version of historical exchange rates updater

x1ddos:
This adds a new exported function PriceAt which returns a historical exchange rate at the given time.
Only BTC/USD pair is currently supported. All the others return flat 0.

Future commits will add all other coints and resolve pending TODOs.

### DIFF
--- a/backend/rates/history.go
+++ b/backend/rates/history.go
@@ -1,0 +1,218 @@
+package rates
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"sort"
+	"strconv"
+	"time"
+)
+
+// See the following for docs and details: https://www.coingecko.com/en/api.
+// Rate limit: 6k QPS per IP address.
+const coingeckoAPIV3 = "https://api.coingecko.com/api/v3"
+
+var (
+	// Copied from https://api.coingecko.com/api/v3/coins/list.
+	// The keyes must match entries in coins slice.
+	geckoCoin = map[string]string{
+		"BTC":  "bitcoin",
+		"LTC":  "litecoin",
+		"ETH":  "ethereum",
+		"USDT": "tether",
+		"LINK": "chainlink",
+		"MKR":  "maker",
+		"ZRX":  "0x",
+		"DAI":  "dai",
+		"BAT":  "basic-attention-token",
+		"USDC": "usd-coin",
+	}
+
+	// Copied from https://api.coingecko.com/api/v3/simple/supported_vs_currencies.
+	// The keys must match entries in fiats slice.
+	geckoFiat = map[string]string{
+		"USD": "usd",
+		"EUR": "eur",
+		"CHF": "chf",
+		"GBP": "gbp",
+		"JPY": "jpy",
+		"KRW": "krw",
+		"CNY": "cny",
+		"RUB": "rub",
+		"CAD": "cad",
+		"AUD": "aud",
+	}
+)
+
+// historyUpdateLoop periodically updates historical market exchange rates
+// forward in time starting with the last fetched timestamp.
+// It never returns.
+func (updater *RateUpdater) historyUpdateLoop() {
+	for {
+		// When to update next, after this loop iteration is done.
+		untilNext := 10 * time.Minute // TODO: add jitter
+
+		// TODO: Do this for all supported coins.
+		start := updater.historyLatestTimestamp("BTC")
+		// When zero, there's no point in fetching data here because the backfillHistory
+		// will kick in and fill it up for the past 90 days anyway.
+		if !start.IsZero() {
+			// Start slightly past the last fetched timestamp
+			start = start.Add(10 * time.Minute)
+			// TODO: Handle the case where (now - start) > 90 days to get hourly rates?
+			now := time.Now()
+			// It's ok if start == now or now < start. CoinGecko simply returns no results.
+			// It's also ok if our "now" doesn't match CoinGecko: it always returns
+			// the latest available data without errors.
+			// TODO: Do this for all supported fiats.
+			_, err := updater.updateHistory("BTC", "USD", start, now)
+			if err != nil {
+				updater.log.Errorf("updateHistory(%v, %v): %v", start, now, err)
+				untilNext = time.Minute // TODO: exponential backoff
+			}
+		}
+
+		time.Sleep(untilNext)
+	}
+}
+
+// backfillHistory fetches historical market exchange rates starting with
+// the earliest fetched timestamp backwards until data is available at the API endpoint.
+//
+// It does so in a loop and returns only after all data is backfilled.
+// Callers are expected to run this in a separate goroutine.
+func (updater *RateUpdater) backfillHistory() {
+	for {
+		// When to update next, after this loop iteration is done.
+		untilNext := time.Second // TODO: add jitter
+
+		// TODO: Do this for all supported coins.
+		end := updater.historyEarliestTimestamp("BTC")
+		var start time.Time
+		if end.IsZero() {
+			// First time; don't have historical data yet.
+			end = time.Now()
+			// 90 days is the max interval which CoinGecko responds to with hourly timeseries.
+			// Make sure we get hourly rates by requesting 90 days minus 1 hour, just in case.
+			start = end.Add(-90*24*time.Hour + time.Hour)
+		} else {
+			// End at the day earlier than the last known timestamp.
+			// We want daily rates for timestamps past the first 90 days: use annual intervals.
+			end = end.Add(-24 * time.Hour)
+			start = end.Add(-365 * 24 * time.Hour)
+		}
+
+		// TODO: Do this for all supported fiats.
+		n, err := updater.updateHistory("BTC", "USD", start, end)
+		if err != nil {
+			updater.log.Printf("updateHistory(%s, %s): %v", start, end, err)
+			untilNext = time.Minute // TODO: exponential backoff
+		}
+		// CoinGecko returns an empty list if we're too far back in history.
+		// Use it to detect when to stop.
+		// Unless the start/end range is suspiciously arbitrary close to current time.
+		// It may indicate an API failure and we'll want to retry.
+		if n == 0 {
+			if time.Since(end) > 365*24*time.Hour {
+				return
+			}
+			untilNext = time.Minute // TODO: exponential backoff
+		}
+
+		time.Sleep(untilNext)
+	}
+}
+
+// updateHistory fetches and stores historical data for later use.
+// It returns a number of the newly fetched and stored entries.
+// The data is stored in updater.history.
+func (updater *RateUpdater) updateHistory(coin, fiat string, start, end time.Time) (n int, err error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	rates, err := updater.fetchGeckoMarketRange(ctx, coin, fiat, start, end)
+	if err != nil {
+		return 0, err
+	}
+
+	updater.historyMu.Lock()
+	defer updater.historyMu.Unlock()
+	a := append(updater.history[coin], rates...)
+	sort.Slice(a, func(i, j int) bool {
+		return a[i].timestamp.Before(a[j].timestamp)
+	})
+	updater.history[coin] = a
+	return len(rates), nil
+}
+
+func (updater *RateUpdater) historyLatestTimestamp(coin string) time.Time {
+	updater.historyMu.RLock()
+	defer updater.historyMu.RUnlock()
+	var t time.Time
+	if n := len(updater.history[coin]); n > 0 {
+		t = updater.history[coin][n-1].timestamp
+	}
+	return t
+}
+
+func (updater *RateUpdater) historyEarliestTimestamp(coin string) time.Time {
+	updater.historyMu.RLock()
+	defer updater.historyMu.RUnlock()
+	var t time.Time
+	if len(updater.history[coin]) > 0 {
+		t = updater.history[coin][0].timestamp
+	}
+	return t
+}
+
+// fetchGeckoMarketRange slurps historical exchange rates using CoinGecko's
+// "market_chart/range" API.
+func (updater *RateUpdater) fetchGeckoMarketRange(ctx context.Context, coin, fiat string, start, end time.Time) ([]exchangeRate, error) {
+	gcoin := geckoCoin[coin]
+	if gcoin == "" {
+		return nil, fmt.Errorf("fetchGeckoMarketRange: unsupported coin %s", coin)
+	}
+	gfiat := geckoFiat[fiat]
+	if gfiat == "" {
+		return nil, fmt.Errorf("fetchGeckoMarketRange: unsupported fiat %s", fiat)
+	}
+	param := url.Values{
+		"from":        {strconv.FormatInt(start.Unix(), 10)},
+		"to":          {strconv.FormatInt(end.Unix(), 10)},
+		"vs_currency": {gfiat},
+	}
+	endpoint := fmt.Sprintf("%s/coins/%s/market_chart/range?%s", updater.coingeckoURL, gcoin, param.Encode())
+	req, err := http.NewRequest("GET", endpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := updater.httpClient.Do(req.WithContext(ctx))
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close() //nolint:errcheck
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("fetchGeckoMarketRange: bad response code %d", res.StatusCode)
+	}
+	var jsonBody struct {
+		Prices [][2]float64 // [timestamp in milliseconds, value], sorted by timestamp asc
+	}
+	// 1Mb is more than enough for a single response.
+	// For comparison, a range of 15 days is about 14Kb.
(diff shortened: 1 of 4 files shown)


x1ddos:
There are no tests yet and I'm not even sure this works as expected. Also, a couple existing tests hit the actual coingecko API because the backend always starts updater in it's NewBackend func. :(

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+const coingeckoAPIV3 = "https://api.coingecko.com/api/v3"
+
+// forwardHistoryLoop periodically updates historical market exchange rates
+// forward in time starting with the last fetched timestamp.
+func (updater *RateUpdater) forwardHistoryLoop() {
+	for {
+		// When to update next, after this loop iteration is done.
+		untilNext := 10 * time.Minute // TODO: add jitter
```
benma, line 25: jitter: why?

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+		if err != nil {
+			updater.log.Printf("updateHist(%s, %s): %v", start, end, err)
+			untilNext = time.Minute // TODO: exponential backoff
+		}
+
+		time.Sleep(untilNext)
+	}
+}
+
+func (updater *RateUpdater) updateHist(coin, fiat string, start, end time.Time) error {
```
benma, line 75: argument for longer names: is this history? histogram? `hist()` in math software is usually histogram, but here it probably means history?

backend/rates/rates.go
```diff
@@ -37,45 +39,103 @@ var fiats = []string{"USD", "EUR", "CHF", "GBP", "JPY", "KRW", "CNY", "RUB", "CA
...
-func NewRateUpdater(socksProxy socksproxy.SocksProxy) *RateUpdater {
-	ratesUpdater := &RateUpdater{
-		last:       map[string]map[string]float64{},
-		log:        logging.Get().WithGroup("rates"),
-		socksProxy: socksProxy,
+func NewRateUpdater(client *http.Client) *RateUpdater {
+	return &RateUpdater{
+		last:         make(map[string]map[string]float64),
+		hist:         make(map[string][]exchangeRate),
+		log:          log.New(os.Stderr, "[rates] ", log.LstdFlags|log.Lmsgprefix),
```
benma, line 73: this now does not go into the log.txt file. i suggest to stick with logrus, and replace all of logrus in a dedicated PR. mixing the two confusing to read and maintain.

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+
+		start := updater.histLatestTimestamp("BTC")
+		if !start.IsZero() {
+			// Start slightly past the last fetched timestamp
+			start = start.Add(10 * time.Minute)
+			now := time.Now()
+			// It's ok if start == now or now < start. CoinGecko simply returns no results.
+			// It's also ok if our "now" doesn't match CoinGecko: it always returns
+			// the latest available data without errors.
+			err := updater.updateHist("BTC", "USD", start, now)
```
benma, line 35: "BTC" is used as the coin id in the coingecko api, but BTC doesn't exist, it is "bitcoin":

https://www.coingecko.com/api/documentations/v3#/coins/get_coins_list

At least if I enter `BTC` here it does not work:

https://www.coingecko.com/api/documentations/v3#/coins/get_coins__id__market_chart_range

benma (commented):
Looks nice! I didn't test it yet, but leaving some comments already.

Is the plan to keep this in RAM or put it into a DB?

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+	updater.histMu.Lock()
+	defer updater.histMu.Unlock()
+	// TODO: updater.hist[coin] and rates may have duplicate timestamps.
+	a := append(updater.hist[coin], rates...)
+	sort.Slice(updater.hist[coin], func(i, j int) bool {
+		return a[i].timestamp.Before(a[j].timestamp)
+	})
+	updater.hist[coin] = a
+
+	updater.Notify(observable.Event{
```
benma, line 92: i think this isn't how the events work, in .Reload the Data should be sent with it, and the subject should point to a valid handlers endpoint.

I would remove this for now, we can add it in a separate PR that deals with updating the chart.

backend/rates/coingecko.go
x1ddos: It follows the RateUpater field names:

`last`: `updateLast`
`hist`: `updateHist`

but sure I can rename. yeah, maybe it isn't clear enough.

backend/rates/coingecko.go
x1ddos: yeah, coingecko uses IDs and `bitcoin` is their identifier for BTC. you'll see it in `fetchMarkerRange`. this will need a map between the standard symbols to coingecko API.

backend/rates/rates.go
x1ddos: hm... haven't thought about the log.txt. good point. will revert.

x1ddos:
> Is the plan to keep this in RAM or put it into a DB?

Yeah for sure, I'd like to put it into a DB, most importantly for the backfilled data. Otherwise, it starts slurping years of history on every app restart. In separate commit though.

Anyway, so this is one approach. Specifically, it starts from an account transactions, computes balance, slices it into equally distant data points and then multiplies each point by the price at that timestamp grabbed from the historical data.

Another approach could be the exact opposite: start with the historical market data, loop over the timeseries and multiply the price by the balance an account had at that time. Not sure if this is actually better. Just thinking out loud.

backend/rates/coingecko.go
benma: yeah including the `hist` field of course ;) same thing.

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := updater.httpClient.Do(req.WithContext(ctx))
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close() //nolint:errcheck
+	var jsonBody struct {
```
benma, line 136: need to check `res.Status` here. as it is now, using `BTC` instead of `bitcoin`, this always returns 404, but it passes silently (even the json decoding does not fail).

backend/rates/coingecko.go
x1ddos: actually, forget what I said. I made it too generic and forgot about the mapping :see_no_evil:. it is currently fetching no useful data indeed.

backend/rates/coingecko.go
benma: fyi there was an interesting error like this in the blockchain.info wallet: they fetched additional randomness from random.org, but some error status code, which was used as the actual randomness due to missing status code checks :sweat_smile: in combination with an additional bug, it ended up being the only "randomness" in the seed creation.

x1ddos: yeah, absolutely. I forgot and made a rusty mistake here :)
nice catch!

backend/rates/coingecko.go
x1ddos: because we'll be fetching historical data for many coins, right: BTC, LTC, ETH and so on.
so, we don't want to hammer the API with 5-10 QPS every 10min exactly. the idea is to be more gentle and spread the load on the API provider over a couple minutes.

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+	// 1Mb is more than enough for a single response.
+	// For comparison, a range of 15 days is about 14Kb.
+	if err := json.NewDecoder(io.LimitReader(res.Body, 1<<20)).Decode(&jsonBody); err != nil {
+		return nil, err
+	}
+
+	rates := make([]exchangeRate, len(jsonBody.Prices))
+	for i, v := range jsonBody.Prices {
+		rates[i] = exchangeRate{
+			timestamp: time.Unix(0, int64(v[0])), // local timezone
```
benma, line 148: this does not work at all, i get only 1970- timestamps. the arguments seem reversed, but when i reverse them i get timestamps in the year 52452 :joy:

backend/rates/coingecko.go
x1ddos: and it'll work in our favour regarding the API rate limits

backend/rates/coingecko.go
x1ddos: that should work. the args are correct: https://golang.org/pkg/time/#Unix

> ... It is valid to pass nsec outside the range [0, 999999999].

you can see the source code: https://golang.org/src/time/time.go?s=39974:40011#L1282

> need to reverse args and divide by 1000.

right, or don't reverse the args and multiple by 1000. I forgot the 2nd arg is nanoseconds.

x1ddos: but yeah, either way there's a mistake. doesn't work as is :sweat_smile:

backend/rates/coingecko.go
```diff
@@ -0,0 +1,153 @@
...
+	var t time.Time
+	if len(updater.hist[coin]) > 0 {
+		t = updater.hist[coin][0].timestamp
+	}
+	return t
+}
+
+func (updater *RateUpdater) fetchMarketRange(ctx context.Context, coin, fiat string, start, end time.Time) ([]exchangeRate, error) {
+	param := url.Values{
+		"from":        {strconv.FormatInt(start.Unix(), 10)},
```
benma, line 121: should double check if it is the right resolution, same as in the response, we have to divide by `1000`, this could also be off by a factor.

though the last 3 months seems to work, so probably ok.

backend/rates/coingecko.go
x1ddos: done

backend/rates/coingecko.go
x1ddos: done

backend/rates/coingecko.go
x1ddos: removed

backend/rates/coingecko.go
x1ddos: yes, it expects unix, i.e. seconds.

backend/rates/coingecko.go
x1ddos: done

backend/rates/coingecko.go
x1ddos: done

backend/rates/rates.go
x1ddos: reverted

x1ddos (commented):
PTAL. Note that I renamed coingecko.go file to history.go. Never meant to keep coingecko.go name but it slipped through.

backend/rates/history_test.go
```diff
@@ -0,0 +1,79 @@
...
+			  [
+			    1598918700000,
+			    10000.0
+			  ],
+			  [
+			    1598922501000,
+			    10001.0
+			  ]
+			]
+    	}`)
```
x1ddos, line 42: spaces vs tabs. will align before merging.

backend/rates/history_test.go
```diff
@@ -0,0 +1,79 @@
...
+		"BTC": {
+			{timestamp: time.Unix(1599091262, 0), value: 10.0}, // 2020-09-03 00:01:02
+		},
+	}
+
+	err := updater.updateHistory("BTC", "USD", time.Unix(wantStartUnix, 0), time.Unix(wantEndUnix, 0))
+	require.NoError(t, err, "updater.updateHistory")
+	wantHistory := map[string][]exchangeRate{
+		"BTC": {
+			{timestamp: time.Unix(1599091262, 0), value: 10.0}, // preexisting point
```
x1ddos, line 57: Hold on. This is wrong. It should be the last point, not the first.

x1ddos: Yep, that was an actual bug. Fixed.

backend/rates/history_test.go
x1ddos: done

x1ddos (commented):
Sorry, PTAL. There's still one more test I'd like to add, for `PriceUSD` method.

x1ddos:
Ok, for the last time PTAL. Added a test for `PriceUSD` and found another bug. Should be all good now.

x1ddos:
Made linters happy but nothing else changed.

backend/rates/history.go
```diff
@@ -0,0 +1,199 @@
...
+// backfillHistory fetches historical market exchange rates starting with
+// the earliest fetched timestamp backwards until the end of times.
+// TODO: what is "the end of times"?
+//
+// It does so in a loop and returns only after all data is backfilled.
+// Callers are expected to run this in a separate goroutine.
+func (updater *RateUpdater) backfillHistory() {
+	for {
+		// When to update next, after this loop iteration is done.
+		untilNext := 10 * time.Second // TODO: add jitter
```
benma, line 88: why wait so long? the rate limit is much shorter right?

backend/rates/history.go
```diff
@@ -0,0 +1,199 @@
...
+
+		// TODO: Do this for all supported coins.
+		end := updater.historyEarliestTimestamp("BTC")
+		if end.IsZero() {
+			end = time.Now()
+		}
+		// 90 days is the max interval which CoinGecko responds with hourly timeseries.
+		start := end.Add(-90 * 24 * time.Hour)
```
benma, line 96: should we do slightly lass than that for the past three months? avoid getting daily rates by accident.

for older rates, what about getting daily rates only, so we get more data per api call? we planned on showing daily rates for older history anyway, and hourly only for the past three months, right?

backend/rates/history.go
```diff
@@ -0,0 +1,199 @@
...
+		if time.Since(start) > 720*24*time.Hour {
+			// ~2 years of historical data is enough for now.
+			// TODO: Where do we stop?
```
benma, line 99: what about all of it, until coingecko does not have any more data?

Or stop at [2009-01-03](https://blockstream.info/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f) :D

backend/rates/rates.go
```diff
@@ -20,60 +20,124 @@ import (
...
-		updater.log.Printf("Error getting http client %v\n", err)
-		updater.last = nil
-		return
+// PriceUSD returns a historical exchange rate for the given coin.
+// The returned value may be imprecise if at arg matches no timestamp exactly.
+// In this case, linear interpolation is used as an approximation.
+// If no data is available with the given args, PriceUSD returns 0.
+//nolint:staticcheck
+func (updater *RateUpdater) PriceUSD(coin string, at time.Time) float64 {
+	coin = "BTC" // TODO: support all coins; remove nolint:staticcheck
```
benma, line 95: remove this?

backend/rates/rates.go
```diff
@@ -20,60 +20,124 @@ import (
...
-	if err != nil {
-		updater.log.Printf("Error getting http client %v\n", err)
-		updater.last = nil
-		return
+// PriceUSD returns a historical exchange rate for the given coin.
+// The returned value may be imprecise if at arg matches no timestamp exactly.
+// In this case, linear interpolation is used as an approximation.
+// If no data is available with the given args, PriceUSD returns 0.
+//nolint:staticcheck
+func (updater *RateUpdater) PriceUSD(coin string, at time.Time) float64 {
```
benma, line 94: what about `PriceAt(coin, fiat string, at time.Time)`? Can panic for now if fiat is not `USD` until implemented.

But once we can switch between currencies, it is easier to pass the fiat as an argument instead of having dedicated functions per fiat currency.

backend/rates/rates.go
```diff
@@ -20,60 +20,124 @@ import (
...
+	data := updater.history[coin]
+	// Find an index of the first entry older or equal the at timestamp.
+	idx := sort.Search(len(data), func(i int) bool {
+		return !data[i].timestamp.Before(at)
+	})
+	if idx == len(data) {
+		return 0 // no data
+	}
+	if idx == 0 || data[idx].timestamp.Equal(at) {
+		return data[idx].value // can't or don't need to interpolate
```
benma, line 107: with `idx==0`, the value in the past for which we do not have rates is very likely totally wrong. better return `0` in this case (no data).

E.g. the price in the year 2010 should not be USD 10'000 just because we only have the last three months available ;)

benma: For the record, this is the only "semantic" issue in the PR, the rest are nits or things than can be improved later.

backend/rates/history.go
x1ddos: Good point. :)
I've checked and the earlierst it returns is dated April 28, 2013 12:00:00 AM. If I give it a range like 2009-01-03 - 2010-01-03, it comes up with an empty list. So, let's fetch until we get an empty list.

backend/rates/history.go
x1ddos: yeah, good idea. done and done.

backend/rates/rates.go
x1ddos: Yeah, I had it as `PriceAt` with coin/fiat before and then figured PriceUSD would've been enough for now but you're right. Might as well do it now.

But I'm not a fan of panics. Added a condition to always return 0 for non-USD fiat. This way all the graphs and backend handler can call PriceAt without issues. It'll simply show a flat graph at 0.

backend/rates/rates.go
x1ddos: done

backend/rates/rates.go
x1ddos: true. better return 0. however, `idx == 0` **and** equal timestamps is still a good case to return the non-zero value.

x1ddos (commented):
Sorry, got sidetracked. PTAL.

benma (approved):
:clap:

PR #1011, opened 2020-09-28
Frontend happy eslint

thisconnect:
tslint is deprecated and going to be replaced by eslint, but unfortunately the TS errors wont be as detailed as when building with TypeScript, see https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md#why-dont-i-see-typescript-errors-in-my-eslint-output


frontends/web/.eslintrc.js
```diff
@@ -0,0 +1,159 @@
...
+  "env": {
+    "browser": true
+  },
+  "parser": "@typescript-eslint/parser",
+  "plugins": ["@typescript-eslint"],
+  "extends": [
+    // "eslint:recommended",
+    // "plugin:@typescript-eslint/eslint-recommended",
+    // "plugin:@typescript-eslint/recommended",
+    // "plugin:@typescript-eslint/recommended-requiring-type-checking",
```
thisconnect, line 11: forgot to enable type checkings, coming soon™

frontends/web/src/app.tsx
```diff
@@ -81,30 +81,30 @@ class App extends Component<Props, State> {
...
-            case 'backend':
-                switch (data) {
-                case 'newTxs':
-                    apiPost('notify-user', {
-                        text: this.props.t('notification.newTxs', {
-                            count: meta.count,
-                            accountName: meta.accountName,
-                        }),
-                    });
+                case 'backend':
```
benma, line 84: is there a way to make the indent rule not indent `case`? imho it worsens readability. not a big deal if not.

frontends/web/src/routes/settings/settings.tsx
```diff
@@ -50,12 +50,12 @@ interface State {
 class Settings extends Component<Props, State> {
     private accountsList = [
         { name: 'bitcoinActive',
-          label: 'Bitcoin',
-          badges: [],
+            label: 'Bitcoin',
```
benma, line 53: this looks wrong?

frontends/web/src/components/backups/backup.tsx
```diff
@@ -36,46 +36,51 @@ export interface Backup {
 
 type Props = BackupsListItemProps & TranslateProps;
 
-class BackupsListItem extends Component<Props> {
```
benma, line 39: eslint complained about this? wow!

what about splitting this commit into a separate PRs? this is massive PR, it's easier to feed it in smaller bits :D

frontends/web/src/components/rates/rates.tsx
```diff
@@ -33,7 +33,9 @@ export type Coin = MainnetCoin | TestnetCoin;
 export type Fiat = 'USD' | 'EUR' | 'CHF' | 'GBP' | 'JPY' | 'KRW' | 'CNY' | 'RUB' | 'CAD' | 'AUD';
 
 export type Rates = {
+    /* eslint no-unused-vars: "off" */
     [coin in MainnetCoin]: {
```
benma, line 37: I tried googling `typescript in` :sweat_smile: what does `in` do in TS?

benma (commented):
Holy moly, nice linting! Interesting that the old linters were so lenient! I am nowhere near going through all of it.

Is it possible to split it into smaller PRs? Maybe one PR per lint rule or sth like that, starting from the old config? It's a bit overwhelming :sweat_smile:

frontends/web/src/app.tsx
thisconnect: Half of the switch case were already indented, I do not agree that this worsens readability, indenting usually helps visualizing where the block starts/ends. But if this is your personal preference then I don't mind, but it is very common in js world

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch
- https://2ality.com/2020/02/typescript-exhaustiveness-checks-via-exceptions.html
- (expect Crockford, he doesn't indent either :P https://www.crockford.com/code.html#switch%20statement)

frontends/web/src/components/backups/backup.tsx
thisconnect: It is already split into single commits, see "eslint: refactor components as a pure function" https://github.com/digitalbitbox/bitbox-wallet-app/pull/1011/commits/1fa9d01c66d2f10d2af0dde4bf480a7553c93c1e

frontends/web/src/routes/settings/settings.tsx
thisconnect: indeed, yucky

frontends/web/src/components/backups/backup.tsx
thisconnect: expect the last commit is a bunch of different eslint fixes

frontends/web/src/components/rates/rates.tsx
thisconnect: Eslint doesn't understand this and errors, but once we enable the commented out plugin:typescript-eslint/recommended 🤞 we can remove the eslint ignore comment.

Mapped types 
- https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types
- https://mariusschulz.com/blog/mapped-types-in-typescript#more-examples-for-mapped-types

thisconnect:
> Holy moly, nice linting! Interesting that the old linters were so lenient! I am nowhere near going through all of it.

I agree, and the worst is the typescript plugins aren't even enabled yet.

The last commit is a cleanup that addresses just the remaining outstanding eslint warnings and errors, but all other commits just fix 1 rule and are acutally done by `eslint --fix`.

I'll update the comment.

frontends/web/src/components/rates/rates.tsx
thisconnect: We don't need to disable the eslint rule here once `"plugin:@typescript-eslint/recommended"` is enabled

frontends/web/src/app.tsx
benma: Interesting that the first two examples have 2 space indents, not 4 space, so the code is not indented 8 spaces in, but just 4. I wouldn't mind going to 2 spaces either :D 

Anyway, I also don't have a very strong opinion. Since you spend the most time in the frontend, it should fit your style :sunglasses:

frontends/web/src/components/rates/rates.tsx
benma: Thx for the reference. Seems like uppercase for the type var might be more appropriate, e.g. `Coin in MainnetCoin`.

thisconnect:
closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1591

PR #1021, opened 2020-10-03
Frontend: enable no else return rule and manual fix

thisconnect:


benma (approved):
one of my favorites :D

PR #1023, opened 2020-10-03
Frontend: eslint fix lonely if

thisconnect:


thisconnect:
I don't like this one, there is no point doing that, there is no else if :)

benma:
> I don't like this one, there is no point doing that, there is no elseif :)

then disable it? i don't mind either way.

benma (approved):

PR #1026, opened 2020-10-03
Frontend: eslint error on missing jsx key, manual fix

thisconnect:


frontends/web/src/components/devices/bitbox02/mnemonicpassphrase.tsx
```diff
@@ -73,8 +73,8 @@ class MnemonicPassphraseButton  extends Component<Props, State> {
                                             !mnemonicPassphraseEnabled && message && (
                                                 <p>
                                                     {
-                                                        message.split('\n').map(line => (
-                                                            <span>
+                                                        message.split('\n').map((line, i) => (
+                                                            <span key={`line-${i}`}>
```
benma, line 77: i think the key can be anything, also just `key={i}`, or not?

frontends/web/src/routes/account/summary/balancestable.tsx
```diff
@@ -52,7 +52,7 @@ class BalancesTable extends Component<Props> {
                         </tr>
                     </thead>
                     <tbody>
-                        {accounts.map(account => <BalanceRow name={account.name} balance={account.balance.available} code={account.coinCode}/>)}
+                        {accounts.map(account => <BalanceRow key={account.coinCode} name={account.name} balance={account.balance.available} code={account.coinCode}/>)}
```
benma, line 55: `coinCode` is not unique, it will be `BTC` for e.g. segwit and native segwit. please check if there i `code` or `accountCode`, which would be sth like `btc-p2wpkh` or similar.

benma (changes requested):

frontends/web/src/components/devices/bitbox02/mnemonicpassphrase.tsx
thisconnect: Yes we can if there is nothing else and we are sure the order never changes, https://reactjs.org/docs/lists-and-keys.html#keys

> We don’t recommend using indexes for keys if the order of items may change.

frontends/web/src/routes/account/summary/balancestable.tsx
thisconnect: Resolved, thanks for #1031

frontends/web/src/components/devices/bitbox02/mnemonicpassphrase.tsx
thisconnect: Using `key={i}` when the message is always the same, squash pushed here https://github.com/digitalbitbox/bitbox-wallet-app/compare/adefd672dff147b228e4bd0a6676e5a1a8186f14..0ffdf60a2d7dd90bf569daa432db50f6e03fb797

keeping a combination of line + i for Spinner

benma: tbh i didn't see the `line` argument. if you wanted to have it in the key, then `line-${i}` didn't do that, you want `${line}-${i}`, or maybe `[i, line]`?

thisconnect: Oh, yes ofc I wanted `${line}-${i}` :P

thisconnect: Thanks for catching this.

frontends/web/src/components/devices/bitbox02/mnemonicpassphrase.tsx
```diff
@@ -72,13 +72,11 @@ class MnemonicPassphraseButton  extends Component<Props, State> {
...
                                                 <p>
-                                                    {
-                                                        message.split('\n').map(line => (
-                                                            <span>
-                                                                <SimpleMarkup tagName="span" markup={line} /><br/>
-                                                            </span>
-                                                        ))
-                                                    }
+                                                    { message.split('\n').map((line, i) => (
+                                                        <span key={i}>
```
benma, line 76: same here and elsewhere: the array index is not a valid key, as it is not unique per element. totally different entries will have the same keys, which can lead to glitches.

the better key here is just `line` i think.

benma (changes requested):

frontends/web/src/components/devices/bitbox02/mnemonicpassphrase.tsx
thisconnect: done, using a combination of line + index https://github.com/digitalbitbox/bitbox-wallet-app/compare/e3b8f4422cd2450d8ff969c8440f74501ad9cc2f..932e44070d7fa5168d7a336655892978aca6a6a8

benma (approved):

PR #1027, opened 2020-10-03
Frontend: enable eslint no-undef and manual fix

thisconnect:


frontends/web/src/components/anchor/anchor.tsx
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { h, RenderableProps } from 'preact';
+import { h, JSX, RenderableProps } from 'preact';
```
thisconnect, line 17: TypeScript complained about returning JSX.Element, here https://github.com/digitalbitbox/bitbox-wallet-app/pull/1027/files#diff-c46c03dfb317b41884c6e334884a9bceR26

Ideally we wouldn't have to import JSX

frontends/web/src/utils/env.js
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+ /* eslint no-undef: "off" */
 export const debug = process.env.NODE_ENV === 'development';
```
thisconnect, line 18: An alternative would be to set `node: true`, here https://github.com/digitalbitbox/bitbox-wallet-app/blob/4457dde3491a63ac184e00da9f0d328d9c2fe227/frontends/web/.eslintrc.js#L3

but node is only used during build, so I prefer this way.

frontends/web/src/components/anchor/anchor.tsx
benma: Why did it build successfully without importing it before?

benma (approved):

frontends/web/src/components/anchor/anchor.tsx
thisconnect: Correction: Typescript doesn't complain but typescript-eslint. Finally found one having same issue: https://github.com/typescript-eslint/typescript-eslint/issues/2477#issuecomment-686845188

I tried going down that rabit hole (using @types/react), but stopped now and postpone until we move to TS4 + PreactX.

PR #1028, opened 2020-10-03
Frontend: eslint enable and autofix no unknown property

thisconnect:


benma (approved):
utACK. did you test?

thisconnect:
> utACK. did you test?

yep, strokeWidth is correctly rendered to `stroke-width`, but doesn't do anything anyways :) .  We could remove all occurrences on the `<svg>` element, because it needs to be directly on the shape element
`<circle>, <ellipse>, <line>, <path>, ...` and so on.
 https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-width

Should I change it in this PR?

PR #1029, opened 2020-10-04
backend/rates: add ILS

benma:
Israeli Shekel


x1ddos (approved):
LGTM but could you also add it here in the staging-portfolio branch: https://github.com/digitalbitbox/bitbox-wallet-app/blob/4457dde/backend/rates/history.go#L37

Otherwise, I can almost guarantee we'll merge portfolio as is and forget to add ILS when fetching rates from coingecko.

PR #1035, opened 2020-10-08
backend/rates: support historical exchange rates for all coins

x1ddos:
All active coin/fiat pairs are taken from the config on startup.
A follow up commit will make the updater enable/disable pairs
when the config is changed at runtime from the UI.

This migrates one bit of frontend's opaque config fiatList to the
backend side since both now need to be aware which fiat currencies
are active.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -196,6 +196,38 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.ratesUpdater = rates.NewRateUpdater(hclient)
 	backend.ratesUpdater.Observe(backend.Notify)
 	backend.ratesUpdater.Start()
+	for _, fiat := range config.AppConfig().Backend.FiatList {
+		if config.AppConfig().Backend.CoinActive(coin.CodeBTC) {
+			backend.ratesUpdater.EnableHistoryPair("BTC", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeTBTC) {
+			backend.ratesUpdater.EnableHistoryPair("TBTC", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeRBTC) {
+			backend.ratesUpdater.EnableHistoryPair("RBTC", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeLTC) {
+			backend.ratesUpdater.EnableHistoryPair("LTC", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeTLTC) {
+			backend.ratesUpdater.EnableHistoryPair("TLTC", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeETH) {
+			backend.ratesUpdater.EnableHistoryPair("ETH", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeTETH) {
+			backend.ratesUpdater.EnableHistoryPair("TETH", fiat)
+		}
+		if config.AppConfig().Backend.CoinActive(coin.CodeRETH) {
+			backend.ratesUpdater.EnableHistoryPair("RETH", fiat)
+		}
+		for _, token := range config.AppConfig().Backend.ETH.ActiveERC20Tokens {
+			// The prefix is stripped on the frontend and in app config.
+			// TODO: Unify the prefix with frontend and erc20.go, and possibly
+			// move all this with coins/coin/code or eth/erc20.
+			backend.ratesUpdater.EnableHistoryPair("eth-erc20-"+token, fiat)
+		}
+	}
 
 	backend.banners = banners.NewBanners()
 	backend.banners.Observe(backend.Notify)
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -105,6 +105,13 @@ type Backend struct {
 	ETH  ethCoinConfig `json:"eth"`
 	TETH ethCoinConfig `json:"teth"`
 	RETH ethCoinConfig `json:"reth"`
+
+	// FiatList contains all enabled fiat currencies.
+	// These are used in the UI as well as by RateUpdater to fetch historical exchange rates.
+	FiatList []string `json:"fiatList"`
+	// MainFiat is the fiat currency used as a default for computing account portfolio data
+	// and transaction amounts.
+	MainFiat string `json:"mainFiat"`
 }
 
 // CoinActive returns the Active setting for a coin by code.
@@ -256,6 +263,9 @@ func NewDefaultAppConfig() AppConfig {
 			RETH: ethCoinConfig{
 				ActiveERC20Tokens: []string{},
 			},
+			// Copied from frontend/web/src/components/rates/rates.tsx.
+			FiatList: []string{"USD", "EUR", "CHF"},
+			MainFiat: "CHF",
 		},
 	}
 }
@@ -282,6 +292,12 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) (*Config
 		accountsConfig:         newDefaultAccountsonfig(),
 	}
 	config.load()
+	appconf := config.appConfig
+	migrateFiatList(&appconf)
+	migrateFiatCode(&appconf)
+	if err := config.SetAppConfig(appconf); err != nil {
+		return nil, errp.WithStack(err)
+	}
 	return config, nil
 }
 
@@ -366,4 +382,39 @@ func (config *Config) save(filename string, conf interface{}) error {
 		return errp.WithStack(err)
 	}
 	return errp.WithStack(ioutil.WriteFile(filename, jsonBytes, 0644)) // #nosec G306
+}
+
+// migrateFiatList moves fiatList from appconf.Frontend to appconf.Backend.
+// This is because with the account portfolio feature, backend needs to know
+// which fiat currencies are enabled to fetch historical exchange rates.
+func migrateFiatList(appconf *AppConfig) {
+	frontconf, ok := appconf.Frontend.(map[string]interface{})
+	if !ok {
+		return // nothing to migrate
+	}
+	fiats, ok := frontconf["fiatList"].([]interface{})
+	if !ok {
+		return // nothing to migrate
+	}
+	appconf.Backend.FiatList = make([]string, len(fiats))
+	for i, f := range fiats {
+		if v, ok := f.(string); ok {
+			appconf.Backend.FiatList[i] = v
+		}
+	}
+	delete(frontconf, "fiatList")
+}
+
+// migrateFiatCode moves fiatCode from appconf.Frontend to appconf.Backend
+// to aid the backend in constructing data series with correct main fiat code
+// for portfolio feature.
+func migrateFiatCode(appconf *AppConfig) {
+	frontconf, ok := appconf.Frontend.(map[string]interface{})
+	if !ok {
+		return // nothing to migrate
+	}
+	if code, ok := frontconf["fiatCode"].(string); ok {
+		appconf.Backend.MainFiat = code
+		delete(frontconf, "fiatCode")
+	}
 }
--- a/backend/rates/history.go
+++ b/backend/rates/history.go
@@ -17,19 +17,27 @@ import (
 const coingeckoAPIV3 = "https://api.coingecko.com/api/v3"
 
 var (
-	// Copied from https://api.coingecko.com/api/v3/coins/list.
-	// The keyes must match entries in coins slice.
+	// Values are copied from https://api.coingecko.com/api/v3/coins/list.
+	// TODO: Replace keys with coin.Code.
 	geckoCoin = map[string]string{
-		"BTC":  "bitcoin",
-		"LTC":  "litecoin",
-		"ETH":  "ethereum",
-		"USDT": "tether",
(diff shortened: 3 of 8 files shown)


x1ddos:
Sorry, there was one typo in config migration. Should be good now.

x1ddos:
Pleasing golang-ci linters with US English. PTAL.

benma:
> This migrates one bit of frontend's opaque config fiatList to the

Any plans to migrate the existing settings? Otherwise the users will lose them.

x1ddos:
> Any plans to migrate the existing settings? Otherwise the users will lose them.

What do you mean? It's migrated in the `migrateFiatList` func in this commit.

backend/rates/history.go
```diff
@@ -49,16 +49,43 @@ var (
...
+func (updater *RateUpdater) EnableHistoryPair(coin, fiat string) {
+	if geckoCoin[coin] == "" || geckoFiat[fiat] == "" {
+		return
+	}
+	updater.historyMu.Lock()
+	defer updater.historyMu.Unlock()
+	ctx, cancel := context.WithCancel(context.Background())
+	go updater.historyUpdateLoop(ctx, coin, fiat)
+	go updater.backfillHistory(ctx, coin, fiat)
+	updater.historyGo[coin+fiat] = cancel
```
x1ddos, line 65: There's a mistake here. This shouldn't happen if already exists. I even wrote so in the doc comment. Will fix

benma:
> > Any plans to migrate the existing settings? Otherwise the users will lose them.
> 
> What do you mean? It's migrated in the `migrateFiatList` func in this commit.

Hmm I commented after testing without checking the code. Before:

before

After:

image

backend/config/config.go
```diff
@@ -394,3 +399,26 @@ func (config *Config) migrateInfura(appConfig AppConfig) AppConfig {
...
+// which fiat currencies are enabled to fetch historical exchange rates.
+func migrateFiatList(appconf *AppConfig) {
+	if len(appconf.Backend.FiatList) > 0 {
+		return // already migrated
+	}
+	frontconf, ok := appconf.Frontend.(map[string]interface{})
+	if !ok {
+		return // can't migrate
+	}
+	if fiats, ok := frontconf["fiatList"].([]string); ok {
```
benma, line 414: I think this conversion does not work.

if you do it without the `ok` check, the panic says:

```
panic: interface conversion: interface {} is []interface {}, not []string [recovered]
	panic: interface conversion: interface {} is []interface {}, not []string
```

backend/backend.go
```diff
@@ -186,6 +186,21 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.ratesUpdater = rates.NewRateUpdater(hclient)
 	backend.ratesUpdater.Observe(backend.Notify)
 	backend.ratesUpdater.Start()
+	for _, fiat := range config.AppConfig().Backend.FiatList {
```
benma, line 189: Mabye this code should be in `initAccounts()`? Then it will work even without an app restart if you toggle accounts in the settings, as toggling accounts will invoke `initAccounts()`.

Just need to take care to that EnableHistoryPair should only run once per pair.

backend/rates/history.go
```diff
@@ -49,16 +49,43 @@ var (
 	}
 )
 
+// EnableHistoryPair spins up a new goroutine to periodically update
+// and backfill historical exchange rates.
+// If a pair is unsupported or already enabled, EnableHistoryPair does nothing.
+// To stop periodic updates for this coin/fiat pair, the callers can use DisableHistoryPair.
+func (updater *RateUpdater) EnableHistoryPair(coin, fiat string) {
+	if geckoCoin[coin] == "" || geckoFiat[fiat] == "" {
+		return
```
benma, line 58: log an error, or possibly even panic? should never happen

backend/rates/history.go
benma: Do you mean this?

>  // If a pair is unsupported or already enabled, EnableHistoryPair does nothing.

As I don't see code that aborts early if already enabled.

backend/config/config.go
x1ddos: it's possible your ~/.config/bitbox/config.json has a different value? could you check please.

This is what you get for "opaque" things when trying to parse them instead of endorsing strong types. :)

backend/backend.go
benma: Hmm same for when changing the fiat settings. Let's postpone fixing both of it to a new PR.

backend/config/config.go
x1ddos: You can see how frontend was storing the list:

image

That's why I picked `[]string` as the type.

backend/rates/history.go
```diff
@@ -49,16 +49,43 @@ var (
...
+	ctx, cancel := context.WithCancel(context.Background())
+	go updater.historyUpdateLoop(ctx, coin, fiat)
+	go updater.backfillHistory(ctx, coin, fiat)
+	updater.historyGo[coin+fiat] = cancel
+}
+
+// DisableHistoryPair deactivates backfill and periodic updates
+// of historical exchange rates for the given coin/fiat pair.
+// If a pair is unsupported or wasn't active, DisableHistoryPair does nothing.
+func (updater *RateUpdater) DisableHistoryPair(coin, fiat string) {
```
benma, line 71: This is never called. It should probably be called in the place where you call `EnableHistoryPair` in the backend, for all other pairs that are not active?

Maybe easiest to by default disable all possible pairs at start, then enable the active pairs after.

benma (changes requested):
tested ACK, apart from the fiat settings migration.

Can you migrate the fiatCode too? Will be helpful to decide which currency to display the chart in.

backend/rates/history.go
x1ddos: True, I added it for consistency, to be used in the next commit where pairs are enabled/disabled at runtime from the UI, but I have another idea how to disable them.

Removed because all pairs are in fact disabled by default at the startup.

backend/config/config.go
x1ddos: I know what's happening here. Check out the rates.tsx in frontend:

image

If you never changed fiat settings, you'll never have it stored in the app config.json and it's why this migration fails. I'll fix it.

x1ddos (commented):
Still need to check how to best handle fiatList migration from the frontend. Juts need to try a couple things.

backend/config/config.go
x1ddos: Although ILS is enabled in your screenshot., which is not in the default "selected" list... I wonder how that happened.

backend/rates/history.go
```diff
@@ -49,16 +50,37 @@ var (
 	}
 )
 
+// EnableHistoryPair spins up a new goroutine to periodically update
+// and backfill historical exchange rates.
+// If a pair is unsupported or already enabled, EnableHistoryPair does nothing.
+func (updater *RateUpdater) EnableHistoryPair(coin, fiat string) {
+	coin, fiat = strings.ToUpper(coin), strings.ToUpper(fiat)
```
benma, line 57: i would remove this, too magical. better to eventually use proper types for coin and fiat.

why not keep them in the original case? BTC will be BTC, and `dai0x6b17` will be `dai0x6b17` instead of `DAI0X6B17`.

Fyi, the coingecko map has wrong entries for dai and sai, should be the above and `sai0x89d2`.

backend/backend.go
```diff
@@ -186,6 +186,20 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
+		if config.AppConfig().Backend.BitcoinActive {
```
benma, line 190: maybe using `CoinActive()` is better, like here? https://github.com/digitalbitbox/bitbox-wallet-app/blob/eeff1391b11b658e07bfb5a843b821c15cb79892/backend/backend.go#L337

backend/rates/history.go
benma: it would be nice to use `coin.Code` (as returned by account.Coin().Code()) for the `coin` type. This will save another mapping from coin codes to rates.go codes.

```
coin.CodeBTC:                     "BTC",
			coin.CodeTBTC:                    "BTC",
			coin.CodeLTC:                     "LTC",
			coin.CodeTLTC:                    "LTC",
			coin.CodeETH:                     "ETH",
			coin.CodeRETH:                    "ETH",
			coin.CodeTETH:                    "ETH",
			coin.Code("eth-erc20-usdt"):      "USDT",
			coin.Code("eth-erc20-usdc"):      "USDC",
			coin.Code("eth-erc20-bat"):       "BAT",
			coin.Code("eth-erc20-sai0x89d2"): "SAI0X89D2",
			coin.Code("eth-erc20-dai0x6b17"): "DAI0X6B17",
			coin.Code("eth-erc20-link"):      "LINK",
			coin.Code("eth-erc20-mkr"):       "MKR",
			coin.Code("eth-erc20-zrx"):       "ZRX",
```

(the erc20 codes are from backend/erc20.go).

backend/backend.go
```diff
@@ -186,6 +186,20 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
+			backend.ratesUpdater.EnableHistoryPair("BTC", fiat)
+		}
+		if config.AppConfig().Backend.LitecoinActive {
+			backend.ratesUpdater.EnableHistoryPair("LTC", fiat)
+		}
+		if config.AppConfig().Backend.EthereumActive {
+			backend.ratesUpdater.EnableHistoryPair("ETH", fiat)
+		}
+		for _, token := range config.AppConfig().Backend.ETH.ActiveERC20Tokens {
+			backend.ratesUpdater.EnableHistoryPair(token, fiat)
```
benma, line 200: for sai and dai, this will pass `sai0x89d2` and `dai0x6b17`, but in rates.go they are `SAI` and `DAI`.

probably better to adjust the coin code for those in rates.go instead.

x1ddos: Oh, I didn't realize these token codes are different between rates.go and the frontend. I can see the map in settings.tsx: https://github.com/digitalbitbox/bitbox-wallet-app/blob/eeff139/frontends/web/src/routes/settings/settings.tsx#L62-L71.  Let me see... but yeah, maybe using `coin.Code` is better.

backend/rates/history.go
x1ddos: but in the app config.json, they're stored like this:

```json
        "eth": {
            "nodeURL": "etherscan+https://api.etherscan.io/api",
            "transactionsSource": "etherScan",
            "activeERC20Tokens": [
                "usdc",
                "link",
                "sai0x89d2",
                "dai0x6b17"
            ]
        },
```

so, if anything, the map should probably look like:

```go
// A map with values taken from https://api.coingecko.com/api/v3/coins/list,
// representing CoinGecko IDs.
// The keyes must match ethCoinConfig.ActiveERC20Tokens.
var geckoCoin = map[string]string{
	coin.CodeBTC:           "bitcoin",
	coin.CodeLTC:           "litecoin",
	coin.CodeETH:           "ethereum",
	coin.Code("usdt"):      "tether",
	coin.Code("link"):      "chainlink",
	coin.Code("mkr"):       "maker",
	coin.Code("zrx"):       "0x",
	coin.Code("dai0x6b17"): "dai",
	coin.Code("sai0x89d2"): "sai",
	coin.Code("bat"):       "basic-attention-token",
	coin.Code("usdc"):      "usd-coin",
}
```

Anyway, I like the idea of using `coin.Code` instead of a string. Will do.

backend/config/config.go
benma: > Although ILS is enabled in your screenshot., which is not in the default "selected" list... I wonder how that happened.

It was not the default selection, it was my selection...

backend/rates/history.go
x1ddos: Ok, I've added them with the eth-erc20- prefix but as strings for now. Left coin.Code as a TODO.

backend/config/config.go
x1ddos: So, there were two problems. 1st is like you said, []interface{} vs []string. 2nd is the frontend never stored fiatList or fiatCode in the config.json until you actually changed it from the UI.

Anyway, all should work now.

x1ddos (commented):
PTAL

backend/config/config.go
```diff
@@ -367,3 +383,38 @@ func (config *Config) save(filename string, conf interface{}) error {
 	}
 	return errp.WithStack(ioutil.WriteFile(filename, jsonBytes, 0644)) // #nosec G306
 }
+
+// migrateFiatList moves fiatList from appconf.Frontend to appconf.Backend.
+// This is because with the account portfolio feature, backend needs to know
+// which fiat currencies are enabled to fetch historical exchange rates.
+func migrateFiatList(appconf *AppConfig) {
+	frontconf, ok := appconf.Frontend.(map[string]interface{})
```
benma, line 391: nice, i was gonna mention to switch the check from backend to frontend! :tada:

benma (approved):
cool cool cool

x1ddos:
```
+ golangci-lint run
backend/config/config.go:295: File is not `gofmt`-ed with `-s` (gofmt)
	appconf:= config.appConfig
```

I was pretty sure I did this. Anyway, updated.

x1ddos:
I think the next coming up is add rate limiter. On my machine here, with all coins enabled and 3 fiats I'm already getting 429 Too Many Requests errors.

PR #1037, opened 2020-10-08
backend/accounts: fix timeseries computation for erc20 tokens

benma:
The fee is paid in Ether and is accounted for in the ETH account
already.

### DIFF
--- a/backend/accounts/transaction.go
+++ b/backend/accounts/transaction.go
@@ -66,6 +66,9 @@ type TransactionData struct {
 	// Fee is nil for a receiving tx. The fee is only displayed (and relevant) when sending funds
 	// from the wallet.
 	Fee *coin.Amount
+	// FeeIsDifferentUnit is true if the fee is paid in a different unit than the main transaction
+	// amount. Example: ERC20-tokens fees are paid in ETH.
+	FeeIsDifferentUnit bool
 	// Time of confirmation. nil for unconfirmed tx or when the headers are not synced yet.
 	Timestamp *time.Time
 	// TxID is the tx ID.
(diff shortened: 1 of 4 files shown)


backend/accounts/transaction.go
```diff
@@ -152,12 +155,12 @@ func NewOrderedTransactions(txs []*TransactionData) OrderedTransactions {
 		case TxTypeSend:
 			balance.Sub(balance, tx.Amount.BigInt())
 			// Subtract fee as well.
-			if tx.Fee != nil {
+			if tx.Fee != nil && !tx.FeeIsDifferentUnit {
```
x1ddos, line 158: But when the fee is a different unit, where does it go?

benma: If you make an erc20 token tx, the fee is paid from the main Ethereum account. There is a 0-value transaction shown in the ETH account for each erc20 tx, that only has fees.

In the portfolio, this is accounted for in the Ethereum timeseries.

benma: In other words, if the fee is paid in a different unit, it does not affect the account's balance or timeseries.

x1ddos: Ah ok. But the actual fee is still accounted for when computing ETH balance, right.

>  if the fee is paid in a different unit, it does not affect the account's balance or timeseries.

The `FeeIsDifferentUnit` made me think it should still be substracted but in a different way.
How about naming it a `PhantomFee` or `FeeIsPhantom`, similar to `PhantomData` in Rust :)

Anyway, as you prefer. If you keep the original name, add a comment to `FeeIsDifferentUnit` field that "when true, the fee amount does not count towards the balance of the account associated with the transaction because it's already paid in another transaction"?

x1ddos (approved):
LGTM

backend/accounts/transaction.go
benma: Oops saw it too late. Here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1039

PR #1040, opened 2020-10-09
backend/eth: properly rate limit EtherScan

benma:
We made one EtherScan instance per ETH and ERC20 token, all of them
firing at once. This was usually not a problem, except for in the
account summary / portfolio page, where all accounts are loaded in
quick succession.

This commit moves out the rate limiting to a new httpClient reused by
all instances.

### DIFF
--- a/util/ratelimit/ratelimit.go
+++ b/util/ratelimit/ratelimit.go
@@ -1,0 +1,56 @@
+// Copyright 2020 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package ratelimit provides a util http.RoundTripper which rate limits its calls.
+package ratelimit
+
+import (
+	"net/http"
+	"time"
+)
+
+// RateLimitedHTTPTransport is a http.RoundTripper that rate limits the requests, waiting at least
+// `callInterval` between requests.
+type RateLimitedHTTPTransport struct {
+	base         http.RoundTripper
+	rateCh       chan struct{}
+	callInterval time.Duration
+}
+
+// NewRateLimitedHTTPTransport make a new rate limited http transport.
+// If base is nil, http.DefaultTransport is used.
+func NewRateLimitedHTTPTransport(
+	base http.RoundTripper, callInterval time.Duration) *RateLimitedHTTPTransport {
+	if base == nil {
+		base = http.DefaultTransport
+	}
+	transport := &RateLimitedHTTPTransport{
+		base:         base,
+		rateCh:       make(chan struct{}),
+		callInterval: callInterval,
+	}
+	go transport.tick()
+	return transport
+}
+
+func (transport *RateLimitedHTTPTransport) tick() {
+	transport.rateCh <- struct{}{}
+	time.AfterFunc(transport.callInterval, transport.tick)
+}
+
+// RoundTrip implements http.RoundTripper, rate limiting the requests.
+func (transport *RateLimitedHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	<-transport.rateCh
+	return transport.base.RoundTrip(req)
+}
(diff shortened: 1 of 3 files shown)


util/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,56 @@
...
+		base:         base,
+		rateLimiter:  time.After(0), // 0 so the first call does not wait.
+		callInterval: callInterval,
+	}
+}
+
+// RoundTrip implements http.RoundTripper, rate limiting the requests.
+func (transport *RateLimitedHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	transport.mu.Lock()
+	defer transport.mu.Unlock()
```
x1ddos, line 50: What about something like this, without a mutex?

```go
func NewRateLimitedHTTPTransport(...) {
  transport = &RatLimitedHTTPTransport {
    base: base,
    callInterval: ...,
    rateCh: make(chan struct{}), // unbuffered to prevent request bursts
  }
  go transport.tick() // allow first request to fire right away
  return transport
}

func (tr *RateLimitedHTTPTransport) tick() {
  tr.rateCh <- struct{}{}
  time.AfterFunc(tr.callInterval, tr.tick)
}

func (tr *RateLimitedHTTPTransport) RoundTrip(...) {
  <-tr.rateCh
  return tr.base.RoundTrip(req)
}
```

benma: At this point I'd leave the original, as the new one has no mutex, but is not less complicated. Kinda equivalent imho.

x1ddos: But we can use it for coingecko API too in the future. I was already planning on it :)
it's a nice separate package. I like it a lot.

x1ddos: > Kinda equivalent imho.

but the original also had a mutex.

benma: So are you saying it's good or it should be changed? :P

x1ddos: I'm saying the separate ratelimit package is good to have :)
But we can do away without a mutex which is always better.

> edit: unbuffered sounds like a memory leak? the timer.Tick has a buffer of 1 and drops more ticks on the floor if the one previously was not processed.

There's no leak. I mean, the time.Ticker drops ticks to:

> ... adjust the intervals or drops ticks to make up for slow receivers

But there is a memory leak in the sense that once a `NewRateLimitedHTTPTransport` is called, it can never be garbage collected because there's always a timer referencing its `tr.tick` method. In this sense, time.Tick also leaks: https://golang.org/pkg/time/#Tick.

x1ddos: > In this sense, time.Tick also leaks: https://golang.org/pkg/time/#Tick.

btw, so does the time.After in your original version: https://golang.org/pkg/time/#After. so, no change in regards to the amount of memory leaks :)

x1ddos: But I don't understand the issue. One less mutex is always good. Why not use it?

benma: All good, I added it like you said. I just misunderstood `unbuffered`.

x1ddos: Oh, I see. If it were with a buffer, say `make(chan struct{}, 1)`, then it would've been possible to have a short burst of 2 QPS. There would be 1 item already sent in the channel due to the buffer of 1, and another one waiting to be sent in `tr.tick`.

Let's say there are 2 independent callers making requests with this transport. It receives first item already in the channel with `<-tr.rateCh` right away and passes the request on to the base transport. This frees rateCh's buffer and the other item which was waiting to be sent goes in and `<-tr.rateCh` receives it. The second caller gets its chance to make the request.

This happens very fast, of course. So, etherscan would see 2 requests coming within the same second.

These quick bursts is actually a feature of a rate limiter quite often. We don't need it at moment but can always add in the future.

x1ddos: > So, etherscan would see 2 requests coming within the same second.

forgot to add: but no more than 2. a 3rd would be waiting at least `callInterval` duration because with a channel buffer of 1, `tr.rateCh <- struct{}{}` blocks if there's already 1 item in the channel: the buffer is full. 

If we made a buffered channel with `make(chan struct{}, 9)`, then it would be possible to have short bursts of around 10 QPS but no more. 11th would again be waiting `callInterval` because the channel fills up completely at 9 items plus one extra "stuck" in `tr.tick` at `tr.rateCh <- struct{}{}`.

x1ddos: >  I added it like you said

wow. you actually copied it. I hope it works :sweat_smile:

x1ddos (approved):
LGTM

util/ratelimit/ratelimit.go
benma: I tested it no worries ;)

PR #1044, opened 2020-10-12
backend/rates: make historical rates settings configurable at runtime

x1ddos:
This is done in two places: at the backend startup and whenever active
coin accounts or selected fiats list changes. In both cases, all
historical rates goroutines are shut down and then started again
but only for the selected coin/fiat pairs.

Maybe a better way would've been to shut down only those goroutines
for which no coin or fiat is active anymore, but this complicates a
simple function and with an added jitter in a follow up commit will be
unnecessary.


benma (approved):
utACK

x1ddos:
Ok, removed strings.ToUpper and rebased on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1047 because we're actually lowercase everywhere.

x1ddos:
Rebased on latest staging-portfolio.

PR #1045, opened 2020-10-12
backend/rates: stop all RateUpdater's goroutines when Close'ing backend

x1ddos:
This is useful for an upcoming commit where RateUpdater starts using a
persistent database for historical exchange rates which typically needs
to terminal all transactions.

It is also useful in testing where multiple instances of backends may
introduce lots of loose updater goroutines, if not shut down at the
end of the test.

Builds on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1044.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -1151,6 +1151,7 @@ func (backend *Backend) Environment() Environment {
 func (backend *Backend) Close() error {
 	errors := []string{}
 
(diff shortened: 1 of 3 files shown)


benma (approved):
utACK

benma:
nice stuff

x1ddos:
Rebased onto the latest staging-portfolio.

PR #1047, opened 2020-10-12
backend/rates: use lowercase coin symbols

x1ddos:
Turns out, they are lowercase everywhere in the backend.

The original CryptoCompare client uses uppercase.
This is left as a TODO for the future.

Note that frontend seem to use all uppercase, both coins and fiats.

### DIFF
--- a/backend/rates/history.go
+++ b/backend/rates/history.go
@@ -20,24 +20,25 @@ var (
 	// Values are copied from https://api.coingecko.com/api/v3/coins/list.
 	// TODO: Replace keys with coin.Code.
 	geckoCoin = map[string]string{
-		"BTC": "bitcoin",
-		"LTC": "litecoin",
-		"ETH": "ethereum",
+		"btc": "bitcoin",
+		"ltc": "litecoin",
+		"eth": "ethereum",
 		// Useful for testing with testnets.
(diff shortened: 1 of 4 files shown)


backend/rates/history.go
```diff
@@ -20,15 +20,15 @@ var (
...
-		"TBTC": "bitcoin",
-		"RBTC": "bitcoin",
-		"TLTC": "litecoin",
-		"TETH": "ethereum",
-		"RETH": "ethereum",
+		"tbtc": "bitcoin",
+		"rbtc": "bitcoin",
+		"tltc": "litecoin",
+		"teth": "ethereum",
+		"reth": "ethereum",
```
benma, line 31: a little bit below:

```
"eth-erc20-dai0x6b17": "dai",
"eth-erc20-sai0x89d2": "sai",
```

instead of the dai that is there.

benma (changes requested):

backend/rates/history.go
x1ddos: :see_no_evil: done
thanks!

x1ddos: (ordered alphabetically)

x1ddos (commented):
PTAL

PR #1048, opened 2020-10-12
backend/rates: add the missing SAI token to CryptoCompare

x1ddos:
This is used in most recent exchange rates to show fiat value when
making a transaction.

Historical rates with CoinGecko already has it.

### DIFF
--- a/backend/rates/rates.go
+++ b/backend/rates/rates.go
@@ -35,7 +35,7 @@ import (
 )
 
 // If modified, also update geckoCoin map.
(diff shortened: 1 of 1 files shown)


benma (approved):

x1ddos:
> Historical rates with CoinGecko already has it.

Turns out this was a lie. Added to CoinGecko in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1047.

PR #1054, opened 2020-10-19
backend/rates: cache historical data in a persistent bolt DB

x1ddos:
RateUpdater loads the data on ReconfigureHistory for active coin/fiat
pairs and dumps new data on disk during each update loop iteration.
All PriceAt look ups use RateUpdater.history in-memory map as before.
In other words, the cache layer is transparent to RateUpdater users.

RateUpdater can still function without the cache but the app users
may be noticeably rate-limited, especially when connected over a VPN
or Tor.

### DIFF
--- a/backend/rates/db.go
+++ b/backend/rates/db.go
@@ -1,0 +1,66 @@
+package rates
+
+import (
+	"encoding/binary"
+	"math"
+	"path/filepath"
+	"time"
+
+	bbolt "github.com/coreos/bbolt"
+)
+
+func openRatesDB(dir string) (*bbolt.DB, error) {
+	opt := &bbolt.Options{Timeout: 5 * time.Second} // network disks may take long
+	return bbolt.Open(filepath.Join(dir, "rates.db"), 0600, opt)
+}
+
+// loadHistoryBucket loads data from an updater.historyDB bucket identified by the key.
+// The returned value is sorted by timestamp in ascending order.
+func (updater *RateUpdater) loadHistoryBucket(key string) ([]exchangeRate, error) {
+	var rates []exchangeRate
+	err := updater.historyDB.View(func(tx *bbolt.Tx) error {
+		bucket := tx.Bucket([]byte(key))
+		if bucket == nil {
+			return nil // no history exists for this key
+		}
+		return bucket.ForEach(func(k, v []byte) error {
+			timestamp := binary.BigEndian.Uint64(k)
+			value := math.Float64frombits(binary.BigEndian.Uint64(v))
+			rates = append(rates, exchangeRate{
+				value:     value,
+				timestamp: time.Unix(int64(timestamp), 0),
+			})
+			return nil
+		})
+	})
+	// bbolt's readme says it stores the keys in byte-sorted order within a bucket.
+	// It doesn't say anything about what order takes place during the reads.
+	// However, tests seem to confirm it's a sequential read, in the same order
+	// in which the keys are stored. Hope it stays that way.
+	// See vendor/github.com/coreos/bbolt/README.md for some more details.
+	//
+	// Given the above, assume rates is already sorted by timestamp in ascending order.
+	return rates, err
+}
+
+// dumpHistoryBucket stores rates in a DB bucket identified by the key.
+// It assumes rates are already sorted by timestamp in ascending order.
+func (updater *RateUpdater) dumpHistoryBucket(key string, rates []exchangeRate) error {
+	return updater.historyDB.Update(func(tx *bbolt.Tx) error {
+		tx.DeleteBucket([]byte(key)) //nolint:errcheck // don't care: next line will fail anyway
+		bucket, err := tx.CreateBucket([]byte(key))
+		if err != nil {
+			return err
+		}
+		for _, rate := range rates {
+			var tsbytes [8]byte
+			binary.BigEndian.PutUint64(tsbytes[:], uint64(rate.timestamp.Unix()))
+			var vbytes [8]byte
+			binary.BigEndian.PutUint64(vbytes[:], math.Float64bits(rate.value))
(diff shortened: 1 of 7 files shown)


x1ddos:
macos CI error is unrelated: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1055

backend/rates/db.go
```diff
@@ -0,0 +1,64 @@
...
+			timestamp := binary.LittleEndian.Uint64(k)
+			value := math.Float64frombits(binary.LittleEndian.Uint64(v))
+			rates = append(rates, exchangeRate{
+				value:     value,
+				timestamp: time.Unix(int64(timestamp), 0),
+			})
+			return nil
+		})
+	})
+	// Unfortunately, bbolt doesn't specify what kind of sorting order is used, if at all,
```
x1ddos, line 36: Or maybe I should've used big endian in bolt db keys. But still, it's a shame I didn't find precise docs on the sorting order.

backend/rates/db.go
```diff
@@ -0,0 +1,64 @@
...
+// dumpHistoryBucket stores rates in a DB bucket identified by the key.
+// It assumes rates are already sorted by timestamp in ascending order.
+func (updater *RateUpdater) dumpHistoryBucket(key string, rates []exchangeRate) error {
+	return updater.historyDB.Update(func(tx *bbolt.Tx) error {
+		tx.DeleteBucket([]byte(key)) //nolint:errcheck // don't care: next line will fail anyway
+		bucket, err := tx.CreateBucket([]byte(key))
+		if err != nil {
+			return err
+		}
+		for _, rate := range rates {
```
benma, line 53: what's the reason for storing each entry as an individual key:value row, instead of dumping all of it in serialized as one value? If you are loading and dumping all of them every time, there would be less overhead in storage and possibly performance?

backend/rates/db.go
benma: https://github.com/etcd-io/bbolt#iterating-over-keys:

> Bolt stores its keys in byte-sorted order within a bucket. This makes sequential iteration over these keys extremely fast.

BigEndian should work I think.

benma (changes requested):

backend/rates/db.go
x1ddos: Reason: so that we have alternatives this dump/load version in the future. Maybe we'll find out in-memory map or its lock or something else is going to be a bottleneck and will want to switch to using bbold for the actual look ups in PriceAt.

Performance: it's an interesting assumption. To verify that, I wrote a simple benchmark where it dumps and loads data in 3 different ways: individual rows, JSON and Gob. Here are the results from my laptop with i7-8565U CPU at 1.8GHz:

```
$ cd backend/rates/
$ go test -bench=. -test.run=Benchmark
goos: linux
goarch: amd64
pkg: github.com/digitalbitbox/bitbox-wallet-app/backend/rates
BenchmarkDumpHistory/individual-8                    402           2912290 ns/op
BenchmarkDumpHistory/json-8                          282           4073714 ns/op
BenchmarkDumpHistory/gob-8                           938           1258093 ns/op
BenchmarkLoadHistory/individual-8                   4164            293393 ns/op
BenchmarkLoadHistory/json-8                          216           5476824 ns/op
BenchmarkLoadHistory/gob-8                          1716            695656 ns/op
PASS
ok      github.com/digitalbitbox/bitbox-wallet-app/backend/rates        8.675s
```

So, a single key with a JSON or Gob -encoded value is faster on writes (DumpHistory/xxx) but slower on reads (LoadHistory/xxx). This makes sense because:

> Bolt is good for read intensive workloads. Sequential write performance is also fast but random writes can be slow.

(https://github.com/etcd-io/bbolt#caveats--limitations)

In our case, we load the database once at startup (read) and dump new data every now and then from goroutines (write). We also load data in ReconfigureHistory in a goroutine from the backend handlers.

If we used one key with the whole bucket serialized as a value, if anything, it would most likely translate in slower startup and no gain during runtime.

Overhead in storage: I didn't include it in the benchmark but from my adhoc testing, 5000 of invidivual rows take about 1Mb of disk space, JSON is about 0.5Mb and Gob is 0.26Mb. Didn't try with protobuf but three versions I used already give good metrics to reason about imho. This storage overhead seems like a drop in the ocean compared to 100s of Mb of blockchain headers.

x1ddos (commented):
PTAL

backend/rates/db.go
benma: You could open an issue to clarify the wording in the docs. I am pretty sure the cursor iterates in the order it is stored, i.e. sorted. Many key-value databases do this and you get indexing for free if you encode the keys right.

This issue seems to confirm it, but it did not have a lot of discussion ;) https://github.com/etcd-io/bbolt/issues/101

backend/rates/db.go
benma: wow, this answer goes above and beyond. nice and ACK!

backend/rates/db.go
x1ddos: Yeah, I guess I was just complaining that one has to look at the readme and chase down issues. Instead, a single doc comment sentence in `ForEach` and `Cursor` would've been much more clear, committing to it so there are no breaking changes in the same major version. I've just sent them https://github.com/etcd-io/bbolt/pull/250.

x1ddos:
Failed CI on macOS is unrelated and should be resolved by https://github.com/digitalbitbox/bitbox-wallet-app/commit/b8d785a3ba99dfcb4bfa11ec9b7e8d6474abd6ab. The rest seems ok.

benma (approved):
tested with https://github.com/digitalbitbox/bitbox-wallet-app/pull/1046, works nicely!

x1ddos:
b102912 just adds a couple more benchmarks. no changes in the actual code.

PR #1058, opened 2020-10-28
backend/coin: add Coin.Name()

benma:
The "Bitcoin"/"Litecoin" etc. strings we define for accounts are in
fact coin names.

We mov this to the coin so it can be used to show the name of the coin
to the user, e.g. in the account summary page (table headers, which
currently just shows the coin code).

The accounts automatically added from a keystore just adopt the coin
name, so it behaves the same as before.


x1ddos (approved):
LGTM

PR #1059, opened 2020-10-30
backend/rates: add jitter to history loops and make them faster

x1ddos:
The forward history update loop was iterating every 10min. This
interval was chosen after an approximate time of a new block on the
chain. However, it may be too slow in the worst case where RateUpdater
provides data stale up to 10min while already available from the API.

This commit reduces the potential staleness to only up to ~1.5min.

### DIFF
--- a/backend/rates/history.go
+++ b/backend/rates/history.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"sort"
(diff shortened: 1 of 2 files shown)


benma (approved):
> This interval was chosen after an approximate time of a new block on the chain.

Interesting, wasn't aware. Fyi, LTC that has 2.5min blocks or ETH that has 1min blocks.

x1ddos:
> > This interval was chosen after an approximate time of a new block on the chain.
> 
> Interesting, wasn't aware. Fyi, LTC that has 2.5min blocks or ETH that has 1min blocks.

Yeah this was in an early version where it was querying only for BTC.

PR #1061, opened 2020-10-30
frontend/accountSummary: show coin names instead of coin codes

benma:
Since grouping accounts by coin currently happens in the frontend, we
fetch the coin names in the same way as we fetch the coin totals from
the backend: with a map. Maybe some day the backend can prepare all
the data in one go, which would make that part a bit nicer.


x1ddos (approved):
LGTM

PR #1064, opened 2020-11-02
docker: unify docker image names

### LINKED ISSUE #1060: docker image tag
Trying to run the app without rebuilding the docker image I did:

```
docker pull shiftcrypto/bitbox-wallet-app
```

Which did not fetch the latest because "latest" doesn't point to 6. So then I did:

```
docker pull shiftcrypto/bitbox-wallet-app:6
```

But then `make dockerdev` didn't work it complained that it couldn't find `bitbox-wallet`. So I had to

```
docker tag shiftcrypto/bitbox-wallet-app:6 bitbox-wallet
```

I think it should be easier to get started than this. `make dockerinit` and `make dockerdev` could use the same name that you use on docker hub. :)

I think the above instructions are correct because I managed to run the app.

benma:
`make dockerdev` entered an image called `bitbox-wallet` made from
`make dockerinit`.

The official image on dockerhub is shiftcrypto/bitbox-wallet-app. Just
pulling that should be enough for someone to get started.

Fixes #1060.

### DIFF
--- a/scripts/dockerdev.sh
+++ b/scripts/dockerdev.sh
@@ -18,7 +18,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 dockerdev () {
     local container_name=bitbox-wallet-dev
 
-    if ! docker images | grep -q bitbox-wallet; then
+    if ! docker images | grep -q bitbox-wallet-app; then
         echo "No bitbox-wallet docker image found! Maybe you need to run 'make dockerinit'?" >&2
         exit 1
     fi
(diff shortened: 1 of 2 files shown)


scripts/dockerdev.sh
```diff
@@ -43,7 +43,7 @@ dockerdev () {
            --add-host="dev1.shiftcrypto.ch:176.9.28.155" \
            --add-host="dev2.shiftcrypto.ch:176.9.28.156" \
            -v $repo_path:/opt/go/src/github.com/digitalbitbox/bitbox-wallet-app \
-           bitbox-wallet bash
+           shiftcrypto/bitbox-wallet-app bash
```
x1ddos, line 46: There's also `echo "No bitbox-wallet docker image found! ..."` message above. Maybe adjust that as well? It greps for `bitbox-wallet`. I would change that too, to `bitbox-wallet-app`.

x1ddos (approved):
LGTM

scripts/dockerdev.sh
benma: :heavy_check_mark: done, good catch

PR #1065, opened 2020-11-04
backend/handlers: use starred fiat for chart

benma:
We will use the fiat marked as starred in the settings for the
charts. This is a quick solution to a fiat cycle button that will take
more time.


thisconnect (approved):
LGTM

PR #1067, opened 2020-11-04
backend/handlers/chart: start at earliest tx time

benma:


backend/handlers/handlers.go
```diff
@@ -874,9 +877,10 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
...
 		earliestTxTime := txs.EarliestTime()
-		// TODO: show chart starting at `earliestPriceavailable`, so the last three months can be
-		// shown more quickly instead of waiting for all historical rates first.
-		// For now, we show the chartDataMissing error.
+		if earliestTxTime.IsZero() {
+			// Ignore the chart for this account, there is no timed transaction.
+			continue
+		}
 		if earliestPriceAvailable.IsZero() ||
 			(!earliestTxTime.IsZero() && earliestTxTime.Before(earliestPriceAvailable)) {
```
x1ddos, line 885: Since you already check for `if earliestTxTime.IsZero()` above now, this can be simplified to:

```go
if earliestPriceAvailable.IsZero() || earliestTxTime.Before(earliestPriceAvailable)
```

easier to read.

backend/handlers/handlers.go
```diff
@@ -961,22 +958,28 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 		jsonTotals[c.Code()] = handlers.formatAmountAsJSON(coin.NewAmount(total), c, false)
 	}
 
-	truncateLeadingZeroes := func(s []chartEntry) []chartEntry {
-		for i, e := range s {
+	toSortedSlice := func(s map[int64]chartEntry) []chartEntry {
+		var result []chartEntry
+		for _, entry := range s {
+			result = append(result, entry)
```
x1ddos, line 964: I think this is one of those rare cases where initial slice allocation matters due to a potentially large enough dataset. I would do

```go
var result = make([]chartEntry, len(s))
result[i] = entry
```

x1ddos (approved):
LGTM with a couple suggestions.

backend/handlers/handlers.go
benma: Excellent! Done.

backend/handlers/handlers.go
benma: Done, PTAL.

x1ddos: :heavy_check_mark:

PR #1068, opened 2020-11-04
Chart styling

thisconnect:
- added chart view filter: week, month, year and all
- use hourly chart data for week view, daily for all others
- display hourly info when using hourly data set in weekly view
- calculate relative percentage change
- refactored accountssummary.tsx


frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -48,40 +144,215 @@ class Chart extends Component<Props, State> {
...
-                { data === undefined && (
-                      <div>{t('chart.dataMissing')}</div>
-                )}
+            <section className={styles.chart}>
+                <header>
+                    <div className={styles.summary}>
+                        {currentTotal ? (
+                            <div className={styles.totalValue}>
+                                {formatAsCurrency(currentTotal)}
+                                <span className={styles.totalUnit}>{getActiveFiat()}</span>
```
benma, line 298: you can now use the chartFiat string delivered from the backend

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -14,28 +14,124 @@
...
+                    },
+                },
+                layout: {
+                    backgroundColor: '#F5F5F5',
+                    fontSize: 11,
+                    fontFamily: '-apple-system, BlinkMacSystemFont, "Segoe UI", "Ubuntu", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif',
+                    textColor: '#1D1D1B',
+                },
+                leftPriceScale: {
+                    // mode: PriceScaleMode.Logarithmic, // TODO: future feature
```
benma, line 89: delete? can always add if needed, but this looks like a candidate for a TODO that might never be addressed ;)

benma (changes requested):
one commit per feature would have been nice. at least changing the table and changign the chart :innocent: 

a small UI bug: when i load the summary page, the axis labels on the left jump a few pixels.

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -113,19 +142,41 @@ class AccountsSummary extends Component<Props, State> {
...
+                                                    coinCode,
+                                                    title: data.coinNames[coinCode],
+                                                    ...account
+                                                }))
+                                            ))
+                                        ) : (
+                                            <p>{t('accountSummary.noAccount')}</p>
+                                        )}
+                                    </tbody>
+                                    {/* <tfoot>
```
benma, line 171: remove

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -48,40 +144,215 @@ class Chart extends Component<Props, State> {
...
+        }
+        this.setState(
+            { display: 'all', source: 'daily' },
+            () => {
+                if (!this.chart) {
+                    return;
+                }
+                const { to } = this.getUTCRange();
+                this.chart.timeScale().setVisibleRange({
+                    from: '2009-01-03',
```
benma, line 236: does `fitContent()` not work?

thisconnect: changed to `this.chart.timeScale().fitContent();`

frontends/web/src/components/rates/rates.tsx
```diff
@@ -73,6 +73,10 @@ export function setActiveFiat(fiat: Fiat): void {
     setConfig({ backend: { mainFiat: fiat } });
 }
 
+export function getActiveFiat(): Fiat {
+    return store.state.active;
+}
```
thisconnect, line 78: I don't need that anymore

thisconnect:
> a small UI bug: when i load the summary page, the axis labels on the left jump a few pixels.

It only happens when using ` priceFormat: { type: 'volume', },`, which shortens 11000 to 11k.
I'll try to hide the whole graph until subscribeVisibleLogicalRangeChange is called for the first time 🤞

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -48,40 +144,211 @@ class Chart extends Component<Props, State> {
...
+            );
+        }
+
+        const diff = difference && Number.isFinite(difference) ? (
+            <span className={styles[difference < 0 ? 'down' : 'up']} title={diffSince}>
+                <span className={styles.arrow}>
+                    {difference < 0 ? (<ArrowUp />) : (<ArrowDown />)}
+                </span>
+                <span className={styles.diffValue}>
+                    {formatAsCurrency(difference)}
```
benma, line 281: rename `formatAsCurrency` to `formatNumber()` or sth? `difference` is not a currency value.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -14,28 +14,124 @@
...
     private ref = createRef();
     private chart?: IChartApi;
+    private lineSeries?: ISeriesApi<'Area'>;
     private resizeTimerID?: any;
     private height: number = 300;
 
+    public readonly state: State = {
+        display: 'all',
+        source: 'daily',
+        currentTotal: this.props.dataDaily && this.props.dataDaily[this.props.dataDaily.length - 1].value,
```
benma, line 52: this should not be a state var, as it never changes. you can make a local function for example: `private getCurrentTotal()` to calculate it.

benma: also: use the hourly dataset instead. is better for currentTotal, as it goes to the current hour. the daily dataset is at utc midnight of the last day, so missing a few hours.

benma (changes requested):

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -48,40 +144,211 @@ class Chart extends Component<Props, State> {
...
+            // should be and therefore barsBefore might still point to the old range
+            // so we have to ignore this call and expect setVisibleRange with correct range
+            this.setState({ difference: 0, diffSince: '' });
+            return;
+        }
+        const valueFrom = data[rangeFrom].value;
+        const valueTo = this.state.currentTotal;
+        const valueDiff = valueTo ? valueTo - valueFrom : 0;
+        this.setState({
+            difference: ((valueDiff / valueFrom) * 100),
```
benma, line 258: this looks correct to me, nice!

benma:
Please fix the conflict and CI.

frontends/web/src/routes/account/summary/chart.tsx
thisconnect: I renamed the function at the import. Will be replaced when support different formatting

thisconnect:
PTAL

frontends/web/src/routes/account/summary/chart.tsx
benma: Why not rename it at the source? We could change how the function works and miss that it also formats non-currency numbers.

thisconnect:
@benma PTAL

benma (approved):
very nice!

PR #1069, opened 2020-11-05
frontend/app: show account summary in production, fix routing

benma:
Default route to account summary after unlock.

When there are no accounts, route to /, to not stay in the account summary.


benma:
@thisconnect one line stolen from #1068 ;)

thisconnect (approved):
LGTM

PR #1070, opened 2020-11-09
Chart total workaround

benma:


backend/handlers/handlers.go
```diff
@@ -829,8 +829,12 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 	fiat := handlers.backend.Config().AppConfig().Backend.MainFiat
 	// Chart data until this point in time.
 	until := handlers.backend.RatesUpdater().HistoryLatestTimestampAll(handlers.allCoinCodes(), fiat)
-	if until.IsZero() {
+	if until.IsZero() || time.Since(until) > 2*time.Hour {
```
x1ddos, line 832: Just an observation: as a user, I'd definitely prefer *some* data instead of none. But maybe it's just me.

backend/handlers/handlers.go
```diff
@@ -954,6 +954,33 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 
 		addChartData(account.Coin().Code(), timeseriesDaily, chartEntriesDaily)
 		addChartData(account.Coin().Code(), timeseriesHourly, chartEntriesHourly)
+
+		// HACK: We still use the latest prices from CryptoCompare for the account fiat balances
+		// above (displayed in the summary table). Those might deviate from the latest latest
```
x1ddos, line 959: "latest latest" two times

backend/rates/rates.go
```diff
@@ -122,6 +123,38 @@ func (updater *RateUpdater) Last() map[string]map[string]float64 {
 	return updater.last
 }
 
+// LastForPair returns the conversion rate for the given (coin, fiat) pair. Returns an error if the
+// rates have not been fetched yet.
+func (updater *RateUpdater) LastForPair(coinCode string, fiat string) (float64, error) {
+	// TODO: use coin.Code
+	last := updater.Last()
```
x1ddos, line 130: I've been trying to avoid this because there's a concurrent use issue: `last` may be read and modified at the same time. Was saving it for later to add a mutex.

backend/handlers/handlers.go
```diff
@@ -954,6 +954,33 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
...
+		// historical prices from coingecko, which results in different total balances in the chart
+		// and the summary table.
+		//
+		// As a temporary workaround, until we use only one source for all prices, we manually add
+		// one final datapoint reflecting the latest price. This can be removed once we stop using
+		// CryptoCompare.
+		now := time.Now().Unix()
+		price, err := handlers.backend.RatesUpdater().LastForPair(string(account.Coin().Code()), fiat)
```
x1ddos, line 967: `LastForPair` has no timestamp in returned value, but it doesn't imply the data is recent enough.
Even if we switched to CoinGecko, there's still no guarantee what `LastForPair` would've returned is recent enough. It could be a data point from 1 hour ago but this code here wouldn't notice it.

Just saying, historical chart doesn't need to or maybe even can't always go up to `now`, regardless of the external API provider, cryptocompare or coingecko.

backend/rates/rates.go
```diff
@@ -122,6 +123,38 @@ func (updater *RateUpdater) Last() map[string]map[string]float64 {
 	return updater.last
 }
 
+// LastForPair returns the conversion rate for the given (coin, fiat) pair. Returns an error if the
+// rates have not been fetched yet.
+func (updater *RateUpdater) LastForPair(coinCode string, fiat string) (float64, error) {
```
x1ddos, line 128: A bit incosistent: uses different coin code keys compared to the `Last` method.
Also, `coinCode, fiat string` is shorter.

backend/rates/rates.go
```diff
@@ -122,6 +123,38 @@ func (updater *RateUpdater) Last() map[string]map[string]float64 {
...
+	// TODO: use coin.Code
+	last := updater.Last()
+	if last == nil {
+		return 0, errp.New("rates not available yet")
+	}
+	key := map[string]string{
```
x1ddos, line 134: I would move this to the top, next to `coins` package var.

x1ddos (approved):
LGTM

backend/handlers/handlers.go
benma: Yeah me too, but if the data is too old, adding the `now`-datapoint is bad. Not sure how to solve this.

Maybe after we get rid of cryptocompare, we can revert this.

backend/handlers/handlers.go
benma: That is very true. With coingecko at least we have a timestamp.

backend/rates/rates.go
benma: I am just converting from coin.Code to CryptoCompare codes, explicitly.

Elsewhere it is done implicitly by using coin.Unit, etc., which I don't like.

backend/rates/rates.go
benma: Done.

backend/handlers/handlers.go
benma: Fixed

benma:
Note: my latest commit is still buggy, will revisit it shortly.

benma:
done, PTAL @thisconnect

PR #1072, opened 2020-11-12
frontend: fix browserlist

benma:
Our desktop app is currently based on 5.11.2:

https://doc.qt.io/archives/qt-5.11/qtwebengine-overview.html

> This version of Qt WebEngine is based on Chromium version
65.0.3325.151, with additional security fixes from newer versions.

The previous value made our JS compilation toolchain output code that
led to a syntax error in the desktop app.

### DIFF
--- a/frontends/web/package.json
+++ b/frontends/web/package.json
@@ -41,6 +41,6 @@
     "sortablejs": "^1.10.1"
   },
   "browserslist": [
-    "last 2 Chrome versions"
+    "chrome >= 65"
   ]
 }


thisconnect (approved):
LGTM

frontends/web/package.json
```diff
@@ -41,6 +41,6 @@
     "sortablejs": "^1.10.1"
   },
   "browserslist": [
-    "last 2 Chrome versions"
+    "chrome >= 65"
```
x1ddos, line 44: The last 2 versions of chrome today I believe are 85 and 86 so it makes sense that our poor qt webengine with features supported only at v65 possibly fails with some syntax errors for js code that targets v85 or v86.

x1ddos (approved):
Makes sense. I'll pass it to @thisconnect.

PR #1074, opened 2020-11-13
Frontend fix sidebar

thisconnect:
- Fix displaying `0` in the sidebar when there are no accounts
- Fix animating the sidebar on desktop (should only animate with the hamburger)


benma (approved):
utACK.

The sidebar animation commit: I can't really tell the difference, but I trust you :D

PR #1076, opened 2020-11-16
frontend/accountsummary: remove export

benma:
We keep the code so we can bring it back easily if needed, but disable
in production.

It was accidentally enabled when we enabled the whole page in
production.


thisconnect (approved):
LGTM

PR #1082, opened 2020-11-17
frontend/qt: move Makefile out of linux resources folder

### LINKED ISSUE #1081: packaging bug
I made an interesting discovery. A Makefile is installed in the root directory when installing the app.

```
niklas@ubuntu:~/Downloads$ dpkg -L bitbox | head
/.
/Makefile
/opt
/opt/bitbox
/opt/bitbox/bitbox.png
/opt/bitbox/libserver.so
/opt/bitbox/resources
/opt/bitbox/resources/qtwebengine_resources_200p.pak
/opt/bitbox/resources/qtwebengine_resources.pak
/opt/bitbox/resources/qtwebengine_resources_100p.pak
```

benma:
The whole resources folder is packaged and installed, so the Makefile
was installed at `/Makefile` before.

Installing the new (fixed) package will remove the /Makefile.

Fixes #1081

### DIFF
--- a/frontends/qt/resources/Makefile
+++ b/frontends/qt/resources/Makefile
@@ -1,0 +1,9 @@
+SIZES=16x16 32x32 48x48 64x64 128x128
+
+TARGETS=$(addsuffix /apps/bitbox.png, $(addprefix linux/usr/share/icons/hicolor/, ${SIZES}))
+
+all: $(TARGETS)
+
+linux/usr/share/icons/hicolor/%/apps/bitbox.png: app_icon_source.png
+	mkdir -p $(@D)
+	convert $< -resize '$*' $@
--- a/frontends/qt/resources/Makefile
+++ b/frontends/qt/resources/Makefile
@@ -1,6 +1,6 @@
 SIZES=16x16 32x32 48x48 64x64 128x128
 
-TARGETS=$(addsuffix /apps/bitbox.png, $(addprefix linux/usr/share/icons/hicolor/, ${SIZES}))
+TARGETS=$(addprefix linux/usr/share/icons/hicolor/, $(addsuffix /apps/bitbox.png, ${SIZES}))
 
 all: $(TARGETS)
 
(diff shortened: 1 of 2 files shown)


frontends/qt/resources/Makefile
```diff
@@ -0,0 +1,9 @@
+SIZES=16x16 32x32 48x48 64x64 128x128
+
+TARGETS=$(addsuffix /apps/bitbox.png, $(addprefix linux/usr/share/icons/hicolor/, ${SIZES}))
```
x1ddos, line 3: Could you swap `addsuffix` and `addprefix`? It just really hurts reading "addsuffix" first, then "addprefix" and swapping them around in my head. :)

frontends/qt/resources/Makefile
```diff
@@ -0,0 +1,9 @@
...
+
+all: $(TARGETS)
```
x1ddos, line 5: Then it feels like this `all` target should also include windows and macos? Besides, only linux uses a makefile. The others have a convert.sh script. Just saying, doesn't feel consistent. I'd recommend converting this to a shell script like the others or adding the other OSes to this makefile.

x1ddos (approved):
LGTM but not really. See the comments.

frontends/qt/resources/Makefile
benma: I am here to fix the Makefile in package issue, not to fix the Makefile itself :D

x1ddos: While there, might as well improve it, right? :)

benma: I am saying that mostly cause I have no clue how proper Makefiles look like and work, so I avoid working on them. Happy if you want to improve them though.

frontends/qt/resources/Makefile
benma: Done, PTAL

x1ddos: :+1: thanks!

x1ddos:
Hm.. macos build failed for some unknown reason. I restarted the jobs.

PR #1083, opened 2020-11-17
backend/eth: initial sync fix

benma:
An account should enforce Balance() can be safely called after
Initialize() and that the WaitSynchronized() call inside waits for it
to be actually synchronized. in BTC this is enforced as the first
counter increment happens in Initialize() (ensureAddresses()). In ETH
however, it did not. This fixes it, so that account.Balance() returns
the correct data after account.Initialize().

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -188,12 +188,13 @@ func (account *Account) Initialize() error {
 	)
 
 	account.coin.Initialize()
-	go account.poll()
+	done := account.Synchronizer.IncRequestsCounter()
+	go account.poll(done)
 
 	return account.BaseAccount.Initialize(accountIdentifier)
 }
 
-func (account *Account) poll() {
+func (account *Account) poll(initDone func()) {
 	timer := time.After(0)
 	for {
 		select {
@@ -212,6 +213,10 @@ func (account *Account) poll() {
 				account.SetOffline(true)
 			} else {
 				account.SetOffline(false)
+			}
+			if initDone != nil {
+				initDone()
+				initDone = nil
 			}
 			timer = time.After(pollInterval)
 		}
--- a/backend/accounts/account.go
+++ b/backend/accounts/account.go
@@ -66,7 +66,9 @@ type Interface interface {
 	FatalError() bool
 	Close()
 	Notifier() Notifier
+	// Must enforce that initial sync is done before returning.
 	Transactions() (OrderedTransactions, error)
+	// Must enforce that initial sync is done before returning.
 	Balance() (*Balance, error)
 	// SendTx signs and sends the active tx proposal, set by TxProposal. Errors if none
 	// available. The note, if set by ProposeTxNote(), is persisted for the transaction.


backend/coins/eth/account.go
```diff
@@ -213,6 +214,10 @@ func (account *Account) poll() {
 			} else {
 				account.SetOffline(false)
 			}
+			if decRequestCounter != nil {
+				decRequestCounter()
```
x1ddos, line 218: Not sure how everything works together here but it seems the account assume to be synchronized ater the first loop iteration unconditionally, even if `account.update()` was unsuccessful. Is that intended?

Also, I just happened to notice there seem to be 2 separate ways in which an account is assumed to be syncronized:

1. In Synchronizer.WaitSynchronized, which waits on a channel until it's closed and it's what's used here.

```go
// WaitSynchronized blocks until all pending synchronization tasks are finished.
func (synchronizer *Synchronizer) WaitSynchronized() {
	synchronizer.log.WithFields(logrus.Fields{"requestCounter": synchronizer.requestsCounter}).
		Debug("wait synchronized")
	if func() int32 {
		defer synchronizer.waitLock.RLock()()
		return synchronizer.requestsCounter
	}() == 0 {
		return
	}
	<-synchronizer.wait
}
```

2. BaseAccount sets `synced` field to `true` and emits an event:

```go
// NewBaseAccount creates a new Account instance.
func NewBaseAccount(config *AccountConfig, coin coin.Coin, log *logrus.Entry) *BaseAccount {
	account := &BaseAccount{
		config: config,
		coin:   coin,
	}
	account.Synchronizer = synchronizer.NewSynchronizer(
		func() { config.OnEvent(EventSyncStarted) },
		func() {
			if !account.synced {
				account.synced = true
				config.OnEvent(EventStatusChanged)
			}
			config.OnEvent(EventSyncDone)
		},
		log,
	)
	return account
}
```

Seems like they both serve the same purpose? If so, maybe they could be combined into one in the future.

x1ddos (commented):
LGTM

backend/coins/eth/account.go
benma: WaitSynchronized() waits until the current round of updating is finished. In BTC: when a addresses are re-synced cause a new tx touches them, in eth: in another poll() iteration. Balance() should wait until that is done to return to ensure a consistent state.

SycDone only refers to the "first" such sync. before that, there is no meaningful data at all.

benma: > Not sure how everything works together here but it seems the account assume to be synchronized ater the first loop iteration unconditionally, even if account.update() was unsuccessful. Is that intended?

Yeah. Currently, failure states are handled via the Offline() and FatalError() (the Offline() one we actually don't handle yet in the portfolio, but should).

backend/coins/eth/account.go
```diff
@@ -188,12 +188,13 @@ func (account *Account) Initialize() error {
 	)
 
 	account.coin.Initialize()
-	go account.poll()
+	decRequestCounter := account.Synchronizer.IncRequestsCounter()
```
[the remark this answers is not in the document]
benma: done :P

backend/coins/eth/account.go
x1ddos: I have to say it isn't clear at all to someone who doesn't know the code by heart. Still, not sure I understand:

> in eth: in another poll() iteration

so, that's 2 iterations in total? if so, why is `if decRequestCounter != nil { decRequestCounter() }` executed in the 1st iteration?

benma: > I have to say it isn't clear at all to someone who doesn't know the code by heart.

Agree, I'd like to improve all of this some day and make every state very clear.

I should have written "in each poll iteration()". Poll calls account.update() in each iteration, which increments the counter, does work, then decrements the counter. Everytime the counter hits 0, it means that data is in a consistent state, and the frontend gets the newest balance, transactions, and so on.

The issue before this PR was that before poll() is even called at all, the counter is also at 0, so Balance() would return 0. 

This commit fixes it by incrementing the counter in Initialize(), and decrements it after the first account update. Bitcoin works the same way: incremented in Initialize(), and decremented to 0 once all addresses and transactions are synced.

x1ddos: Yeah, no I get exactly why this change. I saw both Balance and Transactions being blocked by `WaitSynchronized` which I then traced to synchronizer's counter and a wait channel.

I guess I was complaining about me being confused :)

x1ddos:
:+1: I would ignore travis CI at the moment, for the sake of sanity.

benma:
@x1ddos pls approve

x1ddos (approved):
sorry :)

PR #1084, opened 2020-11-17
backend/handlers/accountSummary: fix missing data issue

benma:
EarliestTime(), like Timeseries(), has to return an error if data is
missing. Before, if a btc/ltc account had only confirmed transactions
with no timestamps (headers not synced up yet, which give the
timestamp), it would have just been ignored in the chart, instead of
showing the error that we are still gathering data.


thisconnect (approved):
tested & LGTM

PR #1086, opened 2020-11-18
Frontend: use custom chart tooltip

thisconnect:
Disabled the standard crosshair price and time labels and added
a custom tooltip, displaying price and date (and time on weekly).

Also enabled entireTextOnly for the priceScale so text should
not be cut off anymore.

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { createChart, IChartApi, LineData, LineStyle, ISeriesApi, UTCTimestamp } from 'lightweight-charts';
+import { createChart, IChartApi, LineData, LineStyle, ISeriesApi, UTCTimestamp, MouseEventHandler } from 'lightweight-charts';
 import { Component, createRef, h, RenderableProps } from 'preact';
 import { translate, TranslateProps } from '../../../decorators/translate';
 import { formatNumber, Fiat } from '../../../components/rates/rates';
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -295,9 +308,63 @@ class Chart extends Component<Props, State> {
...
+            toolTipVisible: true,
+            toolTipValue: price,
+            toolTipTop,
+            toolTipLeft,
+            toolTipTime: time,
+        });
+    }
+
+    private renderDate = (date) => {
+        date = new Date(date);
```
benma, line 344: check if you can use toLocaleString() to format in the same way as the date formatting in the transactions list:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/cc77c1381617a2abb83c4897fa8c0997bc0b6c59/frontends/web/src/components/transactions/transaction.tsx#L81

benma (approved):
Works on desktop. On mobile, mouse movement does not exist. If you want you can look into registering the click handler for that. You can simulate it on desktop by disabling the mousemovement events and using only click events.

could be a bit tricky, as the blue dot indicating which point is selected does not appear on mobile (maybe it can be done via click too somehow). 

i tried just changed `subscribeCrosshairMove` to `subscribeClick`. the tooltip pops up but always with the data of the latest data point instead of where i clicked.

frontends/web/src/routes/account/summary/chart.tsx
thisconnect: fixed but forced to always display 2 digits

benma: LGTM. you are not showing the hour any more in weekly, in case you dropped it by accident. I am okay with either. Feel free to merge.

thisconnect: yes I forgot but @jadzeidan didn't like it anyways… waiting for his input.

PR #1087, opened 2020-11-18
backend/accounts/transaction: fix typo

benma:


x1ddos (approved):
LGTM. I wonder how come golangci didn't catch this. I remeber we run some typo linter. Maybe "timestap" means something :)

benma:
> LGTM. I wonder how come golangci didn't catch this. I remeber we run some typo linter. Maybe "timestap" means something :)

It did, that's how I caught this. Just not in my own PR, but in some other PR after it was merged. I guess I missed it when merging.

PR #1088, opened 2020-11-18
Frontend: fix summary percentage diff

thisconnect:
The summary differenece in percentage was sometimes off as
it did not update at all. Happend with all filter or with
shorter datasets, i.e. first tx less than a year old.

Fix is to simple remove the falsy visiblerange.barsBefore
check as it can be 0.

Also imported lightweight-chart types for LogicalRange and
BarsInfo and reformated the percentage debug tooltip.


benma (approved):
utACK

PR #1090, opened 2020-11-19
Frontend portfolio loading

thisconnect:
- should flicker less during loading, due to fixed widths
- use animated text spinner

tested with:
- just `make servewallet`
- `rm -rf appfolder.dev && cp -r appfolder.dev.brokentxdate appfolder.dev && make servewallet` Note: required new pairing, best unplug BitBox02 first
- `rm -rf appfolder.dev/cache/exchangerates && make servewallet`

### DIFF
--- a/frontends/web/src/components/spinner/ascii.tsx
+++ b/frontends/web/src/components/spinner/ascii.tsx
@@ -1,0 +1,26 @@
+/**
+ * Copyright 2020 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { h } from 'preact';
+import * as style from './ascii.css';
+
+export default function AsciiSpinner() {
+    return (
+        <div class={style.spinnerContainer}>
+            <div class={style.spinner}></div>
+        </div>
+    );
+}
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -26,6 +26,7 @@ import { Guide } from '../../../components/guide/guide';
 import { Fiat, FiatConversion } from '../../../components/rates/rates';
 import { TranslateProps } from '../../../decorators/translate';
 import Logo from '../../../components/icon/logo';
+import Spinner from '../../../components/spinner/ascii';
 import { debug } from '../../../utils/env';
 import { apiGet, apiPost } from '../../../utils/request';
 import { apiWebsocket } from '../../../utils/websocket';
@@ -66,10 +67,6 @@ interface Response {
 }
 
 type Props = TranslateProps & AccountSummaryProps;
-
-export type GroupedByCoinCode = {
-    [key in CoinCode]?: AccountAndBalanceInterface[];
-}
 
 interface BalanceRowProps {
     code: string;
@@ -176,7 +173,7 @@ class AccountsSummary extends Component<Props, State> {
                 <tr key={code}>
                     { nameCol }
                     <td data-label={t('accountSummary.balance')}>
-                        <span>
+                        <span className={style.summaryTableBalance}>
                             {balance.available.amount}{' '}
                             <span className={style.coinUnit}>{coinUnit}</span>
                         </span>
@@ -191,7 +188,10 @@ class AccountsSummary extends Component<Props, State> {
         return (
             <tr key={code}>
                 { nameCol }
-                <td colSpan={2}>spinning... { syncStatus }</td>
+                <td colSpan={2}>
+                    { syncStatus }
+                    <Spinner />
+                </td>
             </tr>
         );
     }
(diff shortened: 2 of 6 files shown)


frontends/web/src/components/spinner/ascii.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { h } from 'preact';
+import * as style from './ascii.css';
+
+export default function AsciSpinner() {
```
benma, line 20: Ascii?

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -205,35 +205,43 @@ class AccountsSummary extends Component<Props, State> {
...
-                                isUpToDate={data.chartIsUpToDate} />}
+                            {data ? (
+                                <Chart
+                                    dataDaily={data.chartDataMissing ? undefined : data.chartDataDaily}
+                                    dataHourly={data.chartDataMissing ? undefined : data.chartDataHourly}
+                                    fiatUnit={data.chartFiat}
+                                    total={data.chartTotal}
+                                    isUpToDate={!data.chartDataMissing || data.chartIsUpToDate} />
+                            ) : (
+                                <p style="text-align: center;">loading chart…</p>
```
benma, line 236: Could you maybe just reword the 'gathering information' text and show that here instead? so we don't add too many i18n entries.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -161,7 +163,9 @@ class Chart extends Component<Props, State> {
                     this.chart.subscribeCrosshairMove(this.handleCrosshair as MouseEventHandler);
                 }
             }
-            this.chart.timeScale().fitContent();
+            if (this.chart) {
```
benma, line 166: isn't this always true?

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -80,3 +92,10 @@ export default class HeadersSync extends Component {
         );
     }
 }
+
+const subscribeHOC = subscribe<SubscribedHeadersSyncProps, HeadersSyncProps & TranslateProps>(({ coinCode }) => ({
+    status: `coins/${coinCode}/headers/status`,
+}), false, true)(HeadersSync);
```
benma, line 98: why `..., false, true`? Before it seemed to be the opposite.

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -1,5 +1,6 @@
+
```
benma, line 1: This change seems completely unrelated to the chart, so it could have been a separate PR :)

benma (approved):

benma:
When rebasing, use `formatCurrency(total, fiatUnit)` in the conflicting line.

frontends/web/src/components/headerssync/headerssync.tsx
thisconnect: Refactored, because in the version where the headerssync was displayed in the sumamrytable.

What is left is  we share the same spinner. Can we keep it in the same PR but different commit?

frontends/web/src/routes/account/summary/accountssummary.tsx
thisconnect: done

frontends/web/src/routes/account/summary/chart.tsx
thisconnect: done

frontends/web/src/components/headerssync/headerssync.tsx
thisconnect: will open a separate PR after this is merged

frontends/web/src/components/headerssync/headerssync.tsx
thisconnect: will open a separate PR

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -205,35 +205,45 @@ class AccountsSummary extends Component<Props, State> {
...
                     </Header>
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
-                            {data && <Chart
-                                dataDaily={data.chartDataMissing ? undefined : data.chartDataDaily}
-                                dataHourly={data.chartDataMissing ? undefined : data.chartDataHourly}
-                                fiatUnit={data.chartFiat}
-                                total={data.chartTotal}
-                                isUpToDate={data.chartIsUpToDate} />}
+                            {(data && data.chartDataDaily && data.chartDataDaily.length) ? (
```
benma, line 228: i think this is not correct. there could be legitimately zero data points and the chart is still fully loaded (e.g. when there are no transactions).

before was correct.

benma (changes requested):

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -205,35 +205,45 @@ class AccountsSummary extends Component<Props, State> {
...
+                                        <line x1="12" y1="15" x2="12" y2="3"></line>
+                                    </svg>
+                                </a>
+                            )
+                        )}
                     </Header>
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
-                            {data && <Chart
-                                dataDaily={data.chartDataMissing ? undefined : data.chartDataDaily}
```
benma, line 230: why is this gone? the two fields are only valid (defined) if chartDataMissing is false, that's why it was like this. the backend might deliver garbage in this case

frontends/web/src/components/headerssync/headerssync.tsx
benma: why?

benma (approved):

frontends/web/src/routes/account/summary/accountssummary.tsx
thisconnect: fixed

PR #1091, opened 2020-11-19
ci: do not install yarn from homebrew

thisconnect:
Because yarn on homebrow has a newer nodejs dependency,
also updated nodejs to 12.x

it still fails but now consistently with github actions


thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1139/files

PR #1092, opened 2020-11-21
Frontend: initialize chart only when there is data to show

thisconnect:
When creating a chart without any data, the chart displays
nothing, but creates a big empty space.

This checks if there is chart data and only creates the chart
and displays the filters with data."

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -91,8 +91,12 @@ class Chart extends Component<Props, State> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


benma:
I think the 2nd commit might not be necessary after the fix in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1095. Can you try that PR and see if you can get rid of the 2nd commit?

Also please make multiple PRs if the commits in one PR are not related to the same thing, then it is much easier to look at them separately, and merge one if the other does not work is takes longer.

benma (approved):

PR #1093, opened 2020-11-22
backend/rates: do not let request context expire when rate-limited

x1ddos:
When blocked by a rate limit during a network round trip, the request context keeps running. This causes such requests to abort before they even get a chance to fire due to the expired context.

What we want is to start a request context right before making an API call. This commit does exactly that by moving the rate limiting function at a "layer" above the HTTP round trip.

The commit also changes API requests timeout from 1min to 10sec - more than enough for CoinGecko to respond. If it doesn't, we better retry than wait another 50sec.

Alternatives which didn't work:

1. Make a new context in RateLimitedHTTPTransport once unblocked. This loses the cancel function, needed in ReconfigureHistory. It is still possible to make this work by select'ing on both old and new contexts but the code turns out too complex and ugly.

2. Give RateUpdater a separate CoinGecko client type instead of an HTTP client. We make only one API call, maybe two at the most in the future. This doesn't warrant a separate type at the moment. RateUpdater is effectively that client.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -197,8 +197,7 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	if err := os.MkdirAll(ratesCache, 0700); err != nil {
 		log.Errorf("RateUpdater DB cache dir: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


benma (approved):
utACK

PR #1094, opened 2020-11-22
backend/rates: better backfill stop condition and daily/hourly logic

x1ddos:
There are many scenarios where the upstream API may return non-zero
results and yet incomplete data. Picking just one, namely "... unless
the start/end range is suspiciously arbitrary close to current time"
improves nothing. So, there's no point in making the logic more
complicated.

This commit also computes correctly when to request hourly vs daily
rates from the upstream in the case where the difference between
the oldest point and current time is less than 90 days.

### DIFF
--- a/backend/rates/history.go
+++ b/backend/rates/history.go
@@ -167,35 +167,34 @@ func (updater *RateUpdater) backfillHistory(ctx context.Context, coin, fiat stri
 		// When to update next, after this loop iteration is done.
 		untilNext := time.Duration(1+rand.Intn(5)) * time.Second
 
+		// We want hourly rates for the last 90 days and daily past that.
+		// 90 days is the max interval CoinGecko responds with hourly timeseries to.
 		end := updater.HistoryEarliestTimestamp(coin, fiat)
 		var start time.Time
-		if end.IsZero() {
-			// First time; don't have historical data yet.
+		switch {
+		// First time; don't have historical data yet.
+		case end.IsZero():
 			end = time.Now()
-			// 90 days is the max interval which CoinGecko responds to with hourly timeseries.
-			// Make sure we get hourly rates by requesting 90 days minus 1 hour, just in case.
-			start = end.Add(-90*24*time.Hour + time.Hour)
-		} else {
-			// End at the day earlier than the last known timestamp.
-			// We want daily rates for timestamps past the first 90 days: use annual intervals.
+			start = end.Add(-90*24*time.Hour + time.Hour) // +1h to be sure
+		// Less than 90 days: still want hourly rates.
+		case time.Since(end) < 90*24*time.Hour:
+			end = end.Add(-time.Hour)
+			start = time.Now().Add(-90 * 24 * time.Hour)
+		// We want daily rates for timestamps past the first 90 days.
+		// Use multiple of years interval but not "too much" to control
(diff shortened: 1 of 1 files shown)


backend/rates/history.go
```diff
@@ -169,33 +169,31 @@ func (updater *RateUpdater) backfillHistory(ctx context.Context, coin, fiat stri
...
 			end = time.Now()
-			// 90 days is the max interval which CoinGecko responds to with hourly timeseries.
-			// Make sure we get hourly rates by requesting 90 days minus 1 hour, just in case.
 			start = end.Add(-90*24*time.Hour + time.Hour)
-		} else {
-			// End at the day earlier than the last known timestamp.
-			// We want daily rates for timestamps past the first 90 days: use annual intervals.
+		// 90 days is the max interval which CoinGecko responds to with hourly timeseries.
+		// Make sure we get hourly rates by requesting 90 days minus 1 hour, just in case.
+		case time.Since(end) < 90*24*time.Hour:
```
benma, line 179: i don't understand this condition. why is it relevant how much time is between `end` and now? we are backfilling the history, not forward filling it.

x1ddos: We want hourly rates for the past 90 days, i.e. between now and (now - 90days). Suppose `end` happened to be somewhere between `now` and `90days ago`. Without this condition, it would go to the default below which requests a much larger range for daily rates but we want hourly for (now-end).

In other words:

```
(years-ago) --- daily rates --- (now-90days) --- hourly rates --- (now)
```

if `end` happens to be between `(now-90days)` and `now`, we still want the hourly rates, specifically for the range of (`end` - `now-90days`).

benma: I see, but why is `start = end.Add(-time.Since(end))`? If `end` is e.g. `now - 1month`, then start would be `now - 2months`? Shouldn't start be `now - 3 months` or possibly `end - 3 months`?

x1ddos: oh, you're totally right and I'm a goofy :)
done.

x1ddos (commented):
PTAL

benma (approved):

backend/rates/history.go
```diff
@@ -167,35 +167,35 @@ func (updater *RateUpdater) backfillHistory(ctx context.Context, coin, fiat stri
...
-			// 90 days is the max interval which CoinGecko responds to with hourly timeseries.
-			// Make sure we get hourly rates by requesting 90 days minus 1 hour, just in case.
-			start = end.Add(-90*24*time.Hour + time.Hour)
-		} else {
-			// End at the day earlier than the last known timestamp.
-			// We want daily rates for timestamps past the first 90 days: use annual intervals.
+			start = end.Add(-90*24*time.Hour + time.Hour) // 1 hour diff just in case
+		// Less than 90 days: still want hourly rates.
+		case time.Since(end) < 90*24*time.Hour:
+			till90days := 90*24*time.Hour - time.Since(end)
```
benma, line 181: isn't `start = end.Add(-90*24*time.Hour)` easier? :D

x1ddos: :+1: although + extra 1h like in the zero case.

x1ddos: actually, no need for +1h. it's always less than 90days here.

PR #1095, opened 2020-11-22
portfolio: add the last missing point to the chart

benma:


backend/handlers/handlers.go
```diff
@@ -971,7 +971,8 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 				return result[i:]
```
x1ddos, line 985: I'm a bit allergic to comparing float values with 0.
not directly related to this commit but anyway, `if e.Value > 0 { return result[i:] }` is better imho.

x1ddos (approved):
LGTM

backend/handlers/handlers.go
benma: done

PR #1096, opened 2020-11-23
rates: add SGD

thisconnect:


backend/rates/history.go
```diff
@@ -63,6 +63,7 @@ var (
 		"AUD": "aud",
 		"ILS": "ils",
 		"BTC": "btc",
+		"SDG": "sdg",
```
thisconnect, line 66: oups, is it sgd...
😬

benma (approved):
LGTM

PR #1100, opened 2020-11-24
backend/rates: evaluate time.Now at the coingecko API call time

x1ddos:
Before this commit, forward history update loop has never had a chance
to catch up. It always calls an update with (start, now) but by the
time it gets to fetch the data, "now" is in the past.

This gets worse as the number of active coin/fiat pairs increases.
The bigger the number, the greater the lag behind "now" is due to rate
limiting.

The change resolves the issue by evaluating "now" at the time
RateUpdater requests data from the upstream and always returns the
latest available exchange rates.

### DIFF
--- a/backend/rates/history_test.go
+++ b/backend/rates/history_test.go
@@ -87,8 +87,8 @@ func TestUpdateHistory(t *testing.T) {
 		},
 	}
 
-	start, end := time.Unix(wantStartUnix, 0), time.Unix(wantEndUnix, 0)
-	n, err := updater.updateHistory(context.Background(), "btc", "USD", start, end)
+	g := fixedTimeRange(time.Unix(wantStartUnix, 0), time.Unix(wantEndUnix, 0))
+	n, err := updater.updateHistory(context.Background(), "btc", "USD", g)
 	require.NoError(t, err, "updater.updateHistory err")
 	assert.Equal(t, 2, n, "updater.updateHistory n")
 	wantHistory := map[string][]exchangeRate{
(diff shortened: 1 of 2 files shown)


backend/rates/history_test.go
```diff
@@ -87,8 +87,8 @@ func TestUpdateHistory(t *testing.T) {
 		},
 	}
 
-	start, end := time.Unix(wantStartUnix, 0), time.Unix(wantEndUnix, 0)
-	n, err := updater.updateHistory(context.Background(), "btc", "USD", start, end)
+	g := fixedTimeRange(time.Unix(wantStartUnix, 0), time.Unix(wantEndUnix, 0))
```
benma, line 90: g?? :joy:

benma (approved):
nice

backend/rates/history_test.go
x1ddos: well, it's "time ran**g**e" and `t` is already taken by the testing.T :)
but really, it's in a test and used only on the very next line. doesn't deserve a longer name imho.

benma: lol.

still i have very much disagree. also in tests you write it once and read it many times, a long name takes no effort compared to the reward.

alternative is also to just inline the whole thing and make the function call a multi-liner.

x1ddos: both a long name used only on the very next line and a multi line expression are harder to read imho.

benma: for me it's the opposite, i always do a double take on a single letter varname, panicking what it means, and having a hard time searching for it to see where it is used when I feel the need to search for it.

x1ddos:
Here's a counter argument:
https://research.swtch.com/names

 > hard time searching for it to see where it is used

no need to search! you just saw it 200-400ms ago on the previous line.

I think you're taking it as a all-or-nothing approach but it's not black 
and white.

PR #1101, opened 2020-11-24
Syncstatus

benma:


frontends/web/src/locales/en/app.json
```diff
@@ -1,6 +1,5 @@
 {
   "account": {
-    "disconnect": "Connection lost. Retrying…",
```
benma, line 3: basically same as `reconnecting` on L17

thisconnect (approved):
LGTM 👍

PR #1114, opened 2020-11-27
frontend: use pulic sans font

thisconnect:
System fonts look only as good as your OS. This change
should improve the font on windows, but only subtle
difference on macOS.


thisconnect:
cc @jadzeidan please test

benma:
Lots of font files added. Do we need all of them?

My1:
is the windows font that bad?

thisconnect:
> is the windows font that bad?

The initial idea was that users have the system font, but we have issues with that and sometimes it really looks ugly (Windows and Linux).

Are you on Windows, does the font rendering look acceptable to you?

This is low prio at the moment, but feedback is welcome.

Also I'd like to only include a free open font, so we'll be limited and wont be able to use any font.

x1ddos:
Why closed?

thisconnect:
spring clean, I'd rather have we focus on relevant open PR's so closed this one, this was more a POC but can always re-open. Also the font was just the first best open font I found. Also no feedback from @jadzeidan  https://github.com/digitalbitbox/bitbox-wallet-app/pull/1114/files#issuecomment-734811914

PR #1115, opened 2020-11-30
backend/bitbox02: update to v9.3.1

benma:


x1ddos (approved):
LGTM but I always wondered, why commits like https://github.com/digitalbitbox/bitbox-wallet-app/pull/1104/commits/70f609c5cdc076670296f47f5d6813dd4b1b59e1 are separate. Each app release always bundles a certain firmware version. Might as well make them a one "atomic" commit?

benma:
> Each app release always bundles a certain firmware version.

That's not true, there have been and can be many more app releases without a firmware upgrade.

Even if, why one commit if you can have two commits? Then you can revert one if there was an issue in one but not the other, etc.

x1ddos:
> > Each app release always bundles a certain firmware version.
> 
> That's not true, there have been and can be many more app releases without a firmware upgrade.
> 
> Even if, why one commit if you can have two commits? Then you can revert one if there was an issue in one but not the other, etc.

Let's say you bump the version of the app to 4.25.0 just like the previous https://github.com/digitalbitbox/bitbox-wallet-app/commit/70f609c. Now you revert the bump. What you end up is 2 apps with 4.24.0 where each one bundles different firmware version.

If you revert this commit, what you end up is 2 apps v4.24.0 each one bundling again different firmware versions. In fact, once this is merged, it is effectively that state.

My point is, you'll probably never want to revert commits like this. You'll want to roll forward. If all this were a single commit and say there were an issue with bundling firmware v9.3.1 into app v4.25.0, you'd revert the bump with `git revert -n <hash>`, replace the bundled assets and increase app version to v.4.26.0 or v.4.25.1. Make a new commit. You end up with a consistent versioning state.

benma:
The version has many states, each commit changes the app but the version is the same. The particular firmware version bundled is no different than any other commit. What matters is only the version tag at the end.

x1ddos:
You're missing my point. It's easier to manage versioning with a single commit. 

You'll never want to leave a reverted bundled firmware version without eventually bumping the app version as well, before tagging a new release. So, you'll always end up with bumping both. Might as well do it all in a single commit.

x1ddos:
The picture would be very different if we replaced bundled assets with every firmware dev change as it comes, i.e. bundled unsigned firmware, and then eventually replaced it with a signed version just before the release. Then I would agree 2 commits are definitely necessary.

But that's not what happens in practice. We work on dev firmware and the app separately, replacing the signed bundle only at release time. There's no advantage in separate commits in this case.

benma:
> So, you'll always end up with bumping both. Might as well do it all in a single commit.

This coupling is not true though. There can and will be app releases that come without new firmware releases. It may also happen that we bump the app version, add a new firmware, then decide to not ship the new firmware after all due to a regression. For example, if 9.3.0 and 9.3.1 had a regression we could simply revert those commits and release without them, keeping the app version bump commit.

x1ddos:
> This coupling is not true though. There can and will be app releases that come without new firmware releases.

Of course but it doesn't contradict a single commit method. In this case the single commit contains zero bundled assets changes.

> For example, if 9.3.0 and 9.3.1 had a regression we could simply revert those commits and release without them

I think you meant 9.3.0-rc1 and 9.3.1-rc1. If we did release 9.3.0 with a firmware we wanted to revert in the end, we'd have to bump the firmware version for 9.3.1 release anyway because there's chance someone has already installed the previous release and upgraded the firmware. We don't support signed firmware downgrades on prod devices, do we?

So, simply reverting a bundled firmware commit wouldn't work. Instead, we'll have to make a new firmware version release with a binary possibly identical to the previous release except it'll have a version increment.

PR #1117, opened 2020-12-01
frontend/accountsummary: update data on syncdone event

benma:
This is the main event to listen to and is fired when an account is
updated.

### DIFF
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -95,7 +95,7 @@ class AccountsSummary extends Component<Props, State> {
     }
 
     public componentWillUnmount() {
-        window.clearInterval(this.summaryReqTimerID);
+        window.clearTimeout(this.summaryReqTimerID);
         this.unsubscribe();
     }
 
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -124,7 +124,11 @@ class AccountsSummary extends Component<Props, State> {
         if (data.type === 'account') {
             switch (data.data) {
                 case 'statusChanged':
+                case 'syncdone':
                     this.onStatusChanged(data.code);
+                    // Force getting account summary now; cancel nex scheduled call.
```
x1ddos, line 129: "next" typo

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -95,7 +95,7 @@ class AccountsSummary extends Component<Props, State> {
     }
 
     public componentWillUnmount() {
-        window.clearInterval(this.summaryReqTimerID);
```
x1ddos, line 98: Good change.

> It's worth noting that the pool of IDs used by setTimeout() and setInterval() are shared, which means you can technically use clearTimeout() and clearInterval() interchangeably. However, for clarity, you should avoid doing so.

(from https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout)

Although "pool of IDs used by setTimeout() and setInterval() are shared" may depend on the browser, I believe.

x1ddos (approved):
LGTM but typo in "update data on syncodne event" commit title

benma:
Omg my typo frequence is increasing rapidly :sob: 

Thanks. Fixed both.

PR #1118, opened 2020-12-01
backend/handlers: chart data not missing if there are 0 transactions

benma:
If the user gets started in the app for the first time and there are
no transactions, there is no need to wait for the headers and exchange
rates to sync. This avoids the 'gathering info' message as the first
impression.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -820,6 +820,8 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 
 	currentTotal := new(big.Rat)
 	currentTotalMissing := false
+	// Total number of transactions across all active accounts.
+	totalNumberOfTransactions := 0
 	for _, account := range handlers.backend.Accounts() {
 		if account.FatalError() {
 			continue
@@ -836,6 +838,7 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 		if err != nil {
 			return nil, err
 		}
+		totalNumberOfTransactions += len(txs)
 
 		// e.g. 1e8 for Bitcoin/Litecoin, 1e18 for Ethereum, etc. Used to convert from the smallest
 		// unit to the standard unit (BTC, LTC; ETH, etc.).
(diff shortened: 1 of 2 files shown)


backend/handlers/handlers.go
```diff
@@ -989,6 +992,14 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 		return []chartEntry{}
 	}
 
+	// Even if we are still gathering data (exchange rates, block headers), we know we result
+	// already if there are no transactions. This avoids showing the user a message that we are
+	// gathering data, only to show nothing in the end.
+	if chartDataMissing && totalNumberOfTransactions == 0 {
```
x1ddos, line 998: Or simply `if totalNumberOfTransactions == 0`?

backend/handlers/handlers.go
```diff
@@ -989,6 +992,14 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 		return []chartEntry{}
 	}
 
+	// Even if we are still gathering data (exchange rates, block headers), we know we result
```
x1ddos, line 995: "we know **the** result" typo?

x1ddos (approved):
LGTM

x1ddos:
> + golangci-lint run
> backend/handlers/handlers.go:797:1: cyclomatic complexity 27 of func `(*Handlers).getAccountSummary` is high (> 25) (gocyclo)
> func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error) {

I have to say I agree with the linter here. That method grew very big.

backend/handlers/handlers.go
benma: Then the log would not be as useful.

backend/handlers/handlers.go
benma: omg omg omg omg fixed

benma:
> I have to say I agree with the linter here. That method grew very big.

Me too actually. I am inclined to postpone refactoring, but let me see if I can get some quick wins.

x1ddos:
> > I have to say I agree with the linter here. That method grew very big.
> 
> Me too actually. I am inclined to postpone refactoring, but let me see if I can get some quick wins.

Yeah, I meant in the future.

backend/handlers/handlers.go
x1ddos: A test would've been even more useful than a log message. Then you wouldn't need to log :)

benma: Mutually exclusive ;)

Unit tests I hope I'll get around to. https://github.com/digitalbitbox/bitbox-wallet-app/pull/1119 will help

PR #1119, opened 2020-12-01
Chart refactor

benma:
This is still not enough to pass the complexity linter :see_no_evil: But it the function is shorter and the code a bit more organized so I opened this PR.

### DIFF
--- a/backend/chart.go
+++ b/backend/chart.go
@@ -1,0 +1,259 @@
+// Copyright 2020 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backend
+
+import (
+	"math/big"
+	"sort"
+	"time"
+
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/errors"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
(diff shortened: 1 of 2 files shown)


backend/chart.go
```diff
@@ -36,20 +36,32 @@ func (backend *Backend) allCoinCodes() []string {
...
+}
 
+// Chart has all data needed to show a time-based chart of their assets to the user.
+type Chart struct {
+	DataMissing bool         `json:"chartDataMissing"`
+	DataDaily   []ChartEntry `json:"chartDataDaily"`
+	DataHourly  []ChartEntry `json:"chartDataHourly"`
+	Fiat        string       `json:"chartFiat"`
+	Total       *float64     `json:"chartTotal"`
+	// only valid is DataMissing is false
```
x1ddos, line 52: "valid **if** ..."

backend/chart.go
```diff
@@ -36,20 +36,32 @@ func (backend *Backend) allCoinCodes() []string {
...
+	Value float64 `json:"value"`
+}
 
+// Chart has all data needed to show a time-based chart of their assets to the user.
+type Chart struct {
+	DataMissing bool         `json:"chartDataMissing"`
+	DataDaily   []ChartEntry `json:"chartDataDaily"`
+	DataHourly  []ChartEntry `json:"chartDataHourly"`
+	Fiat        string       `json:"chartFiat"`
+	Total       *float64     `json:"chartTotal"`
```
x1ddos, line 51: might as well remove the pointer here? there's `DataMissing` for that. or is it also valid when data is missing?

backend/chart.go
```diff
@@ -188,30 +216,8 @@ func (backend *Backend) ChartData() (*Chart, error) {
 			return nil, err
 		}
 
-		addChartData := func(coinCode coin.Code, timeseries []accounts.TimeseriesEntry, chartEntries map[int64]ChartEntry) {
```
x1ddos: Nice. I would also move `toSortedSlice` out. It really isn't a big deal - you see somehing "sorted", you'll know immediately what it does without even looking at its code. But moving `toSortedSlice` out makes it easier to read the rest of the important logic in this method.

x1ddos (approved):
LGTM

backend/chart.go
benma: :see_no_evil:

backend/chart.go
benma: yeah, it is independent of DataMissing, but can also be missing :joy:  I will add comments to each field.

PR #1120, opened 2020-12-01
frontend: make locize-pull

benma:


benma:
> LGTM but translations are old :(

Yeah but they are marked for review. cc @jstrnbrg

PR #1121, opened 2020-12-03
backend/accounts: handle transactions with negative height

benma:
Electrum servers return -1 for the tx height if the tx is unconfirmed
with an unconfirmed parent. The code in the Timeseries() incorrectly
assumed that unconfirmed means a height of 0, leading to an
false `ErrNotAvailable` error if such a transaction is present.


x1ddos (approved):
LGTM

PR #1122, opened 2020-12-03
backend: make changes to accounts atomic

benma:
We had an accountsLock, but it not held for the whole duration of
initAccounts(), which means that in concurrent calls, the accounts
could be uninitialized twice, then all accounts would be added twice,
or other such race conditions.

This commit only acquires lock in exported functions to make it work
with all combinations of account-changing function calls.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -208,9 +208,11 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 
 // configureHistoryExchangeRates changes backend.ratesUpdater settings.
 // It requires both backend.config to be up-to-date and all accounts initialized.
+//
+// The accountsLock must be held when calling this function.
 func (backend *Backend) configureHistoryExchangeRates() {
 	var coins []string
-	for _, acct := range backend.Accounts() {
+	for _, acct := range backend.accounts {
 		coins = append(coins, string(acct.Coin().Code()))
 	}
 	// No reason continue with ERC20 tokens if Ethereum is inactive.
@@ -227,8 +229,8 @@ func (backend *Backend) configureHistoryExchangeRates() {
 }
 
 // addAccount adds the given account to the backend.
+// The accountsLock must be held when calling this function.
 func (backend *Backend) addAccount(account accounts.Interface) {
-	defer backend.accountsLock.Lock()()
 	backend.accounts = append(backend.accounts, account)
 	account.Observe(backend.Notify)
 	backend.onAccountInit(account)
@@ -264,9 +266,8 @@ func (backend *Backend) emitAccountsStatusChanged() {
 	})
 }
 
-// CreateAndAddAccount creates an account with the given parameters and adds it to the backend. If
-// persist is true, the configuration is fetched and saved in the accounts configuration.
-func (backend *Backend) CreateAndAddAccount(
+// The accountsLock must be held when calling this function.
+func (backend *Backend) createAndAddAccount(
 	coin coin.Coin,
 	code string,
 	name string,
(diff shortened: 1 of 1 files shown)


backend/backend.go
```diff
@@ -420,7 +432,7 @@ func (backend *Backend) createAndAddBTCAccount(
 			case signing.ScriptTypeP2WPKH:
 				suffixedName += ": bech32"
 			}
-			err := backend.CreateAndAddAccount(
+			err := backend.createAndAddAccount(
```
x1ddos, line 435: Please add it to `createAndAddBTCAccount` too that it requires the accountsLock.

backend/backend.go
```diff
@@ -1155,7 +1181,10 @@ func (backend *Backend) Close() error {
 	errors := []string{}
 
 	backend.ratesUpdater.Stop()
+
+	unlock := backend.accountsLock.Lock()
```
x1ddos, line 1185: I would just do it at the top with defer'ed unlock. After all, it says: "After this, no other method should be called."

Might be irrelevant but I noticed this manipulates `backend.coins`. I think it should also locked for the whole duration and possibly `delete(backend.coins, coin)`?

backend/backend.go
```diff
@@ -228,7 +228,6 @@ func (backend *Backend) configureHistoryExchangeRates() {
 
 // addAccount adds the given account to the backend.
 func (backend *Backend) addAccount(account accounts.Interface) {
```
x1ddos, line 230: please add the note about accountsLock

x1ddos (commented):
`func (backend *Backend) Accounts() []accounts.Interface` also needs a lock. Otherwise, you get a concurrent usage race condition for `accounts` field, unless I misunderstand what `accountLock` is for?

But locking in `Backend.Accounts` wouldn't work because `initAccounts` requires a lock now and it calls `configureHistoryExchangeRates`, which in turn calls `Backend.Accounts`. Deadlock.

backend/backend.go
benma: done

benma:
> `func (backend *Backend) Accounts() []accounts.Interface` also needs a lock. Otherwise, you get a concurrent usage race condition for `accounts` field, unless I misunderstand what `accountLock` is for?
> 
> But locking in `Backend.Accounts` wouldn't work because `initAccounts` requires a lock now and it calls `configureHistoryExchangeRates`, which in turn calls `Backend.Accounts`. Deadlock.

Great catch. Fixed.

backend/backend.go
benma: done

backend/backend.go
benma: done (not the coins part, maybe in another PR)

backend/backend.go
```diff
@@ -208,9 +208,11 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 
 // configureHistoryExchangeRates changes backend.ratesUpdater settings.
 // It requires both backend.config to be up-to-date and all accounts initialized.
+//
+// The accountsLock must be held when calling this function.
 func (backend *Backend) configureHistoryExchangeRates() {
 	var coins []string
-	for _, acct := range backend.Accounts() {
+	for _, acct := range backend.accounts {
```
x1ddos, line 215: You got off easy this time :)

backend/backend.go
```diff
@@ -852,7 +877,10 @@ func (backend *Backend) Start() <-chan interface{} {
 	if backend.arguments.DevMode() {
 		backend.baseManager.Start()
 	}
+
+	unlock := backend.accountsLock.Lock()
 	backend.initPersistedAccounts()
+	unlock()
```
x1ddos, line 883: but now that configureHistoryExchangeRates requires the accontsLock, this unlock should be at the very end past configureHistory call. I would `defer` it.

x1ddos (approved):
LGTM with one more place to look at.

backend/backend.go
benma: damnit! fixed :)

PR #1127, opened 2020-12-12
frontend: update chart depending on the current source

thisconnect:
When leaving the chart on the weekly view after a while
it would be overwritten by the daily chart data.

It now checks the current source and updates the chart
accordingly.


benma (approved):
utack

benma:
merging so it is in the 4.24.1 release

PR #1135, opened 2021-01-13
frontend: only show buyable coins in moonpay dropdown

thisconnect:
The dropdown checks now each coin if buying is supported and
only shows those accounts.

### DIFF
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -23,8 +23,10 @@ import { load } from '../../decorators/load';
 import { translate, TranslateProps } from '../../decorators/translate';
 import { Button, Checkbox, Select } from '../../components/forms';
 import { Devices } from '../device/deviceswitch';
-import { AccountInterface } from '../account/account';
+import { AccountInterface, CoinCode } from '../account/account';
 import { setConfig } from '../../utils/config';
+import { apiGet } from '../../utils/request';
+import { isBitcoin } from '../account/utils';
 import * as style from './info.css';
 
 interface BuyInfoProps {
@@ -37,9 +39,15 @@ interface LoadedBuyInfoProps {
     config: any;
 }
 
+interface Option {
+    text: string;
+    value: CoinCode;
+}
+
 interface State {
     status: 'choose' | 'agree'
     selected?: string;
+    options: Option[]
 }
 
 type Props = BuyInfoProps & LoadedBuyInfoProps & TranslateProps;
@@ -48,6 +56,7 @@ class BuyInfo extends Component<Props, State> {
     public readonly state: State = {
         status: this.props.config.frontend.skipBuyDisclaimer ? 'choose' : 'agree',
         selected: this.props.code,
+        options: [],
     }
 
     private handleProceed = () => {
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -59,17 +71,31 @@ class BuyInfo extends Component<Props, State> {
...
+        Promise.all(
+            this.props.accounts.map((account) => (
+                apiGet(`account/${account.code}/exchange/moonpay/buy-supported`)
+                    .then(isSupported => (isSupported ? account : false))
+            ))
+        )
+        .then(results => results.filter(result => result))
+        // @ts-ignore
+        .then(accounts  => accounts.map(({ name, code }) => ({ text: name, value: code })))
+        .then(options => setTimeout(() => this.setState({ options }), 1000))
```
x1ddos, line 84: hm... why timeout?

frontends/web/src/routes/buy/info.tsx
```diff
@@ -59,17 +71,31 @@ class BuyInfo extends Component<Props, State> {
...
 
+    private checkSupportedCoins = () => {
+        Promise.all(
+            this.props.accounts.map((account) => (
+                apiGet(`account/${account.code}/exchange/moonpay/buy-supported`)
+                    .then(isSupported => (isSupported ? account : false))
+            ))
+        )
+        .then(results => results.filter(result => result))
+        // @ts-ignore
```
x1ddos, line 82: what's ignored here exactly?

frontends/web/src/routes/buy/info.tsx
```diff
@@ -170,17 +196,14 @@ class BuyInfo extends Component<Props, State> {
...
                                         },
-                                        ...accounts.map(({ code, name}) => ({
-                                            text: `${name} (${code})`,
-                                            value: code,
-                                        }))]
+                                        ...options]
                                     }
                                     onChange={e => this.setState({ selected: e.target.value})}
-                                    value={selected}
+                                    value={options.length ? selected : 'choose'}
```
x1ddos, line 206: if you kept it as is, `value={selected}`, would it still work?
I'm confused betwee this and the `text: t(options.length ? 'buy.info.selectPlaceholder' : 'loading')` above. seems like they're doing similar thing.

frontends/web/src/routes/buy/info.tsx
thisconnect: So in the filter before it filters out all results that are false (the ones that returned false in buy-supported), so on the next line TypeScript still thinks there could be `false`

Similar issue "filter should narrow down output type"
https://github.com/microsoft/TypeScript/issues/20812

Couldn't manage as suggested to do "User-Defined Type Guards"
https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

frontends/web/src/routes/buy/info.tsx
thisconnect: `text: t(options.length ? 'buy.info.selectPlaceholder' : 'loading')` is the text of the first grayed out option, this works if there is no preselected coin.

If there is a preselected coin, i.e. 'btc', the loading message would never show as the dropdown wants to show the option with the value 'btc'. But all accounts are checked asynchronously for if they can be bought, while checking options is an empty array, and value should point to the first grayed out option with the value 'choose' .... actually if there was a coin or token called 'choose' this wouldn't work .

does that make sense?

frontends/web/src/routes/buy/info.tsx
x1ddos: Interesting. So, something like this wouldn't work?

```ts
function isSupported(a: Account | bool): a is Account {
  return a !== false
}

.then(results => results.filter<Account>(isSupported))
```

frontends/web/src/routes/buy/info.tsx
x1ddos: but while the accounts are being checked, there's no way to select anything regardless of whether a coin is preselected, right.

what happens if the value remains as is, `value={selected}`, and `selected` is preset to say `btc`?

frontends/web/src/routes/buy/info.tsx
thisconnect: yep, there should not be a timeout, thanks for catching

frontends/web/src/routes/buy/info.tsx
thisconnect: then it doesn't show loading, but to be honest it happens so fast, I doubt that anyone would ever see the loading text.

x1ddos (approved):
LGTM. Can improve later.

PR #1138, opened 2021-01-15
frontend: support editing notes inline in tx overview

thisconnect:
Added a small edit button in the tx overview when in table layout.
The edit mode turns all notes/tx into a text field, and chnages
are automatically saved.


x1ddos:
Why closed?

thisconnect:
@jadzeidan was not happy with the UI change.

PR #1144, opened 2021-01-19
frontend: catch language c and fallback to en

thisconnect:
BitBoxApp currently does not support LANG=C.UTF-8 and errors with
`js: Uncaught RangeError: Invalid language tag: C`

Adding another fallback in case the proposed locale errors with Date.toLocaleString.

### DIFF
--- a/frontends/web/src/i18n/config.js
+++ b/frontends/web/src/i18n/config.js
@@ -35,6 +35,12 @@ export default {
             }
             apiGet('native-locale').then(locale => {
                 if (typeof locale === 'string' && locale) {
+                    try {
+                        new Date().toLocaleString(i18nextFormat(locale));
+                    } catch (e) {
+                        cb(defaultUserLanguage);
+                        return;
+                    }
                     cb(i18nextFormat(locale));
                     return;
                 }
--- a/frontends/web/test/i18n/config.test.tsx
+++ b/frontends/web/test/i18n/config.test.tsx
@@ -59,6 +59,20 @@ describe('language detector', () => {
         });
     });
 
+    it('uses defaultUserLanguage fallback if native-locale is C.UTF-8', done => {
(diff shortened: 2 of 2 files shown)


frontends/web/src/i18n/i18n.js
```diff
@@ -121,6 +121,9 @@ i18n.on('languageChanged', (lng) => {
             const localeLang = nativeLocale.replace('_', '-').split('-')[0];
             match = lngLang === localeLang;
         }
+        if (lng.toLowerCase() === 'c') {
+            lng = 'en';
```
x1ddos, line 125: I don't think this is needed: the `languageChanged` callback is triggered only from the language picker which includes only the supported translations plus "en". Is it actually possible for `lng` to be C?

frontends/web/src/i18n/config.js
```diff
@@ -34,7 +34,11 @@ export default {
                 return;
             }
             apiGet('native-locale').then(locale => {
-                if (typeof locale === 'string' && locale) {
+                if (
+                    typeof locale === 'string'
+                    && locale
+                    && !['C.UTF-8', 'C'].includes(locale.toUpperCase())
```
x1ddos, line 40: How about we test whether i18next supports this locale, regardless whether it's C, and fall back to `defaultUserLanguage`? Just thinking future-proof, in case a user has weird LANG set.

If there's no function in i18next then ok, I guess.

frontends/web/src/i18n/i18n.js
thisconnect: you are right, fixed

frontends/web/src/i18n/config.js
thisconnect: @x1ddos PTAL

x1ddos (approved):
LGTM. Thanks!

PR #1145, opened 2021-01-19
frontend/buy: change exchanges guide key

benma:
The old key has Safello text in translated languages, and the key
'support' also does not fit.


thisconnect (approved):
Thank you LGTM

PR #1147, opened 2021-01-20
moonpay: propagate fiat and language to the widget 

x1ddos:
Preferences are taken straight from the app config.

If a value is unsupported by Moonpay, they fall back to some default
values, according to their documentation. Worst case scenario, users
are still able to change both fiat and language directly in the widget.

### DIFF
--- a/frontends/web/test/i18n/config.test.tsx
+++ b/frontends/web/test/i18n/config.test.tsx
@@ -34,7 +34,7 @@ describe('language detector', () => {
     it('prefers userLanguage if available', done => {
         (apiGet as jest.Mock).mockImplementation(endpoint => {
             switch (endpoint) {
-                case 'config': { return Promise.resolve({frontend: {userLanguage: 'it'}}); }
+                case 'config': { return Promise.resolve({backend: {userLanguage: 'it'}}); }
                 case 'native-locale': { return Promise.resolve('de'); }
                 default: { return Promise.resolve(); }
             }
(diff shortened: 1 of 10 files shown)


frontends/web/test/i18n/config.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('language detector', () => {
     it('uses native-locale if userLanguage is null', done => {
         (apiGet as jest.Mock).mockImplementation(endpoint => {
             switch (endpoint) {
-                case 'config': { return Promise.resolve({frontend: {userLanguage: null}}); }
+                case 'config': { return Promise.resolve({backend: {userLanguage: ""}}); }
```
thisconnect, line 79: So it used to be stored as null in config.json, how can it be empty string?

Name of the test is 'uses native-locale if userLanguage is null'

Does `migrateUserLanguage` turn null to empty string?

thisconnect: Ok backend conf is now an empty stirng if the lang == native lang 👍 

but the frontend conf is still there, remove?

Screen Shot 2021-01-20 at 10 15 56 AM

thisconnect (approved):
LGTM with small nit

benma:
The first commit's message states a false premise. You can access the config from the account handler:

`handlers.account.Config()`.

benma (approved):
utACK

frontends/web/test/i18n/config.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('language detector', () => {
     it('uses native-locale if userLanguage is null', done => {
```
thisconnect, line 76: @x1ddos change to `'uses native-locale if userLanguage is an empty string'` ?

x1ddos: done

frontends/web/test/i18n/config.test.tsx
x1ddos: Add this comment. Makes sense?

image

thisconnect: thank you :)

PR #1149, opened 2021-01-20
frontend: moonpay check for supported coins on mount

thisconnect:
Before the check to find out which coins can be bought was only
executed when the user agreed to the disclaimer. With the 'do not
show the disclaimer again' option checked, the component shows
to the coin selection directly and needs to know which coins can
be bought rightaway.


x1ddos (approved):
LGTM. Thanks!

PR #1150, opened 2021-01-20
frontend: route back when unpluggin bb02 in buy view

thisconnect:
MoonPay needs an address to send to, therefore the device has
to be plugged in when using the service. Unplugging the device
should now route back to /.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -162,6 +162,10 @@ class App extends Component<Props, State> {
             route(`/device/${deviceIDs[0]}`, true);
             return;
         }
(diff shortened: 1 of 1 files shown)


frontends/web/src/app.tsx
```diff
@@ -162,6 +162,10 @@ class App extends Component<Props, State> {
             route(`/device/${deviceIDs[0]}`, true);
             return;
         }
+        if (deviceIDs.length === 0 && currentURL.startsWith('/buy/')) {
```
benma, line 165: isn't `accounts.length === 0` better, like above with the account summary? the buy component relies on accounts. not devices.

benma (changes requested):

benma (approved):

PR #1153, opened 2021-01-20
frontend: use monospaced font for pre element

thisconnect:
The pairing code font broke on macOS it used something with
serife like Times. We already use monospaced font in textareas
to display receive address or xpub.

Changed font for pre element to always use a monospace font.

Screen Shot 2021-01-20 at 10 26 32 PM


x1ddos (approved):
:+1:

PR #1157, opened 2021-01-22
frontend: fix moonpay background on large screen

thisconnect:
Did not cover the whole available area on larger screens.
Changed by just removing the content class that limits the
width of the content area.


thisconnect:
See white area left and right of the MoonPay iframe

x1ddos (approved):
LGTM

PR #1159, opened 2021-01-22
backend/moonpay: enable bank transfer payment method

x1ddos:
This concludes tweaking Moonpay API parameters for now.

For docs, see the following link but note that it's behind an
authentication, unfortunately:
https://www.moonpay.com/dashboard/getting_started/

The widget now looks like this:

screenshot-2021-01-22T11:10:02


benma (approved):
utACK

PR #1160, opened 2021-01-25
frontend: fix continue button not clickable in wizard

thisconnect:
Unfortunatelly I could only reproduce it once and got the error
js: Uncaught TypeError: Cannot read property 'querySelectorAll' of null
on line
https://github.com/digitalbitbox/bitbox-wallet-app/blob/6ed4ff598ce0a352e13bb5bc3a42edcebc83d47f/frontends/web/src/components/devices/bitbox02/bitbox02.tsx#L357

This fix removes the dom reference and instead adds each checkbox
as a boolean to the state. That way we don't depend on the form
to be there or having to querySelect the input elements and check
if they are checked.

### DIFF
--- a/frontends/web/src/components/devices/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/components/devices/bitbox02/bitbox02.tsx
@@ -72,7 +72,11 @@ interface State {
     // if true, we just pair and unlock, so we can hide some steps.
     unlockOnly: boolean;
     showWizard: boolean;
-    readDisclaimers: boolean;
+    agreement1: boolean;
+    agreement2: boolean;
+    agreement3: boolean;
+    agreement4: boolean;
+    agreement5: boolean;
     waitDialog?: {
         title: string;
         text?: string;
@@ -80,8 +84,6 @@ interface State {
 }
 
 class BitBox02 extends Component<Props, State> {
-    private disclaimerForm!: HTMLElement;
-
     constructor(props) {
         super(props);
         this.state = {
@@ -98,7 +100,11 @@ class BitBox02 extends Component<Props, State> {
             deviceName: '',
             unlockOnly: true,
             showWizard: false,
-            readDisclaimers: false,
+            agreement1: false,
+            agreement2: false,
+            agreement3: false,
+            agreement4: false,
+            agreement5: false,
             waitDialog: undefined,
         };
     }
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -349,19 +355,13 @@ class BitBox02 extends Component<Props, State> {
...
-            if (!checkbox.checked) {
-                result = false;
-            }
-        }
-        this.setState({ readDisclaimers: result });
+    private handleDisclaimerCheck = (event: InputEvent) => {
+        const target = event.target as HTMLInputElement;
+        const name = target.id as 'argement1' | 'argement2' | 'argement3' | 'argement4' | 'argement5';
+        this.setState(prevState => ({
+            ...prevState,
```
benma, line 362: i will never understand what this does and what a simple `this.setState({ [name]: target.checked })` would do wrong here. Can you enlighten me? :D

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
```diff
@@ -605,7 +610,7 @@ class BitBox02 extends Component<Props, State> {
                                         <div className={style.stepContext}>
                                             <p>{t('bitbox02Wizard.stepBackup.createBackup')}</p>
                                             <p className="m-bottom-default">{t('bitbox02Wizard.stepBackup.beforeProceed')}</p>
-                                            <form ref={this.setDisclaimerRef}>
+                                            <form>
                                                 <div className="m-top-quarter">
                                                     <Checkbox onChange={this.handleDisclaimerCheck} className={style.wizardCheckbox} id="agreement1" label={t('bitbox02Wizard.backup.userConfirmation1')} />
```
benma, line 615: don't you also need to put `checked={agreement1}` here and same with all others? e.g. like described here:

https://reactjs.org/docs/forms.html#handling-multiple-inputs

benma (approved):

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
thisconnect: The normal reason why you would do that is if you want to change the state based on the **current state**, i.e. toggle something on/off, this would be important if you click faster than react updating the state. In that case we can pass an async function instead of an object. 

see 
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

```js
// bad
this.setState({
  counter: this.state.counter + this.props.increment,
});

// good
this.setState((state, props) => ({
  counter: state.counter + props.increment
}));
```

but in this particular case, it is because TypeScript didn't like it... I didn't really deeply investigate deeply why 🙈

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
thisconnect: Thanks, yes you are right

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
thisconnect: same issue https://stackoverflow.com/questions/42090191/picks-k-type-with-dynamic-computed-keys

thisconnect: It seems Typescript doesn't like ES6 style dynamic property keys, but doesn't complain about oldschool

```js
const obj = {};
obj[target.id] = target.checked;
this.setState(obj);
```

frontends/web/src/components/devices/bitbox02/bitbox02.tsx
thisconnect: fixed and force pushed https://github.com/digitalbitbox/bitbox-wallet-app/compare/fd680efa61c2ea9244ce3a6c3646f9b6ecf47704..f20078846e3a01774b7715db466574c38d13395c

benma (approved):
utACK (I tested the previous version though, so this should be good too)

PR #1161, opened 2021-01-25
frontend,backend: correct and unify/combine whitelists for opening external URLs

x1ddos:
At the moment, we have only /buy/moonpay/ but the idea is we will
possibly add more in the future.

This commit also unifies Qt and Android frontend whitelists. They were
disaligned after switching routes back and forth.

There's no reason why keep two separate whitelists in the SystemOpen
function. They serve the same purpose and can be easily combined into
one.

So far, there doesn't seem to be a reason to use a more onerous regexp
compared to just a string prefix. This commit in fact does the latter
and all existing patters line up nicely.

If there's really a need in the future, can always switch to a regexp
based approach.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -21,7 +21,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


thisconnect (approved):
LGTM

benma (approved):
utACK

x1ddos:
Sorry, I've realized this wasn't enough so I added another commit. PTAL.

backend/backend.go
```diff
@@ -64,6 +63,36 @@ func init() {
...
+	// Block explorers.
+	"https://blockstream.info/tx/",
+	"https://blockstream.info/testnet/tx/",
+	"http://explorer.litecointools.com/tx/",
+	"https://insight.litecore.io/tx/",
+	"https://etherscan.io/tx/",
+	"https://rinkeby.etherscan.io/tx/",
+	"https://ropsten.etherscan.io/tx/",
+	// Moonpay onramp
+	"https://www.moonpay.com/",
```
benma, line 86: all of the ones that were an exact match before are now prefix checks. imho exact matches are slightly better, but the code is much cleaner now...

benma (approved):

backend/backend.go
x1ddos: right, only the exact match never worked for some at all. for example, we don't actually link to the moonpay.com root, ony some /privacy URL.

the whitelist purpose is to minimize some vulnerable URL injection, right. suppose an `example.org/foo` is whitelisted. if `example.org` is compromized, most likely so is its `/foo` page: not much benefit in exact match.

now, you could say but what if a web page has an XSS and a specially crafted `example.org/foo/%20<script>alert()` would pass. but then what about existing whitelisted URLs like `blockstream.info/tx/`. even with a regexp, they don't verify anything past `/tx/`. so, even with the original regexp, the same issue remains.

my point is, each URL to check should either be decomposed and compared against a domain, path and query params, or the benefits in the regexp vs prefix matching are questionable here.

yes we could change everything to match up to a single character, like in `blockstream\.info/tx/[a-fA-F0-9]{64}` and all the redirects in `ext.shiftcrypto.ch`... the list will get very long and I'm not sure how big of an advantage it would be considering a much more complicated list to maintain.

x1ddos: anywa, can always change to regexp in the future! if it turns out there's actually a good reason for it.

PR #1162, opened 2021-01-25
frontend: better actions buttons layout on mobile

thisconnect:
The new buy was miss-aligned, the buttons now have their own
line and are all equally spaced out and use the same width.

Screen Shot 2021-01-25 at 10 52 36 PM

cc @jadzeidan


jadzeidan:
LGTM

benma (approved):
tACK

PR #1164, opened 2021-01-26
Small moonpay fixes

benma:
Still could not find the 'undefined' issue though...

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -882,7 +882,9 @@ func (handlers *Handlers) getExchangeMoonpayBuySupported(r *http.Request) (inter
 			break
 		}
 	}
-	return acct != nil && exchanges.IsMoonpaySupported(acct.Coin().Code()), nil
+	// TODO: Offline() can be removed from here once there is a unified way of initializing accounts
+	// and showing sync status, offline/fatal states, etc.
+	return acct != nil && !acct.Offline() && exchanges.IsMoonpaySupported(acct.Coin().Code()), nil
 }
 
 func (handlers *Handlers) getExchangeMoonpayBuy(r *http.Request) (interface{}, error) {
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -900,6 +900,11 @@ func (handlers *Handlers) getExchangeMoonpayBuy(r *http.Request) (interface{}, e
 	if acct == nil {
 		return nil, fmt.Errorf("unknown account code %q", acctCode)
 	}
+
+	if err := acct.Initialize(); err != nil {
+		return nil, err
+	}
+
 	params := exchanges.BuyMoonpayParams{
 		Fiat: handlers.backend.Config().AppConfig().Backend.MainFiat,
 		Lang: handlers.backend.Config().AppConfig().Backend.UserLanguage,
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -191,7 +191,9 @@ class BuyInfo extends Component<Props, State> {
                                 </div>
                             </div>
                         ) : (
-                            options.length === 0 ? null : (
+                            options.length === 0 ? (
+                                <div class="content narrow isVerticallyCentered">{t('accountSummary.noAccount')}</div>
+                            ) : (
                                 <div class="content narrow isVerticallyCentered">
                                     <h1 class={style.title}>{t('buy.title', { name })}</h1>
                                     <Select
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -68,7 +68,13 @@ class BuyInfo extends Component<Props, State> {
         if (selected && (status === 'choose' || this.props.config.frontend.skipBuyDisclaimer)) {
             route(`/buy/moonpay/${selected}`);
         } else {
-            this.setState({ status: 'choose' });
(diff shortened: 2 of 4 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -83,11 +90,8 @@ class BuyInfo extends Component<Props, State> {
         // @ts-ignore
         .then(accounts => accounts.map(({ name, code }) => ({ text: name, value: code })))
         .then(options => {
-            if (this.state.status === 'choose' && options.length === 1) {
-                route(`/buy/moonpay/${options[0].value}`);
-            } else {
-                this.setState({ options });
-            }
+            this.setState({ options });
+            this.maybeProceed();
```
thisconnect, line 94: It is unclear when preact updates the state after `this.setState({ options })`, use the 2nd argument of setState that is a callback when the state has updated: `this.setState({ options }, this.maybeProceed);`
or 
```
this.setState({ options }, () => this.maybeProceed());
```

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -28,7 +28,7 @@ import * as style from './moonpay.css';
 
 interface BuyProps {
     accounts: AccountInterface[];
-    code?: string;
+    code: string;
```
thisconnect, line 31: why?

thisconnect (approved):
LGTM

with small nit about calling the next function after setState

frontends/web/src/routes/buy/info.tsx
benma: Good catch. The preact docs suck though, there is nothing about setState? https://preactjs.com/guide/v10/components

backend/handlers/handlers.go
```diff
@@ -882,7 +882,7 @@ func (handlers *Handlers) getExchangeMoonpayBuySupported(r *http.Request) (inter
 			break
 		}
 	}
-	return acct != nil && exchanges.IsMoonpaySupported(acct.Coin().Code()), nil
+	return acct != nil && !acct.Offline() && exchanges.IsMoonpaySupported(acct.Coin().Code()), nil
```
x1ddos, line 885: Doesn't feel semantically correct. The endpoint is supposed to respond whether account's coin is supported. I don't believe you need an account to be "online" in order to get a valid response from `IsMoonpaySupported`.

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -81,7 +81,7 @@ class Moonpay extends Component<Props, State> {
         { height }: State,
     ) {
         const account = this.getAccount();
-        if (!account) {
+        if (!account || moonpay.url === '' || moonpay.address === '') {
```
x1ddos, line 84: I think just the `moonpay.url` should be enough? The address is more like an informational field. I think we could even remove it. Initially, I was planning to show the address in the UI during the onramp flow but never got to it...

x1ddos:
Sorry, forgot: typo in commit 3716f60: "this would only happen if **he** disclaimer ..."

backend/handlers/handlers.go
benma: You don't have to have an online account to call `IsMoonpaySupported()`, but you need an online account to be able to use the account in moonpay. this is a quick fix / hack to not list offline accounts.

x1ddos: Ok. Add a TODO on top? :pray:

x1ddos (approved):
LGTM

backend/handlers/handlers.go
benma: I have an idea for a general fix after this:

A higher order component that basically does what account.tsx already does: initialize an account, show spinner while syncing, show 'offline' if offline, etc. Factor that out of account.tsx, apply it there and in moonpay.tsx (and possibly even in the portfolio).

Will add the TODO

frontends/web/src/routes/buy/moonpay.tsx
benma: Done

benma:
@x1ddos added another commit to fix the init, PTAL.

backend/handlers/handlers.go
```diff
@@ -900,6 +900,11 @@ func (handlers *Handlers) getExchangeMoonpayBuy(r *http.Request) (interface{}, e
 	if acct == nil {
 		return nil, fmt.Errorf("unknown account code %q", acctCode)
 	}
+
+	if err := acct.Initialize(); err != nil {
```
x1ddos, line 904: Oh, I get it now. Before, it was in account handlers which I guess was initialized. But now in the main handlers, it might not have been initialized yet. Does that sounds right?

x1ddos (approved):
:+1:

backend/handlers/handlers.go
x1ddos: (by "it" I meant `getExchangeMoonpayBuy` handler method)

benma: it was also not initialized before either.

it was accidentally initialized by the portfolio (chart.go), but if you click fast enough, it will not have gotten to Litecoin yet ;)

frontends/web/src/routes/buy/info.tsx
thisconnect: - https://preactjs.com/guide/v10/upgrade-guide/#dont-access-thisstate-synchronously
- https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

PR #1169, opened 2021-02-04
frontend/qt: correct origin check when enabling video capture

x1ddos:
The expression `securityOrigin.toString() != "qrc://"` was always true
and so the video capture was never enabled when scanning QR code,
resulting in a blank page.

Also, `QUrl::toString()` uses `PrettyDecoded` formatting by default, but:

> The exact behavior of PrettyDecoded varies from component to
> component and may also change from Qt release to Qt release.

See https://doc.qt.io/qt-5.12/qurl.html.
So, comparing the scheme seems more appropriate here.

Unclear when things changed. It is possible the condition has
never worked before but the permission to use video capture was granted
implicitly due to local origin. The fact is Windows and macOS works
without this change. So, this must be a combination of Qt + Linux.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -205,7 +205,7 @@ int main(int argc, char *argv[])
         view->page(),
         &QWebEnginePage::featurePermissionRequested,
         [&](const QUrl& securityOrigin, QWebEnginePage::Feature feature) {
-            if (securityOrigin.toString() != "qrc://") {
+            if (securityOrigin.scheme() != "qrc") {
                 return;
             }
             if (feature == QWebEnginePage::MediaVideoCapture) {


x1ddos:
Oh, I completely forgot to mention: `securityOrigin.toString()` returns `qrc:`, well at least on my linux.

x1ddos:
Wow, haven't seen the CI all green in a while :)
I guess they all finally caught up with go1.14.13.

benma:
@thisconnect can you test this on macOS with Qt 5.11.2 (the version we currently release under)? For me, camera on macOS does not work on master and not on this branch :thinking:

x1ddos:
@benma can you print `securityOrigin.toString()` somewhere right before `if (securityOrigin.scheme() != "qrc")` in your local build? I would expect it to work on qt5.11 too. Also, we don't really need to stick to that old version on macOS.

benma (approved):
tested on Linux and Windows. macOS does not work for me at all, even master, so I cannot test. Since it works on Windows with the same Qt version, I assume it will work.

x1ddos:
That's not good. What you mean by "at all"? Who's going to build the 
next release, me? :)

x1ddos:
Also, tried upgrading to qt5.12 on your macOS?

benma:
@x1ddos @thisconnect successfully tested - ready to merge

PR #1175, opened 2021-02-15
Frontend backup update

thisconnect:
- cleanup
- fix scrolling/jumping of long backup lists
- fix radio form element focus


benma (approved):
utACK

PR #1176, opened 2021-02-16
frontend: allow Moonpay to request camera access

benma:
They state that from March 31st, 2021, their in-app KYC process will
require camera access.

They said that
https://webrtc.github.io/samples/src/content/getusermedia/gum/ can be
used to test if it is working. This commit makes the camera at that
site work in the moonpay.tsx iframe.

In the Qt app, the security origin scheme check is dropped, as that is
easier than checking for the moonpay url. Note that all external
requests are blocked anyway, so only already whitelisted pages can
even be loaded, so the check was redundant.

### DIFF
--- a/frontends/web/src/utils/custom.d.ts
+++ b/frontends/web/src/utils/custom.d.ts
@@ -30,7 +30,7 @@ declare module '*.svg';
 declare namespace JSX { // tslint:disable-line:no-namespace
     interface HTMLAttributes {
         align?: 'left' | 'right' | 'center';
-        allow?: 'payment';
+        allow?: string;
         autocorrect?: 'on' | 'off';
         spellcheck?: boolean;
     }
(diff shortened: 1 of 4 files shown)


frontends/web/src/utils/custom.d.ts
```diff
@@ -30,7 +30,7 @@ declare module '*.svg';
 declare namespace JSX { // tslint:disable-line:no-namespace
     interface HTMLAttributes {
         align?: 'left' | 'right' | 'center';
-        allow?: 'payment';
+        allow?: 'camera; payment';
```
x1ddos, line 33: I think this should just be an arbitrary optional string at this point. Otherwise, it's basically a copy&paste typo checker  :)

x1ddos (approved):
LGTM

frontends/web/src/utils/custom.d.ts
benma: true! done

x1ddos: :+1:

PR #1177, opened 2021-02-16
Frontend backup success

thisconnect:
frontend: route to create and restore success screen
    
The success screen did not show up after restore from backup
due to the app.tsx rendering again and loosing the wizard state.
    
This refactors create and restore success message into its own
view. The routeing happens on keystoreAvailable, with the drawback
that it could race with the incoming accounts and keystores reload
Events.
An global frontend app state could take care about this in the future.

Also: removed unused CSS, relocated the bitbox02.tsx and renamed
BitBox01.jsx for consistency.


thisconnect:
Screen Shot 2021-02-16 at 6 25 18 PM

thisconnect:
Screen Shot 2021-02-16 at 6 38 52 AM

I think app.tsx ideally wouldn't just redirect when an accounts reload event is pushed by the backend.

We can listen to the later "keystoreAvailable" event and route to the success view in the component, but I had to fight the app.tsx maybeRoute, i.e. currently unplugin the device on the success view does not route and stays. This should be fine IMO.

I'd like to improve the routing soon, also the unluck video flickers every now and then  which is related.

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -144,6 +144,13 @@ class BitBox02 extends Component<Props, State> {
                         case 'attestationCheckDone':
                             this.updateAttestationCheck();
                             break;
+                        case 'keystoreAvailable':
```
benma, line 147: This does not seem robust. 

I don't understand the issue yet.

> The success screen did not show up after restore from backup
> due to the app.tsx rendering again and loosing the wizard state.

Why re-render, why loss of state? That does not sound like it should happen.

benma:
> I think app.tsx ideally wouldn't just redirect when an accounts reload event is pushed by the backend.

I see, so that route the issue. What if we just removed that?

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
thisconnect: Yes agree, this PR doesn't make it robust, it is just a fix by listening to an event and routing to success before other events would redirect to settings.

current issue as far as I can tell:

`app.tsx` currently listens/subscribes to `/accounts`
https://github.com/digitalbitbox/bitbox-wallet-app/blob/b67197a5b94dce6a33ef0b7771704f4678e20c46/frontends/web/src/app.tsx#L269-L276

Then checks for if props changed and if either devices or accounts are different than before it will maybeRoute
https://github.com/digitalbitbox/bitbox-wallet-app/blob/b67197a5b94dce6a33ef0b7771704f4678e20c46/frontends/web/src/app.tsx#L171-L175

Then in maybeRoute it will redirects to settings view `/device/<DEVICEID>` and jumps over the success view.

thisconnect: Currently I don't like the maybeRoute method, and would prefer if each view does routing.

The only reason to forcefully route away outside of your current view/component is unplugin the device. 
But even that could be done in each component/view instead of app.tsx, and I think could be done in reusable function that can be used everywhere.

Would like to discuss this in person if you have time today.

thisconnect:
> I see, so that route the issue. What if we just removed that?

Yes I'd like to remove the routing part (maybeRoute) from app.tsx and think each view should have the ability to control the routing when the view is active. But this is a bigger change and I would do it not in this PR.
So what about this PR, I agree it is not 100% robust and there could be a race. But that was already the case and currently it is broken. I currently tend to that we merge this PR and later change how the routing works.

benma:
Turns out the issue had nothing to do with routing. The issue was that the component was recreated when it should not have been, due to a buggy `key=` property. See the fix here, which should be more robust: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1186

thisconnect:
Thanks,  #1186 looks great, testing..

PR #1179, opened 2021-02-18
handlers: log endpoints again

benma:
We removed them to reduce logging volume, but the endpoints called are
very helpful output to trace what happened.


x1ddos (approved):
LGTM

PR #1181, opened 2021-02-22
frontend/transactions: show ETH nonce in details

benma:
Good advanced info, can help with understanding stuck tx problems
should they appear.

### DIFF
--- a/backend/coins/eth/etherscan/etherscan.go
+++ b/backend/coins/eth/etherscan/etherscan.go
@@ -116,6 +116,7 @@ type jsonTransaction struct {
 	BlockNumber jsonBigInt     `json:"blockNumber"`
 	GasUsed     jsonBigInt     `json:"gasUsed"`
 	GasPrice    jsonBigInt     `json:"gasPrice"`
+	Nonce       jsonBigInt     `json:"nonce"`
 	Hash        common.Hash    `json:"hash"`
 	Timestamp   timestamp      `json:"timeStamp"`
 	From        common.Address `json:"from"`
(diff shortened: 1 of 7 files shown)


benma:
Also finally started a changelog!

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -116,6 +116,7 @@ type jsonTransaction struct {
 	BlockNumber jsonBigInt     `json:"blockNumber"`
 	GasUsed     jsonBigInt     `json:"gasUsed"`
 	GasPrice    jsonBigInt     `json:"gasPrice"`
+	Nonce       jsonBigInt     `json:"nonce"`
```
x1ddos, line 119: Can't really find anything on the wire format in https://etherscan.io/apis#transactions but I guess you know better.

x1ddos: Apart from that, what if the `nonce` is nil or in a different format. Wouldn't that basically break the app in a way users won't be able to do anything with ETH? ...since it's `jsonBigInt`'s unmarshaler always assumes it must be valid.

For example, `txlist` action in `EtherScan.Transactions`. Does it guarantee that the nonce value is always valid?

x1ddos (commented):
kind of random commits in a single PR :)

backend/coins/eth/etherscan/etherscan.go
benma: yeah every tx has a nonce, i think that should be no problem.

benma:
> kind of random commits in a single PR :)

not really if you look at it this way: i made a change and i started a changelog ;)

the other one is also closely related cause both fields are handled the same way now.

x1ddos (approved):
LGTM

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -330,6 +331,12 @@ class Transaction extends Component<Props, State> {
                                     </div>
                                 ) : null
                             }
+                            {
```
benma, line 334: actually that is no good. the nonce has to be `undefined | number`, since it does not exist for BTC but can be 0 in eth (first tx).

benma: done

x1ddos: :+1: yeah, @thisconnect would be better here.
from the UI point of view though, not sure how useful it would be for me as a user. wouldn't know what to do with it. feels more like a debugging output.

PR #1183, opened 2021-02-22
frontend/chart: format BTC with 8 decimal places in chart tooltip

benma:


thisconnect (approved):
LGTM

One day I would even force every last 0 of ever satoshi to be shown.

// bad
0.0000001

// good
0.00000010

PR #1185, opened 2021-02-23
Italian

benma:


thisconnect (approved):
LGTM

FYI this also contains the fix for the moonpay title when it says just "Buy" instead of "Buy Bitcoin"

PR #1186, opened 2021-02-23
frontend/app: fix bb02 success screen by fixing devices key attribute

benma:
After creating a new wallet or restoring from backup, the app is
supposed to show a success screen. This has not happened due to a
regression.

The screen was skipped because the DeviceSwitch and the BitBox02
components were recreated completely, losing the internal state.

The components were recreated because the key attribute,
e.g. `key={['device-switch', devices]}`, "changed". The devices object
was the same, but in Javascript, one can't compare arrays like
this. `[] == []` or `['device-switch'] == [''device-switch']` is
always false, so the component was re-recreated.

We fix it by turning making the key a string derived from the devices,
so preact will correctly identify when the devices changed and when
they didn't change.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -168,6 +168,12 @@ class App extends Component<Props, State> {
         }
     }
 
+    // Returns a string representation of the current devices, so it can be used in the `key` property of subcomponents.
+    // The prefix is used so different subcomponents can have unique keys to not confuse the renderer.
+    private devicesKey = (prefix: string): string => {
+        return prefix + ':' + JSON.stringify(this.props.devices, Object.keys(this.props.devices).sort());
+    }
+
     public componentDidUpdate(prevProps) {
         if (prevProps.accounts !== this.props.accounts || prevProps.devices !== this.props.devices) {
             this.maybeRoute();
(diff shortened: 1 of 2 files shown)


benma:
This is an alternative to https://github.com/digitalbitbox/bitbox-wallet-app/pull/1184

frontends/web/src/app.tsx
```diff
@@ -168,6 +168,12 @@ class App extends Component<Props, State> {
         }
     }
 
+    // Returns a string representation of the current devices, so it can be used in the `key` property of subcomponents.
+    // The prefix is used so different subcomponents can have unique keys to not confuse the renderer.
+    private devicesKey = (prefix: string): string => {
+        return prefix + ':' + JSON.stringify(this.props.devices, Object.keys(this.props.devices).sort());
```
thisconnect, line 174: Why not just `JSON.stringify(this.props.devices)`, why pass a sorted array of the keys as allowlist?

> A function that alters the behavior of the stringification process, or an array of String and Number that serve as an allowlist for selecting/filtering the properties of the value object to be included in the JSON string. If this value is null or not provided, all properties of the object are included in the resulting JSON string.

thisconnect (approved):
LGTM with open question

frontends/web/src/app.tsx
benma: I wasn't sure if the json serialization is deterministic. json object keys are not ordered in general, so in theory serialization could happen in any order.

thisconnect: Ah I see, so but then just `JSON.stringify(Object.keys(this.props.devices).sort())`.  The second argument acts as a allowlist or does it have an effect on the order ?

benma: Good question. Your version does not include the values. Per documentation, the 2nd arg is only an allowlist, but in my tests at least it did serialize in that order (I took it from some stackoverflow post).

thisconnect: yeah but the values are currently either 'bitbox01' or 'bitbox02' and the keys are already unique.

PR #1188, opened 2021-02-23
frontend/bitbox02: fix sd card inserted state

benma:
The BackupsV2 component would be fed the sdCardInserted property via
the URL from the Settings component. This is fragile, e.g. if you
remove the sdcard while in the settings screen and then go to backups,
you will get an error.

The url part is also only used for the BitBox02, but the ManageBackups
component is for both BitBox01 and BitBox02, which makes this also
confusing.

For now we abstract away the microSD card check and use it when
loading the Backups component, so it should always work.

There is more refacoring that can be done later, e.g. cleaning up the
back button mess.

### DIFF
--- a/frontends/web/src/components/devices/bitbox02/sdcardcheck.tsx
+++ b/frontends/web/src/components/devices/bitbox02/sdcardcheck.tsx
@@ -1,0 +1,89 @@
+/**
+ * Copyright 2021 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


benma:
Fixes #674

frontends/web/src/components/devices/bitbox02/sdcardcheck.tsx
```diff
@@ -0,0 +1,89 @@
...
+            <div>
+                {children}
+            </div>
+        );
+    }
+
+}
+
+const HOC = translate<SDCardCheckProps>()(SDCardCheck);
+export { HOC as SDCardCheck };
```
thisconnect, line 89: ❤️

thisconnect (approved):
LGTM

thisconnect:
fixed buttons in https://github.com/thisconnect/bitbox-wallet-app/tree/frontend-backup-styling it's @jadzeidan approved

PR #1189, opened 2021-02-23
frontend: update manage backup layout

thisconnect:
- changed orange button to blue
- increase backup list width and heigth to make better use of the
  screen and prevent a second scrollbar on the container
- rearanged buttons

Screen Shot 2021-02-23 at 9 23 35 PM

Screen Shot 2021-02-23 at 9 27 27 PM


benma (approved):
tACK

PR #1190, opened 2021-02-24
frontend: fix manage backup buttons

thisconnect:
On the manage backups view the buttons have no space and on restore
from backup the buttons are too big. The gap CSS attribute only works
in newer webkit versions.

- changed from gap to margin to fix spacing issue
- removed button height on the restore backup from sd card, so that
  the button uses the default button height

### DIFF
--- a/frontends/web/src/components/backups/backups.css
+++ b/frontends/web/src/components/backups/backups.css
@@ -59,8 +59,13 @@
     display: flex;
     flex-wrap: wrap;
     flex-direction: row-reverse;
-    gap: var(--space-half);
     margin-top: var(--space-default);
+}
+
+@media (min-width: 769px) {
+    .backupButtons > *:not(:first-child) {
+        margin-right: var(--space-half);
+    }
 }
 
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/backups/backups.css
```diff
@@ -59,14 +59,23 @@
     display: flex;
     flex-wrap: wrap;
     flex-direction: row-reverse;
-    gap: var(--space-half);
     margin-top: var(--space-default);
 }
 
+@media (min-width: 769px) {
+    .backupButtons > * + * {
```
benma, line 66: what does `* + *` mean?

benma (approved):
utACK

frontends/web/src/components/backups/backups.css
thisconnect: :)  it's "the owl" selector 🦉 https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/ 

basically targets anything that has is not first child, can change to `*:not(:first-child)` if you want

PR #1192, opened 2021-02-27
Frontend refactor account sync

thisconnect:
supersedes https://github.com/digitalbitbox/bitbox-wallet-app/pull/1172

### DIFF
--- a/frontends/web/src/api/accountsync.ts
+++ b/frontends/web/src/api/accountsync.ts
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2021 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { subscribeEndpoint } from './subscribe';
+import { Unsubscribe } from '../utils/event';
+import { IAccount} from './account';
+
+/**
+ * Subscribes the given function on the "account" event and receives the
+ * list of all available accounts from the backend.
+ * Returns a method to unsubscribe.
+ */
+
+export const syncAccountsList = (
+    cb: (accounts: IAccount[],) => void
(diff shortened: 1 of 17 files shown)


frontends/web/src/api/accountsync.ts
```diff
@@ -0,0 +1,48 @@
...
+    return subscribeEndpoint('accounts', cb);
+};
+
+/**
+ * Subscribes the given function on an "account/<CODE>/synced-addresses-count" event
+ * to receive the progress of the address sync.
+ * Returns a method to unsubscribe.
+ */
+
+export const syncAddressesCount = (
```
benma, line 39: why this style over `export function syncAddressesCount(...)`?

thisconnect: As long as we don't use `this`, the `arguments` object or call the function with `new` it shouldn't matter. 

Looking at your "function declaration" `export function syncAddressesCount(...)` I think it is nicer and improves readability, do you want me to change this for all api stuff?

There is a few other differences, for example:

> The difference is that functionOne is a function expression and so only defined when that line is reached, whereas functionTwo is a function declaration and is defined as soon as its surrounding function or script is executed (due to hoisting).

https://stackoverflow.com/questions/336859/var-functionname-function-vs-function-functionname#answer-336868

thisconnect:
Could you approve this @x1ddos ?

x1ddos (approved):
Ah sorry. If this is identical to #1172 then LGTM.

frontends/web/src/routes/account/account.tsx
```diff
@@ -113,12 +106,26 @@ class Account extends Component<Props, State> {
...
 
+    private subscribe() {
+        this.subscribtions.push(
+            syncAddressesCount(this.props.code, (code, syncedAddressesCount) => {
+                if (code === this.props.code) {
+                    this.setState({ syncedAddressesCount });
+                }
+            }),
+            statusChanged(this.props.code, () => this.onStatusChanged()),
+            syncdone(this.props.code, () => this.onAccountChanged()),
```
thisconnect, line 125: @x1ddos I had to change this https://github.com/thisconnect/bitbox-wallet-app/commit/bbd8e4dabb8a45c705724c2063f34dc122f36571


Switching accounts during account sync did not update the
subscriptions correctly.

Changed:

- use componentDidUpdate instead of componentWillReceiveProps
  (which is deprecated in react).
- the functions passed to syncdone and statusChanged were called
  with the wrong this, so the callbacks could not access
  this.props anymore. Passing an arrow function will call the
  function with the correct this.

x1ddos: Right. That was the reason for re-creating the frontend-new-api branch. :+1:

PR #1194, opened 2021-03-02
Frontend api devices

thisconnect:
- adds api/devices and api/devicessync
- rewrite app.tsx to use api/devices*
- change get started and unlock routing to directly go to account-summary
- updated app / device routing

### DIFF
--- a/frontends/web/src/api/devices.ts
+++ b/frontends/web/src/api/devices.ts
@@ -1,0 +1,27 @@
+/**
+ * Copyright 2021 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { apiGet } from '../utils/request';
+
+type TProductName = 'bitbox' | 'bitbox02' | 'bitbox02-bootloader';
+
+export type TDevices = {
+    readonly [key in string]: TProductName;
+};
+
+export const getDeviceList = (): Promise<TDevices> => {
+    return apiGet('devices/registered');
+};
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -19,6 +19,8 @@ import { Component, h, RenderableProps } from 'preact';
 import { getCurrentUrl, route } from 'preact-router';
 import { getAccounts, IAccount } from './api/account';
 import { syncAccountsList } from './api/accountsync';
+import { getDeviceList, TDevices } from './api/devices';
+import { syncDeviceList } from './api/devicessync';
 import { unsubscribe, UnsubscribeList } from './utils/subscriptions';
 import { ConnectedApp } from './connected';
 import { Alert } from './components/alert/Alert';
@@ -30,7 +32,6 @@ import { MobileDataWarning } from './components/mobiledatawarning';
 import { Sidebar, toggleSidebar } from './components/sidebar/sidebar';
 import TranslationHelper from './components/translationhelper/translationhelper';
 import { Update } from './components/update/update';
-import { subscribe } from './decorators/subscribe';
 import { translate, TranslateProps } from './decorators/translate';
 import { i18nEditorActive } from './i18n/i18n';
 import { Account } from './routes/account/account';
@@ -43,7 +44,7 @@ import { Send } from './routes/account/send/send';
 import { AccountsSummary } from './routes/account/summary/accountssummary';
 import { BitBoxBase, setBaseUserStatus, setInternalBaseStatus, updateSharedBaseState } from './routes/bitboxbase/bitboxbase';
 import { BitBoxBaseConnect, DetectedBitBoxBases } from './routes/bitboxbase/bitboxbaseconnect';
-import { Devices, DeviceSwitch } from './routes/device/deviceswitch';
+import { DeviceSwitch } from './routes/device/deviceswitch';
 import ManageBackups from './routes/device/manage-backups/manage-backups';
 import { Exchanges } from './routes/exchanges/exchanges';
 import ElectrumSettings from './routes/settings/electrum';
@@ -53,21 +54,19 @@ import { apiWebsocket } from './utils/websocket';
 
 interface State {
     accounts: IAccount[];
-    detectedBases: DetectedBitBoxBases;
     bitboxBaseIDs: string[];
+    detectedBases: DetectedBitBoxBases;
+    devices: TDevices;
 }
 
-interface SubscribedProps {
-    devices: Devices;
-}
-
-type Props = SubscribedProps & TranslateProps;
+type Props = TranslateProps;
 
 class App extends Component<Props, State> {
     public readonly state: State = {
         accounts: [],
-        detectedBases: {},
         bitboxBaseIDs: [],
+        detectedBases: {},
+        devices: {},
     };
 
     private unsubscribe!: () => void;
@@ -116,13 +115,18 @@ class App extends Component<Props, State> {
             }
         });
 
-        getAccounts()
-            .then(accounts => this.setState({ accounts }, this.maybeRoute))
+        Promise.all([getAccounts(), getDeviceList()])
+            .then(([accounts, devices]) => {
+                return this.setState({ accounts, devices }, this.maybeRoute);
+            })
             .catch(console.error);
 
         this.unsubscribeList.push(
             syncAccountsList(accounts => {
                 this.setState({ accounts }, this.maybeRoute);
+            }),
+            syncDeviceList(devices => {
+                this.setState({ devices }, this.maybeRoute);
             }),
             // TODO: add syncBackendNewTX
             // TODO: add syncBitBoxBase ?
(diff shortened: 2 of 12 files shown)


thisconnect:
This is my little testing routine

```
- tap side
- unlock
- should see account summary

- unplug/replug

- tap side
- unplug/replug
- should see waiting screen

- tap side
- restart app
- should see unlock screen

- factory reset
- tap side
- should see verify pairing code
- unplug/replug
- tap side
- should see verify pairing code
- now either create or restore


```

thisconnect:
Github shows 5 commits but it is only the 2 latest, the other 3 are now merged (ff) into frontend-new-api

thisconnect:
> Please rebase; I was halfway through reviewing when I noticed that some of the commits were already merged 🙈

Sorry about that, but I left a comment. Not sure why it showed the 3 merged commits, not it looks ok only 2 commits.

https://github.com/digitalbitbox/bitbox-wallet-app/pull/1194#issuecomment-791506273

frontends/web/src/api/devices.ts
```diff
@@ -0,0 +1,282 @@
...
+
+type TSuccess = {
+    success: true;
+}
+
+// if the backend uses maybeBB02Err
+type TFail = {
+    code?: number;
+    message?: string;
+    success: false;
```
benma, line 38: success: boolean?

frontends/web/src/api/devices.ts
```diff
@@ -0,0 +1,282 @@
...
+    | 'unpaired' | 'pairingFailed'
+    | 'uninitialized' | 'seeded' | 'initialized'
+    | 'require_firmware_upgrade' | 'require_app_upgrade';
+
+export type TStatusBB01 = 'unpaired' | 'pairingFailed'
+    | 'uninitialized' | 'seeded' | 'initialized'
+    | 'require_firmware_upgrade' | 'require_app_upgrade';
+
+type TSuccess = {
+    success: true;
```
benma, line 31: boolean?

frontends/web/src/api/devices.ts
```diff
@@ -0,0 +1,282 @@
...
+
+// firmware
+
+export type TBundledFirmwareVersion = {
+    canUpgrade: boolean;
+    currentVersion: string;
+    newVersion: string;
+}
+
+export const getBundledFirmwareVersion = (
```
benma, line 135: many of these funcs in here seem unused. please remove them until they are needed, otherwise it's difficult to see if they are correct without being able to test their use.

frontends/web/src/api/devices.ts
```diff
@@ -0,0 +1,282 @@
...
+import { apiGet, apiPost } from '../utils/request';
+
+type TProuctName = 'bitbox' | 'bitbox02' | 'bitbox02-bootloader';
+
+export type TStatus = 'connected'
+    | 'unpaired' | 'pairingFailed'
+    | 'uninitialized' | 'seeded' | 'initialized'
+    | 'require_firmware_upgrade' | 'require_app_upgrade';
+
+export type TStatusBB01 = 'unpaired' | 'pairingFailed'
```
benma, line 26: can you make a bitbox02.ts and a bitbox01.ts? it would be a lot clearer imho.

benma (changes requested):

frontends/web/src/api/devices.ts
thisconnect: Similar to TFail, TSuccess is always success true

frontends/web/src/api/devices.ts
benma: I see!

thisconnect: With that TypeScript knows which other properties the object has.

frontends/web/src/api/devices.ts
thisconnect: This time, I didn't copy from the existing code, but typed the api by reading go backend, so I need to back it up :)

FYI on this branch there is still unused `api/account.ts`, we should merge master into this branch.

frontends/web/src/api/devices.ts
thisconnect: I like, but there is shared functiontions and types that are not specific to bitbox01 or bitbox02, should we keep the shared ones in `api/devices.ts` or have it in `api/bitbox02.ts`?

frontends/web/src/api/devices.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { apiGet } from '../utils/request';
+
+export type TProuctName = 'bitbox' | 'bitbox02' | 'bitbox02-bootloader';
```
benma, line 19: typo. also does not need to be expoerted?

frontends/web/src/api/devices.ts
benma: concept ack, nice!

frontends/web/src/api/devices.ts
thisconnect: removed `export`

frontends/web/src/api/devices.ts
thisconnect: ok updated and since `api/bitbox01.ts` and `api/bitbox02.ts` isn't used it is not anymore part of this PR

thisconnect:
@benma squashed and removed unused code, actually didn't even add `bitbox01.ts` and `bitbox02.ts` PTAL

thisconnect:
There is currently a routing issue after unlock.

thisconnect:
Because this PR changes how the app component gets devices, before via decorator props and now directly into state added 1 commit to fix and simplify the routing on new devices/accounts.

Tests:
* plugin device: route to devices
* unlock device: route to account-summary (default view still quickly flickers)
* unplug device: route and show default view (from account-summary, account, send, receive, buy, settings, manage-backup)
* unplug device: show default view (from device, doesn't route)
* restart app with unlocked bb: restart servewallet, navigate to / routes to account-summary 
* devmode: reload should stay on the same view (account-summary, account, send, receive, buy, settings)
    - Bug: reloading manage-backup redirects to account-summary
* devmode: restart app, similar to restart app but already on a route, behaves same as devmode just reload with the manage-backup bug

Please let me know if you can think of any other situations that have to be tested.

frontends/web/src/app.tsx
```diff
@@ -116,14 +113,33 @@ class App extends Component<Props, State> {
...
-            .then(accounts => this.setState({ accounts }, this.maybeRoute))
-            .catch(console.error);
+        getDeviceList()
+        .then(devices => {
+            return getAccounts()
+            .then(accounts => {
+                this.setState({ accounts, devices }, this.maybeRoute);
+            });
+        })
+        .catch(console.error);
```
benma, line 124: it is not so nice that we now have to both manually call the endpoints and also subscribe the endpoints, each time doing similar/the same things after it finishes.

the previous subscribe decorator took care of both. can your new subscribe api also take care of both at the same time?

food for thought, not in this PR.

frontends/web/src/app.tsx
```diff
@@ -116,14 +113,33 @@ class App extends Component<Props, State> {
...
 
         this.unsubscribeList.push(
             syncAccountsList(accounts => {
                 this.setState({ accounts }, this.maybeRoute);
             }),
+            syncDeviceList(devices => {
+                const oldDeviceIDList = Object.keys(this.state.devices);
```
benma, line 130: Scattering routing logic makes it harder to reason about the routing conditions and rules.

Why not simply call `this.maybeRoute` like above with the accounts, and keep the routing code there?

benma (approved):

frontends/web/src/app.tsx
thisconnect: Currently:
- App.tsx and devices.jsx (old BB01) are the only places where we subscribe and fetch `devices/registered`.
- Also App.tsx is the only place where it subscribes to syncAccountsList, which are then passed as props to the children.

> each time doing similar/the same things after it finishes

Was doing the same/similar thing, but not anymore with this PR. The routing fix in this PR needs to compare the current devices list with the new one, so it can't be in mayberoute.

Personally I don't like mayberoute it was the cause of various routing bugs in the past and has to maintain a centralized list of exception routing for different views, it is especially hard to follow if there is additional routing in views.

I'd prefer if mayberoute could go away in the future and each view decides for itself if it should be routed to `/` when the device is unplugged or not.
i.e. something like `import { routeToDefaultViewOnDeviceUnplug } from 'utils/routingHelper'` and  `this.unsubscribe = routeToDefaultViewOnDeviceUnplug();`

The device subscription probably makes sense that it stays in app.tsx but the accountlist fetching + subscription could be in a reusable component and only included in views that actually need it.

frontends/web/src/app.tsx
thisconnect: without this PR the app routes many times back and forth and after routing happen routes again and was really hard to follow.

This PR reduces the amount of routing and depends on checking if there is a new device, this can't be done in  maybeRoute.

Also there are other arguments against maybeRoute see my other comment.

thisconnect:
> Please let me know if you can think of any other situations that have to be tested.

@benma I'll try to test with BB01, but please let me know if you can think of any other testing that I forgot.

frontends/web/src/app.tsx
benma: > Personally I don't like mayberoute it was the cause of various routing bugs in the past and has to maintain a centralized list of exception routing for different views, it is especially hard to follow if there is additional routing in views.

Fyi, before we had maybeRoute, we had `route()` scattered and it also had many routing bugs and routing regressions. I had put in one place hoping that it would improve that.

But my comment here was actually not about maybeRoute() at all. Just an observation that for one `subscribe` decorator now, there are two very similar calls that have to be done now, e.g. `getDeviceList` and `syncDeviceList`. Not a problem for now.

frontends/web/src/app.tsx
```diff
@@ -116,14 +113,33 @@ class App extends Component<Props, State> {
             }
         });
 
-        getAccounts()
-            .then(accounts => this.setState({ accounts }, this.maybeRoute))
-            .catch(console.error);
+        getDeviceList()
```
benma, line 116: A comment here explaining why exactly the two calls are chained like this will help in the future, I think.

frontends/web/src/app.tsx
thisconnect: > Fyi, before we had maybeRoute, we had route() scattered and it also had many routing bugs and routing regressions. I had put in one place hoping that it would improve that.

👍 

For now, I think we should stick to the current way, devices routing done by app with a list of exceptions. 

Alternative would be to not route in app.tsx at all and make each view take care about the logic (this could be done with reusable modules to be DRY).

frontends/web/src/app.tsx
thisconnect: You're right, the requests can be done in parallel, no reason to wait, changed back and squashed 
https://github.com/digitalbitbox/bitbox-wallet-app/compare/d27e84efd912e3a1814cf96d168e64aa45ed3323..1711d06eac771a604f93ac9d2f42085f39fcb534

Also tested BB01 pairing, sending, restore form backup and quickly tested BB02 one more time, will merge now.

PR #1197, opened 2021-03-05
frontend: dedup coin and fiat types in rates

thisconnect:
Only app, accounts-summary, accounts tsx and some of the involved
modules have now migrated to api, but this ended up that we have
duplicate types for coin and fiat.

Moved fiat and coin type to api/account and import it from there,
so it shouldn't have anymore duplicate types.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -45,10 +45,6 @@ export const init = (code: string): Promise<null> => {
     return apiPost(`account/${code}/init`);
 };
 
-export const reinitialize = (): Promise<null> => {
-    return apiPost('accounts/reinitialize');
-};
-
 export interface ISummary {
     chartDataMissing: boolean;
     chartDataDaily: ChartData;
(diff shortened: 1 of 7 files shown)


thisconnect:
@benma I commented out everything in `api/accounts` that is currently not used, but ofc I want to rewrite everything to use it, but this will happen in the https://github.com/digitalbitbox/bitbox-wallet-app/tree/frontend-new-api branch.

Should I keep the unused api, or leave it commented out, or remove for now?

Ideally I'd like to merge master into frontend-new-api branch, but don't want it to be gone there, so I'd like to keep them commented out.

Let me know how to proceed.

frontends/web/src/api/account.ts
```diff
@@ -45,9 +45,9 @@ export const init = (code: string): Promise<null> => {
     return apiPost(`account/${code}/init`);
 };
 
-export const reinitialize = (): Promise<null> => {
-    return apiPost('accounts/reinitialize');
-};
+// export const reinitialize = (): Promise<null> => {
```
benma, line 48: I slightly prefer to not have commented or unused code in master, so I'd just delete it and add it in the commit where it is gonna be used.

benma (approved):

frontends/web/src/api/account.ts
thisconnect: done https://github.com/digitalbitbox/bitbox-wallet-app/compare/78f50875e799d2914383d768b30ba632454a0fab..ca06edf99cf5be6a0e9a7b913f3f9ffc2e0e0bba

PR #1199, opened 2021-03-08
frontend/account: bundle status data

benma:
It matches better what the getStatus call handling wants.

This will make it easier to add additional info such as error codes or
error messages to be displayed.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -41,9 +41,14 @@ export const getAccounts = (): Promise<IAccount[]> => {
     return apiGet(`accounts`);
 };
 
-export type AccountState = 'accountSynced' | 'accountDisabled' | 'fatalError' | 'offlineMode';
+export interface IStatus {
+    disabled: boolean;
+    synced: boolean;
+    fatalError: boolean;
+    offline: boolean;
+}
 
-export const getStatus = (code: string): Promise<AccountState[]> => {
+export const getStatus = (code: string): Promise<IStatus> => {
     return apiGet(`account/${code}/status`);
 };
 
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -47,14 +47,12 @@ interface LoadedAccountProps {
 }
 
 interface State {
-    accountSynced: boolean;
-    offlineMode?: boolean;
+    status?: accountApi.IStatus;
     transactions?: accountApi.ITransaction[];
     balance?: accountApi.IBalance;
     hasCard: boolean;
     exported: string;
     accountInfo?:accountApi. ISigningConfigurationList;
-    fatalError: boolean;
     syncedAddressesCount?: number;
 }
 
@@ -62,14 +60,12 @@ type Props = LoadedAccountProps & AccountProps & TranslateProps;
 
 class Account extends Component<Props, State> {
     public readonly state: State = {
-        accountSynced: false,
-        offlineMode: undefined,
+        status: undefined,
         transactions: undefined,
         balance: undefined,
         hasCard: false,
         exported: '',
         accountInfo: undefined,
-        fatalError: false,
         syncedAddressesCount: undefined,
     };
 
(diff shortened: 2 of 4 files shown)


frontends/web/src/api/account.ts
```diff
@@ -41,9 +41,14 @@ export const getAccounts = (): Promise<IAccount[]> => {
     return apiGet(`accounts`);
 };
 
-export type AccountState = 'accountSynced' | 'accountDisabled' | 'fatalError' | 'offlineMode';
+export interface IStatus {
+    disabled: boolean;
+    synced: boolean;
+    fatalError: boolean;
+    offline: boolean;
+}
```
thisconnect, line 49: What's nice about `AccountState = 'accountSynced' | 'accountDisabled' | 'fatalError' | 'offlineMode'` is that there can only be 1 state, but if all the permutations are possible then this doesn't work of corse.

So can the app be synced AND offline AND disabled at the same time?

frontends/web/src/routes/account/account.tsx
```diff
@@ -179,16 +169,17 @@ class Account extends Component<Props, State> {
...
     }
 
     private onAccountChanged = () => {
-        if (!this.props.code || this.state.fatalError) {
+        const status = this.state.status;
+        if (!this.props.code || status === undefined || status.fatalError) {
             return;
         }
-        if (this.state.accountSynced && !this.state.offlineMode) {
+        if (status.synced && !status.offline) {
```
thisconnect, line 182: I'd assume it to be online if `status.synced` is true, is `!status.offline` really necessary?

frontends/web/src/routes/account/account.tsx
```diff
@@ -217,7 +208,7 @@ class Account extends Component<Props, State> {
     }
 
     private export = () => {
-        if (this.state.fatalError) {
+        if (this.state.status === undefined || this.state.status.fatalError) {
```
thisconnect, line 211: fine for now, but in the future I would like to hide the export button or make it disabled,  if app has no status yet or has a fatalError

frontends/web/src/routes/account/account.tsx
```diff
@@ -257,21 +248,20 @@ class Account extends Component<Props, State> {
...
             accountInfo,
-            fatalError,
             syncedAddressesCount,
         }: State) {
         const account = accounts &&
                         accounts.find(acct => acct.code === code);
-        if (!account) {
+        if (!account || status === undefined) {
             return null;
         }
```
thisconnect, line 263: FYI, when new-api is merged (soon™) I'd like to get rid of this `return null` because that is the main reason for the flickering when changing accounts, I believe. The alternative is to let all components render, but they might render them self invisible, or with a gray area to indicate that something is comming.

frontends/web/src/api/account.ts
benma: It wasn't used like that though, cause the getStatus call returned an array of this, not exactly this.

I agree it might be nicer, but out of scope for now - this is only a small refactor to allow showing errors in the end.

frontends/web/src/routes/account/account.tsx
benma: These are independent. You can lose the connection and reconnect after you have synced, for example.

frontends/web/src/routes/account/account.tsx
```diff
@@ -47,29 +47,25 @@ interface LoadedAccountProps {
...
     syncedAddressesCount?: number;
 }
 
 type Props = LoadedAccountProps & AccountProps & TranslateProps;
 
 class Account extends Component<Props, State> {
     public readonly state: State = {
-        accountSynced: false,
-        offlineMode: undefined,
+        status: undefined,
```
thisconnect, line 63: Nit: I'd prefer status not optional and initialize it here with the defaults, I guess `{ disabled: false, synced: false, fatalError: false, offline: false}`, WDYT?

thisconnect (approved):
LGTM with small initialization nit

PR #1200, opened 2021-03-08
Surface blockchain connection errors; add delay before reconnecting

benma:


backend/accounts/account.go
```diff
@@ -61,7 +61,7 @@ type Interface interface {
 	Initialize() error
 	// Synced indicates whether the account has loaded and finished the initial sync.
 	Synced() bool
-	Offline() bool
+	Offline() error
 	FatalError() bool
```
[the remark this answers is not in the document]
x1ddos: I found the description somewhere in the handlers.go. For `FatalError` it says:

> // FatalError indicates that there was a fatal error in handling the account. When this happens,
> // an error is shown to the user and the account is made unusable.

So, almost except for "account is made unusable" but not sure what it means. Can't click anything in the UI? But can you do anything when "offline"?

util/jsonrpc/jsonrpc.go
```diff
@@ -278,16 +275,16 @@ func (client *RPCClient) establishConnection(backend *Backend) error {
...
+		callback(err)
 	}
 }
 
-func (client *RPCClient) setStatus(status Status) {
-	if status != client.status {
-		client.status = status
-		go client.notify(status)
+func (client *RPCClient) setConnectionError(err error) {
+	if err != client.connectionError {
```
x1ddos, line 286: Same here: a second `setConnectionError(errors.New("I'm offline"))` will result in an event notification but a `setConnectionError(io.EOF)` won't.

util/jsonrpc/jsonrpc.go
```diff
@@ -294,37 +289,63 @@ func (client *RPCClient) setConnectionError(err error) {
...
+
+	conn, err := func() (*connection, error) {
+		client.connLock.Lock()
+		defer client.connLock.Unlock()
+		if client.connection != nil {
+			return client.connection, nil
+		}
+
+		start := 0
+		if len(client.backends) > 0 {
```
x1ddos: I think this would be pretty bad if `len(client.backens)` is zero, no? Then the `lastErr` below is always nil since the loop is never run and I wouldn't be surprised if this resulted in a nil dereference panic when accessing `client.connection` - it looks like it'll be `nil`.

util/jsonrpc/jsonrpc.go
```diff
@@ -294,37 +289,63 @@ func (client *RPCClient) setConnectionError(err error) {
...
+		if lastErr != nil {
+			client.log = client.log.WithField("backend", "offline")
+			// tried all backends
+			client.log.Debug("Disconnected from all backends")
+			client.setConnectionError(lastErr)
+			return nil, lastErr
+		}
+		client.setConnectionError(nil)
+		return client.connection, nil
+	}()
```
x1ddos: `client.connLock` is released here but `client.connection` isn't ready yet. I think this is still a source of bugs.

Ideally, you'd hold `client.connLock` for the whole duration of the method but you can't because `client.onConnectCallback()` runs the electrum protocol negotiation which needs `client.connection`. This is the part which I think doesn't belong to the RPC client. Some time later in a rewrite.

backend/accounts/account.go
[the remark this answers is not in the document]
benma: Until now, offline meant the connection could not be established or was lost, but the problem could resolve itself if reconnection succeeds.

FatalError was an unexpected/unhandled error that the app doesn't know how to handle. 

At this point though it seems like the two could maybe be unified into one. Another PR?

x1ddos:
Yeah, maybe somewhere in the future.

util/jsonrpc/jsonrpc.go
benma: Yep I noticed this. If the negotiation fails, the connection is also not closed and the read loop runs forever...

Some day! :pray:

util/jsonrpc/jsonrpc.go
benma: Good catch. Fixed

backend/accounts/baseaccount.go
```diff
@@ -117,13 +117,13 @@ func (account *BaseAccount) ResetSynced() {
...
 	return account.offline
 }
 
 // SetOffline sets the account offline status and emits the EventStatusChanged() if the status
 // changed.
-func (account *BaseAccount) SetOffline(offline bool) {
+func (account *BaseAccount) SetOffline(offline error) {
 	wasOffline := account.offline
 	account.offline = offline
 	if wasOffline != offline {
```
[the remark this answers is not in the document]
benma: I removed the guard - the notification triggers a re-render in the frontend with the new offline error, but it does not really hurt if it happens for the same error again.

PR #1201, opened 2021-03-09
frontend/account: remove invisible offline status

benma:
The spinner takes up the whole screen. Afaik the status removed in
this commit is never visible.


thisconnect (approved):
LGTM

PR #1202, opened 2021-03-09
frontend/account: if offline and tor proxy enabled, give user a hint

benma:
Sometimes users activate the setting by mistake, or activate it but
don't have a Tor proxy running.

### DIFF
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -246,6 +246,7 @@ class Account extends Component<Props, State> {
             t,
             code,
             accounts,
+            config,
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/account.tsx
```diff
@@ -272,6 +273,15 @@ class Account extends Component<Props, State> {
                 })
             ) : '';
 
+        let offlineErrorTextLines = [];
```
thisconnect, line 276: The linter wants `const` here, const because the reference to the array never changes.

benma: i `const`antly fall into this issue ;)

PR #1209, opened 2021-03-12
backend/electrum: set ServerName to help SNI-capable loadbalancers

x1ddos:
The commit makes the electrum client provide SNI server name when
connecting to a node over TLS. It is very useful for a setup with
multiple nodes running on the same machine where a loadbalancer
dispatches connections to a specific node based on the SNI during a TLS
handshake. For example:

    btc1.shiftcrypto.io:443  -> local port 50001
    ltc1.shiftcrypto.io:443  -> same machine, local port 50002
    tbtc1.shiftcrypto.io:443 -> same machine, local port 51001

I don't expect any breaking changes. The ServerName field merely
provides an indication to the server, typically for a certificate
selection in a multi cert setup. Our certificate validation remains
the same.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -1109,7 +1109,7 @@ func (backend *Backend) RatesUpdater() *rates.RateUpdater {
 
 // DownloadCert downloads the first element of the remote certificate chain.
 func (backend *Backend) DownloadCert(server string) (string, error) {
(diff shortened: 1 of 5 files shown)


x1ddos:
Sorry, forgot to remove remnants of an older test.
Should be good now.

benma (approved):

PR #1210, opened 2021-03-16
backend/config: migrate node addresses from 5xxxx ports port 443

x1ddos:
We've received a number of complaints about our nodes being
inaccessible on some systems. Most resolve by allowing an outbound
connection in a firewall or similar.
This commit tries to reduce this issue to a minimum by abusing HTTPS
and making the app connect via port 443 which these firewalls are
expected to be more lenient with.

These are the same nodes originally migrated to in commit 309ea92,
just behind port 443. On the infrastructure side, the routing is done
by matching the server name during TLS SNI handshake and
reverse-proxying TCP connection to the specific 5xxxx port on the same
host.

This of course requires client cooperation where it sends appropriate
server name in TLS SNI. Done in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1209.

### DIFF
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -168,12 +168,12 @@ func NewDefaultAppConfig() AppConfig {
 			BTC: btcCoinConfig{
 				ElectrumServers: []*ServerInfo{
 					{
-						Server:  "btc1.shiftcrypto.io:50001",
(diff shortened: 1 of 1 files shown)


benma (approved):
utACK

PR #1211, opened 2021-03-16
frontend: hide disabled copy button on receive view

thisconnect:
There is no point if showing a disabled copy button, also the real
copy button is currently displayed in a different place (dialog).

Closes https://github.com/digitalbitbox/bitbox-wallet-app/issues/1001

PR #1212, opened 2021-03-16
Frontend: update copy component state init, var and type naming

thisconnect:
Moved state initialization out of constructor and use readonly
state initialization. We already do that in various places, reason
is less boilerplate and it tells TypeScript that the state key
shall not be mutated, example this.state.foo = 'bar'.
The linter should already know that this.state should be readonly
via eslint-plugin-react.

A while ago the copy component's input element has been changed
to textarea to support multiple lines, but the variable names and
types where still using input as name and type.
Changed variable names and types to use textarea, this should not
have a practical difference.

on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1211

### DIFF
--- a/frontends/web/src/components/copy/Copy.tsx
+++ b/frontends/web/src/components/copy/Copy.tsx
@@ -35,14 +35,11 @@ interface State {
 }
 
 class CopyableInput extends Component<Props, State> {
-    private inputField!: HTMLInputElement;
-
-    constructor(props: Props) {
-        super(props);
-        this.state = {
-            success: false,
-        };
+    public readonly state: State = {
+        success: false,
     }
+
+    private textArea!: HTMLTextAreaElement;
 
     public componentDidMount() {
         this.setHeight();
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/copy/Copy.tsx
```diff
@@ -35,15 +35,12 @@ interface State {
...
-    constructor(props: Props) {
-        super(props);
-        this.state = {
-            success: false,
-        };
+    public readonly state: State = {
+        success: false,
     }
 
+    private textArea!: HTMLTextAreaElement;
```
x1ddos, line 42: What's the difference? I mean, why is textarea better than input here. Would've been nice if the commit msg explained it.

x1ddos (approved):
> move state initialization out of constructor

I see that the constructor is actually removed but it always calls `super(props)`. Are there any potentially negative effects due to the missing `super(props)`? Would've been nice if the commit msg explained this too.

x1ddos:
> Would've been nice if the commit msg explained this too.

The most important thing I look in a commit msg is the "why". The "how" 
is typically clear from "git diff".

x1ddos:
... and I expect to find out the "what" in the commit title.
to sum it up roughtly:

1. what: commmit title
2. why and a higher level "how" if complicated: commit body
3. how: git diff

frontends/web/src/components/copy/Copy.tsx
thisconnect: The copy component is already using a textarea instead of an input element, this was changed in https://github.com/digitalbitbox/bitbox-wallet-app/commit/ce2fdd7ea4064b15fdb6dd6b93ffdd4b9e8206a5#diff-b5a510b0817e11affa3032785dc7cf269fce7aee513f5ea3f73a2f9a11a118fb
in order to support multiple lines, but all the variable names and types are still using input.

Updated the commit message.

thisconnect:
> I see that the constructor is actually removed but it always calls super(props). Are there any potentially negative effects due to the missing super(props)? Would've been nice if the commit msg explained this too.

`super(props)` has to be called because it extendes (p)react class' constructor, in that case you have to and this only to set the initial state. I just found out that in never react versions initializing the state this way errors.

Once we have hooks (in preact X) we don't have to use `class` anymore and don't have to extend from preact component class.

PR #1214, opened 2021-03-16
windows: kill any running instance of bitbox.exe during the install

x1ddos:
If a user already has BitBoxApp installed and it's currently running,
the new installation won't be able to proceed because the bitbox.exe is
not writable, assuming the same install directory is chosen.

This commit forces any running instance of bitbox.exe to quit before
doing anything else.

The taskkill command should be available on at least Windows 7 and on.
Docs: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill


benma (approved):
utACK

PR #1215, opened 2021-03-16
frontend: show fee on transaction overview

thisconnect:
The total balance should be incoming - outgoing tx, but
we currently do not show outgoing fees and especially
with small amounts it visible that fees are missing.
Fees of incoming tx do not matter as they go to the
miner.

- added fee info for send and sendtoself tx
- moved Fiat to the last numeric column
- vertically reduced space between cells in responsive view
- hide pending/complete label text on small screen
- removed UPPERCASE table header styling for more consistent look

### DIFF
--- a/frontends/web/src/components/transactions/transaction.css
+++ b/frontends/web/src/components/transactions/transaction.css
@@ -223,7 +223,7 @@
 
 @media (max-width: 1080px), (min-width: 1200px) and (max-width: 1322px) {
     .container {
-        margin: 0 0 var(--space-half) 0;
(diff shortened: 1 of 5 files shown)


thisconnect:
Still discussing another change with @jadzeidan

x1ddos:
So, have you guys finished discussing or is it this ongoing, or onhold? :)

jadzeidan:
> So, have you guys finished discussing or is it this ongoing, or onhold? :)

Oh hold I would say. We have a couple of concepts but haven't decided on one yet.

PR #1216, opened 2021-03-17
frontend-new-api -> master

benma:
So we can proceed with RC1.

https://github.com/digitalbitbox/bitbox-wallet-app/pull/1194 can simply change the target branch after (can be done in the GH UI)


thisconnect (approved):
LGTM

PR #1221, opened 2021-03-23
frontend: rewrite wait-dialog to typescript

thisconnect:
This potentially fixes TypeError: Cannot read property 'classList'
of null, in rare cases during wait-dialog's animation. The error
did not seem to have any other impact and app was fully functional.

### DIFF
--- a/frontends/web/src/components/devices/bitbox02/reset.tsx
+++ b/frontends/web/src/components/devices/bitbox02/reset.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,7 +23,7 @@ import { Dialog } from '../../dialog/dialog';
 import * as dialogStyle from '../../dialog/dialog.css';
 import { Button, Checkbox } from '../../forms';
 import { SettingsButton } from '../../settingsButton/settingsButton';
-import WaitDialog from '../../wait-dialog/wait-dialog';
+import { WaitDialog } from '../../wait-dialog/wait-dialog';
 
 interface ResetProps {
     apiPrefix: string;
@@ -113,7 +114,6 @@ class Reset extends Component<Props, State> {
                 {
                     isConfirming && (
                         <WaitDialog
-                            active={isConfirming}
                             title={t('reset.title')} >
                             {t('bitbox02Interact.followInstructions')}
                         </WaitDialog>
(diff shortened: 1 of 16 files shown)


thisconnect:
Tested BB02:
- factory reset
- restore backup
- create backup
- rename backup
- show recovery words (only started the process, didn't verify the words)
- devdevide -> bootloader currently errors 
- send2self btc

frontends/web/src/components/devices/bitbox02/reset.tsx
```diff
@@ -113,7 +114,6 @@ class Reset extends Component<Props, State> {
                 {
                     isConfirming && (
                         <WaitDialog
-                            active={isConfirming}
```
benma, line 116: typescript FTW

frontends/web/src/components/wait-dialog/wait-dialog.tsx
```diff
@@ -14,66 +15,94 @@
...
 import reject from '../../assets/icons/tap.png';
 import { animate } from '../../utils/animation';
 import * as style from '../dialog/dialog.css';
 
-@translate()
-export default class WaitDialog extends Component {
-    state = {
+interface WaitDialogProps {
+    includeDefault?: boolean;
+    prequel?: JSX.Element | false;
```
benma, line 27: why `| false`?

`undefined` should be enough?

benma (approved):
utACK, with a small improvement

frontends/web/src/components/wait-dialog/wait-dialog.tsx
thisconnect: TS said it can be false :)
but looking into it more closely:

The only place that `prequel` is used is here https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/routes/account/send/send.tsx#L747

The variable is set here https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/routes/account/send/send.tsx#L575
and if `signProgress.steps > 1` is 1 or less the variable will be `false`.
`true` wouldn't make sense, but I'd expect `false` to do the same as undefined just falsy :).

Should I rewrite send.tsx a bit, so it would be either a JSX.Element or undefined but not false?

Tho I'd prefer in this case `prequel?: JSX.Element | null;` so we could set the variable `condition ? (<span>) : null` 
 but that is a matter of taste we already use `undefined` in various places.

benma: I still think anything beyond `undefined | Element` is too many needless states. send.tsx can just set it to undefined if it is not used.

undefined vs null: no strong opinion, but undefined seems to be the convention in typescript.

PR #1225, opened 2021-03-29
backend/bitbox02bootloader: log firmware/signing key hashes

benma:
Easier debugging when e.g. a firmware update goes wrong.


x1ddos (approved):
LGTM

PR #1234, opened 2021-04-20
backend/device: remove commented/unused interace methods

benma:
They have been commented long enough.


x1ddos (approved):
I noticed the embedded `observable.Interface` is at the bottom - very hidden imho. I'd put at the top.

LGTM

PR #1235, opened 2021-04-20
backend/config: make persisted accounts compatible with unified accounts

benma:
With the unified accoutns feature added some time ago, one
account (e.g. "Bitcoin") can combine multiple configurations,
e.g. p2wsh and p2wsh-psh, each at a separate keypath.

We will use persisted accounts to manage multiple accounts and
watchonly accounts, so they must also be able to store multiple
configs per account.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -313,22 +313,26 @@ func (backend *Backend) createAndAddAccount(
 		if err != nil {
 			return err
 		}
-		if len(configurations) != 1 {
-			// TODO: unified-accounts
-			return errp.New("Watch-only accounts don't support mixed inputs yet")
-		}
-		configuration := configurations[0]
 		accountsConfig := backend.config.AccountsConfig()
 		for _, account := range accountsConfig.Accounts {
-			if account.Configuration.Hash() == configuration.Hash() && account.CoinCode == coin.Code() {
-				return errp.WithStack(ErrAccountAlreadyExists)
+			if account.CoinCode == coin.Code() {
+				// We detect a duplicate account (subaccount in a unified account) if any of the
+				// configurations is already present.
+				for _, config := range account.Configurations {
+					for _, config2 := range configurations {
+						if config.Hash() == config2.Hash() {
+							return errp.WithStack(ErrAccountAlreadyExists)
+						}
+					}
+				}
+
 			}
 		}
 		accountsConfig.Accounts = append(accountsConfig.Accounts, config.Account{
-			CoinCode:      coin.Code(),
-			Code:          code,
-			Name:          name,
-			Configuration: configuration,
+			CoinCode:       coin.Code(),
+			Code:           code,
+			Name:           name,
+			Configurations: configurations,
 		})
 		if err := backend.config.SetAccountsConfig(accountsConfig); err != nil {
 			return err
@@ -705,7 +709,7 @@ func (backend *Backend) initPersistedAccounts() {
 			continue
 		}
 		getSigningConfigurations := func() (signing.Configurations, error) {
-			return signing.Configurations{account.Configuration}, nil
+			return account.Configurations, nil
 		}
 		err = backend.createAndAddAccount(coin, account.Code, account.Name, getSigningConfigurations, false, false)
 		if err != nil {
--- a/backend/config/accounts.go
+++ b/backend/config/accounts.go
@@ -22,10 +22,10 @@ import (
 
 // Account holds information related to an account.
 type Account struct {
-	CoinCode      coin.Code              `json:"coinCode"`
-	Name          string                 `json:"name"`
-	Code          string                 `json:"code"`
-	Configuration *signing.Configuration `json:"configuration"`
+	CoinCode       coin.Code              `json:"coinCode"`
+	Name           string                 `json:"name"`
+	Code           string                 `json:"code"`
+	Configurations signing.Configurations `json:"configurations"`
 }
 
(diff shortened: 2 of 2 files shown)


backend/backend.go
```diff
@@ -313,22 +313,26 @@ func (backend *Backend) createAndAddAccount(
...
 		accountsConfig := backend.config.AccountsConfig()
 		for _, account := range accountsConfig.Accounts {
-			if account.Configuration.Hash() == configuration.Hash() && account.CoinCode == coin.Code() {
-				return errp.WithStack(ErrAccountAlreadyExists)
+			if account.CoinCode == coin.Code() {
+				// We detect a duplicate account (subaccount in a unified account) if any of the
+				// configurations is already present.
+				for _, config := range account.Configurations {
+					for _, config2 := range configurations {
+						if config.Hash() == config2.Hash() {
```
x1ddos, line 323: It still doesn't feel right that the Hash method is relying on json.Marshal. very easy to break: it's enough to change a field name in https://github.com/digitalbitbox/bitbox-wallet-app/blob/bdb9c97/backend/signing/configuration.go#L184-L190 and the whole thing will result in a different hash.

Consider reimplementing Hash without using json.

backend/config/accounts.go
```diff
@@ -22,10 +22,10 @@ import (
...
 // Account holds information related to an account.
 type Account struct {
-	CoinCode      coin.Code              `json:"coinCode"`
-	Name          string                 `json:"name"`
-	Code          string                 `json:"code"`
-	Configuration *signing.Configuration `json:"configuration"`
+	CoinCode       coin.Code              `json:"coinCode"`
+	Name           string                 `json:"name"`
+	Code           string                 `json:"code"`
+	Configurations signing.Configurations `json:"configurations"`
```
x1ddos, line 28: Did you consider using a map? Might make it easier with detecting existing config hashes:

```go
type Account struct {
  ...
  // SigningConfig is keyed by the configuration hash.
  SigningConfig map[string]signing.Configuration
}

// elsewhere
for _, conf := range configurations {
  if _, exist := account.SigningConfig[conf.Hash()]; exist {
    return ErrAccountAlreadyExists
  }
}
```

backend/backend.go
```diff
@@ -313,22 +313,26 @@ func (backend *Backend) createAndAddAccount(
 		if err != nil {
```
x1ddos, line 313: The comment is for the line above this. I'm always confused about getSigningConfigurations. Can it be a normal arg with type `signing.Configurations`?

backend/backend.go
```diff
@@ -313,22 +313,26 @@ func (backend *Backend) createAndAddAccount(
...
 		for _, account := range accountsConfig.Accounts {
-			if account.Configuration.Hash() == configuration.Hash() && account.CoinCode == coin.Code() {
-				return errp.WithStack(ErrAccountAlreadyExists)
+			if account.CoinCode == coin.Code() {
+				// We detect a duplicate account (subaccount in a unified account) if any of the
+				// configurations is already present.
+				for _, config := range account.Configurations {
+					for _, config2 := range configurations {
+						if config.Hash() == config2.Hash() {
+							return errp.WithStack(ErrAccountAlreadyExists)
```
x1ddos, line 324: What about when `persist` is false. Wouldn't you want to detect an existing account anyway?

backend/config/accounts.go
benma: I did but I don't see a benefit yet. Will do it it appears.

backend/backend.go
benma: This is coming in one of the next PRs. Here is the sneak peak of the commit message:

----
backend/accounts: load signing configurations immediately

The reason that the signing configurations were loaded lazily was that
it requires querying the xpubs from the device. With the BitBox01,
this is especially slow, but also the BitBox02 shows a small delay
when loading e.g. 3 xpubs.

Since we want to persist the accounts/configs anyway (to support
multiple accounts and watch-only), we need to get them
immediately. Since they will be persisted, this will only happens once
per keystore.

backend/backend.go
benma: possibly yes, but if so I can only do it later, since the config is still behind the `getSigningConfigurations()` function which we don't call otherwise. But in the next PRs this function will be gone and we can do it.

backend/config/accounts.go
x1ddos: Well, to me one immediate benefit is no nested loop. And in general, if there's a lookup-by-signing-config algorithm anywhere, it's just easier with a map instead of constantly looping over configurations slice.

x1ddos (approved):
LGTM except I still recommend adding signing configs to account as a map instead of the slice.

backend/config/accounts.go
x1ddos: and of course it's nicer to read compared to:

image

benma:
> LGTM except I still recommend adding signing configs to account as a map instead of the slice.

Currently the change would go quite deep - the accounts assume an ordered array too, with the first config being the default when making a receive address.

I will change it if the need arises.

x1ddos:
>> LGTM except I still recommend adding signing configs to account as a map instead of the slice.
> 
> Currently the change would go quite deep - the accounts assume an ordered array too, with the first config being the default when making a receive address.

Ah well that makes a big difference. Didn't realize there was a specific 
order. In this case, would've really appreciated a doc comment on the 
Configurations field of Account explaining it.

PR #1243, opened 2021-04-26
backend/accounts: load signing configurations immediately

benma:
The reason that the signing configurations were loaded lazily was that
it requires querying the xpubs from the device. With the BitBox01,
this is especially slow, but also the BitBox02 shows a small delay
when loading e.g. 3 xpubs.

Since we want to persist the accounts/configs anyway (to support
multiple accounts and watch-only), we need to get them
immediately. Since they will be persisted, this will only happens once
per keystore.


x1ddos (approved):
LGTM and I know tests are coming :)

PR #1244, opened 2021-04-26
backend: move split/unified accounts logic

benma:
The accounts are always stored in the unified format (multiple signign
configurations per account) in the accounts.json database. When
loading them, we split them into individual accounts if needed.

This previously happened when adding the BTC accounts when a keystore
is registered, but this will not work anymore as we will persist all
accounts and not load them dynamically after a keystore is registered
anymore. Instead, we will persist the default accounts when a new
keystore is registered.

createAndAddBTCAccount() and createAndAddETHAccount() will be changed
to persist the default accounts instead of loading them dynamically.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -430,27 +430,20 @@ func (backend *Backend) createAndAddBTCAccount(
 	}
 	log.Info("init account")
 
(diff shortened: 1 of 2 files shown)


x1ddos (approved):
LGTM

PR #1245, opened 2021-04-26
backend: persist default set of accounts for a new keystore 

benma:
Before, we would load them dynamically whenever a keystore is
connected.

Now, we only load accounts from the accounts.json database, and
pre-populate the database with a default set of accounts to kick
things off.

We take into account the global settings for which coins are active,
so that users who already disabled some accounts have a smooth
migration. The global settings to activate/deactivate coins and tokens
will be removed and replaced with functionality to add and remove
accounts per keystore.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -328,6 +328,7 @@ func (backend *Backend) createAndAddAccount(
 	code string,
 	name string,
 	signingConfigurations signing.Configurations,
+	activeTokens []string,
 ) {
 	var account accounts.Interface
 	accountConfig := &accounts.AccountConfig{
@@ -361,6 +362,16 @@ func (backend *Backend) createAndAddAccount(
 	case *eth.Coin:
 		account = eth.NewAccount(accountConfig, specificCoin, backend.log)
 		backend.addAccount(account)
+
+		// Load ERC20 tokens enabled with this Ethereum account.
+		for _, erc20TokenCode := range activeTokens {
+			token, err := backend.Coin(coinpkg.Code(erc20TokenCode))
+			if err != nil {
+				backend.log.WithError(err).Error("could not find ERC20 token")
+				continue
+			}
+			backend.createAndAddAccount(token, erc20TokenCode, token.Name(), signingConfigurations, nil)
+		}
 	default:
 		panic("unknown coin type")
 	}
@@ -506,6 +517,18 @@ func (backend *Backend) persistETHAccountConfig(
 			extendedPublicKey,
 		),
 	}
+	// In the past, ERC20 tokens were configured to be active or inactive globally, now they are
+	// active/inactive per ETH account. We use the previous global settings to decide the default
+	// set of active tokens, for a smoother migration for the user.
+	var activeTokens []string
+	if coin.Code() == coinpkg.CodeETH {
+		for _, tokenCode := range backend.config.AppConfig().Backend.ETH.ActiveERC20Tokens {
+			prefix := "eth-erc20-"
+			// Old config entries did not contain this prefix, but the token codes in the new config
+			// do, to match the codes listed in erc20.go
+			activeTokens = append(activeTokens, prefix+tokenCode)
+		}
+	}
 
 	rootFingerprint, err := keystore.RootFingerprint()
 	if err != nil {
@@ -519,6 +542,7 @@ func (backend *Backend) persistETHAccountConfig(
 		SupportsUnifiedAccounts: keystore.SupportsUnifiedAccounts(),
 		RootFingerprint:         rootFingerprint,
 		Configurations:          signingConfigurations,
+		ActiveTokens:            activeTokens,
 	})
 	if errp.Cause(err) == ErrAccountAlreadyExists {
 		// This is adding initial default accounts. If this already happened, there is nothing else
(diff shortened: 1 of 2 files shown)


backend/backend.go
```diff
@@ -446,25 +457,39 @@ func (backend *Backend) createAndAddBTCAccount(
...
+		CoinCode:                coin.Code(),
+		Name:                    name,
+		Code:                    code,
+		SupportsUnifiedAccounts: keystore.SupportsUnifiedAccounts(),
+		RootFingerprint:         rootFingerprint,
+		Configurations:          signingConfigurations,
+	})
+
+	if errp.Cause(err) == ErrAccountAlreadyExists {
+		// This is adding initial default accounts. If this already happened, there is nothing else
```
benma, line 475: This might be temporary - it is not checking if the accounts exist by using the fingerprint yet, but fetches and uses the xpubs instead.

backend/config/accounts.go
```diff
@@ -37,6 +37,21 @@ type Account struct {
...
 	Configurations  signing.Configurations `json:"configurations"`
+	// ActiveTokens list the tokens that should be loaded along with the account.  Currently, this
+	// only applies to ETH, and the elements are ERC20 token codes (e.g. "eth-erc20-usdt",
+	// "eth-erc20-bat", etc).
+	ActiveTokens []string `json:"activeTokens"`
+}
+
+// TokenActive returns true if this token is configured to be active.
+// code is the token id, e.g. "eth-erc20-usdt".
+func (account Account) TokenActive(code string) bool {
```
x1ddos, line 48: I can't find where this is used. Maybe somewhere later? If so, I'd add it in the next commits.

backend/config/accounts.go
```diff
@@ -37,6 +37,21 @@ type Account struct {
...
+	for _, tokenCode := range account.ActiveTokens {
```
x1ddos, line 49: I know we have only a few tokens at the moment (8?) but in the future, this might be a little performance drain, let's say 100 tokens, also depending on the usage of `TokenActive` which I can't find.

Consider at least sorting the slice when saving so that the search is O(log n) instead of O(n), better yet use a map which is about O(1)? If the idea is for the frontend to call `TokenActive` for every known token, it'll turn into O(n^2). 

I wouldn't bother and suggest optimizing when the time comes but this is a persistent configuration. If you don't do it now, it'll be more work later due to another config migration.

backend/config/accounts.go
```diff
@@ -30,8 +30,13 @@ type Account struct {
 	//
 	// This is used to unify multiple Bitcoin script types (p2wsh, p2wsh-p2sh) in one account. The
 	// keystore must be able to sign transactions with mixed inputs.
-	SupportsUnifiedAccounts bool                   `json:"supportsUnifiedaccounts"`
-	Configurations          signing.Configurations `json:"configurations"`
+	SupportsUnifiedAccounts bool `json:"supportsUnifiedaccounts"`
```
x1ddos, line 33: I wonder why add this to the persistent config. By the time a keystore is connected/opened, which root fingerprint is how its accounts are identified by, you should be able to tell whether unified accounts are supported by calling `keystore.SupportsUnifiedAccounts()`.

I see a challenge in `Backend.initPersistedAccounts()` because it doesn't actually get a hold on the active keystore and probably needs some refactoring. Not sure what I'm missing but it seems to me unnecessary to persist something that is easly known at runtime?

x1ddos (commented):
> Before, ERC20 token activation was globla

sounds like "blah blah" :)

backend/config/accounts.go
benma: This is with lookahead  to watch-only. Then, the connecting a keystore will become optional (resp. only required to receive/send). So we need to persist all info from the keystore that we need.

x1ddos: But from what I understand by looking around, "unified accounts" support is needed only during signing, i.e. a connected keystore. Why would you care about it in watch-only mode, only for UI?

backend/config/accounts.go
benma: Ah right, I used it in some previous iteration, so I can delete it.

backend/config/accounts.go
benma: Even 100 is instant / not noticeable? 

If you think of many calls to this in a loop could be slow, then we can also sort it before the loop (or we sort it when loading the json) and get log(n) access anyway, without worrying about it when storing it. And what if someone edits the config? Then we also need to sort it at load time anyway to validate it...

But since this function will be deleted anyway (unused), there is nothing to do anyway, right?

backend/config/accounts.go
x1ddos: > (resp. only required to receive/send)

resp?

backend/config/accounts.go
x1ddos: Yeah, nothing to do if removed. I was commenting preemptively :)

backend/config/accounts.go
benma: Well yes, there is no other way really. If you see the accounts as split with the keystore connected, you expect to see them the same way in watch-only.

Besides, if it is loaded and shown as a unified account, the balance is unified and the inputs are mixed when sending. I don't see a way to undo that only when sending without extreme confusion.

x1ddos: :+1:

benma: > resp?

respectively

backend/config/accounts.go
benma: done

x1ddos (approved):
LGTM but the linter is slightly unhappy

benma:
> LGTM but the linter is slightly unhappy

Fixed in a new commit. I wonder why this popped up now - the affected code was not modified in this PR :confused: 

Is the CI not actually fixing the version of golang-ci?

x1ddos:
The CI is just running "scripts/travis-ci.sh ci" in a container:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/0b03df0/.github/workflows/ci.yml#L41

But I saw that lint error since the beginning of the PR. It's new code, 
right.

PR #1228, opened 2021-04-06
Frontend Preact X

thisconnect:
- upgraded Preact 8 to X
- switched to popular [enzyme](https://www.npmjs.com/package/enzyme) testing utility
- refactored status.tsx
- updated i18next
- rewrote language.tsx with typescript 
- typescript fixes for Preact X
- fixed a bug in wait-dialog.tsx due to Preact children not always returning an arrays anymore

Note this changes a lot of the frontend dependencies, when testing this branch use `make buildweb` once to install all deps.

PR #1246, opened 2021-04-27
Frontend improvements from issues with Preact X 

thisconnect:
In Preact X branch there were a few issues, i.e. setState asynchronously updating `this.state` and subsequent code relying on an update `this.state` breaking. This PR has some of the changes in frontend-preact-x as a single PR:

- rewrite language component to typescript
- rewrite status component with hidden prop
- frontend: ensure dialog refs are present
- frontend: ensure state is up to date before continuing

See also old Preact X PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/1228

### DIFF
--- a/frontends/web/src/components/dialog/dialog.tsx
+++ b/frontends/web/src/components/dialog/dialog.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,18 +37,15 @@ interface State {
 }
 
 class Dialog extends Component<Props, State> {
-    private overlay!: HTMLElement;
-    private modal!: HTMLElement;
-    private modalContent!: HTMLElement;
+    private overlay?: HTMLDivElement | null;
+    private modal?: HTMLDivElement | null;
+    private modalContent?: HTMLDivElement | null;
     private focusableChildren!: NodeListOf<HTMLElement>;
 
-    constructor(props: Props) {
-        super(props);
-        this.state = {
-            active: false,
-            currentTab: 0,
-        };
-    }
+    public readonly state: State = {
+        active: false,
+        currentTab: 0,
+    };
 
     public componentDidMount() {
         setTimeout(this.activate, 10);
@@ -57,18 +55,6 @@ class Dialog extends Component<Props, State> {
         document.removeEventListener('keydown', this.handleKeyDown);
     }
 
-    private setOverlay = (element: HTMLDivElement) => {
-        this.overlay = element;
-    }
-
-    private setModal = (element: HTMLDivElement) => {
-        this.modal = element;
-    }
-
-    private setModalContent = (element: HTMLDivElement) => {
-        this.modalContent = element;
-    }
-
     private handleFocus = (e: FocusEvent) => {
         const input = e.target as HTMLElement;
         const index = input.getAttribute('index');
@@ -76,14 +62,16 @@ class Dialog extends Component<Props, State> {
     }
 
     private focusWithin = () => {
-        this.focusableChildren = this.modalContent.querySelectorAll('a, button, input, textarea');
-        const focusables = Array.from(this.focusableChildren);
-        for (const c of focusables) {
-            c.classList.add('tabbable');
-            c.setAttribute('index', focusables.indexOf(c).toString());
-            c.addEventListener('focus', this.handleFocus);
+        if (this.modalContent) {
+            this.focusableChildren = this.modalContent.querySelectorAll('a, button, input, textarea');
+            const focusables = Array.from(this.focusableChildren);
+            for (const c of focusables) {
+                c.classList.add('tabbable');
+                c.setAttribute('index', focusables.indexOf(c).toString());
+                c.addEventListener('focus', this.handleFocus);
+            }
+            document.addEventListener('keydown', this.handleKeyDown);
         }
-        document.addEventListener('keydown', this.handleKeyDown);
     }
 
     private focusFirst = () => {
@@ -128,7 +116,13 @@ class Dialog extends Component<Props, State> {
     }
 
     private deactivate = () => {
+        if (!this.modal || !this.overlay) {
+            return;
+        }
         animate(this.modal, 'fadeOutUp', () => {
+            if (!this.modal) {
+                return;
+            }
             this.modal.classList.remove(style.activeModal);
             this.setState({ active: false, currentTab: 0 }, () => {
                 document.removeEventListener('keydown', this.handleKeyDown);
@@ -138,16 +132,29 @@ class Dialog extends Component<Props, State> {
             });
         });
         animate(this.overlay, 'fadeOut', () => {
+            if (!this.overlay) {
+                return;
+            }
             this.overlay.classList.remove(style.activeOverlay);
         });
+
     }
 
     private activate = () => {
         this.setState({ active: true }, () => {
+            if (!this.modal || !this.overlay) {
+                return;
+            }
             animate(this.overlay, 'fadeIn', () => {
+                if (!this.overlay) {
+                    return;
+                }
                 this.overlay.classList.add(style.activeOverlay);
             });
             animate(this.modal, 'fadeInUp', () => {
+                if (!this.modal) {
+                    return;
+                }
                 this.modal.classList.add(style.activeModal);
                 this.focusWithin();
                 this.focusFirst();
@@ -175,26 +182,28 @@ class Dialog extends Component<Props, State> {
         const isSlim = slim ? style.slim : '';
         const isCentered = centered && !onClose ? style.centered : '';
         return (
-            <div className={[style.overlay].join(' ')} ref={this.setOverlay}>
-                <div className={[style.modal, isSmall, isMedium, isLarge].join(' ')} ref={this.setModal}>
+            <div className={style.overlay} ref={element => this.overlay = element}>
+                <div
+                    className={[style.modal, isSmall, isMedium, isLarge].join(' ')}
+                    ref={element => this.modal = element}>
                     {
                         title && (
                             <div className={[style.header, isCentered].join(' ')}>
                                 <h3 class={style.title}>{title}</h3>
-                                {
-                                    onClose && (
-                                        <button className={style.closeButton} onClick={this.deactivate} disabled={disabledClose}>
-                                            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round">
-                                                <line x1="18" y1="6" x2="6" y2="18"></line>
-                                                <line x1="6" y1="6" x2="18" y2="18"></line>
-                                            </svg>
-                                        </button>
-                                    )
-                                }
+                                { onClose ? (
+                                    <button className={style.closeButton} onClick={this.deactivate} disabled={disabledClose}>
+                                        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round">
+                                            <line x1="18" y1="6" x2="6" y2="18"></line>
+                                            <line x1="6" y1="6" x2="18" y2="18"></line>
+                                        </svg>
+                                    </button>
+                                ) : null }
                             </div>
                         )
                     }
-                    <div className={[style.contentContainer, isSlim].join(' ')} ref={this.setModalContent}>
+                    <div
+                        className={[style.contentContainer, isSlim].join(' ')}
+                        ref={element => this.modalContent = element}>
                         <div className={style.content}>
                             {children}
                         </div>
--- a/frontends/web/src/components/language/language.tsx
+++ b/frontends/web/src/components/language/language.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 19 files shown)


frontends/web/src/routes/account/account.tsx
```diff
@@ -312,8 +317,12 @@ class Account extends Component<Props, State> {
                     {status.synced && this.dataLoaded() && isBitcoinBased(account.coinCode) && <HeadersSync coinCode={account.coinCode} />}
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
-                            <Status dismissable={`info-${code}`} type="info" className="m-bottom-default">
-                                {t(`account.info.${code}`, { defaultValue: '' })}
+                            <Status
+                                className="m-bottom-default"
+                                hidden={!coinCodeInfo}
```
x1ddos, line 322: So, this status is now hidden if a translation is missing? unless I don't understand what it does exactly:

```js
const coinCodeInfo: string = t(`account.info.${code}`, { defaultValue: '' });
```

My guess is `coinCodeInfo` is empty string if either a translation is missing or there's no such key. But how can it be that a key is missing and why should it impact the hidden prop of the status? I'm confused.

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -315,8 +315,8 @@ class Receive extends Component<Props, State> {
         return (
             <div class="contentWithGuide">
                 <div class="container">
-                    <Status type="warning">
-                        {paired === false && t('warning.receivePairing')}
+                    <Status type="warning" hidden={paired !== false}>
```
x1ddos, line 318: Isn't it simpler to write `hidden={paired}`? It's typescript, right. So, everything is typed?

Same in send.tsx.

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -175,26 +182,28 @@ class Dialog extends Component<Props, State> {
...
+                                        </svg>
+                                    </button>
+                                ) : null }
                             </div>
                         )
                     }
-                    <div className={[style.contentContainer, isSlim].join(' ')} ref={this.setModalContent}>
+                    <div
+                        className={[style.contentContainer, isSlim].join(' ')}
+                        ref={element => this.modalContent = element}>
```
x1ddos, line 206: Is this like known pattern to prefer inlined code in react rendering instead of functions? To me, `ref={this.setModalContent}` is subjectively better.

Or is there some kind of technical reason to do inline `ref={element => ... }`?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -379,11 +379,12 @@ class Send extends Component<Props, State> {
         } else if (target.id === 'amount') {
             this.convertToFiat(value);
         }
-        this.setState(prevState => ({
+        this.setState((prevState) => ({
```
x1ddos, line 382: Some times I see brackets like here in `(prevState)` but some other times I don't, like in a previous commit in `{element => this.elem = element}`? What's the difference? Shouldn't it always be in the same way for better style consistency?

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -36,18 +37,15 @@ interface State {
...
     private focusableChildren!: NodeListOf<HTMLElement>;
 
-    constructor(props: Props) {
-        super(props);
-        this.state = {
-            active: false,
-            currentTab: 0,
-        };
-    }
+    public readonly state: State = {
```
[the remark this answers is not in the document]
thisconnect: Yup, I believe the reason is that when extending from the (p)react base class component, that base class needs to be able to access it, we have that in many places.

Typescript will complain about the following when changing

Screen Shot 2021-04-27 at 10 25 12 AM

frontends/web/src/components/status/status.css
```diff
@@ -42,6 +42,7 @@
 }
 
 .close {
+    border: none;
```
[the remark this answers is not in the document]
thisconnect: I guess I should have just copied the original commit message

```
n Preact X props.children is not always an array anymore see
https://preactjs.com/guide/v10/upgrade-guide/#propschildren-is-not-always-an-array

For this reason the way status.tsx checked if it had children to
display broke and empty status messages are visible. Also various
ways of rendering message text lead to inconsistent code, as
conditionally rendering the status component or its children
with logical or ternary operator.

Added a simple hidden property that is opt-in and can be used to
hide the message and render null.

Also changed the dismiss close element from using anchor with
href=# to button, so it does not change the url#hash.
```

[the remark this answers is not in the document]
thisconnect: see last point, I'll update the commit message

frontends/web/src/components/status/status.tsx
```diff
@@ -91,35 +94,31 @@ export default class Status extends Component<Props, State> {
             show,
         }: State) {
         const childrenList = children as ComponentChild[];
-        if (!show || (childrenList && childrenList.length === 1 && !childrenList[0])) {
+        if (hidden || !show || (childrenList && childrenList.length === 1 && !childrenList[0])) {
```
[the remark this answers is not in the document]
thisconnect: Agree this should be written much simpler, the point now is that `(childrenList && childrenList.length === 1 && !childrenList[0])` isn't needed anymore, I'll remove that as well.

I am not yet sure about hidden and show. Hidden is an opt in feature to force hiding the component. I have to check show, this reads the config and can opt in hidding it seems…
Will update status.tsx…

frontends/web/src/routes/account/account.tsx
thisconnect: > Some situations the whole status component was wrapped in a
condition, others only the content of the status component was
in a condition. Now the condition can be passed in the hidden
prop.

I should update the commit description.

> So, this status is now hidden if a translation is missing? 

it already was doing that before, but by checking if the child is truthy else return null. I'll remove that logic as it is not needed anymore as it can be controlled with the hidden props now.

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: initially paired is undefined because there is an async check to fetch paired. The warning should only be shown if paired is false. If paired is true or undefined it should be hidden.

frontends/web/src/components/dialog/dialog.tsx
thisconnect: Same, I preferred having the function not defined in the render function.

> **Caveats with callback refs**
If the ref callback is defined as an inline function, it will get called twice during updates, first with null and then again with the DOM element. This is because a new instance of the function is created with each render, so React needs to clear the old ref and set up the new one. You can avoid this by defining the ref callback as a bound method on the class, but note that it shouldn’t matter in most cases.

https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

In the long term this is this will change anyway when we use hooks. For now I find is more readable, but can change back why do you find it subjectively better?

x1ddos: It's ok. So, there's a technical explanation for this. Just wanted to know. :+1:

thisconnect: It is in theory slightly less performant as `a new instance of the function is created with each render`, unless there is optimizing step in the build process, I am not aware of any but there could ofc be something.

In the future we can use useRef hook
https://reactjs.org/docs/hooks-reference.html#useref

frontends/web/src/components/status/status.tsx
```diff
@@ -80,46 +82,41 @@ export default class Status extends Component<Props, State> {
...
+        children,
+        className,
+        dismissable,
+        hidden,
+        type = 'warning',
+    }: RenderableProps<Props>,
+    {
+        show,
+    }: State) {
+        if (hidden || !show) {
```
thisconnect, line 95: @x1ddos I had to keep both `show` from state (updated after config is read) and `hidden` from prop, but got at least rid of all the children checking.

thisconnect:
Addressed most feedback, except still using show/hidden in status and updated commit messages.

PTAL

frontends/web/src/components/language/language.tsx
```diff
@@ -14,34 +15,60 @@
...
 import * as style from './language.css';
 
-@translate()
-export default class LanguageSwitcher extends Component {
+type TActiveLanguageCodes = 'bg' | 'de' | 'en' | 'es'
+    | 'fr' | 'hi' | 'it' | 'ja' | 'ms' | 'pt'
+    | 'ru' | 'sl' | 'tr' | 'zh';
+
+type TLanguage = {
+    code: TActiveLanguageCodes | string;
```
x1ddos, line 30: If code can also be a string, then what's the point of defining `TActiveLanguageCodes`. Can the `code` get assigned a random value somehow, that is not among the `TActiveLanguageCodes`?

x1ddos: Will this work:

```ts
code: TActiveLanguageCodes
```

(no `| string`)

frontends/web/src/routes/account/account.tsx
x1ddos: I still don't understand how this works:

```ts
const coinCodeInfo: string = t(`account.info.${code}`, { defaultValue: '' });
hidden={!coinCodeInfo}
```

Why would a translated string impact the `hidden` attribute value? It doesn't sound like the right condition to me.

frontends/web/src/components/language/language.tsx
thisconnect: Typescript will error in the tests

Screen Shot 2021-04-27 at 5 34 19 PM

https://github.com/digitalbitbox/bitbox-wallet-app/blob/0b03df07ea60ba491bd5eab52b1b43596c09284b/frontends/web/test/components/language/language.test.tsx#L31

But we could just ignore in the tests
```diff
diff --git a/frontends/web/src/components/language/language.tsx b/frontends/web/src/components/language/language.tsx
index 9335e3b7..b7e760fa 100644
--- a/frontends/web/src/components/language/language.tsx
+++ b/frontends/web/src/components/language/language.tsx
@@ -27,7 +27,7 @@ type TActiveLanguageCodes = 'bg' | 'de' | 'en' | 'es'
     | 'ru' | 'sl' | 'tr' | 'zh';
 
 type TLanguage = {
-    code: TActiveLanguageCodes | string;
+    code: TActiveLanguageCodes;
     display: string;
 };
 
diff --git a/frontends/web/test/components/language/language.test.tsx b/frontends/web/test/components/language/language.test.tsx
index 02d0245a..239c3637 100644
--- a/frontends/web/test/components/language/language.test.tsx
+++ b/frontends/web/test/components/language/language.test.tsx
@@ -36,6 +36,7 @@ describe('components/language/language', () => {
             it(`returns exact match (${lang.code})`, (done) => {
                 i18n.changeLanguage(lang.code, (err) => {
                     expect(err).toBe(null);
+                    /* @ts-ignore */
                     const ctx = deep(<LanguageSwitch i18n={i18n} languages={supportedLangs} />);
                     expect(ctx.state('selectedIndex')).toEqual(idx);
                     done();
@@ -46,6 +47,7 @@ describe('components/language/language', () => {
         it('matches main language tag', (done) => {
             i18n.changeLanguage('pt_BR', (err) => {
                 expect(err).toBe(null);
+                /* @ts-ignore */
                 const ctx = deep(<LanguageSwitch i18n={i18n} languages={supportedLangs} />);
                 expect(ctx.state('selectedIndex')).toEqual(2); // 'pt'
                 done();
@@ -55,6 +57,7 @@ describe('components/language/language', () => {
         it('returns default if none matched', (done) => {
             i18n.changeLanguage('it', (err) => {
                 expect(err).toBe(null);
+                /* @ts-ignore */
                 const ctx = deep(<LanguageSwitch i18n={i18n} languages={supportedLangs} />);
                 expect(ctx.state('selectedIndex')).toEqual(0); // 'en'
                 done();
```

frontends/web/src/routes/account/account.tsx
thisconnect: I keep the logic, before this change the condition was it rendered null if children are falsy `!childrenList[0]` so for example an empty string. With `!childrenList[0]` here, (this is also what broke in PreactX and proably will in React too)

https://github.com/digitalbitbox/bitbox-wallet-app/blob/0b03df07ea60ba491bd5eab52b1b43596c09284b/frontends/web/src/components/status/status.tsx#L94

thisconnect: > Why would a translated string impact the hidden attribute value? It doesn't sound like the right condition to me.

I agree, but need to investigate…

thisconnect: Ok this is pretty old and only used when the legacy setting split accounts is enabled.

The info is only shown on the different Bitcoin address types, if the translation is missing (and defaults to an empty string) then no status component is rendered.

https://github.com/digitalbitbox/bitbox-wallet-app/blob/0b03df07ea60ba491bd5eab52b1b43596c09284b/frontends/web/src/locales/en/app.json#L8-L15

frontends/web/src/components/language/language.tsx
x1ddos: I think it's perfectly fine to ts-ignore in tests.

frontends/web/src/routes/account/account.tsx
x1ddos: Don't you find this logic a bit concerning? Is it possible to change to something more reasonable like this:

```ts
hidden={legacySplitIsEnabled}
```

I'm sure frontend can get this setting from somewhere.

thisconnect: added  WithCoinTypeInfo 
```js
const WithCoinTypeInfo = [
    'btc-p2pkh',
    'btc-p2wpkh',

];
```
squashed force pushed in https://github.com/digitalbitbox/bitbox-wallet-app/compare/6108003c9fdb5b3e4bbbe0d8a9f668c07d4ac366..4663377c71c832cb9c7a73fa3733b92186e730fe

also addressed the other comment, which I cannot comment on anymore.

x1ddos: :pray: but please add a short comment to explain what `WithCoinTypeInfo` is for.

x1ddos (approved):
LGTM

thisconnect:
Ok tested one more time:
- [x] with dev device: devices attestation failed warning 
- [x] sdCard plugged in warning
- [x] dismissable "Native Segwit Bitcoin" info
- [x] BB01 EOL info
- [x] BB01 Please pair the BitBox

PR #1248, opened 2021-04-27
backend: remove multisig

benma:
It was a "hacky" 2-of-2 multisig if two bitboxes are connected, which
was used to play with multisig in the past. The multisig argument to
enable this mode has been long removed. The cosigner index concept was
also wrong, as the cosigner index is not a property of the keystore,
but a property of the signing configuration (and could be infered
there from the xpubs that are present).

### DIFF
--- a/backend/devices/bitbox/device.go
+++ b/backend/devices/bitbox/device.go
@@ -1373,16 +1373,13 @@ func (dbb *Device) ExtendedPublicKey(keypath signing.AbsoluteKeypath) (*hdkeycha
 }
 
 // KeystoreForConfiguration implements device.Interface.
-func (dbb *Device) KeystoreForConfiguration(
-	cosignerIndex int,
-) keystoreInterface.Keystore {
+func (dbb *Device) KeystoreForConfiguration() keystoreInterface.Keystore {
 	if dbb.Status() != StatusSeeded {
 		return nil
 	}
 	return &keystore{
-		dbb:           dbb,
-		cosignerIndex: cosignerIndex,
-		log:           dbb.log,
+		dbb: dbb,
+		log: dbb.log,
 	}
 }
 
--- a/backend/devices/bitbox/device.go
+++ b/backend/devices/bitbox/device.go
@@ -1372,8 +1372,8 @@ func (dbb *Device) ExtendedPublicKey(keypath signing.AbsoluteKeypath) (*hdkeycha
 	return dbb.xpub(keypath.Encode())
 }
 
-// KeystoreForConfiguration implements device.Interface.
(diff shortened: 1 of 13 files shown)


backend/devices/bitbox/device.go
```diff
@@ -1373,16 +1373,13 @@ func (dbb *Device) ExtendedPublicKey(keypath signing.AbsoluteKeypath) (*hdkeycha
 }
 
 // KeystoreForConfiguration implements device.Interface.
-func (dbb *Device) KeystoreForConfiguration(
-	cosignerIndex int,
-) keystoreInterface.Keystore {
+func (dbb *Device) KeystoreForConfiguration() keystoreInterface.Keystore {
```
x1ddos, line 1376: The method name makes little sense now, tbh.

backend/devices/usb/manager.go
```diff
@@ -280,7 +276,7 @@ func (manager *Manager) listen() {
 				continue
 			}
 			// Skip if we already have another device registered and we only support one device.
-			if manager.onlyOne && len(manager.devices) != 0 {
+			if len(manager.devices) != 0 {
 				continue
```
x1ddos, line 280: I find this loop a bit useless. Would've done it the opposite way: run it only when `len(devices) == 0`.

backend/keystore/keystore.go
```diff
@@ -48,14 +48,10 @@ type Keystore interface {
 	// https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#key-identifiers
 	RootFingerprint() ([]byte, error)
 
-	// CosignerIndex returns the index at which the keystore signs in a multisig configuration.
-	// The returned value is always zero for a singlesig configuration.
-	CosignerIndex() int
-
 	// SupportsAccount returns true if they keystore supports the given coin/account.
 	// meta is a coin-specific metadata related to the account type.
 	// If false, the backend will add one account per supported script type.
```
x1ddos, line 53: > If false, the backend will add one account per supported script type.

Not sure I understand the comment. What is `false`, the returned value? But then "backend will add ..." doesn't concern this `SupportsAccount` method at all. Or did it refer to the `multisig` bool?

x1ddos (approved):
LGTM

backend/keystore/keystore.go
benma: No, it means that if this function returns false, that the backend will split the accounts, one per script type. I know it does not concern this function, but thought it might be helpful. Can also delete it :shrug:

backend/devices/bitbox/device.go
benma: Renamed

PR #1249, opened 2021-04-27
backend: supported coins endpoint

benma:


benma:
@thisconnect this is the first endpoint in the frontend: you can use it for the first step of the 'add account' wizard to get the coins to show.

backend/devices/bitbox02/keystore.go
```diff
@@ -51,15 +51,14 @@ func (keystore *keystore) RootFingerprint() ([]byte, error) {
 	return keystore.device.RootFingerprint()
 }
 
-// SupportsAccount implements keystore.Keystore.
-func (keystore *keystore) SupportsAccount(coin coin.Coin, meta interface{}) bool {
+// SupportsCoin implements keystore.Keystore.
+func (keystore *keystore) SupportsCoin(coin coin.Coin) bool {
 	switch specificCoin := coin.(type) {
 	case *btc.Coin:
 		if (coin.Code() == coinpkg.CodeLTC || coin.Code() == coinpkg.CodeTLTC) && !keystore.device.SupportsLTC() {
```
x1ddos, line 58: This is confusing. I see `coin.Xxx` and `coinpkg.Xxx` but it's the same package, imported twice.

backend/devices/bitbox02/keystore.go
```diff
@@ -51,15 +51,14 @@ func (keystore *keystore) RootFingerprint() ([]byte, error) {
 	return keystore.device.RootFingerprint()
 }
 
-// SupportsAccount implements keystore.Keystore.
-func (keystore *keystore) SupportsAccount(coin coin.Coin, meta interface{}) bool {
+// SupportsCoin implements keystore.Keystore.
+func (keystore *keystore) SupportsCoin(coin coin.Coin) bool {
```
x1ddos, line 55: > We want to be able to list the coins available for a keystore in a dropdown

then it feels like a simpler way is to just return a slice of all supported coins?

I saw a new `Backend.AvailableCoins` which does return a slice. So, why not make this keystore method also return a slice?

```go
func (k *keystore) SupportedCoins() []coin.Coin
```

backend/accounts.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
 	coinpkg "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
```
x1ddos, line 19: `backend/coins/coin` package imported twice.

backend/handlers/handlers.go
```diff
@@ -806,6 +809,16 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
 	return handlers.backend.ChartData()
 }
 
+// getAvailableCoinsHandler returns an array of coin codes for which you can add an account.
```
x1ddos, line 812: You mean a slice, right? Also, who's "you"? :)

frontends/web/src/api/backend.ts
```diff
@@ -0,0 +1,22 @@
+/**
```
x1ddos, line 1: The filename `api/backend.ts` is strange. Are all API calls moving here at some point? Maybe @thisconnect knows more.

backend/devices/bitbox02/keystore.go
```diff
@@ -70,6 +69,20 @@ func (keystore *keystore) SupportsAccount(coin coin.Coin, meta interface{}) bool
...
 }
 
+// SupportsAccount implements keystore.Keystore.
+func (keystore *keystore) SupportsAccount(coin coin.Coin, meta interface{}) bool {
+	if !keystore.SupportsCoin(coin) {
+		return false
+	}
+	switch coin.(type) {
+	case *btc.Coin:
+		scriptType := meta.(signing.ScriptType)
```
x1ddos, line 79: Isn't this going to panic if passed a different type? Wouldn't you want:

```go
typ, ok := meta.(signing.ScriptType)
```

If it's always `signing.ScriptType`, definitely consder changing the `meta` arg to a script type.

backend/devices/bitbox02/keystore.go
benma: Also possible. `AvailableCoins()` still needs to exist to filter testnet/regtest. Do you have a strong preference to change it?

backend/handlers/handlers.go
benma: Depends on your perspective, it delivers a json array :)

frontends/web/src/api/backend.ts
benma: I just put it here because it is direct backend/handlers.go endpoint with no namespace.

backend/devices/bitbox02/keystore.go
benma: It is only a ScriptType for BTC/LTC. I wish we had Rust enums (union types).

backend/devices/bitbox02/keystore.go
benma: oops, fixed

frontends/web/src/api/backend.ts
x1ddos: I see. So, the plan is to move here all the other API endpoints which are in backend/handlers.go?

backend/handlers/handlers.go
```diff
@@ -38,6 +38,7 @@ import (
 	baseHandlers "github.com/digitalbitbox/bitbox-wallet-app/backend/bitboxbase/handlers"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc"
 	accountHandlers "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/handlers"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
```
x1ddos, line 41: imported twice.

backend/devices/bitbox02/keystore.go
x1ddos: If it were me, would've never guessed it and needed to `grep SupportsAccount`, hoping to find all the references across the codebase. I guess we'll leave it for another day.

backend/devices/bitbox02/keystore.go
x1ddos: I have many preferences :)
It just felt natural to return a slice. Up to you.

But I've realized, I'd actually name Backend.AvailableCoins `SupportedCoins`. See the other comment in the frontend API.

x1ddos (approved):
LGTM but @thisconnect has something.

frontends/web/src/api/backend.ts
thisconnect: Maybe just `coins.ts` and `apiGet('coins')`?

frontends/web/src/api/backend.ts
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CoinCode } from './account';
+import { apiGet } from '../utils/request';
+
+export const getAvailableCoins = (): Promise<CoinCode[]> => {
+    return apiGet('available-coins');
```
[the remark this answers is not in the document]
thisconnect: If we have support for using multiple BB's at the same time, I guess frontend should pass deviceID ?

frontends/web/src/api/backend.ts
benma: one file per function is excessive?

frontends/web/src/api/backend.ts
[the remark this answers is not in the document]
benma: see the commit msg - we can support multiple devices at once and then we'd pass a deviceID (or keystoreID), but adding this too in the UI and backend is not needed now.

backend/devices/bitbox02/keystore.go
benma: done

frontends/web/src/api/backend.ts
thisconnect: just wanted to give some options, I didn't mean to suggest to write a file per function.

PR #1250, opened 2021-04-29
Frontend multiaccount UI

thisconnect:
Added the add account workflow with the select-coin, choose-name
and success steps. The select-coin step is hidden if there is only
one coin that can be added.

TODO not part of this PR:
- call add-account backend with coin and name
- cleanup i18n
- add manage account button in settings
- edit account name in account info view

### DIFF
--- a/frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
+++ b/frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2021 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { h, RenderableProps } from 'preact';
+import * as accountApi from '../../../../api/account';
+import { Select } from '../../../../components/forms';
+import { translate, TranslateProps } from '../../../../decorators/translate';
+
+// copied over from old addacconut.tsx
+const COIN_AND_ACCOUNT_CODES = {
+    'btc': {
+        name: 'Bitcoin',
+        coinCode: 'btc',
+    },
+    // TODO: what about those?
+    'btc-p2wpkh-p2sh': {
+        name: 'Bitcoin',
+        coinCode: 'btc',
+        scriptType: 'p2wpkh-p2sh',
+    },
+    'btc-p2wpkh': {
+        name: 'Bitcoin: bech32',
+        coinCode: 'btc',
+        scriptType: 'p2wpkh',
+    },
+    'btc-p2pkh': {
+        name: 'Bitcoin Legacy',
+        coinCode: 'btc',
+        scriptType: 'p2pkh',
+    },
+    'btc-addr': {
+        name: 'Bitcoin Address',
+        coinCode: 'btc',
+        scriptType: 'p2wpkh', // TODO dummy script type to pass DecodeScriptType
+    },
+    'ltc': {
+        name: 'Litecoin',
+        coinCode: 'ltc',
+    },
+    'ltc-p2wpkh-p2sh': {
+        name: 'Litecoin',
+        coinCode: 'ltc',
+        scriptType: 'p2wpkh-p2sh',
+    },
+    'ltc-p2wpkh': {
+        name: 'Litecoin: bech32',
+        coinCode: 'ltc',
+        scriptType: 'p2wpkh',
+    },
+    'ltc-addr': {
+        name: 'Litecoin Address',
+        coinCode: 'ltc',
+        scriptType: 'p2wpkh', // TODO dummy script type to pass DecodeScriptType
+    },
+    'eth': {
+        name: 'Ethereum',
+        coinCode: 'eth',
+        scriptType: 'p2wpkh',
+    },
+    // Testnet
+    'tbtc': {
+        name: 'Bitcoin Testnet',
+        coinCode: 'tbtc',
+    },
+    'tbtc-p2wpkh-p2sh': {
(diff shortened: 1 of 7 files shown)


thisconnect:
oups not sure why the other commits are here

PTAL ee01655 frontend: add account workflow prototype …

frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+ * limitations under the License.
+ */
+
+import { h, RenderableProps } from 'preact';
+import * as accountApi from '../../../../api/account';
+import { Select } from '../../../../components/forms';
+import { translate, TranslateProps } from '../../../../decorators/translate';
+
+// copied over from old addacconut.tsx
+const COIN_AND_ACCOUNT_CODES = {
```
benma, line 23: I will extend the 'supported-coins' endpoint to return not only the coincode, but also the name, then we can delete this code.

addaccount.tsx should probably be deleted completely, we will not be using it anymore, right?

frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+interface State {
+    accountName: string;
+    coinCode: 'choose' | accountApi.CoinCode;
+    onlyOneSupportedCoin: boolean;
+    step: 0 | 1 | 2;
+    supportedCoins: string[];
+}
+
+class ManageAccount extends Component<Props, State> {
```
benma, line 46: Why `Manage`? Would `Add` be more fitting?

benma (commented):
Looks nice!

Consider also that if there is only one coin to select (BB02 Bitcoin-only), the first step could be skipped. It would be strange fora btc-only user to 'select a coin'.

frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
benma: Done here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1253/

benma:
The account should probably not be added when you click 'Next'. The button should be renamed to 'Add account'?

After it is added, you shouldn't be able to go back.

frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -0,0 +1,221 @@
...
+                console.info(`${coinCode} selected`);
+                break;
+            case 'choose-name':
+                // TODO: post accountName and coinCode to backend
+                console.info(`add new account for ${coinCode} with the name: ${accountName}`);
+                break;
+            case 'success':
+                route('/account-summary');
+                return;
+                break;
```
benma, line 106: break not needed after return

frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -0,0 +1,221 @@
...
+        }
+        this.setState((state) => {
```
benma, line 108: why do you translate between `0 | 1 | 2` and `select-coin | choose-name | success` here and in getStep()? Maybe just use the logical names everywhere?

frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -0,0 +1,221 @@
...
+            switch (state.step) {
+                case 0: return ({ step: 1 });
+                case 1: return ({ step: 2 });
+                case 2: return ({ step: 2 });
```
benma, line 112: go from 2 to 2? :thinking:

frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -0,0 +1,221 @@
...
+    onlyOneSupportedCoin: boolean;
+    step: 0 | 1 | 2;
+    supportedCoins: string[];
+}
+
+class ManageAccount extends Component<Props, State> {
+    public readonly state: State = {
+        accountName: '',
+        coinCode: 'choose',
+        onlyOneSupportedCoin: false,
```
benma, line 50: This is strictly derived from `supportedCoins` below, so it's better to not have it in the state, but compute it on the fly.

benma (commented):
I made a PR building on top of this to add the backend part and improve some small things:

https://github.com/digitalbitbox/bitbox-wallet-app/pull/1258

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: We can use the same view to also edit and rename an existing account, so manage account is more appropriate imo

thisconnect:
> Consider also that if there is only one coin to select (BB02 Bitcoin-only), the first step could be skipped. It would be strange fora btc-only user to 'select a coin'.

already implemented see commit message, or was there something missing?

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: Yeah I know it is silly and not necessary, but else Typescript complains about `Not all code paths return a value.ts(7030)`

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: To skip the first step for BB02 Bitcoin-only but still be able to just add + 1 to the step state, but yeah I'll try to remove the step number completelly.

frontends/web/src/routes/account/manage/manage.tsx
benma: In https://github.com/digitalbitbox/bitbox-wallet-app/pull/1258 I saw that this state transition has to happen after a successful apiPost anyway, but: you could add a `default:` case that does nothing instead.

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: renamed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1277

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: Done in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1277

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: fixed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1277

thisconnect:
Closing this in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1277

PR #1251, opened 2021-05-01
backend: add accounts by number

benma:


backend/accounts.go
```diff
@@ -82,3 +88,72 @@ func (backend *Backend) SupportedCoins(keystore keystore.Keystore) []coinpkg.Cod
...
+		accountCode = fmt.Sprintf("%s-%d", coinCode, accountNumber)
+	}
+	log := backend.log.
+		WithField("accountCode", accountCode).
+		WithField("coinCode", coinCode).
+		WithField("accountNumber", accountNumber)
+
+	// This used to be a user-facing setting. Now we simply use it for migration to decide which
+	// coins to add by default. Only for the first account.
+	if accountNumber == 0 && !backend.config.AppConfig().Backend.CoinActive(coinCode) {
```
benma, line 114: just realized this will not work, e.g. if someone starts with only bitcoin, and then e.g. wants to add the first litecoin account. i will move this to the callsite when adding the initial default accounts.

backend/accounts.go
```diff
@@ -82,3 +88,72 @@ func (backend *Backend) SupportedCoins(keystore keystore.Keystore) []coinpkg.Cod
...
+// keystore supports. The keypaths will be standard BIP44 keypaths for the respective account types.
+func (backend *Backend) createAndPersistAccountConfig(coinCode coinpkg.Code, accountNumber uint16, keystore keystore.Keystore) error {
+	coin, err := backend.Coin(coinCode)
+	if err != nil {
+		return err
+	}
+
+	// The first account's code does not have a suffix, to keep the same code as before we introduced
+	// multiple accounts. This is to not invalidate the account cache the user had before.
+	accountCode := string(coinCode)
```
benma, line 103: this also doesn't work  - needs to be unique across keystores.

backend/signing/keypath.go
```diff
@@ -32,6 +32,9 @@ const (
 
 	// NonHardened denotes a non-hardened key derivation.
 	NonHardened = false
+
+	// HARDENED is the BIP44 offset to make a keypath element hardened.
+	HARDENED uint32 = hdkeychain.HardenedKeyStart
```
x1ddos, line 37: Why not use directly `hdkeychain.HardenedKeyStart`? I'd find it very confusing to see `signing.Hardened` and `signing.HARDENED`  - they look too similar.

backend/backend.go
```diff
@@ -760,66 +749,27 @@ func (backend *Backend) initPersistedAccounts() {
...
 		if backend.arguments.Regtest() {
-			RBTC, _ := backend.Coin(coinpkg.CodeRBTC)
-			backend.persistBTCAccountConfig(keystore, RBTC,
-				"rbtc",
-				[]scriptTypeWithKeypath{
-					newScriptTypeWithKeypath(signing.ScriptTypeP2WPKHP2SH, "m/49'/1'/0'"),
-					newScriptTypeWithKeypath(signing.ScriptTypeP2PKH, "m/44'/1'/0'"),
-				},
-			)
+			if err := backend.createAndPersistAccountConfig(coinpkg.CodeRBTC, 0, keystore); err != nil {
```
x1ddos, line 755: if you just `backend.createAndPersistAccountConfig(...)` it'll be simpler:

```go
// Testnets.
if backend.arguments.Testing() {
  if Regtest() {
    return backend.createAndPersistAccountConfig(...)
  }
  for _, coinCode := range []coinpkg.Code{coinpkg.CodeTBTC, coinpkg.CodeTLTC, coinpkg.CodeTETH, coinpkg.CodeRETH} {
    if err := backend.createAndPersistAccountConfig(coinCode, 0, keystore); err != nil {
        return err
    }
  }
  return nil
}

// Mainnets
for _, coinCode := range []coinpkg.Code{coinpkg.CodeBTC,
}
```

Basically less nesting > easier to read: least indented code is the "main" or "production".

x1ddos (approved):

PR #1253, opened 2021-05-04
backend/handlers: add more info to /supported-coins

benma:
Also show the display name and if an account can be added for this
coin, so the frontend can choose to disable the coin already upfront.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -38,7 +38,6 @@ import (
 	baseHandlers "github.com/digitalbitbox/bitbox-wallet-app/backend/bitboxbase/handlers"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc"
 	accountHandlers "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/handlers"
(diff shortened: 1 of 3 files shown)


backend/handlers/handlers.go
```diff
@@ -812,11 +812,29 @@ func (handlers *Handlers) getAccountSummary(_ *http.Request) (interface{}, error
...
 	if len(keystores) != 1 {
 		return []string{}, nil
 	}
-	return handlers.backend.SupportedCoins(keystores[0]), nil
+	keystore := keystores[0]
+	var result []element
+	for _, coinCode := range handlers.backend.SupportedCoins(keystore) {
+		coin, err := handlers.backend.Coin(coinCode)
+		if err != nil {
+			return nil, err
```
x1ddos, line 829: Just skip it? Seems less impactful: I'd still want to be able to add an account for any other supported coin.

x1ddos (approved):
LGTM

backend/handlers/handlers.go
benma: I like it!

PR #1254, opened 2021-05-04
frontend/bitbox02: display securechip model

benma:
Available from v9.6.0.

### DIFF
--- a/frontends/web/src/components/devices/bitbox02/settings.tsx
+++ b/frontends/web/src/components/devices/bitbox02/settings.tsx
@@ -34,11 +34,12 @@ interface SettingsProps {
 
 interface State {
     versionInfo?: VersionInfo;
-    deviceInfo: {
+    deviceInfo?: {
         name: string;
         initialized: boolean;
         version: string;
         mnemonicPassphraseEnabled: boolean;
+        securechipModel: string;
     };
 }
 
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -579,6 +579,7 @@
         "label": "microSD card",
         "true": "Inserted"
       },
+      "securechip": "Secure chip",
```
x1ddos, line 582: Alternatively, "Secure element"? The "chip" is more nerdy and I like it  :) but I've seen this alt name in many places. Not sure though.

frontends/web/src/components/devices/bitbox02/settings.tsx
```diff
@@ -34,11 +34,12 @@ interface SettingsProps {
 
 interface State {
     versionInfo?: VersionInfo;
-    deviceInfo: {
+    deviceInfo?: {
```
x1ddos, line 37: Hm... so how did it work before this, without the `?`?

x1ddos (approved):
LGTM but see comments

frontends/web/src/locales/en/app.json
benma: We've used secure chip everywhere since the beginning, e.g. here: https://shiftcrypto.ch/bitbox02/

frontends/web/src/components/devices/bitbox02/settings.tsx
benma: Yeah there is even a `!== undefined` check somewhere, which I would expect the linter to trigger if it cannot be `undefined`. :shrug:

PR #1256, opened 2021-05-05
backend/handlers: expose CreateAndPersistAccountConfig

benma:
We re-purpose the existing /account-add endpoint, whose old behavior
will be retired.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -35,6 +35,21 @@ const hardenedKeystart uint32 = hdkeychain.HardenedKeyStart
 // limit, but simply use a hard limit for simplicity.
 const accountsHardLimit = 5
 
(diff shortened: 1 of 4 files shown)


backend/accounts.go
```diff
@@ -35,6 +35,21 @@ const hardenedKeystart uint32 = hdkeychain.HardenedKeyStart
...
 
+// ErrorCode are errors that are represented by an error code. This helps the frontend to translate
+// error messages.
+type ErrorCode string
+
+func (e ErrorCode) Error() string {
+	return string(e)
+}
+
+var (
```
x1ddos, line 46: `const` should work, and it's better here, right?

x1ddos (approved):
LGTM

PR #1258, opened 2021-05-05
Frontend multiaccount UI, rebased and extended

benma:


frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -97,8 +99,27 @@ class ManageAccount extends Component<Props, State> {
...
+                    errorCode?: 'alreadyExists' | 'limitReached';
+                    errorMessage?: string;
+                }
+
+                apiPost('account-add', {
+                    coinCode,
+                    name: accountName,
+                }).then((data: ResponseData) => {
+                    if (data.success) {
+                        this.setState({ step: 2 });
```
thisconnect, line 113: Shouldn't it be step + 1 ​`this.setState((state) => ({ step: state.step + 1 }));` ?

Otherwise skiping over select coin doesn't work for the bitcoin-only verison

frontends/web/src/routes/account/manage/manage.tsx
```diff
@@ -108,7 +129,6 @@ class ManageAccount extends Component<Props, State> {
         this.setState((state) => {
             switch (state.step) {
                 case 0: return ({ step: 1 });
-                case 1: return ({ step: 2 });
```
thisconnect: why did you remove this?

thisconnect: hm just realized it is outdated... 🤔

benma: It happens when the apiPost above succeeds. If adding the account fails, we shouldn't show the success message.

frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
```diff
@@ -133,9 +42,10 @@ function CoinDropDown({
                     disabled: true,
                     value: 'choose',
                 },
-                ...(supportedCoins).map(item => ({
-                    value: item,
-                    text: COIN_AND_ACCOUNT_CODES[item].name,
+                ...(supportedCoins).map(({ coinCode, name, canAddAccount}) => ({
+                    value: coinCode,
+                    text: name,
+                    disabled: !canAddAccount,
```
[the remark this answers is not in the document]
benma: :+1: discussed with @x1ddos . We should probably add a sentence like 'Up to five accounts per coin are supported' (which is what we settled on for now, could change later).

frontends/web/src/routes/account/manage/manage.tsx
benma: So step 2 is not always the last step?

In any case, I think it would help a lot to get rid of the numeric step and switch to logical steps only. If there is only one coin, start at `choose-name`

frontends/web/src/routes/account/manage/manage.tsx
thisconnect: yep makes sense

frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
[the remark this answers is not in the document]
thisconnect: When would you expect to see this message?

once you try to add the 6th account, or already at the beginning of the workflow, or just in the guide?

cc @jadzeidan wdyt?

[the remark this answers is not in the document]
thisconnect: can be done later, lets merge this.

thisconnect (approved):
LGTM lets merge and iterate

frontends/web/src/routes/account/manage/components/coin-dropdown.tsx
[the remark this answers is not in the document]
benma: If the coin is disabled like in this PR, then you can't even add the account, so it would have to be before.

No strong opinions anyway.

@jadzeidan @thisconnect fyi, the plan for now is not go into the gaplimit-for-accounts business and just limit to the accounts per coin and bitbox to e.g. 5, which is a loooot simpler conceptually and in code (e.g. we don't need to do any scanning to decide if the user can add an account). We can still switch to gap limits later if 5 accounts is not good enough.

PR #1259, opened 2021-05-05
backend/devices: cache root fingerprint during app runtime

x1ddos:


benma (approved):
tACK makes a huge difference for the bb01.

PR #1261, opened 2021-05-06
backend: move root fingerprint into the signing config

benma:
It belongs with the xpub and keypath, as each one can have a different
rootFingerprint in theory (e.g. multisig accounts).


x1ddos (approved):
LGTM. Makes sense.

PR #1262, opened 2021-05-06
backend/erc20: move token code to erc20.Token

benma:
So that one can access it via coin.erc20Token.Code, which will be
needed to configure the exchange rates fetcher.


x1ddos:
FTR this was for `backend.configureHistoryExchangeRates` but we've realized there's no need for that: `backend.accounts[n].Coin().Code()` already returns `eth-erc20-xxx` for ERC20 tokens.

PR #1269, opened 2021-05-11
frontend: fix send/receie titles to use the coin name

benma:
Account name is not good, especially with user-named accounts. "Get My
Savings" does not make sense.


thisconnect:
> Seems good?

yes, please ignore what I wrote it's all good :)

PR #1270, opened 2021-05-11
Manage multiaccount prototype

thisconnect:


frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function getAccountFavorites() {
+    return apiGet('config')
+        .then(({ frontend = {} }): Promise<TFavorites> => {
+            return frontend.accountFavorites || {};
+        });
+}
+
+export function setAccountFavorite(
+    accountId: string,
```
benma, line 31: pls call it `accountCode` to match the naming elsewhere in frontend and backend

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,47 @@
...
+    favorite: boolean,
+) {
+    return apiGet('config').then(config => {
+        const { frontend = {} } = config;
+        return apiPost('config', {
```
benma, line 36: use `setConfig()` from utils/config.js, which already takes care to only update the diff. It is also used in settings.tsx.

frontends/web/src/app.tsx
```diff
@@ -268,8 +269,14 @@ class App extends Component<Props, State> {
                             <ElectrumSettings
                                 path="/settings/electrum" />
                             <Settings
+                                accounts={accounts}
                                 deviceIDs={deviceIDs}
                                 path="/settings" />
+                            <ManageAccounts
+                                accounts={accounts}
+                                deviceIDs={deviceIDs}
+                                key={this.devicesKey('manage-accounts')}
```
benma, line 278: this has nothing to do with devices?

frontends/web/src/app.tsx
```diff
@@ -268,8 +269,14 @@ class App extends Component<Props, State> {
                             <ElectrumSettings
                                 path="/settings/electrum" />
                             <Settings
+                                accounts={accounts}
                                 deviceIDs={deviceIDs}
                                 path="/settings" />
+                            <ManageAccounts
+                                accounts={accounts}
+                                deviceIDs={deviceIDs}
```
benma, line 277: deviceIDs should not be needed?

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -0,0 +1,149 @@
...
+        getAccountFavorites()
+            .then(accountFavorites => {
+                this.setState({ accountFavorites });
+            })
+            .catch(console.error);
+    }
+
+    public componentDidUpdate(prevProps) {
+        if (prevProps.deviceIDs.length && !this.props.deviceIDs.length) {
+            route('/', true);
```
benma, line 56: I am afraid putting routing here instead of app.tsx (where deviceIDs come from and all other routing happens) makes routing harder to debug.

Besides, why do this route here based on devices anyway? This is an accounts page, not a devices page :)

benma (changes requested):

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: Yes completely agree, but copypasta from settings.tsx and there is some some logic that a home icon is shown when a device is plugged in.

Then since we don't have watch-only I wasn't sure if one should stay on the page when the device is unplugged.

thisconnect: I'll remove the home icon as well.

frontends/web/src/api/config.ts
thisconnect: PTAL

benma (approved):
utACK.

I'll provide an account.isToken to skip them in this view

PR #1271, opened 2021-05-12
backend: remove dynamic split accounts

benma:
Before, we would dynamically explode a unified account into multiple
individual accounts (one per signing config) if the keystore didn't
support unified accounts. This does not work well with account
management, where one wants to list all accounts, edit their names,
etc.

We kept a toggle in the settings to force-explode accounts, so that
users who (ab)used multiple signing configurations as multiple
accounts could get the behavior back. It was meant to be temporary,
but the above forces us to speed up its removal.

After this commit, users will find their split accounts to be unified,
and they can split the funds manually again by sendint the funds to a
new account (by using coin control).

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -415,13 +415,37 @@ func (backend *Backend) persistBTCAccountConfig(
 		)
 		signingConfigurations = append(signingConfigurations, signingConfiguration)
 	}
-	return backend.persistAccount(config.Account{
-		CoinCode:                coin.Code(),
-		Name:                    name,
-		Code:                    code,
-		SupportsUnifiedAccounts: keystore.SupportsUnifiedAccounts(),
-		Configurations:          signingConfigurations,
-	}, accountsConfig)
+
+	if keystore.SupportsUnifiedAccounts() {
+		return backend.persistAccount(config.Account{
+			CoinCode:       coin.Code(),
+			Name:           name,
+			Code:           code,
+			Configurations: signingConfigurations,
+		}, accountsConfig)
+	}
+
+	// Unified accounts not supported, so we add one account per configuration.
+	for _, cfg := range signingConfigurations {
+		suffixedName := name
+		switch cfg.ScriptType() {
+		case signing.ScriptTypeP2PKH:
+			suffixedName += ": legacy"
+		case signing.ScriptTypeP2WPKH:
+			suffixedName += ": bech32"
+		}
+
+		err := backend.persistAccount(config.Account{
+			CoinCode:       coin.Code(),
+			Name:           suffixedName,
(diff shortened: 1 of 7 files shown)


backend/backend.go
```diff
@@ -640,39 +663,9 @@ func (backend *Backend) initPersistedAccounts() {
...
 
-		// We split accounts if the user setting dictates it or if the keystore connected to the
-		// account does not support unified accounts.
-
-		var isBTCBased bool
-		switch account.CoinCode {
-		case coinpkg.CodeBTC, coinpkg.CodeTBTC, coinpkg.CodeRBTC, coinpkg.CodeLTC, coinpkg.CodeTLTC:
-			isBTCBased = true
-		}
-		splitAccounts := isBTCBased && (backend.config.AppConfig().Backend.SplitAccounts ||
```
x1ddos, line 652: Also remove `SplitAccounts` field from the `Backend` struct in config/config.go?

frontends/web/src/locales/en/app.json
```diff
@@ -942,10 +942,6 @@
       "servers": {
         "text": "This app communicates with the Shift Crypto servers to check for updates, load transactions, and send information to paired mobile apps.\nThe app also retrieves the latest exchange rates from CryptoCompare. All conversions are calculated locally which means no data about the amount of your transaction is ever transmitted.\nNote: For Ethereum and ERC20 Tokens, we use Etherscan.io APIs.",
         "title": "Which servers does this app talk to?"
-      },
-      "whyMultipleAccounts": {
```
x1ddos, line 946: This entry is still used in manage-accounts.tsx: https://github.com/digitalbitbox/bitbox-wallet-app/blob/6f725bc/frontends/web/src/routes/settings/manage-accounts.tsx#L132

x1ddos (approved):
LGTM but see comments and please add a changelog entry!

frontends/web/src/locales/en/app.json
benma: Good catch. I assume this was a placeholder entry display the guide on the page? The help text does not fit the page otherwise. @thisconnect correct?

backend/backend.go
benma: :heavy_check_mark:

benma:
> LGTM but see comments and please add a changelog entry!

Done!

frontends/web/src/locales/en/app.json
benma: I removed the guide entry from the manage accounts page.

thisconnect: Yes I kept this as placeholder, it can be removed.

PR #1272, opened 2021-05-12
backend/btc: don't block in FatalError()

benma:
FatalError is called when getting the account status and blocked until
the accoutn was synced. The account component is not rendered until
the status endpoint returns, which would be after syncing. This
prevents the account component from showing the address scan progress
indication.


x1ddos (approved):
:+1:

PR #1274, opened 2021-05-13
Multi-accounts manage tokens prototype

thisconnect:
Added tokens below ETH accounts if the account is active.
Currently some tokens have dummy active state just for testing.
This will later come when the api supports enabling tokens.

Also fixed a token image gpax that added width and
height so the svg size is the same as all other tokens.


benma:
I rebased and extended this PR here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1276

benma:
https://github.com/digitalbitbox/bitbox-wallet-app/pull/1276 was merged instead

PR #1276, opened 2021-05-17
Multi-accounts manage tokens prototype - rebased & extended

benma:
Rebased and extended https://github.com/digitalbitbox/bitbox-wallet-app/pull/1274

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -244,3 +244,15 @@ func (backend *Backend) CreateAndPersistAccountConfig(
 			coinCode, nextAccountNumber, name, keystore, nil, accountsConfig)
 	})
 }
+
+// SetTokenActive activates/deactivates an token on an account. `tokenCode` must be an ERC20 token
+// code, e.g. "eth-erc20-usdt", "eth-erc20-bat", etc.
+func (backend *Backend) SetTokenActive(accountCode string, tokenCode string, active bool) error {
+	return backend.config.ModifyAccountsConfig(func(accountsConfig *config.AccountsConfig) error {
+		acct := accountsConfig.Lookup(accountCode)
+		if acct == nil {
+			return errp.Newf("Could not find account %s", accountCode)
+		}
+		return acct.SetTokenActive(tokenCode, active)
+	})
+}
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -359,7 +359,8 @@ func (backend *Backend) createAndAddAccount(
 				backend.log.WithError(err).Error("could not find ERC20 token")
 				continue
 			}
-			backend.createAndAddAccount(token, erc20TokenCode, token.Name(), signingConfigurations, nil)
+			erc20AccountCode := fmt.Sprintf("%s-%s", code, erc20TokenCode)
+			backend.createAndAddAccount(token, erc20AccountCode, token.Name(), signingConfigurations, nil)
 		}
 	default:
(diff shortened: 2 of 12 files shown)


backend/config/accounts.go
```diff
@@ -43,3 +43,15 @@ func newDefaultAccountsonfig() AccountsConfig {
 		Accounts: []Account{},
 	}
 }
+
+// ByCode returns the account with the given code, or nil if no such account exists.
+// A reference is returned, so the account can be modified by the caller.
+func (cfg AccountsConfig) ByCode(code string) *Account {
```
x1ddos, line 49: Will there be other `ByXxx` methods? If not, I think a better name might have been `Find` or `LookUp` or just `Account`.

backend/accounts.go
```diff
@@ -244,3 +244,32 @@ func (backend *Backend) CreateAndPersistAccountConfig(
...
+		}
+	}
+	if active {
+		activeTokens = append(activeTokens, tokenCode)
+	}
+	acct.ActiveTokens = activeTokens
+	return nil
+}
+
+// SetTokenActive activates/deactivats an token on an account. `tokenCode` must be an ERC20 token
```
x1ddos, line 269: deactivat**e**s typo

thisconnect (approved):
Tested  frontend with `make servewallet-mainnet` tested adding another ETH account and activated some tokens

LGTM

backend/backend.go
```diff
@@ -359,7 +359,8 @@ func (backend *Backend) createAndAddAccount(
 				backend.log.WithError(err).Error("could not find ERC20 token")
 				continue
 			}
-			backend.createAndAddAccount(token, erc20TokenCode, token.Name(), signingConfigurations, nil)
+			erc20AccountCode := fmt.Sprintf("%s-%s", code, erc20TokenCode)
```
x1ddos, line 362: A grep shows this already in other 3 places. Time to factor out into a helper func?

image

backend/accounts.go
```diff
@@ -244,3 +244,32 @@ func (backend *Backend) CreateAndPersistAccountConfig(
 			coinCode, nextAccountNumber, name, keystore, nil, accountsConfig)
 	})
 }
+
+func setTokenActive(accountCode string, tokenCode string, active bool, accountsConfig *config.AccountsConfig) error {
```
x1ddos, line 248: Why this choice of a separate function in accounts.go? Wouldn't it be logical to make this a method of `config.AccountsConfig`? Then your last argument `accountsConfig` becomes the receiver.

x1ddos (approved):
LGTM

backend/config/accounts.go
benma: done

backend/accounts.go
benma: done

backend/backend.go
benma: They all seem to have a different context though, and happen to use the same syntax by coincidence, not because there is something special about %s-%s.

backend/accounts.go
benma: done

backend/backend.go
x1ddos: Then I would definitely prefix them with that context:

```go
fmt.Sprintf("context-%s-%s", ...)
```

so to not give an impression they are interchangeable

PR #1277, opened 2021-05-17
Multiaccounts add-account improvements

thisconnect:
- refactor add-account address remaining feedback from https://github.com/digitalbitbox/bitbox-wallet-app/pull/1250
- updated simplemarkup component to support className
- better keyboard support
- hide backbutton on the first and last step
- show error inline before input element


frontends/web/src/routes/account/add/add.tsx
```diff
@@ -58,46 +60,35 @@ class ManageAccount extends Component<Props, State> {
     public componentDidMount() {
         backendAPI.getSupportedCoins()
             // TEST with only 1 coin
-            // .then(() => (['btc']))
+            // .then(() => ([{coinCode: 'tbtc' as accountApi.CoinCode, name: 'Bitcoin Testnet', canAddAccount: true}]))
```
benma, line 63: remove?

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -58,46 +60,35 @@ class ManageAccount extends Component<Props, State> {
...
             .then((coins) => {
                 const onlyOneSupportedCoin = (coins.length === 1);
```
benma, line 65: nit: no parens () needed

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -58,46 +60,35 @@ class ManageAccount extends Component<Props, State> {
...
                 this.setState({
-                    // @ts-ignore
-                    coinCode: onlyOneSupportedCoin ? coins[0] : 'choose',
-                    supportedCoins: coins
+                    coinCode: onlyOneSupportedCoin ? coins[0].coinCode : 'choose',
+                    step: onlyOneSupportedCoin ? 'choose-name' : 'select-coin',
```
benma, line 68: what if `coin.canAddAccount` is false but `onlyOneSupportedCoin` is true, e.g. you try to add account when you already have 5 accounts with btc-only?

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -109,29 +100,71 @@ class ManageAccount extends Component<Props, State> {
...
-                    } else {
-                        if (data.errorCode) {
-                            alertUser(this.props.t(`addAccount.error.${data.errorCode}`));
-                        } else if (data.errorMessage) {
-                            alertUser(this.props.t('unknownError', { errorMessage: data.errorMessage }));
+                })
+                    .then((data: ResponseData) => {
+                        if (data.success) {
+                            this.setState({ step: 'success' });
+                            //route('/account/' + data.accountCode);
```
benma, line 107: cherry-pick these changes into this PR, or merge the other one then rebase this one?

https://github.com/digitalbitbox/bitbox-wallet-app/pull/1279

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -109,29 +100,71 @@ class ManageAccount extends Component<Props, State> {
...
+                        } else {
+                            this.showError(data);
                         }
-                    }
-                });
+                    })
+                    .catch(error => this.showError(error));
                 break;
             case 'success':
+                // TODO: route to manage accounts to configure tokens
```
benma, line 115: This is conflicting with the above TODO to route to the new account.

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -58,46 +60,35 @@ class ManageAccount extends Component<Props, State> {
...
-        const { step } = this.state;
-        if (this.onlyOneSupportedCoin()) {
-            return step === 0 ? 'choose-name' : 'success';
-        }
-        switch (step) {
-            case 0: return 'select-coin';
-            case 1: return 'choose-name';
-            case 2: return 'success';
-        }
+            .catch(error => this.showError(error));
```
benma, line 72: .catch() does nothing in production. Errors are always handled in the `then()` clause (success=false)

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -109,29 +100,71 @@ class ManageAccount extends Component<Props, State> {
...
+            case 'select-coin':
+                return true;
+            case 'choose-name':
+                return this.onlyOneSupportedCoin();
+            case 'success':
+                return false;
+        }
+    }
+
+    private focusRef = (ref) => {
```
benma, line 152: can this be moved into the Input component? we can use it in lots of other places. e.g. `<Input autoFocus ...>`.

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -109,29 +100,71 @@ class ManageAccount extends Component<Props, State> {
...
+                    .catch(error => this.showError(error));
                 break;
             case 'success':
+                // TODO: route to manage accounts to configure tokens
                 route('/account-summary');
                 return;
         }
     }
 
+    private showError = (error) => {
```
benma, line 121: I would inline all of this in the apiPost(..).then(...) function above, as it is only used there (after the `.catch(...)` are gone).

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -109,29 +100,71 @@ class ManageAccount extends Component<Props, State> {
...
+                        if (data.success) {
+                            this.setState({ step: 'success' });
+                            //route('/account/' + data.accountCode);
+                        } else {
+                            this.showError(data);
                         }
-                    }
-                });
+                    })
+                    .catch(error => this.showError(error));
```
benma, line 112: same - can remove this.

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -109,29 +100,71 @@ class ManageAccount extends Component<Props, State> {
...
+        }
+    }
+
+    private focusRef = (ref) => {
+        setTimeout(() => {
+            if (ref === document.activeElement) {
+                return;
+            }
+            if (document.activeElement instanceof HTMLElement) {
+                document.activeElement?.blur();
```
benma, line 158: why is this needed? does `.focus()` not take care of that?

benma (changes requested):

frontends/web/src/routes/account/add/add.tsx
thisconnect: Yea, but I don't have a btconly device and guess that I'd still need to test this.

frontends/web/src/routes/account/add/add.tsx
thisconnect: yes I know, but I always add them if it helps me read it.

The transpiler/bundler will remove/rewrite this anyway.

frontends/web/src/routes/account/add/add.tsx
thisconnect: Yes good point, currently the dropdown options are grayed out, without a message and it wont let the user select one.

We need a guide entry for sure, and maybe even a fullscreen message "sorry you used all your accounts" or similar

frontends/web/src/routes/account/add/add.tsx
thisconnect: ideally yes, but if there is a real error catch will catch it and we should display it to the user.

frontends/web/src/routes/account/add/add.tsx
thisconnect: maybe, but I don't want to loose too much time with focus with preact 8 anymore.

frontends/web/src/routes/account/add/add.tsx
thisconnect: It didn't but I can test again, but I am affraid it will behave differently in the Qt app, common focus issue. BTW I really look forward using react hooks for this.

frontends/web/src/routes/account/add/add.tsx
jadzeidan: I think most clear way would be to not to grey out options, but to let the user select whichever coin they want. And if the coin the selected already has max. accounts, then a message is displayed saying they can't make more accounts. Something like "You cannot make any more [coin] accounts. Maximum number in 5."

Of course we can make it a user guide entry, and maybe that is enough for MVP.

thisconnect: Would be nice if max-5-accounts it was communicated earlier, before any action

frontends/web/src/routes/account/add/add.tsx
benma: as i said, `.catch()` after apiPost and apiGet is never called in production (in the Qt app). This line here will never execute, so you can safely remove it.

frontends/web/src/routes/account/add/add.tsx
benma: Maybe it already works, like this?

https://github.com/digitalbitbox/bitbox-wallet-app/blob/staging-multiaccounts/frontends/web/src/components/backups/create.jsx#L91-L92

frontends/web/src/routes/account/add/add.tsx
thisconnect: removed

frontends/web/src/routes/account/add/add.tsx
thisconnect: This doesn't work afaik if another element  already has focus.

frontends/web/src/routes/account/add/add.tsx
thisconnect: removed

frontends/web/src/routes/account/add/add.tsx
thisconnect: cherry-picked

frontends/web/src/routes/account/add/add.tsx
thisconnect: done

frontends/web/src/routes/account/add/add.tsx
thisconnect: removed and tested with qt-osx, got focus

thisconnect:
PTAL

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -57,81 +61,99 @@ class ManageAccount extends Component<Props, State> {
...
-                    }
-                });
+                    });
                 break;
             case 'success':
+                if (accountCode) {
+                    route(`/account/${accountCode}`);
+                    return;
+                }
                 route('/account-summary');
```
benma, line 128: i can't see when this happens - seems success=true is set at the same time as accountCode, always?

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -57,81 +61,99 @@ class ManageAccount extends Component<Props, State> {
...
-                        this.setState({ step: 2 });
-                        //route('/account/' + data.accountCode);
-                    } else {
-                        if (data.errorCode) {
-                            alertUser(this.props.t(`addAccount.error.${data.errorCode}`));
+                })
+                    .then((data: ResponseData) => {
+                        if (data.success) {
+                            this.setState({
+                                accountCode: data.accountCode,
```
benma, line 108: why this and not route right here, like the commented line below?

frontends/web/src/routes/account/add/add.tsx
thisconnect: From your other PR I assumed accountCode isn't always there https://github.com/digitalbitbox/bitbox-wallet-app/pull/1279/files#diff-74c54a5fb0b456d69855bb21b08f65442ab1db6a07ee52185345853d4eb641c0R116

Shall I remove `if (accountCode) {` and expect backend to always return accountCode?

frontends/web/src/routes/account/add/add.tsx
thisconnect: User should see the success message and then when clicking primary button on success step it should go to the account.

Else it would go directly to account without showing the success step.

benma (approved):

PR #1278, opened 2021-05-17
backend: sort accounts by coin, account number

benma:
In lieu of a user-chosen order, we fix the order and group accounts by
coin, ordered by account number.

Tokens are unchanged, are listed after the parent ETH account, in an
arbitrary order.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -16,6 +16,8 @@ package backend
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/btcsuite/btcutil/hdkeychain"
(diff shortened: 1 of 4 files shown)


x1ddos (approved):
LGTM but:

> Tokens are listed in an arbitrary order.

why not also sort tokens alphabetically? it's better than nothing. as a user I'd definitely expect some order.

benma:
> why not also sort tokens alphabetically? it's better than nothing. as a user I'd definitely expect some order.

i agree, meant to say that it was not part of that commit. maybe in the next PR

PR #1281, opened 2021-05-18
frontend: enabling tokens now use toggles on manage accounts

thisconnect:
- added small lines to visualize their close relation to the
  parent ethereum account
- commented out but kept the toggle favor account feature
- linked the coin accounts to their account-overview, but
  have not id for tokens to correctly link to

### DIFF
--- a/frontends/web/src/routes/settings/manage-accounts.tsx
+++ b/frontends/web/src/routes/settings/manage-accounts.tsx
@@ -19,7 +19,6 @@ import { route } from 'preact-router';
 import * as accountAPI from '../../api/account';
 import * as backendAPI from '../../api/backend';
 import { apiGet } from '../../utils/request';
-import { setConfig } from '../../utils/config';
 import { alertUser } from '../../components/alert/Alert';
 import { Button } from '../../components/forms';
 import Logo from '../../components/icon/logo';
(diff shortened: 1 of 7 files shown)


thisconnect:
I'll add show/hide token button in this PR with another commit

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -132,20 +163,27 @@ class ManageAccounts extends Component<Props, State> {
         }
         return Object.entries(this.erc20TokenCodes)
             .map(([tokenCode, name]) => {
-                const active = activeTokens && activeTokens.includes(tokenCode);
+                const active = !!activeTokens && activeTokens.includes(tokenCode);
```
benma, line 166: are the !! really needed? i didn't see typescript complain about sth like this yet.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -132,20 +163,27 @@ class ManageAccounts extends Component<Props, State> {
...
-                        <Logo
-                            active={active}
-                            alt={name}
-                            className={style.tokenIcon}
-                            coinCode={tokenCode}
-                            stacked />
-                        <span className="flex-1 p-left-quarter">
-                            {name} ({tokenCode})
-                        </span>
+                        <div
```
benma, line 170: why not `<A href...`?

The link is underlined but it also needs a hand/finger mouse pointer.

benma (approved):
looks nice!

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: IAccount says activeTokens can be undefined/missing is that correct?

https://github.com/digitalbitbox/bitbox-wallet-app/blob/staging-multiaccounts/frontends/web/src/api/account.ts#L39

Screen Shot 2021-05-19 at 12 02 02 PM

The Toggle component wants a boolean, so Typescript complains, could either:
- keep as is
- update IAccounts interface (if backend always returns at least an empty array)
- move `!!` to Toggle, but I think `active ` should be a boolean and never undefined.

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: Inactive tokens shouldn't be linked, so we won't always have a link, I'd prefer at the moment to keep using onClick

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -157,13 +157,19 @@ class ManageAccounts extends Component<Props, State> {
...
         }
         return Object.entries(this.erc20TokenCodes)
             .map(([tokenCode, name]) => {
                 const active = activeTokens !== undefined && activeTokens.includes(tokenCode);
+                const accountCode = accounts
+                    .filter(account => account.coinCode === tokenCode);
+                    // .filter or .find by my ETH account somehow, need ETH parent account
+                if (accountCode.length > 1) {
+                    console.log(accountCode);
+                }
```
thisconnect, line 172: @benma as discussed, we'd at least need the parent ETH account or similar

thisconnect:
LGTM

PR #1287, opened 2021-05-20
frontend: resurrect old translations

thisconnect:
We might need them for watch-only and they are already translated.


benma:
NACK ;) imho this is deadweight, we will never use them. we can bring them back from the git history if we end up needing them, but i highly doubt it.

thisconnect:
ack

PR #1291, opened 2021-05-20
backend: add function to rename an account

benma:


backend/accounts.go
```diff
@@ -302,3 +302,25 @@ func (backend *Backend) SetTokenActive(accountCode string, tokenCode string, act
...
 }
+
+// RenameAccount renames an account in the accounts database.
+func (backend *Backend) RenameAccount(accountCode string, name string) error {
+	if name == "" {
+		return errp.New("Name cannot be empty")
+	}
+	return backend.config.ModifyAccountsConfig(func(accountsConfig *config.AccountsConfig) error {
+		found := false
+		for i := range accountsConfig.Accounts {
```
x1ddos, line 313: or, in fewer lines:

```go
for _, acct := range accountsConfig.Accounts {
  if acct.Code == accountCode {
    acct.Name = name
    return nil
  }
}
return errp.Newf("Account not found: %s", accountCode)
```

No need for `found` var.

x1ddos (approved):

backend/accounts.go
benma: excellent!

PR #1293, opened 2021-05-20
frontend: add rename account dialog to manage accounts

thisconnect:
Added edit button to all accounts that are real coins and can
be named, edit input and save form displayed with a dialog
as inline is a bit trickier and we dont want the current
solution from edit notes.

### DIFF
--- a/frontends/web/src/routes/settings/manage-accounts.tsx
+++ b/frontends/web/src/routes/settings/manage-accounts.tsx
@@ -20,10 +20,13 @@ import * as accountAPI from '../../api/account';
 import * as backendAPI from '../../api/backend';
 import { apiGet } from '../../utils/request';
 import { alertUser } from '../../components/alert/Alert';
-import { Button } from '../../components/forms';
+import { Button, Input } from '../../components/forms';
 import Logo from '../../components/icon/logo';
 import { Header } from '../../components/layout';
 import { Toggle } from '../../components/toggle/toggle';
+import { Dialog } from '../../components/dialog/dialog';
+import { Message } from '../../components/message/message';
+import * as dialogStyle from '../../components/dialog/dialog.css';
 import { translate, TranslateProps } from '../../decorators/translate';
 import * as style from './manage-accounts.css';
 
@@ -41,6 +44,9 @@ type TShowTokens = {
 }
 
 interface State {
+    editAccountCode?: string;
+    editAccountNewName: string;
+    editErrorMessage?: string;
     favorites?: TFavorites;
     accounts: accountAPI.IAccount[];
     showTokens: TShowTokens;
@@ -48,6 +54,8 @@ interface State {
 
 class ManageAccounts extends Component<Props, State> {
     public readonly state: State = {
+        editAccountNewName: '',
+        editErrorMessage: undefined,
         favorites: undefined,
         accounts: [],
         showTokens: {}
@@ -107,9 +115,11 @@ class ManageAccounts extends Component<Props, State> {
                             <span className="unit">({account.coinUnit})</span>
                         </span>
                     </div>
-                    {/* <ButtonLink>
-                        Edit
-                    </ButtonLink> */}
+                    <button
+                        className={style.editBtn}
+                        onClick={() => this.setState({ editAccountCode: account.code, editAccountNewName: account.name })}>
+                        {t('manageAccounts.editAccount')}
+                    </button>
                     {/* <Toggle
                         checked={active}
                         id={account.code}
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -226,6 +253,25 @@ class ManageAccounts extends Component<Props, State> {
                                 </div>
                             </div>
                         ) : null }
+                        { editAccountCode ? (
+                            <Dialog
+                                onClose={() => this.setState({ editAccountCode: undefined, editAccountNewName: '' })}
```
benma, line 258: why name to `''` and not to `undefined`?

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -41,13 +44,17 @@ type TShowTokens = {
 }
 
 interface State {
+    editAccountCode?: string;
+    editAccountNewName?: string;
```
benma, line 48: remove `?` - it is set to `''` anyway

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -201,9 +210,27 @@ class ManageAccounts extends Component<Props, State> {
...
     }
 
+    private updateAccountName = (event: Event) => {
+        event.preventDefault();
+        const { editAccountCode, editAccountNewName } = this.state;
+
+        backendAPI.renameAccount(editAccountCode!, editAccountNewName!)
+            .then(result => {
+                if (!result.success) {
+                    this.setState({ editErrorMessage: result.errorMessage });
```
benma, line 220: it looks nice but code-wise it might be overkill - it should be impossible to get an error (if you prevent sending an empty name). for such errors, we often simply do alertUser(...) to show the message, in case an unexpected error comes out.

no need to change, just a thought.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -226,6 +253,25 @@ class ManageAccounts extends Component<Props, State> {
...
+                        { editAccountCode ? (
+                            <Dialog
+                                onClose={() => this.setState({ editAccountCode: undefined, editAccountNewName: '' })}
+                                title={t('manageAccounts.editAccountNameTitle')}>
+                                <form onSubmit={this.updateAccountName}>
+                                    <Message type="warning" hidden={!editErrorMessage}>
+                                        {editErrorMessage}
+                                    </Message>
+                                    <Input
+                                        onInput={e => this.setState({ editAccountNewName: e.target.value})}
```
benma, line 265: space before `}`? :D

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -226,6 +253,25 @@ class ManageAccounts extends Component<Props, State> {
                                 </div>
                             </div>
                         ) : null }
+                        { editAccountCode ? (
+                            <Dialog
+                                onClose={() => this.setState({ editAccountCode: undefined, editAccountNewName: '' })}
+                                title={t('manageAccounts.editAccountNameTitle')}>
+                                <form onSubmit={this.updateAccountName}>
+                                    <Message type="warning" hidden={!editErrorMessage}>
```
benma, line 261: should be type="error" or not?

benma (approved):

benma (changes requested):
the error message is not reset and shows up in a new dialog

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -41,13 +44,17 @@ type TShowTokens = {
 }
 
 interface State {
+    editAccountCode?: string;
+    editAccountNewName?: string;
+    editErrorMessage?: string;
```
benma, line 49: just an idea, not sure: maybe it's easier to put those into one object:

`editState?: { accountCode: string; ... }`

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -226,6 +257,28 @@ class ManageAccounts extends Component<Props, State> {
                                 </div>
                             </div>
                         ) : null }
+                        { editAccountCode ? (
+                            <Dialog
+                                onClose={() => this.setState({ editAccountCode: undefined, editAccountNewName: '' })}
```
benma, line 262: reset error message here too

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: Looking at it I don't like it, but commit it anyway. 
Screen Shot 2021-05-20 at 4 25 01 PM

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: setState doesn't deep merge

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: this will be passed to the input element, and it should do `input.value = ''` and not `input.value = undefined`.

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: will remove `?`

frontends/web/src/routes/settings/manage-accounts.tsx
benma: imho whose two {} are different, one is TS and the other is TSX. we don't usually put spaces in JSX {}, but we do for JS objects. anyway :D

benma (approved):

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: there is like 50 `}` no wonder I missed it :P

PR #1296, opened 2021-05-25
backend: limit number of accounts for bitbox01 to 1

benma:
In the app, we allow up to 5 accounts per coin for the BitBox02, and
one account per coin for BitBox01.


x1ddos:
Just to double check: how does this work with account "explosion" for bb01, with different script types?

benma:
> Just to double check: how does this work with account "explosion" for bb01, with different script types?

Excellent catch. In fact `len(accounts)` was not a good idea after all to get the next account number. Before it would not have worked with the BitBox01 (now disabled anyway).

I added a commit that gets the next account number by finding the maximum existing number.

backend/accounts.go
```diff
@@ -257,12 +257,31 @@ func (backend *Backend) nextAccountNumber(coinCode coinpkg.Code, keystore keysto
 		return account.Configurations.ContainsRootFingerprint(rootFingerprint)
 	}
 
-	nextAccountNumber := uint16(len(backend.filterAccounts(accountsConfig, filter)))
+	nextAccountNumber := uint32(0)
```
x1ddos, line 260: `int` would've worked just fine here. like I was saying, specific int types get in the way of readability. plus, you convert it to `uint16` in the end anyway. but, `AccoutNumber` returns `uint32`, unfortunately :(

so, you see, different conversions for not benefit.

backend/accounts.go
```diff
@@ -257,12 +257,31 @@ func (backend *Backend) nextAccountNumber(coinCode coinpkg.Code, keystore keysto
...
+
+	if !keystore.SupportsMultipleAccounts() && nextAccountNumber >= 1 {
+		return 0, errp.WithStack(ErrAccountLimitReached)
+	}
 
 	if nextAccountNumber >= accountsHardLimit {
 		return 0, errp.WithStack(ErrAccountLimitReached)
 	}
-	return nextAccountNumber, nil
+	return uint16(nextAccountNumber), nil
```
x1ddos, line 284: I know this cannot happen (famous last words), but this is a theoretical unchecked int overflow.

backend/accounts.go
```diff
@@ -257,12 +257,31 @@ func (backend *Backend) nextAccountNumber(coinCode coinpkg.Code, keystore keysto
 		return account.Configurations.ContainsRootFingerprint(rootFingerprint)
 	}
 
-	nextAccountNumber := uint16(len(backend.filterAccounts(accountsConfig, filter)))
```
x1ddos, line 260: To cover the case of bb01, would this work as well?

```go
if !keystore.SupportsMultipleAccounts() && len(backend.filterAccounts(accountsConfig, filter)) > 0 {
  return 0, ErrAccountLimitReached
}
```

benma: yes but the code as it is now also covers it, right?

x1ddos (approved):

benma:
@x1ddos added unit tests and slightly changed the function, pls check.

backend/keystore/keystore.go
```diff
@@ -38,6 +38,7 @@ const (
 var ErrSigningAborted = errors.New("signing aborted by user")
 
 // Keystore supports hardened key derivation according to BIP32 and signing of transactions.
+//go:generate moq -pkg mocks -out mocks/keystore.go . Keystore
```
benma, line 41: before you say anything @x1ddos, this tool is the one we looked at that produces mocks nearly like I'd write them by hand :D

backend/accounts_test.go
```diff
@@ -21,11 +21,21 @@ import (
...
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/signing"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func mustKeypath(keypath string) signing.AbsoluteKeypath {
+	kp, err := signing.NewAbsoluteKeypath(keypath)
+	if err != nil {
+		panic(err)
```
x1ddos, line 34: A panic prevents all following tests from running, apart from avoiding an ugly stacktrace. A better way imho would be to fail only one specific test where this is used. for example,

```go
func makeKeypath(t *testing.T, p string) signing.AbsoluteKeypath {
  t.Helper()
  ...
  if err != nil {
    t.Fatalf("%s: %v", p, err)
  }
  return kp
}
```

backend/keystore/keystore.go
x1ddos: I remember :)
I still don't like the separate package `mocks`. It's often more useful in the same package as that one under test. plus, all those 10+ ugly sync.RWMutex.

x1ddos (approved):
LGTM

benma:
> It's often more useful in the same package as that one under test. 

Such structs are often used in tests from multiple packages (any package that uses it and has tests). Where would you put it then? Example: BlockchainMock

x1ddos:
> > It's often more useful in the same package as that one under test.
> 
> Such structs are often used in tests from multiple packages (any package that uses it and has tests). Where would you put it then? Example: BlockchainMock

If it is really used extensively in different package tests, maybe it deserves a dedicated testing package. Example: `net/http/httptest`. If I write `httptest.NewRequest(...)` it's very clear where it's coming from, compared to an anonymous `mock.Something`. Yeah sure, you can always rename the pkg at import but it's again an atypical use, like `int` vs `uint32`. Plus, If it's always called `mock` and you import multiple of them, you're really constrained to do named imports, picking random names.

PR #1297, opened 2021-05-25
backend/eth: increase update interval to 1min

benma:
With newly up to 5 accounts per ETH and token, and conservatively 5
requests per update round, updating every 30s is too quick and 1min
should be ok.


x1ddos (approved):
Good idea!

PR #1299, opened 2021-05-26
frontend: hide add-account button in sidebar

thisconnect:
Reasons are default accounts should be enough for average
users, also once the accounts are created the button does
not always have to be visible.
Add account in manage accounts should be sufficient.

### DIFF
--- a/frontends/web/src/components/sidebar/sidebar.tsx
+++ b/frontends/web/src/components/sidebar/sidebar.tsx
@@ -175,19 +175,6 @@ class Sidebar extends Component<Props> {
                         <span className="sidebarHeader" hidden={!keystores.length}>
                             {t('sidebar.accounts')}
                         </span>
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -175,7 +175,7 @@ class Sidebar extends Component<Props> {
                         <span className="sidebarHeader" hidden={!keystores.length}>
                             {t('sidebar.accounts')}
                         </span>
-                        <span className="sidebarHeaderAction" hidden={!keystores.length}>
+                        <span className="sidebarHeaderAction" hidden={!keystores.length || !debug}>
```
benma, line 178: just delete the whole section? no need to diverge in debug mode here.

benma (changes requested):

benma (approved):

PR #1300, opened 2021-05-26
frontend: do not use guide entry link url

thisconnect:
There is multiple issues:
- in some translations the url is missing or translated
- Entry component does not fallback to english if the
  translated url is missing
- url should probably not be in translations (mostly)

Changed guide buy entry about exchanges to pass link-text,
title and text and hardcode the exchanges url.
Removed in app.json so we don't forget to update it.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -854,8 +854,7 @@
     "buy": {
       "exchanges": {
         "link": {
-          "text": "List of exchanges",
-          "url": "/exchanges"
+          "text": "List of exchanges"
         },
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -854,8 +854,7 @@
     "buy": {
       "exchanges": {
         "link": {
-          "text": "List of exchanges",
-          "url": "/exchanges"
```
thisconnect, line 858: should I remove it in locize?

benma (approved):
nice! utACK

PR #1301, opened 2021-05-26
backend: error on duplicate account name

benma:
To avoid confusion, we disallow adding an account with a name that
another account already has.

We check all accounts, even ones that are not currently visible - we
might show them in the future in watch-only mode.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -463,7 +463,7 @@ func (handlers *Handlers) postAddAccountHandler(r *http.Request) (interface{}, e
 	if err != nil {
 		handlers.log.WithError(err).Error("Could not add account")
 		if errCode, ok := errp.Cause(err).(backend.ErrorCode); ok {
-			return response{Success: false, ErrorCode: errCode.Error()}, nil
+			return response{Success: false, ErrorCode: string(errCode)}, nil
 		}
 		return response{Success: false, ErrorMessage: err.Error()}, nil
 	}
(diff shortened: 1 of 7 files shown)


benma:
Will also apply this to renaming accounts, hang on.

backend/handlers/handlers.go
```diff
@@ -537,12 +537,16 @@ func (handlers *Handlers) postRenameAccountHandler(r *http.Request) (interface{}
...
 		ErrorMessage string `json:"errorMessage,omitempty"`
+		ErrorCode    string `json:"errorCode,omitempty"`
 	}
 
 	if err := json.NewDecoder(r.Body).Decode(&jsonBody); err != nil {
 		return response{Success: false, ErrorMessage: err.Error()}, nil
 	}
 	if err := handlers.backend.RenameAccount(jsonBody.AccountCode, jsonBody.Name); err != nil {
+		if errCode, ok := errp.Cause(err).(backend.ErrorCode); ok {
+			return response{Success: false, ErrorCode: errCode.Error()}, nil
```
x1ddos, line 548: A bit confusing. I would expect an error code to already be some kind of machine-readable code. Wouldn't `ErrorCode: string(errCode)` be more logical here? ErrorCode satisfies the `error` type by implementing the `Error` method but it doesn't really mean you have to call it in all cases.

x1ddos (approved):
LGTM

backend/handlers/handlers.go
benma: done. fyi `Error()` returned the error code before, but `string(errCode)` is more robust.

PR #1303, opened 2021-05-26
frontend/add-account: automatically suggest name

benma:
Add it as a placeholder, use it as a default name if none was entered.

The suggested name is the coin name followed by the next account number.


thisconnect:
CI failing

thisconnect (approved):
LGTM

PR #1305, opened 2021-05-26
backend: fix regression with naming the default initial accounts

benma:
If no accounts exist, the first account of each supported coin is
added with a default name. Making a default name was accidentally
removed in 4ab179273826c59137bc7c592a6ca1c0484a11f1, leading to an
error when adding the initial accounts.


x1ddos (approved):
LGTM

PR #1306, opened 2021-05-26
android: bump versionCode

benma:
Submitted the previous code for review, but a regression was found, so
we have to make a new release.


x1ddos (approved):
:+1:

PR #1307, opened 2021-05-27
frontend: remove error when continuing

thisconnect:
There can be error, i.e. account name already in use, the error
currently stays if the user progresses to the next step or is
going back.

Changed to hide the error when progressing with continue or back.


benma (approved):
utack

PR #1308, opened 2021-05-27
Remove BitBoxBase

benma:
It is a maintenance burden. We can resurrect the parts we need from
the git history if needed.


thisconnect (approved):
frontend LGTM

x1ddos (approved):
backend LGTM except there are still protobufs in the vendored bitbox02-api-go. I think you'll want to remove those two, marking its fields as "reserved" so we won't reuse them in the future. but I guess it'll be a separate commit.

PR #1309, opened 2021-05-27
frontend/account/add: show name on success

benma:
If the user went with the suggested placeholder name, the success
message did not show the name.


thisconnect (approved):
atACK

PR #1310, opened 2021-05-27
frontend/account/add: make placeholder a value instead

benma:
The suggested name should be entered as a value into the input field,
not a placeholder, so it is more intuitive that one can simply proceed
with the suggested name.


thisconnect (approved):
LGTM

PR #1314, opened 2021-05-28
backend/account/notes: restore transaction notes lost in migration

benma:
Transaction notes are stored in json files named using the account
identifier. In v4.28.0, account identifiers changed as part of the
multi-accounts feature. It used to contain the hash of the signing
configurations. As a result, transaction notes set before were not
shown anymore, as the wrong file was loaded.

This patch restores the legacy transaction notes by loading the legacy
transaction note files.

In v4.27.0, you could have a unified account and split them into
individual accounts (one per signing configuration). Transaction notes
were not shared between these accounts even them. In v4.28.0, we
removed the split-account setting, so we consider notes from split
accounts as well - otherwise they would be lost for good.

### DIFF
--- a/backend/accounts/baseaccount.go
+++ b/backend/accounts/baseaccount.go
@@ -18,6 +18,7 @@ import (
 	"encoding/csv"
 	"fmt"
 	"io"
+	"os"
 	"path"
 	"sync"
 	"time"
@@ -65,10 +66,18 @@ type BaseAccount struct {
 	synced  bool
 	offline error
 
-	notes *notes.Notes
+	// notes handles transaction notes.
+	//
+	// It is a slice for migration purposes: from v4.27.0 to v4.28.0, the account identifiers
+	// changed. The slice contains all possible instances of where notes are stored. The first
+	// element is the newest, and other elements are notes stored under legacy names. After
+	// `Initialize()`, this will always have at least one element.
+	notes []*notes.Notes
 
 	proposedTxNote   string
 	proposedTxNoteMu sync.Mutex
+
+	log *logrus.Entry
 }
 
 // NewBaseAccount creates a new Account instance.
@@ -76,6 +85,7 @@ func NewBaseAccount(config *AccountConfig, coin coin.Coin, log *logrus.Entry) *B
 	account := &BaseAccount{
 		config: config,
 		coin:   coin,
+		log:    log,
 	}
 	account.Synchronizer = synchronizer.NewSynchronizer(
 		func() { config.OnEvent(EventSyncStarted) },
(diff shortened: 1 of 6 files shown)


benma:
I need to test & unit test ETH and tokens too - they might not work yet.

benma:
> I need to test & unit test ETH and tokens too - they might not work yet.

Tested ETH and tokens manually and it works. Unit tests to follow.

backend/accounts/baseaccount.go
```diff
@@ -168,14 +213,33 @@ func (account *BaseAccount) GetAndClearProposedTxNote() string {
...
-	if err := account.notes.SetTxNote(txID, note); err != nil {
-		return err
+	for idx, notes := range account.notes {
+		// Store the note, delete it from legacy notes.
+		storeNote := note
+		if idx > 0 {
+			storeNote = ""
+		}
+		if err := notes.SetTxNote(txID, storeNote); err != nil {
+			return err
```
x1ddos, line 223: I would ignore and continue here for idx > 0. Maybe a more straightforward way - tbh I didn't get `storeNote = ""` at the first second - is to split new notes vs legacy for clarity:

```go
// The notes slice is guaranteed to have at least one element by BaseAccount.Initialize. 
if err := account.notes[0].SetTxNote(txID, note); err != nil {
  return err
}
// Delete the notes in legacy files. Don't really care if it fails.
for i, notes := range account.notes[1:] {
  if err := notes.SetTxNote(txID, ""); err != nil {
    log.Printf("Can't delete a note from a legacy file idx=%d: %v", i, err)
  }
}
return nil
```

backend/accounts/baseaccount.go
```diff
@@ -65,7 +65,12 @@ type BaseAccount struct {
 	synced  bool
 	offline error
 
-	notes *notes.Notes
+	// notes handles transaction notes.
+	//
+	// It is a slice for migration purposes: from v4.27.0 to v4.28.0, the account identifiers
+	// changed. The slice contains all possible instances of where notes are stored. The first
+	// element is the newest, and other elements are notes stored under legacy names.
```
x1ddos, line 72: please add a phrase saying notes always has at least one element.

x1ddos (approved):
LGTM but see comments.

backend/accounts/baseaccount.go
```diff
@@ -131,20 +136,60 @@ func (account *BaseAccount) SetOffline(offline error) {
...
+	}
+	for _, identifier := range legacyAccountIdentifiers {
+		legacyNotes, err := notes.LoadNotes(path.Join(
+			account.config.NotesFolder,
+			fmt.Sprintf("%s.json", identifier),
+		))
+		if err != nil {
+			return err
+		}
+		account.notes = append(account.notes, legacyNotes)
```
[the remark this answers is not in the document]
benma: great catch! done

backend/accounts/baseaccount.go
benma: done

backend/accounts/baseaccount.go
benma: done

PR #1315, opened 2021-05-28
backend/etherscan: improve error message related to rate limiter

benma:
There was a bogus unmarshal error before. Now:

unexpected response: {"status":"0","message":"NOTOK","result":"Max rate limit reached"}

### DIFF
--- a/backend/coins/eth/etherscan/etherscan.go
+++ b/backend/coins/eth/etherscan/etherscan.go
@@ -18,6 +18,7 @@ package etherscan
 import (
 	"context"
 	"encoding/json"
+	"io/ioutil"
 	"math/big"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


backend/coins/eth/etherscan/etherscan.go
```diff
@@ -67,9 +68,13 @@ func (etherScan *EtherScan) call(params url.Values, result interface{}) error {
 	if response.StatusCode != http.StatusOK {
 		return errp.Newf("expected 200 OK, got %d", response.StatusCode)
 	}
-	if err := json.NewDecoder(response.Body).Decode(result); err != nil {
+	body, err := ioutil.ReadAll(response.Body)
```
x1ddos, line 71: I would limit here with `io.LimitedReader`. Not sure but 1Mb is probably more than enough?

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -67,9 +68,13 @@ func (etherScan *EtherScan) call(params url.Values, result interface{}) error {
...
+	if err != nil {
 		return errp.WithStack(err)
 	}
+	if err := json.Unmarshal(body, result); err != nil {
+		return errp.Newf("unexpected response: %s", string(body))
```
x1ddos, line 76: Since this is visible to the user, I would clarify:

> unexpected response from Etherscan: %s

x1ddos (approved):
LGTM. See comments.

backend/coins/eth/etherscan/etherscan.go
benma: done

backend/coins/eth/etherscan/etherscan.go
benma: maybe another day - don't want to spend time now estimating the max size

PR #1318, opened 2021-06-01
backend: give account code its own type

benma:


backend/handlers/handlers.go
```diff
@@ -888,7 +888,7 @@ func (handlers *Handlers) getExchangeMoonpayBuy(r *http.Request) (interface{}, e
 	// TODO: Refactor to make use of a map.
 	var acct accounts.Interface
 	for _, a := range handlers.backend.Accounts() {
-		if a.Config().Code == acctCode {
+		if string(a.Config().Code) == acctCode {
```
x1ddos, line 891: Or do the opposite?

```go
acctCode := accounts.Code(mux.Vars(r)["code"])
...
if a.Config().Code == acctCode {
   ...
}
```

Less conversions, easier to read. Here and in the other place.

backend/coins/btc/handlers/handlers.go
```diff
@@ -196,7 +196,7 @@ func (handlers *Handlers) getAccountTransactions(_ *http.Request) (interface{},
 }
 
 func (handlers *Handlers) postExportTransactions(_ *http.Request) (interface{}, error) {
-	name := time.Now().Format("2006-01-02-at-15-04-05-") + handlers.account.Config().Code + "-export.csv"
+	name := time.Now().Format("2006-01-02-at-15-04-05-") + string(handlers.account.Config().Code) + "-export.csv"
```
x1ddos, line 199: This would read nicer:

```
fmt.Sprintf("%s-%s-export.csv", time.Now().Format("2006-01-02-at-15-04-05"), handlers.account.Config().Code)
```

- no need for explicit `string(..code..)`
- no funny trailing `-` in time format
- easier to see how the final filename looks like

backend/backend.go
```diff
@@ -444,7 +444,7 @@ func (backend *Backend) persistBTCAccountConfig(
 		err := backend.persistAccount(config.Account{
 			CoinCode:       coin.Code(),
 			Name:           suffixedName,
-			Code:           fmt.Sprintf("%s-%s", code, cfg.ScriptType()),
+			Code:           accounts.Code(fmt.Sprintf("%s-%s", code, cfg.ScriptType())),
```
x1ddos, line 447: Please consider addting these to accounts package, something like:

```go
package accounts

// ScriptTypedCode produces an account code used in/for ...
func ScriptTypedCode(code Code, typ signing.ScriptType) Code

// FingerprintedCode makes an account code from a keystore root fingerprint.
// ... not sure where this one is used compared to ScriptTypedCode because they both
// appear in somethingPersistAccount and would be very nice to have docs explaining it here.
func FingerprintedCode(finger []byte, coinCode coinpkg.Code, acctNumber int) Code {
  return Code(fmt.Sprintf("v0-%x-%s-%d", finger, coinCode, accountNumber))
}

// ERC20Code makes an ERC20 code where ethAccountCode is token's parent account.
func ERC20Code(ethAccountCode Code, tokenCode string) Code 
```

It groups them together and adds doc comments explaining which is used where. Also, easier to trace and have rest of the code to always refer to the same calling place instead of an inlined fmt.Sprintf.

x1ddos (approved):
LGTM but see comments.

backend/backend.go
benma: Excellent idea. Done

backend/coins/btc/handlers/handlers.go
benma: :+1: done

backend/handlers/handlers.go
benma: :+1: done

PR #1319, opened 2021-06-01
Account unit tests

benma:


backend/accounts_test.go
```diff
@@ -179,3 +182,86 @@ func TestNextAccountNumber(t *testing.T) {
...
 	require.Equal(t, ErrAccountLimitReached, errp.Cause(err))
 }
+
+func newBackend(t *testing.T, testing, regtest bool) *Backend {
+	t.Helper()
+	b, err := NewBackend(
+		arguments.NewArguments(
+			test.TstTempDir("appfolder"),
+			testing, regtest,
+			false, false,
```
x1ddos, line 201: It would be so much nicer if arguments were just a struct: I'll have to go and look up what `NewArguments` args are now...

backend/accounts_test.go
```diff
@@ -179,3 +182,86 @@ func TestNextAccountNumber(t *testing.T) {
...
+			&types.GapLimits{Receive: 20, Change: 6}),
+		nil,
+	)
+	require.NoError(t, err)
+	return b
+}
+
+func TestSupportedCoins(t *testing.T) {
+	// All coins supported, mainnet.
+	t.Run("", func(t *testing.T) {
```
x1ddos, line 211: The comments should actually be sub test names. It's what it's made for when watching tests fail - tells you which one:

```go
t.Run("all coins supported, mainnet", ...)
t.Run("all coins supported, testnet", ...)
t.Run("all coins supported, regtest", ...)
...
```

backend/accounts_test.go
```diff
@@ -179,3 +182,86 @@ func TestNextAccountNumber(t *testing.T) {
 	_, err = nextAccountNumber(coinpkg.CodeTBTC, ks(fingerprint2, true), accountsConfig)
 	require.Equal(t, ErrAccountLimitReached, errp.Cause(err))
 }
+
+func newBackend(t *testing.T, testing, regtest bool) *Backend {
```
x1ddos, line 195: Consider defining consts for the booleans:

```go
const (
  testnetEnabled = true
  testnetDisabled = false
  regtestEnabled = true
  regtestDisabled = false
)
```

It just makes reading anywhere else easier:

```go
newBackend(t, testnetEnabled, regtestDisabled)
NewArguments(.., testnetDisabled, regtestDisabled, ...)
```

Also, `testing` sounds like it's running some tests? It's testnet, no?

x1ddos (approved):
LGTM

backend/accounts_test.go
benma: Done.


> Also, testing sounds like it's running some tests? It's testnet, no?

What do you mean? It enables testnet mode, which means that tbtc, tltc, teth, reth are available and the mainnet coins are not.

backend/accounts_test.go
benma: Done

PR #1320, opened 2021-06-01
backend: explictly only support one keystore

benma:
We implicitly only support one keystore (e.g. RegisterKeystore()
resets previous keystores, DeregisterKeystore() does not
have a keystore ID argument, etc). Having a keystores type handling
multiple keystores is more confusing than helpful, as most of the
behavior is not well defined, since we don't actually make use of it.

This patch makes this explicit and simplifies some code in the
process.

The accountsLock is extended to cover the keystore as well.


backend/backend.go
```diff
@@ -157,8 +158,8 @@ type Backend struct {
 	coins     map[coinpkg.Code]coinpkg.Coin
 	coinsLock locker.Locker
 
-	accounts     []accounts.Interface
-	accountsLock locker.Locker
+	accounts                []accounts.Interface
+	accountsAndKeystoreLock locker.Locker
```
x1ddos, line 162: Normally, you put the mutex right above what it guards for clarity, separated by blank lines:

```go
type Backend struct {
  ...

  accountsKeystoreLock locker.Locker
  accounts []accounts.Interface
  keystore keystore.Keystore

  log
  ...
}
```

backend/backend.go
```diff
@@ -822,17 +816,18 @@ func (backend *Backend) uninitAccounts() {
...
 
 // registerKeystore registers the given keystore at this backend.
 func (backend *Backend) registerKeystore(keystore keystore.Keystore) {
+	defer backend.accountsAndKeystoreLock.Lock()()
+
 	backend.log.Info("registering keystore")
-	if err := backend.keystores.Add(keystore); err != nil {
-		backend.log.Panic("Failed to add a keystore.", err)
-	}
+	backend.keystore = keystore
```
x1ddos, line 830: Just checking: what if backend.keystore is not nil, where does the "old" keystore go? Shouldn't you check and return an error? oh wait... there's no returned error :)

Or it doesn't matter? I know you might say this can't happen but still... maybe at least log a message?

backend/backend.go
```diff
@@ -822,17 +816,18 @@ func (backend *Backend) uninitAccounts() {
...
+func (backend *Backend) Keystore() keystore.Keystore {
+	defer backend.accountsAndKeystoreLock.RLock()()
+	return backend.keystore
 }
 
 // registerKeystore registers the given keystore at this backend.
 func (backend *Backend) registerKeystore(keystore keystore.Keystore) {
+	defer backend.accountsAndKeystoreLock.Lock()()
+
 	backend.log.Info("registering keystore")
```
x1ddos, line 829: how about adding keystore.RootFingerprint() to the message? might help with debugging? and the same in "deregestering keystore".

x1ddos (approved):
LGTM. Nice simplification.

backend/backend.go
benma: :heavy_check_mark:

backend/backend.go
benma: It will simply be replaced. In devmode, this happens for example if you register a software keystore and then unlock a bb02.

I added a comment to be explicit about this.

backend/backend.go
benma: Good idea, done

PR #1321, opened 2021-06-03
frontend: support deactivating account prototype

thisconnect:
In BBApp 4.27 altcoins could be deactivated by a toggle in
the settings view. With multiaccounts support this got lost,
once an account is created it cannot be removed again.

BIP44 accounts can't technically be deleted, coins would
still be there even if the account is removed. Also remove
could be scary as a user might be affraid that coins could
be lost.

Added toggles to deactivate coins on the manage accounts
view.

TODO:
- backend api set-coin-active
- indicate or warn when deactivating an account that has coins
- add error when the last account is deactivated, at least 1
  account must be active

### DIFF
--- a/frontends/web/src/api/backend.ts
+++ b/frontends/web/src/api/backend.ts
@@ -34,6 +34,11 @@ export const getSupportedCoins = (): Promise<ICoin[]> => {
     return apiGet('supported-coins');
 };
 
+// TODO: implemnt backend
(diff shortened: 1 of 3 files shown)


benma:
Waiting for the unit-test PRs to be merged, as there is a semantic merge conflict with this PR (about reinitalizeaccounts), and I want to add the activation logic with a unit test.

benma:
Rebased and re-opened here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1337

PR #1323, opened 2021-06-03
backend/rates: reduce max range during initial backfill to 1 year

x1ddos:
The fetchGeckoMarketRange limits a single response to 1Mb or 1048576
bytes. It turns out a dataset for 2 years is about 1064916 bytes, as
observed in the logs of our coingecko mirror.

I believe the limits were ok at first but our mirror is now serving
rates at 5min interval data points for any range. Compared to
hourly/daily rates, this is a lot more data which doesn't fit into the
1Mb response limit.

Instead of increasing the response limit, I think it's best to reduce
initial backfill range. It doubles the backfill QPS but makes response
size smaller - faster network roundtrip, less room for failed requests
when internet connection is bad.


benma (approved):
tACK

PR #1322, opened 2021-06-03
More unit tests

benma:
Builds on https://github.com/digitalbitbox/bitbox-wallet-app/pull/1320 and https://github.com/digitalbitbox/bitbox-wallet-app/pull/1319

PR #1328, opened 2021-06-07
backend: check account support by keystore also when loading

benma:


benma:
Builds on #1322

x1ddos:
> Edtion switch

Did you mean Edition? and by edition, did you mean... Rust edition?

benma:
> > Edtion switch
> 
> Did you mean Edition? and by edition, did you mean... Rust edition?



> > Edtion switch
> 
> Did you mean Edition? and by edition, did you mean... Rust edition?

Renamed the title.. Sorry sometimes I just keep the default name which is based off my branch name...

x1ddos (approved):
Oh, very good find :+1: 

On a related note, I wonder what happens in reverse. A btc-only restored on the regular. Since the initial default accounts were stored while using btc-only, so only BTC, will the user be able to see and add other coin accounts, i.e. ETH?

x1ddos:
> will the user be able to see and add other coin accounts, i.e. ETH?

I mean I know they should be. Just checking if there are any potential bugs.

benma:
> I mean I know they should be. Just checking if there are any potential bugs.

Yes it works - one can add the altcoins, but they won't be added by default - that only happens if there are no persisted accounts at all yet.

PR #1329, opened 2021-06-07
Frontend revamp accountinfo

thisconnect:


benma:
Just testing a bit: if there are three xpubs it's not working well:

image

Also, in Ethereum and tokens, you show the xpub, but is that relevant information? I don't think any ETH wallet takes xpubs for anything, but I could be wrong. Maybe better to simply show the address again?

benma:
Also in ETH/tokens, the guide entry should be removed, as it only works for BTC/LTC:

image

thisconnect:
> if there are three xpubs it's not working well

fixed, it should now support any amonut of xpubs, as long as there are translations for it.

> Also, in Ethereum and tokens, you show the xpub, but is that relevant information?

removed xpub and added a button to route to the receive view to display the eth address

> Also in ETH/tokens, the guide entry should be removed, as it only works for BTC/LTC:

updated guide entry for bitcoin based coins, in PR
https://github.com/digitalbitbox/bitbox-wallet-app/pull/1334

benma (changes requested):

frontends/web/src/api/account.ts
```diff
@@ -63,21 +63,29 @@ export type ScriptType = 'p2pkh' | 'p2wpkh-p2sh' | 'p2wpkh';
...
+    scriptType: ScriptType;
+}
+
+export type TEthereumSimple = {
+    keyInfo: IKeyInfo;
+}
+
+export type TSigningConfiguration = {
+    bitcoinSimple: TBitcoinSimple;
+    ethereumSimple?: never;
```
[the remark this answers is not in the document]
thisconnect: yes I seems like it is needed afaik

```
✖ ERROR ./src/routes/account/account.tsx
ERROR in ./src/routes/account/account.tsx(238,61):
TS2339: Property 'bitcoinSimple' does not exist on type 'TSigningConfiguration'.
  Property 'bitcoinSimple' does not exist on type '{ ethereumSimple: TEthereumSimple; }'.
error Command failed with exit code 1.
```

frontends/web/src/locales/en/app.json
```diff
@@ -24,7 +24,17 @@
     "extendedPublicKey": "Extended public key",
     "label": "Account info",
     "title": "Account information",
-    "verify": "Verify on device"
+    "verify": "Verify on device",
+    "xpubTypeChangeBtn": {
+      "p2pkh": "View legacy P2PKH extended public key",
+      "p2wpkh": "Switch back to Native segwit (bech32)",
```
benma, line 30: maybe also `View` instead of `Switch back`? then the order does not matter.

frontends/web/src/locales/en/app.json
```diff
@@ -24,7 +24,17 @@
...
+      "p2wpkh-p2sh": "View older segwit extended public key"
+    },
+    "xpubTypeInfo": {
+      "p2pkh": "Currently displaying legacy P2PKH extended public key ({{current}} of {{numberOfXPubs}})",
```
benma, line 34: what if you used only one translation `Currently displaying {{scriptType}} extended public key ({{current}} of {{numberOfXPubs}})`?

This would avoid many translations, and also we don't need to come up with half-accurate names for the script types (e.g. 'segwit extended public key' is technically also true for taproot, as taproot is also segwit, etc).

frontends/web/src/routes/account/info/guide.tsx
```diff
@@ -0,0 +1,51 @@
...
+    t: Translate;
+}
+
+export function BitcoinBasedAccountInfoGuide({
+    t,
+    coinName,
+}: BitcoinBasedAccountInfoGuideProps) {
+    return (
+        <Guide>
+            <Entry key="guide.accountInfo.xpub" entry={{
```
benma, line 33: could just be `entry={t('guide.accountInfo.xpub')}` i think

frontends/web/src/locales/en/app.json
```diff
@@ -491,9 +501,21 @@
       "title": "What does incoming mean?"
     },
     "accountInfo": {
+      "multipleXPubs": {
+        "text": "Each xpub is tied to the \"Type\" shown: either \"Native segwit (bech32)\" or \"Wrapped segwit\". These are script types used by {{coinName}}. The BitBoxApp combines them, supporting multiple script types in the same account. But because each script type gives a different xpub, there are multiple xpubs per account.\n\nIf you consistently receive on the standard address, you only need the \"bech32\" xpub. However, if you also receive funds on the \"compatible address\" type, these will be shown on the \"older segwit\" extended public key.",
```
benma, line 505: > But because

afaik starting a sentence in english with 'But' is not proper grammar, you can just remove the word and it works the same way. same below.

benma (approved):

frontends/web/src/routes/account/info/info.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    <Header title={<h2>{t('accountInfo.title')}</h2>} />
+                    <div class="innerContainer scrollableContainer">
+                        <div class="content padded">
+                            <div class="box larger">
+                                { isBitcoinBased(account.coinCode) ? (
+                                    <h2 className={style.title}>
+                                        {t('accountInfo.extendedPublicKey')}
+                                    </h2>
+                                ) : null }
+                                { numberOfXPubs > 1 ? (
```
benma, line 104: maybe ` && config.bitcoinSimple`? then the `?` below on line 106 maybe is not needed, if the compiler is smart enough.

frontends/web/src/routes/account/info/info.tsx
```diff
@@ -0,0 +1,140 @@
...
+        if (!this.state.info) {
+            return;
+        }
+        const numberOfXPubs = this.state.info.signingConfigurations.length;
+        this.setState(({ viewXPub }) => ({
+            viewXPub: (viewXPub + 1) % numberOfXPubs
+        }));
+    }
+
+    private xpubTypes = ['p2wpkh', 'p2wpkh-p2sh', 'p2pkh']
```
benma, line 82: the order could be wrong. simply get it from the configs?

```
info.signingConfigurations.map(cfg => cfg.bitcoinSimple?.scriptType)
```

inside render()

benma (changes requested):

frontends/web/src/locales/en/app.json
benma: then adding another type like taproot would work automatically too.

frontends/web/src/locales/en/app.json
thisconnect: I prefer `Switch back` as it goes back to the default one, but looking forward when we add another scripttype you're right it's better to make them all the same.

frontends/web/src/routes/account/info/info.tsx
thisconnect: No I'd never use `&&` for conditional rendering, reason: if `numberOfXPubs` is the number 0 then JSX would actually render `0` and you'd want to render nothing.

benma: >  if numberOfXPubs is the number 0 then JSX would actually render 0 and you'd want to render nothing.

that would be with `||` not with`&&`?

frontends/web/src/locales/en/app.json
thisconnect: yes and I'll move the more human readable name to utils and use it here ok?

```diff
diff --git a/frontends/web/src/routes/account/utils.ts b/frontends/web/src/routes/account/utils.ts
index 515b7732..0b409441 100644
--- a/frontends/web/src/routes/account/utils.ts
+++ b/frontends/web/src/routes/account/utils.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { CoinCode } from '../../api/account';
+import { CoinCode, ScriptType } from '../../api/account';
 
 export function isBitcoin(code: string): boolean {
     switch (code) {
@@ -63,3 +63,15 @@ export function getCoinCode(coinCode: CoinCode): CoinCode | undefined {
             return 'eth';
     }
 }
+
+
+export function getScriptTypeName(scriptType: ScriptType): string {
+    switch (scriptType) {
+        case 'p2pkh':
+            return 'Legacy';
+        case 'p2wpkh-p2sh':
+            return 'Segwit';
+        case 'p2wpkh':
+            return 'Native segwit (bech32)';
+    }
+}
```

frontends/web/src/routes/account/info/guide.tsx
thisconnect: Right. Keep in mind link text do not fallback that's why we can't use the shorthand version here https://github.com/digitalbitbox/bitbox-wallet-app/commit/9b43641ee6b379b9c5f080cdc49a733baa6847ad#diff-7ec1c85f063f25e402825aad376083870f9cf75cdc90c3792d73c60902ddec96

frontends/web/src/locales/en/app.json
benma: okay but drop the (bech32) maybe? Afaik this word is not explained anywhere.

thisconnect: Good point, I'll change back, but keep the script type name function it in utils.

frontends/web/src/routes/account/info/info.tsx
thisconnect: Updated had to change to `{ (config.bitcoinSimple !== undefined && numberOfXPubs > 1) ? (` to make TypeScript happy, PTAL

thisconnect:
PTAL

frontends/web/src/routes/account/info/signingconfiguration.tsx
```diff
@@ -37,64 +39,110 @@ interface State {
...
+                            <code>{config.keyInfo.keypath}</code>
+                        </p>
+                        { ('scriptType' in config) ? (
+                            <p key="accountType" className={style.entry}>
+                                <strong>Type:</strong>
+                                <span>{getScriptName(config.scriptType)}</span>
+                            </p>
+                        ) : null}
+                        { ('scriptType' in config) ? (
+                            <p key="accountType" className={style.entry}>
```
benma, line 108: duplicate key (i got warnings in the console).

benma (approved):
Nice! LGTM after fixing the duplicate key.

frontends/web/src/routes/account/info/signingconfiguration.tsx
thisconnect: Good catch thanks

PR #1330, opened 2021-06-07
frontend/accounts/add: suggested name missing for Bitcoin-only

benma:
The suggested (default) account name was only set when choosing the
coin from the dropdown. In BitBox02 Bitcoin-only, the default name was missing.


thisconnect (approved):
LGTM

PR #1336, opened 2021-06-14
backend: add SetAccountActive

benma:
Enables activation and deactivation of accounts.


x1ddos (approved):
LGTM but `manageAccountsLen` looks silly to me :) @thisconnect

PR #1339, opened 2021-06-15
aopp: register URI scheme on linux

benma:
See
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#mime-types

Installing the .deb package automatically registers the scheme based
on this entry in the .desktop file. Manual changes to the desktop file
require calling `sudo update-desktop-database`.


x1ddos (approved):
LGTM

PR #1340, opened 2021-06-16
aopp: register URI scheme on Windows

benma:
It seems there can only be one app for one protocol. We set it if no
other app is registered, or otherwise ask the user for permission.

### DIFF
--- a/frontends/qt/setup.nsi
+++ b/frontends/qt/setup.nsi
@@ -14,6 +14,7 @@ SetCompressor /SOLID lzma
 !define BINDIR "build\windows"
 !define ICONDIR "resources\win"
 !define APP_EXE "BitBox.exe"
+!define AOPP_EXE "$\"$INSTDIR\${APP_EXE}$\" $\"%1$\""
 
 # MUI Symbol Definitions
 !define MUI_ICON "${ICONDIR}\icon.ico"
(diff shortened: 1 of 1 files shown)


frontends/qt/setup.nsi
```diff
@@ -106,6 +107,23 @@ Section -post SEC0001
     WriteRegStr HKCU "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$(^Name)" UninstallString $INSTDIR\uninstall.exe
     WriteRegDWORD HKCU "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$(^Name)" NoModify 1
     WriteRegDWORD HKCU "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$(^Name)" NoRepair 1
+
+    # Links aopp: URI scheme.
+    # It links it silently if no other app is registered, to not overwrite.
+    # If another app is registered, we ask the user for permission.
+    ReadRegStr $0 HKCU "SOFTWARE\Classes\aopp\shell\open\command" ""
+    ${If} $0 != "${AOPP_EXE}"
```
x1ddos, line 115: One problem with this is during app upgrade. Users can choose the install location, right. If it's different from the previous install, `AOPP_EXE` won't match. But it's an issue bigger than this PR. I guess it's good as is.

x1ddos (approved):
Nice! Very user-caring way of doing it.

frontends/qt/setup.nsi
benma: Hmm good point. In that case the user is asked, which sounds good enough.

PR #1342, opened 2021-06-17
Aopp scheme macos

benma:


frontends/qt/main.cpp
```diff
@@ -32,6 +32,35 @@ static WebClass* webClass;
...
+            QFileOpenEvent* openEvent = static_cast<QFileOpenEvent*>(event);
+            if (!openEvent->url().isEmpty()) {
+                // This is only supported on macOS and is used to handle URIs that are opened with
+                // the BitBoxApp, such as "aopp:..." links. The event is received and handled both
+                // if the BitBoxApp is launched and also when it is already running, in which case
+                // it is brought to the foreground automatically.
+
+                // TODO: handle the URI.
+                // After removing the messagebox here, remove the QMessageBox include above.
+                QMessageBox::information(NULL, QString("Handle URI"), openEvent->url().toString());
```
x1ddos, line 55: How about something like this instead:

```cpp
std::cerr << "Handle URI: " << openEvent->url().toString() << std::endl
```

- no need for qmessagebox
- might even keep it in the release; we already log blocked `systemOpen` URLs in this way
- feels much less distruptive even if it's in a staging branch

frontends/qt/main.cpp
```diff
@@ -32,6 +32,35 @@ static WebClass* webClass;
...
+    BitBoxApp(int &argc, char **argv): QApplication(argc, argv)
+    {
+    }
+
+#if defined(Q_OS_MACOS)
+    bool event(QEvent *event) override
+    {
+        if (event->type() == QEvent::FileOpen) {
+            QFileOpenEvent* openEvent = static_cast<QFileOpenEvent*>(event);
+            if (!openEvent->url().isEmpty()) {
```
x1ddos, line 47: was a bit worried `openEvent->url()...` could generate a segfault but from https://github.com/qt/qtbase/blob/31f90e9/src/gui/kernel/qevent.h#L927-L943 and https://github.com/qt/qtbase/blob/31f90e9/src/gui/kernel/qevent.cpp#L3615-L3626 it seems QFileOpenEvent's `m_url` is always valid :+1:

x1ddos (approved):
LGTM but would replace qmessagebox with `std::cerr << ...`

benma:
> LGTM but would replace qmessagebox with `std::cerr << ...`

It's in staging and temporary until it is handled for real.

It can't be on stdout/stderr because you can't see that on the first launch.

PR #1343, opened 2021-06-17
backend: allow links to shiftcrypto.support and shiftcrypto.shop

thisconnect:
shiftcrypto.support should be allowed so we can slowly move to
the new knowledge base, this will replace guides.shiftcrypto.ch.
Keeping guides as there are still some older links pointing to
it.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -63,8 +63,9 @@ var fixedURLWhitelist = []string{
 	// Shift Crypto owned domains.
 	"https://shiftcrypto.ch/",
 	"https://ext.shiftcrypto.ch/",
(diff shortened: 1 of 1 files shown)


backend/backend.go
```diff
@@ -65,6 +65,7 @@ var fixedURLWhitelist = []string{
 	"https://ext.shiftcrypto.ch/",
 	"https://shop.shiftcrypto.ch/",
```
x1ddos, line 66: Since you're here, please replace this with `https://shiftcrypto.shop/`.

thisconnect: done

x1ddos (approved):
LGTM

PR #1349, opened 2021-06-21
frontend/subscribe: remove unused actions 

benma:


frontends/web/src/utils/event.ts
```diff
@@ -24,7 +24,7 @@ export type Subject = string;
 /**
  * This type enumerates the various actions of an event.
  */
-export type Action = 'replace' | 'prepend' | 'append' | 'remove' | 'reload';
+export type Action = 'notify' | 'replace' | 'reload';
```
thisconnect, line 27: What is 'notify' ?

benma: Oops good catch. It is something I might add for AOPP - a simple notification from backend to frontend, without a generic action like 'replace' or 'reload'.

thisconnect: I mean I get the intention, but technically replace has a payload and the frontend can just set this as the new state, reload tells the frontend to GET some data, what should notify do?

benma: It is not final yet, so I might not use it, but it would simply be to trigger the frontend to do something which is not simply setting the state from the payload.

Was just playing though, I am still trying to figure out how to best invoke the aopp workflow once the app is launched through an aopp click.

thisconnect (approved):
LGTM

PR #1350, opened 2021-06-21
Frontend cleanup multiaccounts

thisconnect:
- added missing divider on the device setting view
- bitcoin crypto naming changes due to multi acccount

### DIFF
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -27,7 +27,7 @@ import { translate, TranslateProps } from '../../decorators/translate';
 import { Button, Checkbox, Select } from '../../components/forms';
 import { setConfig } from '../../utils/config';
 import { apiGet } from '../../utils/request';
-import { isBitcoin } from '../account/utils';
+import { isBitcoinOnly } from '../account/utils';
 import * as style from './info.css';
 
 interface BuyInfoProps {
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -98,9 +98,18 @@ class BuyInfo extends Component<Props, State> {
         setConfig({ frontend: { skipBuyDisclaimer: e.target.checked }});
     }
 
+    private getCryptoName = (): string => {
+        const { accounts, code, t } = this.props;
+        if (!code) {
+            const onlyBitcoin = accounts.every((({ coinCode }) => isBitcoinOnly(coinCode)));
```
benma, line 104: nit: one set of `()` too many i think

frontends/web/src/routes/buy/info.tsx
```diff
@@ -98,9 +98,18 @@ class BuyInfo extends Component<Props, State> {
...
+            return onlyBitcoin ? 'Bitcoin' : t('buy.info.crypto');
+        }
+        const account = accounts.find(account => account.code === code);
+        return isBitcoinOnly(account!.coinCode) ? 'Bitcoin' : t('buy.info.crypto');
```
benma, line 108: `account!.coinCode` could be undefined, but `isBitcoinOnly` does not take undefined? Maybe better to catch the undefined case before separately.

benma (approved):
utACK

frontends/web/src/routes/buy/info.tsx
thisconnect: Good catch 👍

frontends/web/src/routes/buy/info.tsx
thisconnect: Find could return `undefined` if it would not find the account, but this should never happen. Reason `code` is only set if the user clicked on "Buy" inside the account, so the account should be there.

Added a check `if (account) {` anyways, to make 100% sure we have an account here and removed the `!` and added a fallback to `buy.info.crypto`

PR #1352, opened 2021-06-21
backend/keystore: add BTCSignMessage

benma:
Needed for AOPP.


x1ddos (approved):
LGTM but you'll want to re-gen the mock. Also, no ETH signing?

PR #1358, opened 2021-06-24
backend/aopp: support ETH requests

benma:
Can test with https://testing.aopp.group/?asset=eth, but only after the site is updated a little later today - ~currently it has a bug in ETH signature verification that will be fixed.~ currently we don't know if the recoverable id in Ethereum Signed messages should be offset by 27 or not (MEW [accepts both](https://github.com/MyEtherWallet/MyEtherWallet/blob/052f697f3b269db4378f331dbb89d24b406bb7c7/src/components/VerifyMessageInput/VerifyMessageInput.vue#L118)) the AOPP demo site expects it currently)

### DIFF
--- a/backend/aopp.go
+++ b/backend/aopp.go
@@ -31,8 +31,7 @@ import (
 // aoppCoinMap maps from the asset codes specified by AOPP to our own coin codes.
 var aoppCoinMap = map[string]coinpkg.Code{
 	"btc": coinpkg.CodeBTC,
-	// TODO: add support for ETH
-	// "eth": coinpkg.CodeETH,
+	"eth": coinpkg.CodeETH,
 }
 
 type account struct {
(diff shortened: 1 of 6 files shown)


backend/aopp.go
```diff
@@ -268,6 +267,22 @@ func (backend *Backend) AOPPChooseAccount(code accounts.Code) {
...
 		}
 		signature = sig
+	case coinpkg.CodeETH:
+		sig, err := backend.keystore.SignETHMessage(
+			[]byte(backend.aopp.message),
+			addr.AbsoluteKeypath(),
+		)
+		if err != nil {
+			if firmware.IsErrorAbort(err) {
+				backend.log.WithError(err).Error("user aborted msg signing")
```
x1ddos, line 277: you made a local `log` var on line 220 but continue to use `backend.log`. I think you meant to use the local `log` everywhere.

x1ddos (approved):
LGTM but see comment about the local `log` var.

benma:
The demo works if `27` is added to the recoverable ID (last byte of the sig). If we decide to add 27 on our side, we can update bitbox02-api-go - this PR needs to further changes.

PR #1360, opened 2021-06-25
frontend/bitbox02: add button to go to startup settings

benma:
So users can toggle the startup firmware hash display.

The acutal bootloader page in the app needs to be improved to not show
the firmware upgrade if it is already up to date. This follows in
another commit.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/settings.tsx
+++ b/frontends/web/src/routes/device/bitbox02/settings.tsx
@@ -22,6 +22,7 @@ import { Footer } from '../../../components/layout';
 import { Header } from '../../../components/layout/header';
 import { SettingsButton } from '../../../components/settingsButton/settingsButton';
 import { SettingsItem } from '../../../components/settingsButton/settingsItem';
+import { GotoStartupSettings } from './gotostartupsettings';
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/device/bitbox02/settings.tsx
```diff
@@ -150,6 +151,9 @@ class Settings extends Component<Props, State> {
                                                 mnemonicPassphraseEnabled={deviceInfo.mnemonicPassphraseEnabled}
                                                 getInfo={this.getInfo} />
                                         </div>
+                                        <div className="box slim divide">
```
thisconnect, line 154: Add it to the same box as the MnemonicPssphraseButton, see also https://github.com/digitalbitbox/bitbox-wallet-app/commit/a70b736ce7ee69a10604c6638d868b5fc9c52579

thisconnect (approved):
LGTM with small nit

frontends/web/src/routes/device/bitbox02/settings.tsx
benma: Done!

PR #1361, opened 2021-06-28
go: update Go to 1.16.5

benma:


benma:
I tested Linux and Android manually, both work.

README.md
```diff
@@ -47,7 +47,7 @@ The below instructions assume a unix environment.
 
 The following dependencies need to be installed:
 
-- [Go](https://golang.org/doc/install) version 1.14
+- [Go](https://golang.org/doc/install) version 1.16
```
x1ddos, line 50: docs/BUILD.md also references go1.14.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/digitalbitbox/bitbox-wallet-app
 
-go 1.14
+go 1.16
```
x1ddos, line 3: Also FYI here: makes sense to bump only if the plan is to use newer features of go introduced in 1.16. https://golang.org/doc/modules/gomod-ref

x1ddos (approved):
LGTM

go.mod
benma: What about potential security updates?

x1ddos: those are covered by patch releases: https://golang.org/doc/devel/release#go1.14.minor

PR #1364, opened 2021-06-28
aopp: use the format hint to deliver a compatible address

benma:


backend/aopp_test.go
```diff
@@ -122,6 +124,7 @@ func TestAOPPSuccess(t *testing.T) {
 			Address:      expectedAddress,
 			CallbackHost: callbackHost,
 			coinCode:     coinpkg.CodeBTC,
+			format:       "any",
```
x1ddos: Would be very nice to have one test with a specific format and another one with uknown format to get `errAOPPUnsupportedFormat`.

x1ddos (approved):
LGTM

PR #1366, opened 2021-06-30
frontend/aopp: remove unused/wrong state

benma:
Was a leftover from a previous version before putting the state into
the SubscribedProps.


thisconnect (approved):
LGTM

PR #1365, opened 2021-06-28
AOPP unit test failures

benma:

PR #1367, opened 2021-06-30
backend/aopp: add user-approval state 

benma:
If an AOPP link is clicked while no BitBox is connected, we do as
before and ask the user to insert and unlock.

However, if the BitBox is already inserted and unlocked, we don't want
a link click to immediately go to the account-chooser, but have an
intro dialog first asking the user if they want to proceed.

Builds on https://github.com/digitalbitbox/bitbox-wallet-app/pull/1365


x1ddos (approved):
LGTM but consider moving error sub tests to their own test func.

PR #1370, opened 2021-07-02
frontend: prevent flicker when unlocking the device

thisconnect:
The BBApp did route back to / after unlocking the device, that
caused a quick flicker because it rendered the waiting view one
more time.

Unlocking can directly be routed to account-summary view, to
prevent flickering.

- removed noAccount on account summary in favor of loading msg
- updated buy.info.noAccount wording

TODO: update locize

### DIFF
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -244,7 +244,11 @@ class AccountsSummary extends Component<Props, State> {
                                         { accounts.length > 0 ? (
                                             accounts.map(account => this.balanceRow(account))
                                         ) : (
-                                            <p>{t('accountSummary.noAccount')}</p>
+                                            <tr>
+                                                <td colSpan={3} className={style.loadingAccounts}>
+                                                    {t('loading')}
+                                                </td>
+                                            </tr>
                                         )}
                                     </tbody>
                                     {(data && data.chartTotal) ? (
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -195,8 +195,7 @@ class BitBox02 extends Component<Props, State> {
                 errorText: undefined,
             });
             if (status === 'initialized' && unlockOnly && showWizard) {
-                // bitbox is unlocked, now route to / and wait for incoming accounts
-                route('/', true);
+                route('/account-summary', true);
             }
         });
     }
--- a/frontends/web/src/routes/account/summary/accountssummary.css
+++ b/frontends/web/src/routes/account/summary/accountssummary.css
@@ -62,6 +62,11 @@
     white-space: nowrap;
 }
 
+.table tr td.loadingAccounts {
+    color: var(--color-gray);
+    text-align: center;
+}
+
 .table tfoot th,
 .table tfoot td {
     background-color: var(--color-mediumgray);


frontends/web/src/locales/en/app.json
```diff
@@ -37,6 +37,7 @@
     "balance": "Balance",
     "exportSummary": "Export accounts summary to downloads folder as CSV file",
     "fiatBalance": "Fiat balance",
+    "loadingAccounts": "Loading accounts",
     "name": "Account name",
     "noAccount": "There are no accounts to show.",
```
thisconnect, line 42: Not sure if we still need "noAccount" or when to show it

thisconnect:
Tested also with `make qt-osx`

frontends/web/src/locales/en/app.json
x1ddos: my grep shows `noAccount` is also used in [frontends/web/src/routes/buy/info.tsx](https://github.com/digitalbitbox/bitbox-wallet-app/blob/c6aab96/frontends/web/src/routes/buy/info.tsx#L207) although in the buy context maybe a better text would be: "There are no suitable accounts to choose from".

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -244,7 +244,11 @@ class AccountsSummary extends Component<Props, State> {
                                         { accounts.length > 0 ? (
                                             accounts.map(account => this.balanceRow(account))
                                         ) : (
-                                            <p>{t('accountSummary.noAccount')}</p>
+                                            <tr>
+                                                <td colSpan={3} className={style.loadingAccounts}>
+                                                    {t('accountSummary.loadingAccounts')}
```
x1ddos, line 249: How about reusing a more generic [loading](https://github.com/digitalbitbox/bitbox-wallet-app/blob/c6aab96/frontends/web/src/locales/en/app.json#L893) key? I don't think it matters that much here, what's loading.

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -195,8 +195,7 @@ class BitBox02 extends Component<Props, State> {
                 errorText: undefined,
             });
             if (status === 'initialized' && unlockOnly && showWizard) {
-                // bitbox is unlocked, now route to / and wait for incoming accounts
-                route('/', true);
+                route('/account-summary', true);
```
x1ddos, line 198: Will this work nicely with AOPP where, after unlocking, it proceeds to the addr. verification flow?

thisconnect: Before it routed to `/` and then to `/account-summary`,
with this change it just goes directly to `/account-summary`.

AOPP just covers everything no matter what it is, expect if the user needs help unlocking/settingup, then aopp can minimize into a small bar at the top or something else.

frontends/web/src/locales/en/app.json
thisconnect: update, but I am still not sure if there ever is a state on the accountSummary where we should show the noAccount (There are no accounts to show.) message.

x1ddos (approved):
LGTM but I would split the changes around `noAccount` into a separate PR - they seem completely unrelated to flickering.

thisconnect:
> LGTM but I would split the changes around noAccount into a separate PR - they seem completely unrelated to flickering.

Removed, they have nothing todo with this and has to be changed in locize in a different PR.

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -244,7 +244,11 @@ class AccountsSummary extends Component<Props, State> {
                                         { accounts.length > 0 ? (
                                             accounts.map(account => this.balanceRow(account))
                                         ) : (
-                                            <p>{t('accountSummary.noAccount')}</p>
```
benma, line 247: why? if accounts.length is 0 it means there are no accounts, not that they are loading.

if you want to distinguish between loading and no accounts, you need to make the accounts undefined before it is set (i.e. treat undefined abd `[]` differently)

Also if if is empty, then won't it route back to `/` again?

https://github.com/digitalbitbox/bitbox-wallet-app/blob/d45cba211340be929f1fbecaaaadebac8ff510ba/frontends/web/src/app.tsx#L139-L146

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1599

PR #1371, opened 2021-07-06
frontend: fix tx progress icon

thisconnect:
Some longer tx statusText translations pushed the progress
ring icon, so that it was rendered smaller than the specified width.
This only happens in a app size that sows the tx overview as table.

Added an inline style to set a min-width and prevent longer
texts shrinking the icon. Longer statusText will overlap slightly
on the right side.


x1ddos (approved):
Thanks!!

PR #1374, opened 2021-07-07
backend/aopp: always start with aoppStateUserApproval

benma:
We changed our minds and will always ask the user if they want to
continue. Before, we only asked if there was already an unlocked
BitBox02.


thisconnect (approved):
LGTM

PR #1379, opened 2021-07-09
frontend: move commonly used icons to icon component

thisconnect:
Many components contain inline SVG (in JSX), instead of using
an external asset and link with image source. Having SVG in JSX is
sometimes usefull if some attributes change at runtime, i.e. the
tx progress ring. But having static SVG in JSX hurts memory usage
and rendering performance, for example the vdom diff has to check
all of the static SVG attributes on every render and it will
increase the JavaScript bundle size.
See also https://twitter.com/_developit/status/1382838799420514317

Moved some icons to the icon component assets and changed the props
typing so TypeScript can check valid HTML attributes.


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -59,7 +58,7 @@ interface SignProgress {
 type Props = SendProps & TranslateProps;
 
 interface State {
-    account?: Account;
+    account?: accountApi.IAccount;
```
thisconnect, line 61: Not sure why TS didn't complain about this

benma (approved):
LGTM, though could you explain what's wrong with using the SVG directly?

thisconnect:
Updated commit message and PR description.

PR #1380, opened 2021-07-09
frontend: rewrite more form components to typescript

thisconnect:
- start using JSX.IntrinsicElements with 'div' or 'input' etc.
  so that rest props are type checked
- removed textMedium radio prop, this was always used, so it is
  now the default style
- fixed TypeScript recommendations in Backup component, the only
  component that actually uses the radio component (sometimes)

### DIFF
--- a/frontends/web/src/components/forms/field.tsx
+++ b/frontends/web/src/components/forms/field.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,18 +15,15 @@
  * limitations under the License.
  */
 
-import { h } from 'preact';
+import { h, JSX } from 'preact';
+import * as style from './field.css';
 
-export default function Label({
-    className,
-    style,
-    children,
-    ...props
-}) {
-    const classes = ['label', className].join(' ');
+export function Field({
+    children, ...props
+}: JSX.IntrinsicElements['div']) {
     return (
-        <label className={classes} style={style} {...props}>
+        <div className={style.field} {...props}>
             {children}
-        </label>
+        </div>
     );
 }
--- a/frontends/web/src/components/forms/label.tsx
+++ b/frontends/web/src/components/forms/label.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


frontends/web/src/components/forms/label.tsx
```diff
@@ -14,17 +15,18 @@
...
-import { h } from 'preact';
+import { h, JSX } from 'preact';
+import * as style from './label.css';
 
-export default function Label({
+export function Label({
     className,
-    style,
     children,
+    id,
```
benma, line 24: why not drop this and use `for=...`? would be more intuitive to use the label props for everything.

benma (changes requested):

frontends/web/src/components/forms/field.tsx
```diff
@@ -14,12 +15,12 @@
...
 
-import { h } from 'preact';
+import { h, JSX } from 'preact';
 import * as style from './field.css';
 
-export default function Field({
+export function Field({
     children, ...props
-}) {
+}: JSX.IntrinsicElements['div']) {
```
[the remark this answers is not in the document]
thisconnect: I found it by accident, in our preact version it only uses HTMLAttributes and SVGAttributes, but once we use a newer version we should get more detailed types
- with preact@8.5.3 https://github.com/preactjs/preact/blob/2b967c5074e9d067490c0ab8984cdbb527a46b78/src/preact.d.ts#L767
- with current preact@10.5.14 https://github.com/preactjs/preact/blob/ba8353cf3a105e45958168525c496902ab8524b6/src/jsx.d.ts#L793
- in current types/react https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a9a8595bfc5e2a1e7b0e595a57267fda6ab9ed69/types/react/index.d.ts#L3066

frontends/web/src/components/forms/label.tsx
thisconnect: Agreed, but unfortunately `for` is a reserved word in JavaScript and Typescript goes nuts, therefore JSX uses `htmlFor`, see react https://reactjs.org/docs/dom-elements.html#htmlfor

In preact 8 there are only typings for `for`, but `htmlFor` typings are missing: `Property 'htmlFor' does not exist on type 'HTMLAttributes'.ts(2322)` and got introduced in preact 10.x, see https://github.com/preactjs/preact/commit/5046b75a6aa76c2cc1312ab81679b07827f15812#diff-814abe9ce1d0e89a7fa1ef941142487296d1423ffad605dc1188b544056c2bcfR540

So either keep it with a TODO comment or temporary reassign `for` to `htmlFor` with a comment to change it to `htmlFor`?

```diff
diff --git a/frontends/web/src/components/forms/label.tsx b/frontends/web/src/components/forms/label.tsx
index 2190577a..345e786c 100644
--- a/frontends/web/src/components/forms/label.tsx
+++ b/frontends/web/src/components/forms/label.tsx
@@ -21,12 +21,12 @@ import * as style from './label.css';
 export function Label({
     className,
     children,
-    id,
+    for: htmlFor, // TODO: change to htmlFor when mirgated away from preact@8.x
     ...props
 }: JSX.IntrinsicElements['label']) {
     const classes = [style.label, className].join(' ');
     return (
-        <label for={id} className={classes} {...props}>
+        <label for={htmlFor} className={classes} {...props}>
             {children}
         </label>
     );
```

benma: k didn't realize it was so complicated ;) i leave it to you what is best, anything is good for me.

benma (approved):

PR #1381, opened 2021-07-09
frontend: revamp secondary button style

thisconnect:
Before the secondary buttons were orange, this was used a long
time ago, but apparently still in use for example on the connect
your own node section as back button.

We never want orange buttons. Changed to white button with light
outline border.


benma (approved):
utACK

PR #1382, opened 2021-07-09
Frontend select typescript

thisconnect:
needs frontend-aopp-ui first


frontends/web/src/components/forms/select.tsx
```diff
@@ -14,26 +15,40 @@
...
-    selected = null,
+    selectedOption,
     ...props
-}) {
+}: TSelectProps) {
     return (
         <div className={select}>
-            {label && <label for={id}>{label}</label>}
-            <select id={id} {...props}>
+            {label && <label for={props.id}>{label}</label>}
```
benma, line 44: why do `...props` and then `props.id` instead of putting `id` next to the other props above?

benma (approved):
utACK

frontends/web/src/components/forms/select.tsx
thisconnect: could yes, but then it also needs to be added to the next line, i.e `<select {...props}>` would change to `<select id={id} {...props}>`

benma: up to you!

thisconnect: fixed in https://github.com/digitalbitbox/bitbox-wallet-app/compare/3f5ec258e7b3a9aa0cd38e325bf4abe03b2a9f38..360ce5c0aec87ee2b03c40428be86d79a39413a9 also made separate commit and moved select.jsx to tsx first to preserve the history.

PR #1388, opened 2021-07-18
android: prevent screen from turning off while in foreground

x1ddos:
This is a long requested feature to prevent the screen from turning off
which otherwise may happen often, for exapmle while inputting a
passphrase or confirming a transaction on the device.

The change uses no new app permissions such as WAKE_LOCK. It simply
tells the OS to keep the screen on as long as the app is visible in
the foreground. As soon as the app is gone off the screen, any
configured screen timeouts at the OS level resume normal operations.

The screen still locks normally using an regular action like pushing a
physical lock button on the phone.

See docs for more details:
https://developer.android.com/training/scheduling/wakelock#screen


benma (approved):
utACK, cool!

PR #1389, opened 2021-07-19
frontend/aopp: remove cancel button during signing

benma:
It does nothing. The user has to cancel on the device.


thisconnect (approved):
LGTM

PR #1390, opened 2021-07-19
Aopp display address

benma:


frontends/web/src/api/account.ts
```diff
@@ -239,3 +239,7 @@ export interface IFeeTargetList {
 export const getFeeTargetList = (code: AccountCode): Promise<IFeeTargetList> => {
     return apiGet(`account/${code}/fee-targets`);
 };
+
+export const verifyAddress = (code: AccountCode, addressID: string): Promise<null> => {
```
thisconnect, line 243: Doesn't it return `Promise<boolean>` ?

benma: True, fixed. I didn't even check because the result is unused.

thisconnect (approved):
LGTM

PR #1394, opened 2021-07-19
Frontend aopp styling update

thisconnect:


thisconnect:
ok if we squash everything?

frontends/web/src/locales/en/app.json
```diff
@@ -72,7 +72,7 @@
     "signing": "To proceed, confirm message on your BitBox02.",
     "success": {
       "message": "Proceed on {{host}}",
-      "title": "Successfully sent address to external service."
+      "title": "Message successfully confirmed and sent"
```
benma: :-1: again i don't think the "message" is important at all. why not `Address successfully sent`?

frontends/web/src/locales/en/app.json
```diff
@@ -72,7 +72,7 @@
     "signing": "To proceed, confirm message on your BitBox02.",
     "success": {
       "message": "Proceed on {{host}}",
-      "title": "Message successfully confirmed and sent"
+      "title": "Address request confirmed and sent"
```
benma, line 75: Better but I would still just do `Address successfully sent` (it is not the address request that is sent).

benma (approved):
nice! but please change the message :)

benma (approved):
not sure why i have multiple reviews going on in one PR - maybe cause I went commit by commit...

frontends/web/src/locales/en/app.json
thisconnect: 👍    changed in https://github.com/digitalbitbox/bitbox-wallet-app/compare/488fafe2a809ff7687c60473e0f8f086105418f9..d958cc53580c66825c8208c4fbb95530c7d53df1

PR #1405, opened 2021-07-27
frontend/aopp: show vasp logo if available and rewording

thisconnect:


frontends/web/src/locales/en/app.json
```diff
@@ -64,12 +64,13 @@
...
   "aopp": {
     "addressRequest": "{{host}} is requesting a receiving address.",
+    "addressRequestWithLogo": "is requesting a receiving address",
     "banner": "Address request in progress.",
     "errorTitle": "Error during address request ",
     "labelAddress": "Address",
     "labelMessage": "Message",
     "reverifyInfoText": "Use to verify agianst external service",
-    "signing": "To proceed, sign message on your BitBox02.",
+    "signing": "To proceed, sign message on your BitBox02",
```
thisconnect, line 73: Already updated in Locize

frontends/web/src/locales/en/app.json
```diff
@@ -64,12 +64,13 @@
   },
   "aopp": {
     "addressRequest": "{{host}} is requesting a receiving address.",
+    "addressRequestWithLogo": "is requesting a receiving address",
```
benma, line 67: looks like this and the above could be unified into one entry

frontends/web/src/locales/en/app.json
benma, line 73: please change existing text on locize only. it will land in the repo with the next locize-pull (i do it regularly before releases and on-demand ;)). it is confusing to always do it in both places imho.

frontends/web/src/components/aopp/vasp.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface VASPProps {
+    fallback?: JSX.Element;
+    hostname: string;
+    prominent?: boolean;
+    withLogoText?: string;
+}
+
+const VASPLogoMap = {
+    'testing.21analytics.ch': TwentyOneAnalyticsLogo,
```
benma, line 33: i only see a login in this site - is this the right domain?

i know of:

https://testing.aopp.group/ and https://demo.aopp.group/

frontends/web/src/components/aopp/vasp.tsx
```diff
@@ -0,0 +1,64 @@
...
+import PocketBitcoinLogo from '../../assets/exchanges/logos/pocketbitcoin.svg';
+
+interface VASPProps {
+    fallback?: JSX.Element;
+    hostname: string;
+    prominent?: boolean;
+    withLogoText?: string;
+}
+
+const VASPLogoMap = {
```
benma, line 32: there are entries with subdomains and entries without subdomains.

it sounds more robust to discard subdomains and only go for the main part?

e.g. bitcoinsuisse.com redirects to https://www.bitcoinsuisse.com/ (including www)

frontends/web/src/components/aopp/vasp.tsx
```diff
@@ -0,0 +1,64 @@
...
+import TwentyOneAnalyticsLogo from '../../assets/exchanges/logos/21analytics.png';
+import BitcoinSuisseLogo from '../../assets/exchanges/logos/bitcoin_suisse.png';
+import BittrLogo from '../../assets/exchanges/logos/bittr.png';
+import BityLogo from '../../assets/exchanges/logos/bity.png';
+import PocketBitcoinLogo from '../../assets/exchanges/logos/pocketbitcoin.svg';
+
+interface VASPProps {
+    fallback?: JSX.Element;
+    hostname: string;
+    prominent?: boolean;
```
benma, line 28: remove this? i only see one use of this component and it is always prominent.

frontends/web/src/components/aopp/vasp.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { h, JSX, RenderableProps } from 'preact';
+import * as styles from './vasp.css';
+import TwentyOneAnalyticsLogo from '../../assets/exchanges/logos/21analytics.png';
+import BitcoinSuisseLogo from '../../assets/exchanges/logos/bitcoin_suisse.png';
+import BittrLogo from '../../assets/exchanges/logos/bittr.png';
+import BityLogo from '../../assets/exchanges/logos/bity.png';
+import PocketBitcoinLogo from '../../assets/exchanges/logos/pocketbitcoin.svg';
+
+interface VASPProps {
+    fallback?: JSX.Element;
```
benma, line 26: i'd make this mandatory - you have a fallback to the fallback, which is going a bit too far imho :joy:

benma (changes requested):

frontends/web/src/locales/en/app.json
benma: on 2nd thought, maybe it's better to formulate this one without assuming that the logo is in front. it is possible it can't be translated this way in some languages due to different grammar.

above, there is {{host}}, so in different languages it can be in a different position, but not here.

benma: Maybe: `There is an address request by:` in front of the logo?

then the same text can be used with the domain too - one entry less to translate, and it works for both cases.

frontends/web/src/locales/en/app.json
thisconnect: I know but this way we have nicer commits and I don't forget.

frontends/web/src/components/aopp/vasp.tsx
thisconnect: Since today it's on and off a login, lets support testing.aopp.group and demo.aopp.group too, but should we usethe 21 logo or the AOPP logo?

frontends/web/src/locales/en/app.json
thisconnect: I'll defer to @jadzeidan and @Stadicus

frontends/web/src/components/aopp/vasp.tsx
thisconnect: > there are entries with subdomains and entries without subdomains.

this is the reason this map exsits, so we could map `api.host.tld` to `host.tld` or remove the `testing.` subdomain.

thisconnect: > it sounds more robust to discard subdomains and only go for the main part?

Do you mean for all domains or just the ones we know?

Noob me doesn't know how to remove the subdomain i.e. the TLD could be .com.au or any other weird stuff.

frontends/web/src/components/aopp/vasp.tsx
thisconnect: Please look again, only on the first view (user-confirmation) it is prominent, see here https://github.com/digitalbitbox/bitbox-wallet-app/pull/1405/files#diff-c4c5f0fa92cdc4f4a6fbdd048df4be6bea47bea9db5204b4656e7ac59ba96869R122

but all other views it is used without `prominent`

frontends/web/src/components/aopp/vasp.tsx
thisconnect: only used once, all other occurences do not use it.

frontends/web/src/locales/en/app.json
thisconnect: Just a reminder, I have forgotten text changes in the past i.e. "Buy {name}" instead of "Buy {coinName}" or similar.

Should I undo it, here now in this PR so you can pull later?

benma: ack, no need to undo

frontends/web/src/components/aopp/vasp.tsx
benma: ok let's leave it then

frontends/web/src/components/aopp/vasp.tsx
benma: my point is that maybe they should - having a fallback that has another fallback sounds weird and confusing. not a huge deal, just a nit.

benma (approved):
utACK

frontends/web/src/components/aopp/vasp.tsx
thisconnect: Yep, I don't like it to much too. Small explanation for the others reading this:

Usually we just display "domainname.tld" in light gray if the domain is not known.
But on the first AOPP view (user-approval) it should show a sentence:

if domain is known:

- logo
- domain
- generic sentence

if the domain is NOT known

- sentence including the domainname

Unfortunately we can't use the same translation for the generic sentence and the that contains the domain.

frontends/web/src/components/aopp/vasp.tsx
```diff
@@ -0,0 +1,66 @@
...
+    'bitcoinsuisse.com': BitcoinSuisseLogo,
+    'bity.com': BityLogo,
+    'getbittr.com': BittrLogo,
+    'pocketbitcoin.com': PocketBitcoinLogo,
+};
+
+const VASPHostnameMap = {
+    'demo.aopp.group': 'AOPP.group',
+    'testing.aopp.group': 'AOPP.group',
+};
```
thisconnect, line 44: For the AOPP demo, the visible domain name would be either `Aopp.group` or if we keep the subdomain: `Demo.aopp.group` or `Testing.aopp.group`.
Therefore I mapped both to `AOPP.group`.

benma: imho not needed? could just leave it as it was before, no need for this map. or am i missing sth?

thisconnect: True but I'd assume we need name mapping anyways when some use a different subdomain i.e. `api.exchange.com` and now it is already prepared..

also this is a pure cosmetic change so our users don't see `Demo.aopp.group` or `Testing.aopp.group` nor `Aopp.group`

PR #1406, opened 2021-07-27
backend/aopp: skip choosingAccount step if there is only one account

benma:
No need for the user to select the account in a dropdown with only one choice.


x1ddos (approved):
LGTM

thisconnect (approved):
nice tested with an without multiple Bitcoin accounts LGTM

PR #1408, opened 2021-07-28
qt: let it use default rendering backend

x1ddos:
Commit a82a85020f164a705495bf21e689d93c7476e4cc forced Qt render in software mode, referring to some attack surface reduction. I could not find any existing vulnerabilities or advisory notes as to why and how exactly this helps in practical terms here.

OTOH, letting Qt use hardware acceleration, where available, improves user experience on Windows by far.

Until there's a reasonable and practical explanation why exactly software rendering yields more pro than cons in this context, I propose to switch back to Qt default backend rendering.

See Qt docs about backend renderers:
https://doc.qt.io/qt-5/qtquick-visualcanvas-adaptations.html

/cc @jadzeidan @thisconnect

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -162,10 +162,6 @@ int main(int argc, char *argv[])
     qputenv("QV4_FORCE_INTERPRETER", "1");
     qputenv("DRAW_USE_LLVM", "0");
 #endif
-#if QT_VERSION >= QT_VERSION_CHECK(5,11,0)
-    qputenv("QMLSCENE_DEVICE", "softwarecontext");
-    qputenv("QT_QUICK_BACKEND", "software");
-#endif
 
 
     BitBoxApp a(argc, argv);


thisconnect:
I can't see a visual difference on macOS, but tested by spamming the dialog "change device name"
CPU uses less 25% (was 45% when I do that with 4.28.2)
GPU goes up to 1% :)

jadzeidan:
@x1ddos I tested restarting the app again now and unfortunately the artifacts are back.

x1ddos:
Shame, although looks like 2 chess boards to me. Not too bad.

@Stadicus @benma can you guys try on your Windows this build with GPU enabled: https://ci.appveyor.com/project/benma/bitbox-wallet-app/builds/40162187/artifacts. I'm wondering whether it's @jadzeidan's graphics card.

Can also try enabling `QSG_RHI` instead, a new feature since Qt 5.14. Described on the same docs page.

Stadicus:
Tried the build @x1ddos referenced on my super-old Windows 10 laptop. Runs very smoothly, and I have no artefacts. But when I click on "Manage device" (and only there, anywhere in the app afaict), the dark sidebar quickly blinks white.

benma:
Nice, this runs much smoother on my linux too!

I can also see the white blink when going to Manage Device, but it is not new and exists already in the released version.

Will try Windows later.

benma:
> Can also try enabling `QSG_RHI` instead, a new feature since Qt 5.14. Described on the same docs page.

Attempt can be found here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1412

benma:
> > Can also try enabling `QSG_RHI` instead, a new feature since Qt 5.14. Described on the same docs page.
> 
> Attempt can be found here: #1412

@jadzeidan reports it simply renders a black screen, so that didn't work.

x1ddos:
Well, maybe it's only @jadzeidan's laptop and it works everywhere else.

benma (approved):
I am merging this for the next RC.

If we learn that the artefact happens for other users as well, we will spend time fixing it or reverting this change.

PR #1409, opened 2021-07-29
backend/coins/eth: add fee targets based on EthGasStation

benma:
https://ethgasstation.info/ provides gas price estimates for different
priorities.

We use these to give the user fee options.

If the service is not available, we fall back to the previous gas
price estimation provided by the eth_gasPrice RPC call to the ETH
node (currently etherscan proxy).

FeeTargets() is currently doing all the work whenever it is called,
which should be once when the send screen is loaded and every time a
tx proposal is made (every time the user changes something in the send
screen). This could be optimized with a cache that is invalidated
regularly (or when a new block arrives).

Custom gas price option to be added later.

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -48,7 +48,7 @@ var pollInterval = 60 * time.Second
 
 // ethGasStationAPIKey is used to access the API of https://ethgasstation.info.
 // See https://docs.ethgasstation.info/.
-const ethGasStationAPIKey = "289ea7884cb5903dcb50af9a81cd9518ae000ff9088af5182b5bc8b87385"
+const ethGasStationAPIKey = "3a61bee56f554cc14db4f49e2ace053b3086d3c323aefec83e5ff25ca485"
 
 // Account is an Ethereum account, with one address.
 type Account struct {
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -17,8 +17,10 @@ package eth
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math/big"
+	"net/http"
 	"os"
 	"path"
 	"strings"
@@ -44,6 +46,10 @@ import (
 
 var pollInterval = 60 * time.Second
 
+// ethGasStationAPIKey is used to access the API of https://ethgasstation.info.
+// See https://docs.ethgasstation.info/.
+const ethGasStationAPIKey = "289ea7884cb5903dcb50af9a81cd9518ae000ff9088af5182b5bc8b87385"
+
 // Account is an Ethereum account, with one address.
 type Account struct {
 	*accounts.BaseAccount
@@ -55,6 +61,7 @@ type Account struct {
 	db                   db.Interface
 	signingConfiguration *signing.Configuration
 	notifier             accounts.Notifier
+	httpClient           *http.Client
 
 	// true when initialized (Initialize() was called).
 	initialized bool
@@ -83,6 +90,7 @@ type Account struct {
 func NewAccount(
 	config *accounts.AccountConfig,
 	accountCoin *Coin,
+	httpClient *http.Client,
 	log *logrus.Entry,
 ) *Account {
 	log = log.WithField("group", "eth").
@@ -94,6 +102,7 @@ func NewAccount(
 		coin:                 accountCoin,
 		dbSubfolder:          "", // set in Initialize()
 		signingConfiguration: nil,
+		httpClient:           httpClient,
 		balance:              coin.NewAmountFromInt64(0),
 
 		enqueueUpdateCh: make(chan struct{}),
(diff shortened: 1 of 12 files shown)


backend/coins/eth/feetarget.go
```diff
@@ -0,0 +1,47 @@
...
+
+// FormattedFeeRate returns a string showing the fee rate.
+func (f *feeTarget) FormattedFeeRate() string {
+	if f.gasPrice == nil {
+		return ""
+	}
+	factor := big.NewInt(1e9)
+	return strings.TrimRight(strings.TrimRight(
+		new(big.Rat).SetFrac(f.gasPrice, factor).FloatString(9),
+		"0"), ".") + " Gwei"
```
x1ddos, line 46: spent at least few minutes trying to match parenthesis. is the idea to remove trailing zeros? this should be equivalent but easier to read imho:

```go
s := new(big.Rat).SetFrac(f.gasPrice, factor).FloatString(9)
return strings.TrimRight(s, ".0") + " Gwei"
```

frontends/web/src/locales/en/app.json
```diff
@@ -1134,12 +1135,16 @@
     "feeTarget": {
       "description": {
         "economy": "4 hours (24 blocks)",
+        "economy_eth": "< 30 minutes",
```
x1ddos, line 1138: might be misleading. I would spell it out: "30 minutes or less"

backend/coins/eth/account.go
```diff
@@ -427,34 +434,30 @@ type TxProposal struct {
 	Keypath signing.AbsoluteKeypath
 }
 
-func (account *Account) newTx(
-	recipientAddress string,
-	amount coin.SendAmount,
-	data []byte,
-) (*TxProposal, error) {
-	if !ethcommon.IsHexAddress(recipientAddress) {
+func (account *Account) newTx(args *accounts.TxProposalArgs) (*TxProposal, error) {
```
x1ddos, line 437: TxProposalArgs doc comment still says:

> // Only applies to BTC/LTC.

x1ddos (approved):
LGTM. Yeah, looks pretty much straightforward.

backend/coins/eth/account.go
x1ddos: ehm, I meant the `FeeTargetCode` field.

backend/coins/eth/feetarget.go
benma: Your code turns `10` into `1`.

We use this type of formatting in a few places in the app and also in the firmware: https://github.com/digitalbitbox/bitbox02-firmware/blob/master/src/rust/util/src/decimal.rs#L32

I'll try to add unit tests.

benma: Ah it doesn't cause FloatString adds decimal places. Hmm

thisconnect:
Tested on testnet and got different fee options, looks good  👍

backend/coins/eth/feetarget.go
x1ddos: > Your code turns 10 into 1.

sounds like even more reason to comment the code and make it more readable.

benma: I added unit tests, your code doesn't work for e.g. `21e10` which becomes `210 Gwei` - source code strips beyond the decimal place to `21 Gwei`.

frontends/web/src/locales/en/app.json
benma: done

backend/coins/eth/account.go
benma: removed

benma:
> Tested on testnet and got different fee options, looks good +1

Hmm I forgot about testnet and that is unexpected - I call the same endpoints. I'll make sure it only happens on mainnet. Good catch!

backend/coins/eth/account.go
```diff
@@ -44,6 +46,8 @@ import (
 
 var pollInterval = 60 * time.Second
 
+const ethGasStationAPIKey = "289ea7884cb5903dcb50af9a81cd9518ae000ff9088af5182b5bc8b87385"
```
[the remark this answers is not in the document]
benma: commented

backend/coins/eth/feetarget.go
x1ddos: those 3 lines are still unreadable.

benma: improved

x1ddos: much nicer. thanks! :+1:

benma:
> > Tested on testnet and got different fee options, looks good +1
> 
> Hmm I forgot about testnet and that is unexpected - I call the same endpoints. I'll make sure it only happens on mainnet. Good catch!

On 2nd thought, I'll just leave it. Less complicated, and it doesn't matter what the fees are on testnet really.

thisconnect:
> On 2nd thought, I'll just leave it. Less complicated, and it doesn't matter what the fees are on testnet really.

👍  I prefer it too, we could also consider using btc mainnet fees on the testnet but just a thought... unless there is a good reason to not do it.

x1ddos:
Forgot to mention: update changelog.

benma:
> Forgot to mention: update changelog.

Great catch. Added in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1410/

PR #1410, opened 2021-07-29
ETH custom fees

benma:


thisconnect (commented):
tested TETH with and without custom fee  ✅ 

Also TBTC, but got once an interesting error, but I suspect it was due to sending 2 tx too quickly

Screen Shot 2021-07-30 at 9 05 23 AM

backend/coins/eth/account.go
```diff
@@ -678,14 +681,30 @@ func (account *Account) FeeTargets() ([]accounts.FeeTarget, accounts.FeeTargetCo
...
+// price if the fee target is `FeeTargetCodeCustom`.
+func (account *Account) gasPrice(args *accounts.TxProposalArgs) (*big.Int, error) {
+	if args.FeeTargetCode == accounts.FeeTargetCodeCustom {
+		// Convert from Gwei to Wei.
+		amount, err := coin.NewAmountFromString(args.CustomFee, big.NewInt(1e9))
+		if err != nil {
+			return nil, err
+		}
+		gasPrice := amount.BigInt()
+		if gasPrice.Cmp(big.NewInt(0)) <= 0 {
```
x1ddos, line 694: Maybe a nice future addition would be to compare the value to the economy fee target and, if it's smaller, warn users the fee might be too low.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Add support for the Address Ownership Proof Protocol (AOPP), i.e.: handle 'aopp:?...' URIs. See https://aopp.group/.
+- Add fee options based on priority, and the ability to set a custom gas price.
```
x1ddos, line 5: I'd mention this is for etherium. the only hint is "gas" but some might not even know what gas is if they don't use ETH.

x1ddos (approved):
LGTM

CHANGELOG.md
benma: Of course - how could I miss it...

PR #1414, opened 2021-08-02
frontend/aopp: remove reverify info text

thisconnect:
Removed info text below the verify button on the AOPP success
screen, as we don't need it.


benma (approved):
utACK

PR #1416, opened 2021-08-02
Frontend cleanup devices

thisconnect:
- moved old steps component to device/bitbox02
- asynchronously load unlock gesture animation


thisconnect:
Without this change (note: the long yellow snail)

Screen Shot 2021-08-02 at 6 09 00 PM


With this change (2 shorter yellow snails)

The app loads much faster in webdev mode, but gesture gif will take a moment to show, this should have no visible impact on the production app.

Screen Shot 2021-08-02 at 6 04 14 PM

thisconnect:
Current master bundle.js size with `make webdev` is 30.8MB, with this PR it is 6.2MB, which is still big but the reason is it contains source maps.

frontends/web/src/routes/device/bitbox02/components/password-entry/password-entry.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { h } from 'preact';
+import * as styles from './password-entry.css';
+
+interface IPasswordEntryProps {
+    version: '1.0.0' | '2.0.0' | string;
```
benma, line 21: just `string` is okay, as it includes the other two.

benma (commented):
For me there was a noticeable delay until the gif was loaded, about 5s, during which the image was just white. Same for you? 

We could also think about removing the old gif to save some binary space - 1.0.0 and 2.0.0 were only used for the first beta devices - 3.0.0 was the first real release.

I also saw that the current gif is 9mb in size, which is almost 10% of the whole app. Is there any way to reduce the size of the gif while maintaining adequate quality?

benma:
About the delay, I missed

>  this should have no visible impact on the production app.

Will test it now.

benma (approved):
Indeed, no delay in the Qt app :+1:

thisconnect:
I'll remove the old gif and update this PR.

> I also saw that the current gif is 9mb in size, which is almost 10% of the whole app. Is there any way to reduce the size of the gif while maintaining adequate quality?

We could try as webm movie instead of gif (afaik Twitter compresses animated gif's as movie and reduces the size this way), but then I believe @jadzeidan was already thinking about alternatives to the animated gif, let's merge this first.

thisconnect:
The 5s lag for the video is what we gain now in the rest of the app in `webdev` mode. In other words everything else should load 5s faster in dev mode as the video is only loaded on the unlock (or setup step) and not anymore part of the main bundle.

Updated PR and removed the old gesture as it shouldn't appear anymore, also beta devices should have up to date firmware with the newer pwd input.

thisconnect:
Much faster with WEBM, no delay visible. I'll remove the dynamic loading although I would have liked to figure out how to properly do dynamic module loading with correct typings.

Current issue: autoplay does only work one time, but when unplugging and re-plugging the BB it won't autoplay again. Maybe this changes when it is not dynamically loaded. 

Screen Shot 2021-08-06 at 11 29 56 AM

thisconnect:
Removed dynamic import (slightly bigger bundle again but should not be noticeable in webdev) and fixed autoplay.

9d225e4c7f7d482ec0889cb78330d02f03f4abce

Screen Shot 2021-08-07 at 10 48 31 AM

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -147,6 +149,14 @@ class BitBox02 extends Component<Props, State> {
                     break;
             }
         });
+        import('./components/password-entry/password-entry')
+            .then(({ PasswordEntry }) => {
+                if (!this.PasswordEntry) {
```
benma: why this check? when can it be false?

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -105,10 +105,12 @@ class BitBox02 extends Component<Props, State> {
             agreement4: false,
             agreement5: false,
             waitDialog: undefined,
+            passwordEntryLoaded: false,
         };
     }
 
     private unsubscribe!: () => void;
+    private PasswordEntry: any | null = null;
```
benma: why not put this into the state? then you don't need two vars.

benma: also why `any  | null`? `any` includes `null`. And in TS `undefined` is preferred over `null` :joy:

benma (changes requested):
Why did you change it to load the component instead of the gifs asynchronously? imho the code before was clearer.

also convert the old gesture gif to .webm? would be another 1mb saved.

thisconnect:
My last commit changed already to static import https://github.com/digitalbitbox/bitbox-wallet-app/commit/9d225e4c7f7d482ec0889cb78330d02f03f4abce

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
thisconnect: That would have been IPasswordEntry interface + null if we would have gone with dynamically loading the module, see my comment

> I'll remove the dynamic loading although I would have liked to figure out how to properly do dynamic module loading with correct typings.

https://github.com/digitalbitbox/bitbox-wallet-app/pull/1416#issuecomment-894135231

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
thisconnect: initially before loading the module during runtime this would be null. But outdated now anyways as we change to statically importing the module anyways

frontends/web/src/routes/device/bitbox02/components/password-entry/password-entry.tsx
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import { h, RenderableProps} from 'preact';
+import PasswordGestureVideo from './assets/password-gestures.webm';
+import * as styles from './password-entry.css';
+
+export interface IPasswordEntryProps {}
+
+function isVideoPlaying(video: HTMLVideoElement): boolean {
+    return !!(video.currentTime > 0 && !video.paused && !video.ended && video.readyState > 2);
```
benma, line 24: why `!!`? should be `boolean` already by the looks of it

frontends/web/src/routes/device/bitbox02/components/password-entry/password-entry.tsx
```diff
@@ -0,0 +1,56 @@
...
+export function PasswordEntry({ children }: RenderableProps<IPasswordEntryProps>) {
+    return (
+        <div className={styles.passwordGesturesWrapper}>
+            <video
+                autoPlay
+                // autoPlay+muted only works the first time but after re-render,
+                // i.e. when re-plugin the BitBox the video doesn't play anymore
+                // https://github.com/preactjs/preact/issues/747#issuecomment-370905360
+                // looks like this can be removed with Preact10/React
+                ref={ref => replayVideo(ref)}
```
benma, line 45: `ref={ replayVideo }` might work too

benma (approved):
utACK, nice!

frontends/web/src/routes/device/bitbox02/components/password-entry/password-entry.tsx
thisconnect: interestingly it doesn't play when replugin `ref={replayVideo}` I think that is part of the preact 8 caching issue that should be gone in preact X / react

frontends/web/src/routes/device/bitbox02/components/password-entry/password-entry.tsx
thisconnect: `true`

(pun intended)

thisconnect: fixed in https://github.com/digitalbitbox/bitbox-wallet-app/compare/b4c7cc93b336bf409c01e9c0ac8f233133f4daf7..de055f00a4132747c17abc1964a5ee53b743ef8a

PR #1417, opened 2021-08-03
backend: fix regression - close account

benma:
In c6b816aa35bfdf855662df12e3b9cb43d31a1b96, the account.Close() was
accidentally dropped. When reinitializing the accounts (e.g. after an
account rename), the old ones were not closed, blocking the database
files, so the new ones could not properly load.


thisconnect (approved):
LGTM

tested with testnet webdev, master doesn't load accounts after rename, fixed with this PR

PR #1419, opened 2021-08-03
backend/aopp: provide account code to the frontend

benma:
So it can use it in the 'Verify address' button. Before, it used the
frontend accountCode state var set in the choosing-account step, but
that step is skipped if there is only one account.


thisconnect (approved):
LGTM

PR #1420, opened 2021-08-03
frontend: fix deactivated account links to empty screen

thisconnect:
When clicking a deactivated account (Bitcoin, Litecoin, Ethereum)
the content area turns gray.

Added a check to test if the account is active before routing.
Tokens already had this check.


benma (approved):
utACK

PR #1422, opened 2021-08-03
android: request USB permission if not automatically granted

benma:
There are these scenarios to cover:

- App is closed, user inserts device:

In this case, the user is prompted with all apps that register intent
filters for the device, including the BitBoxApp.

a) If the BitBoxApp is chosen, the app is launched with the
`ACTION_USB_DEVICE_ATTACHED` intent. Previously we would simply update
the device list in `onResume()` without checking the intent, which
included this case. Now we explicitly check the intent in
`onResume()` to avoid endlessly asking the user for permission (see below).

b) If the app is not chosen, permission is denied. If the user then
opens the app (without USB permission granted automatically like
above), previously the app would crash because of it. In this commit,
we request permission in `onCreate()` (by calling
`updateDevice()`). We do it newly in `onCreate()` instead of
`onResume()` because if permission is denied, `onResume()` is called
again, asking the user again for permission endlessly.

- App is opened, user inserts device:

App chooser is shown to the user like above. If they choose the same
BitBoxApp, `onResume()` is called with the intent, same as above.

If they reject, the app remains open without any further action.

- User opens app while the device is already connected

This only happens if the user rejects to open the BitBoxApp after
being prompted.

In this case, `onCreate()` calls `updateDevice()` which asks for USB permission.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -3,6 +3,7 @@ package ch.shiftcrypto.bitboxapp;
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.AlertDialog;
+import android.app.PendingIntent;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.DialogInterface;
@@ -43,6 +44,8 @@ import goserver.Goserver;
 
 public class MainActivity extends AppCompatActivity {
     private final int PERMISSIONS_REQUEST_CAMERA_QRCODE = 0;
+    private static final String ACTION_USB_PERMISSION = "ch.shiftcrypto.bitboxapp.USB_PERMISSION";
+
     // stores the request from onPermissionRequest until the user has granted or denied the permission.
     private PermissionRequest webViewpermissionRequest;
 
@@ -213,6 +216,10 @@ public class MainActivity extends AppCompatActivity {
         final String javascriptVariableName = "android";
         vw.addJavascriptInterface(new JavascriptBridge(this), javascriptVariableName);
         vw.loadUrl("file:///android_asset/web/index.html");
+
+        // We call updateDevice() here in case the app was started while the device was already connected.
+        // In that case, handleIntent() is not called with ACTION_USB_DEVICE_ATTACHED.
+        this.updateDevice();
     }
 
     private void log(String msg) {
@@ -247,12 +254,8 @@ public class MainActivity extends AppCompatActivity {
         // DETACHED intent is a broadcast intent which we register here.
         IntentFilter filter = new IntentFilter();
         filter.addAction(UsbManager.ACTION_USB_DEVICE_DETACHED);
-        filter.addAction("ch.shiftcrypto.bitboxapp.USB_PERMISSION");
+        filter.addAction(ACTION_USB_PERMISSION);
         registerReceiver(this.usbStateReceiver, filter);
-
-        // We call maybeSetDevice() here in case the app was started while the device was already connected.
-        // In that case, handleIntent() is not called with ACTION_USB_DEVICE_ATTACHED.
-        this.updateDevice();
 
         // Listen on changes in the network connection. We are interested in if the user is connected to a mobile data connection.
         registerReceiver(this.networkStateReceiver, new IntentFilter(android.net.ConnectivityManager.CONNECTIVITY_ACTION));
@@ -260,6 +263,7 @@ public class MainActivity extends AppCompatActivity {
         // Handle 'aopp:' URIs. This is called when the app is launched and also if it is already
         // running and brought to the foreground.
         Intent intent = getIntent();
+        handleIntent(intent);
         if(Intent.ACTION_VIEW.equals(intent.getAction())) {
             Uri uri = intent.getData();
             if (uri != null) {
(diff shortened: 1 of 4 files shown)


benma:
I used this as a guide: https://developer.android.com/guide/topics/connectivity/usb/host

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -245,12 +247,12 @@ protected void onResume() {
         // DETACHED intent is a broadcast intent which we register here.
         IntentFilter filter = new IntentFilter();
         filter.addAction(UsbManager.ACTION_USB_DEVICE_DETACHED);
+        filter.addAction("ch.shiftcrypto.bitboxapp.USB_PERMISSION");
         registerReceiver(this.usbStateReceiver, filter);
 
-        // We call updateDeviceList() here in case the app was started while the device was already connected.
+        // We call maybeSetDevice() here in case the app was started while the device was already connected.
```
x1ddos: updateDevice

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -245,12 +247,12 @@ protected void onResume() {
         // DETACHED intent is a broadcast intent which we register here.
         IntentFilter filter = new IntentFilter();
         filter.addAction(UsbManager.ACTION_USB_DEVICE_DETACHED);
+        filter.addAction("ch.shiftcrypto.bitboxapp.USB_PERMISSION");
```
x1ddos: Not sure why this commit adds the action here. It doesn't seem to be handled in `usbStateReceiver` as per docs:

```java
private static final String ACTION_USB_PERMISSION =
    "com.android.example.USB_PERMISSION";

    public void onReceive(Context context, Intent intent) {
        String action = intent.getAction();
        if (ACTION_USB_PERMISSION.equals(action)) {
          ...
        }
   }
```

I assume it'll be used in a following commit but seems misplaced here.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -279,6 +283,21 @@ protected void onPause() {
     }
 
     private void handleIntent(Intent intent) {
+        if (intent.getAction().equals(ACTION_USB_PERMISSION)) {
+            // See https://developer.android.com/guide/topics/connectivity/usb/host#permission-d
+            synchronized (this) {
+                UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
+                if (intent.getBooleanExtra(UsbManager.EXTRA_PERMISSION_GRANTED, false)) {
```
x1ddos, line 275: Can't understand this. The docs say:

> When users reply to the dialog, your broadcast receiver receives the intent that contains the EXTRA_PERMISSION_GRANTED extra, which is a boolean representing the answer. Check this extra for a value of **true** before connecting to the device.

Is there a mistake in their example?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -306,6 +296,16 @@ private void handleIntent(Intent intent) {
             log("usb: detached");
             this.updateDevice();
         }
+        // Handle 'aopp:' URIs. This is called when the app is launched and also if it is already
+        // running and brought to the foreground.
+        if(Intent.ACTION_VIEW.equals(intent.getAction())) {
```
x1ddos, line 296: strange to read. all the other branches have it reversed: `intent.getAction().equals(Intent.ACTION_VIEW)`

x1ddos: also, might as well turn these `if`s into a single switch statement?

```java
switch (intent.getAction()) {
  case ACTION_USB_PERMISSION:
    ...
    break
  case UsbManager.ACTION_USB_DEVICE_ATTACHED:
  case UsbManager.ACTION_USB_DEVICE_DETACHED:
  case Intent.ACTION_VIEW:
}

File: frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/Util.java [hunk 5]
```diff
@@ -13,4 +14,8 @@ public static void systemOpen(Application application, String url) throws Except
             application.startActivity(intent);
         }
     }
+
+    public static void log(String msg) {
+        Log.d("ch.shiftcrypto.bitboxapp", "XYZ: " + msg);
```
x1ddos, line 19: since you're here, remove the funny xyz? not sure what it's for.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
benma: It is handled in `handleIntent()` (in the same commit) which is called from the usb state receiver.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
benma:
```
public boolean getBooleanExtra (String name, 
                boolean defaultValue)
```

The 2nd argument is the default value, not the value to check against, so it is all correct.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
benma: I'd be worried if `==` (switch) is the same as `.equals()`. I'll just reverse the order to make it consistent as you say.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/Util.java
benma: Done

benma:
addressed feedback, PTAL

x1ddos (approved):
LGTM 

> I'd be worried if == (switch) is the same as .equals(). I'll just reverse the order to make it consistent as you say.

String in switch is available since Java 7: http://openjdk.java.net/projects/jdk7/features/. But sure, I don't mind either.

PR #1423, opened 2021-08-04
frontend/qt: bundle more DLLs in Windows installer

benma:
A user reported that installation failed with errors that
msvcp140_1.dll and vcruntime140_1.dll was misisng.

We bundle all DLLs that exist with the msvcp140 and vcruntime140
prefix to solve this problem.

### DIFF
--- a/frontends/qt/compile_windows.bat
+++ b/frontends/qt/compile_windows.bat
@@ -5,5 +5,10 @@ cd build
 qmake ..\BitBox.pro
 nmake
 COPY "%VCToolsRedistDir%\x64\Microsoft.VC142.CRT\msvcp140.dll" windows\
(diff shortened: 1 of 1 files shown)


benma:
VS 2017 directory, used up to 4.28.2:

image

VS 2019 directory, used from 4.29.0:

image


I assume if the user has the VC Redist package (https://support.microsoft.com/en-us/topic/the-latest-supported-visual-c-downloads-2647da03-1eea-4433-9aff-95f26a218cc0), it works without bundling, and that users face this issue because of the update to 2019, for which they might not have the Redist package installed.

benma:
As an urgent measure, I updated the 4.29.0 Windows release based on this commit and a user reports that it resolved their issue. I will update the release notes soon to mention this.

x1ddos (approved):
LGTM

PR #1431, opened 2021-09-03
config: change default fiat currency to usd

thisconnect:
Currently the fiat currency default is Swiss Franc (CHF), but many
people don't know CHF or this default is not very useful for most.

Changed default to US Dollar (USD) as it is the most widely used
pair with Bitcoin and ohter crypto currencies. Kept Euro and CHF
as enabled secondary fiat currencies.

### DIFF
--- a/frontends/web/src/components/rates/rates.tsx
+++ b/frontends/web/src/components/rates/rates.tsx
@@ -42,7 +42,7 @@ export const currencies: Fiat[] = ['AUD', 'BRL', 'CAD', 'CHF', 'CNY', 'EUR', 'GB
 
 export const store = new Store<SharedProps>({
     rates: undefined,
-    active: 'CHF',
+    active: 'USD',
     selected: ['USD', 'EUR', 'CHF'],
 });
 
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -215,7 +215,7 @@ func NewDefaultAppConfig() AppConfig {
 			},
 			// Copied from frontend/web/src/components/rates/rates.tsx.
 			FiatList: []string{"USD", "EUR", "CHF"},
-			MainFiat: "CHF",
+			MainFiat: "USD",
 		},
 	}
 }


thisconnect:
update changelog

benma (approved):
utack

frontends/web/src/components/rates/rates.tsx
```diff
@@ -42,7 +42,7 @@ export const currencies: Fiat[] = ['AUD', 'BRL', 'CAD', 'CHF', 'CNY', 'EUR', 'GB
 
 export const store = new Store<SharedProps>({
     rates: undefined,
-    active: 'CHF',
+    active: 'USD',
```
x1ddos, line 45: I don't remember now: why is the default fiat set in both backend and frontend. shouldn't backend setting be enough?

x1ddos (commented):
how about a little map of country > fiat and set the fiat based on the country? we support only 15 currencies. shouldn't be that big.

frontends/web/src/components/rates/rates.tsx
thisconnect: Yes the setting is coming from the backend, this is only the default initialization of the custom frontend Store.ts.

benma:
CI is failing due to an unrelated reason - a race condition in a unit test recently added.

Fix: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1435/

thisconnect:
> how about a little map of country > fiat and set the fiat based on the country? we support only 15 currencies. shouldn't be that big.

Yes we need to revamp the fiat currency selection we will have more issue when adding new currencies, I'll add an issue.

benma:
> how about a little map of country

But what is the country?

thisconnect:
not part of this PR, we can discuss this offline. This prob is part of a larger revamp.

> But what is the country? Also country does not necessarily determine which currencies the user is interested in.

x1ddos:
Country is whatever [NativeLocale](https://github.com/digitalbitbox/bitbox-wallet-app/blob/bdb9c975/backend/handlers/handlers_test.go#L56) reports unless set explicitely. Of course it doesn't map 1:1 when say English is selected but still better than a hardcoded USD.

> But what is the country? Also country does not necessarily determine which currencies the user is interested in.

PR #1432, opened 2021-09-06
frontend: remove animations

benma:
Since updating to Qt 5.15, animations are very laggy on
Windows (opening dialogs).

It could be fixed by enabling GPU accelerated rendering, but that led
do various rendering artefacts and even crashes.

This is a compromise: by removing animations, the app is not laggy on
Windows even when only using software rendering.


benma:
See also: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1433

thisconnect (approved):
LGTM but we still have some animations and should try to keep them.

This commit doesn't remove all animations, but the ones that cover big parts of the screen and are unperformant.

Examples of animations that we sill have and should keep are: Spinner, header sync progress indicator, animated success icon

PR #1433, opened 2021-09-06
qt: go back to software rendering

benma:
We attempted to use GPU accelerated rendering in a9823e9a3.

Some users had rendering artefacts on Windows, and at least one user
could not launch the app on Ubuntu 18.04 with:

```
ERROR:command_buffer_proxy_lmpl.cc Context Result::KTranslentFailure:Failed to send GpuChannelMsg_
CreateCommandBuffer.
Segmentation fault (core dumped)
```

To fix these issues, we revert the change and re-enable software rendering.

GPU acceleration was enabled to deal with a very laggy animation when
opening dialogs on Windows. We will deal with this by disabling such
animations in another commit.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -163,6 +163,11 @@ int main(int argc, char *argv[])
     qputenv("DRAW_USE_LLVM", "0");
 #endif
 
+    // Force software rendering over GPU-accelerated rendering as various rendering artefact issues
+    // were observed on Windows and the app crashes on some Linux systems.
+    qputenv("QMLSCENE_DEVICE", "softwarecontext");
+    qputenv("QT_QUICK_BACKEND", "software");
+
 
     BitBoxApp a(argc, argv);
     // These three are part of the SingleApplication instance ID - if changed, the user should close
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Verify the EIP-55 checksum in mixed-case Ethereum recipient addresses
+- Disable GPU acceleration introduced in v4.29.0 due to rendering artefacts on Windows
 
 ## 4.29.0 [released 2021-08-03]
 - Add support for the Address Ownership Proof Protocol (AOPP), i.e.: handle 'aopp:?...' URIs. See https://aopp.group/.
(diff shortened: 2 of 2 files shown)


frontends/qt/main.cpp
```diff
@@ -163,6 +163,11 @@ int main(int argc, char *argv[])
     qputenv("DRAW_USE_LLVM", "0");
 #endif
 
+    // Force software rendering over GPU-accelerated rendering as various rendering artefact issues
+    // were observed on Windows and the app crashes on some Linux systems.
```
x1ddos, line 167: So, none of the other values worked, not even OpenVG?

x1ddos: Or RHI:

> From Qt 5.14 onwards, the default adaptation gains the option of rendering via a graphics abstraction layer, the Qt Rendering Hardware Interface (RHI), provided by the QtGui module.

(from https://doc.qt.io/qt-5/qtquick-visualcanvas-scenegraph-renderer.html#rendering-via-the-qt-rendering-hardware-interface)

benma: RHI didn't do anything unfortunately.

OpenVG I missed - do you want to try to quickly enable it in a branch based off of v4.29.0? 

In the PR you can remove the `- if defined APPVEYOR_PULL_REQUEST_NUMBER appveyor exit` line from appveyor to get a Windows build artefact, which we could test.

x1ddos: So, tried `openvg` - doesn't work due to a "Could not create scene graph context for backend 'openvg' - check that plugins are installed correctly in c:/program files/bitbox/plugins". We don't even have plugins and openvg seems to dying: https://www.phoronix.com/scan.php?page=news_item&px=Qt-5.9-OpenVG-Overview

RHI doesn't work due to https://bugreports.qt.io/browse/QTBUG-78682, I believe. QtWebEngine is based on chromium which requires [ANGLE](https://doc.qt.io/qt-5/windows-requirements.html) lib. Otherwise, RHI results in a black screen.

We're out of options.

frontends/qt/main.cpp
```diff
@@ -163,6 +163,11 @@ int main(int argc, char *argv[])
...
+    qputenv("QMLSCENE_DEVICE", "softwarecontext");
```
x1ddos, line 168: I don't think you need this, esp. since we're upgraded to qt5.15 but I guess it doesn't do any harm.

x1ddos (approved):
LGTM

frontends/qt/main.cpp
x1ddos: > We're out of options.

Well, I imagine a more involved future option is to build Qt from sources, linking with ANGLE lib.

PR #1434, opened 2021-09-06
frontend/guide: add entry that plugout is ok after send and receive

thisconnect:
There was some confusion that people are unsure if the BitBox has
to stay plugged in when receiving or sending.

Added guide entry that it is save to plugout the device after send
or creating a receive address.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -738,6 +738,10 @@
         "text": "For the BitBox01, click on the BitBox icon in the sidebar on the left and see the Pairing section. The guide will update and you can continue following the instructions from there.\nFor the BitBox02, you can verify addresses directly on the device during the send/receive process.",
         "title": "How can I verify an address securely?"
       },
+      "plugout": {
+        "text": "No, once you sent coins to your BitBox address, you do not need to leave your BitBox plugged in. You are free to disconnect your BitBox.",
+        "title": "Do I need to leave my BitBox plugged in while receiving?"
+      },
       "why20": {
         "text": "During start-up the app generates addresses derived from your seed to see if they have received funds. As the app can generate an almost infinite number of addresses, it could spend years determining the balance. To limit this search it stops after it sees 20 addresses that have never received funds. This is the \"gap limit\" and 20 is a de-facto standard though the number is arbitrary. These are the 20 addresses you can choose from.",
         "title": "Why only 20 addresses?"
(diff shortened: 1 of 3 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -738,6 +738,10 @@
         "text": "For the BitBox01, click on the BitBox icon in the sidebar on the left and see the Pairing section. The guide will update and you can continue following the instructions from there.\nFor the BitBox02, you can verify addresses directly on the device during the send/receive process.",
         "title": "How can I verify an address securely?"
       },
+      "plugout": {
+        "text": "No, once you send coins to your BitBox address, you do not need to leave your BitBox plugged in. This is because the transaction has already been broadcasted to the blockchain and is on its way to your BitBox address. You are free to disconnect your BitBox.",
```
benma, line 742: 'once you sent'?

broadcasted -> broadcast (past tense of broadcast is broadcast)

I'd actually drop the whole sentence 'This is because the transaction has already been broadcasted to the blockchain and is on its way to your BitBox address.' as it is quite inaccurate :shrug:

frontends/web/src/locales/en/app.json
```diff
@@ -756,6 +760,10 @@
         "text": "The fee is based on the transaction data size and not its amount. The fee targets are calculated by Bitcoin Core's fee estimation algorithm for each network priority you chose. They are shown if they have a different value from the target below.\nEconomy: 24 blocks (around 4 hours for Bitcoin, 1 hour for Litecoin)\nLow: 12 blocks (around 2 hours for Bitcoin, 30 minutes for Litecoin)\nNormal: 6 blocks (around 1 hour for Bitcoin, 15 minutes for Litecoin)\nHigh: 2 blocks (around 20 minutes for Bitcoin, 5 minutes for Litecoin)\n(A block takes on average ten minutes for Bitcoin (2.5 minutes in Litecoin) to mine and the network load may vary considerably in the above periods.)",
         "title": "How is the fee determined?"
       },
+      "plugout": {
+        "text": "No, once you have made a send transaction, you do not need to leave your BitBox plugged in. This is because the transaction has already been broadcasted to the blockchain and is on its way to the destination address. You are free to disconnect your BitBox.",
```
benma, line 764: same here, remove the middle sentence?

benma (changes requested):

frontends/web/src/locales/en/app.json
benma: Also 'Once you have made a transaction' instead of 'send transaction'? 'send transaction' sounds a bit confusing imho

frontends/web/src/locales/en/app.json
jadzeidan: I think it's nice to have a high level sentence explaining why they don't need to have it plugged in. I'm fine to remove for now as not to block the release. But I'll try to think way to explain it more accurately without being too technical.

benma: :+1: can always update it on locize later

frontends/web/src/locales/en/app.json
thisconnect: fixed in https://github.com/digitalbitbox/bitbox-wallet-app/compare/a9cc9deb8bca1a52da3b0843c9677ed8488529f2..a568f6a533009adc0997c43348819fc82f0fbb00

frontends/web/src/locales/en/app.json
thisconnect: In the future a not too technical high level sentence explaining why would be nice.

fixed in 
https://github.com/digitalbitbox/bitbox-wallet-app/compare/a9cc9deb8bca1a52da3b0843c9677ed8488529f2..a568f6a533009adc0997c43348819fc82f0fbb00

benma (approved):

PR #1440, opened 2021-09-07
frontend: use feerate or gas price for custom fee label

thisconnect:
Currently when selecting a custom fee, the label on top of the input
just says 'Network fee', but we use this label also for the total
amount of the estimated fee price below the selection.

Changed the custom fee label to display either 'Fee rate' or
'Gas price' depending on the coinCode of the selected account.

### DIFF
--- a/frontends/web/src/routes/account/send/feetargets.tsx
+++ b/frontends/web/src/routes/account/send/feetargets.tsx
@@ -20,7 +20,7 @@ import * as accountApi from '../../../api/account';
 import { Input, Select } from '../../../components/forms';
 import { load } from '../../../decorators/load';
 import { translate, TranslateProps } from '../../../decorators/translate';
-import { getCoinCode, customFeeUnit } from '../utils';
+import { customFeeUnit, getCoinCode, isEthereumBased } from '../utils';
 import * as style from './feetargets.css';
 
 interface LoadedProps {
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/account/send/feetargets.tsx
```diff
@@ -182,7 +182,9 @@ class FeeTargets extends Component<Props, State> {
                                     align="right"
                                     className={`${style.fee} ${style.feeCustom}`}
                                     disabled={disabled}
-                                    label={t('send.fee.label')}
+                                    label={t('send.feeTarget.customLabel', {
+                                        context: isBitcoinBased(coinCode) ? '' : 'eth'
```
benma, line 186: The opposite check makes more sense (isEthereumBased)? If we added another coin it would say 'gas price', but 'fee rate' should be the default generic text.

benma (changes requested):

frontends/web/src/routes/account/send/feetargets.tsx
thisconnect: Rebased and updated.

thisconnect:
PTAL

frontends/web/src/routes/account/send/feetargets.tsx
```diff
@@ -182,7 +182,9 @@ class FeeTargets extends Component<Props, State> {
                                     align="right"
                                     className={`${style.fee} ${style.feeCustom}`}
                                     disabled={disabled}
-                                    label={t('send.fee.label')}
```
benma, line 185: delete this key from en/app.json?

benma (changes requested):

frontends/web/src/routes/account/send/feetargets.tsx
thisconnect: it is also used in various other places i.e. if not custom fee is selected or in the confirming wait-dialog
https://github.com/digitalbitbox/bitbox-wallet-app/search?q=send.fee.label&type=

benma (approved):
tACK.

I also noticed this space after the fee rate, maybe you can quickly remove it?

Screenshot_2021-09-13_11-35-47

PR #1441, opened 2021-09-07
changelog: support copy address from tx detail

thisconnect:
Was added in 4ca933f7d64d31fa5e6d1717a1014548c27820ed


benma (approved):
LGTM. Imho better to say 'transaction' over 'tx', not everyone might know the short name

PR #1445, opened 2021-09-20
frontend: start bitbox02 specific device api

thisconnect:
- added api/response with success and fail union type
- added api/bitbox02 specifc for BitBox02
  (is similar to api/devices)
- added getDeviceInfo and setMnemonicPassphraseEnabled
- changed device info response to always include the success field
- refactored mnemonicpassphrase component, so it doesn't need
  to use callback function and works on its own

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2021 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { apiGet } from '../utils/request';
+import { TSuccess, TFail } from './response';
+
+export type TDeviceInfo = {
+    initialized: boolean;
+    mnemonicPassphraseEnabled: boolean;
+    name: string;
+    securechipModel: string;
+    version: string;
+}
+
+interface IDeviceInfo extends TSuccess {
+    deviceInfo: TDeviceInfo;
+}
+
+export const getDeviceInfo = (
+    deviceID: string
+): Promise<IDeviceInfo | TFail> => {
+    return apiGet(`devices/bitbox02/${deviceID}/info`);
+};
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -15,9 +15,9 @@
  */
 
 import { apiGet, apiPost } from '../utils/request';
-import { TSuccess, TFail } from './response';
+import { SuccessResponse, FailResponse } from './response';
 
-export type TDeviceInfo = {
+export type DeviceInfo = {
     initialized: boolean;
     mnemonicPassphraseEnabled: boolean;
     name: string;
@@ -25,19 +25,19 @@ export type TDeviceInfo = {
     version: string;
 }
 
-interface IDeviceInfo extends TSuccess {
-    deviceInfo: TDeviceInfo;
-}
+type DeviceInfoResponse = SuccessResponse & {
+    deviceInfo: DeviceInfo;
+};
 
 export const getDeviceInfo = (
     deviceID: string
-): Promise<IDeviceInfo | TFail> => {
+): Promise<DeviceInfoResponse | FailResponse> => {
     return apiGet(`devices/bitbox02/${deviceID}/info`);
 };
 
 export const setMnemonicPassphraseEnabled = (
     deviceID: string,
     enabled: boolean,
-): Promise<TSuccess | TFail> => {
+): Promise<SuccessResponse | FailResponse> => {
     return apiPost(`devices/bitbox02/${deviceID}/set-mnemonic-passphrase-enabled`, enabled);
 };
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -31,9 +31,16 @@ type DeviceInfoResponse = SuccessResponse & {
 
 export const getDeviceInfo = (
     deviceID: string
-): Promise<DeviceInfoResponse | FailResponse> => {
-    return apiGet(`devices/bitbox02/${deviceID}/info`);
+): Promise<DeviceInfo> => {
+    return apiGet(`devices/bitbox02/${deviceID}/info`)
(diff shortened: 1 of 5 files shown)


thisconnect:
This is only the first commit (or PR), updated multistep introduction will follow in another commit (or PR), but I'd like to have review for this before continuing .

frontends/web/src/api/bitbox02.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type TDeviceInfo = {
+    initialized: boolean;
+    mnemonicPassphraseEnabled: boolean;
+    name: string;
+    securechipModel: string;
+    version: string;
+}
+
+interface IDeviceInfo extends TSuccess {
```
x1ddos, line 28: All looks great but I have one suggestion: `IDeviceInfo` is too close to `TDeviceInfo`. Easy to mistake when reading the code where these two are used. How about using a `XxxResponse` convention? IDeviceInfo would become `DeviceInfoResponse`.

x1ddos: Also, this is probably beyond my comprehension but how is it that an interface "extends" from a concrete type?

thisconnect: > Also, this is probably beyond my comprehension but how is it that an interface "extends" from a concrete type?

I might abuse union types here… 🤔

thisconnect: @x1ddos when do we omit the `I` ?
`DeviceInfoResponse` vs `IDeviceInfoResponse`

thisconnect: `TFail` becomes `FailResponse` ?

frontends/web/src/routes/device/bitbox02/settings.tsx
```diff
@@ -53,8 +48,10 @@ class Settings extends Component<Props, State> {
     }
 
     private getInfo = () => {
-        apiGet(this.apiPrefix() + '/info').then(response => {
-            this.setState({ deviceInfo: response.deviceInfo });
+        getDeviceInfo(this.props.deviceID).then(response => {
+            if (response.success) {
```
benma, line 52: maybe not now, but can think about making these api calls reject() instead of resolve(), so one can skip the success check here and use `.catch()` to handle errors. it's a bigger story  to think about though.

frontends/web/src/routes/device/bitbox02/mnemonicpassphrase.tsx
```diff
@@ -16,78 +16,89 @@
...
-                    this.setState({ inProgress: false });
-                    if (success) {
-                        if (enable) {
-                            alertUser(this.props.t('bitbox02Settings.mnemonicPassphrase.successEnable'));
-                        } else {
-                            alertUser(this.props.t('bitbox02Settings.mnemonicPassphrase.successDisable'));
-                        }
-                        this.props.getInfo();
+        setMnemonicPassphraseEnabled(this.props.deviceID, enable)
+            .then((response) => {
```
benma, line 56: i think the parens around (response) are not needed

before the change it was `({ success })`, which is also pretty nice I think

frontends/web/src/routes/device/bitbox02/mnemonicpassphrase.tsx
```diff
@@ -16,78 +16,89 @@
...
+                    } else {
+                        alertUser(t('bitbox02Settings.mnemonicPassphrase.successDisable'));
                     }
-                });
+                    this.getDeviceInfo();
+                }
+            });
     }
 
+    private renderLine = (line, i) => (
```
benma, line 70: type annotations would be nice.

benma (approved):
utACK.

The last commit message does not explain the motivation for the refactor. Why did you do it? Not that I mind, but I don't understand the purpose.

frontends/web/src/routes/device/bitbox02/settings.tsx
thisconnect: agree that would be nice, lets discuss with @x1ddos but not part of this PR.

`.catch()` would also catch if there was a real error, without success:false.

frontends/web/src/routes/device/bitbox02/mnemonicpassphrase.tsx
thisconnect: Yes not needed, but also `;` is not needed and we still write it, see ASI https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#automatic_semicolon_insertion

Having it always is IMO easier to read and easier to change the code again in the future, i.e. adding arg destruction or adding a second arg.

I change it for now, but in the future I would like to just stick to what react-scripts suggest

thisconnect:
I'll update the commit message, one motiviation that is mentioned is that the component now is almost "standalone
this way.". I'll add another comment about that this is part of typing and moving all api calls to `frontends/web/src/api`. 

> The last commit message does not explain the motivation for the refactor. Why did you do it? Not that I mind, but I don't understand the purpose.

frontends/web/src/routes/device/bitbox02/settings.tsx
thisconnect: Discussed with @x1ddos and moved the success check so it is api internal and just return the deviceInfo  https://github.com/digitalbitbox/bitbox-wallet-app/pull/1445/commits/9eb6607f64abb0106ce8e154ac59faab236efab4

frontends/web/src/api/bitbox02.ts
thisconnect: Changed to `SuccessResponse`, `FailResponse` and started removing the `T` prefix here.

Also with the last commit we check now in the api `DeviceInfoResponse` for success and only return `DeviceInfo`.

thisconnect:
PTAL

x1ddos (approved):
LGTM but maybe squash some commits before merging where it makes sense?

frontends/web/src/api/bitbox02.ts
```diff
@@ -0,0 +1,50 @@
...
+    initialized: boolean;
+    mnemonicPassphraseEnabled: boolean;
+    name: string;
+    securechipModel: string;
+    version: string;
+}
+
+type DeviceInfoResponse = {
+    deviceInfo: DeviceInfo;
+} & SuccessResponse;
```
x1ddos, line 30: forgot this comment. not a big deal but pesonally i'd prefer for `SuccessResponse` to be in front because it's just easier to spot, like this:

```ts
type DeviceInfoResponse = SuccessResponse & {
    deviceInfo: DeviceInfo;
};
```

but @thisconnect up to you.

thisconnect:
I don't know @benma wanted to make me small commits so it is easier to review

> LGTM but maybe squash some commits before merging where it makes sense?

frontends/web/src/api/bitbox02.ts
thisconnect: 💯  looks better, squashed into https://github.com/digitalbitbox/bitbox-wallet-app/compare/9eb6607f64abb0106ce8e154ac59faab236efab4..c0544ed4d999040634d9482e09d61bbf65c74037

PR #1457, opened 2021-10-25
update go-ethereum and remove the modvendor tool

benma:


backend/coins/eth/etherscan/etherscan.go
```diff
@@ -555,3 +555,8 @@ func (etherScan *EtherScan) SuggestGasPrice(ctx context.Context) (*big.Int, erro
 	}
 	return (*big.Int)(&result), nil
 }
+
+// SuggestGasTipCap implements rpc.Interface.
+func (etherScan *EtherScan) SuggestGasTipCap(ctx context.Context) (*big.Int, error) {
+	panic("not implemented")
```
x1ddos, line 561: please let's not panic: `return nil, errors.New("not implemented")`

x1ddos (approved):
+88282 −362346 = -274064 :heart: 

8437a5a: update go-ethereum - i'd add, "update go-ethereum and deps"

LGTM but i'd really like to avoid new panics in rpcclient even if it's not supposed to happen (famous last words)

backend/coins/eth/etherscan/etherscan.go
benma: done

PR #1460, opened 2021-11-08
Frontend fullscreen to view

thisconnect:
The fullscreen component used on the AOPP and unlock view
have nicer design, bigger fonts, centered content and cleaner 
components. The component could also be used without
fullscreen and just as regular view.

The component should be used as a layout for a whole view
usually consisting of header, content, button area, footer. 

Added a position attribute to opt-in into fullscreen mode, the
default mode just contains the apps frame next to the sidebar
and guide.

Related to for https://github.com/thisconnect/bitbox-wallet-app/commits/frontend-passphrase-without-dialog-I


x1ddos (approved):
LGTM

PR #1462, opened 2021-11-10
Frontend custom route for passphrase

thisconnect:
There is almost never a reason to use dialogs that cover half
of the screen and have a see through overlay.

This is the first step and moves the passphrase to a custom
route, so that in the next commit we can completely remove
the dialogs.

on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1460


thisconnect:
PTAL

thisconnect:
I rebased `staging-passphrase-desgin` onto master and rebased `frontend-custom-route-for-passphrase`

PR #1464, opened 2021-11-15
frontend: make locize-pull

benma:


thisconnect:
I see, this will have to come later as the necessary change in the frontend hasn't been merged. If there is anything else that needs to be pulled from locize I'll revert that change in Locize.

benma:
> I see, this will have to come later as the necessary change in the frontend hasn't been merged. If there is anything else that needs to be pulled from locize I'll revert that change in Locize.

You changed on locize before it was on master? :scream: :joy: 

This is the only change in locize for now, so it's ok for now.

thisconnect:
Yeah I assumed we could merge my changes by now. I will revert it in Locize.

> You changed on locize before it was on master? 😱 😂

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1552

PR #1467, opened 2021-11-15
refactor multilinemarkup into its own component

thisconnect:
Similar to SimpleMarkup, this component renders strong elements
in the text into real JSX elements, but with newline support.
Each newline is mapped to return a SimpleMarkup component.

In Preact 8 there is no support for fragments so we cannot
use multilineMarkup as real component, but after migrating
to React multilineMarkup can be used as real component.


frontends/web/src/utils/markup.ts
```diff
@@ -17,16 +17,25 @@
...
+ * **SimpleMarkup** renders `foo <strong>bar</strong> baz` safely as
+ * `foo <strong>bar</strong> baz`. Anything else is rendered as
+ * sanitized text.
+ * Only one occurence of <strong> is supported to keep it simple.
+ * ### Example:
+ * ```jsx
+    <SimpleMarkup tagName="p" markup="foo <strong>bar</strong> baz" />
+ * ```
+ */
+
```
x1ddos, line 37: remove this blankline since the comment is supposed to be attached to the function? and the same for `multilineMarkup` below.

frontends/web/test/utils/markup.test.tsx
```diff
@@ -34,3 +34,23 @@ describe('SimpleMarkup', () => {
         expect(span.output().children[1]).toEqual(<strong>baz</strong>)
     });
 });
+
+describe('multilineMarkup', () => {
+    it('contains multiple lines with a strong element each', () => {
```
x1ddos, line 39: :heart:

x1ddos (approved):
> 8fc4f26: Each newline is mapped to retrun a ...

to return?

LGTM otherwise. very nice to see a test case.

frontends/web/src/utils/markup.ts
thisconnect: done

thisconnect:
fixed redrum to return

PR #1468, opened 2021-11-15
frontend: fix centered view regression

thisconnect:
Moving from fullscreen to view component in 4383f68 has
introduced a regression for AOPP workflow. Reason is that before
the fullscreen component was always centered, and now the view
component is by default left aligned.

Changed to add center prop to AOPP view.


benma (approved):
utACK

PR #1469, opened 2021-11-15
Frontend: redirect home to portfolio

BrianCraig:
Redirect to portafolio if there are active accounts.

[Video of implementation](https://streamable.com/q3q4wr)


frontends/web/src/routes/device/deviceswitch.tsx
```diff
@@ -15,20 +15,20 @@
...
+import { TDevices, TProductName } from '../../api/devices';
 import BitBox01 from './bitbox01/bitbox01';
 import { BitBox02 } from './bitbox02/bitbox02';
 import { BitBox02Bootloader } from '../../components/devices/bitbox02bootloader/bitbox02bootloader';
 import { Waiting } from './waiting';
 
 interface Props {
     devices: TDevices;
-    deviceID: string | null;
+    deviceID?: TProductName;
```
benma, line 26: deviceID is actually not the product name, it is an arbitrary ID assigned by the backend. please undo this

frontends/web/src/app.tsx
```diff
@@ -71,27 +72,36 @@ class App extends Component<Props, State> {
     /**
      * Gets fired when the route changes.
      */
-    private handleRoute = () => {
+    private handleRoute = async (e:RouterOnChangeArgs) => {
         if (panelStore.state.activeSidebar) {
             toggleSidebar();
         }
+
+        await sleep(0); // this must be run async
```
benma, line 80: why is this necessary? in my experience, sleep is usually a workaround and the real problem is elsewhere

frontends/web/src/app.tsx
```diff
@@ -71,27 +72,36 @@ class App extends Component<Props, State> {
...
-            case 'backend':
-                switch (data) {
-                case 'newTxs':
-                    apiPost('notify-user', {
-                        text: this.props.t('notification.newTxs', {
-                            count: meta.count,
-                            accountName: meta.accountName,
-                        }),
-                    });
+                case 'backend':
```
benma, line 93: needless reformatting? it's best to stick to the current formatting and keep reformatting changes to separate dedicated commits/PRs.

benma (changes requested):
Thanks, I left a few comments, and also I didn't test it yet.

frontends/web/src/app.tsx
```diff
@@ -71,27 +72,36 @@ class App extends Component<Props, State> {
...
      */
-    private handleRoute = () => {
+    private handleRoute = async (e:RouterOnChangeArgs) => {
         if (panelStore.state.activeSidebar) {
             toggleSidebar();
         }
+
+        await sleep(0); // this must be run async
+        const {accounts} = this.state;
+        const isIndex = e.url === '/' || e.url === '/index.html' || e.url === '/android_asset/web/index.html';
```
benma, line 82: seems like this and the next few lines are copied from `maybeRoute()`. Would it be possible to simply call `this.maybeRoute()` here instead?

frontends/web/src/app.tsx
BrianCraig: It's just a workaround to do an async route change, because if you do it sync preact throws an internal render error.
Screenshot from 2021-11-15 12-57-18
This is the documentation for that specific version, they don't explain why handleRoute is an async function, and they wait to a promise without explaining why. I feel that the library doesn't have a straightforward way to redirect on a redirection on that version.

benma: But why is the sleep needed to change the function to async?

You can also remove the `async` keyword and do `return Promise.resolve()` at the end of the function, which also returns a promise without requiring the use of async/await syntax.

benma:
@BrianCraig please split out the unrelated .gitignore change into another PR.

The essence of the fix seems to be changing the sidebar logo link, e.g. like this:

image

The other changes (setRedirect, etc) seem a bit too complicated. Wdyt?

PR #1475, opened 2021-11-17
frontend: show passphrase configuration in fullscreen

thisconnect:
When enabling or disabling the passphrase setting users should
fully focus on the information shown and not be distracted by other
elements on the page, therefore the view in fullscreen is more
suitable than just a dialog.

Move call to replug-the-device to be the last message of the
success view.

Introduced a new `<Main>` component for contents of the page, this
component should replace `<div class="container">`. Also left example
code of how `<Header>` and `<Guide>` components can be used.

Made also sure the `<View>` content is scrollable inside `<Main>`.

Sometimes the BitBoxApp requires the user to focus on the device
and confirm or reject something on the BitBox02.

Turned caret pointing to bb02 device into a shared component, so
it can be re-used in different places like AOPP.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1027,6 +1027,7 @@
     },
     "successDisabled": {
       "message": "Optional passphrase <strong>successfully enabled</strong>!\nYou’ll be asked to provide a passphrase from now on.",
+      "messageEnd": "Please replug your BitBox02 now.",
       "title": "Passphrase enabled"
     },
     "successEnabled": {
--- a/frontends/web/src/routes/device/bitbox02/passphrase.tsx
+++ b/frontends/web/src/routes/device/bitbox02/passphrase.tsx
@@ -20,12 +20,16 @@ import { route } from 'preact-router';
 import { getDeviceInfo, setMnemonicPassphraseEnabled } from '../../../api/bitbox02';
 import { translate, TranslateProps } from '../../../decorators/translate';
 import { SimpleMarkup } from '../../../utils/simplemarkup';
-import { Header } from '../../../components/layout';
+// This is the first time we use <View> in a <Main> component
+// keeping guide and header as example in the code
+import { /* Header, */ Main } from '../../../components/layout';
 import { Button, Checkbox } from '../../../components/forms';
 import { alertUser } from '../../../components/alert/Alert';
-import { Dialog, DialogButtons } from '../../../components/dialog/dialog';
-import { WaitDialog } from '../../../components/wait-dialog/wait-dialog';
 import { Message } from '../../../components/message/message';
+import { View, ViewButtons, ViewContent, ViewHeader } from '../../../components/view/view';
+// keeing as example for using guides in the new main component
+// import { Guide } from '../../../components/guide/guide';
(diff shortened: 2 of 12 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -1027,6 +1027,7 @@
     },
     "successDisabled": {
       "message": "Optional passphrase <strong>successfully enabled</strong>!\nYou’ll be asked to provide a passphrase from now on.",
+      "messageEnd": "Please replug your BitBox02 now.",
```
x1ddos, line 1030: it could be someone else's bitbox02 :)
how about this: Please replug the BitBox02 device now.

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -251,11 +280,15 @@ class Passphrase extends Component<Props, State> {
...
-                </div>
-            </div>
+                            <SimpleMarkup tagName="p" markup={t('passphrase.successEnabled.messageEnd')} />
+
+                        </ViewContent>
+                    </View>
+                )}
+                {/* <Guide>
+                    <Entry key="whatisapassphrase" entry={t('guide.passphrase.whatisapassphrase')} />
+                </Guide> */}
```
x1ddos, line 290: is the guide entry not ready yet?

x1ddos (commented):
very nice! i love it how there's no more distractions, just the main content.

there's one inconsistency. whem i'm on any page like this one:

image

and before ticking the "I understand ..." checkbox i realize i don't understand and want to go back. i click the "back" button, expecting it to go the the previous screen but it closes the whole thing.

thisconnect:
Thanks for testing,

I changed
- the back button to actually go 1 step back instead of going back to the setting view
- added a close button to go back to the settings view
- updated the styling especially of summary page

For now I opened a new PR here https://github.com/digitalbitbox/bitbox-wallet-app/pull/1478 but this is only temporary I'll check Monday if and how to split them up into different commits.

Screen Shot 2021-11-20 at 12 08 58 PM

Screen Shot 2021-11-20 at 12 09 07 PM

frontends/web/src/locales/en/app.json
thisconnect: I agree, directly addressing the user can have a strong effect, i.e. when we really want to make them aware about something the should or shouldn't do.

But the effect is kind of gone if we always use direct addressing.

I will go through the text and propose some changes.

frontends/web/src/routes/device/bitbox02/passphrase.tsx
thisconnect: Currently there is no additional guide for passphrase, but we now make the user read through the information instead of providing an optional guide.

Is there something missing that you'd like to see in the guide?

If yes then depending on what kind of information it is we should consider adding it to the information shown. Maybe the guide could contain some technical details that we don't want to show to regular users?

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1478

x1ddos:
> superseded by #1478

so, close this?

PR #1476, opened 2021-11-17
android: target api level 30, enforced by Google Play

benma:
Google Play now enforces a minimum level of 30.

The `setAllowFileAccess` change is because it defaulted to true before
and is defaulting to false now.

See also: https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)

### DIFF
--- a/frontends/android/BitBoxApp/app/build.gradle
+++ b/frontends/android/BitBoxApp/app/build.gradle
@@ -6,7 +6,7 @@ android {
     defaultConfig {
         applicationId "ch.shiftcrypto.bitboxapp"
         minSdkVersion 21
-        targetSdkVersion 29
+        targetSdkVersion 30
         versionCode 32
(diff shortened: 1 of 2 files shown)


x1ddos:
hm... but https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean) says:

> Assets and resources are still accessible using file:///android_asset and file:///android_res

so, why do we need file access? :thinking:

x1ddos:
ok so the theory is that webview is still trying to access filesystem anyway, which is also where our request interceptor kicks in and serves `file:///android_asset/...`

LGTM but probably makes sense to revert this and migrate to https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader in a future release.

x1ddos (approved):

PR #1478, opened 2021-11-20
Frontend passphrase fullscreen styling

thisconnect:
Continuation from https://github.com/digitalbitbox/bitbox-wallet-app/pull/1475#issuecomment-974635933

When enabling or disabling the passphrase setting users should
fully focus on the information shown and not be distracted by other
elements on the page, therefore the view in fullscreen is more
suitable than just a dialog.

Changed:

Introduced a new <Main> component for contents of the page, this
component should replace <div class="container">. Also left example
code of how <Header> and <Guide> components can be used.

- changed back button to go 1 step back in the passphrase infos
  instead of going back to the settings view
- turned caret pointing to bb02 device into a shared component, so
  it can be re-used in different places like AOPP.
- changed View component's default, fullscreen and textCenter props
  are now opt-in
- added minHeight prop so the height of the view content can be
  controlled so that the buttons are always in the same position
- nicer styling for understood checkbox, now in a red box that
  changes to green once toggled
- improved positioning of the content area, buttons should now
  always be in the same position
- passphrase enable/disable success should disappear once the
  device is unplugged
- added missing successDisabled.messageEnd translation

Tested:

- different breakpoints small (portrait and landscape), medium
  screen centered and large screen with bigger font
- retesting unlock and AOPP workflow as those also use the view
  component
- tested that content is always scrollable if screen height is
  smaller than the content
- checked paddings and overall look

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1027,6 +1027,7 @@
     },
     "successDisabled": {
       "message": "Optional passphrase <strong>successfully enabled</strong>!\nYou’ll be asked to provide a passphrase from now on.",
+      "messageEnd": "Please replug the BitBox02 now.",
       "title": "Passphrase enabled"
     },
     "successEnabled": {
--- a/frontends/web/src/routes/device/bitbox02/passphrase.tsx
+++ b/frontends/web/src/routes/device/bitbox02/passphrase.tsx
@@ -20,12 +20,26 @@ import { route } from 'preact-router';
 import { getDeviceInfo, setMnemonicPassphraseEnabled } from '../../../api/bitbox02';
 import { translate, TranslateProps } from '../../../decorators/translate';
 import { SimpleMarkup } from '../../../utils/simplemarkup';
-import { Header } from '../../../components/layout';
+// This is the first time we use <View> in a <Main> component
+// keeping guide and header as example in the code
+import { /* Header, */ Main } from '../../../components/layout';
 import { Button, Checkbox } from '../../../components/forms';
 import { alertUser } from '../../../components/alert/Alert';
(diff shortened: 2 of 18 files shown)


thisconnect:
As mentioned in previous PR, I'll check the text again especially concerning direct addressing "you" "yours", but that is independent of the styling PR.

thisconnect:
@x1ddos PTAL

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -251,11 +351,17 @@ class Passphrase extends Component<Props, State> {
...
+                            <SimpleMarkup tagName="p" markup={t(
+                                passphraseEnabled
+                                    ? 'passphrase.successDisabled.messageEnd'
+                                    : 'passphrase.successEnabled.messageEnd')} />
+                        </ViewContent>
+                    </View>
+                )}
+                {/* <Guide>
+                    <Entry key="whatisapassphrase" entry={t('guide.passphrase.whatisapassphrase')} />
+                </Guide> */}
```
x1ddos, line 363: remove this since there's no guide anymore?

thisconnect: I really'd like the to keep it here until we have another example that uses `<View>` and a guide component.

frontends/web/src/locales/en/app.json
```diff
@@ -1027,6 +1027,7 @@
     },
     "successDisabled": {
       "message": "Optional passphrase <strong>successfully enabled</strong>!\nYou’ll be asked to provide a passphrase from now on.",
+      "messageEnd": "Please replug your BitBox02 now.",
```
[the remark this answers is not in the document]
thisconnect: Fixed, but well have to fix the other messageEnd text in Locize.

PR #1480, opened 2021-11-22
frontend: link update button directly to download page

thisconnect:
There is no reason to go through the '/start' redirect, which
could also be used to add custom welcome-to-bitbox banner on the
download page or a custom upgrade-the-bitboxapp instruction.


x1ddos (approved):
LGTM

PR #1482, opened 2021-11-22
frontend: change from yarn to npm

thisconnect:
Changed ci to use npm so it doesn't have to install yarn, and
moved from yarn.lock to package-lock.json.

Other reason besides installing yarn is that node already comes
with yarn and both can read package-lock.json. It used to be
that yarn was faster but npm improved there should not be any
noticable difference between the two.

Documented why some dependencies use a semver range and how to
update those vs. update exact version.

Also had to up nodejs version in appveyor because otherwise
it installed an older version of npm


thisconnect:
@benma PTAL

.travis.yml
```diff
@@ -14,7 +14,7 @@ addons:
       - make
 
 cache:
-  yarn: true
+  yarn: false
```
benma, line 17: just delete? i assume false is the default, or not?

Makefile
```diff
@@ -36,7 +35,6 @@ envinit:
 #  - add to $PATH: /usr/local/opt/go@1.16/bin
 osx-init:
 	/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
-	brew install yarn
```
benma, line 39: brew install npm instead?

README.md
```diff
@@ -51,7 +51,7 @@ The below instructions assume a unix environment.
 The following dependencies need to be installed:
 
 - [Go](https://golang.org/doc/install) version 1.16
-- [Yarn](https://yarnpkg.com/en/) for managing the web UI deps
+- [Node.js](https://nodejs.org/) for managing the web UI deps
```
benma, line 54: add npm?

appveyor.yml
```diff
@@ -1,7 +1,7 @@
 image: Visual Studio 2019
 
 environment:
-  nodejs_version: "12"
+  # nodejs_version: "14"
```
benma, line 4: uncomment or delete

benma (approved):
tACK

.travis.yml
thisconnect: We don't use `.travis.yml` but I thought I change it anyways....
I guess we could remove the travis file completely, but that has nothing todo with change from yarn to npm, so not sure what to do...

Makefile
thisconnect: no, node already ships npm but maybe not always the latest greatest.

README.md
thisconnect: No see other comment.

appveyor.yml
thisconnect: deleted

PR #1483, opened 2021-11-24
Add Buy CTA on Account Info #1473

### LINKED ISSUE #1473: Add intro text on empty summary page
After the setup process, the user is taken to the account summary page without any info and just an empty chart.

Add a simple CTA to buy crypto/Bitcoin. This message should only be shown if the wallet is empty and there has been no transactions in that wallet.

The CTA should take them to the Buy crypto flow `/buy/info`.

empty-wallet

BrianCraig:
Adds a buy CTA action for available to buy coins if the amount is 0
screencapture-localhost-8080-account-v0-6d908ae1-teth-0-2021-11-23-22_45_43

Closes #1473

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -21,6 +21,14 @@
   },
   "accountInfo": {
     "address": "Address",
+    "buyCTA": {
+      "buy": "Buy {{unit}}",
+      "buyCrypto": "Buy Crypto",
+      "information": {
+        "looksEmpty": "Looks like this wallet is empty.",
+        "start": "Get started by depositing some coins to the wallet or buying directly in the BitBoxApp."
+      }
+    },
     "extendedPublicKey": "Extended public key",
     "label": "Account info",
     "title": "Account information",
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -33,6 +33,7 @@ import { Transactions } from '../../components/transactions/transactions';
 import { load } from '../../decorators/load';
 import { translate, TranslateProps } from '../../decorators/translate';
 import { apiGet } from '../../utils/request';
+import { BuyCTA } from './info/buyCTA';
 import * as style from './account.module.css';
 import { isBitcoinBased } from './utils';
 
@@ -310,6 +311,7 @@ class Account extends Component<Props, State> {
                     {status.synced && this.dataLoaded() && isBitcoinBased(account.coinCode) && <HeadersSync coinCode={account.coinCode} />}
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
+                            { this.supportsBuy() && balance && (balance.available.amount === '0') && <BuyCTA code={code} unit={balance.available.unit} /> }
                             <Status
                                 className="m-bottom-default"
                                 hidden={!WithCoinTypeInfo.includes(code)}
--- a/frontends/web/src/routes/account/info/buyCTA.tsx
+++ b/frontends/web/src/routes/account/info/buyCTA.tsx
@@ -1,0 +1,44 @@
+import { FunctionalComponent, h } from 'preact';
+import { route } from 'preact-router';
+import { Coin } from '../../../api/account';
+import { Button } from '../../../components/forms';
+import { translate, TranslateProps } from '../../../decorators/translate';
+import { Balances } from '../summary/accountssummary';
+import * as styles from './buyCTA.module.css';
+
+interface BuyCTAProps {
+    code?: string;
+    unit?: string;
+}
+
+type Props = BuyCTAProps & TranslateProps;
+
+const BuyCTAComponent: FunctionalComponent<Props> = ({code, unit, t}) => {
+    const onCTA = () => route(code ? `/buy/info/${code}` : `/buy/info`);
+    return (
+    <div className={`${styles.main} columns-container`}>
+        <h3 className="column">{t('accountInfo.buyCTA.information.looksEmpty')}</h3>
+        <h3 className="column">{t('accountInfo.buyCTA.information.start')}</h3>
+        <div className="column">
+            <Button primary onClick={onCTA}>{unit ? t('accountInfo.buyCTA.buy', {unit}) : t('accountInfo.buyCTA.buyCrypto')}</Button>
+        </div>
+    </div>);
+};
+
+export const BuyCTA = translate<BuyCTAProps>()(BuyCTAComponent);
+
+const isBitcoinCoin = (coin: Coin) => (coin === 'BTC') || (coin === 'TBTC');
+
+export const AddBuyOnEmptyBalances: FunctionalComponent<{balances?: Balances}> = ({balances}) => {
+    if (balances === undefined) {
+        return null;
+    }
+    const balanceList = Object.entries(balances);
+    if (balanceList.some(entry => entry[1].available.amount !== '0')) {
+        return null;
+    }
+    if (balanceList.map(entry => entry[1].available.unit).every(isBitcoinCoin)) {
+        return <BuyCTA code={balanceList[0][0]} unit={'BTC'} />;
+    }
+    return <BuyCTA />;
+};
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -30,13 +30,14 @@ import Spinner from '../../../components/spinner/ascii';
 import { debug } from '../../../utils/env';
 import { apiWebsocket } from '../../../utils/websocket';
 import { Chart } from './chart';
+import { AddBuyOnEmptyBalances } from '../info/buyCTA';
 import * as style from './accountssummary.module.css';
 
 interface AccountSummaryProps {
     accounts: accountApi.IAccount[];
 }
 
-interface Balances {
+export interface Balances {
     [code: string]: accountApi.IBalance;
 }
 
@@ -189,7 +190,7 @@ class AccountsSummary extends Component<Props, State> {
 
     public render() {
         const { t, accounts } = this.props;
-        const { exported, data } = this.state;
+        const { exported, data, balances } = this.state;
         return (
             <div className="contentWithGuide">
                 <div className="container">
@@ -215,6 +216,7 @@ class AccountsSummary extends Component<Props, State> {
                     </Header>
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
+                            { ( accounts.length === Object.keys(balances || {}).length ) && <AddBuyOnEmptyBalances balances={balances} /> }
                             {data ? (
                                 <Chart
                                     dataDaily={data.chartDataMissing ? undefined : data.chartDataDaily}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Add Buy CTA on empty Account overview and summary views
 - Fix an Android app crash when opening the app after first rejecting the USB permissions
 - Change label to show 'Fee rate' or 'Gas price' for custom fees
 - Change label 'Send all' label to 'Send selected coins' if there is a coin selection
--- a/frontends/web/src/routes/account/info/buyCTA.module.css
+++ b/frontends/web/src/routes/account/info/buyCTA.module.css
@@ -1,0 +1,9 @@
+.main {
+    margin-bottom: calc(var(--space-default) * 2);
+}
+
+.main > * {
+    margin-right: 0 !important;
+    text-align: center;
+    font-weight: 400;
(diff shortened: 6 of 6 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -21,6 +21,13 @@
   },
   "accountInfo": {
     "address": "Address",
+    "buyCTA": {
+      "buy": "Buy {{unit}}",
+      "information": [
+        "Looks like your wallet is empty.",
+        "Get started by depositing some coins into your wallet or buying directly in the BitBoxApp."
```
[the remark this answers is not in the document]
thisconnect: I'll ping SpaghettiMonster to have a look at the text and will paste it here if there is any update.

thisconnect (changes requested):
Nice @BrianCraig 👍  added some comments

frontends/web/src/locales/en/app.json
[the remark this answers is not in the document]
thisconnect: Could you change the text to 

"Looks like this wallet is empty."

"Get started by depositing some coins to the wallet or buying directly in the BitBoxApp."

frontends/web/src/routes/account/info/buyCTA.tsx
```diff
@@ -0,0 +1,30 @@
...
+import * as styles from './buyCTA.module.css';
+
+interface BuyCTAProps {
+    code: string;
+    balance?: IBalance
+}
+
+type Props = BuyCTAProps & TranslateProps;
+
+const BuyCTA :FunctionalComponent<Props>  = ({code, balance, t}) => {
```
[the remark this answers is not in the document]
BrianCraig: Cool, i don't have a personal defined way of what is better, still there is one tradeoff that is not listed, meanwhile you are defining the type of the input(`Props` or `PropsWithChildren<Props>`), and the output (`JSX.Element` or `ReactElement<any, any> | null`), using the `FunctionComponent` defines a more generic way and makes it more integral, you don't need to import or know exact types of the rendering lib you are using. 

For example this is the definition of FunctionalComponent on preact:
```typescript
interface FunctionalComponent<P = {}> {
  (props: RenderableProps<P>, context?: any): VNode<any> | null;
  displayName?: string;
  defaultProps?: Partial<P>;
}
```
and this is on React
```typescript
interface FunctionComponent<P = {}> {
  (props: PropsWithChildren<P>, context?: any): ReactElement<any, any> | null;
  propTypes?: WeakValidationMap<P> | undefined;
  contextTypes?: ValidationMap<any> | undefined;
  defaultProps?: Partial<P> | undefined;
  displayName?: string | undefined;
}
```

the specifics typings of every engine for using with more internal tools like `React.Children API` are defined by default in each `FunctionComponent` typing, and switching between libs os more difficult if every component is defined strictly to their input/output. Of course this is not a big problem as it seems because any function declared as `(props):Props => JSX.Element` would be a FC in any lib.

This lead me to think that if we are going to migrate to React, we will need to migrate a lot of `RenderedProps<P>` to `PropsWithChildren<P>`. Wouldnt be easier to have a typing definition that abstract this difference, example using `WithProps<P> = RenderedProps<P>`, and so when we swithc to React we only have to change one file?

Still, if the team decides to use `(props):Props => JSX.Element` as default typing for components im okay with that, as more integral decisions are made its easier.

frontends/web/src/routes/account/account.tsx
```diff
@@ -310,6 +311,7 @@ class Account extends Component<Props, State> {
                     {status.synced && this.dataLoaded() && isBitcoinBased(account.coinCode) && <HeadersSync coinCode={account.coinCode} />}
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
+                            { this.supportsBuy() && <BuyCTA code={code} balance={balance} /> }
```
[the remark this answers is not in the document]
BrianCraig: This should be visible when all accounts have no balance?

Should this show `Buy BTC` or `Buy Crypto`?

Should this go to `/buy/info` or `/buy/moonpay/[btccode]`?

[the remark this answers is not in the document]
thisconnect: Yes when the overall total is 0 it should be visible.

> Should this show Buy BTC or Buy Crypto?

I have to check but I believe we already have some logic somewhere that decides if coins other than bitcoin can be bought.

> Should this go to `/buy/info` or `/buy/moonpay/[btccode]`?

On the summary just link to `/buy/info` as there is an important disclaimer and there will be a coin seletion if other coins can be bought.

BrianCraig:
@thisconnect I've made some changes, removing the logic of the BuyCTA component.
Also on the accounts page, it will only display if all baalnces are zero, and show a BTC instead of a generic if its only currency is BTC. And also updating the commit message because it had some uppercase letters

Is this ok for you?

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -61,6 +62,17 @@ interface BalanceRowProps {
...
+const BuyCTAActual: FunctionalComponent<{balances?: Balances}> = ({balances}) => {
+    if (balances === undefined)
+        return null;
+    const balanceList = Object.entries(balances);
+    if (balanceList.some(entry => entry[1].available.amount !== '0'))
+        return null;
+    if ((balanceList.length === 1) && balanceList[0][1].available.unit === 'BTC')
+        return <BuyCTA code={balanceList[0][0]} unit={'BTC'} />;
+    return <BuyCTA />;
+};
```
thisconnect, line 74: I was thinking how this could be simpler, but I think it is ok.. can you change the formatting to

```diff
diff --git a/frontends/web/src/routes/account/summary/accountssummary.tsx b/frontends/web/src/routes/account/summary/accountssummary.tsx
index b160edd5..d32635a0 100644
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -62,14 +62,21 @@ interface BalanceRowProps {
     coinCode: accountApi.CoinCode;
 }
 
-const BuyCTAActual: FunctionalComponent<{balances?: Balances}> = ({balances}) => {
-    if (balances === undefined)
+type BuyCTAActualProps = {
+    balances?: Balances
+}
+
+const BuyCTAActual: FunctionalComponent<BuyCTAActualProps> = ({balances}) => {
+    if (balances === undefined) {
         return null;
+    }
     const balanceList = Object.entries(balances);
-    if (balanceList.some(entry => entry[1].available.amount !== '0'))
+    if (balanceList.some(entry => entry[1].available.amount !== '0')) {
         return null;
-    if ((balanceList.length === 1) && balanceList[0][1].available.unit === 'BTC')
+    }
+    if ((balanceList.length === 1) && balanceList[0][1].available.unit === 'BTC') {
         return <BuyCTA code={balanceList[0][0]} unit={'BTC'} />;
+    }
     return <BuyCTA />;
 };
 ```

Comment (thisconnect) [reply frontends/web/src/routes/account/summary/accountssummary.tsx:74]: This component conditionally renders the CTA button and finds out if it is BTC only, I think we could hide it a bit and move to `info/buyCTA`.

Maybe name it `<AddBuyOnEmptyBalances balances={balances} />` ?

**Review by thisconnect — CHANGES_REQUESTED** [review]

File: frontends/web/src/routes/account/summary/accountssummary.tsx (hunk 4 above)
Comment (BrianCraig) [reply frontends/web/src/routes/account/summary/accountssummary.tsx]: Ok, i migrated that small component but had to export `Balances` interface. is that ok?

**Comment by thisconnect** [comment]
Could you add "Why" to the commit body, maybe take some of the text from the issue here https://github.com/digitalbitbox/bitbox-wallet-app/issues/1473

other than that LGTM

**Comment by thisconnect** [comment]
@BrianCraig could you make an entry in https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/CHANGELOG.md

maybe just something like "Add Buy CTA on empty Account overview and summary views"

**Comment by thisconnect** [comment]
@benma PTAL

File: frontends/web/src/routes/account/summary/accountssummary.tsx (hunk 4 above)
Comment (thisconnect) [reply frontends/web/src/routes/account/summary/accountssummary.tsx]: yes thank you

**Comment by BrianCraig** [comment]
> Could you add "Why" to the commit body, maybe take some of the text from the issue here #1473
> 
> other than that LGTM

> @BrianCraig could you make an entry in https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/CHANGELOG.md
> 
> maybe just something like "Add Buy CTA on empty Account overview and summary views"

This changes had been addressed :+1:

File: frontends/web/src/routes/account/info/buyCTA.tsx [hunk 5]
```diff
@@ -0,0 +1,41 @@
...
+
+export const AddBuyOnEmptyBalances: FunctionalComponent<{balances?: Balances}> = ({balances}) => {
+    if (balances === undefined) {
+        return null;
+    }
+    const balanceList = Object.entries(balances);
+    if (balanceList.some(entry => entry[1].available.amount !== '0')) {
+        return null;
+    }
+    if ((balanceList.length === 1) && balanceList[0][1].available.unit === 'BTC') {
```
thisconnect, line 37: Hi I am testing, here when there is more that 1 bitcoin account, this is showing 'Buy Crypto' but ideally it would then show 'Buy Bitcoin'

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -215,6 +216,7 @@ class AccountsSummary extends Component<Props, State> {
                     </Header>
                     <div className="innerContainer scrollableContainer">
                         <div className="content padded">
+                            <AddBuyOnEmptyBalances balances={balances} />
```
thisconnect, line 219: Small issue:
I ran into an edgecase, while testing where the accounts are not fully loaded but the CTA quickly flickers shows and then disappears, tested with a few empty and non empty LTC and BTC accounts. We have syncStatus in the state but I am not sure if that is helpful.

As accounts are there, but balances can be undefined and then get populated one by one, I think the condition to render this component could be something like this `accounts.length === balances && Object.keys(balances).length`

frontends/web/src/routes/account/info/buyCTA.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <h3 className="column">{t('accountInfo.buyCTA.information.start')}</h3>
+        <div className="column">
+            <Button primary onClick={onCTA}>{unit ? t('accountInfo.buyCTA.buy', {unit}) : t('accountInfo.buyCTA.buyCrypto')}</Button>
+        </div>
+    </div>);
+};
+
+export const BuyCTA = translate<BuyCTAProps>()(BuyCTAComponent);
+
+const isBitcoinCoin = (coin: Coin) => (coin === 'BTC') || (coin === 'TBTC');
```
thisconnect, line 30: I thought we can use https://github.com/digitalbitbox/bitbox-wallet-app/blob/d6b33a6d25e394e11bcb8055db41ff77129769ca/frontends/web/src/routes/account/utils.ts#L20

but Coin is coinCode but uppercase https://github.com/digitalbitbox/bitbox-wallet-app/blob/d6b33a6d25e394e11bcb8055db41ff77129769ca/frontends/web/src/api/account.ts#L27-L31

not sure what to do, maybe keep this function but move to `routes/accounts/utils.ts`, @x1ddos what do you think?

also `isBitcoinCoin` name is funny but somehow makes sense :)

thisconnect (changes requested):
Very nice, I think it is almost there, added 2 comments

frontends/web/src/routes/account/summary/accountssummary.tsx
thisconnect: @BrianCraig can you try to reprodce?
(maybe reconfigure accounts or delete appfolder.dev to reproduce)

Could you try to only render the component on the summary page if `accounts.length === balances && Object.keys(balances).length` or similar.

BrianCraig: Screenshot from 2021-12-09 08-03-04
Screenshot from 2021-12-09 08-03-24
Screenshot from 2021-12-09 08-04-25
Screenshot from 2021-12-09 08-04-46

thisconnect: The issue only appears during balance / account loading. And the CTA quickly appears but then as soon as there is a balance it disapears, sometimes jus a flickr, sometimes for a few seconds. depends on how fast the accounts/balance load

BrianCraig: :+1: I'll have to ckeck with the profilig tools to see what states have the balances and acounts

BrianCraig: I chnged to show only when accounts and balances matcches quantity as you said :+1:

thisconnect (approved):
LGTM 👍

PR #1484, opened 2021-11-24
remove unnescesary container component

BrianCraig:
Removal of unnescesary Container Component.

why: Container component is acting like a preact router configurator, and injecting a toggleSidebar prop to all it's children. This prop is never used and a direct usage of the store is used instead always, this commit simplifies the reading of the App.tsx, removing unnescesary overhead.


thisconnect:
Nice thanks!

Tested that sidebar opens/closes with hamburger, closes by clicking overlay background.

LGTM, please update the commit body with the PR description.

BrianCraig:
@thisconnect Is this commit message okay?

thisconnect:
@BrianCraig the text is good, just remove `Why:` and I usually add context on the first line `frontend:`, see other commits.

and just a nit: first line should be lowercase and the lines should be
wrapped at 72 chars

then it is perfect!!!! ❤️

BrianCraig:
@thisconnect can this PR be merged?

thisconnect (approved):
LGTM

PR #1487, opened 2021-11-24
Total balance per coin

mathantunes:
## Backend

New endpoint on accounts to return the subtotal per coin.

## Front end

* Minor changes to use async/await
* Render subtotal row after X accounts of the same coin

## TODO:
* Issue when starting the application, seems like the new endpoint fails - maybe something is not completely loaded yet.
* Styling the subtotal row

## Example with 2 bitcoin accounts
image

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


benma:
Finished in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1620

PR #1488, opened 2021-11-24
passphrase design staging -> master

thisconnect:


BrianCraig:
I've been testing it and it looks really good in UI/UX terms.
There is only one thing that i didn't fully understood, is why there is a `disabled`/`enabled` param in the route? Is this needed? I would suggest that this param should be ommited, and the information shown (enable or disable flow) should be determined by the state of the device

What do you think of this?

thisconnect:
> There is only one thing that i didn't fully understood, is why there is a disabled/enabled param in the route? Is this needed? I would suggest that this param should be ommited, and the information shown (enable or disable flow) should be determined by the state of the device

That would work, but the only reason why I wanted to pass it along is because it is already known in the settings component and that the passphrase doesn't have to wait and fetch it... but I guess it is probably fast enough that nobody would notice and we might not need a loading spinner icon... I'll try.

thisconnect:
@BrianCraig I made a PR that removes the param in the route as you requested
https://github.com/digitalbitbox/bitbox-wallet-app/pull/1496/files

Do you think this is better? If yes why?

I currently slightly prefer the passing param because we already have this info in the settings, on the otherhand with this PR it makes the component easier to use elsewhere (in example on the last step of the setup).

I am not sure what to write in the commit :)
Also the other PR is just to discuss, it is not 100% ready there is some lint error.

also cc @x1ddos

BrianCraig:
> @BrianCraig I made a PR that removes the param in the route as you requested https://github.com/digitalbitbox/bitbox-wallet-app/pull/1496/files
> 
> Do you think this is better? If yes why?
> 
> I currently slightly prefer the passing param because we already have this info in the settings, on the otherhand with this PR it makes the component easier to use elsewhere (in example on the last step of the setup).
> 
> I am not sure what to write in the commit :) Also the other PR is just to discuss, it is not 100% ready there is some lint error.
> 
> also cc @x1ddos

I like it more because having `enabled` or  `disabled` in your route is not something the user can decide. For example try to go manually to the analogous route, that will not work, and there should not be specification on what to do in that case because it should not exist that case. If you are passing state through a url parameter is not a good thing. That is why not having a state param in your url is good. Aside from that, im not so much into commit names, for me if i would be owner IDGAF about the best commit message, AFAIK the code its good it's all fine for me. LGTM

thisconnect:
Thanks for the explanation, makes sense, I will fix and update my PR.

> for me if i would be owner IDGAF about the best commit message

well I do care.

x1ddos:
all looks nice except one thing i noticed. if i go to "buy bitcoin" with the moonpay iframe, no matter how much i click on the sidebar hamburger, can't get it to open the sidebar:

image

i think it's very important for narrow scripts where the sidebar is normally hidden.

x1ddos:
actually, it seems to be regression in at least the last release.

x1ddos (approved):
LGTM

PR #1498, opened 2021-12-01
Migrate RefCallback to RefObject

BrianCraig:
Migrate usages of RefCallback to RefObject for compatibility with React 17.x
```typescript
type RefObject<T> = { current?: T | null };
type RefCallback<T> = (instance: T | null) => void;
```

### DIFF
--- a/frontends/web/src/components/transactions/transaction.tsx
+++ b/frontends/web/src/components/transactions/transaction.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { Component, h } from 'preact';
+import { Component, createRef, h } from 'preact';
 import * as accountApi from '../../api/account';
 import { Input } from '../../components/forms';
 import { translate, TranslateProps } from '../../decorators/translate';
@@ -44,8 +44,8 @@ interface TransactionProps extends accountApi.ITransaction {
 type Props = TransactionProps & TranslateProps;
 
 class Transaction extends Component<Props, State> {
-    private input!: HTMLInputElement;
-    private editButton!: HTMLButtonElement;
+    private input = createRef<HTMLInputElement>();
+    private editButton = createRef<HTMLButtonElement>();
 
     public readonly state: State = {
         transactionDialog: false,
@@ -58,7 +58,7 @@ class Transaction extends Component<Props, State> {
             year: 'numeric',
             month: 'numeric',
             day: 'numeric',
-        };
+        } as Intl.DateTimeFormatOptions;
         return new Date(Date.parse(time)).toLocaleString(this.context.i18n.language, options);
     }
 
@@ -96,21 +96,13 @@ class Transaction extends Component<Props, State> {
     }
 
     private focusEdit = () => {
-        if (this.editButton) {
-            this.editButton.blur();
+        if (this.editButton.current) {
+            this.editButton.current.blur();
         }
-        if (this.state.editMode && this.input) {
-            this.input.scrollLeft = this.input.scrollWidth;
-            this.input.focus();
+        if (this.state.editMode && this.input.current) {
+            this.input.current.scrollLeft = this.input.current.scrollWidth;
+            this.input.current.focus();
         }
-    }
-
-    private setInputRef = input => {
-        this.input = input;
-    }
-
-    private setEditButtonRef = button => {
-        this.editButton = button;
     }
 
     public render() {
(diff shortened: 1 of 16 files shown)


frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -96,23 +96,15 @@ class Transaction extends Component<Props, State> {
...
-            this.editButton.blur();
+        if (this.editButton.current) {
+            this.editButton.current.blur();
         }
-        if (this.state.editMode && this.input) {
-            this.input.scrollLeft = this.input.scrollWidth;
-            this.input.focus();
+        if (this.state.editMode && this.input.current) {
+            this.input.current.scrollLeft = this.input.current.scrollWidth;
+            this.input.current.focus();
```
thisconnect, line 104: I get an error:
- goto transaction overview
- click on the magnifying glass
- click on the edit button

But it checks for `this.input.current`, so I am not sure why this happens 🤔 

Screen Shot 2021-12-01 at 6 03 58 PM

frontends/web/src/routes/account/send/feetargets.tsx
```diff
@@ -58,8 +58,11 @@ class FeeTargets extends Component<Props, State> {
         options: null,
     };
 
+    private input = createRef<HTMLInputElement & {autofocus: boolean}>();
```
thisconnect, line 61: Sorry for this 🙈  this was me trying to get autofocus working, I hope we will have less issues with React

frontends/web/src/routes/device/bitbox01/unlock.jsx
```diff
@@ -52,8 +54,8 @@ class Unlock extends Component {
     }
 
     focus() {
-        if (this.passwordInput) {
-            this.passwordInput.focus();
+        if (this.passwordInput.current) {
+            this.passwordInput.current.focus();
```
thisconnect, line 58: I don't understand why, but this errors on the legacy BitBox01 screen

Screen Shot 2021-12-01 at 6 26 58 PM

Screen Shot 2021-12-01 at 6 27 07 PM

frontends/web/src/routes/device/bitbox01/setup/seed-create-new.jsx
```diff
@@ -48,13 +48,16 @@ class SeedCreateNew extends Component {
...
+    backupPasswordInput = createRef();
+
     componentDidMount () {
         this.checkSDcard();
     }
 
     validate = () => {
         // @ts-ignore
-        if (!this.walletNameInput || !this.walletNameInput.validity.valid || !this.validAgreements()) {
+        if (!this.walletNameInput.current || !this.walletNameInput.current.validity.valid || !this.validAgreements()) {
```
thisconnect, line 60: I get an error here, not sure why this happens 🤔 

Screen Shot 2021-12-01 at 6 50 29 PM

thisconnect (changes requested):
Nice, there are a few cases I get an error, but I don't understand why, I think they are all related to the input. Could you look at the transaction detail where you can set a transaction note?

thisconnect:
@BrianCraig if there is no straight forward way we should merge this PR into `staging-react` and fix it there. There is no need to fix the errors if it is a Preact problem.

frontends/web/src/routes/account/send/feetargets.tsx
BrianCraig: no problema!

frontends/web/src/routes/device/bitbox01/setup/seed-create-new.jsx
BrianCraig: This was because i was not forwarding the ref, fixed when changed to refInput prop

frontends/web/src/routes/device/bitbox01/unlock.jsx
BrianCraig: This was because i was not forwarding the ref, fixed when changed to refInput prop

BrianCraig:
> Nice, there are a few cases I get an error, but I don't understand why, I think they are all related to the input. Could you look at the transaction detail where you can set a transaction note?

Yes, i was using incorrectly the ref, passing it as a `ref` prop and not using `forwardRef`, so all the refs where Input Component Refs and not input element refs

BrianCraig:
> @BrianCraig if there is no straight forward way we should merge this PR into `staging-react` and fix it there. There is no need to fix the errors if it is a Preact problem.

Yes, better doing that, so we can test more this commit

frontends/web/src/routes/device/bitbox01/setup/seed-create-new.jsx
thisconnect: It works now for transaction note `<Input>` 👍 , but here and a few other places in the BB01 setup it is using a different component `<PasswordInput>` and `<PasswordRepeatInput>`. Both use Input internally, and for focus these 2 components could forward the inputRef. But the ref is also used to access some internal function of the component 🙈 

I believe there 3 have still the issue

Screen Shot 2021-12-02 at 10 55 00 PM

thisconnect:
this was merged into `staging-react` lets close this one ok @BrianCraig ?

BrianCraig:
already merged into `staging-react`

PR #1502, opened 2021-12-06
Staging React Multiple fixes

BrianCraig:
Added multiple changes for a migration
to React.

- Add react-router-dom @zxing and lightweight-charts
- export react-i18n HOC & types to decorator/translation.ts
- migrate string to object style props
- migrate preact import & update to react types
- migrate React.SyntheticEvent on event callbacks
- disabled eslint settings that should be re-enabled
jsx-a11y/anchor-is-valid, jsx-a11y/alt-text
- disabled noImplicitAny for tsconfig that
should be re-enabled
- Fixed some uses of React.Children API
- Create Router.tsx with react-router-dom routes
- Changed multiple url navigations to local routes
- fixed some object destructuring of null class component states
- Created an imperative route fn as in preact router


frontends/web/.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": ["react-app"],
+  "rules": {
+    "jsx-a11y/anchor-is-valid": 0,
```
BrianCraig, line 4: There are some linting problems, but they are non-breaking changes. I disabled that rule just for making a more simple PR.

frontends/web/package.json
```diff
@@ -14,10 +14,13 @@
     "typescript": "^4.4.2"
   },
   "dependencies": {
+    "@zxing/library": "0.18.3",
     "i18next": "20.6.0",
```
BrianCraig, line 18: Same versions as `master`

frontends/web/src/routes/router.tsx
```diff
@@ -0,0 +1,122 @@
+import React, { FunctionComponent } from "react";
```
BrianCraig, line 1: I extracted the routes from main, and made a Router Component because it made easier to maintain the routes.

frontends/web/src/routes/router.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <Route path="/">
+            <Route index element={Homepage} />
+            <Route path="device/:deviceID" element={Device} />
+            <Route path="account/:code">
+                <Route index element={Acc} />
+                <Route path="send" element={AccSend} />
+                <Route path="receive" element={AccReceive} />
+                <Route path="info" element={AccInfo} />
+            </Route>
+            <Route path="add-account" element={<AddAccount />} />
```
BrianCraig, line 106: routes like `/add-account` could be `account/add`, same for `account/summary` (or just `account`)

frontends/web/test/components/forms/input.test.tsx
```diff
@@ -0,0 +1,45 @@
+/**
```
BrianCraig, line 1: re-added the input test

frontends/web/src/utils/route.tsx
```diff
@@ -0,0 +1,37 @@
+/**
```
BrianCraig, line 1: a component for making accesible `route` (marked as deprecated because i don't want people using it, being a transitional solution)

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -85,7 +85,7 @@ class Moonpay extends Component<Props, State> {
 
     public render() {
         const { moonpay, t } = this.props;
-        const { height } = this.state;
+        const { height } = this.state || { height: 500 }; //Check why state is null (Load HOC?)
```
BrianCraig, line 88: On first render `this.state` is null, looks strange for me, must be something about the Load HOC. Still the fix works as intended, and on first re-render the height state is fixed

thisconnect (approved):
LGTM with comments, but no blocker

frontends/web/src/routes/router.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const Device = <InjectParams>
+        <DeviceSwitch
+            key={devicesKey('device-switch')}
+            deviceID={null}
+            devices={devices} />
+    </InjectParams>
+
+    const Acc = <InjectParams>
+        <Account
+            code={'' /* dummy to satisfy TS */}
```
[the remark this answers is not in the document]
BrianCraig: Without changing the components, yes, because whe are expecting the params to be defined in the Props of this components. This would make defining params like {code?: string}, but we know that the params are always injected, so it would be bad.

frontends/web/src/routes/buy/moonpay.tsx
BrianCraig: This should be fixed with setting `state = {}` on the body, lets add it later

frontends/web/src/utils/equal.js
```diff
@@ -52,5 +52,5 @@ export function equal(a, b) {
         return true;
     }
 
-    return a !== a && b !== b;
```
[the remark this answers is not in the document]
BrianCraig: Eslint was not happy with that line, i also didn't think about NaN, maybe we could use `Object.is` that is like a real strict equal

frontends/web/src/components/icon/icon.tsx
```diff
@@ -66,14 +65,14 @@ export const ExpandClose = (): JSX.Element => (
...
-    <svg className={`checked ${className || ''}`} viewBox="0 0 52 52" draggable={false} {...props}>
+    <svg className={`checked ${className || ''}`} viewBox="0 0 52 52" {...props}>
         <circle className="checked-circle" cx="26" cy="26" r="25" fill="none"/>
         <path className="checked-check" fill="none" d="M14.1 27.2l7.1 7.2 16.7-16.8"/>
     </svg>
 );
 
 export const CaretDown = ({className, ...props}: SVGProps) => (
-    <svg className={`caretDown ${className || ''}`} viewBox="0 0 1024 1024" draggable={false} {...props}>
+    <svg className={`caretDown ${className || ''}`} viewBox="0 0 1024 1024" {...props}>
```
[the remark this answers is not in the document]
BrianCraig: not an JSX prop for svg

PR #1503, opened 2021-12-07
Migrate ref for bb01

thisconnect:


BrianCraig:
LGTM, still the problem may be with the `Input` merge on the branch (now being a ForwardRef component with two ref inputs). Still this PR removes code and gets the same func so cool.

Can we start using a lib like `Formik` for forms in new created components?

thisconnect:
> Can we start using a lib like Formik for forms in new created components?

Ideally we keep the amount of deps low, lets discuss offline

PR #1505, opened 2021-12-08
frontend: initialize class components state with `{}`

BrianCraig:
why: on preact class components initialize with `{}`
but react initializes with `null`, so we must assing
the state to `{}` on the construction.

### DIFF
--- a/frontends/web/src/decorators/load.tsx
+++ b/frontends/web/src/decorators/load.tsx
@@ -41,6 +41,8 @@ export function load<LoadedProps extends ObjectButNotFunction, ProvidedProps ext
         WrappedComponent: ComponentType<LoadedProps & ProvidedProps>,
     ) {
         return class Load extends Component<ProvidedProps & Partial<LoadedProps>, LoadedProps> {
+            public readonly state = {} as LoadedProps;
+
             public static displayName = `Load(${getDisplayName(WrappedComponent as any)})`;
 
             private determineEndpoints(): EndpointsObject<LoadedProps> {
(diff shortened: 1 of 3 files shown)


frontends/web/src/decorators/load.tsx
```diff
@@ -100,7 +102,7 @@ export function load<LoadedProps extends ObjectButNotFunction, ProvidedProps ext
             private allEndpointsLoaded(): boolean {
                 if (this.endpoints === undefined) { return false; }
                 for (const key of Object.keys(this.endpoints) as KeysOf<LoadedProps>) {
-                    if ((this.state === null) || (this.state[key] === undefined)) {
+                    if (this.state[key] === undefined) {
```
thisconnect, line 105: nice so it is back to the original https://github.com/digitalbitbox/bitbox-wallet-app/blob/d6b33a6d25e394e11bcb8055db41ff77129769ca/frontends/web/src/decorators/load.tsx#L103

👍

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -85,7 +87,7 @@ class Moonpay extends Component<Props, State> {
 
     public render() {
         const { moonpay, t } = this.props;
-        const { height } = this.state || { height: 500 }; //Check why state is null (Load HOC?)
+        const { height } = this.state; //Check why state is null (Load HOC?)
```
thisconnect, line 90: I think the comment is not needed anymore, correct?
please remove

thisconnect (changes requested):
LGTM, please remove the comment if it is not needed anymore

frontends/web/src/routes/buy/moonpay.tsx
BrianCraig: my bad

thisconnect:
LGTM, thanks

PR #1506, opened 2021-12-08
frontend: lock npm dependencies

thisconnect:
Save exact production dependencies as described in:
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/README.md#npm-dependencies

Dev dependencies are fine to use semver range, keeping as is.


thisconnect:
cc @BrianCraig

BrianCraig:
Good catch! LGTM

PR #1509, opened 2021-12-08
frontend: use link component for send button

thisconnect:
With the new routing normal anchors cause a page reload which
messes up the routing.

Changed to use  compoent for send button.


BrianCraig (approved):
LGTM, yes it also increases speed AFAIK

PR #1510, opened 2021-12-09
frontend: add indentation eslint rule

thisconnect:
Some code has inconsistent indentation, sometimes  1 space,
currently the project uses 4 spaces.

See also https://eslint.org/docs/rules/indent

This commit is not complete and will not lint yet.

### DIFF
--- a/frontends/web/.eslintrc.json
+++ b/frontends/web/.eslintrc.json
@@ -3,7 +3,10 @@
   "rules": {
     "jsx-a11y/anchor-is-valid": 0,
     "jsx-a11y/alt-text" : 0,
-    "quotes": ["error", "single"]
+    "quotes": ["error", "single"],
+    "indent": ["error", 4, {
+      "SwitchCase": 1
+    }]
   },
   "overrides": [
     {
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -17,179 +17,178 @@
 
 import { Component } from 'react';
 import { AppRouter } from './routes/router'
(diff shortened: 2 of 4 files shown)


thisconnect:
@BrianCraig @benma @x1ddos WDYT?

Should we do it now?

Actually I would prefer to change to 2 spaces, any opinion?

I know @benma (strongly?) prefers Crockford switch/case (no indentation), I slightly find it more readable to indent switch case, but can live with no indentation for case. Opinion?

BrianCraig:
> @BrianCraig @benma @x1ddos WDYT?
> 
> Should we do it now?
> 
> Actually I would prefer to change to 2 spaces, any opinion?
> 
> I know @benma (strongly?) prefers Crockford switch/case (no indentation), I slightly find it more readable to indent switch case, but can live with no indentation for case. Opinion?

I also prefer 2 spaces over 4, and i don't have any preference over switch cases,

x1ddos:
i've always seen switch case indented in javascript and i'm used to that.
don't really care about 4 vs 2 spaces but jsx code never fit on my screen due to long lines so maybe 2 spaces would actually make it better :+1:

thisconnect:
I checked with @benma he is ok with indented switch case and 2 spaces.

thisconnect:
since there are no open PR to `staging-react` it might be a good time to change that now...

another one would be to force semicolon.

@BrianCraig wdyt, should we do it now?

thisconnect:
closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1591

PR #1514, opened 2021-12-10
frontend: wait-dialog should allow any children

thisconnect:
The wait-dialog component has an interesting logic that checks if
there are no children in which case it displays a graphic that
shows how to confirm on a BitBox01.

The component should also allow just strings as children and then
display that without the BitBox01 graphic. Changed to just check
for children and removed the is valid react component filter.

Tested:
- that graphic shows on BB01 facroty reset
- that graphic doesn't show on go to startup settings or factory reset


BrianCraig (approved):
LGTM

PR #1515, opened 2021-12-10
frontend: rewrite deviceswitch as functional component

thisconnect:
It was a class with only render, better to just use function
component if there is no state.


frontends/web/src/routes/device/deviceswitch.tsx
```diff
@@ -27,22 +27,20 @@ interface Props {
...
 }
 
-class DeviceSwitch extends Component<Props, {}> {
-    public render() {
-        const { deviceID, devices } = this.props;
-        if (this.props.default || deviceID === null || !Object.keys(devices).includes(deviceID)) {
-            return <Waiting />;
-        }
-        switch (devices[deviceID]) {
+const DeviceSwitch: FunctionComponent<Props> = (props) => {
```
BrianCraig, line 30:
```suggestion
const DeviceSwitch: FunctionComponent<Props> = ({ deviceID, devices }) => {
```

BrianCraig (approved):
LGTM

frontends/web/src/routes/device/deviceswitch.tsx
BrianCraig: What do you think about `props` destructuring on arguments?

thisconnect: Yes I prefer, but one prop name is `default` which is a JS reserved word, let me try again

thisconnect: Confirmed it can't destruct a prop with the name `default` I got:

```
Failed to compile.

./src/routes/device/deviceswitch.tsx
SyntaxError: frontends/web/src/routes/device/deviceswitch.tsx: Unexpected token (31:8)

  29 |
  30 | const DeviceSwitch: FunctionComponent<Props> = ({ default, deviceID, devices }) => {
> 31 |     if (default || deviceID === null || !Object.keys(devices).includes(deviceID)) {
     |         ^
  32 |         return <Waiting />;
  33 |     }
  34 |     switch (devices[deviceID]) {

```

thisconnect: and eslint doesn't understand it at all

```
frontends/web/src/routes/device/deviceswitch.tsx
  30:57  error  Parsing error: ':' expected
```

thisconnect: `default` came from here https://github.com/preactjs/preact-router#handling-urls

but since we stop using `preact-router`, this prop could be renamed.

BrianCraig: Yes, maybe you could just see the way to stop using `default` prop, it is a way to say `i'm not rendering a device`, but that could be just known by handling the `deviceID === null` case

thisconnect: I see, yes `deviceID === null` seems sufficient, I'll update the commit.

PR #1516, opened 2021-12-10
Frontend refactor version component

thisconnect:
- refactored version component
- fixed warning about state update on unmounted component

### DIFF
--- a/frontends/web/src/api/version.ts
+++ b/frontends/web/src/api/version.ts
@@ -1,5 +1,4 @@
 /**
- * Copyright 2018 Shift Devices AG
  * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,22 +14,8 @@
  * limitations under the License.
  */
 
-import { FunctionComponent, useEffect, useState } from 'react';
-import { useTranslation } from 'react-i18next';
-import { getVersion } from '../../api/version';
+import { apiGet } from '../utils/request';
 
-const Version: FunctionComponent = () => {
-    const { t } = useTranslation();
-    const [version, setVersion] = useState<string>('');
-
-    useEffect(() => {
-        getVersion().then(setVersion);
-    }, []);
-
-    if (!version) {
-        return null;
-    }
-    return <p>{t('footer.appVersion')} {version}</p>;
-}
-
-export { Version };
+export const getVersion = (): Promise<string> => {
+    return apiGet('version');
+};


frontends/web/src/api/version.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { apiGet } from '../utils/request';
+
+export const getVersion = (): Promise<string> => {
```
BrianCraig, line 19: :+1:

frontends/web/src/components/layout/version.tsx
```diff
@@ -14,19 +15,32 @@
...
 
-interface VersionProps {
-    version: string;
-}
+const Version: FunctionComponent = () => {
+    const { t } = useTranslation();
+    const [version, setVersion] = useState<string>('');
 
-type Props = VersionProps & TranslateProps;
+    useEffect(() => {
```
BrianCraig, line 26: This effect could be simplified to:

```
    useEffect(() => {
        getVersion().then(setVersion);
    }, []);
```

frontends/web/src/components/layout/version.tsx
```diff
@@ -14,19 +15,32 @@
...
+            }
+        }
+        updateVersion();
+        return () => {
+            ignore = true
+        };
+    });
 
-function Version({ t, version }: PropsWithChildren<Props>) {
+    if (!version) {
```
BrianCraig, line 40: If we define a default loading string like `...` for version this check would not be needed, do you think it's better?

BrianCraig (changes requested):

frontends/web/src/components/layout/version.tsx
BrianCraig: In Fact, the API responses are so fast that even with an empty string would be unnoticeable

frontends/web/src/components/layout/version.tsx
thisconnect: I could only get rid of the warning with the 2nd commit that contains the `ignore` variable and does not update the state if the component is already unmounted...

Should I include the empty array?

frontends/web/src/components/layout/version.tsx
thisconnect: yeah true or just a "loading" message 🤷‍♂️  , in the future we could have the gray loading indicator, but this is a bit out of scope for the react-migration.

I believe the version would load faster, but if I remember correctly it is still inside 2 components that have the load decorator I believe that is why it is taking so long to see the version (about .5s)

The missing version text doesn't make the page layout shift, it just appears once the version is loaded, so I would keep it as is for now.

frontends/web/src/components/layout/version.tsx
BrianCraig: Yes, that means that the dependencies of that `useEffect` are none, since `getVersion` it's a constant and `setVersion` is a setter (constant) eslint rules should not warn you anything

frontends/web/src/components/layout/version.tsx
BrianCraig: As far as rendering a null doesn't break the Layout (not this case), im ok, this is just a suggestion :+1:

frontends/web/src/components/layout/version.tsx
thisconnect: As discussed I'll remove the 2nd commit that fixes the warning and change this PR without `ignore` variable and cleanup function.

PR #1518, opened 2021-12-13
Frontend: remove `selectedOption` from `Select`

BrianCraig:
Unnescesary and duplicated prop, removed in favour of `defaultValue`


thisconnect (approved):
LGTM thanks

Issue #1495, opened 2021-11-27
Improve download CSV file


thisconnect:
Got the following report on TG

![Screen Shot 2021-11-27 at 8 41 04 AM](https://user-images.githubusercontent.com/546900/143672882-cddf91a9-6be3-4e67-b04b-f1e90298206a.png)

This probably shouldn't happen, further ideas to improve the UX:
- find out and fix why it shows "Please connect device", this is a bug, it should not route
- add a notification where the file has been downloaded to
- animate the download button into an open file button
- handle if Downloads directory doesn't exist
- ideally: use a native "file save as" dialog

BrianCraig:
The "Please connect device" is because after the CSV is generated, a link element with a local file path href is set, this should not work because it's not part of a specification, and usually blocked by the client (accessing local user files), but somehows it works on linux qt client, opening the file.

I'll address this in the following form:

Add an alert after downloading that says:
"Your summary is ready"
"You can access it on **[path]** from your files explorer"

This will be triggered once the CSV is generated.

I think the idea of making a "file save as" dialog adds a lot of complexity when there are 4 types of clients.

Is this ok for you?

thisconnect:
@BrianCraig thanks for test, so you cannot reproduce the issue on linux in linux qt?

> but somehows it works on linux qt client, opening the file

Yes, iirc there was some custom logic in the Go backend that catches this link and tries to open the file locally in the default application for this filetype.

In this case my suspicion is that something in macOS did change, I guess permission to open a file in Downloads directory.

@jadzeidan can you test this issue on Windows?

thisconnect:
was fixed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1622

PR #1522, opened 2021-12-13
frontend: fix open exported csv file

thisconnect:
The exported CSV file did not open anymore but the view changed
to show the wait screen.

This is a regression in Anchor.tsx introduced in:
https://github.com/digitalbitbox/bitbox-wallet-app/commit/c595e1d8a741bb75af190e54d0ae692854c91fe2

Fixed by bypassing Anchor component logic and directly call
apiPost('open', csvfile). But kept the anchor component that the
layout does not break.

But the layout of the open button on the account summary view
looked broken, so I added minimal styles that font and color
matches the rest.

Anchor.tsx should be refactored to systemOpen component, but in
a different change.


frontends/web/src/components/transactions/transactions.tsx
```diff
@@ -52,9 +53,11 @@ class Transactions extends Component<Props> {
                     <label className="labelXLarge">{t('accountSummary.transactionHistory')}</label>
                     { !csvExportDisabled && (
                         exported ? (
-                            <A href={exported} className="labelXLarge labelLink">{t('account.openFile')}</A>
+                            <A key="open" href="#" onClick={() => apiPost('open', exported)} className="labelXLarge labelLink">
```
benma, line 56: what's the purpose of the new `key` attributes? Was there a bug?

In any case, `A` is probably not the right tag anymore for this, right? Why use an Anchor when the logic of the Anchor component is not needed.  Could use `<a>` or `<span>` or sth. It is also confusing because Anchor.tsx already sets onClick by itself, and now also via the passed props. Maybe something a linter should catch or will catch in the future.

benma (changes requested):

benma:
> Anchor.tsx should be refactored to systemOpen component, but in a different change

That is not needed anymore, right? The name is okay since we will keep using it for internal (no system open) and external urls.

frontends/web/src/components/transactions/transactions.tsx
thisconnect: No reason, just thougth it is best practice want me to remove?

kept using `<A>` and thuoght refactor in a different PR, see reason in commit msg:

> But kept the anchor component that the layout does not break.

thisconnect:
>  internal (no system open) and external urls.

and currently internal urls/views  see `/excahnges`

frontends/web/src/components/transactions/transactions.tsx
benma: is it best practice? i thought it is only needed in lists so that rendering doesn't break, but maybe I missed sth.

benma (approved):
tACK

benma:
> > internal (no system open) and external urls.
> 
> and currently internal urls/views see `/excahnges`

in your quote of me it says 'internal'. just to clarify, we are talking about the same thing right? :smile:

frontends/web/src/components/transactions/transactions.tsx
thisconnect: It doesn't seem to be a problem here, however it can conditionally render the same elements (`<a>`). Should I remove it because it works without or keep as I think it might be clearer with the key prop?

thisconnect:
> in your quote of me it says 'internal'. just to clarify, we are talking about the same thing right? 😄

sorry yesterday was too late  🙈

thisconnect:
added a changelog entry https://github.com/digitalbitbox/bitbox-wallet-app/compare/034312eea63228feed471810d062c10258a5ea41..01d5a822c8932910b65913407ac375ebe4bcf8f9

x1ddos:
did this PR address https://github.com/digitalbitbox/bitbox-wallet-app/issues/1495 somehow? remember to link to the issues.

PR #1523, opened 2021-12-14
frontend: sdcardcheck component should not render before check

thisconnect:
Currently the sdcardcheck component quickly flickers a dialog,
before the response of the check endpoint resolved.

Changed to not render at all until there is a response, once there
is a response if it is false, the dialog should show asking to
insert an sdcard and if true it should render its children
(the backup list fetched from the sdcard).


thisconnect:
@BrianCraig PTAL

BrianCraig:
LGTM, nice catch!

PR #1524, opened 2021-12-14
Frontend: update unsafe class methods

BrianCraig:
Due to the react migration, some Class Components methods
are beingg deprecated around 18.x, see
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

This commit will ignore warnings, but components need
to be migrated if an react 18.x update is done


thisconnect (approved):
LGTM

PR #1525, opened 2021-12-14
frontend: fix settingsbutton linking

thisconnect:
In the Qt app we shouldn't use normal anchors with href attributes
to link to different views. Reason this works in webdev, is that
react-scripts always serves the index.html with main JavaScript,
so after navigation and a full page reload occured the script can
pick up and continue. In the Qt app the index.html isn't served
and therefore something like a 404 happens.

Fixed by changing all remaining href on settings buttons to onClick
using the deprecated route method.

Refactored the settings component to a function components.


BrianCraig:
LGTM :+1:

thisconnect:
@BrianCraig sorry I added 1 more commit PTAL.

BrianCraig:
Nice! LGTM

PR #1528, opened 2021-12-15
Frontend update react scripts

thisconnect:


thisconnect:
@BrianCraig just tried if we could use latest react-scripts, it is more an experiment, could you test if that works for you?

You might have to remove all the `*.css.d.ts` files... the scripts work for me but TypeScript in VS Code complains everywhere always related to the react import.

BrianCraig:
@thisconnect LGTM, update the snapshots https://github.com/thisconnect/bitbox-wallet-app/pull/13

thisconnect:
> update the snapshots thisconnect#13

thanks 🙏

PR #1544, opened 2021-12-17
frontend: add useLoad and useSubscribe hooks

### LINKED ISSUE #1536: Frontend: Type api calls to src/api (remove load HOC)
We should move away from load HOC and move api calls to src/api (using typed responses)

Also rewrite components that use load HOC to function component if possible and use hooks.

See example https://github.com/digitalbitbox/bitbox-wallet-app/pull/1531/commits/b99aa2119b568eb772c1ed2d1647196fd255068d

BrianCraig:
- added an example of `useAsync` and `useSubscription`
- removed unnescesary hiding logic on `HeadersSync`, due to logic of being displayed already hides component on load.

Partially fixes #1536

@thisconnect What do you think about this?

### DIFF
--- a/frontends/web/src/api/coins.ts
+++ b/frontends/web/src/api/coins.ts
@@ -1,5 +1,4 @@
 /**
- * Copyright 2018 Shift Devices AG
  * Copyright 2021 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,19 +14,18 @@
  * limitations under the License.
  */
 
-import { FunctionComponent } from 'react';
-import { useTranslation } from 'react-i18next';
-import { useLoad } from '../../hooks/api';
-import { getVersion } from '../../api/version';
+import { subscribeEndpoint, SubscriptionCallback } from './subscribe';
+import { CoinCode } from './account';
 
-const Version: FunctionComponent = () => {
-    const { t } = useTranslation();
-    const version = useLoad(getVersion);
-
-    if (!version) {
-        return null;
-    }
-    return <p>{t('footer.appVersion')} {version}</p>;
+interface Status {
+    targetHeight: number;
+    tip: number;
+    tipAtInitTime: number;
+    tipHashHex: string;
 }
 
-export { Version };
+export const subscribeCoinHeaders = (coinCode: CoinCode) => (
+    (cb: SubscriptionCallback<Status>) => (
+        subscribeEndpoint(`coins/${coinCode}/headers/status`, cb)
+    )
+);
--- a/frontends/web/src/api/subscribe.ts
+++ b/frontends/web/src/api/subscribe.ts
@@ -17,16 +17,17 @@
 import { apiSubscribe, Event, Unsubscribe } from '../utils/event';
 import { apiGet } from '../utils/request';
 
+export type SubscriptionCallback<T> = (eventObject: T) => void
+
 /**
  * Subscribes the given function on an endpoint on which the backend
  * can push data through. This should be mostly used within api.
  * Note there is a subscibe-legacy.ts module that supports older events.
  */
 
-export function subscribeEndpoint(
+export function subscribeEndpoint<T>(
     endpoint: string,
-    // TODO: double check any type
-    cb: (eventObject: any) => any,
+    cb: SubscriptionCallback<T>,
 ): Unsubscribe {
     return apiSubscribe(endpoint, (event: Event) => {
         switch (event.action) {
--- a/frontends/web/src/components/headerssync/headerssync.tsx
+++ b/frontends/web/src/components/headerssync/headerssync.tsx
@@ -1,4 +1,3 @@
-
 /**
  * Copyright 2018  Shift Devices AG
  * Copyright 2021 Shift Crypto AG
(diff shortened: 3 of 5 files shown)


benma:
Please extend the commit message with more information. why `useAsync()`, why `mounted = false/true`, etc? good commit msgs helps everyone in the future who looks at the code to undersand why it is there.

BrianCraig:
> Please extend the commit message with more information. why `useAsync()`, why `mounted = false/true`, etc? good commit msgs helps everyone in the future who looks at the code to undersand why it is there.

Yes, also i wanted to ask. 
The reason the `mounted` local variable is to not set a state when the component is unmounted.
Screenshot from 2021-12-17 09-43-38
If you don't this message appears, still using it or not the behaviour is the same. The only thing that changes is that this warning is shown. Do you think this is needed?

I'll change it to a useRef for consistency instead of a variable

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -16,89 +16,48 @@
...
+    if (!status) {
+        return null;
     }
 
-    componentDidUpdate(prevProps) {
-        const { status } = this.props;
-        if (status && prevProps.status && status.tip !== prevProps.status.tip) {
-            this.setState({ show: status.tip });
-            if (status.tip === status.targetHeight) {
-                // hide component after 4s when tip reached targetHeight
```
benma, line 57: the commit msg does not explain why `show` is not needed. what about this comment? did this functionality change?

benma (commented):
i still couldn't get around to testing it, but i left some questions that will help review later.

please always explain *why* something changed (reasoning) in the commit message, even more so than what changed or how it changed. it helps me immensely in review so i don't have to guess.

frontends/web/src/hooks/api.ts
```diff
@@ -0,0 +1,67 @@
...
+import { SubscriptionCallback } from '../api/subscribe';
+import { Unsubscribe } from '../utils/event';
+
+/**
+ * useSubscribe is a hook to subscribe to a subscription function.
+ * starts on first render, and returns undefined while there is no first response.
+ * re-renders on every update.
+ */
+
+export const useSubscribe = <T>(subscription: ((callback: SubscriptionCallback<T>) => Unsubscribe)) => {
```
benma, line 27: the return type of this function is missing

frontends/web/src/components/headerssync/headerssync.tsx
BrianCraig: I added an eplanation about why i've removed it:

- removed unnescesary hiding logic on `HeadersSync` `HeadersSync` was returning `null` after 4 seconds the headers been sync, this was a display implementation based on its usage, meanwhile the only usage it's on `account.tsx`, where the component is removed from render just after the headers been sync, 
being unnescesary the hidding of the component, alongside it's a feature that should be in `account.tsx` if wanted, not in the component

thisconnect: Somehow it didn't remove the comp after synced, not sure why.

Screen Shot 2021-12-21 at 5 23 30 PM

thisconnect (changes requested):
Nice hooks 👍

frontends/web/src/components/headerssync/headerssync.tsx
BrianCraig: I've readded the hidding logic

frontends/web/src/hooks/api.ts
```diff
@@ -0,0 +1,71 @@
...
+            }
+        })
+
+        return () => {
+            mounted.current = false;
+            unsubscribe();
+        };
+    // we pass no dependencies because it's only subscribed once
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [])
```
thisconnect, line 45: please add semicolon

thisconnect (approved):
LGTM very nice!!! 
some small nits see comments

Also please update the commit message and change useAsync to useLoad

thisconnect:
@benma if you have time could you look at this PR?
so we could continue migrating more of the api and change load hoc to useLoad hook.

Would it make sense to proceed on a staging branch, i.e. `staging-hooks` ?

frontends/web/src/api/subscribe.ts
```diff
@@ -17,16 +17,18 @@
...
 /**
  * Subscribes the given function on an endpoint on which the backend
  * can push data through. This should be mostly used within api.
  * Note there is a subscibe-legacy.ts module that supports older events.
  */
 
-export function subscribeEndpoint(
+export function subscribeEndpoint<T>(
     endpoint: string,
     // TODO: double check any type
```
benma, line 30: out of date comment?

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -16,89 +16,64 @@
...
-                    </div>
-                    { !loaded ? (<Spinner />) : null }
-                </div>
-                <div className={style.progressBar}>
-                    <div className={style.progressValue} style={{ width: `${value}%` }}></div>
+    return (
+        <div className={style.syncContainer}>
+            <div className={style.syncMessage}>
+                <div className={style.syncText}>
+                    {t('headerssync.blocksSynced', { blocks: formatted })}
```
benma, line 69: space missing:

image

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -16,89 +16,64 @@
...
 
-interface IStatus {
-    targetHeight: number;
-    tip: number;
-    tipAtInitTime: number;
-    tipHashHex: string;
+const useMountedRef = () => {
+    const mounted = useRef(true);
+    useEffect(() => (
+        () => {mounted.current = false}
```
benma, line 34: that does not do anything? or what am i missing?

benma (changes requested):

frontends/web/src/components/headerssync/headerssync.tsx
benma: ah this returns the cleanup callback. mistook ( for { above.

better to init to `false` and set to `true` here? `mounted.current` sounds like it should be true while mounted.

frontends/web/src/api/coins.ts
```diff
@@ -0,0 +1,31 @@
...
+interface Status {
+    targetHeight: number;
+    tip: number;
+    tipAtInitTime: number;
+    tipHashHex: string;
+}
+
+
+export const subscribeCoinHeaders = (coinCode: string) => (
+    (cb: SubscriptionCallback<Status>) => (
```
benma, line 28: why do you put the type here instead of above in `subscribeCoinHeaders()`? would be nice to see the whole function type at a glance without looking into the body of the function.

frontends/web/src/api/coins.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface Status {
+    targetHeight: number;
+    tip: number;
+    tipAtInitTime: number;
+    tipHashHex: string;
+}
+
+
+export const subscribeCoinHeaders = (coinCode: string) => (
```
benma, line 27: coinCode can be of type `CoinCode` from account.ts, right?

frontends/web/src/hooks/api.ts
```diff
@@ -0,0 +1,70 @@
...
+    const mounted = useRef(true);
+    useEffect(() => {
+        apiCall().then((data) => {
+            if (mounted.current) {
+                setResponse(data);
+            }
+        });
+
+        return () => {mounted.current = false};
+    // we pass no dependencies because it's only queried once
```
benma, line 66: the comment is best movd to after the comma before `[]`. pretty sure my editor would automatically indent these lines to be on the same level as the rest of the block otherwise.

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -16,89 +16,64 @@
...
+interface Props {
     coinCode: CoinCode;
 }
 
-interface IStatus {
-    targetHeight: number;
-    tip: number;
-    tipAtInitTime: number;
-    tipHashHex: string;
+const useMountedRef = () => {
```
benma, line 31: move this to a place where it can be reused? you use the same trick three more times in this PR, all of them could use `useMountedRef()` instead, right?

benma:
concept ACK but please see the comments

frontends/web/src/components/headerssync/headerssync.tsx
thisconnect: I think it is alredy doing what you want, the return callback `() => {mounted.current = false}` is called on unmount https://reactjs.org/docs/hooks-effect.html

frontends/web/src/api/coins.ts
thisconnect: it is part of `subscribeCoinHeaders` just on a new line, I requested to put new lines, because I have a hard time reading/understanding when there are multiple arrow functions on the same line.

frontends/web/src/api/coins.ts
BrianCraig: yes :+1:

frontends/web/src/components/headerssync/headerssync.tsx
BrianCraig: I moved it to `hooks/utils.ts` :+1:

frontends/web/src/api/subscribe.ts
```diff
@@ -17,16 +17,17 @@
 import { apiSubscribe, Event, Unsubscribe } from '../utils/event';
 import { apiGet } from '../utils/request';
 
+export type SubscriptionCallback<T> = (eventObject: T) => void
+
 /**
  * Subscribes the given function on an endpoint on which the backend
  * can push data through. This should be mostly used within api.
- * Note there is a subscibe-legacy.ts module that supports older events.
```
thisconnect, line 23: please keep this comment, if I see correctly @benma was referring to the other comment `// TODO: double check any type`

frontends/web/src/hooks/api.ts
```diff
@@ -0,0 +1,66 @@
...
+): (T | undefined) => {
+    const [respose, setResponse] = useState<T>();
+    const mounted = useMountedRef();
+    useEffect(
+        () => {
+            apiCall().then((data) => {
+                if (mounted.current) {
+                    setResponse(data);
+                }
+            });
```
thisconnect, line 62: Question: doesn't it need `return unsubscribe;` ?

BrianCraig: No, because it's not a subscription, it's a request only

thisconnect: ah sorry, ofc, you are right :)

frontends/web/src/components/headerssync/headerssync.tsx
BrianCraig: resolved

thisconnect:
tested 0333181
LGTM

thisconnect (approved):

thisconnect:
merged into `staging-hooks` branch so it doesn't block migrating from HOC to hooks

Issue #1546, opened 2021-12-20
Frontend: unlocked device starts in `manage device` instead of `portfolio`


BrianCraig:
https://github.com/digitalbitbox/bitbox-wallet-app/issues/1499#issuecomment-997086117

PR #1548, opened 2021-12-20
Frontend route unlocked

thisconnect:
- reverted commit b239b0f
- added setStateWithDeviceList
- check if new device has accounts, then do maybeRoute instead of route to device settings

closes https://github.com/digitalbitbox/bitbox-wallet-app/issues/1546

Tested:

(just a screenshot of my doc)

Screen Shot 2021-12-20 at 11 59 30 AM

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -79,25 +79,10 @@ import { route, RouterWatcher } from './utils/route';
                  break;
              }
          });
- 
-         const setDevices = (devices: TDevices) => {
-             const oldDeviceIDList = Object.keys(this.state.devices);
-             this.setState({ devices }, () => {
-                 const newDeviceIDList: string[] = Object.keys(this.state.devices);
-                 if (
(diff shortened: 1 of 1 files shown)


thisconnect:
there are still single commit, let me know and I squash everything into 1 commit.

BrianCraig:
LGTM :clap:

benma:
In webdev mode, if the device is inserted and unlocked, and I then start servewallet and then webdev (in that order) and then go to http://localhost:8080/, I still end up in `Manage device`.

This indicates that there is still some edge case or race that is not correct, and it might happen in Qt/Android too.

thisconnect:
@benma i cannot reproduce, in webdev no matter in which order I end up on the account summary, except if it already is on `http://localhost:8080/device/...`.

> In webdev mode, if the device is inserted and unlocked, and I then start servewallet and then webdev (in that order) and then go to http://localhost:8080/, I still end up in Manage device.

benma (approved):

PR #1554, opened 2021-12-22
Frontend: config api, context and usage

BrianCraig:
- Created typed `get`/`set` for `config` api
- Created Context for storing the config, this is useful because you can have an unique provider of the config, and re-renders components listening to the context through the `Context` or `useConfig` hook. Using a unique provider is better because, if every component manages it's own config state, they might push outdated configs. using `setConfig` with a reducer, will update it with the latest config. 

example: 
```
setConfig((config) => ({
    ...config,
    frontend: {
        ...config.frontend,
        [dismissable!]: true,
    },
}));
```
this reducer function will trigger an update in every context listener, and also update the information with the latest config

- Migrated `Status` to Functional Component

Tested locally


frontends/web/src/contexts/ConfigContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }, [config]);
+    
+    return <ConfigContext.Provider value={{
+        config,
+        setConfig: (setConfig as Dispatch<SetStateAction<Config>>), // avoid undefined
+        configLoaded: !!config
+    }}
+    children={children}
+    />
+}
```
thisconnect, line 60: Please use lowercase filenames maybe  `frontends/web/src/contexts/config.tsx`?

frontends/web/src/hooks/app.ts
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useContext } from 'react';
+import { ConfigContext, ConfigContextInterface } from '../contexts/ConfigContext';
+
+export const useConfig = (): ConfigContextInterface => {
+    return useContext(ConfigContext);
+}
```
thisconnect, line 22: Why not `hooks/config.ts`?
Or what is the plan with `hooks/app.ts` is the idea that more goes into app hook?

frontends/web/src/contexts/ConfigContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+    useEffect(() => {
+        getConfig().then(setConfig);
+    }, [])
+
+    // Tell the backend to store a new config when it changes
+    // Save ref is used to not post the first request
+    useEffect(() => {
+        if (config) {
+            if (save.current) {
+                postConfig(config);
```
thisconnect, line 47: If postConfig fails for some reason it would be great to have an error and roll back to the last config.

It's a bit ugly but The current code base uses alertUser component maybe good enough for now?

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,64 @@
...
+        ltc?: BtcLikeConfig
+        tltc?: BtcLikeConfig
+        eth: EthLikeConfig
+        teth: EthLikeConfig
+        reth: EthLikeConfig
+        fiatList: Fiat[]
+        mainFiat: Fiat,
+        userLanguage: string
+    }
+    frontend: {[key: string]: boolean}
```
thisconnect, line 55: interesting that we never used anything else than `{[key: string]: boolean}` for frontend conf

frontends/web/src/components/status/status.tsx
```diff
@@ -15,15 +15,10 @@
...
+import { FunctionComponent } from 'react';
+import { useConfig } from '../../hooks/app';
 import style from './status.module.css';
-
-interface State {
-    show: boolean;
-}
-
-interface StatusProps {
+interface Props {
```
thisconnect, line 21: Just `Props`, I like

thisconnect (commented):
Very nice, some question

frontends/web/src/components/status/status.tsx
```diff
@@ -33,91 +28,63 @@ interface StatusProps {
...
+}) => {
+    const { config, setConfig } = useConfig();
+    
+    let dismissed = false;
+    if (dismissable) {
+        if (!config) {
+            // Wait until config is loaded to prevent rendering a dismissed Status
+            return null;
         }
+        dismissed = !!config.frontend[dismissable]
```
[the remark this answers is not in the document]
BrianCraig: It doesn't need `!!`, still it is a way to convert a `undefined | boolean` to a `boolean` using a truthy conversion. 
I like it because the later `if (hidden || dismissed) {` is between booleans, and not more complex types.
WDYT?

frontends/web/src/contexts/ConfigContext.tsx
BrianCraig: Renamed :+1:

frontends/web/src/hooks/app.ts
BrianCraig: Yes, its better to have a separate file, i renamed it :+1:

frontends/web/src/contexts/ConfigContext.tsx
BrianCraig: I changed the code, 
Now, on `set`, it gets the new value from the server, updates the config with that value, pushes the new config, and if it fails restores it to the latest gotten from the server, or the latest known if that request failed

benma:
at a glance it seems to be that a `useConfig/setConfig` hook managed locally by every component that needs the config should work and is conceptually a lot simpler? but i may be missing sth. it seems a bit this PR is solving a problem that might not exist.

>  if every component manages it's own config state, they might push outdated configs.

the `setConfig()` function already only updates the subset of config that is needed, and does it in a way where the other config is not changed. can you elaborate on how exactly outdated configs could be pushed in this way?

thisconnect:
Currently each component the needs app config has to fetch it on mount. This is not per se a problem. This PR proposes a different architecture to take care about config globally.

The concept of this PR is that the app only fetches the config once on app launch and have a getter setter available to read update and the context takes care. Read would be slightly faster as the components do not have to fetch config over and over again.

benma:
> Currently each component the needs app config has to fetch it on mount. This is not per se a problem. This PR proposes a different architecture to take care about config globally.
> 
> The concept of this PR is that the app only fetches the config once on app launch and have a getter setter available to read update and the context takes care. Read would be slightly faster as the components do not have to fetch config over and over again.

I guess my point is: why propose a different architecture that is more complicated if it doesn't solve a problem? I am not convinced performance is a problem. I am also not convinced that outdated config can be pushed. 

I'd keep it simple and as it is now, if it works, and not over-engineer it.

Just my two sats - feel free to overrule of course.

BrianCraig:
> at a glance it seems to be that a `useConfig/setConfig` hook managed locally by every component that needs the config should work and is conceptually a lot simpler? but i may be missing sth. it seems a bit this PR is solving a problem that might not exist.
> 
> > if every component manages it's own config state, they might push outdated configs.
> 
> the `setConfig()` function already only updates the subset of config that is needed, and does it in a way where the other config is not changed. can you elaborate on how exactly outdated configs could be pushed in this way?

Adding to what @thisconnect said, the main purpose it should be that all the components who need to know about the config, can actually just use `const { config } = useConfig();` and automatically be re-rendered once any component using the same hook calls the `setConfig` method, this should be an advantage when more atomic components are used, because you don't want to have to make a GET for `config` on each component who needs to know something about the configuration.
Also helps making smaller components and less components inter-dependency.

I admit is something over-engineered for the task that currently does, but it is an example of taking complexity out of the components and moving them to a more specific component, in this case a Context Provider

benma:
> because you don't want to have to make a GET for config on each component who needs to know something about the configuration

Why don't we want that? It is conceptually simpler imo. Are there any concrete downsides?

BrianCraig:
> > because you don't want to have to make a GET for config on each component who needs to know something about the configuration
> 
> Why don't we want that? It is conceptually simpler imo. Are there any concrete downsides?

The concrete downside that i see is that is more simple to use React Apps when visual components are just a representations of actual state of the application. If you have an interaction with the backend on a component might not be a problem, but when you have multiple components that interact independently each other to the backend, with multiple endpoints, and each one maintaining an own state of the endpoint, tend to make sync problems between components. 

The idea is mostly an example of abstracting actual state from components, there are many different things that would benefit following the same principle, like state about devices, accounts, etc...

benma:
Fyi, the purpose of the original load/subscribe HOC was the same: remove the state from the component and turn it into a rendering component only using props.

I guess then it makes sense to translate it to the "new way".

thisconnect:
Purpose yes, but with the big difference that with this context the config is loaded on appstart and **not** loaded by every component that has the load HOC to put the config into its props. 

> Fyi, the purpose of the original load/subscribe HOC was the same: remove the state from the component and turn it into a rendering component only using props.

benma (approved):
concept ACK

PR #1559, opened 2021-12-27
frontend: fix bb01 seed creation

BrianCraig:
On bb01 there is a bug where `SecurityInformation` doesn't render it's children because `children` works different in `preact` than `react`. 
This is involved in the `bb01` seed creation.
This bug was introduced on the `react` migration.


thisconnect (approved):
LGTM

tested backup recover with BB01 👍

PR #1550, opened 2021-12-22
frontend: update i18next & activated tests

BrianCraig:
Updated i18n library, removing unnescesary
i18next-react config in `src/i18n/i18n.js` implementation
because we already are using `I18nextProvider`.

Re-added `i18n.test.tsx`, removing `i18n` mock for a
more integral test.

closes #1543

PR #1568, opened 2021-12-30
frontend: renamed plural to v4

thisconnect:
i18next had a breaking change introduce new v4 json format:
- https://github.com/i18next/i18next/blob/master/CHANGELOG.md#2100

So key: 'item', key_pluar: 'items' does not work anymore, new
syntax is: key_one: 'item', key_other: 'items'.

Made locize pull, renamed key and will need to be updated
with locize-push.

Part of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1550

### DIFF
--- a/frontends/web/src/locales/bg/app.json
+++ b/frontends/web/src/locales/bg/app.json
@@ -808,8 +808,8 @@
   },
   "loading": "зареждане...",
   "notification": {
-    "newTxs": "Нова транзакция в {{accountName}}",
-    "newTxs_plural": "{{count}} нови транзакции в {{accountName}}"
+    "newTxs_one": "Нова транзакция в {{accountName}}",
+    "newTxs_other": "{{count}} нови транзакции в {{accountName}}"
   },
   "pairing": {
     "aborted": {
--- a/frontends/web/src/locales/es/app.json
+++ b/frontends/web/src/locales/es/app.json
@@ -1068,8 +1068,8 @@
     "title": "Nota"
   },
   "notification": {
-    "newTxs": "Nueva transacción en: {{accountName}}",
-    "newTxs_plural": "{{count}} nuevas transacciones en: {{accountName}}"
+    "newTxs_one": "Nueva transacción en: {{accountName}}",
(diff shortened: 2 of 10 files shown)


frontends/web/src/locales/bg/app.json
```diff
@@ -808,8 +808,8 @@
   },
   "loading": "зареждане...",
   "notification": {
-    "newTxs": "Нова транзакция в {{accountName}}",
-    "newTxs_plural": "{{count}} нови транзакции в {{accountName}}"
+    "newTxs_one": "Нова транзакция в {{accountName}}",
```
benma, line 811: is `_one` required?

frontends/web/src/locales/es/app.json
```diff
@@ -1068,8 +1068,8 @@
     "title": "Nota"
   },
   "notification": {
-    "newTxs": "Nueva transacción en: {{accountName}}",
-    "newTxs_plural": "{{count}} nuevas transacciones en: {{accountName}}"
+    "newTxs_one": "Nueva transacción en: {{accountName}}",
```
benma, line 1071: tbh i don't know what happens if you locally change stuff in languages other than `en` and then do a locize push. i guess we will see... so far we have only made local additions/deletions/renames in `en` and then pushed.

benma (approved):

frontends/web/src/locales/es/app.json
thisconnect: I didn't know that, and Locize did not loose and also rename the keys in the other languages?

frontends/web/src/locales/bg/app.json
thisconnect: Yes in the new v4 plural format https://www.i18next.com/translation-function/plurals

frontends/web/src/locales/es/app.json
thisconnect: This PR also needs this PRhttps://github.com/digitalbitbox/bitbox-wallet-app/pull/1550

PR #1570, opened 2022-01-03
Android: hide download button

BrianCraig:
This hides the download button because Android users usually auto-update their app through a store.

This should be added again with a download button that opens up the store where the app has been downloaded. #1571

Screenshot_1641200313

### DIFF
--- a/frontends/web/src/components/update/update.tsx
+++ b/frontends/web/src/components/update/update.tsx
@@ -15,9 +15,10 @@
  * limitations under the License.
  */
 
-import { PropsWithChildren } from 'react';
+import { FunctionComponent } from 'react';
+import { useTranslation } from 'react-i18next';
 import { load } from '../../decorators/load';
-import { translate, TranslateProps } from '../../decorators/translate';
+import { runningInAndroid } from '../../utils/env';
 import A from '../anchor/anchor';
 import Status from '../status/status';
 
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/update/update.tsx
```diff
@@ -30,13 +31,14 @@ interface File {
...
 
-interface LoadedProps {
+interface Props {
     file: File | null;
 }
 
-type Props = LoadedProps & TranslateProps;
-
-function Update({ file, t }: PropsWithChildren<Props>): JSX.Element | null {
+const Update: FunctionComponent<Props> = ({ file }) => {
```
benma, line 38: what's the advantage of an arrow function over `function Update(...)`?

benma (commented):
lgtm, but i still need to test it

benma:
@BrianCraig please also keep refactorings/moves/etc. in a separate commit from functional changes in the future. in general it is a lot easier to review refactors and moves on its own, ensuring only that it does the same thing as before.

benma (approved):

frontends/web/src/components/update/update.tsx
BrianCraig: here is a summary of differences: https://dmitripavlutin.com/differences-between-arrow-and-regular-functions/, but in a more simple answer, regular functions carry historical features that were introduced for `object prototyping`, arrow functions are more limited functional approach.

PR #1572, opened 2022-01-03
frontend: fix send transaction confirmation amount and unit spacing

BrianCraig:
add space between coin amount and coin unit, this was a regression added on the react migration.

before: 
Image Pasted at 2021-12-30 16-00

after:
Screenshot from 2022-01-03 09-27-33

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -767,6 +767,7 @@ class Send extends Component<Props, State> {
                                     <p>
                                         <span key="proposedAmount">
                                             {(proposedAmount && proposedAmount.amount) || 'N/A'}
+                                            {' '}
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -767,6 +767,7 @@ class Send extends Component<Props, State> {
                                     <p>
                                         <span key="proposedAmount">
                                             {(proposedAmount && proposedAmount.amount) || 'N/A'}
+                                            {' '}
```
benma, line 770: i assume this regression happened in https://github.com/digitalbitbox/bitbox-wallet-app/commit/d1bbaa2edbd62f8378046ea1024e16ee3b4b172e, when this was split into multiple lines.

do you remember splitting a single line on multiple lines in other places? maybe the same kind of bug also appears elsewhere.

benma (approved):
lgtm, thanks!

PR #1573, opened 2022-01-04
backend/btc: add P2TR script type and address generation

benma:
Generate Taproot receive addresses according to BIP-86:

https://github.com/bitcoin/bips/blob/a3a397c82384220fc871852c809f73898a4d547c/bip-0086.mediawiki


x1ddos (approved):
LGTM

PR #1576, opened 2022-01-05
qt/server: free C strings again

benma:
According to the cgo docs (https://pkg.go.dev/cmd/cgo):

```
// Go string to C string
// The C string is allocated in the C heap using malloc.
// It is the caller's responsibility to arrange for it to be
// freed, such as by calling C.free (be sure to include stdlib.h
// if C.free is needed).
func C.CString(string) *C.char
```

### DIFF
--- a/frontends/qt/server/server.go
+++ b/frontends/qt/server/server.go
@@ -124,16 +124,20 @@ func serve(
 		&nativeCommunication{
 			respond: func(queryID int, response string) {
 				cResponse := C.CString(response)
+				defer C.free(unsafe.Pointer(cResponse))
 				C.respond(responseFn, C.int(queryID), cResponse)
-				C.free(unsafe.Pointer(cResponse))
 			},
 			pushNotify: func(msg string) {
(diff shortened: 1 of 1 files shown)


frontends/qt/server/server.go
```diff
@@ -128,12 +128,16 @@ func serve(
 				C.free(unsafe.Pointer(cResponse))
 			},
 			pushNotify: func(msg string) {
-				C.pushNotify(pushNotificationsFn, C.CString(msg))
+				cMsg := C.CString(msg)
+				C.pushNotify(pushNotificationsFn, cMsg)
```
x1ddos, line 132: just like with a mutex lock/unlock, i think it's "safer" to defer C.free:

```go
cMsg := C.CString(msg)
defer c.free(unsafe.Pointer(cMsg))
C.pushNotify(....)
```

i mean, everywhere. but this also works i guess.

x1ddos (approved):
LGTM but consider using defer C.free

PR #1577, opened 2022-01-05
Fix race conditions shown by `go test -race`

benma:
More might be lingering, as the race condition checks are based on the code running, and how realistic and exhaustive unit tests are.


x1ddos (approved):
LGTM

x1ddos (approved):
LGTM

PR #1581, opened 2022-01-10
CHANGELOG: fix entries in wrong release

benma:
And add the migration to react.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Changelog
 
 ## [Unreleased]
+- Bundle BitBox02 firmware version v9.9.0
 - Support sending to Bitcoin taproot addresses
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 - Support sending to Bitcoin taproot addresses
+- Fix opening 'transactions export' CSV file
+- Added Dutch translation
```
x1ddos, line 6: "Add", imperative? reads nicer when everything is in the same mood. then also, Migrated -> Migrate

x1ddos (approved):
LGTM but i'm thinking now why don't we specify which firmware version is bundled with the app release? might be handy, just to look up whenever a question comes up, without the need to dig in asset files

PR #1583, opened 2022-01-11
Dockerfile: update npm in the docker image

benma:
The package-lock.json file is at version 2, which requires a higher
npm version than previously installed. It was built and committed
outside of Docker, hence the mismatch.

### DIFF
--- a/scripts/docker_install.sh
+++ b/scripts/docker_install.sh
@@ -42,6 +42,7 @@ apt-get install -y --no-install-recommends \
         libxkbcommon-x11-dev \
         libxrandr-dev
 
+npm install -g npm@8.3.0
 npm install -g locize-cli
 
 mkdir -p /opt/go_dist
--- a/scripts/travis-ci.sh
+++ b/scripts/travis-ci.sh
@@ -14,7 +14,7 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
     # Which docker image to use to run the CI. Defaults to Docker Hub.
     # Overwrite with CI_IMAGE=docker/image/path environment variable.
     # Keep this in sync with .github/workflows/ci.yml.
-    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:11}"
+    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:12}"
     # Time image pull to compare in the future.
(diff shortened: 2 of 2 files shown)


benma:
I'll update the CI to point to the new Docker image soon.

scripts/travis-ci.sh
```diff
@@ -14,7 +14,7 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
     # Which docker image to use to run the CI. Defaults to Docker Hub.
     # Overwrite with CI_IMAGE=docker/image/path environment variable.
     # Keep this in sync with .github/workflows/ci.yml.
-    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:11}"
+    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:12}"
     # Time image pull to compare in the future.
     time docker pull "$CI_IMAGE"
```
thisconnect, line 20: one day we should rename this file since we don't use travis anymore

scripts/docker_install.sh
```diff
@@ -42,6 +42,7 @@ apt-get install -y --no-install-recommends \
         libxkbcommon-x11-dev \
         libxrandr-dev
 
+npm install -g npm@8.3.0
```
thisconnect, line 45: WDYT about adding this also to the macOS travis-ci.sh after `nvm install 14.15.4 # install this node version` so we have the same npm version?

thisconnect (approved):
LGTM with question

scripts/travis-ci.sh
```diff
@@ -14,7 +14,7 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
     # Which docker image to use to run the CI. Defaults to Docker Hub.
     # Overwrite with CI_IMAGE=docker/image/path environment variable.
     # Keep this in sync with .github/workflows/ci.yml.
-    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:11}"
+    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:12}"
```
x1ddos, line 17: unfortunately, .github/workflows/ci.yml wasn't updated. i guess the comment on top doesn't help. so, github should still be building with the version 11, i assume.

PR #1584, opened 2022-01-11
frontend: fix utxos reference unwrapping

BrianCraig:
This fixes a `this.utxos.current.getWrappedInstance is not a function` error thrown on the `Send.tsx`

Send Component has a Ref to UTXOs Class Component, this reference was being unwrapped using getWrappedInstance, due 
 to using a translate HOC, which is not needed because the reference is already being forwarded by the HOC.
 
This difference may be because of using `withTranslation` from `i18n-react` after the React migration, which doesn't require the call to `getWrappedInstance()`, that the old `translate` HOC had.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -40,7 +40,7 @@ import { apiWebsocket } from '../../../utils/websocket';
 import { isBitcoinBased, customFeeUnit } from '../utils';
 import { FeeTargets } from './feetargets';
 import style from './send.module.css';
-import { Props as UTXOsProps, SelectedUTXO, UTXOs } from './utxos';
+import { SelectedUTXO, UTXOs, UTXOsClass } from './utxos';
 import { route } from '../../../utils/route';
 
 interface SendProps {
@@ -94,7 +94,7 @@ interface State {
 }
 
 class Send extends Component<Props, State> {
-    private utxos = createRef<Component<UTXOsProps>>();
+    private utxos = createRef<UTXOsClass>();
     private selectedUTXOs: SelectedUTXO = {};
     private unsubscribe!: () => void;
     private qrCodeReader?: BrowserQRCodeReader;
@@ -237,11 +237,7 @@ class Send extends Component<Props, State> {
                     customFee: '',
                 });
                 if (this.utxos.current) {
-                    try {
-                        (this.utxos.current as any).getWrappedInstance().clear();
-                    } catch (e) {
-                        console.error(e);
-                    }
+                    this.utxos.current.clear();
                 }
                 setTimeout(() => this.setState({
                     isSent: false,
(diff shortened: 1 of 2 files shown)


benma:
> This difference may be because of using `withTranslation`
from `i18n-react`  instead of the own removed translate HOC

I don't understand, can you elaborate?

benma:
`private utxos = createRef<UTXOsClass>();`

It seems wrong that the ref is of type `UTXOsClass` but the actual component mounted is of type `UTXOs`. How does that work?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -467,11 +463,7 @@ class Send extends Component<Props, State> {
     private toggleCoinControl = () => {
         this.setState(({ activeCoinControl }) => {
             if (activeCoinControl && this.utxos.current) {
```
benma, line 465: this branch looks like dead code. toggleCoinControl() is only ever called when activeCoinControl is false, from send.tsx. 

Maybe its a relict from a past where the button was actually a toggler, not a dialog opener.

I propose to rename this function to 'openCoinControl` or similar, and to rename 'Toggle coin control' to 'Open coin control' or similar. cc @thisconnect wdyt?

thisconnect: ack, makes sense.

thisconnect:
> It seems wrong that the ref is of type UTXOsClass but the actual component mounted is of type UTXOs. How does that work?

`UTXOs` is `UTXOsClass` but `withTranslation` HOC from the `react-i18next` module. Somehow it fails to detect that there is a public `clear` method on that class if the UTXOs is used, it seems `react-i18next` doesn't support having a public method on the instance.


Here is what happens if the ref type is UTXOs, `createRef<UTXOs>();` to be honest I don't fully get what the `react-i18next` WithTranslation HOC type omits and subtracts.

```bash
TS2339: Property 'clear' does not exist on type 'ComponentType<Omit<Subtract<UTXOsProps & TranslateProps, WithTranslationProps>, keyof WithTranslation<N>> & WithTranslationProps>'.
  Property 'clear' does not exist on type 'ComponentClass<Omit<Subtract<UTXOsProps & TranslateProps, WithTranslationProps>, keyof WithTranslation<N>> & WithTranslationProps, any>'.
    238 |                 });
    239 |                 if (this.utxos.current) {
  > 240 |                     this.utxos.current.clear();
        |                                        ^^^^^
    241 |                 }
    242 |                 setTimeout(() => this.setState({
    243 |                     isSent: false,
```

We still use class + HOC, but the other option could be to rewrite to FunctionalComponent without class + `useTranslation` hook (as we already did in the newer components). But this seems to be the smallest possible fix with the current send.tsx and utxo.tsx component as is.

Also I believe getting the child instance and trying to call `clear` is a bit messy. Better would be to just have a callback and keep the state in the parent and pass it to the child, so the child is a controlled component. Now we have the data in 2 places, in the child's state and in the instance of the parent.

benma (approved):

thisconnect (approved):
LGTM

tested this PR on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1586

after sending some selected outputs (coincontrol) the outputs were unselected after successful send as expected, no error and no "getWrappedInstance is not a function" error

PR #1585, opened 2022-01-11
frontend: fix select dropdown styling

thisconnect:
The dropdowns in the Qt app had 2 arrows, this is a regression
introduced when we changed to the new build.
    
BitBoxApp is using Qt 5.15.2 that is actually using Chromium 83
- https://wiki.qt.io/QtWebEngine/ChromiumVersions
    
Changing browserlist to build for 83 fixes the styling as the
reset has to be prefixed with -webkit, see:
- https://caniuse.com/css-appearance

### DIFF
--- a/frontends/web/package.json
+++ b/frontends/web/package.json
@@ -34,7 +34,7 @@
   },
   "browserslist": {
     "production": [
-      "chrome >= 87"
+      "chrome >= 83"
     ],
     "development": [
       "last 1 chrome version",


x1ddos:
actually, [we're building the app with qt 5.15.2](https://github.com/digitalbitbox/bitbox-wallet-app/blob/2dacf986/frontends/qt/docker-qt5base/Dockerfile#L86) and https://wiki.qt.io/QtWebEngine/ChromiumVersions says:

image

so, are we actually at chromium 83? if so, is it possible our qtwebengine was a little behind some chromium v83 that caniuse refers to, and in fact requires webkit prefix?

should probably also update browserlist from 87 to 83?

thisconnect:
@x1ddos good catch thanks!

x1ddos (approved):
LGTM 

image

PR #1586, opened 2022-01-12
frontend: disallow undefined value in controlled form inputs

benma:
In React, changing the value of a controlled form input from undefined
to a defined value makes it uncontrolled. A warning popped up in the
console and the send tranasction form broke: it could no be cleared
after sending a transaction (setState() did not affect the form
anymore), disabling send all didn't remove the input value again, etc.

This commit removes `undefined` in the use of controlled form elements.


benma:
cc @BrianCraig

thisconnect (approved):
Great catch, thanks! LGTM

PR #1588, opened 2022-01-13
frontend/send: fix setting note proposal

benma:
Regression from preact to React: setState() is async, so accessing
this.state.note after setState() can give an older non-updated value.

Entering e.g. 'test' in the note, the final tx can have 'tes' as the
actual note.


thisconnect (approved):
LGTM

tested:
- creating a note during send, shows the expected note on tx overview
- qrcode with and without amount, updated to correct coin and fiat amount and checked

PR #1589, opened 2022-01-13
fontend/eth: remove dead code

benma:
The data field is unused.


backend/accounts/account.go
```diff
@@ -35,7 +35,6 @@ type TxProposalArgs struct {
 	// Only applies if FeeTargetCode == Custom. It is provided in sat/vB for BTC/LTC and Gwei for ETH.
 	CustomFee     string
 	SelectedUTXOs map[wire.OutPoint]struct{}
-	Data          []byte
```
thisconnect, line 38: Q: also the supported tokens inapp do not need data to create a TX proposal is that correct?

thisconnect (approved):
LGTM with question

backend/accounts/account.go
benma: that's correct

PR #1590, opened 2022-01-13
frontend: remove unused dead code favorites in manage account

thisconnect:
The code was supposed to keep a list of favorite accounts in the
frontend config to control which accounts are shown in the sidebar.
This was disabled and the plan was to use it later for tokens,
but tokens now use their own api endpoint and do not need to
be configurable in frontend config.

The unused code fetched favorites from frontend config with a
fallback to empty object. In various places there are checks for
if favorites is truthy, but favorites is not used anywhere else.

Removed favorite from manage-account component.


BrianCraig:
:+1: Looks good to me

x1ddos (approved):
LGTM

PR #1591, opened 2022-01-13
Frontend: eslint update identation

BrianCraig:
This updates eslint spacing to two spaces, and adds switch case spacing.

Also it adds a git blame recommended ignored commits, and a readme note for enabling them.

Being a PR that changes almost every file on the frontend, this should be made and merged in a short defined time to not interfere with other PRs, the commits serve as a guide to how to make the PR.

the git ignore blame file is intended to ignore this mass changes
Screenshot from 2022-01-13 08-32-04

note: the build is successful, despite this PR CI failing, due to a mismatch base commit with the `master` branch.

Screenshot from 2022-01-13 08-29-31

note 2: 
the correct command is `npx eslint --ext .jsx,.js,.ts,.tsx src/ --fix` because otherwise it doesn't checks for `.jsx` files to fix


frontends/web/.eslintrc.json
```diff
@@ -3,7 +3,8 @@
   "rules": {
     "jsx-a11y/anchor-is-valid": 0,
     "jsx-a11y/alt-text" : 0,
-    "quotes": ["error", "single"]
+    "quotes": ["error", "single"],
+    "indent": ["error", 2, { "SwitchCase": 1 }]
```
thisconnect, line 7: could you also add jsx quotes? https://eslint.org/docs/rules/jsx-quotes#prefer-double

.git-blame-ignore-revs
```diff
@@ -0,0 +1,12 @@
...
+# run `git config blame.ignoreRevsFile .git-blame-ignore-revs` to ignore 
+# the following autogenerated commits
+
+# frontend: `eslint` change identation to two spaces
+
+705423d833830d0522d9eda10f704271706e8d7e
+
+# frontend: eslint change switchcase identation
+
+0df2064cce37b26aa41061e1bd931500c71b73b1
```
thisconnect, line 12: not sure about this. One concern about different commits was with `git bisect`, when those 2 commits aren't building, bisect doesn't work.

maybe better just squash the 3 commits into 1 ?

cc @x1ddos

thisconnect (changes requested):
concept ack, added a comment about jsx quotes, also could you  the fix again on top of current master?

frontends/web/.eslintrc.json
thisconnect: I'd also like enforce consistent spacing inside braces (object-curly-spacing)
@benma @x1ddos @BrianCraig WDYT?

could we enforce `"object-curly-spacing": ["error", "always"],` 
https://eslint.org/docs/rules/object-curly-spacing#always

.git-blame-ignore-revs
BrianCraig: Those two commits work, this is simply an ignore rule for blame, not bisect

BrianCraig:
> concept ack, added a comment about jsx quotes, also could you the fix again on top of current master?
@thisconnect Let's check that everyone agrees to makes the fixes first, and when, because making the commits is time consuming and any change to `master` would make the commits unmergeable

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1699

PR #1592, opened 2022-01-13
frontend: control select component

thisconnect:
React has a defaultValue property on select dropdowns to set the
default value, but this is intendet to be used with uncontrolled
components. To control a select element React uses value.
Using both defaultValue and value triggers a warning: Select
elements must be either controlled or uncontrolled (specify
either the value prop, or the defaultValue prop, but not both).

AOPP, buy/choose-account and manage-accounts/select-coin all had
a defaultValue and a value props. As far as I can see they could
be all uncontrolled as there is no other form field or other
data that could change the select.
Decided to go with controlled component using value and removing
defaultValue, because I stlightly prefer controlled, i.e. if we
ever add another field that may change the value in the future.

### DIFF
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -216,8 +216,7 @@ class BuyInfo extends Component<Props, State> {
                                             ...options]
                                         }
                                         onChange={(e: React.SyntheticEvent) => this.setState({ selected: (e.target as HTMLSelectElement).value})}
-                                        value={selected}
-                                        defaultValue={'choose'}
+                                        value={selected || 'choose'}
                                         id="coinAndAccountCode"
                                     />
                                     <div className="buttons text-center m-bottom-xxlarge">
--- a/frontends/web/src/components/aopp/aopp.tsx
+++ b/frontends/web/src/components/aopp/aopp.tsx
@@ -154,7 +154,6 @@ class Aopp extends Component<Props, State> {
                                 <Select
                                     label={t('buy.info.selectLabel')}
                                     options={options}
-                                    defaultValue={options[0].value}
                                     value={accountCode}
                                     onChange={e => this.setState({ accountCode: (e.target as HTMLSelectElement)?.value })}
                                     id="account" />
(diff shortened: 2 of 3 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -216,8 +216,7 @@ class BuyInfo extends Component<Props, State> {
                                             ...options]
                                         }
                                         onChange={(e: React.SyntheticEvent) => this.setState({ selected: (e.target as HTMLSelectElement).value})}
-                                        value={selected}
-                                        defaultValue={'choose'}
+                                        value={selected || 'choose'}
```
benma, line 219: this component is routed to with this:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/pr%2F1544/frontends/web/src/routes/router.tsx#L109

In preact, there was this hack that `code` could be undefined, even though it was actually always defined. Did this change in react? If `code` can now always be defined, you could make `code` and `selected` a string instead of `string | undefined`, and remove ` || 'choose'`.

thisconnect: ah yes, that will work, just before I push you mean like this:

```diff
diff --git a/frontends/web/src/routes/buy/info.tsx b/frontends/web/src/routes/buy/info.tsx
index 6ef4ecad9..e4877a9ad 100644
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -32,7 +32,7 @@ import style from './info.module.css';
 
 interface BuyInfoProps {
     accounts: IAccount[];
-    code?: string;
+    code: string;
     devices: TDevices;
 }
 
@@ -216,7 +216,7 @@ class BuyInfo extends Component<Props, State> {
                                             ...options]
                                         }
                                         onChange={(e: React.SyntheticEvent) => this.setState({ selected: (e.target as HTMLSelectElement).value})}
-                                        value={selected || 'choose'}
+                                        value={selected}
                                         id="coinAndAccountCode"
                                     />
                                     <div className="buttons text-center m-bottom-xxlarge">
diff --git a/frontends/web/src/routes/router.tsx b/frontends/web/src/routes/router.tsx
index b72df7429..ad97e8a0c 100644
--- a/frontends/web/src/routes/router.tsx
+++ b/frontends/web/src/routes/router.tsx
@@ -74,6 +74,7 @@ export const AppRouter: FunctionComponent<Props> = ({ devices, deviceIDs, device
 
     const BuyInfoEl = <InjectParams>
         <BuyInfo
+            code="choose"
             devices={devices}
             accounts={activeAccounts} />
     </InjectParams>

```

benma: in that case nevermind, then the current version is better I think. One because it seems like the code prop would be passed twice to the component in case of the `:code` route, and also because the default value should be chosen inside the component, not by the parent component.

PR #1593, opened 2022-01-13
CHANGELOG: finalize 4.31.0

benma:
Release date is in the future, so for now we just mention the tagged date.


thisconnect (approved):
LGTM

PR #1596, opened 2022-01-17
docs: remove build status from readme

x1ddos:
I find it hardly useful and propose to delete it from the readme
completely. The fact that we've had "unknown" travis build status for
months now and nobody complained or cared to update tells me nobody
really needs it in the readme.


thisconnect (approved):
LGTM

PR #1597, opened 2022-01-18
frontend: fix electrum ui initialization

thisconnect:
There was a regression that the configured nodes only showed
TLS, but not what the actual address/domain is.

The problem was that the default state of the servers was
initialized based on props in the constructor. Using setState in
the constructor worked in Preact, but fails in React without an
error or warning.

Changed to intialize the state late in componentDidMount callback
so it shows now correct server settings.

Fixes https://github.com/digitalbitbox/bitbox-wallet-app/issues/1603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Add a file picker dialog to choose where to export a CSV to
+- Fix display of server name and checking the server connection in 'Connect your own full node'
 
 ## 4.31.0 [tagged 2022-01-13]
 - Bundle BitBox02 firmware version v9.9.0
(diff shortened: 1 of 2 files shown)


thisconnect:
TODO: test check connection button

thisconnect:
Couldn't find any other occurrences in the codebase that does this.setState in a constructor

x1ddos:
nice. what about:

> TODO: test check connection button

still todo or good?

benma (approved):
tACK. please rebase and update the changelog, then let's merge :tada:

thisconnect:
@benma added changelog entry and rebased to upstream/master

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Add a file picker dialog to choose where to export a CSV to
+- Fix connect and check to own full node
```
benma, line 5: connect worked, just the server name display and check didn't, right?

PR #1599, opened 2022-01-19
frontend: prevent flicker on unlock

thisconnect:
After device has been connected, there is no need for
bitbox02.tsx to route into `/` or `/account-summary`
because `app.tsx` is already listening to device
and accounts status changes, and routing to `/account-summary`
when the device initializes.
This prevents the rendering of `/` before accounts list
has been synced.


thisconnect:
This currently skips the success screen.

thisconnect:
@BrianCraig should we add a deticated route for the success view?

thisconnect:
out of date

PR #1602, opened 2022-01-20
Update Go CI Lint

maxmousee:
I'm trying to generate a native build using Apple Silicon and the first roadblock I hit was this library, for some reason and updating it seems to make the process go further.
I hope this helps :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ catch:
 	@echo "Choose a make target."
 envinit:
 	# Keep golangci-lint version in sync with what's in .github/workflows/ci.yml.
-	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOPATH)/bin v1.27.0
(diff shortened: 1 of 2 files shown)


benma:
We also need to build and update a new Docker image and running the CI with the new image.

benma:
I pushed an updated Docker image. Could you please change this to tag `:13` instead of `:12`?

https://github.com/digitalbitbox/bitbox-wallet-app/blob/b25823611a3d6b0dce0d3dde46678fac7a304abf/scripts/travis-ci.sh#L17

maxmousee:
> I pushed an updated Docker image. Could you please change this to tag `:13` instead of `:12`?
> 
> https://github.com/digitalbitbox/bitbox-wallet-app/blob/b25823611a3d6b0dce0d3dde46678fac7a304abf/scripts/travis-ci.sh#L17

Sure :)
Thank you very much

benma (approved):
thanks!

PR #1615, opened 2022-01-24
Use karalbe hid

BrianCraig:


benma:
We used to use the hid dependency went from hid to usb here fyi: https://github.com/digitalbitbox/bitbox-wallet-app/commit/bef0672b50ee29b0e92ac7f06087c233a58f5231

benma:
See also: https://github.com/ethereum/go-ethereum/blob/master/go.mod#L47

benma:
Having both deps in the transitive deps graph doesn't sound good. Closing this, as the current usb dep works.

PR #1619, opened 2022-01-26
Staging i18next > master

thisconnect:


thisconnect:
@benma we have to rename some they keys in locize

benma:
> @benma we have to rename some they keys in locize

i don't know how this works. if we locize-push after this PR, maybe locize will delete the previous keys (and all translations) and add the new ones, untranslated, which would be losing translations. need to look at docs to learn how to rename a key for all languages.

x1ddos:
> if we locize-push after this PR, maybe locize will delete the previous keys (and all translations)

it is possible to make a locize backup first? assuming one can restore from it later.

thisconnect:
i think we can pull, checkout the commit and then push, let's discuss offline.

benma (approved):
utACK

PR #1620, opened 2022-01-31
Portfolio: support showing subtotals on account summary (portfolio)

thisconnect:
- add subtotal row if there is more that one account of the same coin
- fix styling and labels
- as this also changes table headers of the account summary there is also changes to table headers of account overview

Everything can be squashed into 1 commit IMO.

TODO:
- review backend implementation

### DIFF
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -32,6 +32,7 @@ import { Chart } from './chart';
 import { AddBuyOnEmptyBalances } from '../info/buyCTA';
 import { apiPost } from '../../../utils/request';
 import style from './accountssummary.module.css';
+import { route } from '../../../utils/route';
 
 interface AccountSummaryProps {
     accounts: accountApi.IAccount[];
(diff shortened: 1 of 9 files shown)


thisconnect:
Screen Shot 2022-01-31 at 5 51 48 PM

Screen Shot 2022-01-31 at 5 51 21 PM

benma:
> RETH and TETH are currently treated as separate coins, should backend return a subtotal ?

They are separate coins/networks, so no subtotal for these.

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -50,33 +51,48 @@ interface State {
...
 
 interface BalanceRowProps {
-    code: string;
+    code: accountApi.AccountCode;
     name: string;
-    balance?: accountApi.IBalance;
+    balance?: string;
     coinUnit: string;
     coinCode: accountApi.CoinCode;
+    coinName?: string;
```
benma, line 65: `?` not needed, it is always defined (i'll fix it)

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -188,6 +221,63 @@ class AccountsSummary extends Component<Props, State> {
...
+                </td>
+            </tr>
+        );
+    }
+
+    private renderAccountSummary() {
+        const { accountsPerCoin } = this;
+        const { totalBalancePerCoin } = this.state;
+        if (!totalBalancePerCoin) {
+            return null;
```
benma, line 265: this stops rendering until all balances are shown, especially the 'scanned ... addresses' progress indicator is gone. this can simply be removed (i'll do that).

frontends/web/src/routes/account/summary/accountssummary.tsx
benma: done

frontends/web/src/routes/account/summary/accountssummary.tsx
benma: done

thisconnect:
@benma good point, changed (fixed-up/force-pushed) to only route to account overview if the accountName is clicked. PTAL

benma (approved):

PR #1621, opened 2022-02-02
qt: release non-posix malloc'ed memory with a compatible free

x1ddos:
This ensures the app doesn't crash after invoking getSaveFilename
dialog, for example when exporting transactions to CSV file, due to
memory release method incompatible with the one used during allocation.

Windows has different memory management models [1]. Relevant to this
case are: the C++ new/malloc which invokes HeapAlloc [2] and g++ mingw
malloc from portable C runtime. While malloc called from the Qt
frontend's main.cpp uses the former, Go's C.free uses the latter.

In fact, win32 docs state:

> Because the different heap allocators provide distinctive
> functionality by using different mechanisms, you must free memory
> with the correct function.

Before the filepicker dialog, we've always malloc'ed memory with cgo
functions like C.CString and released with the compatible C.free.
The filepicker was the first instance of a memory malloc'ed in C++ and
free'd from cgo. So, never had this problem before.

Linux and macOS seem to be happy either way. But of course a compatible
way is the above quote from win32 docs.

[1]: https://docs.microsoft.com/en-us/windows/win32/memory/comparing-memory-allocation-methods
[2]: http://msdn.microsoft.com/en-us/library/aa366597%28v=vs.85%29.aspx

please try on all: windows, linux and macOS

### DIFF
--- a/frontends/qt/libserver.h
+++ b/frontends/qt/libserver.h
@@ -22,6 +22,14 @@ typedef char* (*getSaveFilenameCallback) (const char*);
 static char* getSaveFilename(getSaveFilenameCallback f, const char* suggestedfilename) {
     return f(suggestedfilename);
 }
+
+// equivalent to C.free but suitable for releasing a memory malloc'ed
+// in a non-posix portable environment, incompatible with cgo.
+// this is especially important on windows where the standard C runtime
+// memory management used by cgo and mingw is different from win32 API used
+// when compiling C++ code with MSVC. hence, the memory allocated with malloc
+// in C++ must always be freed by this function in Go instead of C.free.
+typedef void (*cppHeapFree) (void* ptr);
 #endif
 
 #ifdef __cplusplus
@@ -34,6 +42,7 @@ extern void backendCall(int p0, char* p1);
 extern void handleURI(char* p0);
 
 extern void serve(
+    cppHeapFree cppHeapFreeFn,
     pushNotificationsCallback pushNotificationsFn,
     responseCallback responseFn,
     notifyUserCallback notifyUserFn,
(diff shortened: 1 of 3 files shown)


thisconnect:
tested a3b9491c1 getSaveFilename  dialog looks good on macOS 11.6, tested save in different dirs, but open-file only works if the file was saved in Downloads.

frontends/qt/libserver.h
```diff
@@ -22,6 +22,14 @@ typedef char* (*getSaveFilenameCallback) (const char*);
 static char* getSaveFilename(getSaveFilenameCallback f, const char* suggestedfilename) {
     return f(suggestedfilename);
 }
+
+// equivalent to C.free but suitable for releasing a memory malloc'ed
+// in a non-posix portable environment, incompatible with cgo.
+// this is especially important on windows where the standard C runtime
+// memory management used by cgo and mingw is different from win32 API and
+// hence the memory malloc'ed by the "new" operator in C++ or HeapAlloc must
```
benma, line 30: why "new" operator? we use `malloc()` to put the string into the heap.

benma (approved):
utACK, thanks!

frontends/qt/libserver.h
x1ddos: because both `new ...` and `malloc` would use the same heap in that memory world. just wanted to put more emphasis on distinguishing portable C runtime's `malloc` and "the other" `malloc`, and using only `malloc` as an example everywhere seemed more confusing.

as a matter of fact, these two allocate same amount of memory on the same heap in c++:

```c
a = (char*) malloc(10);
b = new char[10];
```

but because the `new` operator also calls a class constructor, when it's allocating memory for objects, you made me realize now i've probably made things more confusing. will remove the "new" operator.

PR #1622, opened 2022-02-03
backend: open CSV automatically after export

benma:
SystemOpen() has a whitelist, but the whitelist does not cover
exported CSV files - only the Downloads folder, but the user can save
it somewhere else.

Instead of adding CSV export paths dynamically to the whitelist, it is
simpler to automatically open the CSV. This also avoids the clumsy
hack of converting the button from 'export' to 'open'.

### DIFF
--- a/backend/accounts/baseaccount.go
+++ b/backend/accounts/baseaccount.go
@@ -57,6 +57,8 @@ type AccountConfig struct {
 	SigningConfigurations signing.Configurations
 	GetNotifier           func(signing.Configurations) Notifier
 	GetSaveFilename       func(suggestedFilename string) string
+	// Opens a file in a default application. The filename is not checked.
+	UnsafeSystemOpen func(filename string) error
 }
 
 // BaseAccount is an account struct with common functionality to all coin accounts.
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -20,7 +20,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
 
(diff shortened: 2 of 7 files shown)


backend/accounts/baseaccount.go
```diff
@@ -57,6 +57,8 @@ type AccountConfig struct {
 	SigningConfigurations signing.Configurations
 	GetNotifier           func(signing.Configurations) Notifier
 	GetSaveFilename       func(suggestedFilename string) string
+	// Opens a file in a default application. The filename is not checked.
+	UnsafeSystemOpen func(filename string) error
```
thisconnect, line 61: What do you mean by "The filename is not checked." ?

backend/backend.go
```diff
@@ -646,17 +643,6 @@ func (backend *Backend) SystemOpen(url string) error {
 		}
 	}
 
-	if runtime.GOOS != "android" { // TODO: fix DownloadsDir() for android
```
thisconnect, line 649: Does something change on Android?

thisconnect (approved):
LGTM, with 2 questions in the comments.

Tested on macOS 11.6, export and opens in user home, downloads, desktop dir. Got a expected permission error when trying to save somewhere without permissions.

backend/accounts/baseaccount.go
benma: simply that there are no checks on the filename and that the caller has to be sure that it is a valid and safe path.

backend/backend.go
benma: no, on android this branch was never executed.

PR #1610, opened 2022-01-24
Frontend: refactor `Waiting` component

BrianCraig:
- Migrate `ClassComponent` to `FunctionComponent`
 - Migrate HOCs to hooks
 - Removed unnescesary `testing` prop load

PR #1624, opened 2022-02-07
Frontend waiting refactor II

thisconnect:
- Migrate ClassComponent to FunctionComponent
- Migrate HOCs to hooks

This replaces https://github.com/digitalbitbox/bitbox-wallet-app/pull/1610
Added `testing` again, so testnet has the "skip for testing" see my comment https://github.com/digitalbitbox/bitbox-wallet-app/pull/1610#discussion_r792509146

### DIFF
--- a/frontends/web/src/routes/device/waiting.tsx
+++ b/frontends/web/src/routes/device/waiting.tsx
@@ -14,71 +14,62 @@
  * limitations under the License.
  */
 
-import { Component} from 'react';
+import { FunctionComponent, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useLoad } from '../../hooks/api';
+import { getTesting } from '../../api/backend';
 import { Entry } from '../../components/guide/entry';
 import { Guide, store as panelStore } from '../../components/guide/guide';
 import { AppLogo, SwissMadeOpenSource } from '../../components/icon/logo';
 import { Footer, Header } from '../../components/layout';
 import { setSidebarStatus } from '../../components/sidebar/sidebar';
-import { load } from '../../decorators/load';
-import { translate, TranslateProps } from '../../decorators/translate';
 import { debug } from '../../utils/env';
-import style from './bitbox01/bitbox01.module.css';
 import { SkipForTesting } from './components/skipfortesting';
+import style from './bitbox01/bitbox01.module.css';
 
-interface TestingProps {
-    testing?: boolean;
-}
+export const Waiting: FunctionComponent = () => {
+    const { t } = useTranslation();
+    const testing = useLoad(debug ? getTesting : () => Promise.resolve(false));
 
-type WaitingProps = TestingProps & TranslateProps;
-
-class Waiting extends Component<WaitingProps> {
-    public UNSAFE_componentWillMount() {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/device/waiting.tsx
```diff
@@ -14,71 +14,62 @@
...
+                <Header title={<h2>{t('welcome.title')}</h2>} />
+                <div className="content padded narrow isVerticallyCentered">
+                    <div>
+                        <AppLogo />
+                        <div className="box large">
+                            <h3 className={style.waitingText}>{t('welcome.insertDevice')}</h3>
+                            <p className={style.waitingDescription}>{t('welcome.insertBitBox02')}</p>
                         </div>
+                        {
+                            (debug && testing) && (
```
thisconnect, line 53: Minor difference, before the HOC querying `/testing` conditionally, only if debug was true.

Now it is always querying `/testing` and the condition is in the rendering.

frontends/web/src/api/testing.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { apiGet } from '../utils/request';
+
+export const getTesting = (): Promise<boolean> => {
```
benma, line 19: imo one backend.ts file is better than one file per api call (version.ts, testing.ts, etc.) A new file imho makes sense for sub-routes, e.g. `/aopp/...` or the accounts api.

frontends/web/src/routes/device/waiting.tsx
benma: Why not

`const testing = debug && useLoad(getTesting);` above, so the call only happens in debug? would that work? (useEffect etc are still a bit mysterious to me)

edit: i tried and this happend

> React Hook "useLoad" is called conditionally. React Hooks must be called in the exact same order in every component rende

This would work instead but is ugly:

    const testing = useLoad(debug ? getTesting : () => Promise.resolve(false));

benma (approved):
tACK

frontends/web/src/routes/device/waiting.tsx
thisconnect: yes it's a bit ugly 🤔 

ugly alternative

```js
const resolvePromise = (value: any) => Promise.resolve(value);

const testing = useLoad(debug ? getTesting : resolvePromise(false));

```

frontends/web/src/api/testing.ts
thisconnect: I agree having `api/using-mobile-data.ts` and `api/testing.ts` is ugly, but

IMO `backend.ts` has no meaning and shouldn't exist as everything is backend api and I'd rather refactor/rename the endpoints https://github.com/digitalbitbox/bitbox-wallet-app/blob/55d685c797cb2db5695e944c57ce3a8bce26c250/backend/handlers/handlers.go#L167-L206

PR #1608, opened 2022-01-24
Frontend: refactor `MobileDataWarning` component

BrianCraig:
- Migrate `ClassComponent` to `FunctionComponent`
 - Migrate HOCs to hooks

PR #1612, opened 2022-01-24
Frontend: refactor `Confirm` component

BrianCraig:
- Migrate `ClassComponent` to `FunctionComponent`
 - Migrate HOCs to hooks

PR #1625, opened 2022-02-08
Frontend confirmation refactor II

thisconnect:
Updated https://github.com/digitalbitbox/bitbox-wallet-app/pull/1608

Last commit moved callback to state hook, but the callback does
not impact rendering, changed to useRef hook.

This component is used in:
- BitBox01 mobile pairing
- create backup
- show mnemonic
- connect node: remove and reset to default

Deprecate Confirm component, exporting a function and updating
the state of the global confirm instance is a bit an antipattern.

### DIFF
--- a/frontends/web/src/components/confirm/Confirm.tsx
+++ b/frontends/web/src/components/confirm/Confirm.tsx
@@ -14,13 +14,19 @@
  * limitations under the License.
  */
 
-import { Component} from 'react';
-import { translate, TranslateProps } from '../../decorators/translate';
+import { FunctionComponent, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import { SimpleMarkup } from '../../utils/markup';
 import { Dialog, DialogButtons } from '../dialog/dialog';
 import { Button } from '../forms';
 
-let confirmation: (message: string, callback: (response: boolean) => void, customButtonText?: string) => void;
+type TCallback = (response: boolean) => void;
+
+/**
+ * @deprecated
+ * shows an alert when called, triggers the callback with user reply
+ */
+export let confirmation: (message: string, callback: TCallback, customButtonText?: string) => void;
 
 interface State {
     active: boolean;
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -14,85 +14,74 @@
...
-}
-
-const TranslatedConfirm = translate()(Confirm);
-
-export { confirmation };
-export { TranslatedConfirm as Confirm };
+            </div>
+        </div>
+        <DialogButtons>
+            <Button primary onClick={respond(true)}>{customButtonText ? customButtonText : t('dialog.confirm')}</Button>
```
benma, line 83: Why does this work? `respond(true)` is of type `()`. It looks like it would be executed immediately on rendering, but onClick should take a function of type `() => void`, e.g. `() => respond(true)`?

frontends/web/src/components/confirm/Confirm.tsx
```diff
@@ -14,85 +14,74 @@
...
+import { useTranslation } from 'react-i18next';
 import { SimpleMarkup } from '../../utils/markup';
 import { Dialog, DialogButtons } from '../dialog/dialog';
 import { Button } from '../forms';
 
-let confirmation: (message: string, callback: (response: boolean) => void, customButtonText?: string) => void;
+type TCallback = (response: boolean) => void;
+
+/**
+ * @deprecated
```
benma, line 26: if deprecated, there should be an alternative. what's the alternative?

benma (approved):
not sure why this is refactored when it is also deprecated at the same time, which means it will be removed?

otherwise tACK

thisconnect:
Yes, well most importantly it shoulnd't be used for new code. And yes it should be removed, but not part of this PR.

> not sure why this is refactored when it is also deprecated at the same time, which means it will be removed?

frontends/web/src/components/confirm/Confirm.tsx
thisconnect: It's syntax sugar called curring, under some circumstances it would be slightly more performant.

I also had to read twice, see my comment: 
https://github.com/digitalbitbox/bitbox-wallet-app/pull/1612#discussion_r791885941

I'll update the PR and remove it.

frontends/web/src/components/confirm/Confirm.tsx
thisconnect: Just use a dialog in the comonent instead of calling the global instance.

frontends/web/src/components/confirm/Confirm.tsx
benma: i see, i mistook the function => arrow as a type declaration arrow :see_no_evil: 

it would have helped to add a type to `respond`.

PR #1626, opened 2022-02-08
Frontend mobile data warning refactor II

thisconnect:
Updates https://github.com/digitalbitbox/bitbox-wallet-app/pull/1608/files

The default behavior of the subscribe HOC was to also load (fetch)
data from the same endpoint as it is subscribed to and did not
render until this request resolved.
    
Added a useLoad hook and added a check to only render if the
initial request loaded to have the same behavior as with the HOC.

### DIFF
--- a/frontends/web/src/components/mobiledatawarning.tsx
+++ b/frontends/web/src/components/mobiledatawarning.tsx
@@ -17,14 +17,13 @@
 
 import { FunctionComponent } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useLoad, useSubscribe } from '../hooks/api';
+import { useSync } from '../hooks/api';
 import { getUsingMobileData, subscribeUsingMobileData } from '../api/mobiledata';
 import Status from './status/status';
 
 export const MobileDataWarning: FunctionComponent = () => {
     const { t } = useTranslation();
-    const isUsingMobileData = useLoad(getUsingMobileData);
-    const usingMobileData = useSubscribe(subscribeUsingMobileData);
+    const isUsingMobileData = useSync(getUsingMobileData, subscribeUsingMobileData);
     if (isUsingMobileData === undefined) {
         return null;
     }
@@ -32,7 +31,7 @@ export const MobileDataWarning: FunctionComponent = () => {
         <Status
             dismissable="mobile-data-warning"
             type="warning"
-            hidden={usingMobileData === false || !isUsingMobileData}>
+            hidden={!isUsingMobileData}>
             {t('mobile.usingMobileDataWarning')}
         </Status>
     );
--- a/frontends/web/src/hooks/api.ts
+++ b/frontends/web/src/hooks/api.ts
@@ -42,7 +42,6 @@ export const useSubscribe = <T>(
     return respose;
 }
 
-
 /**
  * useLoad is a hook to load a promise.
  * gets fired on first render, and returns undefined while loading.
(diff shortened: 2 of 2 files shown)


frontends/web/src/components/mobiledatawarning.tsx
```diff
@@ -15,27 +15,25 @@
...
-interface LoadedProps {
-    usingMobileData: boolean;
-}
-
-type Props = LoadedProps & TranslateProps;
-
-function MobileDataWarning({ usingMobileData, t }: PropsWithChildren<Props>) {
+export const MobileDataWarning: FunctionComponent = () => {
+    const { t } = useTranslation();
+    const isUsingMobileData = useLoad(getUsingMobileData);
```
benma, line 26: i dislike that before it was one subscribe hoc and now it is two separate hooks (`useLoad` and `useSubscribe`), two functions `getUsingMobileData` and `subscribeUsingMobileData`, and double the logic below (`usingMobileData === false || !isUsingMobileData`).

This PR contains proof of that this is bugprone :) `getUsingMobileData` is calling the wrong endpoint.

In the old subscribe HOC, the subscription name is always the same as the endpoint name. I suggest that you add a combined subscribe/load hook that calls apiGet on the endpoint on load, as well as updates the value by listening on the same name in the subscription, to reduce all the complexity described above. WDYT?

benma (changes requested):

frontends/web/src/components/mobiledatawarning.tsx
thisconnect: well technically it called 2 decorator HOC's before, `subscribe` and `load`

> one subscribe hoc 

thanks for catching the wrong endpoint, will fix this. Do you want that the endpoint is a constant that is reused? maybe lets chat offline

thisconnect: I'd wish if `useSubscribe` would take a 2nd argument to set the default, but it could not default to  `useLoad(getUsingMobileData)` because that is async too and will first be undefined.

thisconnect:
@benma WDYT about

```diff
diff --git a/frontends/web/src/api/mobiledata.ts b/frontends/web/src/api/mobiledata.ts
index d9311829e..27e732648 100644
--- a/frontends/web/src/api/mobiledata.ts
+++ b/frontends/web/src/api/mobiledata.ts
@@ -17,12 +17,13 @@
 import { apiGet } from '../utils/request';
 import { subscribeEndpoint, SubscriptionCallback } from './subscribe';
 
-export const getUsingMobileData = (): Promise<boolean> => {
+const getUsingMobileData = (): Promise<boolean> => {
     return apiGet('using-mobile-data');
 };
 
 export const subscribeUsingMobileData = (
     cb: SubscriptionCallback<boolean>
-) => (
-    subscribeEndpoint('using-mobile-data', cb)
-);
+) => {
+    getUsingMobileData().then(cb);
+    return subscribeEndpoint('using-mobile-data', cb);
+};
diff --git a/frontends/web/src/components/mobiledatawarning.tsx b/frontends/web/src/components/mobiledatawarning.tsx
index 0a683a965..a49ea377a 100644
--- a/frontends/web/src/components/mobiledatawarning.tsx
+++ b/frontends/web/src/components/mobiledatawarning.tsx
@@ -17,22 +17,21 @@
 
 import { FunctionComponent } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useLoad, useSubscribe } from '../hooks/api';
-import { getUsingMobileData, subscribeUsingMobileData } from '../api/mobiledata';
+import { useSubscribe } from '../hooks/api';
+import { subscribeUsingMobileData } from '../api/mobiledata';
 import Status from './status/status';
 
 export const MobileDataWarning: FunctionComponent = () => {
     const { t } = useTranslation();
-    const isUsingMobileData = useLoad(getUsingMobileData);
     const usingMobileData = useSubscribe(subscribeUsingMobileData);
-    if (isUsingMobileData === undefined) {
+    if (usingMobileData === undefined) {
         return null;
     }
     return (
         <Status
             dismissable="mobile-data-warning"
             type="warning"
-            hidden={usingMobileData === false || !isUsingMobileData}>
+            hidden={!usingMobileData}>
             {t('mobile.usingMobileDataWarning')}
         </Status>
     );

**Comment by thisconnect** [comment]
@benma I made a new `useSync` hook that is a combination of `useLoad` and `useSubscribe`, see last commit, PTAL.

File: frontends/web/src/hooks/api.ts [hunk 2]
```diff
@@ -60,6 +59,33 @@ export const useLoad = <T>(
...
+}
+
+/**
+ * useSync is a hook to load a promise and sync to a subscription function.
+ * It is a combination of useLoad and useSubscribe.
+ * gets fired on first render, and returns undefined while loading,
+ * re-renders on every update.
+ */
+export const useSync = <T>(
+    apiCall: () => Promise<T>,
```
benma, line 74: the two args could be replaced with `endpoint: string`, so the caller doesn't have to provide two functions that always look the same. or is there a benefit in defining the two functions in mobiledata.ts?

benma:
LGTM, but I'll test it manually too.

frontends/web/src/hooks/api.ts
thisconnect: `mobiledata.ts` is not the best example as it just returns a boolean as arg or in a promise. 

For me the main motivation of `src/api` is that this is the interface to the backend and the place where all types from the api are defined. Before they were defined in various components.

I think `endpoint: string` would be nice, but I am not sure where the types should live. Maybe a giant endpoint/returntypes map somewhere?

benma (approved):
tACK, nice!

PR #1627, opened 2022-02-09
frontend: fix close dialog with esc js error

thisconnect:
Closing a confirm dialog with esc errors with:
> Cannot read properties of null (reading 'classList')

Seems that this is a regression from the Preact->React migration,
it looks like the setState callback is exectued earlier than in
Preact.

Moving updating the classList before setState fixes the error.

Test:
1) connect to own node
2) remove server
3) type esc


benma (approved):
utack

PR #1601, opened 2022-01-20
frontend: use view component for bitbox02 setup

thisconnect:
The BitBox02 setup styling is a bit out of date, there is a new View
component introduced with StupidSimpleMessageSigning and
enable passphrase that looks better with bigger font size on larger
screens and flatter without unnecessary white boxes.

Unlock is already using the View component, but the rest still
used an outdated step component. The step component once actually
showed steps, but this got changed and now this steps component is
only used as a container.

There are a lot of custom styles that are only used in bitbox02
setup and by using the View component the styling can be reused.
View module only need one change to support overwriting text align
to be left aligned in the content area of the view. This is needed
for the success and user confirmation list with checkboxes, that
the layout of those lists do not break.

The step where the user selects between create-new or restore
backup has 2 columns with both options. This could not been done
only with the View component only.

Changed:
- changed all steps from bitbox02.tsx to use View
- added new Grid and Column component to support content that is
  displayed side-by-side in different columns
- use inline (in content) Status and Message component instead
  of Toast component
- attestation failed warning is now shown in the content area
  of unlock and pairing view
- moved the warning graphic to Icons component and deprecated
  a similar icon that is used only for BitBox01
- removed outdated steps component and unused CSS
- moved styles from steps.module.css to bitbox02.module.css

PR #1628, opened 2022-02-11
frontend: fix white screen if passwords do not match during setup

thisconnect:
Regression from Reacht migration, the app currently shows a white
screen when password do not match. Reason setState callback fires
earlier it is unclear which setState sets createWalletStatus.

Test:
- factory reset
- create new wallet, enter name
- enter 2 different passwords
- expect password do not match error

Found in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1601

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -276,6 +276,8 @@ class BitBox02 extends Component<Props, State> {
                 }, () => {
                     this.setPassword();
                 });
+                // show noPasswordMatch error and do NOT continue to createBackup
+                return;
             }
             this.setState({ settingPassword: false, createWalletStatus: 'createBackup' });
         });


frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -276,6 +276,8 @@ class BitBox02 extends Component<Props, State> {
                 }, () => {
                     this.setPassword();
                 });
+                // show noPasswordMatch error and do NOT continue to createBackup
+                return;
             }
             this.setState({ settingPassword: false, createWalletStatus: 'createBackup' });
```
benma, line 282: should this maybe be also after the `set-password` response arrived, in case if `success`?

benma (approved):

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
benma: sry i misread and thought this was outside the response callback. all good

PR #1636, opened 2022-02-15
frontend: move inline svg icons in receive to icon component

thisconnect:
- moves inline SVG to icon component so it can be included as img src
- removed outdated CSS


benma (approved):
utack

PR #1638, opened 2022-02-16
Winbuild

benma:


frontends/qt/make_windows.sh
```diff
@@ -14,7 +14,7 @@
 
 set -e
 
-mkdir build
+mkdir -p build
```
thisconnect, line 17: I thought windows doesn't have `-p` option

thisconnect (approved):
LGTM

PR #1642, opened 2022-02-18
frontend/transaction: make transaction ID copyable

benma:
And separate the link to the external block explorer. Users can then
more easily copy the txID without accidentally opening the explorer.

We add 5px of padding between the input and the copy button, otherwise
the input selection would overlap the button if there is a line break.


thisconnect (approved):
nice, LGTM

benma:
@thisconnect thx - i also added a changelog entry

PR #1644, opened 2022-02-20
backend/accounts: update existing accounts with taproot subaccount

benma:
If the keystore supports it (e.g. after upgrading BitBox02 to
v9.10.0), the persisted accounts are updated with the taproot
subaccount, to make taproot addresses available also for existing
accounts.

The order of the script configs in the account config at the moment
matters only in that the first entry is used as the default address
type for Moonpay, AOPP and as the default change address. Future PRs
will remove this implicit assumption so that the order does not matter.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -706,6 +706,69 @@ func (backend *Backend) persistDefaultAccountConfigs(keystore keystore.Keystore,
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


thisconnect (approved):
LGTM

Tested with new firmware 👍 
Using a BB with old firmware and the same wallet didn't show the Bitcoin accounts anymore.

PR #1646, opened 2022-02-20
frontend/send: fix send to self

benma:
Typechecker should have caught that :O


thisconnect (approved):
We can catch in the fututre once `"noImplicitAny": true` is enabled.

LGTM

PR #1651, opened 2022-02-23
frontend: add warning if taproot is selected as receive address

thisconnect:
Taproot is still new and not supported by all wallets, i.e. funds
might not show up in 3rd party watch-only wallets or not all
exchanges support it yet.

Adding a warning if taproot is selected.

### DIFF
--- a/frontends/web/src/routes/account/receive/receive.tsx
+++ b/frontends/web/src/routes/account/receive/receive.tsx
@@ -34,6 +34,7 @@ import { apiGet } from '../../../utils/request';
 import { getScriptName, isEthereumBased } from '../utils';
 import style from './receive.module.css';
 import { ArrowCirlceLeft, ArrowCirlceLeftActive, ArrowCirlceRight, ArrowCirlceRightActive } from '../../../components/icon';
+import { Message } from '../../../components/message/message';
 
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -272,6 +273,11 @@ class Receive extends Component<Props, State> {
                                         title={getScriptName(scriptType)}>
                                         {t(`receive.scriptType.${scriptType}`)}
                                     </Radio>
+                                    {scriptType === 'p2tr' && this.availableScriptTypes.indexOf('p2tr') === addressDialog.addressType && (
```
benma, line 276: why not the same condition for activeness as above in `checked=...`? `addressDialog.addressType === i`?

` this.availableScriptTypes.indexOf('p2tr')` is the same as `i` as we are mapping over it.

benma (changes requested):

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: 🙈   Fixed and squashed

benma (approved):
utACK, nice!

PR #1653, opened 2022-02-23
backend/btc: speed up SpendableOutputs()

benma:
If an account has many transactions, SpendableOutputs() was very
slow. The reason for this is that the outputs set that it iterates
over can be very large, and for output we loaded the corresponding
transaction from the database.

We only need to load the transaction if the output is still unspent
though. The number of unspent outputs is usually much smaller than the
number of total outputs.


benma:
@thisconnect this is a quick optimization that should speed up the send screen. It could be optimized a lot further, e.g. by keeping an UTXO set, but this might already be enough.

PR #1663, opened 2022-03-03
frontend: reword restore-from-mnemonic instruction

thisconnect:
Before the popup when restoring from mnemonic words repeated
itself:
- Title: Please follow the instructions on your BitBox02
- Text: Please follow the i… to restore from recovery words

Reason: the title used a generic follow-instrucitons
translation which is used in many places. Fixed by adding
a custom followInstructionsMnemonicTitle translation.


thisconnect:
I also need to update locize…

benma (approved):

Issue #1662, opened 2022-03-03
Allow hardware rendering via env var or CLI flag


benma:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/549d6ec3e61dd2825a21a0d0250570da79965d84/frontends/qt/main.cpp#L200-L203

Removing these lines enables the default rendering backend, which is hardware if available. Doing this can lead to problems as described here: https://github.com/digitalbitbox/bitbox-wallet-app/commit/daaf2319afb3f5813a4bddd67e13b300590414e5

A user reported that the app was extremely slow on Windows 11, 4k resolution.

We should allow either setting these to env flags and only putting a default (using `qEnvironmentVariableIsSet` like with `QTWEBENGINE_DISABLE_SANDBOX`), or provide a custom env var or CLI argument to not force software rendering, so a user can try to see if it helps them.

PR #1665, opened 2022-03-03
qt: add BITBOXAPP_RENDER environment var to configure rendering

benma:
We switched to forced software rendering in
daaf2319afb3f5813a4bddd67e13b300590414e5 due to some users suffering
rendering artefacts and crashes with the default rendering backend.

The downside is that the app can be very slow on some machines,
especially when maximized with high resolution.

This env variable is a workaround to enable users to improve their
experience and try the default rendering, which is normally hardware
accelerated.

Closes https://github.com/digitalbitbox/bitbox-wallet-app/issues/1662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Show coin subtotals in 'My portfolio'
 - Transaction details: make transaction ID copyable without opening the block explorer
 - Improve account loading speed when there are many transactions in the account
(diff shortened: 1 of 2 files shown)


thisconnect (approved):
LGTM, tested `BITBOXAPP_RENDER=auto ./frontends/qt/build/osx/BitBox.app/Contents/MacOS/BitBox` on macOS 11.6

PR #1667, opened 2022-03-08
frontend: show errors from qr scanner

thisconnect:
If the app does not have permissions to use the camera the library
returns 'Could not start video source' error.

This change shows the library error to the user in cases something
went wrong, but also if the scaning is aborted and no code has
been detected in that case the library returns 'Video stream has
ended before any code could be detected'.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -518,7 +518,10 @@ class Send extends Component<Props, State> {
                         this.qrCodeReader.reset(); // release camera
                     }
                 })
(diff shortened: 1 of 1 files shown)


thisconnect:
First time using the app on macOS

Screen Shot 2022-03-07 at 11 24 36 AM

If the app doesn't have permission:

Screen Shot 2022-03-07 at 11 26 02 AM

If it was scanning but aborted by the user and no code was found:

Screen Shot 2022-03-07 at 11 26 27 AM

benma (approved):

PR #1668, opened 2022-03-08
frontend: rewrite receive to functional component

thisconnect:
- removed HOC decorators and typed has-secure-output api
- changed receive-address api to work with useLoad
- moved PairedWarning to sub component to load it conditionally
- before: if warning.secureOutput was shown it removed the keydown
  that is used to esc the receive view, and added again after the
  alert callback. This is not necessary as handleKeyDown only
  escapes if verifying is false.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -214,8 +214,10 @@ export interface ReceiveAddressList {
     addresses: IReceiveAddress[];
 }
 
-export const getReceiveAddressList = (code: AccountCode): Promise<ReceiveAddressList[]> => {
-    return apiGet(`account/${code}/receive-addresses`);
+export const getReceiveAddressList = (code: AccountCode) => {
+    return (): Promise<ReceiveAddressList[]> => {
+        return apiGet(`account/${code}/receive-addresses`);
+    };
 };
 
 export interface ISendTx {
@@ -278,4 +280,15 @@ export interface UTXO {
 
 export const getUTXOs = (code: AccountCode): Promise<UTXO[]> => {
     return apiGet(`account/${code}/utxos`);
+};
+
+type TSecureOutput = {
+    hasSecureOutput: boolean;
+    optional: boolean;
+};
+
+export const hasSecureOutput = (code: AccountCode) => {
+    return (): Promise<TSecureOutput> => {
+        return apiGet(`account/${code}/has-secure-output`);
+    };
 };
--- a/frontends/web/src/routes/account/receive/receive.tsx
+++ b/frontends/web/src/routes/account/receive/receive.tsx
@@ -31,8 +31,9 @@ import { Entry } from '../../../components/guide/entry';
 import { Guide } from '../../../components/guide/guide';
 import { Header } from '../../../components/layout';
 import { QRCode } from '../../../components/qrcode/qrcode';
-import { PairedWarning } from './components/bb01paired';
 import { ArrowCirlceLeft, ArrowCirlceLeftActive, ArrowCirlceRight, ArrowCirlceRightActive } from '../../../components/icon';
+import { PairedWarning } from './components/bb01paired';
+import { useVerfiyLabel, VerifyButton } from './components/verfiybutton';
 import style from './receive.module.css';
 
 interface Props {
@@ -76,6 +77,7 @@ export const Receive: FunctionComponent<Props> = ({
 
     const device = deviceIDs.length ? devices[deviceIDs[0]] : undefined;
     const account = accounts.find(({ code: accountCode }) => accountCode === code);
+    const verifyLabel = useVerfiyLabel(device);
 
     // first array index: address types. second array index: unused addresses of that address type.
     const receiveAddresses = useLoad(accountApi.getReceiveAddressList(code));
@@ -143,13 +145,6 @@ export const Receive: FunctionComponent<Props> = ({
     // enable copying only after verification has been invoked if verification is possible and not optional.
     const forceVerification = secureOutput === undefined ? true : (secureOutput.hasSecureOutput && !secureOutput.optional);
     const enableCopy = !forceVerification;
-
-    let verifyLabel = t('receive.verify'); // fallback
-    if (device === 'bitbox') {
-        verifyLabel = t('receive.verifyBitBox01');
-    } else if (device === 'bitbox02') {
-        verifyLabel = t('receive.verifyBitBox02');
-    }
 
     let uriPrefix = '';
     if (account) {
@@ -252,22 +247,11 @@ export const Receive: FunctionComponent<Props> = ({
                                     </form>
                                 )}
                                 <div className="buttons">
-                                    { forceVerification && (
-                                        <Button
-                                            primary
-                                            disabled={verifying || secureOutput === undefined}
-                                            onClick={() => verifyAddress(currentAddressIndex)}>
-                                            {t('receive.showFull')}
-                                        </Button>
-                                    )}
-                                    { !forceVerification && (
-                                        <Button
-                                            primary
-                                            disabled={verifying || secureOutput === undefined}
-                                            onClick={() => verifyAddress(currentAddressIndex)}>
-                                            {verifyLabel}
-                                        </Button>
-                                    )}
+                                    <VerifyButton
+                                        device={device}
+                                        disabled={verifying || secureOutput === undefined}
+                                        forceVerification={forceVerification}
+                                        onClick={() => verifyAddress(currentAddressIndex)}/>
                                     <ButtonLink
                                         transparent
                                         to={`/account/${code}`}>
--- a/frontends/web/src/routes/account/receive/receive.tsx
+++ b/frontends/web/src/routes/account/receive/receive.tsx
@@ -15,377 +15,312 @@
  * limitations under the License.
  */
 
-import React, { Component} from 'react';
-import { route } from '../../../utils/route';
+import React, { FunctionComponent, useEffect, useRef, useState} from 'react';
+import { useTranslation } from 'react-i18next';
+import { useLoad } from '../../../hooks/api';
 import * as accountApi from '../../../api/account';
 import { TDevices } from '../../../api/devices';
+import { route } from '../../../utils/route';
+import { getScriptName, isEthereumBased } from '../utils';
 import { alertUser } from '../../../components/alert/Alert';
 import { CopyableInput } from '../../../components/copy/Copy';
 import { Dialog, DialogButtons } from '../../../components/dialog/dialog';
 import { Button, ButtonLink, Radio } from '../../../components/forms';
+import { Message } from '../../../components/message/message';
 import { Entry } from '../../../components/guide/entry';
 import { Guide } from '../../../components/guide/guide';
 import { Header } from '../../../components/layout';
 import { QRCode } from '../../../components/qrcode/qrcode';
-import Status from '../../../components/status/status';
-import { load } from '../../../decorators/load';
-import { translate, TranslateProps } from '../../../decorators/translate';
-import { apiGet } from '../../../utils/request';
-import { getScriptName, isEthereumBased } from '../utils';
-import style from './receive.module.css';
+import { PairedWarning } from './components/bb01paired';
 import { ArrowCirlceLeft, ArrowCirlceLeftActive, ArrowCirlceRight, ArrowCirlceRightActive } from '../../../components/icon';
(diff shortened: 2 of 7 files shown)


frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -15,376 +15,309 @@
...
+    devices,
+    deviceIDs,
+}) => {
+    const { t } = useTranslation();
+    const [verifying, setVerifying] = useState<boolean>(false);
+    const [activeIndex, setActiveIndex] = useState<number>(0);
+    // index into `availableScriptTypes`, or 0 if none are available.
+    const [addressType, setAddressType] = useState<number>(0);
+    const [addressDialog, setAddressDialog] = useState<AddressDialog>();
+    const [availableScriptTypes, setAvailableScriptTypes] = useState<accountApi.ScriptType[]>();
```
benma, line 63: this was not in the state before, so it shouldn't be in the state now? it is a derived value

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -15,376 +15,309 @@
...
+        }
+        document.addEventListener('keydown', handleKeyDown);
+        return () => document.removeEventListener('keydown', handleKeyDown);
+    }, [code, verifying]);
 
-    private registerEvents = () => {
-        document.addEventListener('keydown', this.handleKeyDown);
-    }
+    // Find index in list of receive addresses that matches the given script type, or -1 if not found.
+    const getIndexOfMatchingScriptType = useCallback((
```
benma, line 85: i find this pretty complicated, why use `useCallback` when it could just be a regular function?

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -15,376 +15,309 @@
...
 
-    private unregisterEvents = () => {
-        document.removeEventListener('keydown', this.handleKeyDown);
-    }
+    useEffect(() => {
+        if (receiveAddresses) {
+            // All script types that are present in the addresses delivered by the backend. Will be empty for if there are no such addresses, e.g. in Ethereum.
+            setAvailableScriptTypes(scriptTypes.filter(sc => getIndexOfMatchingScriptType(sc) >= 0));
+        }
+    }, [receiveAddresses, getIndexOfMatchingScriptType]);
```
benma, line 99: what does `getIndexOfMatchingScriptType` in the deps least mean? it is a function after all.

imo everything would be easier to read if `getIndexOfMatchingScriptType` took `receiveAddresses` as an argument and it was used directly without putting it here or using `useCallback`

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -15,376 +15,309 @@
...
+            setActiveIndex(activeIndex + 1);
         }
-    }
+    };
 
-    private getAccount = () => {
-        return this.props.accounts.find(({ code }) => code === this.props.code);
-    }
+    // enable copying only after verification has been invoked if verification is possible and not optional.
+    const forceVerification = secureOutput && secureOutput.hasSecureOutput && !secureOutput.optional;
```
benma, line 142: now while `hasSecureOutput()` is not done yet, `forceVerification` is false. Before, the whole screen was not rendered.

In absence of a spinner, maybe defaulting to `true` is better, as that optimizes for the BitBox02?

frontends/web/src/api/account.ts
```diff
@@ -214,8 +214,10 @@ export interface ReceiveAddressList {
     addresses: IReceiveAddress[];
 }
 
-export const getReceiveAddressList = (code: AccountCode): Promise<ReceiveAddressList[]> => {
-    return apiGet(`account/${code}/receive-addresses`);
+export const getReceiveAddressList = (code: AccountCode) => {
```
benma, line 217: does it make sense to have some api functions return the promise directly while others return a function returning a promise? if you didn't change this function, would `useLoad(() => accountApi.getReceiveAddressList(code))` work? imo that would be a bit less magical and more expected.

i noticed because `accountApi.getReceiveAddressList(this.getAccount()!.code)()` looked a bit ugly.

thisconnect: `accountApi.getReceiveAddressList(this.getAccount()!.code)()` looks indeed ugly, but will look nice once it  uses `useLoad()` hook, but that needs to be converted from class to FunctionalComponent. I wouldn't optimize for class.

We already use currying with `useSubscribe` https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/components/headerssync/headerssync.tsx#L33

and

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/api/coins.ts#L29



Also it would be nice if it works with `useSync` currently is a combination of `useLoad` + `useSubscribe` which should just work with curried callbacks.

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: I agree, but `availableScriptTypes` is not just a derived value that is used for rendering but also for a dependency array of a `useEffect`, so if `availableScriptTypes` changes useEffect will execute again.

The dependency array can contain normal variables too, but then I get:
Screen Shot 2022-03-14 at 5 05 36 PM

Changing to use `useRef`

```diff
diff --git a/frontends/web/src/routes/account/receive/receive.tsx b/frontends/web/src/routes/account/receive/receive.tsx
index a949a275..c362d86b 100644
--- a/frontends/web/src/routes/account/receive/receive.tsx
+++ b/frontends/web/src/routes/account/receive/receive.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import React, { FunctionComponent, useCallback, useEffect, useState} from 'react';
+import React, { FunctionComponent, useCallback, useEffect, useRef, useState} from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLoad } from '../../../hooks/api';
 import * as accountApi from '../../../api/account';
@@ -60,7 +60,6 @@ export const Receive: FunctionComponent<Props> = ({
     // index into `availableScriptTypes`, or 0 if none are available.
     const [addressType, setAddressType] = useState<number>(0);
     const [addressDialog, setAddressDialog] = useState<AddressDialog>();
-    const [availableScriptTypes, setAvailableScriptTypes] = useState<accountApi.ScriptType[]>();
     const [currentAddresses, setCurrentAddresses] = useState<accountApi.IReceiveAddress[]>();
     const [currentAddressIndex, setCurrentAddressIndex] = useState<number>(0);
 
@@ -91,16 +90,18 @@ export const Receive: FunctionComponent<Props> = ({
         return receiveAddresses.findIndex(addrs => addrs.scriptType !== null && scriptType === addrs.scriptType);
     }, [receiveAddresses]);
 
+    const availableScriptTypes = useRef<accountApi.ScriptType[]>();
+
     useEffect(() => {
         if (receiveAddresses) {
             // All script types that are present in the addresses delivered by the backend. Will be empty for if there are no such addresses, e.g. in Ethereum.
-            setAvailableScriptTypes(scriptTypes.filter(sc => getIndexOfMatchingScriptType(sc) >= 0));
+            availableScriptTypes.current = scriptTypes.filter(sc => getIndexOfMatchingScriptType(sc) >= 0);
         }
     }, [receiveAddresses, getIndexOfMatchingScriptType]);
 
     useEffect(() => {
-        if (receiveAddresses && availableScriptTypes) {
-            let addressIndex = availableScriptTypes.length > 0 ? getIndexOfMatchingScriptType(availableScriptTypes[addressType]) : 0;
+        if (receiveAddresses && availableScriptTypes.current) {
+            let addressIndex = availableScriptTypes.current.length > 0 ? getIndexOfMatchingScriptType(availableScriptTypes.current[addressType]) : 0;
             if (addressIndex === -1) {
                 addressIndex = 0;
             }
@@ -166,7 +167,7 @@ export const Receive: FunctionComponent<Props> = ({
         }
     }
 
-    const hasManyScriptTypes = availableScriptTypes && availableScriptTypes.length > 1;
+    const hasManyScriptTypes = availableScriptTypes.current && availableScriptTypes.current.length > 1;
 
     return (
         <div className="contentWithGuide">
@@ -223,7 +224,7 @@ export const Receive: FunctionComponent<Props> = ({
                                         setAddressDialog(undefined);
                                     }}>
                                         <Dialog medium title={t('receive.changeScriptType')} >
-                                            {availableScriptTypes && availableScriptTypes.map((scriptType, i) => (
+                                            {availableScriptTypes.current && availableScriptTypes.current.map((scriptType, i) => (
                                                 <div key={scriptType}>
                                                     <Radio
                                                         checked={addressDialog.addressType === i}
```

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: yes you are right, I'll just use a normal function, shouldn't matter performance wise.

thisconnect: Now I remember why: "TypeScript told me so" also makes a bit sense because the function is used as `useEffect` dependency. So that means: if the dependency array of `useCallback` does not change, then `useCallback` is still the same function and `useEffect` do not have to execute. 

See what happens without `useCallback` (another remark: it still looks complicated `useCallback` doesn't matter IMO)

Screen Shot 2022-03-14 at 5 58 50 PM

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: Yes correct: "it is a function after all",  but if the function is not the same function it needs to execute again, therefore it needs to be in the dependency array. Or else:

> React Hook useEffect has a missing dependency: 'getIndexOfMatchingScriptType'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

also

> The 'getIndexOfMatchingScriptType' function makes the dependencies of useEffect Hook (at line 100) change on every render. To fix this, wrap the definition of 'getIndexOfMatchingScriptType' in its own useCallback() Hook.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

Your suggestion makes sense I'll update the PR.

frontends/web/src/api/account.ts
benma: okay but then please declare the return type on this line instead of inside the function body if possible

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: changed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1668/commits/4cffb63e8cdd741b03e9123dcf57b616cdea6f6e

(not squashed yet)

thisconnect: changed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1668/commits/4cffb63e8cdd741b03e9123dcf57b616cdea6f6e

(not squashed yet)

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: done in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1668/commits/d4a2b379a27b867095f321bc367221020d4d69f3

(not squashed yet)

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: changed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1668/commits/d4a2b379a27b867095f321bc367221020d4d69f3

(not yet squashed)

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1668/commits/763d60138728f059e1f198851dbfcafe9e151a40

will squash that one too.

thisconnect: But I added 1 more commit that moved the verifylabel and button into its own component, here https://github.com/digitalbitbox/bitbox-wallet-app/pull/1668/commits/5ee5c3655243f228cee0b323fed19f7a08c2f2bf

should I keep the commit or also squash?

PTAL

benma (approved):
LGTM but let's merge after we tag v4.32.0?

PR #1669, opened 2022-03-08
frontend: add guide entry about change output

thisconnect:
Added guide entry to explain when the change output is a Taproot,
only shown in the guide of bitcoin accounts.

The rest was updated in locize and will need a locize-pull

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -763,6 +763,10 @@
       }
     },
     "send": {
+      "change": {
+        "text": "The change will be returned to a Taproot address if you have at least one Taproot UTXO. If you use coin control, the change will be returned to a Taproot address if there is at least one Taproot UTXO among the selected UTXOS. In all other cases, the change is returned to a Native Segwit address.",
+        "title": "How is the change output determined?"
+      },
       "fee": {
         "text": "The fee is based on the transaction data size and not its amount. The fee targets are calculated by Bitcoin Core's fee estimation algorithm for each network priority you chose. They are shown if they have a different value from the target below.\nEconomy: 24 blocks (around 4 hours for Bitcoin, 1 hour for Litecoin)\nLow: 12 blocks (around 2 hours for Bitcoin, 30 minutes for Litecoin)\nNormal: 6 blocks (around 1 hour for Bitcoin, 15 minutes for Litecoin)\nHigh: 2 blocks (around 20 minutes for Bitcoin, 5 minutes for Litecoin)\n(A block takes on average ten minutes for Bitcoin (2.5 minutes in Litecoin) to mine and the network load may vary considerably in the above periods.)",
         "title": "How is the fee determined?"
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -763,6 +763,10 @@
       }
     },
     "send": {
+      "change": {
+        "text": "The change will be returned to a Native Segwit address, unless you have at least one Taproot UTXO. In that case the change will be returned to a Taproot address.\n\nPlease note, if you are using coin control and select only Native Segwit UTXOs, the change will be returned to a Native Segwit address, regardless of whether you have a Taproot UTXO in that account. However, if you select Wrapped Swegit UTXOs, the change will still be returned to a Native Segwit address.",
```
benma, line 767: typo: Wrapped Swegit -> Wrapped Segwit

@jadzeidan this feels very convoluted, is there no easier and shorter way? How about:

"The change will be returned to a Taproot address if you have at least one Taproot UTXO.  If you use coin control, the change will be returned to a Taproot address if there is at least one Taproot UTXO among the selected UTXOS. In all other cases, the change is returned to a Native Segwit address."

benma:
> The rest was updated in locize and will need a locize-pull

image


Why drop `coinName`? In Litecoin the xpubs and this guide entry are also shown.

frontends/web/src/locales/en/app.json
jadzeidan: Sure, fine with me.

thisconnect: fixed and pushed

thisconnect:
> Why drop coinName? In Litecoin the xpubs and this guide entry are also shown.

Oversight by me, but reading the new sentence that has Taproot:

"Each xpub is tied to the "Type" shown: either "Native Segwit (bech32)" , "Wrapped Segwit" or “Taproot”. These are script types used by Bitcoin" 

Writing "used by Litecoin" would be wrong as it doesn't support Taproot.

benma:
> > Why drop coinName? In Litecoin the xpubs and this guide entry are also shown.
> 
> Oversight by me, but reading the new sentence that has Taproot:
> 
> "Each xpub is tied to the "Type" shown: either "Native Segwit (bech32)" , "Wrapped Segwit" or “Taproot”. These are script types used by Bitcoin"
> 
> Writing "used by Litecoin" would be wrong as it doesn't support Taproot.

in BB01 there is also P2PKH ;)

What about "Each xpub is tied to the "Type" shown, for example "Native Segwit (bech32)" or "Wrapped Segwit".

thisconnect:
cc @jadzeidan @Stadicus PTAL

Stadicus:
I can't judge the technical aspects, but the text looks good to me. :)

benma (approved):

jadzeidan:
> What about "Each xpub is tied to the "Type" shown, for example "Native Segwit (bech32)" or "Wrapped Segwit".

Sounds good to me.

PR #1671, opened 2022-03-10
Frontend remove links from locize

thisconnect:
There is no need to maintain links to the support section and other
resources in the translation. The original idea was to use different
links for each language but it turns out that this was never used.

Moving all remaining links from translation file to the frontend.

tested:

- Waitscreen order and backup center
- Accountsummary coingecko link
- Connect to own node (en and de)
- Guide appendix
- BitBox01 guide in non-unified bech32 accounts

### DIFF
--- a/frontends/web/src/routes/device/waiting.tsx
+++ b/frontends/web/src/routes/device/waiting.tsx
@@ -16,6 +16,7 @@
 
 import { FunctionComponent, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
+import i18n from '../../i18n/i18n';
 import { useLoad } from '../../hooks/api';
 import { getTesting } from '../../api/backend';
 import { Entry } from '../../components/guide/entry';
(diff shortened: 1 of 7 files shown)


frontends/web/src/routes/device/waiting.tsx
```diff
@@ -64,8 +65,24 @@ export const Waiting: FunctionComponent = () => {
...
+                        url: 'https://shiftcrypto.shop/',
+                    },
+                    text: t('guide.waiting.getDevice.text'),
+                    title: t('guide.waiting.getDevice.title'),
+                }} />
+                <Entry entry={{
+                    link: {
+                        text: t('guide.waiting.lostDevice.link.text'),
+                        url: (i18n.language === 'de')
+                            ? 'https://shiftcrypto.support/help/de-de/5-backup/8-wie-kann-ich-ein-bitbox02-wallet-in-ein-drittanbieter-wallet-importieren'
```
benma, line 80: interestingly, you can also use `de-de` with the english suffix :thinking: 

https://shiftcrypto.support/help/de-de/5-backup/8-how-do-i-restore-my-wallet-if-my-bitbox02-is-lost

benma (approved):
very nice, utACK

frontends/web/src/routes/device/waiting.tsx
thisconnect: yeah, the url you pasted url should not exist unfortunatelly that tool has no consisten links and as far as i know it uses 2 different links for the same page.

you can also mix or omit completely or write nonsense https://shiftcrypto.support/help/de-de/5-wagagagagaga/8-biiiiiitttttccccoooiiiin 🙈

PR #1673, opened 2022-03-12
frontend/send: enable QR code scanner for ETH-based accounts

benma:
It implements a subset of
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md, not
looking at any parameters.

Same as in BTC/LTC, the QR code scanner result parser is forgiving and
accepts anything. The send form validation will catch it if the parsed
result is not a valid recipient address.


thisconnect (approved):
LGTM

PR #1680, opened 2022-03-21
Frontend esc hook

thisconnect:
Added hook to take care about ESC keydown event and fire a callback.

Based on PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/1679

### DIFF
--- a/frontends/web/src/hooks/keyboard.ts
+++ b/frontends/web/src/hooks/keyboard.ts
@@ -1,0 +1,35 @@
+/**
+ * Copyright 2022 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect } from 'react'
+
+/**
+ * useEsc handles ESC key.
+ * gets fired on ESC keydown and executes the provided callback.
+ */
+export const useEsc = (
+    callback: () => void
+) => {
+    useEffect(() => {
+        const handleKeyDown = (e: KeyboardEvent) => {
+            if (e.key === 'Escape') {
+                callback();
+            }
+        }
+        document.addEventListener('keydown', handleKeyDown);
+        return () => document.removeEventListener('keydown', handleKeyDown);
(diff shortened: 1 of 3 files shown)


frontends/web/src/hooks/keyboard.ts
```diff
@@ -0,0 +1,35 @@
...
+) => {
+    useEffect(() => {
+        const handleKeyDown = (e: KeyboardEvent) => {
+            if (e.key === 'Escape') {
+                callback();
+            }
+        }
+        document.addEventListener('keydown', handleKeyDown);
+        return () => document.removeEventListener('keydown', handleKeyDown);
+    }, [callback]);
```
benma, line 34: Is the `[callback]` dep needed? According to the React docs:

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. 

And this effect is sth. you want to run and cleanup once on mount and unmount.

frontends/web/src/routes/account/info/info.tsx
```diff
@@ -15,126 +15,101 @@
...
+}) => {
+    const { t } = useTranslation();
+    const info = useLoad(getInfo(code));
+    const [viewXPub, setViewXPub] = useState<number>(0);
+    const account = accounts.find(({ code: accountCode }) => accountCode === code);
 
-    private handleKeyDown = (e: KeyboardEvent) => {
-        if (e.keyCode === 27) {
-            route(`/account/${this.props.code}`);
+    useEffect(() => {
```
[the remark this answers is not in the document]
thisconnect: I should have made a comment that this branched off of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1679

This commit is that PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/1679

frontends/web/src/hooks/keyboard.ts
thisconnect: the dependency list just tells the useEffect hook to execute (cleanup/unmount and mount again) when the array changes, in this case when the callback changes. 

Unfortunately functions cannot be compared unless it is two references to the same function. I was thinking of caching the function with useCallback but does not really improve anything afaik here and makes it harder to read.

benma: But how can the callback change? It is a function param.

thisconnect: @benma please take a look at this codepen, clicking or pressing ESC counts up and in the codepen's console you should see mount and unmount logs. Now remove `callback` from the dependencyList and test with pressing ESC, you'll see it will only count to 1, because the useEffect in the hook isn't updated after the parent re-rendered.

 https://codepen.io/drugtgkddngme/pen/eYyBgeY?editors=1011

Screen Shot 2022-03-23 at 8 27 39 AM

benma: nice explanation, thanks! ack

benma (approved):
utACK, awesome

PR #1683, opened 2022-03-24
frontend: only show buy button if there is no tx and no incoming tx

thisconnect:
The buy button is currently shown if the balance is zero, but
didn't take into consideration if there is an incoming tx or
if there was transactions in the past.

Changed the conditions on when to show the buy button on the
account overview, taking incoming and old transactions into
account.


benma (approved):
good catch!

PR #1685, opened 2022-04-06
frontend: fix alert user ui text overflow

thisconnect:
The alertUser component recently changed to use view component
internally, this didn't handle long text well, long text was
displayed overflowing the width of the dialog.

Changed styling of the view header and content to break-words
which fixes this issue.

Also changed, that multiline messages do not use div elements
as that is not allowed as child of H1 elements, which is what
ViewHeader uses.

Tested:
- long invalid addresses in electrum settings
- normal stupid simple message signing
- enable passphrase


benma (approved):
utACK, thanks!

PR #1691, opened 2022-04-26
frontend/bitbox02/settings: rewrite to functional component

benma:

PR #1692, opened 2022-04-26
frontend/bitbox02: show root fingerprint

benma:
This helps to identify the wallet and is especially useful when using
the optional passphrase.

Builds on https://github.com/digitalbitbox/bitbox-wallet-app/pull/1691

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -404,6 +404,7 @@
       }
     },
     "hardware": {
+      "rootFingerprint": "Root fingerprint",
       "sdcard": {
         "false": "Not inserted",
         "label": "microSD card",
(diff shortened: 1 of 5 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -692,6 +693,12 @@
         "title": "How to securely pair with your phone"
       }
     },
+    "bitbox02": {
+      "rootFingerprint": {
+        "text": "The root fingerprint is a unique identifier for the wallet that is currently loaded. It can help you distinguish different wallets if you use passphrases.",
```
benma, line 698: @jadzeidan text sounds good?

thisconnect: ping @jadzeidan

jadzeidan: How about:

"The root fingerprint is a unique identifier for the wallet **currently in use**. It can help you distinguish **between** different wallets if you use passphrases."

jadzeidan: Makes sense to me

frontends/web/src/routes/device/bitbox02/settings.tsx
```diff
@@ -129,6 +144,9 @@ export const Settings: FunctionComponent<Props> = ({ deviceID }) => {
                     </Footer>
                 </div>
             </div>
+            <Guide>
+                <Entry key="accountDescription" entry={t('guide.bitbox02.rootFingerprint')} />
+            </Guide>
         </div>
     );
 }
```
thisconnect, line 152: In the comment you mention to move it into account info.

> Now I am thinking though if it might be better to show the fingerprint in the account info along with the xpub. If we show the xpub in descriptor format, the root fingerprint would be part of that anyway

move to account-info?

https://github.com/digitalbitbox/bitbox-wallet-app/pull/1692#discussion_r862725886

thisconnect (commented):
added questions

PR #1693, opened 2022-05-05
update npm deps

thisconnect:
Updated npm dependencies such as chart lib, router and i18next,
or dev deps like react-scripts and types.

Newest QR reader lib couldn't be updated due to missing sourcemaps.

Also cannot update react to 18, due to 3 tests that are using enzyme
that wont work with react 18 and need to be migrated to
testing-library first.

### DIFF
--- a/frontends/web/package.json
+++ b/frontends/web/package.json
@@ -4,27 +4,27 @@
   "version": "0.0.0",
   "devDependencies": {
     "@testing-library/jest-dom": "^5.14.1",
-    "@testing-library/react": "^11.2.7",
-    "@testing-library/user-event": "^12.8.3",
-    "@types/enzyme": "^3.10.10",
-    "@types/jest": "^26.0.24",
-    "@types/node": "^12.20.21",
+    "@testing-library/react": "^12.1.5",
+    "@testing-library/user-event": "^14.1.1",
+    "@types/enzyme": "3.10.12",
+    "@types/jest": "27.5.0",
+    "@types/node": "17.0.31",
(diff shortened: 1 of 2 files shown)


frontends/web/package.json
```diff
@@ -4,27 +4,27 @@
...
   "devDependencies": {
     "@testing-library/jest-dom": "^5.14.1",
-    "@testing-library/react": "^11.2.7",
-    "@testing-library/user-event": "^12.8.3",
-    "@types/enzyme": "^3.10.10",
-    "@types/jest": "^26.0.24",
-    "@types/node": "^12.20.21",
+    "@testing-library/react": "^12.1.5",
+    "@testing-library/user-event": "^14.1.1",
+    "@types/enzyme": "3.10.12",
```
benma, line 9: how do you decide when to use ^ and when not to use it?

benma (approved):

frontends/web/package.json
thisconnect: This is only for some devDependencies, there is a small comment in the readme that for example TypeScript it is ok if it uses a slightly newer patch release

https://github.com/digitalbitbox/bitbox-wallet-app#npm-dependencies

This is just an indication that patches or minor versions are ok too, but the deps are locked anyway.

thisconnect: FYI, if we have the same NPM version and you remove node_modules and package-lock.json and do a new npm install you should get 🤞  the same lock file... until in the meantime some modules got updated.

PR #1695, opened 2022-05-09
frontend: type button component

thisconnect:
Migrated Button and ButtonLink components to TypeScript.

### DIFF
--- a/frontends/web/src/components/forms/button.tsx
+++ b/frontends/web/src/components/forms/button.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2021 Shift Crypto AG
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/forms/button.tsx
```diff
@@ -40,26 +47,25 @@ export function ButtonLink({
         ], className
     ].join(' ');
 
-    if (disabled) {
+    if (disabled === true) {
```
benma, line 50: why `=== true`? looks like the way it was before works the same way.

benma (approved):
utACK

frontends/web/src/components/forms/button.tsx
thisconnect: No reason, actually I prefer `if (disabled) {` and will change it back

PR #1696, opened 2022-05-10
frontend: changed setdevicename to functioncomponent

thisconnect:
- migrated Button and ButtonLink components to TypeScript
    - first time used ComponentPropsWithoutRef<'button'> instead of
      JSX.IntrinsicElements['button'] to be less verbose
    - changed ButtonLink internally to make typing simpler, it used
      to return a Button component instance if disabled, changed to
      return just a button element instead
    - fixed named import/export


frontends/web/src/routes/device/bitbox02/setdevicename.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const { t } = useTranslation();
+
+    const [active, setActive] = useState(false);
+    const [currentName, setCurrentName] = useState(deviceName);
+    const [name, setName] = useState('');
+    const [inProgress, setInProgress] = useState(false);
+
+    const inputRef = useRef<HTMLInputElement>(null);
+
+    const updateName = () => {
```
benma, line 45: this might be a good candidate for async/await syntax, but it's also fine as it is.

benma (approved):
tACK

frontends/web/src/routes/device/bitbox02/setdevicename.tsx
thisconnect: I'll check, the only thing I don't like (or am not sure exactly) is for `.catch` to use `try {} catch(e) {} ` if the response fails and have to check the finally thing.

thisconnect: IIRC there was an advantage over promises when using await/async, if I remember correctly better stacktraces.

I'll change to await/async ack? 

```diff
diff --git a/frontends/web/src/routes/device/bitbox02/setdevicename.tsx b/frontends/web/src/routes/device/bitbox02/setdevicename.tsx
index 8f47c4b4..da411182 100644
--- a/frontends/web/src/routes/device/bitbox02/setdevicename.tsx
+++ b/frontends/web/src/routes/device/bitbox02/setdevicename.tsx
@@ -42,16 +42,18 @@ export const SetDeviceName: FunctionComponent<Props> = ({
 
     const inputRef = useRef<HTMLInputElement>(null);
 
-    const updateName = () => {
+    const updateName = async () => {
         setInProgress(true);
-        setDeviceName(deviceID, name)
-            .then(() => getDeviceInfo(deviceID))
-            .then(({ name }) => setCurrentName(name))
-            .catch(() => alertUser('Device name could not be set'))
-            .finally(() => {
-                setActive(false);
-                setInProgress(false);
-            });
+        try {
+            await setDeviceName(deviceID, name);
+            const { name: newDeviceName } = await getDeviceInfo(deviceID);
+            setCurrentName(newDeviceName);
+        } catch (e) {
+            alertUser('Device name could not be set')
+        } finally {
+            setActive(false);
+            setInProgress(false);
+        }
     };
 
     return (
```

benma: lgtm

PR #1697, opened 2022-05-12
frontend: disable chart filter buttons if chartdatamissing

thisconnect:
The chart filter buttons to change the chart to display
week, month, year and all is currently only checking
if the chart contains any data and the total is not 0.

But if the chart is updating the buttons are already active
even though the chart has not loaded and is displaying
updating data.

Changed to check for chartDataMissing and keep the buttons
disabled until everything is up to date.

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -101,8 +101,8 @@ class Chart extends Component<Props, State> {
         }
     }
 
-    private hasData = () => {
-        return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length;
+    private hasData = (): boolean => {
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -101,8 +101,8 @@ class Chart extends Component<Props, State> {
         }
     }
 
-    private hasData = () => {
-        return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length;
+    private hasData = (): boolean => {
+        return this.props.data.chartDataDaily && !!this.props.data.chartDataDaily.length;
```
benma, line 105: wouldn't `... length > 0` be a bit more natural? :smile:

benma (approved):
lgtm

PR #1700, opened 2022-05-17
Makefile: fix gomobile setup

benma:
The current way with turning off the modules fails with:

```
gomobile: go install golang.org/x/mobile/cmd/gobind@latest failed: exit status 1
go: modules disabled by GO111MODULE=off; see 'go help modules'
```

According to the docs at
https://pkg.go.dev/golang.org/x/mobile/cmd/gomobile, `go install` is
to be used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,8 @@ envinit:
 	GO111MODULE=on go get -u github.com/vektra/mockery/...
 	GO111MODULE=off go get -u github.com/matryer/moq
 	GO111MODULE=off go get golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -26,8 +26,8 @@ envinit:
 	GO111MODULE=on go get -u github.com/vektra/mockery/...
 	GO111MODULE=off go get -u github.com/matryer/moq
 	GO111MODULE=off go get golang.org/x/tools/cmd/goimports
```
benma, line 28: maybe the ones above should also use `go install` instead of `go get` :thinking:

Beerosagos (approved):

PR #1702, opened 2022-05-17
go: update Go to 1.18

benma:
Go 1.18 by default enables adding VCS info to the binary by querying
`git`. See https://tip.golang.org/doc/go1.18 ("VCS"). This fails in
the GitHub CI with the following error:

```
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```

It turns out that `git status` etc. on Github CI fails with:

```
fatal: unsafe repository ('/opt/go/src/github.com/digitalbitbox/bitbox-wallet-app' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /opt/go/src/github.com/digitalbitbox/bitbox-wallet-app
```

The issue is fixed by following this instruction.

See also: https://github.com/actions/checkout/issues/760#issuecomment-1097501613

golangci: remove newly added linters

Updated golangci-lint at the same time, so we only have to rebuild the
Docker image once instead of twice.

We updated from v1.43.0 to v1.46.1, which enabled a lot of new
linters.

In the future we should disable all linters by default and whitelist
the linters we want, instead of doing the opposite.

The linters disabled in this commit should be checked individually to
decide which to enable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ env:
   # https://docs.github.com/en/packages/guides/pushing-and-pulling-docker-images
   #
   # Keep this in sync with default in scripts/travis-ci.sh.
-  CI_IMAGE: ghcr.io/digitalbitbox/bitbox-wallet-app-ci:11
+  CI_IMAGE: ghcr.io/digitalbitbox/bitbox-wallet-app-ci:14
   TRAVIS_BUILD_DIR: ${{github.workspace}}
 
 jobs:
@@ -115,7 +115,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.18.x
       - name: Build macOS app
         run: >
           ./scripts/travis-ci.sh qt-osx;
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ env:
   # https://docs.github.com/en/packages/guides/pushing-and-pulling-docker-images
   #
   # Keep this in sync with default in scripts/travis-ci.sh.
-  CI_IMAGE: ghcr.io/digitalbitbox/bitbox-wallet-app-ci:11
+  CI_IMAGE: ghcr.io/digitalbitbox/bitbox-wallet-app-ci:14
```
benma, line 24: fyi I built this image using `make dockerinit` and pushed it here

scripts/travis-ci.sh
```diff
@@ -14,16 +14,19 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
     # Which docker image to use to run the CI. Defaults to Docker Hub.
     # Overwrite with CI_IMAGE=docker/image/path environment variable.
     # Keep this in sync with .github/workflows/ci.yml.
-    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:13}"
+    : "${CI_IMAGE:=shiftcrypto/bitbox-wallet-app:14}"
```
benma, line 17: and here

benma:
(I updated the commit msg, put the wrong version info about the golangci update before)

Beerosagos (approved):

PR #1704, opened 2022-05-18
backend/btc: make some client functions synchronous

benma:
The async style is hard to read and reason about and complicates error handling, for practically no benefit.


benma:
@Beerosagos I pushed a new commit, please check

Beerosagos (approved):
tACK!

PR #1708, opened 2022-06-02
frontend: add fitContent option to the view component

thisconnect:
If a View has a success icon by `<ViewContent withIcon="success"`
but not a lot of other content sometimes the ViewButtons are not
in the visible area on smaller screens.

Refactored the unused `top` option into `fitContent` to indicate
that the View should try to fit all content into the visible area.
With `fitContent` the big success icon will shrink if there is not
enough available space.

Note the footer could still be outside of the visible area and only
visible when scrolling down.

Also added some docstrings to the View component.

### DIFF
--- a/frontends/web/src/components/view/view.tsx
+++ b/frontends/web/src/components/view/view.tsx
@@ -23,9 +23,9 @@ import style from './view.module.css';
 
 type ViewProps = {
     dialog?: boolean;
+    fitContent?: boolean;
     fullscreen?: boolean;
     minHeight?: string;
-    top?: boolean;
     onClose?: () => void;
     position?: 'fill' | '';
     textCenter?: boolean;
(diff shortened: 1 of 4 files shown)


frontends/web/src/components/view/view.tsx
```diff
@@ -23,20 +23,31 @@ import style from './view.module.css';
...
-    top?: boolean;
     onClose?: () => void;
     position?: 'fill' | '';
     textCenter?: boolean;
     width?: string;
     withBottomBar?: boolean;
 }
 
+/**
+ * View component is used as a container component to wrap ViewHeader, ViewContent and ViewButtons
```
benma, line 37: did we use this docstring format in the past or is it the first time?

benma (approved):
utACK

frontends/web/src/components/view/view.tsx
thisconnect: We used it in the past

Screen Shot 2022-06-07 at 12 17 24 PM

thisconnect: https://github.com/digitalbitbox/bitbox-wallet-app/blob/2686682a73e28c526cbd9a80beab7e6430f174c6/frontends/web/src/decorators/subscribe.tsx#L25-L32

frontends/web/src/components/view/view.tsx
```diff
@@ -23,20 +23,31 @@ import style from './view.module.css';
...
+ * @param dialog wether to render the view as a dialog
+ * @param fitContent tries to squeze the whole view into the visible area, if true the icon specified through ViewContent withIcon will shrink if the visible area is bigger than the window height
```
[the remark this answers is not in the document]
thisconnect: https://github.com/digitalbitbox/bitbox-wallet-app/compare/e80d52763e46ed2e5cfbef5ccc8e71ff6b651499..6fae0143e9e23dc3b3df8b4058fc45668a7341f9

PR #1718, opened 2022-06-20
frontend: increase minimum height of passphrase intro content

thisconnect:
The passphrase introduction explains what a passphrase is through
multiple steps. The second step is getting another line in a
different commit as that is pulled form locize.

This change increased the minimum height of the content area so
that all steps use the same height.


thisconnect:
ALso: updated locize to add the new sentense.

benma (approved):
utACK

PR #1721, opened 2022-07-04
frontend: add eslint rule to make semicolons mandatory

thisconnect:
We do not use ASI (automatic semicolon insertion).

Added an eslint rule to enforce semicolons and antoher to disallow
unnecessary semicolons after TypeScript interfaces or function
declarations.

This commit's changes were autogenerated with eslint fix, the
command run was `make webfix`.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -16,7 +16,7 @@
  */
 
 import { Component } from 'react';
-import { AppRouter } from './routes/router'
+import { AppRouter } from './routes/router';
 import { getAccounts, IAccount } from './api/account';
 import { syncAccountsList } from './api/accountsync';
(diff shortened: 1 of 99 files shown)


thisconnect:
FYI to only lint the frontend use `make weblint`.

frontends/web/src/app.tsx
```diff
@@ -60,7 +60,7 @@ class App extends Component<Props, State> {
     if (panelStore.state.activeSidebar) {
       toggleSidebar();
     }
-  }
+  };
```
thisconnect, line 63: At first I was confused why it added a semicolon here, but found this https://github.com/prettier/prettier/issues/1444#issuecomment-297835274

> class properties - unlike method declarations in classes - do have semicolons after them.

Beerosagos (approved):
tACK!

PR #1723, opened 2022-07-05
frontend: type bitbox01 device info api

thisconnect:
There is already api/devices.ts but this getDeviceInfo is specific
to BitBox01, so this change added api/bitbox01.ts.

Copied types from DeviceInfo struct in:
- [backend/devices/bitbox/device.go](https://github.com/digitalbitbox/bitbox-wallet-app/blob/197c8b7d4113d71bf717d391a70eda3b5f9b4cac/backend/devices/bitbox/device.go#L97-L112)

### DIFF
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2022 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


frontends/web/src/routes/account/account.tsx
```diff
@@ -138,13 +139,8 @@ class Account extends Component<Props, State> {
     Promise.all(this.deviceIDs(this.props.devices).map(deviceID => {
       switch (this.props.devices[deviceID]) {
       case 'bitbox':
-        return apiGet(`devices/${deviceID}/info`)
-          .then(info => {
-            if (!info) {
-              return false;
-            }
```
thisconnect, line 145: I am not 100% sure if the returned object can be undefined, seems it was added in https://github.com/digitalbitbox/bitbox-wallet-app/commit/06739c2fffdb39333fe1dc49a107f2c455b874d3

Beerosagos: I don't have a bb01 to make a real test on the device, but I made some checks and tested a bit using similar calls on bb02. The endpoint should always return a valid object or an error (which would be handled by `apiGet` error handler). If for any reason the call requests an invalid ID it gets a 404 error and the `then` code is not executed. I think this is safe enough.

PR #1726, opened 2022-07-18
frontend: better l10n of numbers

ide:
Use the locale-specific formatting for currency numbers instead of always using `'` as the thousandths separator. This is done using `Intl.NumberFormatter` with the user's chosen language. Note: this uses plain number formatting instead of Intl's currency formatting because Intl's currency formatting attaches a currency symbol ($, €), while the current UI doesn't show the symbol.

Currency rendering in portfolio using the Testnet testing data:
German:
de

English:
en

However, some strings in the account summary are still not localized because they come from the API:

account summary

### DIFF
--- a/frontends/web/src/components/headerssync/headerssync.tsx
+++ b/frontends/web/src/components/headerssync/headerssync.tsx
@@ -29,7 +29,7 @@ interface Props {
 }
 
 export const HeadersSync: FunctionComponent<Props> = ({ coinCode }) => {
-  const { t } = useTranslation();
+  const { i18n, t } = useTranslation();
   const status = useSubscribe(subscribeCoinHeaders(coinCode));
   const [hidden, setHidden] = useState<boolean>(false);
   const mounted = useMountedRef();
@@ -47,12 +47,7 @@ export const HeadersSync: FunctionComponent<Props> = ({ coinCode }) => {
   const total = status.targetHeight - status.tipAtInitTime;
   const value = 100 * (status.tip - status.tipAtInitTime) / total;
   const loaded = !total || value >= 100;
-  let formatted = status.tip.toString();
-  let position = formatted.length - 3;
-  while (position > 0) {
-    formatted = formatted.slice(0, position) + '\'' + formatted.slice(position);
-    position = position - 3;
-  }
+  const formatted = new Intl.NumberFormat(i18n.language).format(status.tip);
 
   return (
     <div className={style.syncContainer}>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Localize numbers using the user's chosen language and system region
 - Render number of blocks scanned and percentage progress using fixed-width digits for a more stable UI
 - Transaction Details: show fiat value at time of transaction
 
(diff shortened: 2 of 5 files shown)


benma:
Thanks a lot! 

Could you please split the two changes (header sync and currency formatting) into two commits, or even two PRs? It's better if commits do one change at a time.

Please also update the CHANGELOG.md.

In the docs:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat

I see it takes a language code as an argument. Did you look into that? Would it make sense to provide the current app language here?

ide:
> I see it takes a language code as an argument. Did you look into that? Would it make sense to provide the current app language here?

This is a nice idea and even better than using the system's locale. If you have a Swiss-based computer, could you please help test that this change uses the system's region?

Specifically, the language switcher in the app allows changing the language ("en", "de") but not the region ("US", "DE"). This matters when the region affects how numbers are formatted:
```
> Intl.NumberFormat('en-US').format(1234.56)
'1,234.56'
> Intl.NumberFormat('en-CH').format(1234.56)
'1’234.56'
``` 

If you start with your system's default locale ("??-CH"), then choose a non-English language ("de"), and then choose English ("en"), I would expect the browser to infer that "en-CH" is your desired locale and format numbers like "1’234.56".

Beerosagos:
Hi @ide, thanks for your contribution!

I agree with you that user locale should be exploited in number formatting, but I am afraid that the approach you proposed is not feasible. As you noticed, part of the formatting is done in the backend, because this allows to avoid loss of precision, since depending on the coins we may need up to 18 fractional digits! We think that it would be better to refactor the code, moving all the formatting in the backend and handling user locale there.

If you are willing to contribute further we can include you and discuss the details of the implementation.

ide:
@Beerosagos I think I won't be able to make a larger contribution and I'll go ahead and close this PR. Feel free to take some of the code if you find it useful. `Intl` may still make sense for formatting fiat currency values (`{ format: 'currency', currency: 'CHF' }`) because the API understands many formatting conventions, like that USD always has two decimal digits and JPY has none.

frontends/web/src/components/headerssync/headerssync.tsx
```diff
@@ -47,12 +47,7 @@ export const HeadersSync: FunctionComponent<Props> = ({ coinCode }) => {
   const total = status.targetHeight - status.tipAtInitTime;
   const value = 100 * (status.tip - status.tipAtInitTime) / total;
   const loaded = !total || value >= 100;
-  let formatted = status.tip.toString();
-  let position = formatted.length - 3;
-  while (position > 0) {
-    formatted = formatted.slice(0, position) + '\'' + formatted.slice(position);
-    position = position - 3;
-  }
+  const formatted = new Intl.NumberFormat(i18n.language).format(status.tip);
```
thisconnect, line 51: thanks @ide  ❤️ 

I put this in a separate PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/1730, but I am not sure how to make you co-author in git.


https://github.com/digitalbitbox/bitbox-wallet-app/pull/1730

ide: No worries =) Glad I could be of some help

PR #1727, opened 2022-07-19
frontend: kerning of progress numbers using tabular-nums

ide:
When showing rapidly changing numbers like the number of blocks scanned, tabular numbers make the UI more stable. For example, a number like "11" is narrow while "12" is wider. By using tabular numbers, the width of the UI stays stable when it goes from rendering "11" to "12".

I added `font-variant: tabular-nums` to a few progress indicators. There is one indicator in account.tsx (`initializingSpinnerText`) that I didn't modify since it would require bigger changes to the Spinner component, such as allowing a parent to pass in JSX children instead of supporting only the `text` prop.

Block syncing progress:

https://user-images.githubusercontent.com/379606/179463028-19867286-09da-43d5-9b5e-39cacaec644c.mp4

Account summary scanning progress:

accountsummary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Render number of blocks scanned and percentage progress using fixed-width digits for a more stable UI
 - Transaction Details: show fiat value at time of transaction
 
 ## 4.34.0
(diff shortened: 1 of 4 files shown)


benma (approved):
Thanks!

Strange, I checked the size of the sync label using the firefox developer tool on master (without your change). The computed width stayed fixed when going from 11% to 12%:
 
image

image

I guess it depends on the actual font being used.

Anyway, LGTM :+1:

PR #1730, opened 2022-07-25
frontend: language-sensitive number formatting for blocks synced

thisconnect:
Use the locale-specific formatting for currency numbers instead of
always using a singlequote as the thousandths separator. This is
done using Intl.NumberFormatter with the user's default locale.


benma (approved):
utACK

PR #1731, opened 2022-07-26
Frontend split receive component into bitbox02 and legacy (bitbox01)

thisconnect:
This has some duplicated code, but it will allow to completely revamp the receive view for BitBox02 and keep the legacy BitBox01 receive flow as it is.

### DIFF
--- a/frontends/web/src/routes/account/receive/receive.tsx
+++ b/frontends/web/src/routes/account/receive/receive.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2022 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,7 +20,6 @@ import { useTranslation } from 'react-i18next';
 import { useLoad } from '../../../hooks/api';
 import { useEsc } from '../../../hooks/keyboard';
 import * as accountApi from '../../../api/account';
-import { TDevices } from '../../../api/devices';
 import { route } from '../../../utils/route';
 import { getScriptName, isEthereumBased } from '../utils';
 import { alertUser } from '../../../components/alert/Alert';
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -284,20 +275,10 @@ export const Receive: FunctionComponent<Props> = ({
...
-          </>
-        )}
-      </Guide>
+      <ReceiveGuide
+        hasMultipleAddresses={currentAddresses && currentAddresses.length > 1}
+        hasDifferentFormats={receiveAddresses && receiveAddresses.length > 1}
+      />
     </div>
   );
 };
```
thisconnect, line 284: maybe there could be other parts removed that are only relelvant to BB01 but I am not sure, cc @benma

frontends/web/src/routes/account/receive/receive-legacy.tsx
```diff
@@ -0,0 +1,303 @@
+/**
```
benma, line 1: i'd name the file `receive-bb01.tsx` for clarity.

frontends/web/src/routes/account/receive/receive-legacy.tsx
```diff
@@ -64,7 +62,6 @@ const getIndexOfMatchingScriptType = (
 export const Receive: FunctionComponent<Props> = ({
   accounts,
   code,
-  devices,
```
benma: the devices are still passed from receive/index.ts, right? typescript doesn't mind?

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -76,9 +70,8 @@ export const Receive: FunctionComponent<Props> = ({
   const [currentAddresses, setCurrentAddresses] = useState<accountApi.IReceiveAddress[]>();
   const [currentAddressIndex, setCurrentAddressIndex] = useState<number>(0);
 
-  const device = deviceIDs.length ? devices[deviceIDs[0]] : undefined;
   const account = accounts.find(({ code: accountCode }) => accountCode === code);
-  const verifyLabel = useVerfiyLabel(device);
+  const verifyLabel = useVerfiyLabel('bitbox02');
```
benma, line 73: could remove that function and inline `t('receive.verifyBitBox02')`

benma (approved):
utACK

frontends/web/src/routes/account/receive/receive-legacy.tsx
thisconnect: I believe the short answer is it matches so TypeScript is happy.

https://stackoverflow.com/questions/49580725/is-it-possible-to-restrict-typescript-object-to-contain-only-properties-defined

https://github.com/Microsoft/TypeScript/issues/12936

thisconnect: There is an eslint rule that would help, "react/no-unused-prop-types": "error" and catches a few cases, but unfortunately it has a few cases with the load decorator where it fails to recognise that those props are actually used and for some reason it does not catch this case :/

 I'll open a separate PR

thisconnect: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1733

Beerosagos:
tACK! :+1:

PR #1732, opened 2022-07-26
frontend: use translation hook

thisconnect:
Changed to useTranslation hook instead of passing the 't' function
to function components.


benma (approved):
utACK

PR #1733, opened 2022-07-28
frontend: fix react/no-unused-prop-types

thisconnect:
Eslint rule react/no-unused-prop-types found a few issues, but also
false positives related to load decorator or custom store.

This change removes unused props, but does not permanently add the
rule to eslint.


benma (approved):
very nice! would be awesome if we could add it to the linter list and disable it manually for the few false positives.

thisconnect:
@benma added 2 commits, and noticed that there was one that is not a false positive.

PTAL

benma (approved):

benma (commented):
:+1:

PR #1735, opened 2022-08-02
Formatting refactor

Beerosagos:
This PR (WIP):
- Fixes the BTC 2 digits backend formatting bug
- Proposes a draft solution to centralize formatting handling in the backend, exploiting higher precision when available

I still see room for some improvements in FromUnit implementation precision.

See commit messages for details.

Once agreed on this first draft, the next step is to refactor the frontend part, moving all the occasional formatting done there into the backend.

### DIFF
--- a/backend/chart.go
+++ b/backend/chart.go
@@ -41,8 +41,15 @@ func (backend *Backend) allCoinCodes() []string {
 
 // ChartEntry is one data point in the chart timeseries.
 type ChartEntry struct {
-	Time  int64   `json:"time"`
-	Value float64 `json:"value"`
+	Time           int64   `json:"time"`
+	Value          float64 `json:"value"`
+	FormattedValue string  `json:"formattedValue"`
+}
+
+// RatChartEntry exploits composition to extend ChartEntry and save high precision values.
+type RatChartEntry struct {
+	ChartEntry
+	RatValue *big.Rat
 }
 
 // Chart has all data needed to show a time-based chart of their assets to the user.
@@ -59,6 +66,8 @@ type Chart struct {
 	// Current total value of all assets in the fiat currency. Nil if missing (this is independent
 	// of `DataMissing`).
 	Total *float64 `json:"chartTotal"`
+	// ChartTotal formatted for frontend visualization
+	FormattedTotal string `json:"formattedChartTotal"`
 	// Only valid if DataMissing is false
 	IsUpToDate bool `json:"chartIsUpToDate"`
 }
@@ -68,7 +77,7 @@ func (backend *Backend) addChartData(
 	fiat string,
 	coinDecimals *big.Int,
 	timeseries []accounts.TimeseriesEntry,
-	chartEntries map[int64]ChartEntry,
+	chartEntries map[int64]RatChartEntry,
 ) {
 	for _, e := range timeseries {
 		price := backend.RatesUpdater().HistoricalPriceAt(
@@ -76,18 +85,23 @@ func (backend *Backend) addChartData(
 			fiat,
 			e.Time)
 		timestamp := e.Time.Unix()
-		ChartEntry := chartEntries[timestamp]
+		chartEntry := chartEntries[timestamp]
 
-		ChartEntry.Time = timestamp
-		fiatValue, _ := new(big.Rat).Mul(
+		chartEntry.Time = timestamp
+		fiatValue := new(big.Rat).Mul(
 			new(big.Rat).SetFrac(
 				e.Value.BigInt(),
 				coinDecimals,
 			),
 			new(big.Rat).SetFloat64(price),
-		).Float64()
-		ChartEntry.Value += fiatValue
-		chartEntries[timestamp] = ChartEntry
+		)
+
+		if chartEntry.RatValue == nil {
+			chartEntry.RatValue = new(big.Rat).Set(fiatValue)
+		} else {
+			chartEntry.RatValue.Add(fiatValue, chartEntry.RatValue)
+		}
+		chartEntries[timestamp] = chartEntry
 	}
 }
 
@@ -98,8 +112,8 @@ func (backend *Backend) ChartData() (*Chart, error) {
 	chartDataMissing := false
 
 	// key: unix timestamp.
-	chartEntriesDaily := map[int64]ChartEntry{}
-	chartEntriesHourly := map[int64]ChartEntry{}
+	chartEntriesDaily := map[int64]RatChartEntry{}
+	chartEntriesHourly := map[int64]RatChartEntry{}
 
 	fiat := backend.Config().AppConfig().Backend.MainFiat
 	// Chart data until this point in time.
@@ -231,11 +245,17 @@ func (backend *Backend) ChartData() (*Chart, error) {
 
 	}
 
-	toSortedSlice := func(s map[int64]ChartEntry) []ChartEntry {
+	toSortedSlice := func(s map[int64]RatChartEntry, fiat string) []ChartEntry {
 		result := make([]ChartEntry, len(s))
 		i := 0
+		// Discard the RatValue, which is not used anymore
 		for _, entry := range s {
-			result[i] = entry
+			floatValue, _ := entry.RatValue.Float64()
+			result[i] = ChartEntry{
+				Time:           entry.Time,
+				Value:          floatValue,
+				FormattedValue: coin.FormatAsCurrency(entry.RatValue, fiat),
+			}
 			i++
 		}
 		sort.Slice(result, func(i, j int) bool { return result[i].Time < result[j].Time })
@@ -248,8 +268,9 @@ func (backend *Backend) ChartData() (*Chart, error) {
 		if isUpToDate && !currentTotalMissing {
 			total, _ := currentTotal.Float64()
 			result = append(result, ChartEntry{
-				Time:  time.Now().Unix(),
-				Value: total,
+				Time:           time.Now().Unix(),
+				Value:          total,
+				FormattedValue: coin.FormatAsCurrency(currentTotal, fiat),
 			})
 		}
 
@@ -272,16 +293,19 @@ func (backend *Backend) ChartData() (*Chart, error) {
 	}
 
 	var chartTotal *float64
+	var formattedChartTotal string
 	if !currentTotalMissing {
 		tot, _ := currentTotal.Float64()
 		chartTotal = &tot
+		formattedChartTotal = coin.FormatAsCurrency(currentTotal, fiat)
 	}
 	return &Chart{
-		DataMissing: chartDataMissing,
-		DataDaily:   toSortedSlice(chartEntriesDaily),
-		DataHourly:  toSortedSlice(chartEntriesHourly),
-		Fiat:        fiat,
-		Total:       chartTotal,
-		IsUpToDate:  isUpToDate,
+		DataMissing:    chartDataMissing,
+		DataDaily:      toSortedSlice(chartEntriesDaily, fiat),
+		DataHourly:     toSortedSlice(chartEntriesHourly, fiat),
+		Fiat:           fiat,
+		Total:          chartTotal,
+		FormattedTotal: formattedChartTotal,
+		IsUpToDate:     isUpToDate,
 	}, nil
 }
--- a/backend/coins/btc/coin.go
+++ b/backend/coins/btc/coin.go
@@ -168,6 +168,13 @@ func (coin *Coin) ToUnit(amount coinpkg.Amount, isFee bool) float64 {
 	return result
 }
 
+// SetAmount implements coinpkg.Coin.
+func (coin *Coin) SetAmount(amount *big.Rat, isFee bool) coinpkg.Amount {
+	satsAmount := new(big.Rat).Mul(amount, new(big.Rat).SetFloat64(unitSatoshi))
+	intSatsAmount, _ := new(big.Int).SetString(satsAmount.FloatString(0), 0)
+	return coinpkg.NewAmount(intSatsAmount)
+}
+
 // Blockchain connects to a blockchain backend.
 func (coin *Coin) Blockchain() blockchain.Interface {
 	return coin.blockchain
--- a/backend/coins/coin/conversions.go
+++ b/backend/coins/coin/conversions.go
@@ -8,8 +8,13 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/rates"
 )
 
-func formatAsCurrency(amount float64) string {
-	formatted := strconv.FormatFloat(amount, 'f', 2, 64)
+func formatAsCurrency(amount float64, currency string) string {
+	var formatted string
+	if currency == "BTC" {
+		formatted = strings.TrimRight(strconv.FormatFloat(amount, 'f', 8, 64), "0")
+	} else {
+		formatted = strconv.FormatFloat(amount, 'f', 2, 64)
+	}
 	position := strings.Index(formatted, ".") - 3
 	for position > 0 {
 		formatted = formatted[:position] + "'" + formatted[position:]
@@ -27,7 +32,7 @@ func Conversions(amount Amount, coin Coin, isFee bool, ratesUpdater *rates.RateU
 		float := coin.ToUnit(amount, isFee)
 		conversions = map[string]string{}
 		for key, value := range rates[unit] {
-			conversions[key] = formatAsCurrency(float * value)
+			conversions[key] = formatAsCurrency(float*value, key)
 		}
 	}
 	return conversions
@@ -46,7 +51,7 @@ func ConversionsAtTime(amount Amount, coin Coin, isFee bool, ratesUpdater *rates
 			if value == 0 {
 				conversions[fiat] = ""
 			} else {
-				conversions[fiat] = formatAsCurrency(float * value)
+				conversions[fiat] = formatAsCurrency(float*value, fiat)
 			}
 		}
 	}
--- a/backend/coins/coin/conversions.go
+++ b/backend/coins/coin/conversions.go
@@ -1,19 +1,20 @@
 package coin
 
 import (
-	"strconv"
+	"math/big"
 	"strings"
 	"time"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/rates"
 )
 
-func formatAsCurrency(amount float64, currency string) string {
+// FormatAsCurrency handles formatting for currencies.
+func FormatAsCurrency(amount *big.Rat, coinUnit string) string {
 	var formatted string
-	if currency == "BTC" {
-		formatted = strings.TrimRight(strconv.FormatFloat(amount, 'f', 8, 64), "0")
+	if coinUnit == "BTC" {
+		formatted = strings.TrimRight(strings.TrimRight(amount.FloatString(8), "0"), ".")
 	} else {
-		formatted = strconv.FormatFloat(amount, 'f', 2, 64)
+		formatted = amount.FloatString(2)
 	}
 	position := strings.Index(formatted, ".") - 3
 	for position > 0 {
@@ -29,10 +30,10 @@ func Conversions(amount Amount, coin Coin, isFee bool, ratesUpdater *rates.RateU
 	rates := ratesUpdater.LatestPrice()
 	if rates != nil {
 		unit := coin.Unit(isFee)
-		float := coin.ToUnit(amount, isFee)
 		conversions = map[string]string{}
 		for key, value := range rates[unit] {
-			conversions[key] = formatAsCurrency(float*value, key)
+			conversion := new(big.Rat).Mul(new(big.Rat).SetFloat64(coin.ToUnit(amount, isFee)), new(big.Rat).SetFloat64(value))
+			conversions[key] = FormatAsCurrency(conversion, key)
 		}
 	}
 	return conversions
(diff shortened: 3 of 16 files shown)


Beerosagos:
@benma during our last eng sync you proposed to find a solution to avoid having currency/coins string around in the code, like here: https://github.com/Beerosagos/bitbox-wallet-app/blob/formatting-refactor/backend/coins/coin/conversions.go#L14

I made a check and I could not find a struct with all the available fiats in the backend. I noticed that supported currencies are listed in frontend `rates.tsx' (https://github.com/Beerosagos/bitbox-wallet-app/blob/formatting-refactor/frontends/web/src/components/rates/rates.tsx#L41) and shared with backend only when activated by the user. 

I would suggest to open a separated issue/PR to move the list of available currencies in the backend, behind a dedicated endpoint. WDYT?

Beerosagos:
@benma after our talk about float precision I decided to make some tests and verify how much we can relay on `float64` to correctly handle coins and fiats, since it is currently used to compute conversions in the backend (e.g. https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/coin/conversions.go#L30).

From this playground https://go.dev/play/p/DAU9m6ybqKh it seems that the 18 digits precision required by ETH coins is not completely satisfied, also with relatively small amounts. `big.Float` seems to have a better behavior. 

I would suggest to update conversion and formatting functions to use big.Float or big.Rat to reach a better precision. 
In case you agree, I am not sure if this should be included in this PR, or maybe it is better to make a dedicated one. 
WDYT?

backend/handlers/handlers.go
```diff
@@ -666,19 +666,19 @@ func (handlers *Handlers) getConvertFromFiatHandler(r *http.Request) (interface{
...
 	}
-	unit := coin.Unit(isFee)
+	unit := currentCoin.Unit(isFee)
 	switch unit { // HACK: fake rates for testnet coins
 	case "TBTC", "TLTC", "TETH", "RETH":
 		unit = unit[1:]
 	}
 	rate := handlers.backend.RatesUpdater().LatestPrice()[unit][from]
-	result := 0.0
+	result := coin.NewAmount(big.NewInt(0))
```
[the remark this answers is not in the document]
Beerosagos: Cool, didn't notice it! I'll change it :+1:

frontends/web/package-lock.json
```diff
@@ -1,16171 +1,8 @@
 {
   "name": "bitboxapp",
   "version": "0.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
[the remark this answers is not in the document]
Beerosagos: Sorry! I'll revert and look into that

backend/coins/coin/conversions.go
```diff
@@ -27,7 +32,7 @@ func Conversions(amount Amount, coin Coin, isFee bool, ratesUpdater *rates.RateU
 		float := coin.ToUnit(amount, isFee)
 		conversions = map[string]string{}
 		for key, value := range rates[unit] {
-			conversions[key] = formatAsCurrency(float * value)
+			conversions[key] = formatAsCurrency(float*value, key)
```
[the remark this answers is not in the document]
Beerosagos: I'll check that, thx

backend/coins/btc/coin.go
```diff
@@ -168,6 +168,13 @@ func (coin *Coin) ToUnit(amount coinpkg.Amount, isFee bool) float64 {
 	return result
 }
 
+// FromUnit implements coinpkg.Coin.
+func (coin *Coin) FromUnit(floatAmount float64, isFee bool) coinpkg.Amount {
```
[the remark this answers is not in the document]
Beerosagos: I introduced this change not to improve precision, but to exploit the `formatAmount` method for coins and avoid "wild" formatting around. Ok for making it a `ParseAmount` method

backend/coins/coin/conversions.go
```diff
@@ -32,7 +33,7 @@ func Conversions(amount Amount, coin Coin, isFee bool, ratesUpdater *rates.RateU
 		float := coin.ToUnit(amount, isFee)
 		conversions = map[string]string{}
 		for key, value := range rates[unit] {
-			conversions[key] = formatAsCurrency(float*value, key)
+			conversions[key] = FormatAsCurrency(big.NewFloat(float*value), key)
```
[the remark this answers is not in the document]
Beerosagos: This is related to my previous comment (https://github.com/digitalbitbox/bitbox-wallet-app/pull/1735#issuecomment-1203994765), as agreed in our last eng sync I'll use a `Big.Rat`

backend/coins/btc/coin.go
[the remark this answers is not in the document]
Beerosagos: While working on the code I realized that I had to convert from string to number to obtain the coin amount (fiat amount input / exchange rate) in `convertFromFiat` handler (see https://github.com/Beerosagos/bitbox-wallet-app/blob/formatting-refactor/backend/handlers/handlers.go#L684)

So I thought it was better to make a `SetAmount` method already taking in input a Big.Rat, instead of a string. WDYT?

backend/handlers/handlers.go
```diff
@@ -631,54 +631,78 @@ func (handlers *Handlers) getRatesHandler(_ *http.Request) (interface{}, error)
 func (handlers *Handlers) getConvertToFiatHandler(r *http.Request) (interface{}, error) {
 	from := r.URL.Query().Get("from")
 	to := r.URL.Query().Get("to")
-	amount := r.URL.Query().Get("amount")
-	amountAsFloat, err := strconv.ParseFloat(amount, 64)
-	if err != nil {
+	coin := r.URL.Query().Get("amount")
```
benma, line 634: `amount` is a better variable name? coin usually refers to a `coin.Coin` instance.

backend/handlers/handlers.go
```diff
@@ -631,54 +631,78 @@ func (handlers *Handlers) getRatesHandler(_ *http.Request) (interface{}, error)
...
 		"success":    true,
-		"fiatAmount": strconv.FormatFloat(amountAsFloat*rate, 'f', 2, 64),
+		"fiatAmount": coinpkg.FormatAsCurrency(amount, to),
 	}, nil
 }
 
 func (handlers *Handlers) getConvertFromFiatHandler(r *http.Request) (interface{}, error) {
 	isFee := false
 	from := r.URL.Query().Get("from")
+	// TODO here we should check the `from` field, but there is not a list of valid fiats
```
benma, line 655: imo TODO's in code are very rarely useful and should only be added if the fix is coming very soon, if at all. i have seen too many permanent TODO's that stay forever or lose their meaning over time :sweat_smile: 

new tasks are best put into an issue tracker.

frontends/web/src/components/rates/formatCurrency.tsx
```diff
@@ -0,0 +1,48 @@
+/**
+ * Copyright 2018 Shift Devices AG
+ * Copyright 2021 Shift Crypto AG
```
benma, line 3: 2022 Shift Crypto

frontends/web/src/components/rates/formatCurrency.tsx
```diff
@@ -0,0 +1,48 @@
...
+  amount: number;
+  currency: string;
+}
+
+export type FormattedCurrency = {
+  amount: string;
+  success: string;
+}
+
+export const getFormatCurrency = (props: Props) => {
```
benma, line 32: move this to api/backend.ts

frontends/web/src/components/rates/rates.tsx
```diff
@@ -141,13 +131,16 @@ function Conversion({
...
+  let toFormat = false;
+  let value = 0;
 
   if (amount.conversions) {
     if (amount.conversions[active] !== '')
       formattedValue = amount.conversions[active];
   } else {
     if (rates && rates[coin]) {
-      formattedValue = formatCurrency(rates[coin][active] * Number(amount.amount), active);
+      value = rates[coin][active] * Number(amount.amount);
```
benma, line 142: a lot of effort to make a format backend API endpoint, but this still uses floats :thinking: 

what is the plan here?

maybe use the `/coins/convertToFiat` endpoint in some way?

frontends/web/src/components/rates/formatCurrency.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FunctionComponent } from 'react';
+import { useLoad } from '../../hooks/api';
+import { apiGet } from '../../utils/request';
+
+interface Props {
+  amount: number;
```
benma, line 23: this is a floating point. didn't you want to use `string`? otherwise the backend formatting doesn't do much to improve precision, does it?

frontends/web/src/components/rates/rates.tsx
Beerosagos: As far as I understood, this is only used to convert to fiat the total coin balance in the account summary (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/routes/account/summary/accountssummary.tsx#L262). I would suggest to directly add the `conversions` fiat values to the `total-balance` endpoint and remove this piece of code. I would do that in a separate PR, though, wdyt?

backend/handlers/handlers.go
Beerosagos: ok! I added it to the issue I made to list the supported fiat (https://gitlab.com/shiftcrypto/godbb/-/issues/1337

frontends/web/src/components/rates/formatCurrency.tsx
Beerosagos: You are right, I'll change it, thx!

backend/handlers/handlers.go
```diff
@@ -515,17 +518,21 @@ func (handlers *Handlers) getAccountsTotalBalanceHandler(_ *http.Request) (inter
...
 	for k, v := range totalPerCoin {
 		currentCoin, err := handlers.backend.Coin(k)
 		if err != nil {
 			return nil, err
 		}
-		formattedTotalPerCoint[k] = currentCoin.FormatAmount(coin.NewAmount(v), false)
+
+		totalAmount[k] = accountHandlers.FormattedAmount{
+			Amount:      currentCoin.FormatAmount(coin.NewAmount(v), false),
+			Conversions: conversionsPerCoin[k],
```
benma, line 532: maybe i am blind, but is this used anywhere? the `ITotalBalance` interface in the frontend is unchanged, and it doesn't fit `IAmount`.

It seems everything works also without adding `Conversions` here, not sure how we came to the conclusion that this was needed here :sweat_smile: do you know?

Beerosagos: I think this is correct. Why do you say that it doesn't fit IAmount? It is used here, to display the fiat value of the total coin balance for multiple accounts https://github.com/digitalbitbox/bitbox-wallet-app/pull/1735/commits/f785a580e1261cf545337236b55aa1ab15028db9#diff-e5d06f77122fc6a06ea5e7c80d3466cd0fa4d69a252e8108de78062a5bdbea83R255

If you don't add `Conversions` in `getAccountsTotalBalanceHandler` return value, it won't be correctly displayed.

backend/chart.go
```diff
@@ -79,14 +88,22 @@ func (backend *Backend) addChartData(
 		ChartEntry := chartEntries[timestamp]
```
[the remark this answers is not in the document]
Beerosagos: Sure! :)

backend/chart.go
```diff
@@ -79,14 +88,22 @@ func (backend *Backend) addChartData(
...
-		).Float64()
-		ChartEntry.Value += fiatValue
+		)
+
+		if ChartEntry.RatValue == nil {
+			ChartEntry.RatValue = new(big.Rat).Set(fiatValue)
+		} else {
+			ChartEntry.RatValue.Add(fiatValue, ChartEntry.RatValue)
+		}
+		ChartEntry.FormattedValue = coin.FormatAsCurrency(ChartEntry.RatValue, fiat)
```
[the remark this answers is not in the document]
Beerosagos: Right! ok

frontends/web/src/components/rates/rates.tsx
```diff
@@ -133,23 +107,19 @@ function Conversion({
...
-  rates,
   active,
   noAction,
   children,
 }: PropsWithChildren<Props>): JSX.Element | null {
 
-  const coin = amount.unit;
   let formattedValue = '---';
 
   if (amount.conversions) {
```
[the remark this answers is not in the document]
Beerosagos: Right! Do you want me to directly update `IAmount` and remove the check here, overriding your pending PR?

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -353,7 +353,7 @@ class AccountsSummary extends Component<Props, State> {
                         {(data && data.chartTotal !== null) ? (
```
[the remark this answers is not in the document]
Beerosagos: Great catch, thx!

frontends/web/src/components/rates/rates.tsx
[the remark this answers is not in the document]
benma: Yeah if it's possible sure. Can also finish this first and do more cleanups after, up to you.

Another cleanup I am seeing: `IAmount` contains `unit: string`, but I don't see that it is actually used (same for `FormattedAmount.Unit`, the backend equivalent).

backend/coins/eth/coin.go
```diff
@@ -147,6 +148,14 @@ func (coin *Coin) ToUnit(amount coin.Amount, isFee bool) float64 {
 	return result
 }
 
+// SetAmount implements coin.Coin.
+func (coin *Coin) SetAmount(amount *big.Rat, isFee bool) coin.Amount {
```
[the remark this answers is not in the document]
Beerosagos: Done, is that ok? eth/coin_test.go seems a bit messy compared to btc/coin.go (no dedicated test functions). Should I refactor splitting into a dedicated function? 
Also,  SetAmount has different behaviors for ERC20 token (depending on the number of digits and fee flag), Should I introduce an ERC20 token in the test?

backend/coins/btc/coin.go
```diff
@@ -168,6 +168,13 @@ func (coin *Coin) ToUnit(amount coinpkg.Amount, isFee bool) float64 {
 	return result
 }
 
+// SetAmount implements coinpkg.Coin.
+func (coin *Coin) SetAmount(amount *big.Rat, isFee bool) coinpkg.Amount {
```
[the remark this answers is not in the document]
Beerosagos: Done!

frontends/web/src/components/rates/rates.tsx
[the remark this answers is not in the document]
Beerosagos: Done refactoring IAmount and removing the conversions check from rates.tsx. 

I don't see what you mean with the `unit: string` field. `IAmount.unit` is of `Coin` type and is used in several places around the code. Am I missing something?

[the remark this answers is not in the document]
Beerosagos: > Done refactoring IAmount and removing the conversions check from rates.tsx.

UPDATE: I realized that `IAmount.conversions` can be null for unconfirmed transactions, generating a runtime error. I put the check back in `rates.tsx`.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -98,13 +105,27 @@ class Chart extends Component<Props, State> {
...
     }
   }
 
   private hasData = (): boolean => {
     return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length > 0;
   };
 
+  private setFormattedData(data : ChartData) {
+    if (!this.formattedData)
+      this.formattedData = new Map<number,string>();
```
benma, line 118: what is a map and why is it better than a simple object `{ ... }`?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -98,13 +109,24 @@ class Chart extends Component<Props, State> {
...
+      this.setFormattedData(data);
     }
   }
 
   private hasData = (): boolean => {
     return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length > 0;
   };
 
+  private setFormattedData(data : ChartData) {
+    if (!this.formattedData)
```
benma, line 121: never skip curly braces :joy:

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -98,13 +109,24 @@ class Chart extends Component<Props, State> {
...
+      this.formattedData = {} as FormattedData;
```
benma, line 122: don't you want to do this every time? reset to `{}`

benma: talked to @thisconnect, he will try to enable a linter for formatting, e.g. no space in `data :` (before colon), require curly braces, etc.

Beerosagos: fixed!

frontends/web/src/components/rates/rates.tsx
```diff
@@ -56,11 +46,6 @@ apiGet('config').then((appconf) => {
   }
 });
 
-// TODO: should not invoking apiGet imediatelly, see the apiGet() function for more details
-apiGet('rates').then(rates => store.setState({ rates }));
```
benma, line 60: remove the backend /rates handler and the backend 'rates' push notifications? they now seem unused

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -677,7 +677,7 @@ class Send extends Component<Props, State> {
                         onInput={this.handleFiatInput}
                         disabled={sendAll}
                         error={amountError}
-                        value={fiatAmount}
+                        value={fiatAmount.replace('\'','')} /* input with type=number doesn't like some formatting chars like [']. */
```
benma, line 680: this is a bit hacky and not so robust, i.e. likely to break (e.g. if we add different locales formatting).

why not deliver the fiat amount in the same format as before this PR from the backend?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -98,13 +109,24 @@ class Chart extends Component<Props, State> {
...
   private hasData = (): boolean => {
     return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length > 0;
   };
 
+  private setFormattedData(data: ChartData) {
+    this.formattedData = {} as FormattedData;
+
+    data.forEach( entry => {
+      if (this.formattedData) {
+        this.formattedData[entry.time as number] =  entry.formattedValue;
```
benma, line 125: if you rebase on upstream/master, a new linter should complain about the double space here when running `make weblint`. if not, we should find a linter that does :D

benma (changes requested):

frontends/web/src/routes/account/send/send.tsx
Beerosagos: I don't like that too, but we decided to move and centralize all the formatting in the backend, and `FormatAsCurrency` adds apostrophes https://github.com/Beerosagos/bitbox-wallet-app/blob/formatting-refactor/backend/coins/coin/conversions.go#L21
This clearly has to be changed to allow locales support, just like `FormatAsCurrency` in backend. Maybe is it better to change both in a dedicated PR for locales?

frontends/web/src/components/rates/rates.tsx
Beerosagos: I left that because I am not sure if it could be useful for addressing the issue of transaction rates, which are not updating in send view. I can create the issue saying that there is this problem and that we should remove this code if it is still unused after fixing it. wdyt? Just to avoid deleting and then rewriting the same code

frontends/web/src/routes/account/send/send.tsx
benma: Sure, but who says that `getConvertToFiatHandler()` needs to format the amount with a locale? it can just return a plain string that is suitable for this input, e.g. instead of `coinpkg.FormatAsCurrency(convertedAmount, to)` use `convertedAmount.FloatString(2)` (or 8 in case of BTC)?

Beerosagos: Fixed, I made a dedicated `FormatAsPlainCurrency` for this

backend/handlers/handlers.go
```diff
@@ -632,53 +639,66 @@ func (handlers *Handlers) getConvertToFiatHandler(r *http.Request) (interface{},
...
 	}
+
 	rate := handlers.backend.RatesUpdater().LatestPrice()[from][to]
+	convertedAmount := new(big.Rat).Mul(amountRat, new(big.Rat).SetFloat64(rate))
+	var formattedAmount string
+	if plain == "true" {
+		formattedAmount = coinpkg.FormatAsPlainCurrency(convertedAmount, to)
+	} else {
+		formattedAmount = coinpkg.FormatAsCurrency(convertedAmount, to)
+	}
```
benma, line 658: don't see this being used anywhere? why keep it? could get rid of the `plain` param

backend/coins/coin/conversions.go
```diff
@@ -1,15 +1,33 @@
...
+		formatted = strings.TrimRight(strings.TrimRight(amount.FloatString(8), "0"), ".")
+	} else {
+		formatted = amount.FloatString(2)
+	}
+	return formatted
+}
+
+// FormatAsCurrency handles formatting for currencies.
+func FormatAsCurrency(amount *big.Rat, coinUnit string) string {
+	var formatted string
```
benma, line 25: `formatted := FormatAsPlainCurrency(amount, coinUnit)`?

Beerosagos: ack!

backend/handlers/handlers.go
Beerosagos: I thought that in this way it can be used (now and in the future) to serve both kinds of formatting needs. I'll change it

Beerosagos: Should be ok now

benma (approved):
LGTM! pls rebase, there seems to be a conflict

PR #1736, opened 2022-08-08
frontend: migrate env to typescript

thisconnect:
also added global.d.ts to define qt and android globals.


benma (approved):
nice, lgtm!

I tested the Qt app and the Android app, everything works.

PR #1737, opened 2022-08-08
Move some icons to icon component

thisconnect:
Many components contain inline SVG (in JSX), instead of using an
external asset and link with image source. Having SVG in JSX is
sometimes useful if some attributes change at runtime, i.e. the
tx progress ring. But having static SVG in JSX hurts memory usage
and rendering performance, for example the vdom diff has to check
all of the static SVG attributes on every render and it will
increase the JavaScript bundle size.
See also https://twitter.com/_developit/status/1382838799420514317

Moved some icons to the icon component assets and changed the props
typing so TypeScript can check valid HTML attributes.

Also fixed wrong attribute that was wrongly copied from JSX to SVG,
SVG attributes use dashes, not camelCase, example: strokeWidth
should be stroke-width.


benma (approved):
LGTM

PR #1738, opened 2022-08-10
frontend: fix receive view with software keystore

thisconnect:
In e4f0d4e7 the receive view was split into BitBox01 and BitBox02
components so that the BitBox01 legacy version could remain as is.

This introduced an issue that no receive view is shown with the
software keystore used in webdev (dev mode).

Fixed receive view for software keystore and render the BitBox02
receive component.

### DIFF
--- a/frontends/web/src/routes/account/receive/index.tsx
+++ b/frontends/web/src/routes/account/receive/index.tsx
@@ -34,15 +34,14 @@ export const Receive: FunctionComponent<Props> = props => {
   } = props;
   const device = deviceIDs.length ? devices[deviceIDs[0]] : undefined;
   switch (device) {
-  case 'bitbox02':
-    return (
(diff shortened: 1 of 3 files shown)


thisconnect:
Added a commit to pass deviceID and removed some BB01 code as discussed, PTAL

> It would be nice if you passed along the device ID to receive.tsx anyway, so then you could e.g. hide the verify button completely if there is no device (deviceID is undefined).

… AppVeyor doesn't like it, but seems unrelated

```
Installing the following packages:
make
By installing, you accept licenses for the packages.
make not installed. An error occurred during installation:
 The remote server returned an error: (524).
```

benma (approved):

benma:
Thanks. I took the liberty of merging this, as I needed the receive screen to work for my upcoming PR :innocent:

PR #1740, opened 2022-08-14
backend/btc/maketx: don't use btcec in test

benma:
We want to update the btcd dep, but it heavily refactored the btcec
package - it now does not use big.Int anymore and there seems to be no
way to get the R and S values as big.Int anymore.

This test is changed to not use the btcec dep at all, circumventing
that problem.

I manually converted the DER signature from before to the
types.Signature that is needed in the test.


Beerosagos (approved):
tACK!

PR #1742, opened 2022-08-17
vendor: update btcsuite (btcd, btcec, btcutil)

benma:
btcutil is now part of btcd.

btcec had a major version bump and is now at btcec/v2.

The updated btcutil/btcd have full support for Taproot, which means
that we can drop our custom Taproot code (see address.go and util.go).

The `txValidityCheck()` was also disabled when we added Taproot as the
tx verified by btcd didn't understand Taproot yet. Now it does, so we
can re-enable it again.

Go-ethereum had to be updated at the same time, as that dependency
uses btcec/v2 itself in the new version.

```
go get github.com/ethereum/go-ethereum@v1.10.21
go get github.com/digitalbitbox/bitbox02-api-go@e9258db31750ed3ba2b4e988b4bd74c9c4107f64
go mod tidy
go mod vendor
```


backend/coins/btc/addresses/address.go
```diff
@@ -99,49 +97,8 @@ func NewAccountAddress(
...
-			return h.Sum(nil)
-		}
-		if isOdd(pubKey) {
-			pubKey = negate(pubKey)
-		}
-		_, tweak := btcec.PrivKeyFromBytes(pubKey.Curve, taggedHash("TapTweak", pubKey.SerializeCompressed()[1:]))
-		x, y := pubKey.Curve.Add(pubKey.X, pubKey.Y, tweak.X, tweak.Y)
-		xOnlyPubKey := (&btcec.PublicKey{Curve: pubKey.Curve, X: x, Y: y}).SerializeCompressed()[1:]
-		address, err = btcutil.NewAddressTaproot(xOnlyPubKey, net)
+		outputKey := txscript.ComputeTaprootKeyNoScript(configuration.PublicKey())
```
benma, line 100: @guggero thanks for reminding me that btcd now has full Taproot support - now we can remove some of our own code, e.g. the one above where the output key is computed!

benma:
@Beerosagos please take a look

backend/coins/ltc/params.go
```diff
@@ -83,19 +83,13 @@ var MainNetParams = chaincfg.Params{
 	MinerConfirmationWindow:       8064, //
 	Deployments: [chaincfg.DefinedDeployments]chaincfg.ConsensusDeployment{
 		chaincfg.DeploymentTestDummy: {
-			BitNumber:  28,
```
benma, line 86: todo: get the right params from ltcd

benma: done

Beerosagos (approved):
tACK :+1:

PR #1743, opened 2022-08-18
frontend/api/account: add conversions to UTXO amount type

benma:
It is delivered by the backend and used in <FiatConversion> to display
the conversions already.


Beerosagos:
we included this in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1735 
Closing!

PR #1744, opened 2022-08-18
backend/devices/keystore: don't panic if previous output missing

benma:
Not being able to reference the previous output of an input is a bug,
but at the moment one can run into it when making really fast
consecutive transactions: if a new tx is made before the previous one
is broadcast and received from the Electrum backend, it might reuse
UTXOs, which become spent the moment the previous tx is processed.

This commit simply removes the panic and turns it into an error that
the user sees in the frontend.

### DIFF
--- a/backend/devices/bitbox/keystore.go
+++ b/backend/devices/bitbox/keystore.go
@@ -150,7 +150,8 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
 	for index, txIn := range transaction.TxIn {
 		spentOutput, ok := btcProposedTx.PreviousOutputs[txIn.PreviousOutPoint]
 		if !ok {
-			keystore.log.Panic("There needs to be exactly one output being spent per input!")
+			keystore.log.Error("There needs to be exactly one output being spent per input.")
+			return errp.New("There needs to be exactly one output being spent per input.")
 		}
 		address := btcProposedTx.GetAddress(spentOutput.ScriptHashHex())
 		isSegwit, subScript := address.ScriptForHashToSign()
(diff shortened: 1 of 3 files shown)


thisconnect:
Thanks LGTM

I ran into "too-long-mempool-chain, too many unconfirmed ancestors [limit: 25]". I could reproduce an error when sending bigger amounts, but no more panic in the log.


```
An unknown error occurred: Failed to broadcast transaction: the transaction was rejected by network rules.
insufficient fee, rejecting replacement 9c8818bdc8af797d8092df680569ce6de4a3cbf712cea4adf8820dd98c293a60; new feerate 0.00001000 BTC/kvB <= old feerate 0.00001004 BTC/kvB
[01000000000102a24678b12b84cabd20872e571f880cda7517fc93617ee1117217e85a33cb7e370000000000fdffffff56680f8d7ec483e336d41663d842048ee1c0503e5d829a499e2ff61bf570bb920100000000fdffffff0268f70c000000000016001468d0d7b1891d1623ea92896065897361e9155ae0cbbd480000000000160014c82a82282791e2ff6fb0ee913f0028b28dfbc5cb0248304502210088b5023656fb2bd4884148a32806171fd9b5510f684971103ce48844f223cd2b022074c439ddbe7c951343c4445400e0b69332827c9fcb2acc705b7ab47fdfcf8dce0121033010ae25a6244a992a7f519c8c525ebfa53aa54b5781332ccd7016e6bc82e5e6024730440220514f3537ebf2bdf322e6e1f34d7fcfb45a8b945b4574f750a87d4af98fc1fabe0220108e1b3ff6b67b487664632ffec6a28d33911b559deeac2a628a9e4736b8623301210242071efc2403845b8f0254ad16dea6fb5c483bb2148632dc6c8e6ae7fa3a46bc00000000]
```

PR #1746, opened 2022-08-23
frontend/send: allow more than 2 decimal places in the fiat field

benma:
When BTC is the selected "fiat" currency, entering more than two
decimal places results in a red error border even though it is a legal
BTC amount.

The two decimal places where not correct before either e.g. for JPY,
where there are no decimals at all.


Beerosagos (approved):
tACK!

PR #1747, opened 2022-08-24
backend: for pending tx, set fiatAtTime to null, not an empty struct

benma:
This indicates clearly to the frontend that the value does not
exist. FiatConversion then is not invoked with garbage data, which
allows FiatConversion to be simplified by not taking into
consideration `null` conversions.


Beerosagos:
we included this in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1735
closing!

PR #1748, opened 2022-08-25
frontend: eslint rule for curly braces

thisconnect:
Always use consitent braces with newlines for better readability.


thisconnect:
Updated code with `make webfix`

benma (approved):

benma:
Can you also make a linter for correct spacing? E.g. no double spaces, one space before and after operators, no space before a colon in a type, etc.?

`if (a) {` instead of `if( a)` etc.

benma (approved):

PR #1750, opened 2022-08-28
android: make status bar color the same white as the app background

benma:
Before it was a dark green from Android project template that we never
changed.

```
<item name="android:windowLightStatusBar">true</item>
```

is added so that the text in the status bar turns black, otherwise it
would become invisible (white on white).

windowLightStatusBar is available only from API level 23 (even though
statusBarColor is available already from API level 21), so we enable
it for 23+.


thisconnect (approved):
LGTM

PR #1751, opened 2022-08-29
backend/keystore/software: add Taproot (P2TR) support

benma:
We added P2TR support for the BitBox02 but couldn't do the same for
the software (debug) keystore as BTCD didn't have Taproot support
yet. Now it does, and since we upgraded the dep in
c44d766c87303821855b4bcdee045b619317b5bf, we can use Taproot also in
the software keystore.

`keystore.sign()` is inlined into the main signing function, as the
signatures now need to be distinguished between Schnorr/BIP86 and
ECDSA.

The private key is tweaked accoriding to BIP-86 to match the public
key that backs the BIP-86 P2TR address created by address.go.

### DIFF
--- a/backend/keystore/software/software.go
+++ b/backend/keystore/software/software.go
@@ -22,6 +22,7 @@ import (
 	"math/big"
 
 	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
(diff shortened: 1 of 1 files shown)


Beerosagos (approved):
tACK!

PR #1752, opened 2022-08-30
backend/test: refactor eth coin test using testify/suite

Beerosagos:


backend/coins/eth/coin_test.go
```diff
@@ -35,38 +41,46 @@ func TestCoin(t *testing.T) {
 		nil,
 		nil,
 	)
+}
 
-	require.Equal(t,
+func TestSuite(t *testing.T) {
+	suite.Run(t, new(testSuite))
```
benma, line 47: maybe run the suite multiple times, once with ETH and once with an ERC20 token /with different number of decimals)? then the SetAmount test also covers tokens and `isFee = true` behaves differently.

benma (approved):
LGTM!

backend/coins/eth/coin_test.go
Beerosagos: Ok! I'll push an update

Beerosagos:
Added erc20 token test

PR #1754, opened 2022-08-31
backend/eth: fixed bug on pending transaction type

Beerosagos:
By default pending ETH tx were always displayed with `send` type. This
commit add a check on the account address to correctly display
`send_to_self` txs.


benma (approved):
utACK

PR #1755, opened 2022-09-01
qt: setRequestInterceptor is obsolete

thisconnect:
The obsolete setRequestInterceptor causes the follwing warning:
'Use of deprecated not thread-safe setter, use setUrlRequestInterceptor instead.'

Use setUrlRequestInterceptor instead, see
https://doc.qt.io/qt-5/qwebengineprofile-obsolete.html#setRequestInterceptor


benma (approved):
tACK

PR #1760, opened 2022-09-07
frontend/send: update balance after sending a transaction

Beerosagos:


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -166,6 +166,18 @@ class Send extends Component<Props, State> {
           break;
         }
         break;
+      case 'backend':
```
thisconnect, line 169: small nit: To me `backend` doesn't make sense as everything that is pushed is coming from the backend. (same reason I think we should not have `api/backend`)

maybe `account` would be more suitable?

thisconnect: ignore, I forgot that it is already like that.. :P

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -166,6 +166,18 @@ class Send extends Component<Props, State> {
...
+        switch (data) {
+        case 'newTxs':
```
benma, line 171: this is the wrong event, you want `syncdone` like in account.tsx.

thisconnect:
The balance on the send view did not update after the send success message.
    
    Changed send component to listen to syncdone event and update the balance
    to show the correct remaining balance.

Could you ad something like this to the commit body?

Also please add a short note to the changelog.

frontends/web/src/routes/account/send/send.tsx
Beerosagos: Changed, thanks!

Beerosagos:
Now should be ok

thisconnect (approved):
LGTM

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -166,6 +166,18 @@ class Send extends Component<Props, State> {
...
         }
         break;
+      case 'account':
+        switch (data) {
+        case 'syncdone':
+          if (this.props.code) {
+            accountApi.getBalance(this.props.code)
+              .then(newBalance => {
+                this.setState({ balance: newBalance });
+              }).catch(console.error);
```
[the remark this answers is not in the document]
Beerosagos: Nice! Updated, thx

PR #1762, opened 2022-09-08
frontend/bitbox02: abort SetPassword when the user aborts

benma:
Before, the error was always assumed to be the password mismatch error, and the SetPassword workflow was invoked again immediately.

Now we go back to the initial screen if the user aborts.

The BitBox02 only returns the user abot error from this API endpoint from v9.13.0, before the two types of errors can not be distinguished.


Beerosagos:
Nice! Only thing I noticed is that there is still a generic error when the user aborts while restoring from backup. I would suggest adding a check on the error code and avoiding the error message on restore. See https://github.com/benma/bitbox-wallet-app/blob/bb02abort/frontends/web/src/routes/device/bitbox02/backups.tsx#L80

benma:
@Beerosagos that means ACK? :smile:

benma:
> @Beerosagos that means ACK? smile

(the Restore thing is a separate issue imo)

Beerosagos (approved):
LGTM!

Beerosagos:
> > @Beerosagos that means ACK? smile
> 
> (the Restore thing is a separate issue imo)

in that case ACK :)

PR #1763, opened 2022-09-12
backend: generate syncdone event on rates update

Beerosagos:
Recent modifications in the fiat conversion and formatting caused a regression in the live update of fiat converted values. This fixes the issue, generating a new syncdone event everytime an updated rates value is observed, and allowing frontend to reload the relevant data.


backend/accounts/baseaccount.go
```diff
@@ -208,6 +208,15 @@ func (account *BaseAccount) Initialize(accountIdentifier string) error {
 		account.notes = append(account.notes, legacyNotes)
 	}
 
+	// An account syncdone event is generated when new rates are available. This allows the frontend to reload the relevant data.
+	if account.config.RateUpdater != nil {
+		account.config.RateUpdater.Observe(func(e observable.Event) {
+			if e.Subject == "rates" {
```
benma, line 214: consider putting this string into a constant

Beerosagos: now should be better!

benma (approved):
utACK

PR #1766, opened 2022-09-14
frontend: improved insufficient gas error for erc20 txs

Beerosagos:
If the user tries to send an erc20 transaction without having enough ETH to pay for the fee, etherscan returns a rough error message, which was previously displayed in the frontend as is.
Now the backend returns a dedicated error code and the frontend displays a comprehensible and localized error message.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1194,6 +1194,7 @@
       "total": "Total"
     },
     "error": {
+      "erc20InsufficientGasFunds": "It seems like you do not have enough Ether to pay for this ERC20 transaction. Please make sure you hold enough Ether in your wallet",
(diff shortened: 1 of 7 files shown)


thisconnect:
@benma please review Go changes

frontends/web/src/locales/en/app.json
```diff
@@ -1189,6 +1189,7 @@
       "total": "Total"
     },
     "error": {
+      "erc20InsufficientGasFunds": "It seems like you do not have enough Ether to pay for this ERC20 transaction. Please make sure you hold enough Ether in your Wallet",
```
benma, line 1192: 'in your Wallet' -> 'in your wallet.'

benma (approved):
tACK with nit

PR #1768, opened 2022-09-16
Frontend blocking firmware upgrade

thisconnect:
Once the user is prompted to confirm upgrading the firmware on the
device, the dialog in the app should be blocking.

Changed to use DialogButtons to have nice alignment and do
not need to import the CSS file from another component.


benma (approved):
utACK

PR #1769, opened 2022-09-20
Frontend type transports

thisconnect:


frontends/web/src/utils/request.ts
```diff
@@ -88,7 +102,10 @@ export function apiGet(endpoint) {
...
 }
 
-export function apiPost(endpoint, body) {
+export function apiPost(
+  endpoint: string,
+  body?: object | number | string | boolean,
+): Promise<any> {
   if (runningInQtWebEngine()) {
     return call(JSON.stringify({
       method: 'POST',
```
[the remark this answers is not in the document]
thisconnect: Could add a comment to change this to return `Promise<unknown>` but then api needs to be changed like this

```diff
diff --git a/frontends/web/src/api/account.ts b/frontends/web/src/api/account.ts
index c91ae2a0..34741cfc 100644
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -47,8 +47,8 @@ export interface IAccount {
     blockExplorerTxPrefix: string;
 }
 
-export const getAccounts = (): Promise<IAccount[]> => {
-  return apiGet('accounts');
+export const getAccounts = () => {
+  return apiGet('accounts') as Promise<IAccount[]>;
 };
 ```

File: frontends/web/src/utils/transport.common.ts [hunk 2]
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+export type TQueryPromiseMap = {
+  [key: number]: {
+    resolve: (value: unknown) => void,
+    reject: (value: unknown) => void,
+  };
+};
+
+export type TMsgCallback = (payload: any) => void; // should be { type: any; data: any; meta: any; } or
```
thisconnect, line 24: Temporary moved payload and msgCallback type here to share, typed this in the next PR, so all apiWebsocket return either event or event-legacy type.

thisconnect:
Let me know if and of which parts I should make separate PR's

frontends/web/src/utils/transport.common.ts
thisconnect: The next PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/1775

frontends/web/src/globals.d.ts
```diff
@@ -17,6 +17,8 @@
 export declare global {
     interface Window {
         qt?: unknown;
-        android?: unknown;
+        android?: (query: string) => void;
+        onAndroidCallResponse?: unknown;
```
benma, line 21: can be `(queryID: number, response: string) => void` instead of `unknown`, right?

frontends/web/src/globals.d.ts
```diff
@@ -17,6 +17,8 @@
...
+        onAndroidPushNotification?: (msg: string) => void;
```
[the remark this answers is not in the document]
benma: I think it should be like this:

```
android?: {
            call: (queryID: number, query: string) => void;
        };
```

Here is the function signature of the call method:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/3ab4d430c14c9a926afd9a8165690f7bae680c42/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java#L70

frontends/web/src/utils/androidtransport.ts
```diff
@@ -16,40 +15,39 @@ import { runningInAndroid } from './env';
...
   });
 }
 
-export function androidSubscribePushNotifications(msgCallback) {
-  // @ts-ignore
+export function androidSubscribePushNotifications(msgCallback: TMsgCallback) {
   if (typeof window.onAndroidPushNotification === 'undefined') {
-    // @ts-ignore
-    window.onAndroidPushNotification = msg => {
+    window.onAndroidPushNotification = (msg: string) => {
```
benma, line 50: is typing `msg` here needed if it is already typed in global.d.ts? just curious; i don't mind repeating it for clarity anyway.

frontends/web/src/utils/qttransport.ts
```diff
@@ -37,36 +39,39 @@ function initTransport() {
...
-        webChannel.objects.backend.pushNotify.connect(msg => {
+        webChannel.objects.backend.pushNotify.connect((msg: string) => {
           currentListeners.forEach(listener => listener(JSON.parse(msg)));
         });
         resolve(webChannel);
+
       };
-      // @ts-ignore
-      cache = new QWebChannel(qt.webChannelTransport, initWebChannel); // eslint-disable-line no-undef
+      cache = new QWebChannel((window.qt as any).webChannelTransport, initWebChannel);
```
benma, line 57: why not type `qt` as `any` in the first place then? or make it of type `{ webChannelTransport: any; }` instead?

frontends/web/src/utils/qttransport.ts
```diff
@@ -15,20 +16,21 @@
...
 import { QWebChannel } from './qwebchannel';
+import { TMsgCallback, TQueryPromiseMap } from './transport-common';
 import { runningInQtWebEngine } from './env';
 
-let cache = null;
-let webChannel = null;
-let queryID = 0;
-let queryPromises = {};
-let currentListeners = [];
+let cache: any = null;
```
benma, line 22: after this PR is merged, I have some ideas how to improve this. `cache` can actually be `isChannelInitialized: boolean` instead, and `initTransport()` etc. could be made async functions that are much easier to read.

frontends/web/src/utils/request.ts
```diff
@@ -35,18 +36,31 @@ function extConfig(key, defaultValue) {
...
 }
 
-export function apiURL(endpoint) {
+export function apiURL(endpoint: string): string {
   return (isTLS() ? 'https://' : 'http://') + 'localhost:' + apiPort + '/api/' + endpoint;
 }
 
-function handleError(endpoint) {
-  return function(json) {
+// https://fettblog.eu/typescript-hasownproperty/
```
benma, line 47: pls add a bit more docstring about the purpose of this, in case this page becomes unavailable.

frontends/web/src/utils/request.ts
```diff
@@ -35,18 +36,31 @@ function extConfig(key, defaultValue) {
...
+  return obj.hasOwnProperty(prop);
+}
+
+function handleError(endpoint: string) {
+  return function(json: unknown) {
     return new Promise((resolve, reject) => {
-      if (json && json.error) {
+      if (
+        typeof json === 'object'
+        && json !== null
```
benma, line 60: is this needed? if it is an object (previous line), it is not null, right?

frontends/web/src/utils/request.ts
```diff
@@ -88,7 +102,10 @@ export function apiGet(endpoint) {
   }).then(response => response.json()).then(handleError(endpoint));
 }
 
-export function apiPost(endpoint, body) {
+export function apiPost(
+  endpoint: string,
+  body?: object | number | string | boolean,
```
benma, line 107: body should be `any`, right? As JSON.Stringify() works for any type. In the future maybe a post endpoint accepts an array etc.

frontends/web/src/utils/request.ts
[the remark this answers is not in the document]
benma: what would be the type of `getAccounts` in that case? I find it strange if the top level declaration doesn't have an explicit return type.

benma (changes requested):

frontends/web/src/utils/androidtransport.ts
thisconnect: I think I used unknown for onAndroidPushNotification  in global.d.ts now it is not needed anymore

frontends/web/src/utils/request.ts
thisconnect: yes seems so, lets check together…

Screen Shot 2022-10-05 at 14 21 59

frontends/web/src/utils/request.ts
thisconnect: JSON.stringify should throw a TypeError when

- value contains a circular reference
- or [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) value is encountered.


```js
JSON.stringify(BigInt("9007199254740991"))
```

also 

> Only [enumerable own properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties) are visited. This means [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), etc. will become "{}".

frontends/web/src/utils/request.ts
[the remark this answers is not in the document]
thisconnect: discussed offline, will change to `Promise<unknown>` in a new PR

thisconnect:
@benma updated as discussed PTAL

frontends/web/src/utils/androidtransport.ts
```diff
@@ -16,40 +15,39 @@ import { runningInAndroid } from './env';
...
         };
       }
-
       queryID++;
       queryPromises[queryID] = { resolve, reject };
-      // @ts-ignore
-      android.call(queryID, query);
+      if (window.android) {
+        window.android.call(queryID, query);
+      }
```
benma, line 41: window.android!.call(queryID, query) - we know that window.android exists at this point

benma (approved):

frontends/web/src/utils/request.ts
```diff
@@ -25,7 +26,7 @@ import { runningInAndroid, runningInQtWebEngine } from './env';
 // is provided in case the file wasn't generated but used directly,
 // for convenience when developing. Both key and defaultValue must be
 // strings and converted into the desired type.
-function extConfig(key, defaultValue) {
+function extConfig(key: string, defaultValue: string): string {
   if (typeof key === 'string' && key.startsWith('{{ ') && key.endsWith(' }}')) {
```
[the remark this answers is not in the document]
thisconnect: agree and removed.

PR #1770, opened 2022-09-20
frontend: enable sats unit option for bitcoin accounts

Beerosagos:
The main unit displayed in the app for bitcoin accounts is Bitcoin (BTC), but sometimes it can be easier to think in terms of Satoshi (SATS), especially for lower amounts.

This allows the user to choose the preferred unit using a switch in the expert settings section of the frontend.


backend/coins/btc/coin.go
```diff
@@ -46,6 +46,7 @@ type Coin struct {
 	code                  coinpkg.Code
 	name                  string
 	unit                  string
+	formatUnit            string
```
benma, line 49: docstrings for these two explaining what each of them are for would be useful

backend/coins/coin/codes.go
```diff
@@ -40,6 +40,13 @@ const (
 	// There are some more coin codes for the supported erc20 tokens in erc20.go.
 )
 
+const (
```
benma, line 43: why did you put these in the generic `coin` package instead of using them locally in the `btc` package?

backend/coins/coin/codes.go
```diff
@@ -40,6 +40,13 @@ const (
...
+	// UnitBtc represents Bitcoin unit.
+	UnitBtc string = "BTC"
+	// UnitSats represents Satoshi unit.
+	UnitSats string = "SATS"
```
benma, line 47: I don't think `SATS` is commonly used. `sat` is commonly used I think (lower case, not plural)

backend/handlers/handlers.go
```diff
@@ -197,6 +197,7 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/coins/tbtc/headers/status", handlers.getHeadersStatus(coinpkg.CodeTBTC)).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/ltc/headers/status", handlers.getHeadersStatus(coinpkg.CodeLTC)).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/btc/headers/status", handlers.getHeadersStatus(coinpkg.CodeBTC)).Methods("GET")
+	getAPIRouter(apiRouter)("/coins/btc/set-unit", handlers.getBtcFormatUnit).Methods("GET")
```
benma, line 200: since this makes state changes, it should be a POST request

backend/coins/coin/coin.go
```diff
@@ -41,6 +41,12 @@ type Coin interface {
 	// The fee unit is usually the same as the main unit, but can differ.
 	Unit(isFee bool) string
 
+	// SetformatUnit sets the unit used to format the amount, e.g. "BTC" or "SATS".
```
benma, line 44: explain what it does for non-BTC accounts. does it even need to be in the interface if it is only used for BTC? consider removing it from the interface.

backend/handlers/handlers.go
```diff
@@ -677,6 +712,8 @@ func (handlers *Handlers) getConvertFromFiatHandler(r *http.Request) (interface{
 		}, nil
 	}
 
+	currentCoin.ToUnit(currentCoin.SetAmount(fiatRat, false), false)
```
benma, line 715: does this line even do anything?

frontends/web/src/locales/en/app.json
```diff
@@ -1277,7 +1277,8 @@
       "fee": "Enable custom fees",
       "setProxyAddress": "Set proxy address",
       "title": "Expert settings",
-      "useProxy": "Enable tor proxy"
+      "useProxy": "Enable tor proxy",
+      "useSats": "Use SATS unit for Bitcoin Accounts"
```
benma, line 1281: in the app we use "title case" everywhere, i.e. no capitalized words in the middle of a sentence => "Bitcoin accounts".

benma (changes requested):

backend/accounts.go
```diff
@@ -416,6 +416,12 @@ func (backend *Backend) createAndAddAccount(
 			backend.arguments.GapLimits(),
 			backend.log,
 		)
+
+		if specificCoin.Code() == coinpkg.CodeBTC ||
+			specificCoin.Code() == coinpkg.CodeTBTC {
+			specificCoin.SetFormatUnit(backend.config.AppConfig().Backend.BtcUnit)
+		}
```
benma, line 423: this is the wrong place for this code. first, you wouldn't expect a function that  creates an account to modify the underlying coin (which is shared between multiple accounts). secondly, this is redundant, as it will be called for the same coin multiple times if there are multiple btc accounts.

maybe `backend.Coin()` is the right function to put this in, which is when the coin instance is created?

benma:
If you select BTC as the "fiat" currency, all the BTC formatted there are not formatted as satoshi's currently, but probably should. Example:

image

thisconnect:
an alternative could be to keep BTC as base currency and add "sat" as another base currency (but I guess that would be a different PR)..

> If you select BTC as the "fiat" currency, all the BTC formatted there are not formatted as satoshi's currently, but probably should.

frontends/web/src/routes/settings/settings.tsx
```diff
@@ -116,6 +118,23 @@ class Settings extends Component<Props, State> {
...
+    }
+    const target = (event.target as HTMLInputElement);
+    var unit: BtcUnit = 'BTC';
+    if (target.checked) {
+      unit = 'SATS';
+    }
+
+    setConfig({
+      backend: { btcUnit: unit }
+    }).then(config => this.setState({ config }));
```
thisconnect, line 134: shouldn't it be `etConfig({ frontend: { btcUnit: unit } })` instead of backend?

benma: not true, see e.g. setProxyConfig.

It's in the backend object if it is accessed by the backend.

frontends/web/src/api/coins.ts
```diff
@@ -29,3 +30,12 @@ export const subscribeCoinHeaders = (coinCode: CoinCode) => (
     subscribeEndpoint(`coins/${coinCode}/headers/status`, cb)
   )
 );
+
+interface ISetUnit {
+  success: boolean;
+  errorMessage?: string;
+}
```
thisconnect, line 37: I think it is ok for now, but we should cleanup the different responses.

Ideally we could use SuccessResponse and FailResponse from `api/response.ts`, but it is a bit different :(

Also we have ISuccess from `api/backend.ts` but this optionally contains errorCode, which reminds me shouldn't coins/btc/set-unit also return errorCode?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -819,7 +828,7 @@ class Send extends Component<Props, State> {
                       <small>{(proposedTotal && proposedTotal.unit) || 'N/A'}</small>
                     </span>
                     {(proposedTotal && proposedTotal.conversions) && (
-                      <span> <span className="text-gray">/</span> <strong>{proposedTotal.conversions[fiatUnit]}</strong> <small>{fiatUnit}</small></span>
+                      <span> <span className="text-gray">/</span> <strong>{proposedTotal.conversions[fiatUnit]}</strong> <small>{fiatUnit === 'BTC' ? btcUnit : fiatUnit}</small></span>
```
thisconnect, line 831: Could save this in a variable and reuse it `const baseCurrencyUnit = fiatUnit === 'BTC' ? btcUnit : fiatUnit;` I am not sure about the name `baseCurrencyUnit`, but fiatUnit is a bit outdated as it can also be BTC.

backend/accounts.go
```diff
@@ -417,6 +417,7 @@ func (backend *Backend) createAndAddAccount(
 			backend.arguments.GapLimits(),
 			backend.log,
 		)
+
```
thisconnect, line 420: nit: remove?

backend/config/config.go
```diff
@@ -88,6 +88,9 @@ type Backend struct {
 	// or set to empty by the frontend if its value matches native locale
 	// as reported by the OS.
 	UserLanguage string `json:"userLanguage"`
+
+	// BtcUnit is the unit used to represents Bitcoin amounts, e.g. `BTC` or `SATS`.
```
thisconnect, line 92: `SATS` -> `sat` ?

backend/coins/coin/conversions.go
```diff
@@ -8,10 +8,18 @@ import (
...
 
+// Btc2Sat is the sat equivalent of 1 BTC.
+const Btc2Sat = 1e8
+
 // FormatAsPlainCurrency handles formatting for currencies in a simplified way.
 // This should be used when `FormatAsCurrency` can't be used because a simpler formatting is needed (e.g. to populate forms in the frontend).
-func FormatAsPlainCurrency(amount *big.Rat, coinUnit string) string {
+func FormatAsPlainCurrency(amount *big.Rat, coinUnit string, btcUnit string) string {
 	var formatted string
+	if btcUnit == UnitSats || btcUnit == "t"+UnitSats {
```
thisconnect, line 18: question: does `"t"+UnitSats` mean it could be `tsat`?

thisconnect (changes requested):

frontends/web/src/api/coins.ts
Beerosagos: > I think it is ok for now, but we should cleanup the different responses.
Do you mean to uniform the different endpoint responses and errors handling in the `handlers.go` files?

I think in this case an error message could be the right thing, but I don't know if we have a specific ratio to handle error messages vs error codes. @benma ?

backend/coins/btc/coin.go
Beerosagos: Done :+1:

backend/coins/coin/codes.go
Beerosagos: With the last update it is also used by `coin` pkg, I left it there, since including `btc` package generates a inclusion cycle.

backend/coins/coin/codes.go
Beerosagos: Ok!

backend/handlers/handlers.go
Beerosagos: Right, changed

backend/coins/coin/coin.go
Beerosagos: Removed, thx

backend/handlers/handlers.go
Beerosagos: Actually, no :sweat_smile:  removed

frontends/web/src/locales/en/app.json
Beerosagos: :+1:

backend/accounts.go
Beerosagos: Good point, thx. Changed

backend/coins/coin/conversions.go
Beerosagos: Yep. When using testnet BTC was already formatted as TBTC. I did the same with sat -> tsat

.golangci.yml
```diff
@@ -59,7 +59,7 @@ linters-settings:
       - unlambda
       - unslice
   gocyclo:
-    min-complexity: 27
+    min-complexity: 28
```
Beerosagos, line 62: This was required because of changes in chart.go. I'll make a separate PR to split the function in order to reduce this value again.

backend/handlers/handlers.go
```diff
@@ -685,6 +745,11 @@ func (handlers *Handlers) getConvertFromFiatHandler(r *http.Request) (interface{
 		unit = unit[2:]
 	}
 
+	btcCoin, _ := handlers.backend.Coin(coinpkg.CodeBTC)
+	if from == coinpkg.UnitBtc && btcCoin.(*btc.Coin).GetFormatUnit() == coinpkg.UnitSats {
+		fiatRat = new(big.Rat).Quo(fiatRat, new(big.Rat).SetFloat64(coinpkg.Btc2Sat))
```
benma, line 750: I see this `new(big.Rat).Quo(amountRat, new(big.Rat).SetFloat64(unitSatoshi))` line three times in this PR. Please move all of them to the btc package in a common function, ideally with a unit test.

Also, instead of SetFloat64(), maybe you can just use `big.NewRat()`? since the unitSatoshi const is not a float but an int.

benma (changes requested):
Partway through, but I will already submit the current review.

backend/accounts/baseaccount.go
```diff
@@ -59,6 +59,8 @@ type AccountConfig struct {
 	GetSaveFilename       func(suggestedFilename string) string
 	// Opens a file in a default application. The filename is not checked.
 	UnsafeSystemOpen func(filename string) error
+	// BtcCurrencyUnit is the unit which should be used to format fiat amounts values expressed in BTC..
```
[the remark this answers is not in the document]
Beerosagos: Currently, it is `BtcUnit` for backend configuration, `formatUnit` for btc/coin and `BtcCurrencyUnit` for accounts/baseaccount. The rationale is the following:
- `formatUnit` express the formatting unit of a BTC coin (while `unit` is the main unit)
- `BtcCurrencyUnit` express the unit of BTC when it is intended as a currency value (i.e. in amount conversions)
- `BtcUnit`is the backend configuration which is used to store the sat configuration for btc coins and currency, since we decided that we want the same behavior for both of them.

If we decide on a different behavior in the future, having separated names for coins and currency formatting could be helpful, I think. But if you feel like it is better to have a consistent name everywhere I am ok with that.

Beerosagos:
Pushed an update, but just found a bug. Converting to draft until it is fixed

frontends/web/src/api/coins.ts
Beerosagos: Modified using `ISuccess`, without error code

backend/handlers/handlers.go
```diff
@@ -490,6 +491,36 @@ func (handlers *Handlers) getAccountsHandler(_ *http.Request) (interface{}, erro
...
+		return nil, errp.WithStack(err)
+	}
+
+	unit := request.Unit
+
+	// update BTC format unit for Coins
+	btcCoin, _ := handlers.backend.Coin(coinpkg.CodeBTC)
+	btcCoin.(*btc.Coin).SetFormatUnit(unit)
+	btcCoin, _ = handlers.backend.Coin(coinpkg.CodeTBTC)
+	btcCoin.(*btc.Coin).SetFormatUnit(unit)
```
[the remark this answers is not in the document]
Beerosagos: Done

backend/coins/coin/conversions.go
```diff
@@ -8,21 +8,29 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/rates"
 )
 
+// Btc2Sat is the sat equivalent of 1 BTC.
+const Btc2Sat = 1e8
+
 // FormatAsPlainCurrency handles formatting for currencies in a simplified way.
 // This should be used when `FormatAsCurrency` can't be used because a simpler formatting is needed (e.g. to populate forms in the frontend).
```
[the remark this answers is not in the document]
Beerosagos: Changed `btcUnit` in a bool variable `formatBtcAsSats` plus a helper function. Using a dedicated type would have required to create units for each coin

backend/handlers/handlers.go
```diff
@@ -515,10 +546,17 @@ func (handlers *Handlers) getAccountsTotalBalanceHandler(_ *http.Request) (inter
...
 		} else {
 			totalPerCoin[coinCode] = new(big.Int).Add(totalPerCoin[coinCode], amount.BigInt())
 		}
-		conversionsPerCoin[coinCode] = coin.Conversions(coin.NewAmount(totalPerCoin[coinCode]), account.Coin(), false, account.Config().RateUpdater)
+
+		unit := ""
+		btcCoin, ok := account.Coin().(*btc.Coin)
+		if ok {
+			unit = btcCoin.GetFormatUnit()
+		}
```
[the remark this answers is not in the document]
Beerosagos: Good point! Done, thx

backend/handlers/handlers.go
```diff
@@ -527,8 +565,17 @@ func (handlers *Handlers) getAccountsTotalBalanceHandler(_ *http.Request) (inter
 			return nil, err
 		}
 
+		var unit string
+		btcCoin, ok := currentCoin.(*btc.Coin)
+		if ok {
+			unit = btcCoin.GetFormatUnit()
+		} else {
+			unit = currentCoin.Unit(false)
+		}
```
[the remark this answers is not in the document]
Beerosagos: Ahah ok no pro :)

backend/handlers/handlers.go
```diff
@@ -639,20 +686,33 @@ func (handlers *Handlers) getConvertToPlainFiatHandler(r *http.Request) (interfa
...
-			"errMsg":  "invalid amount",
+			"errMsg":  "coin not supported",
 		}, nil
 	}
 
+	coinAmount, err := currentCoin.ParseAmount(amount, false)
+	if err != nil {
+		return map[string]interface{}{
+			"success": false,
+			"errMsg":  err.Error(),
```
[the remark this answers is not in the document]
Beerosagos: Deleted

backend/handlers/handlers.go
Beerosagos: Done

Beerosagos:
Bug should be fixed: it was failing to retrieve erc20 coins in `getConvertToPlainFiatHandler`.
PR Re-opened and ready for a new round of review :+1:

backend/handlers/handlers.go
```diff
@@ -639,20 +685,37 @@ func (handlers *Handlers) getConvertToPlainFiatHandler(r *http.Request) (interfa
 	from := r.URL.Query().Get("from")
 	to := r.URL.Query().Get("to")
 	amount := r.URL.Query().Get("amount")
-	amountRat, valid := new(big.Rat).SetString(amount)
-	if !valid {
+
+	currentCoin, err := handlers.backend.Coin(coinpkg.Code(strings.ToLower(from)))
```
benma, line 689: i don't think `from` is a reliable field for the coin code. The frontend send `coinUnit` in for this field, which might or might not be the coincode. 

Why not just provide `coinCode` as a GET parameter? Then it is 100% correct, without the heuristic of trying to convert the unit to the coin code.

Beerosagos: Good point, thx! Now should be better

benma (changes requested):

backend/coins/btc/coin.go
```diff
@@ -168,11 +192,24 @@ func (coin *Coin) ToUnit(amount coinpkg.Amount, isFee bool) float64 {
...
 // SetAmount implements coinpkg.Coin.
 func (coin *Coin) SetAmount(amount *big.Rat, isFee bool) coinpkg.Amount {
-	satsAmount := new(big.Rat).Mul(amount, new(big.Rat).SetFloat64(unitSatoshi))
+	satsAmount := coinpkg.Btc2Sat(amount)
 	intSatsAmount, _ := new(big.Int).SetString(satsAmount.FloatString(0), 0)
 	return coinpkg.NewAmount(intSatsAmount)
 }
 
+// ParseAmount implements coinpkg.Coin.
+func (coin *Coin) ParseAmount(amount string, isFee bool) (coinpkg.Amount, error) {
```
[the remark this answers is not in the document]
Beerosagos: Done :+1:

backend/coins/eth/coin.go
```diff
@@ -118,6 +119,16 @@ func (coin *Coin) Unit(isFee bool) string {
 	return coin.unit
 }
 
+// SetFormatUnit implements coin.Coin.
+func (coin *Coin) SetFormatUnit(string) {
+	// No need for format unit on ETH coins.
+}
```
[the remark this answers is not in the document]
Beerosagos: I agree it is not the best to have empty functions around the code, but having `SetFormatUnit` in the interface allows to avoid casting in `set-unit` handler, keeping the code cleaner and easier to read. Let me know if you prefer to have it removed from the interface anyway :)

backend/coins/eth/coin.go
```diff
@@ -156,6 +167,16 @@ func (coin *Coin) SetAmount(amount *big.Rat, isFee bool) coin.Amount {
 	return coinpkg.NewAmount(intWeiAmount)
 }
 
+// ParseAmount implements coinpkg.Coin.
+func (coin *Coin) ParseAmount(amount string, isFee bool) (coinpkg.Amount, error) {
```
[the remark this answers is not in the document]
Beerosagos: Done :+1:

backend/coins/coin/coin.go
```diff
@@ -55,6 +61,10 @@ type Coin interface {
 	// e.g. BTC 1/2 => 50000000
 	SetAmount(amount *big.Rat, isFee bool) Amount
 
+	// ParseAmount parse a String representing a given amount, considering the formatting unit.
+	// e.g. if the formatUnit is set as "sat", the amount will be considered as being sats
+	ParseAmount(amount string, isFee bool) (Amount, error)
```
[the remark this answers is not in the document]
Beerosagos: Right! Removed :)

backend/handlers/handlers.go
```diff
@@ -685,6 +745,17 @@ func (handlers *Handlers) getConvertFromFiatHandler(r *http.Request) (interface{
...
 	}
 
+	btcCoin, err := handlers.backend.Coin(coinpkg.CodeBTC)
+	if err != nil {
+		return map[string]interface{}{
+			"success": false,
+			"errMsg":  "internal error",
+		}, nil
+	}
+	if from == coinpkg.UnitBtc && btcCoin.GetFormatUnit() == coinpkg.UnitSats {
```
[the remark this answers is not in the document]
Beerosagos: Totally agree, thx for pointing it out

frontends/web/src/api/account.ts
```diff
@@ -22,12 +22,14 @@ export type CoinCode = 'btc' | 'tbtc' | 'ltc' | 'tltc' | 'eth' | 'teth' | 'reth'
...
 export type AccountCode = string;
 
-export type Fiat = 'AUD' | 'BRL' | 'BTC' | 'CAD' | 'CHF' | 'CNY' | 'EUR' | 'GBP' | 'HKD' | 'ILS' | 'JPY' | 'KRW' | 'NOK' | 'RUB' | 'SEK' | 'SGD' | 'USD';
+export type Fiat = 'AUD' | 'BRL' | 'BTC' | 'CAD' | 'CHF' | 'CNY' | 'EUR' | 'GBP' | 'HKD' | 'ILS' | 'JPY' | 'KRW' | 'NOK' | 'RUB' | 'SEK' | 'SGD' | 'USD' | 'sat';
 
 export type MainnetCoin = 'BTC' | 'LTC' | 'ETH';
 
 export type TestnetCoin = 'TBTC' | 'TLTC' | 'TETH' | 'RETH';
 
+export type BtcUnit = 'BTC' | 'sat';
```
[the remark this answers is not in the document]
Beerosagos: Done :+1:

frontends/web/src/api/account.ts
```diff
@@ -22,12 +22,14 @@ export type CoinCode = 'btc' | 'tbtc' | 'ltc' | 'tltc' | 'eth' | 'teth' | 'reth'
 
 export type AccountCode = string;
 
-export type Fiat = 'AUD' | 'BRL' | 'BTC' | 'CAD' | 'CHF' | 'CNY' | 'EUR' | 'GBP' | 'HKD' | 'ILS' | 'JPY' | 'KRW' | 'NOK' | 'RUB' | 'SEK' | 'SGD' | 'USD';
+export type Fiat = 'AUD' | 'BRL' | 'BTC' | 'CAD' | 'CHF' | 'CNY' | 'EUR' | 'GBP' | 'HKD' | 'ILS' | 'JPY' | 'KRW' | 'NOK' | 'RUB' | 'SEK' | 'SGD' | 'USD' | 'sat';
```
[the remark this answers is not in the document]
benma: chartFiat  in `ISummary` is also of type `FiatUnit` right?

I think `FiatUnit` right next to `Fiat` which are also units is confusing. Please add a comment about what it means. at some point we should probably rename 'Fiat' to 'Conversion' or sth like that, as the BTC in the Fiats becomes more and more confusing :joy:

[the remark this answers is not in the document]
Beerosagos: > chartFiat in ISummary is also of type FiatUnit right?

Right, done. Thx!

Definitely the Fiat naming is confusing and we should move to something that can include BTC and sat without creating confusion. I renamed `FiatUnit` as `ConversionUnit`, hope it is a bit more clear. We could make a separate PR refactoring all the `Fiat` naming to `Conversion` both in frontend and backend

benma (approved):
tACK!

thisconnect (approved):

PR #1772, opened 2022-09-21
frontend: modify fiat historical value placeholder in tx

Beerosagos:
Fiat historical value is shown in tx detail. If the transaction is recent the value could be not available. The placeholder we used was '---' which led to poorly readable results for incoming/outgoing txs ('+---' or '----'). The new placeholer is 'N/A', which should be more readable.

### DIFF
--- a/frontends/web/src/components/rates/rates.tsx
+++ b/frontends/web/src/components/rates/rates.tsx
@@ -99,6 +99,8 @@ interface ProvidedProps {
     unstyled?: boolean;
     skipUnit?: boolean;
     noAction?: boolean;
+    sign?: string;
+
 }
 
 type Props = ProvidedProps & SharedProps;
@@ -110,13 +112,15 @@ function Conversion({
   skipUnit,
   active,
   noAction,
-  children,
+  sign,
 }: PropsWithChildren<Props>): JSX.Element | null {
 
   let formattedValue = '---';
+  let isAvailable = false;
 
   // amount.conversions[active] can be empty in recent transactions.
   if (amount && amount.conversions[active] !== '') {
+    isAvailable = true;
     formattedValue = amount.conversions[active];
   }
 
@@ -138,8 +142,8 @@ function Conversion({
     );
   }
   return (
-    <span className={style.rates}>
-      {children}
+    <span className={ `${style.rates} ${!isAvailable ? style.notAvailable : ''}`}>
+      {isAvailable ? sign : ''}
       {formattedValue}
       {' '}
       {
--- a/frontends/web/src/components/rates/rates.module.css
+++ b/frontends/web/src/components/rates/rates.module.css
@@ -41,6 +41,10 @@
     right: 2px;
 }
 
+.notAvailable {
+    color: var(--color-default);
+}
+
 @media (max-width: 880px) {
     .fiatRow {
         bottom: 0;
--- a/frontends/web/src/components/transactions/transaction.tsx
+++ b/frontends/web/src/components/transactions/transaction.tsx
@@ -142,7 +142,7 @@ class Transaction extends Component<Props, State> {
     ) : (
       <ArrowSelf />
     );
-    const sign = ((type === 'send') && '−') || ((type === 'receive') && '+') || null;
+    const sign = ((type === 'send') && '−') || ((type === 'receive') && '+') || '';
     const typeClassName = (type === 'send' && style.send) || (type === 'receive' && style.receive) || '';
     const sDate = time ? this.parseTimeShort(time) : '---';
     const statusText = {
@@ -204,7 +204,7 @@ class Transaction extends Component<Props, State> {
             </div>
             <div className={parentStyle.fiat}>
               <span className={`${style.fiat} ${typeClassName}`}>
-                <FiatConversion amount={amount} noAction>{sign}</FiatConversion>
+                <FiatConversion amount={amount} sign={sign} noAction />
               </span>
             </div>
             <div className={parentStyle.currency}>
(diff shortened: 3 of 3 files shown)


frontends/web/src/components/rates/rates.tsx
```diff
@@ -113,7 +113,7 @@ function Conversion({
   children,
 }: PropsWithChildren<Props>): JSX.Element | null {
 
-  let formattedValue = '---';
+  let formattedValue = ' N/A';
```
benma, line 116: `+ N/A` and `- N/A` are not much better :innocent: 

I think the `---` is not the problem (both `---` and `N/A` are okay`), but that there is a sign in `+` or `-` in front, coming from https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/components/transactions/transaction.tsx#L207. 

Not sure why this is passed as `children` anyway. Maybe a good opportunity to make it a property instead, and not render it if there is no amount to render.

benma (changes requested):

frontends/web/src/components/rates/rates.tsx
Beerosagos: ok, pushed an update to improve it

frontends/web/src/components/rates/rates.tsx
```diff
@@ -99,6 +99,9 @@ interface ProvidedProps {
     unstyled?: boolean;
     skipUnit?: boolean;
     noAction?: boolean;
+    typeClassName?: string;
+    sign?: string | undefined;
```
thisconnect, line 103: nit: `?` indicates optional and should already allow `undefined `

Beerosagos: Thx, updated!

frontends/web/src/components/rates/rates.tsx
```diff
@@ -99,6 +99,9 @@ interface ProvidedProps {
     unstyled?: boolean;
     skipUnit?: boolean;
     noAction?: boolean;
+    typeClassName?: string;
+    sign?: string;
```
benma, line 103: how about `sign?: '+' | '-'`?

then at this point you can remove `typeClassName` here and move the style.send and style.receive from transaction.tsx to this component instead. would be a lot simpler to understand imo than the current code, where the logic is quite hard to understand at a glance.

Beerosagos: Seems really nice, but the style is also used outside the `FiatConversion` component, to render the coin amounts (e.g. https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/components/transactions/transaction.tsx#L211). Maybe I can refactor `typeClassName` to `txTypeStyle` or smth similar, to make it more clear. wdyt?

thisconnect: > the style is also used outside the FiatConversion component

Is this about the color? It would be nice if the `FiatConversion` component would behave like text and just inherit the parents color.

benma: oh too bad. yeah or maybe just a comment above the prop

benma: @thisconnect the point is that if there is on value to show, it should show `---` or `N/A` without style (otherwise it would be green/red). At least that's how I understood it.

frontends/web/src/components/rates/rates.tsx
```diff
@@ -138,8 +143,8 @@ function Conversion({
     );
   }
   return (
-    <span className={style.rates}>
-      {children}
+    <span className={ `${style.rates} ${formattedValue !== unavailableValue ? typeClassName : ''}`}>
+      {formattedValue !== unavailableValue ? sign : ''}
```
thisconnect, line 147: This inline condition is a bit hard to read. Could change `const isNotAvailable = formattedValue !== unavailableValue;` and just use `isNotAvailable` here.

frontends/web/src/components/rates/rates.tsx
thisconnect: Ah I see, in case for `N/A` the FiatComponent could apply normal text color and overwrite the parents color.

Beerosagos: Now should be better: the style is inherited by the parent, unless the amount is not available.

frontends/web/src/components/rates/rates.tsx
Beerosagos: Done :+1:

frontends/web/src/components/rates/rates.tsx
```diff
@@ -110,13 +112,16 @@ function Conversion({
   skipUnit,
   active,
   noAction,
-  children,
+  sign,
 }: PropsWithChildren<Props>): JSX.Element | null {
 
-  let formattedValue = '---';
+  let unavailableValue = ' ---';
+  let formattedValue = unavailableValue;
```
thisconnect, line 119: Is the space before the dashes `' ---';` still needed?

Also `unavailableValue` is not used anymore, I think you can just directly do `let formattedValue = '---';`

thisconnect (changes requested):

benma (approved):
utACK, deferring to @thisconnect for the final nits :)

frontends/web/src/components/rates/rates.tsx
thisconnect: > how about `sign?: '+' | '-'`

also tried this, and it looks like the share decorator messes something up with the types. I propose to keep as string for now and then change it after rates.tsx is refactored.

frontends/web/src/components/rates/rates.tsx
Beerosagos: Right! Done :+1:

thisconnect (approved):
tACK, LGTM

PR #1773, opened 2022-09-22
frontend/api: implement typed function to verify attestation on bb02

Beerosagos:
Before this, `/attestation` endpoint was called in an untyped way from `bitbox02.tsx`. Now there is a dedicated typed funcion in `api/bitbox02.ts`.

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -81,3 +81,9 @@ export const setMnemonicPassphraseEnabled = (
       return Promise.resolve();
     });
 };
+
+export const verifyAttestation = (
+  deviceID: string,
+): Promise<boolean> => {
+  return apiGet(`devices/bitbox02/${deviceID}/attestation`);
+};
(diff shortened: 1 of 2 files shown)


thisconnect (approved):
LGTM

frontends/web/src/api/bitbox02.ts
```diff
@@ -81,3 +81,9 @@ export const setMnemonicPassphraseEnabled = (
       return Promise.resolve();
     });
 };
+
+export const verifyAttestation = (
+  deviceID: string,
+): Promise<boolean> => {
```
benma, line 87: Fyi the return type of this API call is `null | boolean`, see here:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/e2f1008475cad14c7ecf9f73011f819a18d03e6a/backend/devices/bitbox02/handlers/handlers.go#L128

https://github.com/digitalbitbox/bitbox02-api-go/blob/e9258db31750ed3ba2b4e988b4bd74c9c4107f64/api/firmware/attestation.go#L166-L167

image

cc @Beerosagos @thisconnect

benma: Was also typed like this already here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1773/files#diff-ddc6137588aa6e257eb264fca22e2ab9ab7d8183da99c07fe4df15b58c2e0ab3R52

PR #1778, opened 2022-09-28
frontend: add small check icon on green circle

thisconnect:
The anmiated check icon was used also in small version on the send success dialog and device geniune in settings. This icon does not look good and is hard to read when it is small, also does not need to be animated.

Added a new check icon on a green circle that looks good as small icon and to reduce confusion about different check icons, renamed the big one to AnimatesChecked icon.


thisconnect:
Screen Shot 2022-09-28 at 16 55 27

Screen Shot 2022-09-28 at 16 43 21

Beerosagos (approved):
tACK! The new icon is much nicer :+1:

PR #1779, opened 2022-09-28
frontend: refactor toggle component

thisconnect:
- changed from class based to function component
- removed '[key: string]: any;' type and added className type
- changed from interface to type
- cleaned up CSS, alphabetically sorted CSS properties and removed prefixed styles

### DIFF
--- a/frontends/web/src/components/toggle/toggle.module.css
+++ b/frontends/web/src/components/toggle/toggle.module.css
@@ -1,54 +1,50 @@
 .container {
-    position: relative;
-    display: inline-block;
-    width: 60px;
-    min-height: var(--item-height-xsmall);
-    margin: 0;
+  display: inline-block;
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/toggle/toggle.module.css
```diff
@@ -1,54 +1,50 @@
...
-    border-radius: 2px;
+  background-color: var(--color-disabled);
+  border-radius: 2px;
+  bottom: 0;
+  cursor: pointer;
+  left: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  transition: .2s;
```
thisconnect, line 29: removed prefixed CSS property `-webkit-transition` as we only target chromium browser and the build would automatically prefix if it was still necessary

PR #1781, opened 2022-09-29
frontend: refactor fiat default

thisconnect:
This is not about fixing the financial system but only about which
legacy currency should be set as the default unit of account. lol.

The current styling of the default fiat is star becomes orange
(outlined icon) and the toggle becomes inactive as it is disabled
so that the default cannot be turned off. The styling looks
confusing as it is not clear that this is the default.

To improve the visual styling this commit:

- added small '(default)' text after the currency
- made active star more prominent, changed from outline to filled

Also (additional commit)
- move inline svg to image to reduce the bundle size, JavaScript parse time and React dom diffing.

### DIFF
--- a/frontends/web/src/components/icon/assets/icons/star-inactive.svg
+++ b/frontends/web/src/components/icon/assets/icons/star-inactive.svg
@@ -1,0 +1,12 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24"
+  height="24"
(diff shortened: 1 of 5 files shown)


thisconnect:
Screen Shot 2022-09-29 at 10 25 51

thisconnect:
FYI: removed parentheses

Beerosagos (approved):
tACK

PR #1784, opened 2022-09-30
Windows: improve Visual Studio installation detection

mdcmeridius:
Change compile_windows.bat to detect other Visual Studio installations when the VS 2019 Community Edition is not found.

It uses the registry to detect the visual studio installation path.
It assumes the default icon registry value always ends in "ide\devenv.exe,1200". There is no built-in batch file (or shell) command to extract regular expressions, so I had to resort to a non-flexible string replace. Ideally this should use another shell (e.g.: Powershell) or third-party utility to extract the path from the registry.

### DIFF
--- a/frontends/qt/compile_windows.bat
+++ b/frontends/qt/compile_windows.bat
@@ -1,6 +1,22 @@
 :: Compiles the Qt5 app. Part of `make windows`, which also compiles/bundles the deps
 
-call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
+setlocal
+set varsbat="C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
+if exist %varsbat% (goto :CALL_VC_VARS_BAT)
+echo Trying to figure out Visual Studio 2019 location...
+for /f "tokens=2* eol=," %%a in ('reg query "HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Classes\CLSID\{2E1517DA-87BF-4443-984A-D2BF18F5A908}\DefaultIcon" ^|findstr /rc:"REG_SZ *"') do set varsbat=%%~b
+If Defined varsbat (
+  set varsbat="%varsbat:common7\ide\devenv.exe,1200=VC\Auxiliary\Build\vcvars64.bat%"
+)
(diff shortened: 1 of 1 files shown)


benma:
> Change compile_windows.bat to detect other Visual Studio installations when the VS 2019 Community Edition is not found.

Can there be a different VS 2019 edition?

I am not sure going to VS2022 is a good idea for two reasons:

- [here](https://github.com/digitalbitbox/bitbox-wallet-app/blob/e2f1008475cad14c7ecf9f73011f819a18d03e6a/docs/BUILD.md?plain=1#L81) it is indicated that Qt5.15 depends on VS 2019 
- Even though the app is not yet reproducible, having fixed dependencies is better than dynamic dependencies.

WDYT?

Fyi, this these commands are only used to produce a production build. When doing development in the codebase, compiling the native app is not necessary, and one can use `make servewallet` and `make webdev` instead.

mdcmeridius:
> Can there be a different VS 2019 edition?

Yes, unfortunately Visual Studio Professional and Visual Studio Entreprise use different installation directories.
I can confirm this.

> I am not sure going to VS2022 is a good idea for two reasons:

I cannot confirm the C++ build works in VS2022 because I do not have the C++ components installed.
Therefore, I agree we should remove the VS2022 fallback section.

Should I change it, or will you edit the file while merging?

mdcmeridius:
> And they use the same registry key?

Can't tell you for sure. Must try the 2019 community.,
I'm using the CLSID from these registry keys:

image

I look for the path by following "Default icon" node.

Anyway, 6 more lines of batch file code is better than just a hard-coded path to the Community Edition installed on the C: drive, don't you agree? :-)

benma:
I will close this issue, as I don't feel like the complications in the build script are worth the payoff, considering the two points I made above, especially:

> Or alternatively, use a virtual machine? Note that compiling the final QT app is only needed when making a production release. 95% of the time during development, using make webdev and make servewallet suffices.

Updating Qt/VS or changing the build system in the future would take a lot more time if we had to take this registry stuff into account as well.

mdcmeridius:
Well, let me tell you that vehemently disagree with your decision because having a HARD-CODED PATH is a very unprofessional way of doing things.

I had a considerable amout of work proving that this works perfectly. Thanks a lot!

benma:
I agree hard-coded paths are suboptimal, but there are other ways of solving this, e.g. by defining and using an environment variable.

I don't think my statement above is unreasonable - the payoff for complicating the build script is not worth the benefit in its current form. Sorry you feel differently. Thanks anyway.

PR #1785, opened 2022-10-03
frontends\web\src: color variables refactoring

jadzeidan:
Seperated text and background colors into their own variables. This makes colors easier to maintain because now it's possible to change text and background colors without affecting each other.

Also some text used `--soft-black` instead of a designated `/* font colors */` variable. Changed this text to `--color-default` so it is consistant with the other text colors.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -689,6 +689,26 @@
       "text": "Click the \"Export\" button and open the downloads folder where you will find the CSV export. Then click the link below, upload your BitBox CSV file and import the data in order to use it for your CoinTracking portfolio manager and to create your tax reports.",
       "title": "How to import my transactions into CoinTracking?"
     },
+    "device": {
+      "attestation": {
+        "link": {
+          "text": "Read more about the authenticity check"
+        },
+        "text": "The BitBoxApp performs an attestation check on the BitBox02 to verify if the device is genuine. The check is done locally and does not connect to any servers.",
+        "title": "How does the authenticity check work?"
+      },
+      "name": {
+        "text": "This is the name of your wallet and backup. The name is used for future backups and can be used to help distinguish between different wallets. It can be changed at any time but note that backups made before the change will still use the previous name.",
+        "title": "What is the BitBox02 name used for?"
+      },
+      "secure-chip": {
+        "link": {
+          "text": "Read more about the secure chip"
+        },
+        "text": "This information shows the model number of the secure chip, the most up to date chip is ATECC608B with improved security features compared to older models.",
+        "title": "Why show the secure chip model?"
+      }
+    },
     "exchanges": {
       "commission": {
         "text": "This page uses affiliate links where available. However, not all listed exchanges have an affiliate program. Whether an exchange offers one does not influence our decision to list them.",
--- a/frontends/web/src/routes/device/bitbox02/settings-guide.tsx
+++ b/frontends/web/src/routes/device/bitbox02/settings-guide.tsx
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2022 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useTranslation } from 'react-i18next';
+import i18n from '../../../i18n/i18n';
+import { Entry } from '../../../components/guide/entry';
+import { Guide } from '../../../components/guide/guide';
+
+export const ManageDeviceGuide = () => {
+  const { t } = useTranslation();
+  return (
+    <Guide>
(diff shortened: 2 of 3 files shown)


jadzeidan:
> I don't think you intended to make 6 commits 😁 try squashing them into 1 commit and putting the PR description into the commit msg. Let me know if you need help with that.

I'm trying to squash but it's multiplying 😭

frontends/web/src/components/forms/checkbox.module.css
```diff
@@ -69,19 +69,19 @@
...
-    color: var(--color-white);
+    color: var(--color-alt);
 }
 
 .info input:checked + label::before {
     background-color: var(--color-info);
     border-color: var(--color-info);
-    color: var(--color-white);
+    color: var(--color-alt);
 }
```
thisconnect, line 86: There is more in this file, see line 14 `background: rgb(255, 255, 255);` can be changed to `background: var(--background-secondary);`

frontends/web/src/components/forms/radio.module.css
```diff
@@ -50,7 +50,7 @@
 /* checked */
```
thisconnect, line 51: there is more in this file, search for `background: rgb(255, 255, 255);`

frontends/web/src/style/variables.css
```diff
@@ -25,6 +25,7 @@
     --color-primary: var(--color-blue);
     --color-secondary: var(--color-gray);
     --color-light: var(--color-mediumgray);
```
thisconnect, line 27: Searched, but `--color-light` isn't used anywhere, but `--color-mediumgray` is used in many places, remove `--color-light`  here or change all places to use `--color-light` instead of  `--color-mediumgray` ?

thisconnect: Also `--color-gray` and `--color-gray-alt` is used in various places, so for now I would just remove `--color-light` here. And then look at mediumgray vs. gray vs. gray-alt later in a different PR.

thisconnect (changes requested):
Very nice cleanup effort, thanks!

Also have a look at spinner.module.css and other modules where there is `background: #fff;` could you do global search?

and in view.module.css it should use `var(--background)` and not `background: var(--color-lightgray);`. Related: `--color-lightgray` is used many times, could you check if that should be replaced with `--background` everyhere or in some places?

frontends/web/src/routes/account/summary/accountssummary.module.css
```diff
@@ -57,14 +57,14 @@
 
 .table thead th,
 .table tfoot th {
-    background-color: var(--color-white);
-    color: var(--color-softblack);
+    background-color: var(--background-secondary);
+    color: var(--color-default);
```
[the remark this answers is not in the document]
jadzeidan: deleted

frontends/web/src/routes/account/summary/accountssummary.module.css
```diff
@@ -57,14 +57,14 @@
...
     font-size: var(--size-default);
     font-weight: bold;
 }
 
 .table strong {
-    color: var(--color-softblack);
+    color: var(--color-default);
 }
```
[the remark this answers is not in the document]
jadzeidan: removed

frontends/web/src/components/forms/checkbox.module.css
jadzeidan: changed

frontends/web/src/components/forms/radio.module.css
jadzeidan: updated to `--background-secondary`

frontends/web/src/style/variables.css
jadzeidan: I searched `--color-mediumgray` and found that it is mostly used for backgrounds (and boarders). I found one instance where it is used as a font color and changed that to `--color-light`.

Since background and fonts shouldn't use the same variable, we need to make a new background variable to use `--color-mediumgray`.

jadzeidan: I also see we use `--color-gray-alt` which doesn't have a corresponding font color variable. So we either need to make one or change it to an existing font color variable such as `--color-secondary`.

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -36,10 +36,10 @@
 
 .filters button {
   appearance: none;
-  background: var(--color-white);
-  border: 2px solid var(--color-white);
+  background: var(--background-secondary);
+  border: 2px solid var(--background-secondary);
   border-radius: 2rem;
-  color: var(--color-softblack);
+  color: var(--color-default);
```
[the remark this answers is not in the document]
jadzeidan: `--color-secondary`
image

`--color-default` (in pr)
image

`--color-secondary` looks too light for me (looks like the buttons are disabled). If you think `--color-default` is too dark then we need to make a new variable for fonts that use `--color-softblack)`.

To me, `--color-default` and `--color-softblack` are very similar that you only notice a difference when you see it side by side. So I would keep it `--color-default`.

frontends/web/src/components/transactions/transactions.module.css
```diff
@@ -25,7 +25,7 @@
 }
 
 .headers {
-  color: var(--color-softblack);
+  color: var(--color-default);
```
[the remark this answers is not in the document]
jadzeidan: Similar to https://github.com/digitalbitbox/bitbox-wallet-app/pull/1785#discussion_r1010256329

Changing it back to `softblack` would require a new font variable. To me it looks good with `default` so I would keep it like this.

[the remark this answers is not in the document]
jadzeidan: removed line

jadzeidan:
@thisconnect I think I addressed all the comments. PTAL

PR #1786, opened 2022-10-04
android: target api level 31, enforced by Google Play

benma:
According to
https://support.google.com/googleplay/android-developer/answer/11926878, the requirements as of now is that updating an app requires API level 31 (Android 12) by November 1st 2022.

The exported attribute is set according to
https://developer.android.com/about/versions/12/behavior-changes-12#exported:

> If your app targets Android 12 or higher and contains activities,
services, or broadcast receivers that use intent filters, you must explicitly declare the android:exported attribute for these app components.


Beerosagos (approved):
tACK (on Android 11)

PR #1787, opened 2022-10-04
frontend: update package-lock.json

thisconnect:
This reduces the vulnerabilities reported by `npm audit`. Those are all npm dev dependencies and production modules should not be affected.

Deleted node_modules and package-lock.json and did a new clean
install running: `npm install --ignore-scripts`

opened so I don't forget but need to test this more.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,6 @@ jobs:
         with:
           submodules: recursive
         uses: actions/checkout@v2
-      - name: Enable caching
-        uses: actions/cache@v2
-        with:
-          key: ${{runner.os}}-test-lint
-          path: frontends/web/node_modules
       - name: Run CI script
         # The script also runs golang-ci but it's ok: doesn't take too long and may be useful
         # to keep its linter errors in this log, too.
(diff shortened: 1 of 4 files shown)


benma:
> opened so I don't forget but need to test this more.

fyi, you can use 'Draft' instead of PR when a PR is not ready

thisconnect:
it is ready, I just wanted to review/test before.

> fyi, you can use 'Draft' instead of PR when a PR is not ready

thisconnect:
Yes changing to `npm ci --ignore-scripts` would make sense, but it deletes node_modules everytime. So in that case the CI should not cache node_modules anymore. agree? 

see node_modules is cached in various places in https://github.com/digitalbitbox/bitbox-wallet-app/blob/ff6fada72d50325c2944723bd72922d22d17b8bc/scripts/travis-ci.sh#L57

> Didn't we also mean to change npm install to npm ci to force using the lockfile and auto-deleting node_modules upfront, for better safety?

thisconnect:
Update package-lock.json again and rebased.

thisconnect:
rebased

thisconnect:
@benma PTAL

benma (approved):

PR #1789, opened 2022-10-04
android: fix timezone so the BitBox02 shows the correct time

benma:
On Android, due to a bug in
Go (https://github.com/golang/go/issues/20455), the local timezone is not correct. It is always UTC.

As a result, when restoring a backup on the BitBox02, the BitBox02 displays the current time in UTC instead of the local time.

This commit contains a workaround to fix the timezone on Android.

Alternative considered: getting the timezone offset in native Android code and sending it to Go-land via backend.Environment. I did not pursue this as a quick search didn't turn up an easy way to get the timezone offset on Android which takes into account daylight-savings etc.


Beerosagos (approved):
tACK on Android 11

PR #1734, opened 2022-07-28
Frontend refactor account

thisconnect:
I don't like the last commit, had to pass the dependency array to the hook otherwise the sdcard check runs only once on mount.

@benma wdyt?

PR #1790, opened 2022-10-05
frontend: refactor account to functional component

thisconnect:
Ongoing effort to rewrite all class based components to functional components so we can use hooks.
    
- typed check-sdcard api
- typed buy-supported api
- remove load decorator and load config directly

also: remove old account type information
    
This was used before unified accounts were introduced. Removing.


benma (approved):
utACK (I did test https://github.com/digitalbitbox/bitbox-wallet-app/pull/1734 yesterday however)

PR #1791, opened 2022-10-05
backend/conversions: fix nil conversions when rates not available

Beerosagos:
Unexpected errors from coingecko service were causing the backend to send amounts with nil conversions field, which caused runtime errors on the frontend. With this update, a != nil conversions field should always be available even when coingecko errors accour, allowing the frontend to handle the situation without blocks.


benma (approved):
LGTM

PR #1796, opened 2022-10-12
frontend/guide/entry: convert to FunctionalComponent

benma:
The entry property was also changed from `EntryProp | string` to
`EntryProp`, as the locize in-app editor (`cimode`) does not exist
anymore (removed in 29a0764e).


Beerosagos (approved):
tACK

PR #1798, opened 2022-10-19
frontend: fix coin and unit styling in tx overview

thisconnect:
Sometimes the unit breaks onto a separate line, this happens with
large amounts or coins with lots of decimal for example with ETH.

Changed styling so that very large amounts are cut off with "…",
but only visually so that copy/paste contains takes the whole
amount. This should also fix the issue with large amounts on
small screen, in extreme cases the fiat amount should squeeze
the coin amount together.

Also added "font-variant: tabular-nums" definition for the amount
so that all numbers take equal space, without using a monospace
font.

Screen Shot 2022-10-19 at 13 51 03

and small screen:

Screen Shot 2022-10-19 at 14 33 16

### DIFF
--- a/frontends/web/src/components/transactions/transaction.module.css
+++ b/frontends/web/src/components/transactions/transaction.module.css
@@ -51,7 +51,19 @@
 }
 
 .fiat {
+    font-variant: tabular-nums;
+}
 
+.amount {
+    font-variant: tabular-nums;
+    max-width: 100%;
+}
+
+.amountOverflow {
+    overflow: hidden;
+    padding-right: 40px;
+    position: relative;
+    text-overflow: ellipsis;
 }
 
 .send {
@@ -63,11 +75,18 @@
 }
 
 .currency {
-
+    font-variant: tabular-nums;
 }
 
 .currencyUnit {
     color: var(--color-secondary);
+    font-variant: normal;
+}
+
+.amountOverflow .currencyUnit {
+    position: absolute;
+    right: 0;
+    top: 0;
 }
 
 .action {
@@ -274,4 +293,12 @@
     .fee {
         min-width: 0;
     }
+}
+
+@media (max-width: 560px) {
+
+    .status {
+        display: none;
+    }
+
 }
--- a/frontends/web/src/components/rates/rates.module.css
+++ b/frontends/web/src/components/rates/rates.module.css
@@ -1,7 +1,8 @@
 .rates {
     cursor: default;
-    margin: 0;
     line-height: 1;
+    margin: 0;
+    white-space: nowrap;
 }
 
 .unit {
@@ -12,6 +13,7 @@
 .unitAction {
     color: var(--color-secondary);
     font-size: var(--size-default);
+    font-variant: normal;
     position: relative;
 }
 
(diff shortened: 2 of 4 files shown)


frontends/web/src/components/transactions/transaction.module.css
```diff
@@ -63,11 +75,18 @@
 }
 
 .currency {
-
+    font-variant: tabular-nums;
```
shonsirsha, line 78: nice!

frontends/web/src/components/transactions/transaction.module.css
```diff
@@ -51,7 +51,19 @@
 }
 
 .fiat {
+    font-variant: tabular-nums;
+}
 
+.amount {
+    font-variant: tabular-nums;
```
shonsirsha, line 58: nice!

shonsirsha (approved):
Looks good!! Personally, I would approve it. 

However, there's a small question from my side: what's the likelihood of having an (extra) long eth balance like somewhere 35-40 digits or more? Because it starts becoming messy at that point in smaller screens.

Pixel 5 screen width (393px):
Screen Shot 2022-10-26 at 07 13 04
eth value is ~35-40 sig. figures

Tablet (?) screen width (540px):
Screen Shot 2022-10-26 at 07 13 44
eth value is ~35-40 sig. figures

 (i changed the eth value directly from dev tool)

If this is a real concern, I think we could define a pixel value for the `max-width` of `.amount` (also could define this for each breakpoint).

This also guarantees that there'll always be the same amount of minimum spacing between the status and the fiat amount, so everything looks neater.

But if for some reason the eth value has been first formatted to display only the first 35 digits or less, then I think we're clear.

thisconnect:
bitcoin has a max of 21m with 8 decimals and ethereum 18 decimals and not sure about the maximum.

> (i changed the eth value directly from dev tool)

Ah this explains why the BTC value starts with a 0 :)

If there is not enough room, the basecurrency/fiat value should show in full and push the coin so that it shortens with `…` at the end so it hides some decimals.

On 540px it shows the word "complete" (meaning the block was mined and has at least 6 confirmations). Could also hide "complete" on this breakpoint. 

> If this is a real concern, I think we could define a pixel value for the max-width of .amount (also could define this for each breakpoint).

one issue with fixed pixel is that the font renders differently on Linux, macOS and Windows, so I am not sure about what a good max-width would be.

I'll try to hide the word "complete" up to the next breakpoint as without it should self-regulate (i.e. fiat pushes coin amount and that shortens with `…`)

thisconnect:
The first screenshot you made, I guess the green completed progress-ring is very close to the fiat value, but I didn't want to completely change the whole design. This change should just shorten the coin value if there is not enough place

> Pixel 5 screen width (393px):

To make the whole list nicer, I think I would remove the spacing on the left/right and decrease the one between the transaction, but not in this PR. But then the whole transaction layout could be overhauled.

This would look neater and give more space.

Screen Shot 2022-10-26 at 08 51 51

thisconnect:
rebased + changed breakpoint when the word "complete" is hiding.

This is a fake transaction with 100m ETH and realistic fiat amount:

Screen Shot 2022-10-26 at 09 09 32

PTAL

shonsirsha:
>> one issue with fixed pixel is that the font renders differently on Linux, macOS and Windows, so I am not sure about what a good max-width would be.

Ah, didnt know this. Thanks for the info!
Looks good to me, LGTM 🙂

PR #1799, opened 2022-10-20
frontend: show point-to-device graphic on pairing step

thisconnect:
Before there was a disabled grayed out continue button on the pairing step when the user should focus on the device and confirm the code.

Changed so that the is a point-to-device graphic without a disabled button. Once the pairing is confirmed it changes to a continue button.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -516,13 +516,17 @@ class BitBox02 extends Component<Props, State> {
               )}
             </ViewContent>
             <ViewButtons>
-              <Button
-                primary
-                onClick={() => this.channelVerify(true)}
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -516,13 +516,18 @@ class BitBox02 extends Component<Props, State> {
...
-              </Button>
+              {deviceVerified ? (
+                <Button
+                  primary
+                  onClick={() => this.channelVerify(true)}
+                  hidden={status === 'pairingFailed'}>
+                  {t('button.continue')}
+                </Button>
+              ) : (
+                status !== 'pairingFailed' ? (
```
benma, line 527: could move this to the top and put everything in this condition (both the continue button and the PointToBitBox02)

benma (approved):
tACK

PR #1800, opened 2022-10-24
frontend: converted SDCardCheck to a functional component

shonsirsha:
As a part of the refactoring process and to follow the latest conventions in the frontend, class components need to be converted into functional components whenever possible.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
+++ b/frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
@@ -14,80 +14,64 @@
  * limitations under the License.
  */
 
-import { Component } from 'react';
-import { translate, TranslateProps } from '../../../decorators/translate';
+import { FunctionComponent, useCallback, useEffect, useState } from 'react';
+import { TranslateProps } from '../../../decorators/translate';
 import { Dialog, DialogButtons } from '../../../components/dialog/dialog';
-import { apiGet } from '../../../utils/request';
 import { Button, ButtonLink } from '../../../components/forms';
+import { checkSDCard } from '../../../api/bitbox02';
+import { useTranslation } from 'react-i18next';
 
-interface SDCardCheckProps {
+type SDCardCheckProps = {
     deviceID: string;
 }
 
-interface State {
-    sdCardInserted?: boolean;
-}
-
-
 type Props = SDCardCheckProps & TranslateProps;
 
-class SDCardCheck extends Component<Props, State> {
-  state = {} as State;
+const SDCardCheck: FunctionComponent<Props> = ({ deviceID, children }) => {
+  const { t } = useTranslation();
+  const [sdCardInserted, setSdCardInserted] = useState<boolean | undefined>();
+  const check = useCallback(() => checkSDCard(deviceID).then(setSdCardInserted), [deviceID]);
 
-  public componentDidMount() {
-    this.check();
+  useEffect(() => {
+    check();
+  }, [check]);
+
+
+  // pending check-sdcard request
+  if (sdCardInserted === undefined) {
+    return null;
   }
 
-  private check = () => {
-    apiGet('devices/bitbox02/' + this.props.deviceID + '/check-sdcard')
-      .then(inserted => this.setState({ sdCardInserted: inserted }));
-  };
-
-  public render() {
-    const {
-      t,
-      children,
-      deviceID,
-    } = this.props;
-    const { sdCardInserted } = this.state;
-
-    // pending check-sdcard request
-    if (sdCardInserted === undefined) {
-      return null;
-    }
-    if (!sdCardInserted) {
-      return (
-        <Dialog title="Check your device" small>
-          <div className="columnsContainer half">
-            <div className="columns">
-              <div className="column">
-                <p>{this.props.t('backup.insert')}</p>
-              </div>
+  if (!sdCardInserted) {
+    return (
+      <Dialog title="Check your device" small>
+        <div className="columnsContainer half">
+          <div className="columns">
+            <div className="column">
+              <p>{t('backup.insert')}</p>
             </div>
           </div>
-          <DialogButtons>
-            <Button
-              primary
-              onClick={this.check}>
-              {t('button.ok')}
-            </Button>
-            <ButtonLink
-              transparent
-              to={`/device/${deviceID}`}>
-              {t('button.back')}
-            </ButtonLink>
-          </DialogButtons>
-        </Dialog>
-      );
-    }
-    return (
-      <div>
-        {children}
-      </div>
+        </div>
+        <DialogButtons>
+          <Button
+            primary
+            onClick={check}>
+            {t('button.ok')}
+          </Button>
+          <ButtonLink
+            transparent
+            to={`/device/${deviceID}`}>
+            {t('button.back')}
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
```diff
@@ -14,80 +14,68 @@
...
 
-  public componentDidMount() {
-    this.check();
-  }
+  const [sdCardInserted, setSdCardInserted] = useState<boolean | undefined>(undefined);
+
+  useEffect(() => {
+    check();
+    // no dependencies passed - only gets executed once on mount
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
thisconnect, line 36: Shouldn't the effect re-run if `deviceID` changes, I guess useCallback, but see my later comment about `useLoad`.

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
```diff
@@ -14,80 +14,68 @@
...
 
-  private check = () => {
-    apiGet('devices/bitbox02/' + this.props.deviceID + '/check-sdcard')
-      .then(inserted => this.setState({ sdCardInserted: inserted }));
+  const check = () => {
+    apiGet('devices/bitbox02/' + deviceID + '/check-sdcard')
+      .then(inserted => setSdCardInserted(inserted));
   };
```
thisconnect, line 41: Could you try with our custom hook `useLoad` instead?

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
```diff
@@ -14,80 +14,68 @@
...
-
-}
+  return (
+    <div>
+      {children}
+    </div>
+  );
+};
 
 const HOC = translate()(SDCardCheck);
```
thisconnect, line 80: Remove the HOC and replace with `useTranslation` hook.

thisconnect (changes requested):
nice thanks 👍, added some comments.

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
```diff
@@ -14,80 +14,68 @@
...
 type Props = SDCardCheckProps & TranslateProps;
 
-class SDCardCheck extends Component<Props, State> {
-  state = {} as State;
+const SDCardCheck: FunctionComponent<Props> = ({ t, deviceID, children }) => {
 
-  public componentDidMount() {
-    this.check();
-  }
+  const [sdCardInserted, setSdCardInserted] = useState<boolean | undefined>(undefined);
```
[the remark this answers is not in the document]
shonsirsha: I put `undefined` as a default value just so it's clear at first glance what the state of `sdCardInserted`is. And yeah no real difference between what you shown and what I have, so I could remove it. 👍

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
shonsirsha: Didn't get the impression that it should re-run when `deviceID` changes, but anyhow I agree that useLoad is cleaner!

However, if we were to use `useLoad` I wonder what should the `onClick` be for the "ok button"? Because atm, it's firing `check` which affects the local state, which affects the conditional rendering. 

Using `useLoad`, we won't need to use the local state (`inserted`) anymore as the hook returns the value already, but this means we can't control the state on a button click, like we have right now.

Ofc the `onClick` could be just checkSDCard() (API call), but that doesn't really do anything in regards to the UI rendering - so it's pointless at that point, or we could use useEffect+useCallback but this seems like it just overcomplicates things without adding real value. Example:

```js
  const [inserted, setInserted] = useState<boolean | undefined>();
  const checkInserted = useLoad(() => checkSDCard(deviceID), [deviceID]);
  const check = useCallback(() => checkSDCard(deviceID).then(inserted => setInserted(inserted)), [deviceID]);

  useEffect(() => {
    setInserted(checkInserted);
  }, [checkInserted]);
```

In this case, `onClick` will be `check` and we use `inserted` for the conditional rendering instead of the returned value from `useLoad` (`checkInserted`) . 😕

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
shonsirsha: See comment above

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
shonsirsha: I changed it to use useCallback + useEffect which still uses the local useState (without useLoad). Please take a look, otherwise we could use useLoad but question above still stands: what should the onClick be for the "ok button" if useLoad is being used (do we need an ok button at all then)?  🤔

thisconnect: You are right `useLoad` doesn't work with clicking the button, the way you have it with useCallback + useEffect is good 👍 .

> Didn't get the impression that it should re-run when deviceID changes, but anyhow I agree that useLoad is cleaner!

In practice right now deviceID would not change as we only have support for 1 BB02 at the time, but if we one day allow multiple, it may change. I believe now it's better as it would create a new callback if the id changes. Before the check would only have been executed once. I believe now it is good and the effect runs again if there is a new check function / id changed.

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
```diff
@@ -14,80 +14,64 @@
...
-
-
 type Props = SDCardCheckProps & TranslateProps;
 
-class SDCardCheck extends Component<Props, State> {
-  state = {} as State;
+const SDCardCheck: FunctionComponent<Props> = ({ deviceID, children }) => {
+  const { t } = useTranslation();
+  const [sdCardInserted, setSdCardInserted] = useState<boolean | undefined>();
+  const check = useCallback(() => checkSDCard(deviceID).then(ins => setSdCardInserted(ins)),[deviceID]);
```
thisconnect, line 33: Just pass the setter?

```diff
diff --git a/frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx b/frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
index 6f50dfb3..4fb9a2ad 100644
--- a/frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
+++ b/frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
@@ -30,7 +30,7 @@ type Props = SDCardCheckProps & TranslateProps;
 const SDCardCheck: FunctionComponent<Props> = ({ deviceID, children }) => {
   const { t } = useTranslation();
   const [sdCardInserted, setSdCardInserted] = useState<boolean | undefined>();
-  const check = useCallback(() => checkSDCard(deviceID).then(ins => setSdCardInserted(ins)),[deviceID]);
+  const check = useCallback(() => checkSDCard(deviceID).then(setSdCardInserted), [deviceID]);
 
   useEffect(() => {
     check();

```

Also small nit, please add space after comma, I'll open a PR with space eslint rule later.

thisconnect (approved):
LGTM with small nit

PR #1801, opened 2022-10-24
frontend: use generic continue button translation

thisconnect:
The same translation for 'Continue' is used in various components,
changed to use the same generic text everywhere instead of custom
translation key.

Note: I kept different commits so that it is easier to review, but I'll squash
into 1 commit once approved.


shonsirsha (approved):
looks good 🙂 lgtm

PR #1805, opened 2022-10-26
backend: add a BtcUnit type and use it everywhere to format sats

benma:
Instead of 'BTC' | 'sat', we use 'default' | 'sat', to avoid confusion with BTC coin units which also include 'TBTC' for testnet.

This helps with readability of the code, as the type and possible values are clear at every step.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -120,7 +120,7 @@ class Send extends Component<Props, State> {
     noMobileChannelError: false,
     fiatUnit: fiat.state.active,
     coinControl: false,
-    btcUnit : 'BTC',
+    btcUnit : 'default',
     activeCoinControl: false,
     hasCamera: false,
     activeScanQR: false,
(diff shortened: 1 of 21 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -608,7 +608,7 @@ class Send extends Component<Props, State> {
         {t('send.signprogress.label')}: {signProgress.step}/{signProgress.steps}
       </span>
     ) : undefined;
-    const baseCurrencyUnit = fiatUnit === 'BTC' ? btcUnit : fiatUnit;
+    const baseCurrencyUnit: accountApi.Fiat | 'sat' = fiatUnit === 'BTC' && btcUnit === 'sat' ? 'sat' : fiatUnit;
```
Beerosagos, line 611: nit: could use 
```suggestion
    const baseCurrencyUnit: accountApi.ConversionUnit = fiatUnit === 'BTC' && btcUnit === 'sat' ? 'sat' : fiatUnit;
```

Beerosagos (approved):
tACK!

PR #1806, opened 2022-10-26
frontend/bitbox02/settings: handle attestation==null

benma:
The backend can in theory return nil/null in the attestation endpoint, if the endpoint is called before the device.Init() function has finished, where the attesation check is performed.

The bb02 init is a bit messy, in the above state probably most/all other bb02 calls would be broken too. For now, we simply adjust the return type of `verifyAttestation()` to match the backend.


thisconnect (approved):
utACK LGTM 👍

PR #1807, opened 2022-10-26
backend/bitbox02: format BTC as sats if the setting is turned on

benma:
Also bitbox02-api-go is updated to make use of the new protobuf messages:

go get github.com/digitalbitbox/bitbox02-api-go@7b5da12 go mod vendor
go mod tidy


Beerosagos (approved):
tACK

PR #1808, opened 2022-10-26
frontend: add guide entry for sats

thisconnect:
Added 'What is a Satoshi' guide entry on the settings page, as not all users might know what it is.

Also reordered settings entry and moved which exchange rates up as second entry.

Screen Shot 2022-10-26 at 14 47 57

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -793,6 +793,10 @@
       }
     },
     "settings": {
+      "sats": {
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -793,6 +793,10 @@
       }
     },
     "settings": {
+      "sats": {
+        "text": "A Satoshi ('sats' for short) is the smallest unit of Bitcoin. One Satoshi is a hundred millionth of a bitcoin (0.00000001 BTC). It is named after the creator of Bitcoin, Satoshi Nakamoto.",
```
Beerosagos, line 797: Maybe better to use 'sat' instead of 'sats', since it is the abbreviation used everywhere in the app?

thisconnect: fixed thanks

Beerosagos (approved):
tACK!

PR #1809, opened 2022-10-26
backend/usb/hid: prepend 0 byte to packet

benma:
If a USB packet starts with a zero byte, that byte will be stripped by the C USB library (i.e. signal11 on Linux), leading to a broken package.

We only need to add it if not on Windows, as karalabel/usb prepends this zero byte already for Windows (I think this is a bug in karalabe/usb and karalabe/hid, which didn't consider that a packet could start with a zero byte right after the report ID).

Currently we don't observe a bug because the first byte in each packet is the `cid` (channel identifier), which is currnetly harddcoded to `0xff000000` in bitbox02-api-go. If we start using a random CID, and it starts with a zero byte, the packet will be corrupted and the BitBox02 will not be able to parse it.

The same fix was also added to the bitbox01 in the BitBoxApp (https://github.com/digitalbitbox/bitbox-wallet-app/blob/bd70f6c400268355cee94891397447796f02fbde/backend/devices/bitbox/communication.go#L235-L241), where it was crucial because packets in the bitbox01 are padded with zero bytes to the left. We can remove the fix there, as it uses the same hid functions as the BitBox02.

### DIFF
--- a/backend/devices/bitbox/communication.go
+++ b/backend/devices/bitbox/communication.go
@@ -20,7 +20,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"io"
-	"runtime"
 	"unicode"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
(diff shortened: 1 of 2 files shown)


benma:
I need to test this on bitbox01 (bootloader, firmware) and bitbox02 (bootloader, firmware) on all supported platforms.

PR #1811, opened 2022-10-27
frontend/account: call onAccountChanged after init/onStatusChanged

benma:
Before this change, onAccountChanged was called immediately when the
account was switched and the code changed, as it depended on the code,
which changed the callback, which led to it being called due to this line:

`useEffect(onAccountChanged, [onAccountChanged, status]);`

Calling this callback before `accountApi.init()` is incorrect and
leads to an error.

Instead of depending on the code and status, we pass it as params
instead.

We also await the result of `init()`, as technically it is also async
and could finis after `onAccountChanged()` was called.

### DIFF
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -114,8 +114,6 @@ export function Account({
       .catch(console.error);
   }, [onAccountChanged, code]);
 
-  useEffect(onStatusChanged, [onStatusChanged, code]);
-
   const subscriptions = useRef<UnsubscribeList>([]);
   useEffect(() => {
     unsubscribe(subscriptions.current);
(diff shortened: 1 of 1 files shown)


thisconnect (approved):
tested with `rm -rf appfolder.dev/cache/account-* &&  make servewallet` accounts loaded nice with "scanned <n> addresses" + no more error.

maybe squash both commits into 1?

LGTM

Beerosagos (approved):
very nice! tACK :+1:

benma:
> maybe squash both commits into 1?

Why? They are totally independent. If you needed to revert one fix, the other should not also be reverted.

thisconnect:
ack

> They are totally independent. If you needed to revert one fix, the other should not also be reverted.

PR #1812, opened 2022-10-27
frontend: add options to navigate on add-account

thisconnect:
The add-account success view only has 1 button 'Done' to continue that goes directly to the newly created account. This makes adding multiple accounts a bit tedious becuase the user needs to navigate back to settings -> manage accounts -> add account.

- added a back button on select-currency (1st step of add-account) that goes back to the manage-accounts view
- added a 'Add another account' secondary button on the succeess step of add-account to easily add multiple accounts


shonsirsha (approved):
LGTM 👍

PR #1814, opened 2022-10-31
backend,frontend: remove Ropsten/Rinkeby

benma:
Was replaced by the Goerli testnet. Ropsten/Rinkeby have been shut down on Oct. 5th 2022.


backend/backend.go
```diff
@@ -371,31 +368,12 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
...
 			"https://goerli.etherscan.io/tx/",
 			etherScan,
 			nil)
-	case code == coinpkg.CodeERC20TEST:
-		etherScan := etherscan.NewEtherScan("https://api-ropsten.etherscan.io/api", backend.etherScanHTTPClient)
-		coin = eth.NewCoin(etherScan, code, "ERC20 TEST", "TEST", "TETH", params.RopstenChainConfig,
-			"https://ropsten.etherscan.io/tx/",
-			etherScan,
-			erc20.NewToken("0x2f45b6fb2f28a73f110400386da31044b2e953d4", 18),
-		)
```
thisconnect, line 398: Was this for some kind of test?

Not needed anymore?

backend/coins/eth/coin_test.go
```diff
@@ -46,7 +46,7 @@ func (s *testSuite) SetupTest() {
 
 	s.ERC20Coin = NewCoin(
 		nil,
-		coin.CodeERC20TEST,
+		"ERC20TEST",
```
thisconnect, line 49: In other places you removed something about ERC20Test... is this still used?

thisconnect (approved):
LGTM with question.

backend/backend.go
benma: I once deployed (I think? can't quite remember) a ERC20 token for manual testing. Haven't used it in years.

backend/coins/eth/coin_test.go
benma: It is for unit testing erc20 token stuff. The name here doesn't really matter, same as the fake contract address below it.

Beerosagos (approved):
tACK. One question: I noticed this https://github.com/benma/bitbox-wallet-app/blob/rem-testnets/backend/config/config.go#L77-L79 I am not sure what is that for, can we delete it?

benma:
> tACK. One question: I noticed this https://github.com/benma/bitbox-wallet-app/blob/rem-testnets/backend/config/config.go#L77-L79 I am not sure what is that for, can we delete it?

I left it on purpose so we don't accidentally use the keys again in the future

>	// Removed in v4.35 - don't reuse these two keys.

PR #1815, opened 2022-11-01
frontend: fixed download cert, check, and add buttons

shonsirsha:
The ‘download cert’ and ‘check’ buttons were not showing the loading animation when they were  supposed to. The problem
was due to the buttons' incorrect styling. It caused the animation to look invisible.

To fix the issue, their style has been altered to be the same as the buttons at the top of the page (the ‘Check’ and
‘Remove’ buttons). The `disabled` conditions for each button have also  been slightly refactored to increase its robustness
& readability which also fixed an existing bug.

Screen capture of the **now working (visible)** animation:

https://user-images.githubusercontent.com/28676406/199158463-05b41792-6aa4-4716-8fb0-d8fb920f41ab.mov

### DIFF
--- a/frontends/web/src/routes/settings/electrum-add-server.tsx
+++ b/frontends/web/src/routes/settings/electrum-add-server.tsx
@@ -73,6 +73,10 @@ export const ElectrumAddServer = ({
     setLoadingCheck(false);
   };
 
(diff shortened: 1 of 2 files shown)


benma (approved):
excellent, thanks! utACK

PR #1816, opened 2022-11-01
frontend/bitbox02/checkbackup: fix async bug

benma:
If the first checkBackup() call has a delay in the backend and the
second checkBackup() call starts before the first handles the
response, the dialog will be empty. I encountered this bug in the wild
recently.

Awaiting the first response before requesting the second fixes this.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/checkbackup.tsx
+++ b/frontends/web/src/routes/device/bitbox02/checkbackup.tsx
@@ -18,6 +18,7 @@ import { Component } from 'react';
 import * as bitbox02API from '../../../api/bitbox02';
 import { translate, TranslateProps } from '../../../decorators/translate';
 import { Backup, BackupsListItem } from '../components/backup';
+import { alertUser } from '../../../components/alert/Alert';
 import { Dialog, DialogButtons } from '../../../components/dialog/dialog';
 import { Button } from '../../../components/forms';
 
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/device/bitbox02/checkbackup.tsx
```diff
@@ -44,46 +45,39 @@ class Check extends Component<Props, State> {
...
-        this.setState({ foundBackup });
-        message = this.props.t('backup.check.success');
-      } else {
-        message = this.props.t('unknownError', { errorMessage: 'Not found' });
+  private checkBackup = async () => {
+    try {
+      const backupID = await bitbox02API.checkBackup(this.props.deviceID, true);
+      const foundBackup = this.props.backups.find((backup: Backup) => backup.id === backupID);
+      if (!foundBackup) {
+        alertUser(this.props.t('unknownError', { errorMessage: 'Not found' }));
```
thisconnect, line 53: The alertUser global component is a bit a react antipattern, why not reuse the dialog that is here?

for example:

```js
this.setState({
  activeDialog: true,
  userVerified: true,
  message: this.props.t('unknownError', { errorMessage: 'Not found' })
});
```

but yes, `userVerified` doesn't make too much sense. But this could be renamed from `userVerified` to `userVerify`

thisconnect (approved):
Nice 👍 

LGTM with proposal to not use alertUser

frontends/web/src/routes/device/bitbox02/checkbackup.tsx
benma: @thisconnect I chose it because the alert looks just like a dialog, but the message is in the body instead of in the title. I had your proposed code before and the message was in the dialog title with an empty body, which looks strange.

Why is it an antipattern?

thisconnect: ack, it's fine we use it in various places..

This is what we should change one day: 

At the moment alertUser is a variable in a global scope pointing to a function of the <Alert /> instance. This "works" 
but is a bit a hack and not react style.

https://github.com/digitalbitbox/bitbox-wallet-app/blob/453eb3935c114408d28051b726714a7e4b1806f1/frontends/web/src/components/alert/Alert.tsx#L49

PR #1817, opened 2022-11-01
frontend: add manage-device guide

thisconnect:
Added a guide to manage device with entries about:

- BitBox02 name
- Secure chip
- Attestation / authenticity check

Screen Shot 2022-11-01 at 11 44 09


benma (approved):
tACK

PR #1820, opened 2022-11-01
backend/btc: make ScriptHashGetHistory synchronous

benma:
This will make it simpler to propagate and handle errors. It is also
easier to read.

There is no loss of performance, as this call already happens in a
goroutine as part of the callback to `ScriptHashSubscribe()`.


Beerosagos (approved):
tACK

PR #1821, opened 2022-11-02
frontend: improved portfolio chart on mobile

shonsirsha:
portfolio chart on mobile looked like it had uneven spacings
and whitespaces which made it not centered. This is due to 
the vertical label (y-axis) drawn by the library.

This commit is inteded to fix that chart’s UI by partly re-styling
it for mobile, where the vertical label and side paddings are
removed.

To interact with the chart on mobile, long-press on the chart and
move around the chart. To stop interacting, users can simply lift
their fingers (stop touching the chart).

(Updated) screenshots of the chart:

Screen Shot 2022-11-02 at 15 03 40
Screen Shot 2022-11-02 at 15 05 51

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.module.css
+++ b/frontends/web/src/routes/account/summary/chart.module.css
@@ -10,11 +10,13 @@
 
 .chartUpdatingMessage {
   align-items: center;
+  color: var(--color-gray-alt);
   display: flex;
   flex-direction: row;
   justify-content: center;
   text-align: center;
 }
+
 
 .chart header {
   display: flex;
@@ -23,7 +25,12 @@
   min-height: 68px;
 }
 
-.summary {}
+.summary {
+  margin: 0 auto;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+}
 
 .filters {
   align-items: baseline;
@@ -34,16 +41,42 @@
   margin: var(--spacing-half) 0 0 0;
 }
 
+@media (max-width: 640px){
+
+  .chart{
+    margin-bottom: var(--spacing-default);
+  }
+
+  .summary{
+    align-items: center;
+  }
+
+  .filters{
+    justify-content: center;
+    margin-top: var(--spacing-default);
+  }
+}
+
+
 .filters button {
   appearance: none;
   background: var(--color-white);
   border: 2px solid var(--color-white);
   border-radius: 2rem;
   color: var(--color-softblack);
+  font-size: var(--size-default);
   line-height: 1.75;
   margin-bottom: var(--spacing-half);
   margin-left: var(--spacing-half);
   padding: 0 calc(var(--spacing-default) * .75);
+}
+
+.filters button:hover:not([disabled]) {
+  cursor: pointer;
+}
+
+.filters button:hover:is([disabled]) {
+  cursor: not-allowed;
 }
 
 .filters button:focus {
@@ -99,6 +132,14 @@
 .chartCanvas {
   position: relative;
 }
+
+@media (max-width: 640px){
+  .chartCanvas {
+    width: 100vw;
+    margin-left: calc(var(--space-half) * -1);
+  } 
+}
+  
 
 .tooltip {
   background: white;
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -22,6 +22,8 @@ import { Skeleton } from '../../../components/skeleton/skeleton';
 import { formatNumber } from '../../../components/rates/rates';
 import { bitcoinRemoveTrailingZeroes } from '../../../utils/trailing-zeroes';
 import styles from './chart.module.css';
+import Filters from './filters';
+import { TChartDisplay, TChartFiltersProps } from './types';
 
 export interface FormattedLineData extends LineData {
   formattedValue: string;
@@ -34,21 +36,22 @@ type ChartProps = {
     noDataPlaceholder?: ReactChild;
 };
 
-interface State {
-    display: 'week' | 'month' | 'year' | 'all';
-    source: 'daily' | 'hourly';
-    difference?: number;
-    diffSince?: string;
-    toolTipVisible: boolean;
-    toolTipValue?: string;
-    toolTipTop: number;
-    toolTipLeft: number;
-    toolTipTime: number;
+type State = {
+  display: TChartDisplay;
+  source: 'daily' | 'hourly';
+  difference?: number;
+  diffSince?: string;
+  toolTipVisible: boolean;
+  toolTipValue?: string;
+  toolTipTop: number;
+  toolTipLeft: number;
+  toolTipTime: number;
+  isMobile: boolean;
 }
 
 type Props = ChartProps & TranslateProps;
 
-interface FormattedData {
+type FormattedData = {
   [key: number]: string;
 }
 
@@ -59,6 +62,7 @@ class Chart extends Component<Props, State> {
   private lineSeries?: ISeriesApi<'Area'>;
   private resizeTimerID?: any;
   private height: number = 300;
+  private mobileHeight: number = 150;
   private formattedData?: FormattedData;
 
   static readonly defaultProps: ChartProps = {
@@ -81,10 +85,11 @@ class Chart extends Component<Props, State> {
     toolTipTop: 0,
     toolTipLeft: 0,
     toolTipTime: 0,
-
+    isMobile: false,
   };
 
   public componentDidMount() {
+    this.checkIfMobile();
     this.createChart();
   }
 
@@ -128,13 +133,19 @@ class Chart extends Component<Props, State> {
     });
   }
 
+  private checkIfMobile = () => {
+    this.setState({ isMobile: window.innerWidth <= 640 });
+  };
+
   private createChart = () => {
     const { data: { chartIsUpToDate, chartDataMissing } } = this.props;
     if (this.ref.current && this.hasData() && (chartIsUpToDate && !chartDataMissing)) {
       if (!this.chart) {
+        const chartWidth = !this.state.isMobile ? this.ref.current.offsetWidth : document.body.clientWidth;
+        const chartHeight = !this.state.isMobile ? this.height : this.mobileHeight;
         this.chart = createChart(this.ref.current, {
-          width: this.ref.current.offsetWidth,
-          height: this.height,
+          width: chartWidth,
+          height: chartHeight,
           handleScroll: false,
           handleScale: false,
           crosshair: {
@@ -155,7 +166,7 @@ class Chart extends Component<Props, State> {
             horzLines: {
               color: '#dedede',
               style: LineStyle.Solid,
-              visible: true,
+              visible: !this.state.isMobile,
             },
           },
           layout: {
@@ -167,7 +178,7 @@ class Chart extends Component<Props, State> {
           leftPriceScale: {
             borderVisible: false,
             drawTicks: false,
-            visible: true,
+            visible: !this.state.isMobile,
             entireTextOnly: true,
           },
           localization: {
@@ -180,6 +191,10 @@ class Chart extends Component<Props, State> {
           timeScale: {
             borderVisible: false,
             timeVisible: false,
+            visible: !this.state.isMobile,
+          },
+          trackingMode: {
+            exitMode: 0
           }
         });
       }
@@ -189,8 +204,8 @@ class Chart extends Component<Props, State> {
         priceFormat: {
           type: 'volume',
         },
-        topColor: 'rgba(94, 148, 192, 0.5)',
-        bottomColor: 'rgba(94, 148, 192, 0.02)',
+        topColor: '#DFF1FF',
+        bottomColor: '#F5F5F5',
         lineColor: 'rgba(94, 148, 192, 1)',
         crosshairMarkerRadius: 6,
       });
@@ -206,6 +221,7 @@ class Chart extends Component<Props, State> {
   };
 
   private onResize = () => {
+    this.checkIfMobile();
     if (this.resizeTimerID) {
       clearTimeout(this.resizeTimerID);
     }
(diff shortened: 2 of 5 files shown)


frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -128,12 +132,16 @@ class Chart extends Component<Props, State> {
     });
   }
 
+  private checkIfMobile = () => {
+    this.setState({ isMobile: window.innerWidth <= 640 });
+  };
+
```
shonsirsha, line 138: `640`px as this is the breakpoint that the page uses when the layout starts changing for "mobile". 

This fn is used to easily determine the screen size. This way we can dynamically change the chart's option without having to initialise/create a second chart only for mobile.

shonsirsha, line 138: `640`px as this is the breakpoint that the page uses when the layout starts changing for "mobile". 

This fn is used to easily determine the screen size. This way we can dynamically change the chart's option without having to initialise/create a second chart only for mobile. 🙂

Beerosagos:
I think it looks really nice! Tested it on Android 11, I am not able to tap on the chart line to get the specific date values. I checked and this is also not happening for me on the master branch, but I think this can be a bigger problem here since we are removing the chart axes.

Beerosagos:
I also noticed that rotating my screen the chart line doesn't update correctly. Again, this seems to happen also on the master branch, but maybe you can look into that while you are working on this? 
immagine

jadzeidan:
LGTM 🎉

shonsirsha:
> I think it looks really nice! Tested it on Android 11, I am not able to tap on the chart line to get the specific date values. I checked and this is also not happening for me on the master branch, but I think this can be a bigger problem here since we are removing the chart axes.

Ah, thanks! 🙂  yeah maybe I should've added this in the PR description as well.

1. To drag the chart on mobile, **you'll have to do a long-press then drag. To "stop" dragging, tap anywhere on the chart once.** This is called the **"tracking mode"** by the library. I unfortunately also can't find a way for it to be just dragged "normally" without a long-press on mobile. See [here](https://github.com/tradingview/lightweight-charts/issues/434#issuecomment-666924467). 

Maybe we should add something (on mobile) that tells the user that they have to long-press to explore the graph. Pinging @jadzeidan for his input 🙂 

> I also noticed that rotating my screen the chart line doesn't update correctly. 
2. Yeah I noticed this as well but thought it wasn't gonna be a problem since this existed in master and the screen normally doesn't get bigger - well it for sure does when we move to landscape! 😄  I will investigate this further on Monday when I am back, hopefully getting it fixed.  Thanks!

jadzeidan:
> To drag the chart on mobile, you'll have to do a long-press then drag. To "stop" dragging, tap anywhere on the chart once. This is called the "tracking mode" by the library. I unfortunately also can't find a way for it to be just dragged "normally" without a long-press on mobile. See https://github.com/tradingview/lightweight-charts/issues/434#issuecomment-666924467.

Thanks for the explanation @shonsirsha. The intention is that when you remove your finger from the screen, then "tracking more" is disabled. So there would be no need to tap the chart again to stop tracking. Is that possible?

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -14,16 +14,23 @@
   flex-direction: row;
   justify-content: center;
   text-align: center;
+  color: #BFBFBF;
```
thisconnect, line 17: isn't there any existing color in variables.css that could be used?

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -44,11 +68,17 @@
   margin-bottom: var(--spacing-half);
   margin-left: var(--spacing-half);
   padding: 0 calc(var(--spacing-default) * .75);
+  font-size: var(--size-default);
```
thisconnect, line 71: Could you order CSS properties alphabetically?

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -44,11 +68,17 @@
...
+}
+
+.filters button:hover {
+  cursor: pointer;
```
thisconnect, line 75: pointer also shows when the buttons are disabled, maybe add `:not([disabled])`?

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -44,11 +68,17 @@
...
 }
 
 .filters button:focus {
   border: 2px solid var(--color-blue);
   outline: none;
+
```
thisconnect, line 81: remove?

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -100,6 +130,14 @@
   position: relative;
 }
 
+@media (max-width: 640px){
+  .chartCanvas {
+    width: 100vw;
+    margin-left: -16px;
+  } 
+}
```
thisconnect, line 138: Could you try to other way around, have no padding for this view with `<div class="content noSpace">` and add padding/margin for elements that need it?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -35,15 +37,16 @@ type ChartProps = {
...
+  source: 'daily' | 'hourly';
+  difference?: number;
+  diffSince?: string;
+  toolTipVisible: boolean;
+  toolTipValue?: string;
+  toolTipTop: number;
+  toolTipLeft: number;
+  toolTipTime: number;
+  isMobile: boolean;
 }
```
thisconnect, line 50: Since you are here anyway please change interface to type.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -206,14 +216,28 @@ class Chart extends Component<Props, State> {
...
+    this.checkIfMobile();
     if (this.resizeTimerID) {
       clearTimeout(this.resizeTimerID);
     }
     this.resizeTimerID = setTimeout(() => {
       if (!this.chart || !this.ref.current) {
         return;
       }
-      this.chart.resize(this.ref.current.offsetWidth, this.height);
+      this.chart.resize(!this.state.isMobile ? this.ref.current.offsetWidth : window.outerWidth, !this.state.isMobile ? this.height : this.mobileHeight);
```
thisconnect, line 227: this is a bit hard to read, could you save the const first in width and height varibables?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -415,71 +441,62 @@ class Chart extends Component<Props, State> {
...
+              </span>
+            }
           </div>
+          {!isMobile && <Filters chartFiltersData={chartFiltersData} />}
         </header>
-        <div className={styles.chartCanvas} style={{ minHeight: `${this.height}px` }}>
+        <div className={styles.chartCanvas} style={{ minHeight: `${!isMobile ? this.height : this.mobileHeight}px` }}>
           {(!chartIsUpToDate || chartDataMissing) ? (
-            <div className={styles.chartUpdatingMessage} style={{ height: `${this.height}px` }}>
+            <div className={styles.chartUpdatingMessage} style={{ height: `${!isMobile ? this.height : this.mobileHeight}px` }}>
```
thisconnect, line 499: also a bit hard to read and some duplication with line 497, please save the height in a const before the render

frontends/web/src/routes/account/summary/filters.tsx
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+import { FunctionComponent } from 'react';
+import { useTranslation } from 'react-i18next';
+import { TChartFilters } from './types';
+import styles from './chart.module.css';
+
+
+type FilterProps = {
+  chartFiltersData: TChartFilters
```
thisconnect, line 24: the displayWeek, displayMonth, displayYear, displayAll, callbacks are not really part of the data, just type them here?

also maybe rename to onDisplayWeek so it is clear that it's a callback fn?

thisconnect (changes requested):
looks nice 👍 

some issue: On medium/small screen around 600px I sometimes get a long horizontal scrollbar.

frontends/web/src/locales/en/app.json
```diff
@@ -352,7 +352,7 @@
   },
   "chart": {
     "dataMissing": "Gathering historical data... stay tuned.",
-    "dataUpdating": "updating data…",
+    "dataUpdating": "Loading chart",
```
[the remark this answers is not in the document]
shonsirsha: Oops, haha. Thanks!

frontends/web/src/routes/account/summary/filters.tsx
shonsirsha: Unsure  what you meant by “type them here”? If you meant defining these fns inside `<Filter/>`, then it would be quite a hassle as the fns involved its parent's states, the chart settings/methods, etc so that means we have to pass all of these down to the `<Filter/>`

Imo, this becomes  (unnecessarily)  more complex.  I suggest that we should keep child components as “simple” or as “dumb” as possible, meaning we should try to keep these "complex" logic (functions) within the parent components,  and away from its children. This way it’s easier to test, refactor and re-use these child components.  

But I do agree with the naming! Was meaning to change that but totally forgot about it. I’ll also change the name to `chartFiltersProps` instead of `chartFiltersData`, as it’s true - these callback fns are not really “data”.

frontends/web/src/routes/account/summary/chart.module.css
shonsirsha: Hm.. I'm not sure if that's the correct solution here. 🤔 

Atm, the chart (along with everything inside the page) is positioned within a container `content padded`. This means, they are inside a container with top, left, and right padding. As we want the chart to be end-to-end (fill up the whole screen horizontally), we don't want this left and right padding, thus we added margin-left of -16px to the chart. I've now changed this margin value to use the css variable: `calc(var(--space-half)*-1)`, negating whatever is the current left and right padding of the page's container (which is just `16px`).

**Now, the chart itself already has no padding and margin, so adding `content noSpace` wont change anything.** We could use this `content noSpace` for the page's container, but this will introduce too changes to other components within the page (adding margin/padding to them).

So I think adding a negative margin **to the chart** is the simplest / imo cleanest solution for it right now. 🙂 pls lmk what you think or if I misunderstood you!

shonsirsha:
@jadzeidan Yes, that’s possible, please feel free to check the latest changes! :)

On another note, I was referring to the fact that any user has to first, long-press, to be able to see values (and move around the graph), and I think this isn’t really intuitive.

I think it’d be beneficial to let the user know that they have to long-press to be able to see values (enter tracking mode).  However, this might only be my opinion - happy to discuss:)

shonsirsha:
> looks nice 👍
> 
> some issue: On medium/small screen around 600px I sometimes get a long horizontal scrollbar.

Thanks for the comments @eni! Made most of the changes you suggested, PTAL 🙏 . One thing, I can’t seem to repro the huge scrollbar thing. This is how it looked on mine:

Screen Shot 2022-11-07 at 03 58 31

It does sometimes have a horiz scrollbar, but only for a split second when resizing quickly from a bigger screen to smaller one. However, I don't think this would be a problem, since in practice app users aren’t doing that.

shonsirsha:
> I also noticed that rotating my screen the chart line doesn't update correctly. Again, this seems to happen also on the master branch, but maybe you can look into that while you are working on this? immagine

I took a look at this, and it seems like we've done everything correctly in terms of setting / resizing the chart width -  and it did reflect on the DOM's width. However, the image that the chart lib rendered in the canvas isn't reflecting it. To be more precise, the canvas (which is the `chart`, in the code) did get bigger to fill the space needed, but the actual graph drawing (the blue lines) never got bigger / got stretched big enough. 

This seems like a bug from the library to me...

thisconnect:
hm, indeed I can't reproduce the big scrollbar (chart displayed a lot more data).

At the moment I can only reproduce a very small horizontal scrollbar on Chrome at around 630px.

> One thing, I can’t seem to repro the huge scrollbar thing.

frontends/web/src/locales/en/app.json
```diff
@@ -352,7 +352,7 @@
   },
   "chart": {
     "dataMissing": "Gathering historical data... stay tuned.",
-    "dataUpdating": "updating data…",
+    "dataUpdating": "updating data...",
```
thisconnect, line 355: hm, I think it doesn't matter if we do locize-pull next time it will be overwritten, but if you update the PR anyway please change to `…`.

frontends/web/src/routes/account/summary/chart.module.css
thisconnect: 👍  for `var(--space-half`

It would go to the edge, but I see, although I think negative margins are nowadays *almost* never needed anymore, it will introduce more changes on the whole view. Negative margin is simplest. 

> Nevertheless, originally, the chart itself already has no padding and margin, so adding content noSpace wont change anything.

thisconnect:
maybe we should listen to orientationchange (but I just found this is deprecated and new api is [ScreenOrientation](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation)).

Anyway, doesn't have to be part of this PR. Maybe it would be better to rewrite to functional component and then add hooks.

> I also noticed that rotating my screen the chart line doesn't update correctly. Again, this seems to happen also on the master branch, but maybe you can look into that while you are working on this?

thisconnect (changes requested):
please rebase onto upstream/master

thisconnect:
Code LGTM @Beerosagos could you test the current version one more time on Android?

frontends/web/src/locales/en/app.json
shonsirsha: Oops, done.

shonsirsha:
Made the last change and rebased @thisconnect

benma:
please update CHANGELOG.md with an entry describing this change for users that look at release notes

Beerosagos:
> Code LGTM @Beerosagos could you test the current version one more time on Android?

All good on my Android! As discussed above, I still think that it is not super intuitive that you have to do a long-press to get the chart details (but I like it a lot that it shows only while you have the finger on it). If it is not possible to achieve the same result with a short press, maybe some kind of hint about that could be useful. wdyt? 

cc @jadzeidan

jadzeidan:
> > Code LGTM @Beerosagos could you test the current version one more time on Android?
> 
> All good on my Android! As discussed above, I still think that it is not super intuitive that you have to do a long-press to get the chart details (but I like it a lot that it shows only while you have the finger on it). If it is not possible to achieve the same result with a short press, maybe some kind of hint about that could be useful. wdyt?
> 
> cc @jadzeidan

Yeah, ideally it would work with a short press. But if only long press is possible I think a hint that shows up the first time you see the new chart would be nice. I'll think of a concept that could work.

shonsirsha:
> please update CHANGELOG.md with an entry describing this change for users that look at release notes

Done. Changes made and rebased ✅

thisconnect (approved):
LGTM

PR #1823, opened 2022-11-03
frontend: fix space between amount and unit

thisconnect:
In the last commit that improved support for large amounts or coins with lots of decimal, the space for the unit was fixed to 40px and the cause for some visual bug and inconsistent space between amount and unit. Sometimes too large or almost overlapping in case of GOETH.

This change creates a dynamic 'spacer' element with the correct width by passing the unit + 1 white-space to an invisible elemenet.


Beerosagos (approved):
LGTM!

PR #1827, opened 2022-11-07
backend: recreate corrupt bolt databases

benma:
A user had a corrupt rates.db database. bbolt is designed that this
does not happen using their API. Corruptions can happen due to
external causes, such as disk errors.

`bbolt check rates.db` listed errors.

Reading the btcUSD bucket eats all RAM until the process is killed.

Reading the btcEUR bucket reads a key that is not 8 bytes long, panicking on this line:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/6858d293a7561f575ee35445bcd1f6ab2676e341/backend/rates/db.go#L28

To mitigate this issue, we check for corruption after opening the database and recreate the database (deleting the old one) if it was corrupted. This is okay, as the rates.db file is merely a cache and can be re-populated at any time.

For the same reason, we do the same for the account cache databases and the notifier database.


Beerosagos (approved):
tACK

util/bolt/bbolt_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NoError(t, err)
+	require.Error(t, checkCorruption(db))
+	require.NoError(t, db.Close())
+
+	// Open with corruption recovery.
+	db, err = Open(filename, log)
+	require.NoError(t, err)
+	require.NoError(t, db.Close())
+
+	// Check that the recovery created a new DB. The hash below is the hardcoded hash of a fresh DB.
```
Beerosagos, line 72: What do you think about comparing the hash before and after the test? I am afraid that the hash of an empty DB could change in the future with bbolt updates.

benma: good idea - done (hash not necessary anymore, can compare directly)

PR #1828, opened 2022-11-07
frontend: fix progress bar styling

thisconnect:
The native progress element that shows the progress of upgrading the firmware has a visual regression and does not use the full available width anymore but is rendered at around 150px.

Looks like the default styling for the progress element changed in chrome and uses `display: inline-block;` and `width: 160px;`.

This is older but could be related:
- https://bugs.webkit.org/attachment.cgi?id=50959&action=diff#WebCore/css/html.css_sec1


benma (approved):
utACK

PR #1829, opened 2022-11-07
frontend: convert multilineMarkup to JSX

shonsirsha:
To follow the latest conventions in the frontend, we would like to convert `multilineMarkup()` into a "normal" JSX component, `<MultilineMarkup />`.

Previously, it's not possible to use the component as `<MultilineMarkup />` due to the usage of Preact 8 - which does not support document fragments.

### DIFF
--- a/frontends/web/src/utils/markup.tsx
+++ b/frontends/web/src/utils/markup.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import React, { createElement } from 'react';
+import React, { createElement, FunctionComponent } from 'react';
 
 type MarkupProps = {
     tagName: keyof JSX.IntrinsicElements;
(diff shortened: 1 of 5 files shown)


thisconnect (approved):
LGTM

frontends/web/src/utils/markup.tsx
```diff
@@ -47,23 +47,16 @@ export function SimpleMarkup({ tagName, markup, ...props }: MarkupProps) {
...
-    SimpleMarkup({
-      key: `${line}-${i}`,
-      tagName,
-      markup: line,
-      ...props
-    })
-  ));
-}
+
+export const MultilineMarkup: FunctionComponent<MarkupProps> = ({ tagName, markup, ...props }) => {
```
benma, line 54: Is FunctionComponent correct here? I think @thisconnect once told me that we should only use it if there are children (or not even then?).

See also https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13

I might be wrong and am mainly here to learn :nerd_face:

thisconnect: yeah we use FunctionComponent everywhere at the moment and probably shouldn't. Lets address this in a different PR and update the whole codebase.

thisconnect (approved):
LGTM

thisconnect (approved):
LGTM

PR #1831, opened 2022-11-08
frontend: refactor color variables

jadzeidan:
Refactoring color variables by separating variables so components (such as fonts and backgrounds) don't use the same variable. Continuation of pr: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1785


jadzeidan:
@thisconnect PTAL

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -60,10 +60,10 @@
 
 .filters button {
   appearance: none;
-  background: var(--color-white);
-  border: 2px solid var(--color-white);
+  background: var(--background-secondary);
+  border: 2px solid var(--background-secondary);
   border-radius: 2rem;
-  color: var(--color-softblack);
+  color: var(--color-default);
```
thisconnect, line 66: 👍

frontends/web/src/components/forms/checkbox.module.css
```diff
@@ -31,21 +31,21 @@
     top: -1px;
 }
```
thisconnect, line 32: nit: there is another border color here that uses `rgb(220, 220, 220)`

frontends/web/src/components/forms/checkbox.module.css
```diff
@@ -69,19 +69,19 @@
...
-    color: var(--color-white);
+    color: var(--color-alt);
 }
 
 .info input:checked + label::before {
     background-color: var(--color-info);
     border-color: var(--color-info);
-    color: var(--color-white);
+    color: var(--color-alt);
 }
```
thisconnect, line 86: and some more at the end of this file

```
checkbox input[disabled] + label::before {
    background: rgb(248, 248, 248);
    border-color: rgb(220, 220, 220);
```

Not sure keep as is or update?

frontends/web/src/components/forms/radio.module.css
```diff
@@ -50,7 +50,7 @@
 /* checked */
```
thisconnect, line 50: there is another one a few lines before this line (sorry can't commend there so I write here),

```
.radio input + label::before {
    border: 1px solid rgb(220, 220, 220);
```

frontends/web/src/components/forms/radio.module.css
```diff
@@ -50,7 +50,7 @@
...
 
 .radio input:checked + label::before {
-    background-color: var(--color-white);
+    background-color: var(--background-secondary);
 }
```
thisconnect, line 55: and at the end, I am not sure where disabled radio inputs are used.

```
.radio input:disabled + label::before {
    background: rgb(248, 248, 248);
    border-color: rgb(220, 220, 220);
```

frontends/web/src/components/forms/input.module.css
```diff
@@ -49,7 +49,7 @@
 }
 
 .input input[disabled] {
-    background-color: var(--color-white) !important;
+    background-color: var(--background-secondary) !important;
     border-color: var(--color-lightgray) !important;
```
thisconnect, line 53: change `--color-lightgray` to `--background` ?

frontends/web/src/style/variables.css
```diff
@@ -25,6 +25,8 @@
     --color-primary: var(--color-blue);
     --color-secondary: var(--color-gray);
     --color-light: var(--color-mediumgray);
```
thisconnect, line 27: in the old PR I wrote that `--color-light` isn't used anywhere, could you double check and remove?

thisconnect (changes requested):
Very nice cleanup effort, thanks! I went through the comments of the old PR, could you check this again:

    Also have a look at spinner.module.css and other modules where there is background: #fff; could you do global search for `background:` to see if there are other occurences?

    and in view.module.css it should use var(--background) and not background: var(--color-lightgray);. Related: --color-lightgray is used many times, could you check if that should be replaced with --background everyhere or in some places?

frontends/web/src/style/variables.css
```diff
@@ -37,6 +39,9 @@
     /* background colors */
     --background: var(--color-lightgray);
     --bg-transparent-dark: rgba(0, 0, 0, 0.4);
+    --background-secondary: var(--color-white);
+    --background-dark: var(--color-softblack);
+    --background-tertiary: var(--color-gray-alt);
```
thisconnect, line 45: One more thing: could you add a comment to `/* font colors */` and to `/* background colors */` that these are the variables that should be used, in order for dark mode to work correctly?

maybe also add comment on the section `/* colors */` of colors that should NOT be used directly, to hint to other devs which ones should be used.

frontends/web/src/components/forms/checkbox.module.css
jadzeidan: I kept this since we don't have a variable for this color. I think we can keep it for now (since I think this color works for both light and dark mode) and then in a future pr we can make it a variable.

frontends/web/src/components/forms/checkbox.module.css
jadzeidan: Like the comment above, I kept these hard coded values because we don't currently have those colors as variables. I think we can keep it for now because I think the check boxes don't need to be changed for dark mode. And then we can make them variables in a future pr.

frontends/web/src/components/forms/radio.module.css
jadzeidan: Same as comment above (I think the current radio button colors also work for dark mode). So we can keep it how it is now and make the color variables in a future pr.

frontends/web/src/components/forms/radio.module.css
jadzeidan: Same as comment above.

frontends/web/src/components/forms/input.module.css
jadzeidan: It's a `border-color`, so I wasn't sure whether borders should use `font colors` or `background colors`. So I just left it how it is. But I think `border-color` can use background colors if we decide that's ok.

frontends/web/src/style/variables.css
jadzeidan: We use `--color-mediumgray` a lot and for a fonts, backgrounds and borders. So we should replace `--color-mediumgray` with `--color-light` for fonts. For backgrounds, we should make a new variable that uses `--color-mediumgray`.

frontends/web/src/style/variables.css
jadzeidan: Added comments to `/* font colors */` and `*background colors*/`

jadzeidan:
@thisconnect PTAL

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1852

PR #1832, opened 2022-11-08
frontend: new guide entry for entering tracking mode on mobile

shonsirsha:
New entry guide to help users understand the action needed to be able to
explore (entering the tracking mode of) the portfolio graph.

This allows them to see the historical value of their
portfolio.

Screen Shot 2022-11-09 at 11 18 27

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -827,6 +827,10 @@
       "close": "Close guide",
       "open": "Guide"
     },
+    "trackingModePortfolioChart": {
+      "text": "On desktop, hover the cursor over the chart. On mobile, hold your finger on the chart and drag horizontally.",
+      "title": "How to see historical values on the chart?"
+    },
     "unlock": {
       "forgotDevicePassword": {
         "text": "You have to reset the device and restore the wallet from a backup, using the recovery password.",
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -381,6 +381,7 @@ class AccountsSummary extends Component<Props, State> {
             text: t('guide.accountSummaryAmount.text'),
             title: t('guide.accountSummaryAmount.title')
           }} />
+          <Entry key="trackingModePortfolioChart" entry={t('guide.trackingModePortfolioChart')} />
         </Guide>
       </div>
     );


frontends/web/src/locales/en/app.json
```diff
@@ -827,6 +827,10 @@
       "close": "Close guide",
       "open": "Guide"
     },
+    "trackingModePortfolioChartMobile": {
+      "text": "Hold your finger on the chart and drag to see the value of your portfolio at different times in history.",
+      "title": "How to see historical values on the chart on mobile?"
```
thisconnect, line 832: Sorry I proposed with "on mobile" as an intermediate solution, but as soon as it is in locize others might translate this temp. text. So I think it would be better if the entry just describes both desktop and mobile:

title: How to see historical values on the chart?
text: On desktop hover the cursor or on mobile touch and drag on the chart to see the value of your portfolio at different times in history.

thisconnect (changes requested):

frontends/web/src/locales/en/app.json
jadzeidan: A simpler version which I think is more clear.

On desktop, hover the cursor over the chart. On mobile, hold your finger on the chart and drag horizontally.

shonsirsha: Just made the change. PTAL 🙏 

Screen Shot 2022-11-09 at 03 31 06

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -381,6 +381,7 @@ class AccountsSummary extends Component<Props, State> {
             text: t('guide.accountSummaryAmount.text'),
             title: t('guide.accountSummaryAmount.title')
           }} />
+          <Entry key="trackingModePortfolioChartMobile" entry={t('guide.trackingModePortfolioChartMobile')} />
```
thisconnect, line 384: Nit: now that the guide entry is for both mobile and desktop remove "Mobile" from the key and rename to `trackingModePortfolioChart` or similar?

thisconnect (approved):
LGTM with nit

thisconnect:
Nit: please also update the commit message

frontends/web/src/routes/account/summary/accountssummary.tsx
shonsirsha: Oh definitely makes sense, forgot to remove it actually. Thanks!

PR #1835, opened 2022-11-09
CHANGELOG: move entry out of 4.35 to unreleased

benma:
This change added a 0.5s-1s delay when launching the app, which should be addressed before releasing it. Moving out of the 4.35 release.


Beerosagos (approved):
LGTM

PR #1836, opened 2022-11-10
Revert "backend: recreate corrupt bolt databases"

benma:
This reverts commit 4ee3dc9d09002a4af45cb512e97cc085bbc2e5d9.

And remove the CHANGELOG entry again.

The corruption check can take a long time on a big database (i.e. ~1s for a 50MB rates.db), which casues a noticable delay when loading the database when the app launches. Need to come up with a better solution.


Beerosagos (approved):
tACK

PR #1841, opened 2022-11-15
backend/eth: return ErrInsufficientFunds if gas estimation failed

benma:
Similar to 7a25ac8720989b2215806f802e42d5bf1ccc263a.

EtherScan returns this error during when estimating the gas if there are not enough funds to pay for the transaction.

This is likely a change in behavior in Etherscan, as the check below `// Check that the entered value and the estimated fee are not greater than the balance.` used to work. We keep that code in case we move away from Etherscan or Etherscan changes the behavior and returns a gas  anyway.

This is not a robust check as it compares strings, but it is the best we can do for now.


Beerosagos (approved):
tACK!

PR #1842, opened 2022-11-15
frontend: improve dialog on mobile

shonsirsha:
To maximise spacing and achieving a more modern look on mobile, we want to re-style the Dialog component to become a "modal bottom sheet" (or a bottom sheet).

Some screenshots and a screen recording:

https://user-images.githubusercontent.com/28676406/201877070-d82b3737-ed6d-4649-815c-bfd7c81afce1.mov

Screen Shot 2022-11-15 at 07 25 38
Screen Shot 2022-11-15 at 07 25 59
Screen Shot 2022-11-15 at 07 26 10

### DIFF
--- a/frontends/web/src/components/dialog/dialog.module.css
+++ b/frontends/web/src/components/dialog/dialog.module.css
@@ -8,13 +8,12 @@
     left: 0;
     height: 100%;
     width: 100%;
-    background-color: var(--bg-transparent-dark);
+    background: transparent;
     z-index: 4010;
-    opacity: 0;
 }
 
 .overlay.activeOverlay {
-    opacity: 1;
+    background-color: var(--bg-transparent-dark);
 }
 
 .modal {
@@ -26,14 +25,7 @@
     text-align: left;
     max-height: 100vh;
     overflow: auto;
-    opacity: 0;
 }
-
-.modal.activeModal {
-    opacity: 1;
-}
-
-.active {}
 
 .modal.small {
     max-width: 340px;
@@ -236,9 +228,34 @@
     width: 100%;
 }
 
+@media (min-width: 769px) {
+    .closingOverlay {
+        display: none;
+    }
+}
+
 @media (max-width: 768px) {
+
+    .header .title{
+        font-size: var(--header-default-font-size);
+    }
+
     .modal {
         padding: 0;
+    }
+
+    .modal, .modal.small, .modal.medium, .modal.large{
+        max-width: 100vw;
+        height: 100vh;
+        transform: translateY(100vh);
+        transition: transform 300ms;
+        border-top-right-radius: var(--space-half);
+        border-top-left-radius: var(--space-half);
+    }
+
+    .modal.open {
+      transition: transform 300ms;
+      transform: translateY(0);
     }
 
     .contentContainer.padded {
--- a/frontends/web/src/components/dialog/dialog.tsx
+++ b/frontends/web/src/components/dialog/dialog.tsx
@@ -29,11 +29,13 @@ interface Props {
     onClose?: (e?: Event) => void;
     disabledClose?: boolean;
     children: React.ReactNode;
+    open: boolean;
 }
 
 interface State {
     active: boolean;
     currentTab: number;
+    renderDialog: boolean;
 }
 
 class Dialog extends Component<Props, State> {
@@ -41,14 +43,30 @@ class Dialog extends Component<Props, State> {
   private modal = createRef<HTMLDivElement>();
   private modalContent = createRef<HTMLDivElement>();
   private focusableChildren!: NodeListOf<HTMLElement>;
+  private timerId: any;
 
   public state: State = {
     active: false,
     currentTab: 0,
+    renderDialog: false,
   };
 
-  public componentDidMount() {
-    setTimeout(this.activate, 10);
+  public componentDidUpdate(prevProps: Readonly<Props>): void {
+    const { open } = this.props;
+    const { active } = this.state;
+
+    if (!active && open && !prevProps.open) {
+      this.setState({ renderDialog: true }, () => {
+        this.activate();
+      });
+      return;
+    }
+
+
+    if (active && !open && prevProps.open) {
+      this.deactivate();
+      return;
+    }
   }
 
   public componentWillUnmount() {
@@ -88,6 +106,19 @@ class Dialog extends Component<Props, State> {
     });
   };
 
+  private deactivateModal = () => {
+    if (!this.modal.current || !this.overlay.current) {
+      return;
+    }
+    this.overlay.current.classList.remove(style.closingOverlay);
+    this.setState({ active: false, currentTab: 0, renderDialog: false }, () => {
+      document.removeEventListener('keydown', this.handleKeyDown);
+      if (this.props.onClose) {
+        this.props.onClose();
+      }
+    });
+  };
+
   private getNextIndex(isNext: boolean) {
     const { currentTab } = this.state;
     const focusables = Array.from(this.focusableChildren);
@@ -119,23 +150,36 @@ class Dialog extends Component<Props, State> {
     if (!this.modal.current || !this.overlay.current) {
       return;
     }
-    this.modal.current.classList.remove(style.activeModal);
+
+    if (this.timerId) {
+      clearTimeout(this.timerId);
+    }
+
+    this.modal.current?.classList.remove(style.open);
     this.overlay.current.classList.remove(style.activeOverlay);
-    this.setState({ active: false, currentTab: 0 }, () => {
-      document.removeEventListener('keydown', this.handleKeyDown);
-      if (this.props.onClose) {
-        this.props.onClose();
-      }
-    });
+    this.overlay.current.classList.add(style.closingOverlay);
+    this.timerId = setTimeout(this.deactivateModal, 300);
   };
 
   private activate = () => {
+
     this.setState({ active: true }, () => {
       if (!this.modal.current || !this.overlay.current) {
         return;
       }
+
+      if (this.timerId) {
+        clearTimeout(this.timerId);
+      }
+
       this.overlay.current.classList.add(style.activeOverlay);
-      this.modal.current.classList.add(style.activeModal);
+
+      // Minor delay
+      this.timerId = setTimeout(() => {
+        this.modal.current?.classList.add(style.open);
+      }, 10);
+
+
       this.focusWithin();
       this.focusFirst();
     });
@@ -153,11 +197,17 @@ class Dialog extends Component<Props, State> {
       disabledClose,
       children,
     } = this.props;
+    const { renderDialog } = this.state;
     const isSmall = small ? style.small : '';
     const isMedium = medium ? style.medium : '';
     const isLarge = large ? style.large : '';
     const isSlim = slim ? style.slim : '';
     const isCentered = centered && !onClose ? style.centered : '';
+
+    if (!renderDialog) {
+      return null;
+    }
+
     return (
       <div className={style.overlay} ref={this.overlay}>
         <div
--- a/frontends/web/src/components/language/language.tsx
+++ b/frontends/web/src/components/language/language.tsx
@@ -19,8 +19,8 @@ import { useState } from 'react';
 import { i18n as Ii18n } from 'i18next';
 import { useTranslation } from 'react-i18next';
 import { Dialog } from '../dialog/dialog';
-import style from './language.module.css';
 import { TActiveLanguageCodes, TLanguagesList } from './types';
+import style from './language.module.css';
 
 type TLanguageSwitchProps = {
     languages?: TLanguagesList;
@@ -113,44 +113,40 @@ const LanguageSwitch = ({ languages }: TLanguageSwitchProps) => {
         </svg>
         {allLanguages[selectedIndex].code === 'en' ? 'Other languages' : 'English'}
       </button>
-      {
-        activeDialog && (
-          <Dialog small slim title={t('language.title')} onClose={() => setActiveDialog(false)}>
-            {
-              allLanguages.map((language, i) => {
-                const selected = selectedIndex === i;
-                return (
-                  <button
-                    key={language.code}
-                    className={[style.language, selected ? style.selected : ''].join(' ')}
-                    onClick={() => changeLanguage(language.code, i)}
-                    data-testid={`language-selection-${language.code}`}
-                  >
-                    {language.display}
-                    {
-                      selected && (
-                        <svg
-                          className={style.checked}
-                          xmlns="http://www.w3.org/2000/svg"
-                          width="24"
-                          height="24"
-                          viewBox="0 0 24 24"
-                          fill="none"
-                          stroke="currentColor"
-                          strokeWidth="2"
-                          strokeLinecap="round"
-                          strokeLinejoin="round">
-                          <polyline points="20 6 9 17 4 12" />
-                        </svg>
-                      )
-                    }
-                  </button>
-                );
-              })
-            }
-          </Dialog>
-        )
-      }
+      <Dialog small slim title={t('language.title')} onClose={() => setActiveDialog(false)} open={activeDialog}>
+        {
+          allLanguages.map((language, i) => {
+            const selected = selectedIndex === i;
+            return (
+              <button
+                key={language.code}
+                className={[style.language, selected ? style.selected : ''].join(' ')}
+                onClick={() => changeLanguage(language.code, i)}
+                data-testid={`language-selection-${language.code}`}
+              >
+                {language.display}
+                {
+                  selected && (
+                    <svg
+                      className={style.checked}
+                      xmlns="http://www.w3.org/2000/svg"
+                      width="24"
+                      height="24"
+                      viewBox="0 0 24 24"
+                      fill="none"
+                      stroke="currentColor"
+                      strokeWidth="2"
+                      strokeLinecap="round"
+                      strokeLinejoin="round">
+                      <polyline points="20 6 9 17 4 12" />
+                    </svg>
+                  )
+                }
+              </button>
+            );
+          })
+        }
+      </Dialog>
     </div>
   );
 };
--- a/frontends/web/src/hooks/matchmedia.ts
+++ b/frontends/web/src/hooks/matchmedia.ts
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2022 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect, useState } from 'react';
+
+const getMatches = (query: string): boolean => window.matchMedia(query).matches;
+
+export const useMatchMedia = (query: string): boolean => {
+  const [matches, setMatches] = useState<boolean>(getMatches(query));
+
+  useEffect(() => {
+    const handleChange = () => {
+      setMatches(getMatches(query));
+    };
+    const matchMedia = window.matchMedia(query);
+    handleChange();
+    matchMedia.addEventListener('change', handleChange);
+    return () => {
+      matchMedia.removeEventListener('change', handleChange);
+    };
+  }, [query]);
+
+  return matches;
+};
--- a/frontends/web/src/routes/account/receive/receive-bb01.tsx
+++ b/frontends/web/src/routes/account/receive/receive-bb01.tsx
@@ -201,16 +201,16 @@ export const Receive = ({
                       {t('receive.changeScriptType')}
                     </button>
                   )}
-                  { hasManyScriptTypes && addressDialog && (
-                    <form onSubmit={e => {
-                      e.preventDefault();
-                      setActiveIndex(0);
-                      setAddressType(addressDialog.addressType);
-                      setAddressDialog(undefined);
-                    }}>
-                      <Dialog medium title={t('receive.changeScriptType')} >
-                        {availableScriptTypes.current && availableScriptTypes.current.map((scriptType, i) => (
-                          <div key={scriptType}>
+                  <form onSubmit={e => {
+                    e.preventDefault();
+                    setActiveIndex(0);
+                    setAddressType(addressDialog ? addressDialog?.addressType : addressType);
+                    setAddressDialog(undefined);
+                  }}>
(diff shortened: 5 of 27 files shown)


frontends/web/src/components/language/language.tsx
```diff
@@ -80,10 +80,15 @@ const LanguageSwitch: FunctionComponent<TLanguageSwitchProps> = ({ languages })
...
+  const dialogRef = useRef<Dialog>() as React.MutableRefObject<Dialog>;
+
   const changeLanguage = (langCode: TActiveLanguageCodes, index: number) => {
     setSelectedIndex(index);
-    setActiveDialog(false);
     i18n.changeLanguage(langCode);
+    dialogRef.current.deactivate();
+    setTimeout(() => {
+      setActiveDialog(false);
+    }, 350);
```
shonsirsha, line 91: This is needed to preserve the animation of dismissing the modal after clicking a language:

1. We need to trigger `deactivate()` which shows the animation & deactivates the dialog. 

With the way `Dialog` works right now, I haven't found a "better" way to trigger the animation from a parent component, such as `language.tsx` here.

2. And after 350ms (after the animation has finished), we can set `activeDialog` to `false` 

We need to give a 350ms delay here before setting `activeDialog` to `false`, because otherwise the component won't have a chance to animate. It'd just be dismissed instantly without any animation due to conditional rendering `activeDialog && <Dialog...`, and this isn't what we want.

I can see this being refactored, but we first need to start from the `Dialog` component, probably utilising `React.createContext` so we don't have to conditional render `<Dialog/>` on each page where it's being used thus reducing redundancy.

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -96,6 +97,18 @@ export const Receive: FunctionComponent<Props> = ({
...
+      e.preventDefault();
+      setActiveIndex(0);
+      setAddressType(addressDialog.addressType);
+      dialogRef.current.deactivate();
+      setTimeout(() => {
+        setAddressDialog(undefined);
+      }, 350);
+    }
+  };
+
```
shonsirsha, line 111: `setTimeout` and `deactivate()` needed for the same reasons [above](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1842/files#r1022541054).

frontends/web/src/components/language/language.tsx
shonsirsha, line 91: This is needed to preserve the animation of dismissing the modal after clicking a language:

1. We need to trigger `deactivate()` which shows the animation & deactivates the dialog. 

With the way `Dialog` works right now, I haven't found a "better" way to trigger the animation from a parent component, such as `language.tsx` here.

2. And after 350ms (after the animation has finished), we can set `activeDialog` to `false` 

We need to give a 350ms delay here before setting `activeDialog` to `false`, because otherwise the component won't have a chance to animate. It'd just be dismissed instantly without any animation due to conditional rendering `activeDialog && <Dialog...`, and this isn't what we want.

I can see this being refactored, but we first need to start from the `Dialog` component, probably utilising `React.createContext` so we don't have to conditional render `<Dialog/>` on each page where it's being used thus reducing redundancy of these pieces of code.

I am definitely open for suggestions! 🙂

shonsirsha:
Changed PR status to draft. Noticed some small stuff, will put it back to "ready for review" soon 🙏

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -45,16 +46,23 @@ class Dialog extends Component<Props, State> {
...
   }
 
   public componentWillUnmount() {
     document.removeEventListener('keydown', this.handleKeyDown);
   }
 
+  private checkIfMobile = () => {
+    this.setState({ isMobile: window.innerWidth <= 768 });
+  };
+
```
shonsirsha, line 65: We could use the newly created (also in this PR) `useMatchMedia` here once this component gets converted into a functional one.

shonsirsha:
Put back to "ready for review" 🙏 thx:)

frontends/web/src/components/language/language.tsx
```diff
@@ -79,11 +80,17 @@ const LanguageSwitch = ({ languages }: TLanguageSwitchProps) => {
...
   const [activeDialog, setActiveDialog] = useState<boolean>(false);
+  const isMobile = useMatchMedia('(max-width: 768px)');
+
+  const dialogRef = useRef<Dialog>() as React.MutableRefObject<Dialog>;
 
   const changeLanguage = (langCode: TActiveLanguageCodes, index: number) => {
     setSelectedIndex(index);
-    setActiveDialog(false);
     i18n.changeLanguage(langCode);
+    dialogRef.current.deactivate();
```
thisconnect, line 90: instead of using ref and accessing the deactivate function, could this be passed as prop somehow?

Would something like this could work instead of using ref `<Dialog open={activeDialog}` without the conditional rendering `activeDialog && (` so the component can check props.open and call deactivate, wdyt?

frontends/web/src/hooks/matchmedia.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect, useState } from 'react';
+
+export const useMatchMedia = (query: string): boolean => {
```
thisconnect, line 19: nice 🤩

frontends/web/src/hooks/matchmedia.ts
```diff
@@ -0,0 +1,44 @@
...
+  const getMatches = (query: string): boolean => {
+    if (typeof window !== 'undefined') {
+      return window.matchMedia(query).matches;
+    }
+    return false;
+  };
```
thisconnect, line 25: Can getMatches be moved out of the hook before useMatchMedia?
and `window` should always be defined, or not?

```diff
diff --git a/frontends/web/src/hooks/matchmedia.ts b/frontends/web/src/hooks/matchmedia.ts
index 9aca01c9c..07a38d03e 100644
--- a/frontends/web/src/hooks/matchmedia.ts
+++ b/frontends/web/src/hooks/matchmedia.ts
@@ -16,14 +16,9 @@
 
 import { useEffect, useState } from 'react';
 
-export const useMatchMedia = (query: string): boolean => {
-  const getMatches = (query: string): boolean => {
-    if (typeof window !== 'undefined') {
-      return window.matchMedia(query).matches;
-    }
-    return false;
-  };
+const getMatches = (query: string): boolean => matchMedia(query).matches;
 
+export const useMatchMedia = (query: string): boolean => {
   const [matches, setMatches] = useState<boolean>(getMatches(query));
```

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -96,6 +100,18 @@ export const Receive = ({
...
 
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    if (addressDialog) {
+      e.preventDefault();
+      setActiveIndex(0);
+      setAddressType(addressDialog.addressType);
+      dialogRef.current.deactivate();
+      setTimeout(() => {
+        setAddressDialog(undefined);
+      }, isMobile ? 350 : 0);
```
thisconnect, line 111: would be nice if the `<Dialog>` component would take care of the timeout.

thisconnect (changes requested):
ui looks nice.

made code review but haven't tested yet.

frontends/web/src/hooks/matchmedia.ts
shonsirsha: Yes it can, and... yeah it **should** always be :) I however sometimes get window `undefined` but it's usually for web projects that requires SSR. Shouldn't be the case here..? Anyhow I'll remove it, thx!

frontends/web/src/hooks/matchmedia.ts
```diff
@@ -0,0 +1,44 @@
...
+    matchMedia.addEventListener('change', handleChange);
+    return () => {
+      matchMedia.removeEventListener('change', handleChange);
+    };
+  }, [query]);
+
+  return matches;
+};
+
+export default useMatchMedia;
```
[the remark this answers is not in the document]
shonsirsha: Oops, good catch, exporting twice here lol. thx!

frontends/web/src/components/dialog/dialog.module.css
```diff
@@ -237,10 +235,28 @@
...
+
     .modal {
         padding: 0;
     }
 
+    .modal.small, .modal.medium, .modal.large{
+        max-width: 100vw;
+        height: 100vh;
+        margin-top: 250vh;
+        transition: 350ms all;
```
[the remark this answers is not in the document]
shonsirsha: Yeah, I agree. I'll use transform and explicitly define transform as the tranisiton property. Both behave the same but looks cleaner this way.

frontends/web/src/components/dialog/dialog.module.css
```diff
@@ -237,10 +235,28 @@
...
+    }
+
     .modal {
         padding: 0;
     }
 
+    .modal.small, .modal.medium, .modal.large{
+        max-width: 100vw;
+        height: 100vh;
+        margin-top: 250vh;
```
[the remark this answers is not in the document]
shonsirsha: See [here](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1842/files#r1023396566)

frontends/web/src/components/language/language.tsx
shonsirsha: Yeah.. this should be possible. I just didn't want to change the usage of all <Dialog/> but it's probably cleaner this way.

I think in fn components, this could very easily be solved using useEffect. We can use `componentDidUpdate` here. Let's see what I can do :)

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -115,27 +136,28 @@ class Dialog extends Component<Props, State> {
...
       if (!this.modal.current || !this.overlay.current) {
         return;
       }
-      this.overlay.current.classList.add(style.activeOverlay);
+      const { isMobile } = this.state;
       this.modal.current.classList.add(style.activeModal);
+      this.modal.current.classList.add(style.open);
+      this.overlay.current?.classList.add(style.activeOverlay);
+      setTimeout(() => {
+      }, isMobile ? 80 : 0); // Minor delay
```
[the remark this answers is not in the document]
shonsirsha: Oops, I think I was playing around with a minor delay to when adding a class but didn't actually end up using it. Will remove, thx!

frontends/web/src/components/language/language.tsx
shonsirsha: Yeah.. this should be possible. I just didn't want to change the usage of all `<Dialog/>` ,as there's quite a lot of files that are already using Dialog. buuut it's probably cleaner not to do what I did here and alter Dialog a bit, even though that means there's gonna be a lot of small changes in many files.

I think in fn components, this could very easily be solved using useEffect. We can use `componentDidUpdate` here. Let's see what I can do :)

frontends/web/src/routes/account/receive/receive.tsx
shonsirsha: Yes, we won't need this timeout on each component (and firing deactivate) once we do [this](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1842#discussion_r1023451442). :)

shonsirsha:
Thanks for the comments @thisconnect! Changes are addressed, PTAL 🙏 . 2 things to note:

1. There are more files changed due to the way we use `<Dialog/>` now. See [here](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1842#discussion_r1023210820).
2. **Because of number 1, we now don't need `useMatchMedia` anymore. 😅  Shall I remove it?**  

Tbf, we could've used `useMatchMedia` on `<Dialog/>` (instead of what we have rn to get `isMobile` there) if it were a func component , but `<Dialog/>` is a class component now. Converting it to functional would change the code even more and this PR would've been a bit too large IMO.. so I avoided that. 

Preview after code changes have been made (looks the same to me 👍 ): 


https://user-images.githubusercontent.com/28676406/202074961-94e93fe8-fd94-499b-9ba1-8e3b3766e7b4.mov

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -115,27 +136,28 @@ class Dialog extends Component<Props, State> {
...
+    this.modal.current?.classList.remove(style.open);
     this.overlay.current.classList.remove(style.activeOverlay);
-    this.setState({ active: false, currentTab: 0 }, () => {
-      document.removeEventListener('keydown', this.handleKeyDown);
-      if (this.props.onClose) {
-        this.props.onClose();
-      }
-    });
+
+    setTimeout(this.deactivateModal, isMobile ? 350 : 0);
```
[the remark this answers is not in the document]
shonsirsha: Hm, the reason we need isMobile here is so that it only delays on mobile, and not for both mobile and desktop. I mean 350ms would probably be somewhat unnoticable delay for most, but it's still a delay that's not needed on desktop (due to no animation) 🙂

frontends/web/src/hooks/matchmedia.ts
thisconnect: no SSR due to no server :)

thisconnect:
What happens if during opening/closing animation the same dialog should render with a different message, i.e. an error popped up. I assume it should clearTimeout to abort the animation of the old dialog and let the new render. Could you test this somehow?

shonsirsha:
> What happens if during opening/closing animation the same dialog should render with a different message, i.e. an error popped up...

Hey! Yeah I tested this before and it worked as expected. **Although when an "error" popped out, a different dialog (not only different message, but the actual "old" dialog styling) would've popped out.** 

This is because we are still using `alertUser` for errors, as the styling makes more sense than this new one (modal slides up from bottom to top). I believe we discussed this already though. 🙂 

[Preview] When an error occurs almost immediately upon opening the newly styled dialog:

https://user-images.githubusercontent.com/28676406/203033978-1ef3e4eb-3abf-45b2-9868-f79d05595838.mov

[Preview] when an error occurs several seconds after opening the newly styled dialog:

https://user-images.githubusercontent.com/28676406/203034012-df54abdd-c644-40cd-a565-a99ee9894d87.mov

[Preview] when for some reason another modal bottom sheet (this newly styled dialog) pops up after the currently opened one:

https://user-images.githubusercontent.com/28676406/203034205-6c78d5b8-264f-4b48-b919-cb98ac5d9e5c.mov

From the behaviour above, it seems like it's behaving as it's supposed to. 🤔 lmk if im missing smt!

thisconnect:
awesome, thanks for the videos.. yes I agree last video is kind of intended... unless in the future we mark errors/warning that they don't need to animated..

keep in mind alertUser is an unexpected error, but each view should catch expected errors and might use the existing dialog.

could you make 1 more test, like the last video, but with a dialog appearing while the dialog is animating out. (if possible)

frontends/web/src/components/dialog/dialog.module.css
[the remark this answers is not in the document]
shonsirsha: It seems like using `transform` for transition property didn't work as expected. Will keep using `all` for now.

frontends/web/src/components/dialog/dialog.module.css
```diff
@@ -236,11 +228,40 @@
...
 }
 
+@media (min-width: 769px) {
+    .closingModal {
+        background: red;
+    }
+
+    .closingOverlay {
+        display: none;
+    }
```
shonsirsha, line 238: Without this additional class, a user on desktop will be unable to scroll the app immediately for 240ms after they closed the modal. (they need to wait for 240ms upon closing, which isn't ideal).

frontends/web/src/routes/account/receive/receive-bb01.tsx
```diff
@@ -201,16 +201,16 @@ export const Receive = ({
...
-                    }}>
-                      <Dialog medium title={t('receive.changeScriptType')} >
-                        {availableScriptTypes.current && availableScriptTypes.current.map((scriptType, i) => (
-                          <div key={scriptType}>
+                  <form onSubmit={e => {
+                    e.preventDefault();
+                    setActiveIndex(0);
+                    setAddressType(addressDialog ? addressDialog?.addressType : addressType);
+                    setAddressDialog(undefined);
+                  }}>
```
thisconnect, line 209: note to myself: test bb01 receive. There is small change that the form is always rendered now, but shouldn't be a problem.

frontends/web/src/routes/account/receive/receive-bb01.tsx
```diff
@@ -249,12 +249,13 @@ export const Receive = ({
...
                     <div className={style.hide}></div>
                   )}
-                  { account && forceVerification && verifying && (
-                    <Dialog
-                      title={verifyLabel}
-                      disableEscape={true}
-                      medium centered>
-                      <div className="text-center">
+                  <Dialog
+                    open={(!!(account && forceVerification && verifying))}
```
thisconnect, line 253: Nit: as those are all booleans anyways it's fine to not use `!!`

frontends/web/src/routes/device/bitbox01/settings/components/randomnumber.jsx
```diff
@@ -55,27 +55,22 @@ class RandomNumber extends Component {
         <SettingsButton onClick={this.getRandomNumber}>
           {t('random.button')}
         </SettingsButton>
-        {
-          // @ts-ignore Object is possibly 'undefined'.
```
thisconnect, line 59: looks like TS doesn't complain about this anymore, also the next `// @ts-ignore` seems to be not needed anymore.

frontends/web/src/routes/device/bitbox02/checkbackup.tsx
```diff
@@ -81,38 +81,34 @@ class Check extends Component<Props, State> {
...
-                    <Button
-                      primary
-                      onClick={this.abort}
-                      disabled={!userVerified}
-                    >
-                      { userVerified ? t('button.ok') : t('accountInfo.verify') }
-                    </Button>
-                  </DialogButtons>
+        <Dialog open={activeDialog} title={message}>
+          {
```
thisconnect, line 85: since you are here, could you remove this `{`, seems to be some leftover that is not needed.

frontends/web/src/routes/device/bitbox02/sdcardcheck.tsx
```diff
@@ -43,34 +43,30 @@ const SDCardCheck = ({ deviceID, children }: TProps) => {
...
+              onClick={check}>
+              {t('button.ok')}
+            </Button>
+            <ButtonLink
+              transparent
+              to={`/device/${deviceID}`}>
+              {t('button.back')}
+            </ButtonLink>
+          </DialogButtons>
+        </Dialog> : children}
```
thisconnect, line 69: note to my self: check sdcard check without sdcard, as the dom is slightly different now and has additional `<div>` around the dialog. Should not matter.

thisconnect (changes requested):
very nice 👍  added some small nits, but LGTM.

@Beerosagos  @jadzeidan could you test on real devices ideally with Qt / Android app?

Test for example: configure Tor proxy on small screen and on normal screen.

frontends/web/src/components/dialog/dialog.module.css
```diff
@@ -236,11 +228,40 @@
     width: 100%;
 }
 
+@media (min-width: 769px) {
+    .closingModal {
+        background: red;
```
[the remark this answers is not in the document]
shonsirsha: Oops! Was just testing. Thx!

frontends/web/src/components/language/language.tsx
```diff
@@ -81,10 +81,12 @@ const LanguageSwitch = ({ languages }: TLanguageSwitchProps) => {
   const [selectedIndex, setSelectedIndex] = useState<number>(getSelectedIndex(allLanguages, i18n));
   const [activeDialog, setActiveDialog] = useState<boolean>(false);
 
+  // const dialogRef = useRef<Dialog>() as React.MutableRefObject<Dialog>;
+
   const changeLanguage = (langCode: TActiveLanguageCodes, index: number) => {
     setSelectedIndex(index);
-    setActiveDialog(false);
     i18n.changeLanguage(langCode);
+    setActiveDialog(false);
```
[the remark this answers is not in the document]
shonsirsha: Nope, will do!

frontends/web/src/routes/device/bitbox02/checkbackup.tsx
shonsirsha: Yeah, good catch!

frontends/web/src/routes/account/receive/receive-bb01.tsx
shonsirsha: `account` isn't actually boolean, it's `accountApi.IAccount | undefined`..

We could do `(!!account) && forceVerification && verifying` though.

frontends/web/src/routes/account/receive/receive-bb01.tsx
```diff
@@ -201,16 +201,16 @@ export const Receive = ({
...
+                  <form onSubmit={e => {
+                    e.preventDefault();
+                    setActiveIndex(0);
+                    setAddressType(addressDialog ? addressDialog?.addressType : addressType);
+                    setAddressDialog(undefined);
+                  }}>
+                    <Dialog open={!!(hasManyScriptTypes && addressDialog)} medium title={t('receive.changeScriptType')} >
+                      {availableScriptTypes.current && availableScriptTypes.current.map((scriptType, i) => (
+                        <div key={scriptType}>
+                          {addressDialog && <>
```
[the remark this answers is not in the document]
shonsirsha: >>  The dialog already only renders if addressDialog is true

Oh okay. 👍 

Anyhow, unfortunately TS is still going to be unhappy with such optional chaining... 🤔 

So should we use `open={!!(addressDialog)}` for this one now?

shonsirsha:
Thanks @thisconnect! Made some more changes, PTAL 🙏

frontends/web/src/routes/account/receive/receive-bb01.tsx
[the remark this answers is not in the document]
thisconnect: it's fine now, actually BB01 doesn't have unified accounts, hasManyScriptTypes and this dialog to choose between different scriptTypes in BB01 receive can probably be removed. But later in a different PR.

thisconnect:
thanks @shonsirsha 👍 

a229726d8cd5424d502aa9610bad5962088ce7bb looks good to me, I'll ping people to test a bit and then merge it by the end of the week.

Beerosagos:
> @Beerosagos @jadzeidan could you test on real devices ideally with Qt / Android app?

Tested on Linux QT and Android 11. Looks pretty nice!! :)

thisconnect (approved):
LGTM

shonsirsha:
> LGTM

Feel free to let me know if it's good to be merged or more testing still needs to be done. Thanks! 🙂

PR #1845, opened 2022-11-15
git: ignore unfinished locales

thisconnect:
Was removed in b21fd73fa1b1fa87f684b76c1ce1a00a93bb0634 because in VS Code the content of the files could not be searched, probably because one rule ignores the locales directory, and other rules unignore the currently active languages that are supported in the BitBoxApp.

Searching in VS Code is still a bit buggy and content cannot be found unless one of the parent directories is specifically added in 'files to include'. This is good enough.

Added back the ignores that unfinished locales are not accidentally commited.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,25 @@ Session.vim
 tags
 # Persistent undo
 [._]*.un~
+
+# Ignore unfinished locales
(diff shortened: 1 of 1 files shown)


thisconnect:
@benma PTAL

PR #1839, opened 2022-11-14
frontend: changed return types & children types for FC

shonsirsha:
As a part of our refactoring effort in the FE to increase our code's readability and consistency as well as following the best practices and conventions, we'd like to change several things mentioned below for our functional components (FC):

1. If a component optionally accepts children we keep using `FunctionComponent` as we do already (or start to use one when it isn't used yet).

2. If a component does not accept any children we remove `FunctionComponent` and `PropsWithChildren`.

3. Remove return type `JSX.Element` or `JSX.Element | null`.

4. Children's type is mostly `ReactNode`. When component must only have exactly 1 child, the type is `ReactChild`.

Also changed Interface to Type whenever possible.

PR #1846, opened 2022-11-16
buy: provide a first integration of Pocket widget

Beerosagos:
This commit is a first step in the integration process of Pocket widget into the buy workflow.

The widget is embedded in a dedicated iframe in `pocket.tsx`, while a drafted `exchange.tsx` allows the user to choose which exchange (moonpay or pocket) should be used in the purchase process. The iframe integration includes an address verification step similar to the one used with AOPP flow. In order to do this, a typescript `request-address` library has been added to the `package.json`.

Backend side, a new `pocket.go` contains the functions needed to verify coin support and iframe URL, while the code dedicated to the address verification is in `backend.go`.


backend/exchanges/pocket.go
```diff
@@ -0,0 +1,34 @@
...
+	pocketAPILiveURL = "http://widget.pocketbitcoin.com/widget_vqx25E6kzvGBYGjN2QoXVH"
+)
+
+// BuyPocket verifies if the passed account is enabled for Pocket and returns the
+// url needed to incorporate the widget in the frontend.
+func BuyPocket(acct accounts.Interface) (string, error) {
+	if acct.Coin().Code() != coin.CodeBTC {
+		return "", fmt.Errorf("unsupported cryptocurrency code %q", acct.Coin().Code())
+	}
+
```
benma, line 24: Use `IsPocketSupported()`?

is this check even needed? in the frontend this call doesn't happen unless you selected pocket (where the check already happened).

backend/backend.go
```diff
@@ -683,3 +685,62 @@ func (backend *Backend) HandleURI(uri string) {
...
+		return "", ""
+	}
+	if err := account.Initialize(); err != nil {
+		log.
+			WithError(err).
+			WithField("code", account.Config().Code).
+			Error("Pocket Widget: could not initialize account")
+		return "", ""
+	}
+	if account.Coin().Code() != coinpkg.CodeBTC {
```
benma, line 710: use `IsPocketSupported()`?

backend/exchanges/pocket.go
```diff
@@ -0,0 +1,34 @@
...
+	// Address signing is only available for mainnet accounts on firmware, we can't use test URL at the moment.
+	apiURL := pocketAPILiveURL
+	return apiURL, nil
+}
+
+// IsPocketSupported is true if coin.Code is supported by Pocket.
+func IsPocketSupported(acct accounts.Interface) bool {
+	// Pocket would also support tbtc, but at the moment testnet address signing is disabled on firmware.
```
benma, line 32: the actual check should then be (in pseudocode ;)):

(coinCode==BTC or coinCode==TBTC) and acct.keystore.CanSignMessage(coinCode)

then it will start working automatically if we enable tbtc msg signing in the bitbox02, or if we add signmsg to the software test keystore - the latter could actually be a good time now to test pocket integration in testnet.

frontends/web/package.json
```diff
@@ -20,7 +20,8 @@
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "react-i18next": "11.16.8",
-    "react-router-dom": "6.3.0"
+    "react-router-dom": "6.3.0",
+    "request-address": "^0.0.3"
```
thisconnect, line 24: Minor detail: for non-devdeps I would prefer to lock it in the end, but feel free to keep the semver range in staging.

To add a locked npm version do `npm install --save-exact request-address` or preferably `npm install --save-exact --ignore-scripts request-address`. The ignore-scripts flags does not execute postinstall hooks, there is almost never a reason to execute a script when installing npm deps.

frontends/web/src/api/backend.ts
```diff
@@ -77,3 +82,19 @@ export const getConfig = (): Promise<any> => {
...
+  return (): Promise<string> => {
+    return apiGet(`exchange/pocket/buy/${accountCode}`);
+  };
+};
+
+export const isPocketSupported = (accountCode: string) => {
+  return (): Promise<boolean> => {
+    return apiGet(`exchange/pocket/supported/${accountCode}`);
+  };
+};
```
thisconnect, line 100: maybe this is the right moment to start `api/exchanges.ts` and also move pocket-addr-sig in there?

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -0,0 +1,61 @@
...
+          onClick={() => goToExchange('pocket')} >
+          {t('Pocket')}
+        </Button>) }
+      </div>
+
+    </div>
+  );
+};
+
+export const Exchange = translate()(ExchangeComponent);
```
thisconnect, line 61: Better use hooks instead of decorator, since this is a functional component

```ts

const ExchangeComponent: FunctionComponent<Props> = ({ code }) => {
```

backend/backend.go
```diff
@@ -683,3 +685,62 @@ func (backend *Backend) HandleURI(uri string) {
 		backend.log.Warningf("Unknown URI scheme: %s", uri)
 	}
 }
+
+// PocketWidgetSignAddress returns an unused address and makes the user sign a message to prove ownership.
+func (backend *Backend) PocketWidgetSignAddress(code accounts.Code, msg string, format string) (string, string) {
+	log := backend.log.WithField("accountCode", code)
+	var account accounts.Interface
+	for _, acct := range backend.accounts {
```
benma, line 693: move `getAccountFromCode()` from the handlers here and use that in both places. That one also checks if the account is active.

frontends/web/src/api/backend.ts
```diff
@@ -30,6 +30,11 @@ export interface ISuccess {
     errorCode?: string;
 }
 
+export type AddrSignResp = {
```
benma, line 33: nit: we usually spell out all words, `AddressSignResponse`

backend/handlers/handlers.go
```diff
@@ -204,10 +205,13 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/socksproxy/check", handlers.postSocksProxyCheck).Methods("POST")
 	getAPIRouter(apiRouter)("/exchange/moonpay/buy-supported/{code}", handlers.getExchangeMoonpayBuySupported).Methods("GET")
 	getAPIRouter(apiRouter)("/exchange/moonpay/buy/{code}", handlers.getExchangeMoonpayBuy).Methods("GET")
+	getAPIRouter(apiRouter)("/exchange/pocket/buy/{code}", handlers.getExchangePocketBuy).Methods("GET")
+	getAPIRouter(apiRouter)("/exchange/pocket/supported/{code}", handlers.getExchangePocketSupported).Methods("GET")
 	getAPIRouter(apiRouter)("/aopp", handlers.getAOPPHandler).Methods("GET")
 	getAPIRouter(apiRouter)("/aopp/cancel", handlers.postAOPPCancelHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/aopp/approve", handlers.postAOPPApproveHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/aopp/choose-account", handlers.postAOPPChooseAccountHandler).Methods("POST")
+	getAPIRouter(apiRouter)("/pocket-addr-sign", handlers.postPocketWidgetSignAddress).Methods("POST")
```
benma, line 214: nit: we usually spell out words: address not addr.

super-nit: `address-sign` vs `SignAddress` - why the different order of the words? :smile: 

could also do `/pocket/...` in case more endpoints should come.

backend/backend.go
```diff
@@ -683,3 +685,62 @@ func (backend *Backend) HandleURI(uri string) {
 		backend.log.Warningf("Unknown URI scheme: %s", uri)
 	}
 }
+
+// PocketWidgetSignAddress returns an unused address and makes the user sign a message to prove ownership.
+func (backend *Backend) PocketWidgetSignAddress(code accounts.Code, msg string, format string) (string, string) {
```
benma, line 690: return (string, string, error) and return all errors and propagate them all the way to the frontend, which should be able to react/show failure modes, if the user aborted, etc.

backend/backend.go
```diff
@@ -683,3 +685,62 @@ func (backend *Backend) HandleURI(uri string) {
 		backend.log.Warningf("Unknown URI scheme: %s", uri)
 	}
 }
+
+// PocketWidgetSignAddress returns an unused address and makes the user sign a message to prove ownership.
```
benma, line 689: document both return values more clearly, especially that the sig is returned in base64

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { translate, TranslateProps } from '../../decorators/translate';
+import { FunctionComponent } from 'react';
```
benma, line 18: we don't use that anymore afaik, see https://github.com/digitalbitbox/bitbox-wallet-app/pull/1839

backend/handlers/handlers.go
```diff
@@ -1068,6 +1072,34 @@ func (handlers *Handlers) getExchangeMoonpayBuy(r *http.Request) (interface{}, e
...
+	return acct, nil
+}
+
+func (handlers *Handlers) getExchangePocketBuy(r *http.Request) (interface{}, error) {
+	acct, err := handlers.getAccountFromCode(mux.Vars(r)["code"])
+	if err != nil {
+		return nil, err
+	}
+
+	return exchanges.BuyPocket(acct)
```
benma, line 1084: fyi, errors returned like this are swallowed in Qt (we should refactor error handling in the handlers). if you want to handle errors, for now we normally return a struct with a success flag and error message.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  return (
+    <div className="contentWithGuide">
+      <div className="container">
+        <div className={style.header}>
+          <Header title={<h2>{t('buy.info.title', { name })}</h2>} />
+        </div>
+        <div ref={ref} className="innerContainer">
+          <div className="noSpace" style={{ height }}>
```
benma, line 124: slightly confused, what is double braces? and isn't `height` an int not a style?

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -0,0 +1,144 @@
...
+      // handle address request from moonpay
+      if (message.type === V0MessageType.RequestAddress && message.withMessageSignature) {
+        signAddr(
+          message.withScriptType ? message.withScriptType : 'p2wpkh',
+          String(message.withMessageSignature),
+          code)
+          .then( resp => {
+            console.log(JSON.stringify(resp));
+
+            if (!resp.addr || !resp.sig ) {
```
benma, line 86: @thisconnect wasn't there a linter for spaces? also above before `resp`

backend/backend.go
```diff
@@ -683,3 +685,62 @@ func (backend *Backend) HandleURI(uri string) {
...
+	}
+	addr := unused[signingConfigIdx].Addresses[0]
+
+	sig, err := backend.keystore.SignBTCMessage(
+		[]byte(msg),
+		addr.AbsoluteKeypath(),
+		account.Config().SigningConfigurations[signingConfigIdx].ScriptType(),
+	)
+	if err != nil {
+		if firmware.IsErrorAbort(err) {
```
benma, line 737: this would definitely be helpful to propagate. maybe an 'user aborted' flag/msg should even be accepted by pocket? we should ask them about it.

thisconnect (changes requested):

Beerosagos:
Converted to draft until I finish addressing all the comments, I'll be pushing partial improvements in the meantime :pray:

frontends/web/src/routes/buy/pocket.tsx
thisconnect: made a PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/1847

frontends/web/package.json
Beerosagos: Done :+1:

frontends/web/src/api/backend.ts
Beerosagos: ACK

frontends/web/src/routes/buy/exchange.tsx
Beerosagos: Removed until we don't decide which texts should be used and add the locales entry. There was the same mistake also in `pocket.tsx`, fixed there.

backend/exchanges/pocket.go
Beerosagos: Right. Slightly modified the function to add TBTC support, as you proposed here https://github.com/digitalbitbox/bitbox-wallet-app/pull/1846#discussion_r1023795009

backend/backend.go
Beerosagos: ACK

backend/exchanges/pocket.go
Beerosagos: Nice! Didn't notice `CanSignMessage` method :) Done :+1:

backend/backend.go
Beerosagos: Done :+1:

frontends/web/src/api/backend.ts
Beerosagos: ACK

backend/handlers/handlers.go
Beerosagos: Ok! about the super-nit: I think I changed it in a second moment and didn't notice the discrepancy :D

backend/backend.go
Beerosagos: Right, done!

backend/backend.go
Beerosagos: Done, hope it is fine :)

frontends/web/src/routes/buy/exchange.tsx
Beerosagos: Nice! Fixed

backend/handlers/handlers.go
Beerosagos: Fixed

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: spaces fixed, in the meantime

backend/backend.go
Beerosagos: Added an `abort` flag in the `Response` struct. Atm the frontend just shows an alert, but we could send a message to the widget for this specific case, if we decide so.

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -0,0 +1,61 @@
...
+          primary
+          onClick={() => goToExchange('moonpay')} >
+          {t('Moonpay')}
+        </Button>) }
+      </div>
+      <div>
+        { showPocket && (<Button
+          primary
+          onClick={() => goToExchange('pocket')} >
+          {t('Pocket')}
```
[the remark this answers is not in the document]
Beerosagos: Sorry, was just a placeholder :) Removed

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const showPocket = useLoad(isPocketSupported(code));
+  const showMoonpay = useLoad(isMoonpayBuySupported(code));
+
+  const goToExchange = (exchange: string) => {
+    route(`/buy/${exchange}/${code}`);
+  };
+
+  return (
+    <div>
+      <div>{t('Choose you exchange!')}</div>
```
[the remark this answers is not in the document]
Beerosagos: Also here, just a placeholder. Removed until we don't define layout and texts of the page

backend/exchanges/pocket.go
```diff
@@ -0,0 +1,88 @@
...
+	return (coinCode == coin.CodeBTC || coinCode == coin.CodeTBTC) && canSign
+}
+
+// PocketWidgetSignAddress returns an unused address and makes the user sign a message to prove ownership.
+// Input params:
+// 	`account` is the account from which the address is derived, and that will be linked to the Pocket order.
+// 	`message` is the message that will be signed by the user with the private key linked to the address.
+//	`format` is the script type that should be used in the address derivation, as received by the widget
+//		(see https://github.com/pocketbitcoin/request-address#requestaddressv0messagescripttype).
+// 	`aoppBTCScriptTypeMap` is the map used in the AOPP flow to get the `ScriptType` object related to the `format` param.
```
Beerosagos, line 51: Not sure if this is totally correct. @benma do you think we should make a dedicated struct or see any other better solution?

Beerosagos:
Thanks @thisconnect and @benma for the review. Took some time to address the comment and to refactor the backend. 

I modified most of the function names, hope they are more intuitive now. I also moved `PocketWidgetSignAddress` in `pocket.go`, so that now all the logic should be in the same file. 

Could you take another look when you have a moment?

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -0,0 +1,151 @@
...
+      setHeight(ref.current.offsetHeight);
+    }, 200);
+  };
+
+  const onMessage = (e: MessageEvent) => {
+    if (!iframeURL) {
+      return;
+    }
+
+    const iframeOrigin = iframeURL.match(/^https?:\/\/[^/?]+/)![0];
```
thisconnect, line 75: Maybe change to `new URL(location.href).origin`. If the url can be invalid it should be in a try/catch, but I think in this case it is fine.

frontends/web/src/routes/buy/pocket.tsx
benma: TIL!

benma (approved):
LGTM, awesome

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: Ah nice! Thank you :) Now should be better

thisconnect (approved):
LGTM

PR #1847, opened 2022-11-17
frontend: eslint consistent space in parentheses

thisconnect:
Added space-in-parens rule and fixed with "make webfix", see
- https://eslint.org/docs/latest/rules/space-in-parens


benma (approved):
nice!

PR #1848, opened 2022-11-21
frontend: restyled buy/send/receive buttons on mobile

shonsirsha:
To improve the UI on mobile, we'd like to increase the size (height) of these buttons.

Before:

Screen Shot 2022-11-21 at 13 16 32

After:

Screen Shot 2022-11-21 at 13 16 18


thisconnect (approved):
untested LGTM

PR #1849, opened 2022-11-22
frontend: restyled header

shonsirsha:
In order to improve our UI, we'd like to restyle our app's header component.

Many files get changed due to the way our header currently works -> It's not actually "fixed" / "sticky" styled, but it was put outside of a scrollable container. So, to make `Header` not "sticky" / "fixed" we have to move the `Header` inside a scrollable container div **for each component** that uses the `Header`.


frontends/web/src/routes/account/account.tsx
```diff
@@ -214,33 +221,20 @@ export function Account({
...
-                  <span>{t('button.receive')}</span>
-                </Link>
-                { moonpayBuySupported && (
-                  <Link key="buy" to={`/buy/info/${code}`} className={style.buy}>
-                    <span>{t('button.buy')}</span>
-                  </Link>
-                )}
+              <div className="flex flex-row flex-between flex-item-center flex-column-mobile flex-reverse-mobile">
+                <Balance balance={balance} />
+                <label className="labelXLarge flex-self-start-mobile hide-on-large show-on-small">
```
thisconnect, line 231: I'd assume `show-on-small` should take care is `hide-on-large` really needed?

thisconnect (changes requested):
Added a question, still need to do some more testing, but overall 40a38b5842a7141ad4fa57a47d9225628dab3e99 is good.

frontends/web/src/routes/account/account.tsx
shonsirsha: Yeah that's right. I didn't notice `.show-on-small` has `display:none` on desktop lol. Change has been made.

thisconnect:
Tested a bit, I personally would remove the sticky from all view expect for TX overview where the view can become really long.

I would remove the sticky header from all other views, such as send or configure custom nodes.

But up to you LGTM otherwise.

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -83,32 +86,40 @@ class Moonpay extends Component<Props, State> {
...
+                  }}
+                  title="Moonpay"
+                  width="100%"
+                  height={iframeLoaded ? height : 0}
+                  frameBorder="0"
+                  className={style.iframe}
+                  allow="camera; payment"
+                  src={`${moonpay.url}&colorCode=%235E94BF`}>
+                </iframe>
+              </div>
```
shonsirsha, line 122: Also to prevent this UI bug from happening:

Screen Shot 2022-11-28 at 07 21 13

shonsirsha:
> Tested a bit, I think it is ok to remove the sticky from all views expect for TX overview as this one can become really long over time. So I'd remove it also on send and custom node views etc. WDYT?
> 
> other than that LGTM

Oops, I missed those - totally agree! Made changes, PTAL 🙏 🙂

thisconnect (approved):
partly-tested 0d5836e36f9d0d2bbf78416dc4a23d292bb6f6cb LGTM

Please add a changelog entry that the header design was updated.

Note: this changed also headers of BB01 device views, i.e. setup, but not all of them. i.e BB01 goal.jsx and security.jsx are missing. But I think it is fine. The other comp that is sticky is account-overview and technically also buy info disclaimer.

shonsirsha:
> partly-tested [0d5836e](https://github.com/digitalbitbox/bitbox-wallet-app/commit/0d5836e36f9d0d2bbf78416dc4a23d292bb6f6cb) LGTM
> 
> Please add a changelog entry that the header design was updated.
> 
> Note: this changed also headers of BB01 device views, i.e. setup, but not all of them. i.e BB01 goal.jsx and security.jsx are missing. But I think it is fine. The other comp that is sticky is account-overview and technically also buy info disclaimer.

Yes thanks, that's correct 👍  these are technically still sticky. However, the container below these components' header is not scrollable (not `scrollableContainer`). So at this point there'd be no difference.

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -83,32 +86,40 @@ class Moonpay extends Component<Props, State> {
...
-                frameBorder="0"
-                className={style.iframe}
-                allow="camera; payment"
-                src={`${moonpay.url}&colorCode=%235E94BF`}>
-              </iframe>
+          <div ref={this.ref} className="innerContainer scrollableContainer">
+            <div className={style.header}>
+              <Header title={<h2>{t('buy.info.title', { name })}</h2>} />
+            </div>
+            <div ref={this.ref} className="innerContainer">
```
thisconnect, line 107: This ref and className appear twice

PR #1851, opened 2022-11-24
buy/exchange: add logic for exchange selection

Beerosagos:
This adds:

** Dynamic header
- header logic to display different coin names based on what the user wants to buy (as already implemented for other pages in /buy/ workflow).

** Region support for exchanges
- region list encoded following ISO 3166-1
- moonpay availability based on region (exploiting moonpay API REST call)
- localization of region names when displayed in `exchanges.tsx`

~~Please note that Pocket API REST service for supported regions is not ready yet. For this reason at the moment Pocket is displayed as available in every region.~~

** Deal details for exchanges
- Details of the deals offered by Pocket and Moonpay are available in `exchanges.tsx` with `FAST` and `BEST DEAL` tags.

** What is still missing: 
- `exchange.tsx` layout is just drafted to verify the logic. It still misses completely the layout, which will be added in a separated PR.

### DIFF
--- a/backend/exchanges/exchanges.go
+++ b/backend/exchanges/exchanges.go
@@ -1,0 +1,110 @@
+// Copyright 2022 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exchanges
+
+import (
+	"net/http"
+
+	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
+)
+
+// regionCodes is an array containing ISO 3166-1 alpha-2 code of all regions.
+// Source: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
+var regionCodes = []string{
+	"AD", "AE", "AF", "AG", "AI", "AL", "AM", "AO", "AQ", "AR", "AS", "AT", "AU", "AW", "AX", "AZ",
+	"BA", "BB", "BD", "BE", "BF", "BG", "BH", "BI", "BJ", "BL", "BM", "BN", "BO", "BQ", "BR", "BS",
+	"BT", "BV", "BW", "BY", "BZ", "CA", "CC", "CD", "CF", "CG", "CH", "CI", "CK", "CL", "CM", "CN",
+	"CO", "CR", "CU", "CV", "CW", "CX", "CY", "CZ", "DE", "DJ", "DK", "DM", "DO", "DZ", "EC", "EE",
+	"EG", "EH", "ER", "ES", "ET", "FI", "FJ", "FK", "FM", "FO", "FR", "GA", "GB", "GD", "GE", "GF",
+	"GG", "GH", "GI", "GL", "GM", "GN", "GP", "GQ", "GR", "GS", "GT", "GU", "GW", "GY", "HK", "HM",
+	"HN", "HR", "HT", "HU", "ID", "IE", "IL", "IM", "IN", "IO", "IQ", "IR", "IS", "IT", "JE", "JM",
+	"JO", "JP", "KE", "KG", "KH", "KI", "KM", "KN", "KP", "KR", "KW", "KY", "KZ", "LA", "LB", "LC",
+	"LI", "LK", "LR", "LS", "LT", "LU", "LV", "LY", "MA", "MC", "MD", "ME", "MF", "MG", "MH", "MK",
+	"ML", "MM", "MN", "MO", "MP", "MQ", "MR", "MS", "MT", "MU", "MV", "MW", "MX", "MY", "MZ", "NA",
+	"NC", "NE", "NF", "NG", "NI", "NL", "NO", "NP", "NR", "NU", "NZ", "OM", "PA", "PE", "PF", "PG",
+	"PH", "PK", "PL", "PM", "PN", "PR", "PS", "PT", "PW", "PY", "QA", "RE", "RO", "RS", "RU", "RW",
+	"SA", "SB", "SC", "SD", "SE", "SG", "SH", "SI", "SJ", "SK", "SL", "SM", "SN", "SO", "SR", "SS",
+	"ST", "SV", "SX", "SY", "SZ", "TC", "TD", "TF", "TG", "TH", "TJ", "TK", "TL", "TM", "TN", "TO",
+	"TR", "TT", "TV", "TW", "TZ", "UA", "UG", "UM", "US", "UY", "UZ", "VA", "VC", "VE", "VG", "VI",
+	"VN", "VU", "WF", "WS", "YE", "YT", "ZA", "ZM", "ZW"}
+
+// ExchangeRegionList contains a list of ExchangeRegion objects.
+type ExchangeRegionList struct {
+	Regions []ExchangeRegion `json:"regions"`
+}
+
+// ExchangeRegion contains the ISO 3166-1 alpha-2 code of a specific region and a boolean
+//for each exchange, indicating if that exchange is enabled for the region.
+type ExchangeRegion struct {
+	Code             string `json:"code"`
+	IsMoonpayEnabled bool   `json:"isMoonpayEnabled"`
+	IsPocketEnabled  bool   `json:"isPocketEnabled"`
+}
+
+// PaymentMethod type is used for payment options in exchange deals.
+type PaymentMethod string
+
+const (
+	// CardPayment is a payment with credit/debit card.
+	CardPayment PaymentMethod = "card"
+	// BankTransferPayment is a payment with bank transfer.
+	BankTransferPayment PaymentMethod = "bank-transfer"
+)
+
+// ExchangeDeal represents a specific purchase option of an exchange.
+// - Fee indicates form the percentage that goes to the exchange in a float representation (e.g. 0.01 -> 1%).
+// - Payment is the payment method offered in the deal (usually different payment methods bring different fees).
+// - IsFast is usually associated with card payments. It is used by the frontend to display the `fast` tag in deals list.
+type ExchangeDeal struct {
+	Fee     float32       `json:"fee"`
+	Payment PaymentMethod `json:"payment"`
+	IsFast  bool          `json:"isFast"`
+}
+
+// ListExchangesByRegion populates an array of `ExchangeRegion` objects representing the availability
+// of the various exchanges in each of them.
+// NOTE: if one of the endpoint fails for any reason, the related exchange will be set as available in any
+// region by default.
+func ListExchangesByRegion(httpClient *http.Client) ExchangeRegionList {
+	moonpayRegions, moonpayError := GetMoonpaySupportedRegions(httpClient)
+	log := logging.Get().WithGroup("exchanges")
+	if moonpayError != nil {
+		log.Error(moonpayError)
+	}
+
+	pocketRegions, pocketError := GetPocketSupportedRegions(httpClient)
+	if pocketError != nil {
+		log.Error(pocketError)
+	}
+
+	exchangeRegions := ExchangeRegionList{}
+	for _, code := range regionCodes {
+		// default behavior is to show the exchange if the supported regions check fails.
+		moonpayEnabled, pocketEnabled := true, true
+		if moonpayError == nil {
+			_, moonpayEnabled = moonpayRegions[code]
+		}
+		if pocketError == nil {
+			_, pocketEnabled = pocketRegions[code]
+		}
+		exchangeRegions.Regions = append(exchangeRegions.Regions, ExchangeRegion{
+			Code:             code,
+			IsMoonpayEnabled: moonpayEnabled,
+			IsPocketEnabled:  pocketEnabled,
+		})
+	}
+
+	return exchangeRegions
+}
--- a/backend/exchanges/moonpay.go
+++ b/backend/exchanges/moonpay.go
@@ -1,7 +1,22 @@
+// Copyright 2022 Shift Crypto AG
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package exchanges
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 
@@ -21,13 +36,16 @@ const (
 	// - KYC always succeeds; simply click on "submit anyway"
 	// - need to provide a valid email addr to receive a check code;
 	//   any temp email service like fakermail.com will do
-	moonpayAPITestPubKey = "pk_test_e9i4oaa4J7eKo8UI3Wm8QLagoskWGjXN"
-	moonpayAPITestURL    = "https://buy-staging.moonpay.com"
+	moonpayBuyAPITestPubKey = "pk_test_e9i4oaa4J7eKo8UI3Wm8QLagoskWGjXN"
+	moonpayBuyAPITestURL    = "https://buy-staging.moonpay.com"
 
 	// moonpayAPILivePubKey is the production API key for real transactions.
 	// It is ok for it to be public.
-	moonpayAPILivePubKey = "pk_live_jfhWEt55szMLar8DhQWWiDwteX1mftY"
-	moonpayAPILiveURL    = "https://buy.moonpay.com"
+	moonpayBuyAPILivePubKey = "pk_live_jfhWEt55szMLar8DhQWWiDwteX1mftY"
+	moonpayBuyAPILiveURL    = "https://buy.moonpay.com"
+
+	// moonpayAPILiveURL is the API url for REST calls.
+	moonpayAPILiveURL = "https://api.moonpay.com/v3"
 )
 
 // Here's the list of all supported currencies:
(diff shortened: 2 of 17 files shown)


thisconnect (changes requested):

frontends/web/src/routes/account/utils.ts
```diff
@@ -15,7 +15,21 @@
...
+import { CoinCode, ScriptType, IAccount } from '../../api/account';
+
+export function getAccount(accounts: IAccount[], accountCode: string | undefined): IAccount | undefined {
+  if (!accounts) {
+    return undefined;
+  }
+  return accounts.find(({ code }) => accountCode === code);
+}
+
+export function getCryptoName(account: IAccount | undefined, t: (s: string) => string): string {
```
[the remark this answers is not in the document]
Beerosagos: Cool, thx :)

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -14,44 +14,202 @@
...
 
 interface TProps {
     code: string;
+    accounts: IAccount[];
+
+}
+
+interface Option {
+    text: string;
+    value: string | undefined;
```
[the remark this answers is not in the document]
Beerosagos: ACK

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -14,44 +14,202 @@
...
-  const showPocket = useLoad(isPocketSupported(code));
-  const showMoonpay = useLoad(isMoonpayBuySupported(code));
+export const Exchange = ({ code, accounts }: TProps) => {
+  const { t } = useTranslation();
+
+  const [showPocket, setShowPocket] = useState(false);
+  const [showMoonpay, setShowMoonpay] = useState(false);
+  const [selectedRegion, setSelectedRegion] = useState('');
+  const [selectedExchange, setSelectedExchange] = useState('');
+  const [regions, setRegions] = useState({} as Option[]);
```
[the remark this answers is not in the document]
Beerosagos: ACK

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -14,44 +14,202 @@
...
+  const [locale, setLocale] = useState('');
+  const [bestDeal, setBestDeal] = useState({} as exchangesAPI.ExchangeDeal);
```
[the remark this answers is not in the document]
Beerosagos: ACK

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -14,44 +14,202 @@
...
+        setShowPocket(!!isPocketSupportedCoin && region.isPocketEnabled);
+        setShowMoonpay(!!isMoonpaySupportedCoin && region.isMoonpayEnabled);
+        return;
+      }
+    });
+
+  }, [isPocketSupportedCoin, isMoonpaySupportedCoin, selectedRegion, regionList]);
+
+  const findBestDeal = (deals: exchangesAPI.ExchangeDeal[]): exchangesAPI.ExchangeDeal => {
+    var best = deals[0];
```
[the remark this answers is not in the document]
Beerosagos: ACK

frontends/web/src/routes/account/utils.ts
```diff
@@ -15,7 +15,21 @@
...
 
-import { CoinCode, ScriptType } from '../../api/account';
+import { CoinCode, ScriptType, IAccount } from '../../api/account';
+
+export function getAccount(accounts: IAccount[], accountCode: string | undefined): IAccount | undefined {
+  if (!accounts) {
+    return undefined;
+  }
+  return accounts.find(({ code }) => accountCode === code);
+}
```
thisconnect, line 25: There is an api function with a similar name `getAccounts` in `api/account.ts`, please rename getAccount to something else, maybe findAccount ?

Also there is an  almost identical function in send.tsx see https://github.com/digitalbitbox/bitbox-wallet-app/blob/936332e17d9949836d44f572cfbaedaa3c5fda44/frontends/web/src/routes/account/send/send.tsx#L475-L480

Worth refactoring? Also you could move this whole file to `src/utils`.

frontends/web/src/routes/account/utils.ts
```diff
@@ -15,7 +15,21 @@
  * limitations under the License.
  */
 
-import { CoinCode, ScriptType } from '../../api/account';
+import { CoinCode, ScriptType, IAccount } from '../../api/account';
+
+export function getAccount(accounts: IAccount[], accountCode: string | undefined): IAccount | undefined {
+  if (!accounts) {
+    return undefined;
+  }
```
thisconnect, line 23: accounts is alway at least an empty array `[]` and never falsy, `if (!accounts) {` can be removed.

frontends/web/src/routes/account/utils.ts
```diff
@@ -15,7 +15,21 @@
  * limitations under the License.
  */
 
-import { CoinCode, ScriptType } from '../../api/account';
+import { CoinCode, ScriptType, IAccount } from '../../api/account';
+
+export function getAccount(accounts: IAccount[], accountCode: string | undefined): IAccount | undefined {
```
thisconnect, line 20: accountCode should be mandatory, this requires making code not optional in info.tsx, see later comment.

frontends/web/src/routes/buy/moonpay.tsx
```diff
@@ -85,33 +84,15 @@ class Moonpay extends Component<Props, State> {
...
     const { agreedTerms, height, iframeLoaded } = this.state;
-    const account = this.getAccount();
-
+    const account = getAccount(this.props.accounts, this.props.code);
     if (!account || moonpay.url === '') {
       return null;
     }
 
-    const name = this.getCryptoName();
+    const name = getCryptoName(this.props.t, account);
```
thisconnect, line 95: Instead of passing  `t` function, how about using the 2nd arg to pass the cryptoName string, like this:

```js
getCryptoName(account, t('buy.info.crypto'));
```

(I know we pass t function in other places, that we should clean up one day)

frontends/web/src/routes/buy/info.tsx
```diff
@@ -75,19 +75,6 @@ class BuyInfo extends Component<Props, State> {
...
-    }
-    const account = accounts.find(account => account.code === code);
-    if (account) {
-      return isBitcoinOnly(account.coinCode) ? 'Bitcoin' : t('buy.info.crypto');
-    }
-    return t('buy.info.crypto');
-  };
-
   public render() {
     const { t } = this.props;
```
thisconnect, line 79: nit: destruct `accounts` and `code` here, so you can use those variables instead of `this.props.accounts` and `this.props.code`

frontends/web/src/routes/buy/info.tsx
```diff
@@ -97,7 +84,10 @@ class BuyInfo extends Component<Props, State> {
     if (options === undefined) {
       return <Spinner text={t('loading')} />;
     }
-    const name = this.getCryptoName();
+
+    const account = getAccount(this.props.accounts, this.props.code);
+    const name = getCryptoName(this.props.t, account);
+
```
thisconnect, line 90: I think it is better to make props.code mandatory in this component, it should always use a code or not?

Then code should be mandatory in getAccount.

Didn't test but TypeScript seems happy:

```diff
diff --git a/frontends/web/src/routes/account/utils.ts b/frontends/web/src/routes/account/utils.ts
index 584d48bf0..7fb42d74d 100644
--- a/frontends/web/src/routes/account/utils.ts
+++ b/frontends/web/src/routes/account/utils.ts
@@ -17,10 +17,7 @@
 
 import { CoinCode, ScriptType, IAccount } from '../../api/account';
 
-export function getAccount(accounts: IAccount[], accountCode: string | undefined): IAccount | undefined {
-  if (!accounts) {
-    return undefined;
-  }
+export function getAccount(accounts: IAccount[], accountCode: string): IAccount | undefined {
   return accounts.find(({ code }) => accountCode === code);
 }
 
diff --git a/frontends/web/src/routes/buy/info.tsx b/frontends/web/src/routes/buy/info.tsx
index efd3f6adb..5fe798f24 100644
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -28,7 +28,7 @@ import style from './info.module.css';
 
 interface BuyInfoProps {
     accounts: IAccount[];
-    code?: string;
+    code: string;
 }
 
 interface Option {
diff --git a/frontends/web/src/routes/router.tsx b/frontends/web/src/routes/router.tsx
index 5850ae52d..42b904a5b 100644
--- a/frontends/web/src/routes/router.tsx
+++ b/frontends/web/src/routes/router.tsx
@@ -81,6 +81,7 @@ export const AppRouter = ({ devices, deviceIDs, devicesKey, accounts, activeAcco
 
   const BuyInfoEl = <InjectParams>
     <BuyInfo
+      code={''}
       accounts={activeAccounts} />
   </InjectParams>;
 ```

**Review by thisconnect — CHANGES_REQUESTED** [review]

**Comment by thisconnect** [comment]
I think frontend is almost good, see last minor changes.. but for backend review @benma PTAL

File: frontends/web/src/routes/account/utils.ts (hunk 6 above)
Comment (Beerosagos) [reply frontends/web/src/routes/account/utils.ts:22]: Function renamed and `send.tsx` refactored. I would move the file in a separate PR, since that involves changes to several other files.

File: frontends/web/src/routes/account/utils.ts (hunk 7 above)
Comment (Beerosagos) [reply frontends/web/src/routes/account/utils.ts]: Done, thanks :)

File: frontends/web/src/routes/account/utils.ts (hunk 8 above)
Comment (Beerosagos) [reply frontends/web/src/routes/account/utils.ts]: Done :+1:

File: frontends/web/src/routes/buy/moonpay.tsx (hunk 9 above)
Comment (Beerosagos) [reply frontends/web/src/routes/buy/moonpay.tsx]: Nice, done!

File: frontends/web/src/routes/buy/info.tsx (hunk 10 above)
Comment (Beerosagos) [reply frontends/web/src/routes/buy/info.tsx]: No need for destructing variables, I think. Just used `const { t, accounts, code } = this.props;`

File: frontends/web/src/routes/buy/info.tsx (hunk 11 above)
Comment (Beerosagos) [reply frontends/web/src/routes/buy/info.tsx:90]: Good point, done!

**Comment by Beerosagos** [comment]
> I think frontend is almost good, see last minor changes.. but for backend review @benma PTAL

Thanks @thisconnect! Pushed an update with last nits fixed and a new endpoint `exchange/buy-supported/${code}`, used in `Info.tsx` to filter the accounts based on moonpay+pocket support.

File: frontends/web/src/routes/buy/info.tsx (hunk 10 above)
Comment (thisconnect) [reply frontends/web/src/routes/buy/info.tsx]: nice 👍

File: frontends/web/src/routes/buy/exchange.tsx [hunk 12]
```diff
@@ -14,44 +14,203 @@
...
+          Next
+            </Button>
+          </div>
+        </div>
+      </div>
+      <Guide name={name} />
     </div>
+
   );
 };
```
thisconnect, line 216: I didn't fully test this component.

thisconnect (approved):
Frontend looks good to me, but note: I didn't test the exchanges component, as afaik  there is more work planned on this component.

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -14,44 +14,203 @@
...
+  const [showPocket, setShowPocket] = useState(false);
+  const [showMoonpay, setShowMoonpay] = useState(false);
+  const [selectedRegion, setSelectedRegion] = useState('');
+  const [selectedExchange, setSelectedExchange] = useState('');
+  const [regions, setRegions] = useState<TOption[]>([]);
+  const [locale, setLocale] = useState('');
+  const [bestDeal, setBestDeal] = useState<exchangesAPI.ExchangeDeal>();
+
+  const isPocketSupportedCoin = useLoad(exchangesAPI.isPocketSupported(code));
+  const isMoonpaySupportedCoin = useLoad(exchangesAPI.isMoonpayBuySupported(code));
```
thisconnect, line 55: one more thing, iirc useLoad by default only fires once, so if code changes for some reason it will not requst again.

Found this again when looking at the other PR that tests the hooks.

If this is a problem, i.e. if code can change at anytime (but I think it doesn't right?) then you would need to specify a 2nd arg and pass `[code]` as dependencies so it can watch for changes in the dependencyList and make a new request if code changed.

backend/handlers/handlers.go
```diff
@@ -1032,6 +1036,44 @@ func (handlers *Handlers) postExportAccountSummary(_ *http.Request) (interface{}
...
+}
+
+func (handlers *Handlers) getExchangeBuySupported(r *http.Request) (interface{}, error) {
+	acct, err := handlers.backend.GetAccountFromCode(mux.Vars(r)["code"])
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Offline() can be removed from here once there is a unified way of initializing accounts
+	// and showing sync status, offline/fatal states, etc.
```
benma, line 1069: please avoid TODOs in the code, more often than not they stay nearly forever. better to create issues in the issue tracker about things that need to be improved than use the codebase as an issue tracker :)

frontends/web/src/i18n/i18n.js
```diff
@@ -100,8 +100,8 @@ i18n.on('languageChanged', (lng) => {
     if (!match) {
       // There are too many combinations. So, we compare only the main
       // language tag.
-      const lngLang = lng.replace('_', '-').split('-')[0];
-      const localeLang = nativeLocale.replace('_', '-').split('-')[0];
+      const lngLang = localeMainLanguage(lng);
+      const localeLang = localeMainLanguage(nativeLocale);
```
benma, line 104: deserves its own commit. i'd always try to think about if a commit needed to be reverted, you'd only want to revert one logical change, not many. or if you bisect a bug, you should land on a commit with one change, not many.

backend/exchanges/exchanges.go
```diff
@@ -0,0 +1,82 @@
...
+	Payment PaymentMethod `json:"payment"`
+	IsFast  bool          `json:"isFast"`
+}
+
+// ListExchangesByRegion populates an array of `ExchangeRegion` objects representing the availability
+// of the various exchanges in each of them.
+func ListExchangesByRegion(httpClient *http.Client) (ExchangeRegionList, error) {
+	moonpayRegions, err := GetMoonpaySupportedRegions(httpClient)
+	if err != nil {
+		return ExchangeRegionList{}, err
```
benma, line 62: here and below: it would not be good if the buy page couldn't load just because one of the two api endpoints is not responding correctly. is that the case now? if so, let's think about graceful failures.

would be especially bad if one exchange endpoint failing would also stop the other from being accessible, but also the buy widget could work even if the countries api does not respond correctly.

backend/handlers/handlers.go
```diff
@@ -1032,6 +1036,44 @@ func (handlers *Handlers) postExportAccountSummary(_ *http.Request) (interface{}
...
+func (handlers *Handlers) getExchangesByRegion(r *http.Request) (interface{}, error) {
+	type errorResult struct {
+		Error string `json:"error"`
+	}
+	regions, err := exchanges.ListExchangesByRegion(handlers.backend.HTTPClient())
+	if err != nil {
+		handlers.log.Error(err)
+		return errorResult{Error: err.Error()}, nil
+	}
+	return regions, err
```
benma, line 1048: `return regions, nil` is idiomatic, otherwise it's a bit confusing and looks like a real error is returned.

backend/handlers/handlers.go
```diff
@@ -206,7 +207,10 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/certs/download", handlers.postCertsDownloadHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/electrum/check", handlers.postElectrumCheckHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/socksproxy/check", handlers.postSocksProxyCheck).Methods("POST")
+	getAPIRouter(apiRouter)("/exchange/regions", handlers.getExchangesByRegion).Methods("GET")
```
benma, line 210: call it `/exchange/by-regions`? i prefer the naming to be consistent.

frontends/web/src/api/exchanges.ts
```diff
@@ -24,6 +24,38 @@ export type AddressSignResponse = {
   address: string;
   error: string;
 }
+export type ExchangeRegionList = {
```
benma, line 27: this does not cover the error case which the backend can respond with

frontends/web/src/routes/buy/exchange.tsx
Beerosagos: I don't think that the code can change inside this page, should be ok as is

backend/exchanges/exchanges.go
```diff
@@ -0,0 +1,82 @@
...
+const (
+	// CardPayment is a payment with credit/debit card.
+	CardPayment PaymentMethod = "card"
+	// BankTransferPayment is a payment with bank transfer.
+	BankTransferPayment PaymentMethod = "bank-transfer"
+)
+
+// ExchangeDeal represents a specific purchase option of an exchange.
+type ExchangeDeal struct {
+	Fee     float32       `json:"fee"`
```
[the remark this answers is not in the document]
Beerosagos: Done

backend/exchanges/moonpay.go
```diff
@@ -77,6 +81,48 @@ type BuyMoonpayParams struct {
 	Lang string // user preferred language in ISO 639-1; falls back to "en"
 }
 
+// BuyMoonpayRegion represents informations collected by Moonpay supported countries REST call.
+type BuyMoonpayRegion struct {
+	Alpha2       string `json:"alpha2"`
```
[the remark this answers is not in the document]
Beerosagos: Done

frontends/web/src/i18n/i18n.js
Beerosagos: Good point! Done

frontends/web/src/routes/account/utils.ts
```diff
@@ -15,7 +15,18 @@
...
+
+export function findAccount(accounts: IAccount[], accountCode: string): IAccount | undefined {
+  return accounts.find(({ code }) => accountCode === code);
+}
+
+export function getCryptoName(cryptoLabel: string, account?: IAccount): string {
+  if (account) {
+    return isBitcoinOnly(account.coinCode) ? 'Bitcoin' : cryptoLabel;
+  }
+  return cryptoLabel;
```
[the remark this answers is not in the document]
Beerosagos: Done

backend/exchanges/util.go
```diff
@@ -0,0 +1,42 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
+)
+
+// APIGet performs a HTTP Get call to a given endpoint and unmarshal the result
+// populating a given object.
+func APIGet(httpClient *http.Client, endpoint string, result interface{}) error {
+	req, err := http.NewRequest("GET", endpoint, nil)
```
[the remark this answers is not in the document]
Beerosagos: Could be done using a context with timeout, like rates update? https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/rates/rates.go#L270

With a more general scope, adding a timeout to the `httpClient` could be a solution? https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/util/socksproxy/socksproxy.go#L103

I could open a dedicated issue for this. wdyt?

frontends/web/src/api/exchanges.ts
```diff
@@ -24,6 +24,38 @@ export type AddressSignResponse = {
...
+  isFast: boolean;
+}
+
+export type ExchangeDeals = {
+  pocket: ExchangeDeal[];
+  moonpay: ExchangeDeal[];
+}
+
+export const getExchangesByRegion = () => {
+  return (): Promise<ExchangeRegionList> => {
```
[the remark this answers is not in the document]
Beerosagos: Now should be better! Thanks :)

backend/handlers/handlers.go
Beerosagos: Thanks, fixed

frontends/web/src/api/exchanges.ts
Beerosagos: There is no error anymore. See my previous comment

Beerosagos:
Thanks @benma @thisconnect for the review! I hope I addressed all the point, PTAL :pray:

benma (approved):
lgtm

PR #1852, opened 2022-11-25
frontends: color variables refactoring

thisconnect:
Separated text and background colors into their own variables. This makes colors easier to maintain because now it's possible to change text and background colors without affecting each other.

Also some text used --color-softblack instead of a designated /* font colors */ variable. Changed this text to --color-default so it is consistent with the other text colors.

- changed text using color-grey to color-secondary
- created --color-tertiary and --background-tertiary that use --color-gray-alt
- removed unused text-darkgray, text-black and text-bold

TODO: for the future --color-blue to --color-info

### DIFF
--- a/frontends/web/src/routes/settings/manage-accounts.module.css
+++ b/frontends/web/src/routes/settings/manage-accounts.module.css
@@ -12,10 +12,6 @@
     flex-shrink: 0;
 }
 
-.setting.disabled > p {
-    color: var(--color-gray-alt);
-}
-
 .coinLogo {
     position: relative;
     z-index: 2;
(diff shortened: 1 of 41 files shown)


frontends/web/src/style/variables.css
```diff
@@ -21,10 +21,13 @@
     --color-lightorange: #F9AC60;
 
     /* font colors */
+    /* all fonts should use these colors and not from the main colors directly */
```
shonsirsha, line 24: 👍

frontends/web/src/routes/settings/manage-accounts.module.css
```diff
@@ -113,7 +109,7 @@
 .tokenActive {}
```
shonsirsha, line 109: Not related but this could be removed I guess

shonsirsha (approved):
LGTM 👍

PR #1854, opened 2022-11-27
backend/coin/conversions: never return nil for Conversions

benma:
The frontend types IAmount.Conversions to always be a map, never null. If null is returned, the frontend errors in rates.tsx when accessing `amount.conversions[active]`, crashing the whole frontend.

This fix has already been done before in
32387d1b9814eeeb44d69ff79dc8daaa1924f879, but accidentally undone again in f5a2a823cd406bde2a30f273a6e646606a38210f.


Beerosagos (approved):
LGTM! Thanks

thisconnect:
Thinking this actually nicer than https://github.com/digitalbitbox/bitbox-wallet-app/commit/2c5cf147fe8f095a762f1f35619b58b911076197

PR #1855, opened 2022-11-27
backend/electrum: make Headers/GetMerkle rpc calls blocking/synchronous

benma:


backend/coins/btc/headers/headers.go
```diff
@@ -433,6 +436,7 @@ func (headers *Headers) VerifiedHeaderByHeight(height int) (*wire.BlockHeader, e
 }
 
 func (headers *Headers) kick() {
+	time.Sleep(30 * time.Second)
```
Beerosagos, line 439: Was this left here intentionally? In my test, it seems to slow down the sync, which takes forever.

backend/coins/btc/headers/headers.go
```diff
@@ -180,7 +180,10 @@ func (headers *Headers) Initialize() {
 			headers.update(header.BlockHeight)
 		},
 	)
-	headers.kickChan <- struct{}{}
+	go func() {
```
Beerosagos, line 183: Why do we need the 30s sleep?

backend/coins/btc/headers/headers.go
benma: omg no, thanks for finding it. quite embarrassing :sweat_smile:

backend/coins/btc/headers/headers.go
benma: we don't, it's a debug leftover :cold_sweat: thanks, undone!

Beerosagos (approved):
tACK!

PR #1857, opened 2022-11-28
ci: use macos 11

thisconnect:
The macOS 10.15 Actions runner image is beeing deprecated.

See https://github.com/actions/runner-images/issues/5583


thisconnect:
alternative would be to use `macos-latest` cc @benma

benma (approved):

Beerosagos (approved):
LGTM

PR #1859, opened 2022-11-29
frontend: display offline errors on account summary

thisconnect:
Collect status.offlineError of all different accounts and if there are any, render the errors instead of the account summary.

This can happen if Tor is down or misconfigured or internet not reachable.

### DIFF
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -27,6 +27,7 @@ import { FiatConversion } from '../../../components/rates/rates';
 import { Check } from '../../../components/icon/icon';
 import Logo from '../../../components/icon/logo';
 import Spinner from '../../../components/spinner/ascii';
+import { Spinner as ErrorSpinner } from '../../../components/spinner/Spinner';
 import { debug } from '../../../utils/env';
 import { Chart } from './chart';
 import { AddBuyReceiveOnEmptyBalances } from '../info/buyReceiveCTA';
@@ -52,6 +53,7 @@ interface State {
     exported: string;
     balances?: Balances;
     syncStatus?: SyncStatus;
+    offlineErrors: string[];
     totalBalancePerCoin?: accountApi.ITotalBalance;
 }
 
@@ -75,6 +77,7 @@ class AccountsSummary extends Component<Props, State> {
   public readonly state: State = {
     data: undefined,
     exported: '',
+    offlineErrors: [],
     totalBalancePerCoin: undefined,
   };
   private unsubscribe!: () => void;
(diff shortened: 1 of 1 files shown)


thisconnect:
@benma this currently only looks at offlineError but ignores fatalError.

Also it collects the offlineErrors from all accounts and shows them all. This only works the first time, but when navigating to other accounts and back somehow it doesn't receive the errors anymore 🤔

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -157,6 +160,10 @@ class AccountsSummary extends Component<Props, State> {
 
   private async onStatusChanged(code: string, initial: boolean = false) {
     const status = await accountApi.getStatus(code);
+    const { offlineErrors } = this.state;
+    if (status.offlineError && !offlineErrors.includes(status.offlineError)) {
+      this.setState({ offlineErrors: [...offlineErrors, status.offlineError] });
```
thisconnect, line 165: could also push `status.fatalError && t('account.fatalError')` into this array, or replace the array completely if 1 account had fatalError.

On the other hand the offlineError might contain useful information like Tor misconfiguration.

Beerosagos: I would not replace the array in case of fatalError, I agree with you that offlineError could be useful

Beerosagos:
> Also it collects the offlineErrors from all accounts and shows them all. This only works the first time, but when navigating to other accounts and back somehow it doesn't receive the errors anymore thinking

Tried to investigate this, but couldn't really figure it out. It seems to me that after a while backend account handlers stop working correctly, stalling frontend `status` requests. Could be related to a mutex that will not be unlocked? I tried to debug removing some locks and seems to improve (even if it doesn't solve it completely).

cc @benma

Beerosagos (approved):
LGTM

frontends/web/src/routes/account/summary/accountssummary.tsx
benma: what if you get online again? this code seems to never reset offlineErrors when back online.

benma:
I also couldn't see the reason after staring at it for 30min. Will need more time on this. 

It also happens without this PR, if you have an error and you switch to account->portfolio->account, the backend does not respond on many handlers.

Ideally we refactor and unit test account status management, it grew to be quite a mess.

benma:
It's probably because /balance and /transactions block until synced, but it might never sync. We should make these non blocking and return an error instead, with notifications to load them when ready or something like that.

No clue why /status would not respond, it looks like it should simply deliver a response immediately :thinking:

thisconnect:
rebased

thisconnect:
closing this is very old, in the meantime we adding offlineerros in the account view  https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/65da562618acf34c309ac21d0efc804de7fe6af2

PR #1861, opened 2022-11-29
frontend: created unit tests for hooks

shonsirsha:
as a part of improving the frontend's code quality and robustness, we'd like to increase its test coverage.

This commit is for adding unit tests for all of the hooks that we use in the frontend. Technically we now have 100% test coverage for our hooks - however this certainly still can be improved even further in the future.

All tests are passing 70/70 ✅ 
Screen Shot 2022-11-29 at 14 14 08

note: I'll create a separate PR for some component tests later on to avoid a huge PR.


frontends/web/package.json
```diff
@@ -5,6 +5,7 @@
   "devDependencies": {
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.1.5",
+    "@testing-library/react-hooks": "^8.0.1",
```
shonsirsha, line 8: [about this (npm page)](https://www.npmjs.com/package/@testing-library/react-hooks) 🙂

frontends/web/src/hooks/utils.test.ts
```diff
@@ -14,24 +14,15 @@
...
-    const matchMedia = window.matchMedia(query);
-    handleChange();
-    matchMedia.addEventListener('change', handleChange);
-    return () => {
-      matchMedia.removeEventListener('change', handleChange);
-    };
-  }, [query]);
-
-  return matches;
-};
```
shonsirsha, line 37: As we're not using this right now, I'll remove it.

frontends/web/src/hooks/keyboard.test.ts
```diff
@@ -14,24 +14,21 @@
...
-    handleChange();
-    matchMedia.addEventListener('change', handleChange);
-    return () => {
-      matchMedia.removeEventListener('change', handleChange);
-    };
-  }, [query]);
+});
 
-  return matches;
-};
```
shonsirsha, line 37: As we're not using this right now, I'll remove it.

frontends/web/src/hooks/keyboard.test.ts
```diff
@@ -14,24 +14,21 @@
...
-  useEffect(() => {
-    const handleChange = () => {
-      setMatches(getMatches(query));
-    };
-    const matchMedia = window.matchMedia(query);
-    handleChange();
-    matchMedia.addEventListener('change', handleChange);
-    return () => {
-      matchMedia.removeEventListener('change', handleChange);
-    };
```
shonsirsha, line 33: I'm unsure why this file is marked as moved/renamed when it's not related at all. But here, I created a test file for keyboard hook, and removed useMatchMedia as it's unused. :)

thisconnect (commented):
Thanks looks great! I'll need a bit more time to study it.

Beerosagos (approved):
Looks very good to me, but I don't feel like being knowledgeable enough on hooks and react testing to give proper feedback. I'll stay at the window for @thisconnect comments :)

frontends/web/src/hooks/api.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(result.current).toBe(MOCK_RETURN_STATUS);
+  });
+
+  it('useLoad should load promise and return the correct resolved value', async () => {
+    const mockApiCall = jest.fn().mockImplementation(() => () => Promise.resolve(true));
+    const { result, waitForNextUpdate } = renderHook(() => useLoad(mockApiCall()));
+    await waitForNextUpdate();
+    expect(result.current).toBe(true);
+  });
```
thisconnect, line 53: @shonsirsha if possible, could you add another test for `useLoad` that tests the 2nd argument as well?

I've added the 2nd arg a while ago (not 100% sure if this was a good idea), as useLoad only does 1 request on mount,  
see my comment here for an example: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1851/files#r1039794716

thisconnect (changes requested):
nice 👍 , it all makes sense, lots of mocking especially the ones in  `hooks/api.ts` are hard. @Beerosagos I recommend to read the simpler one, don't look at `hooks/api.ts` 🤯 

added some comments an a request for another test for useLoad and a rename. But already LGTM.

frontends/web/src/hooks/api.test.ts
shonsirsha: Yes sure, I'll write a test for the 2nd arg. In this test, I'll assert that the apiCall should be called however many times the dependencies (2nd arg) changes.  🙂

frontends/web/src/hooks/utils.test.ts
```diff
@@ -0,0 +1,30 @@
...
+import { renderHook } from '@testing-library/react-hooks';
+import { useMountedRef } from './utils';
+
+describe('useMountedRef', () => {
+
+  it('should return true on mount and false on unmount', () => {
+    const { result, unmount } = renderHook(() => useMountedRef());
+    expect(result.current).toEqual({ current: true });
+    unmount();
+    expect(result.current).toEqual({ current: false });
```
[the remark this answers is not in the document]
shonsirsha: That's correct! Actual value of the hook (the bool) is `result.current.current`. We compare as objects because the hook returns `{current: bool}`.

shonsirsha:
Changes made + rebased master - PTAL @thisconnect 🙏

thisconnect (approved):
Thanks, LGTM 👍

PR #1862, opened 2022-11-29
frontend: full screen menu on mobile

shonsirsha:
to improve our app's UI and achieving a more modern look, we'd like to make the sidebar menu / navigation to be of full screen size on mobile.

[Preview] Screen recording on mobile:

https://user-images.githubusercontent.com/28676406/204563461-d42905f8-3f87-4935-9160-481d918f20b6.mov

[Preview] Screenshot on desktop (no change):
Screen Shot 2022-11-29 at 15 45 51

### DIFF
--- a/frontends/web/src/components/sidebar/sidebar.css
+++ b/frontends/web/src/components/sidebar/sidebar.css
@@ -1,3 +1,15 @@
+.closeButton {
+    background: transparent;
+    border: none;
+    display: none;
+    padding: 12px;
+    transform: translateX(14px);
+}
+
+.closeButton:focus{
+    outline: none;
+}
+
 .sidebarOverlay {
     position: fixed;
     background-color: rgba(0, 0, 0, 0.3);
@@ -42,11 +54,11 @@
 }
 
 .sidebar .sidebarLogoContainer {
+    align-items: center;
     display: flex;
     flex-direction: row;
-    justify-content: center;
-    align-items: center;
     height: 70px;
+    justify-content: space-between;
     padding: calc(var(--spacing-default) + var(--spacing-half)) var(--spacing-large);
     background-size: cover;
     background-color: rgba(0, 0, 0, 0.1);
@@ -75,25 +87,6 @@
     line-height: var(--sidebar-header-line-height);
     color: var(--color-secondary);
     text-transform: uppercase;
-}
-
-.sidebarHeaderAction {
-    height: var(--sidebar-header-line-height);
-}
-
-.sidebarHeaderAction > a {
-    height: 100% !important;
-}
-
-.sidebarHeaderAction > a > svg {
-    height: var(--sidebar-header-line-height);
-    width: auto;
-    stroke: var(--color-gray);
-    transition: stroke 0.2s ease;
-}
-
-.sidebarHeaderAction:hover > a > svg {
-    stroke: var(--color-white);
 }
 
 .sidebarItem {
@@ -228,8 +221,16 @@ a.sidebar-active .single img,
         margin-left: 0;
         width: var(--sidebar-width-large);
     }
+}
 
-    .sidebarContainer:not(.forceHide) .sidebar .sidebarLogoContainer a {
-        display: none !important;
+@media (max-width: 560px) {
+    .closeButton {
+        display:block;
+    }
+
+    .sidebar {
+        margin-left: -100vw;
+        transition: margin-left 0.3s ease;
+        width: 100vw;
     }
 }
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/sidebar/sidebar.css
```diff
@@ -1,3 +1,15 @@
+.closeButton {
+    background: transparent;
+    border: none;
+    display: none;
+    padding: 12px;
```
shonsirsha, line 5: padding is not used for spacing, but to increase the button's touchable / surface area which improves UX (user doesnt have to touch exactly the X button, but surrounding is fine).

frontends/web/src/components/sidebar/sidebar.css
```diff
@@ -1,3 +1,15 @@
...
+    transform: translateX(14px);
```
shonsirsha, line 6: this is to "negate" the padding above  (otherwise the X will appear a little bit to the left). 

no big deal, but 14 is used here and not 12 (actual padding val) because of the extra whitespace brought in by the actual `CloseXWhite` component itself.

frontends/web/src/components/sidebar/sidebar.css
```diff
@@ -217,8 +228,14 @@ a.sidebar-active .single img,
...
 @media (max-width: 1199px) {
+    .closeButton {
+        display:block;
+    }
+
     .sidebar {
-        transition: margin-left 0.2s ease;
+        margin-left: -100vw;
+        transition: margin-left 0.3s ease;
+        width: 100vw;
```
thisconnect, line 238: 1199px is still desktop mode, but with hamburger (unfortunatelly, but one day we fix that).

Could you pick a smaller breakpoint that the navigation menu only goes into fullscreen on mobile portrait, but not on small desktop. Ideally landscape would be as it was before, but mobile portrait would be fullscreen.

thisconnect (changes requested):

frontends/web/src/components/sidebar/sidebar.css
shonsirsha: Ok. I agree with you.

We currently don't have an exact "mobile" breakpoint, but I'll use a the smallest breakpoint that is currently used in the project which is <=560px.

shonsirsha: Changes made, PTAL 🙏

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -158,13 +159,17 @@ class Sidebar extends Component<Props> {
...
-            <div className="sidebarLogoContainer">
+          <div className="flex flex-between sidebarLogoContainer">
+            <Link
+              to={accounts.length ? '/account-summary' : '/'}
+              onClick={this.handleSidebarItemClick}>
               <AppLogoInverted className="sidebarLogo" />
-            </div>
-          </Link>
+            </Link>
+            <button className="closeButton" onClick={toggleSidebar}>
```
thisconnect, line 168: I just realized this component doesn't use CSS module and the CSS is all global. This component should probably use CSS module and use `className={style.closeButton}`. It would take renaming `sidebar.css` to `sidebar.module.css` and then import style from sidebar.module.css. But can be in a different PR or 2nd commit, up to you.

thisconnect (changes requested):

frontends/web/src/components/sidebar/sidebar.tsx
shonsirsha: Haha, it seems like only this component isn't using css module 😄 https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/style/index.css

Will migrate it to be using one here.

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -158,13 +159,17 @@ class Sidebar extends Component<Props> {
       <div className={['sidebarContainer', hidden ? 'forceHide' : ''].join(' ')}>
         <div className={['sidebarOverlay', activeSidebar ? 'active' : ''].join(' ')} onClick={toggleSidebar}></div>
         <nav className={['sidebar', activeSidebar ? 'forceShow' : '', shown ? 'withGuide' : ''].join(' ')}>
-          <Link
-            to={accounts.length ? '/account-summary' : '/'}
-            onClick={this.handleSidebarItemClick}>
-            <div className="sidebarLogoContainer">
+          <div className="flex flex-between sidebarLogoContainer">
```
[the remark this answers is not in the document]
shonsirsha: sure

frontends/web/src/components/sidebar/sidebar.tsx
thisconnect: wait the styles in `src/style/..` are supposed to be global.
but all css in `src/components/*` are supposed to be CSS modules

thisconnect (approved):
LGTM

shonsirsha:
Hm, looked like this was merged but the changes aren't on master. 🤔

PR #1864, opened 2022-11-29
frontend: fix buy container

thisconnect:
In f30df4f4a64f50789541fe153a24389b5fc1954d the iframe container accidentally contained 2 this.ref, which is wrong.

As the iframe scrolls anyway the header is always sticky. Removed scrollContainer classes, so that there is no double scroll.

### DIFF
--- a/frontends/web/src/routes/buy/moonpay.tsx
+++ b/frontends/web/src/routes/buy/moonpay.tsx
@@ -100,26 +100,24 @@ class Moonpay extends Component<Props, State> {
     return (
       <div className="contentWithGuide">
         <div className="container">
-          <div ref={this.ref} className="innerContainer scrollableContainer">
+          <div className="innerContainer">
             <div className={style.header}>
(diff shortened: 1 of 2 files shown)


thisconnect:
PTAL also test in Qt app please

shonsirsha (approved):
code LGTM. 👍

Beerosagos (approved):
tACK on QT linux

The bottom part of the iframe still remains not scrollable, but it is usable
immagine

thisconnect:
> The bottom part of the iframe still remains not scrollable, but it is usable

exactly, also there is a privacy policy link.. I think if the JS height calculation shoud take the header height into account... will look into that later.

also I don't get why the issue only exist in Qt but not in webdev... something to keep in mind, that there are subtle differences.

thisconnect:
@Beerosagos this commit is relevant to staging-pocket branch. Want to open master->staging-pocket PR ?

PR #1869, opened 2022-12-05
frontend: fix header background

thisconnect:
Looks like the change header design branch and refactor CSS variables canceled each others header background out.

Fixed and removed header color as it is supposed to be in:
- f30df4f4a64f50789541fe153a24389b5fc1954d


shonsirsha (approved):
LGTM.

PR #1871, opened 2022-12-05
frontend: added unit tests for several components

shonsirsha:
as a part of our refactoring efforts  in the FE, we'd like to increase the test coverage of our react components.

**Created unit tests for functional component ones**. This PR did not cover all functional components yet though - so the rest can be done in another PR to hopefully avoid a huge, longstanding PR. 🙂  thanks.

### DIFF
--- a/frontends/web/src/components/balance/balance.test.tsx
+++ b/frontends/web/src/components/balance/balance.test.tsx
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2022 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render } from '@testing-library/react';
+import { IBalance } from '../../api/account';
+import I18NWrapper from '../../i18n/forTests/i18nwrapper';
+import { Balance } from './balance';
+
+describe('components/balance/balance', () => {
+  it('renders balance properly', () => {
+    const MOCK_BALANCE: IBalance = {
+      hasAvailable: true,
+      hasIncoming: true,
+      available: {
+        amount: '0.005',
+        unit: 'BTC'
+      },
+      incoming: {
+        amount: '0.003',
+        unit: 'BTC',
+        conversions: {
+          BTC: '0.003',
+          AUD: '512',
+          BRL: '512',
+          CAD: '512',
+          CHF: '512',
+          CNY: '512',
+          EUR: '512',
+          GBP: '512',
+          HKD: '512',
+          ILS: '512',
+          JPY: '512',
+          KRW: '512',
+          NOK: '512',
+          RUB: '512',
+          SEK: '512',
+          SGD: '512',
+          USD: '512',
+        }
+      }
+    };
+    const { getByTestId } = render(<Balance balance={MOCK_BALANCE} />, { wrapper: I18NWrapper });
+    expect(getByTestId('availableBalance')).toHaveTextContent('0.005BTC');
+    expect(getByTestId('incomingBalance')).toHaveTextContent('+0.003 BTC / 512 USD');
+  });
+});
--- a/frontends/web/src/components/forms/button.test.tsx
+++ b/frontends/web/src/components/forms/button.test.tsx
@@ -1,0 +1,24 @@
+import { MemoryRouter } from 'react-router-dom';
+import { render, screen } from '@testing-library/react';
+import { ButtonLink, Button } from './button';
+
+describe('components/forms/button', () => {
+  describe('ButtonLink', () => {
+    it('renders as button when disabled with disabled attribute', () => {
+      render(<ButtonLink to="/settings" disabled>A ButtonLink</ButtonLink>);
+      expect(screen.getByRole('button', { name: /A ButtonLink/i })).toHaveAttribute('disabled');
+    });
+
+    it('renders as link when not disabled', () => {
+      render(<ButtonLink transparent to={'/settings'}>A ButtonLink</ButtonLink>, { wrapper: MemoryRouter });
+      expect(screen.getByRole('link', { name: /A ButtonLink/i })).toBeInTheDocument();
+    });
+  });
+
+  describe('Button', () => {
+    it('renders a button with proper styling', () => {
+      render(<Button primary>A Button</Button>);
+      expect(screen.getByRole('button', { name: /A Button/i })).toHaveClass('primary');
+    });
+  });
+});
--- a/frontends/web/src/components/forms/checkbox.test.tsx
+++ b/frontends/web/src/components/forms/checkbox.test.tsx
@@ -1,0 +1,26 @@
+/**
+ * Copyright 2022 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render, screen } from '@testing-library/react';
+import Checkbox from './checkbox';
+
+describe('components/forms/checkbox', () => {
+  it('renders checkbox with proper attributes', () => {
+    const { container } = render(<Checkbox label="my checkbox label" id="checkbox" />);
+    expect(container.firstChild).toHaveClass('default');
+    expect(screen.getByRole('checkbox', { name: 'my checkbox label' })).toBeInTheDocument();
+  });
+});
--- a/frontends/web/src/components/guide/entry.test.tsx
+++ b/frontends/web/src/components/guide/entry.test.tsx
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2022 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { render } from '@testing-library/react';
+import { Entry, TEntryProp } from './entry';
+
+describe('components/guide/entry', () => {
+  describe('renders correct entry values', () => {
+    it('collapsed', () => {
+      const EntryProps: TEntryProp = {
+        title: 'A title',
+        text: 'Some text',
+      };
+      const { container } = render(<Entry key={'key'} entry={EntryProps} />);
+
+      expect(container).toHaveTextContent('A title');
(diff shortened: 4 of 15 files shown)


frontends/web/src/components/forms/button.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+describe('components/forms/button', () => {
+  describe('ButtonLink', () => {
+    it('renders as button when disabled with disabled attribute', () => {
+      render(<ButtonLink to="/settings" disabled>A ButtonLink</ButtonLink>);
+      expect(screen.getByRole('button', { name: /A ButtonLink/i })).toHaveAttribute('disabled');
+    });
+
+    it('renders as link when not disabled', () => {
+      render(<ButtonLink transparent to={'/settings'}>A ButtonLink</ButtonLink>, { wrapper: MemoryRouter });
+      expect(screen.getByRole('link', { name: /A ButtonLink/i })).toBeInTheDocument();
```
thisconnect, line 14: I tried to find the documentation for toBeInTheDocument, found this https://testing-library.com/docs/guide-disappearance/ but it seems to be about async rendering/wait-utility. All I found about MemoryRouter is this example, https://testing-library.com/docs/example-react-router/ is this needed?

how does this work and is there a better link to the documentation?

frontends/web/src/components/forms/checkbox.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { render, screen } from '@testing-library/react';
+import Checkbox from './checkbox';
+
+describe('components/forms/checkbox', () => {
+  it('renders checkbox with proper attributes', () => {
+    const { container } = render(<Checkbox label="my checkbox label" id="checkbox" />);
+    expect(container.firstChild).toHaveClass('default');
+    expect(screen.getByRole('checkbox', { name: 'my checkbox label' })).toBeInTheDocument();
```
thisconnect, line 24: Sorry I don't know testing lib well, can you explain what exactly is tested to be inthedocument?

I don't understand exactly what getByRole finds, but I am thinking about DOM elements, maybe that is wrong?

The component renders a span containing `input[type="checkbox"]` element with a sibling  `label` element that contains the text.  So does getByRole find the "input type checkbox", the label element or the whole component or something in the "accessibility tree" (found https://testing-library.com/docs/queries/about#priority 🤔  )

frontends/web/src/components/guide/entry.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      const link = getByTestId('link');
+
+      expect(container).toHaveTextContent('A title');
+      expect(container).toHaveTextContent('Some text');
+      expect(container).toHaveTextContent('some url');
+      expect(link).toBeInTheDocument();
+      expect(link).toHaveAttribute('title', 'http://someurl.com');
+    });
+  });
+});
```
thisconnect, line 83: nice

frontends/web/src/components/headerssync/headerssync.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+describe('components/headerssync/headerssync', () => {
+  beforeEach(() => {
+    useMountedRefSpy.mockReturnValueOnce({ current: true });
+  });
+
+  it('renders null when subscribe value of null', () => {
+    useSubscribeSpy.mockReturnValueOnce(null);
+
+    const { container } = render(<HeadersSync coinCode="btc" />);
+    expect(container.firstChild).toBeNull();
```
thisconnect, line 36: noob q: I'd expect container to be null, because headersync componentn returns null if there is no status, but container returns div here .. what am I missing?

frontends/web/src/components/headerssync/headerssync.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+      };
+      useSubscribeSpy.mockReturnValueOnce(MOCKED_SUBSCRIBE_VALUE);
+
+      const { getByTestId, container } = render(<HeadersSync coinCode="btc" />);
+      const progressBar = getByTestId('progress-bar');
+      expect(container).toHaveTextContent('50%');
+      expect(progressBar.firstChild).toHaveStyle('width: 50%');
+    });
+  });
+});
```
thisconnect, line 69: nice, I don't know how but one test could be that the component hides after 4s after it fully synced to targetHeight.

frontends/web/src/components/language/language.test.tsx
```diff
@@ -25,8 +25,6 @@ jest.mock('react-i18next');
 const useTranslationSpy = useTranslation;
 
 describe('components/language/language', () => {
-
-
```
thisconnect, line 29: I wouldn't remove the lines as there are no other changes in this file, but no strong opinon.

thisconnect (approved):
nice LGTM, with many questions (sorry I am a testing lib noob)

frontends/web/src/components/balance/balance.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+          SGD: '512',
+          USD: '512',
+        }
+      }
+    };
+    const { getByTestId } = render(<Balance balance={MOCK_BALANCE} />);
+    expect(getByTestId('availableBalance')).toHaveTextContent('0.005BTC');
+    expect(getByTestId('incomingBalance')).toHaveTextContent('+0.003 BTC / 512 USD');
+  });
+});
```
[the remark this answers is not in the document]
shonsirsha: Hm, good catch! Unfort. using `useTranslation` wont solve this issue as some components, such as "balance" imports component(s) which imports another component(s).. etc that use translation. See below for details. I think in `language.test.tsx` we could use it because I believe we don't import  anything with translation in `language.tsx`. The only component using translation is **that component itself**.

**Anyhow, to fix this warning issue we could render the component (in this case `Balance`), inside a `<I18nextProvider/>`**. I will make this change! 🙂 



FYI, this is the error when mocking `useTranslation` in `Balance`:
```js
    TypeError: (0 , _translate.translate)(...) is not a function

      129 | }
      130 |
    > 131 | const HOC = translate()(share<SharedProps, TranslateProps>(store)(Guide));
          |             ^
      132 | export { HOC as Guide };
```
The trace:

```js
    at Object.<anonymous> (src/components/guide/guide.tsx:131:13)
      at Object.<anonymous> (src/components/layout/header.tsx:20:1)
      at Object.<anonymous> (src/components/layout/index.tsx:18:1)
      at Object.<anonymous> (src/components/view/view.tsx:19:1)
      at Object.<anonymous> (src/components/alert/Alert.tsx:21:1)
      at Object.<anonymous> (src/utils/request.ts:19:1)
      at Object.<anonymous> (src/utils/config.js:17:1)
      at Object.<anonymous> (src/components/rates/rates.tsx:22:1)
      at Object.<anonymous> (src/components/balance/balance.tsx:20:1)
      at Object.<anonymous> (src/components/balance/balance.test.tsx:19:1)
      at TestScheduler.scheduleTests (node_modules/@jest/core/build/TestScheduler.js:333:13)
      at runJest (node_modules/@jest/core/build/runJest.js:404:19)
```

frontends/web/src/components/forms/button.test.tsx
```diff
@@ -0,0 +1,24 @@
+import { MemoryRouter } from 'react-router-dom';
+import { render, screen } from '@testing-library/react';
+import { ButtonLink, Button } from './button';
+
+describe('components/forms/button', () => {
+  describe('ButtonLink', () => {
+    it('renders as button when disabled with disabled attribute', () => {
+      render(<ButtonLink to="/settings" disabled>A ButtonLink</ButtonLink>);
+      expect(screen.getByRole('button', { name: /A ButtonLink/i })).toHaveAttribute('disabled');
```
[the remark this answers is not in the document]
shonsirsha: From reading several articles, `getByRole` is indeed more suitable than `getByTestId` in some scenarios such as this one. 

I mixed the usage in our codebase because I used whatever is easiest to test. This component for example, clearly will only render one single button (or link), so there's no need to use `getByTestId` for example. Moreover, the one "plus" in using getByRole is accessibility, whereas using testid it will find  whatever element there is as long as testid is there (doesn't matter the "role").

"Role" is for example, here an element could be a button or a link - using getByTestId it will not "care" about it (sure we can maybe test its innerHTML or something - but it seems hacky). However, using `getByRole` we wanna assert that there's a component that **identifies itself as.. a button or a link that _also_ has bla bla bla (whatever the next assertion)** . 

Articles on why role > testid:
1. https://derekndavis.com/posts/getbytestid-overused-react-testing-library (this is a pretty good blog)
2. https://dev.to/jacques_blom/don-t-use-getbytestid-32oj

I tbh have no strong opinion on using role vs testid, however I do think it's nice that we can have some sort of confidence that our element indeed renders as whatever it intends to be.. e.g a button, or a link, or a checkbox and so on.

frontends/web/src/components/forms/button.test.tsx
shonsirsha: The reason why we need a wrapper here (which is the `MemoryRouter`)  is because this will render a `<Link/>` component from `react-router-dom`. And when rendering only a component on itself like this (the nature of unit tests), it is missing the "routing context" (in the app we use `BrowserRouter`).

So, we can also use `BrowserRouter` here, but `MemoryRouter` is preferred because it keeps the routing history independent of any external source.

Good sources to read (small reads):
1. https://learnwithparam.com/blog/different-types-of-router-in-react-router/ (what is MemoryRouter in simple words)
2. https://testing-library.com/docs/react-testing-library/api/#wrapper (what is a wrapper) -> supported by RTL since it's such a common usage
3. https://stackoverflow.com/questions/70904665/difference-between-memoryrouter-and-router-concerning-testing-with-react-testing (BrowserRouter vs MemoryRouter in RTL)
4. https://stackoverflow.com/questions/70805929/how-to-fix-error-usehref-may-be-used-only-in-the-context-of-a-router-compon

frontends/web/src/components/headerssync/headerssync.test.tsx
shonsirsha: Yeah, in first glance it really could look like that. The thing is `render` wraps a `div` (which is this `container`)  on whatever is being `render`ed.

So the `firstChild` is the actual element (`HeaderSync` in this case).

_"Container: By default, React Testing Library will create a div and append that div to the document.body and this is where your React component will be rendered"_ - https://testing-library.com/docs/react-testing-library/api/#container

frontends/web/src/components/language/language.test.tsx
shonsirsha: Removed these spaces because other test files have no spaces between the describe and it. 🙂

frontends/web/src/components/forms/checkbox.test.tsx
shonsirsha: No worries, I am also still learning 🙂 

`toBeInTheDocument` is quite simply asserting that whatever it is (in this case checkbox with a label "my checkbox label") to be within the document body, aka in the page. 

I haven't tested but I think here are all the "valid roles" in HTML5: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

>> So does getByRole find the "input type checkbox", the label element or the whole component or something in the "accessibility tree" (found...

That link you attached described exactly what it is and why it's the preferred method in most situations.

In this case, byRole checks the whole screen and see if there's a **checkbox** with the label **my checkbox label**. byRole can "connect" these two (label and checkbox) because of the `htmlFor` we use in the `label` and `id` in the `input checkbox` (`htmlFor` === `id`). 

byRole doesn't "see" per element (as opposed to getByTestId, getByid and similars), but it tries to act like a "user" e.g looking at it holistically. In this case, it's looking for "a checkbox with a label of bla bla..". Here, we can tell that we can only use this way of testing (byRole) if we follow accessibility best practices, which we should always try to do anyway.

shonsirsha:
@thisconnect I made the change to wrap the components that caused warnings in tests due to the `useTranslation` inside the `i18NProvider`.  Here is the [wrapper](https://github.com/digitalbitbox/bitbox-wallet-app/blob/2d374f01cc6acb3e975d43ead4f173d99113b3b4/frontends/web/src/i18n/forTests/I18nWrapper.tsx). PTAL once more? 🙏 no any major change just wanna make sure you're aware before I'm merging.

[Explanation in my comment above](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1871#discussion_r1040386449)🙂 

I hope this PR also helps us to understand more about the testing lib. As you can tell, there are many many ways to achieve the same test result or to assert something is X or Y. We can use byRole, byTestId, even checking their html, children nodes ,etc etc. Which ones are better or correct? I think most of these are preference / opinionated. I don't think there's one "correct" way per se though.

I am also still in the space of trying to explore more, but using getByRole whenever possible does make sense to me as RTL even recommend us to use it first and foremost before using anything else.  However, some things like our `progress-bar` is  difficult to be "seen" using `byRole` so I used `byTestId` for example.

PR #1876, opened 2022-12-11
Fix race conditions pt1

benma:
This fixes some potential race conditions. More to come.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"path"
 	"sort"
-	"sync"
 	"sync/atomic"
 
 	"github.com/btcsuite/btcd/btcutil"
@@ -102,8 +101,8 @@ type Account struct {
 	feeTargets     []*FeeTarget
 	feeTargetsLock locker.Locker
 	// Access this only via getMinRelayFeeRate(). sat/kB.
-	minRelayFeeRate   *btcutil.Amount
-	minRelayFeeRateMu sync.Mutex
+	minRelayFeeRate     *btcutil.Amount
+	minRelayFeeRateLock locker.Locker
 
 	// true when initialized (Initialize() was called).
 	initialized     bool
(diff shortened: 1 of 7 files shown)


backend/coins/btc/account.go
```diff
@@ -99,13 +99,15 @@ type Account struct {
...
+	feeTargets   []*FeeTarget
+	feeTargetsMu locker.Locker
 	// Access this only via getMinRelayFeeRate(). sat/kB.
 	minRelayFeeRate   *btcutil.Amount
 	minRelayFeeRateMu sync.Mutex
 
 	// true when initialized (Initialize() was called).
-	initialized bool
+	initialized   bool
+	initializedMu locker.Locker
```
Beerosagos, line 110: could this be an `atomic.bool` instead?

backend/coins/btc/account.go
```diff
@@ -99,13 +99,15 @@ type Account struct {
 	activeTxProposal     *maketx.TxProposal
 	activeTxProposalLock locker.Locker
 
-	feeTargets []*FeeTarget
+	feeTargets   []*FeeTarget
+	feeTargetsMu locker.Locker
```
Beerosagos, line 103: I see that in this structure we use `Locker` or `Mutex` in different places. Would be possible to uniform it and use Lockers everywhere? Apart from this: `feeTargetsMu` suggests that it is a Mutex. Maybe change to `feeTargetsLock` instead? Same for `initializedMu`

Beerosagos (approved):
tACK with small nits

backend/coins/btc/account.go
benma: No as the whole Initialize function needs to be serialized, not just writing to and reading from this variable.

frontends/web/src/api/account.ts
```diff
@@ -99,7 +99,7 @@ export type TSigningConfiguration = {
     ethereumSimple: TEthereumSimple;
 }
 
-export interface ISigningConfigurationList {
+export type ISigningConfigurationList = null | {
```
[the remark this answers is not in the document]
benma: Done

backend/coins/btc/account.go
benma: Done

PR #1878, opened 2022-12-11
backend/btc/transactionsdb: don't create buckets on Begin()

benma:
Begin() currently always opens a write-transaction, so we could init the buckets there. Most of the time we don't need write-transactions as read-only transactions suffice. This commit prepares using read-only transactions, which improve code clarity and performance.


Beerosagos (approved):
tACK!

PR #1879, opened 2022-12-11
backend/eth/account: avoid potential race conditions

benma:


backend/coins/eth/account.go
```diff
@@ -738,18 +749,19 @@ func (account *Account) TxProposal(
...
 	}}
 }
 
 // VerifyAddress implements accounts.Interface.
 func (account *Account) VerifyAddress(addressID string) (bool, error) {
-	if account.signingConfiguration == nil {
+	if !account.isInitialized() {
 		return false, errp.New("account must be initialized")
 	}
-	account.Synchronizer.WaitSynchronized()
```
Beerosagos, line 751: Not sure why this is not needed anymore. Could be also removed from `backend/coins/btc/account.go` in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1876/ ?

backend/coins/eth/account.go
```diff
@@ -68,23 +67,25 @@ type Account struct {
 	httpClient           *http.Client
 
 	// true when initialized (Initialize() was called).
-	initialized bool
+	initialized   bool
+	initializedMu locker.Locker
```
Beerosagos, line 71: nit: `initializedMu` -> `initializedLock` ?

backend/coins/eth/account.go
benma: It is not needed because a few lines above we check if the account is initialized (now with `if !account.isInitialized()`, before with `if account.signingConfiguration == nil`. Waiting for transactions/balance etc to sync is not needed to be able to verify the address.

Need to take a closer look in the btc account, as there, the address list can change during synchronization. The goal is definitely to remove all blocking `WaitSynchronized()` and replace them with error responses however, so it will go away one way or another :)

backend/coins/eth/account.go
benma: Done

Beerosagos (approved):
LGTM!

PR #1880, opened 2022-12-12
backend/btc: use block-client-go for the Electrum connection check

benma:
block-client-go is a new library for connecting to Electrum servers with support for proper timeouts, failover, a better API, etc.

This commit adds this dependency and for a start uses it to replace the CheckElectrumServer function with an easier and more robust version.

Note that the connection check still has not a custom-defined timeout as it uses the `socksproxy.GetTCPProxyDialer()` dialer, where no timeout is configured. This should be fixed in a future commit.


Beerosagos (approved):
tACK! very nice :)

PR #1883, opened 2022-12-13
frontends: dark mode

jadzeidan:
This commit contains refactoring required for dark mode and changes in the style sheet so the variables use dark mode colors.

There are still more parts that need to be done, such as:
- change the chart background color to a dark mode color
- change color of SVGs (such as "x" in dialogs and "save" button in notes)
- make dark mode images (such as logos)


jadzeidan:
For SVGs, we currently use `` tags. Does it make sense to convert certain ones (such as the "x" in dialogs") to `<svg>` tags instead? Then we can use CSS to color them.

frontends/web/src/components/icon/icon.tsx
```diff
@@ -33,6 +33,7 @@ import closeXWhiteSVG from './assets/icons/close-x-white.svg';
 import closeXDarkSVG from './assets/icons/close-x-dark.svg';
 import guideSVG from './assets/icons/guide.svg';
 import menuSVG from './assets/icons/menu.svg';
```
shonsirsha, line 35: we could remove this unused `menuSVG` import.

shonsirsha, line 35: we could remove this unused `menuSVG` import.

shonsirsha (changes requested):
Nice, good job so far! 👍 

prob still in WIP, regardless left a small comment.

frontends/web/src/components/icon/icon.tsx
jadzeidan: removed now

frontends/web/src/routes/account/summary/chart.module.css
```diff
@@ -142,8 +142,8 @@
   
 
 .tooltip {
-  background: white;
-  border: 1px solid #dedede;
+  background: var(--background-secondary);
+  border: 1px solid var(--background)
```
shonsirsha, line 146: missed a `;` here 😉

shonsirsha (changes requested):
minor one

frontends/web/src/routes/account/summary/chart.module.css
jadzeidan: Fixed

shonsirsha:
BTW looks good. I'll try to have a look again today on the code. Can you change this PR to be merging into staging-darkmode instead of master though? Thx!

shonsirsha:
We can safely close this due to: #1912 🙂

PR #1889, opened 2022-12-14
frontend: use view component for device settings

thisconnect:
Settings can use View, Grid and Column component for layout instead of plain div's. The View component is mostly used in fullscreen mode but an early version did support normal mode with sidebar and guide. Fixed some minor CSS issues (spacing etc.) so that normal mode should now be fully working.

This could also be used for normal settings and exchange integration in the future.

### DIFF
--- a/frontends/web/src/components/view/view.tsx
+++ b/frontends/web/src/components/view/view.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2022 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,9 +15,9 @@
  */
 
 import { FunctionComponent, ReactNode } from 'react';
-import { AppLogo } from '../icon';
-import { Footer } from '../layout';
-import { SwissMadeOpenSource } from '../icon/logo';
+import { LanguageSwitch } from '../language/language';
+import { Version } from '../layout/version';
+import { AppLogo, SwissMadeOpenSource } from '../icon/logo';
 import { AnimatedChecked, Close } from '../icon/icon';
 import style from './view.module.css';
 
@@ -85,9 +85,13 @@ export const View = ({
       )}
       {withBottomBar && (
         <div style={{ marginTop: 'auto' }}>
-          <Footer>
+          <footer className={style.footer}>
             <SwissMadeOpenSource />
-          </Footer>
+            <div className="m-right-half hide-on-small">
+              <Version />
+            </div>
+            <LanguageSwitch />
+          </footer>
         </div>
       )}
     </div>
--- a/frontends/web/src/components/view/view.module.css
+++ b/frontends/web/src/components/view/view.module.css
@@ -24,11 +24,15 @@
     justify-content: center;
 }
 
+/* this will be applied if fullscreen is false */
 .fill {
+    align-self: center;
     display: flex;
     flex-direction: column;
     flex-grow: 1;
-    padding-bottom: var(--space-default);
+    padding: 0 var(--space-default) var(--space-default) var(--space-default);
+    max-width: var(--content-width);
+    width: 100%;
 }
 
 .inner {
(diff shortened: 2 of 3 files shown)


thisconnect:
This should almost exactly look and behave like the master, except for:
- settings title now uses ViewTitle and is located in a different position (bigger and below)
- footer spacing is a bit off, View probably shouldn't use the old footer and could use it's own

thisconnect:
New

Screen Shot 2022-12-14 at 18 24 41

thisconnect:
Old

    
Screen Shot 2022-12-14 at 18 25 28

thisconnect:
New 
  
Screen Shot 2022-12-14 at 18 26 04

thisconnect:
Most of the CSS changes only impact `<View fullscreen={false}>` (all other instances of View use fullscreen), but the last commit refactors the footer. The footer change should also be tested with fullscreen, i.e. setup a new BB02

frontends/web/src/components/view/view.tsx
```diff
@@ -85,9 +85,13 @@ export const View = ({
...
         <div style={{ marginTop: 'auto' }}>
-          <Footer>
+          <footer className={style.footer}>
             <SwissMadeOpenSource />
-          </Footer>
+            <div className="m-right-half hide-on-small">
+              <Version />
+            </div>
+            <LanguageSwitch />
+          </footer>
```
shonsirsha, line 94: Hm, not rly a big deal but instead of sort of duplicating `Footer` here, would you think it makes sense to just re-style the `Footer` (to not use the inline `flex`, `flex-row`, `flex-end`,etc)? 

As it seems like this is the only difference between the `Footer` component and what we have above? 🤔

shonsirsha, line 94: Hm, not rly a big deal but instead of sort of duplicating `Footer` here, would you think it makes sense to just re-style the `Footer` (to not use the inline `flex`, `flex-row`, `flex-end`,etc)? 

As it seems like this is the only difference between the `Footer` component and what we have above? 🤔

shonsirsha:
Just a personal opinion, without the text beside the hamburger icon, the top part of the app looks a bit empty especially on mobile. Is this an agreed-upon design? 🙂

frontends/web/src/components/view/view.tsx
thisconnect: yes I thought about this, but then on the other hand `<Footer>` is only used in theses components, which I think all should be migrated to `<View>`

Screen Shot 2022-12-19 at 08 19 32

thisconnect:
The screenshot is out of date, I actually reverted it in the 2nd commit https://github.com/digitalbitbox/bitbox-wallet-app/pull/1889#issuecomment-1351844802 but after looking at it again, I think it looks better in the screenshot. The hamburger has not really anything todo with the title of the view, IMO the hamburger makes more sense alone.

> the top part of the app looks a bit empty. Is this an agreed-upon design?

do you prefer title next to the hamburger?

frontends/web/src/components/view/view.tsx
shonsirsha: Mhm, okay. So once we migrated all of these to be using the `View` components, we can remove `Footer` completely? And maybe if it's relevant enough, should we think of a way to pass children into the footer (if it's inside `View`)?

shonsirsha:
> The screenshot is out of date, I actually reverted it in the 2nd commit [#1889 (comment)](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1889#issuecomment-1351844802) but after looking at it again, I think it looks better in the screenshot. The hamburger has not really anything todo with the title of the view, IMO the hamburger makes more sense alone.
> 
> > the top part of the app looks a bit empty. Is this an agreed-upon design?
> 
> do you prefer title next to the hamburger?

Hm, yeah but this is a totally personal opinion. The reason for this is prob. the font size of the "Manage Device" vs "Secrets". To me they look like they're in the same hierarchy, when they aren't. Tagging @jadzeidan for his opinion 🙂

thisconnect:
I agree the view title on small screen should be bigger. 

> The reason for this is prob. the font size of the "Manage Device" vs "Secrets"

frontends/web/src/components/view/view.tsx
thisconnect: The view component could export `<ViewFooter>`, but right now it is not needed anywhere.

thisconnect:
merging as is without new design.

PR #1893, opened 2022-12-19
frontend: convert alert into functional and added tests

shonsirsha:
to improve our code in the FE, we'd like to convert following component into a functional one and added tests for it.

### DIFF
--- a/frontends/web/src/components/alert/Alert.tsx
+++ b/frontends/web/src/components/alert/Alert.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2022 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,8 +15,8 @@
  * limitations under the License.
  */
 
-import { Component } from 'react';
-import { withTranslation, WithTranslation } from 'react-i18next';
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import { MultilineMarkup } from '../../utils/markup';
 import { View, ViewButtons, ViewHeader } from '../view/view';
 import { Button } from '../forms';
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/alert/Alert.tsx
```diff
@@ -31,79 +14,54 @@ import { Button } from '../forms';
...
-      callback,
       asDialog = true,
     } = options;
-    this.callback = callback;
-    this.setState({
-      active: true,
-      asDialog,
-      message,
-    });
+    setCallback(() => options.callback);
```
thisconnect, line 32: This was confusing me for a bit, i.e. I expected `setCallback(() => options.callback());` or  `setCallback(options.callback);`, but that won't work and execute the callback too early, when the alert opens instead of when the user clicks ok.

I remembered that setState also accepted a function... current react docs doesn't explain the hook well or I am blind, for future me: https://beta.reactjs.org/apis/react/useState#setstate

question as the UI doesn't render differently nor has to update (I think), does it even make sense to keep the callback in the state? (assuming there is a simpler alternative i.e. in a local variable)

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -31,79 +14,54 @@ import { Button } from '../forms';
...
-}
+  const handleClose = () => {
+    if (callback) {
+      callback();
+    }
+    setActive(false);
+  };
 
-const TranslatedAlert = withTranslation()(Alert);
+  return active && message ? (
```
thisconnect, line 45: Nit: for readability it would help me read the code if the condition before conditional (ternary) operator was in parentheses

frontends/web/src/components/alert/Alert.tsx
```diff
@@ -31,79 +14,54 @@ import { Button } from '../forms';
...
+          </Button>
+        </ViewButtons>
+      </View>
+    </form>
+  ) : null;
+};
 
 export { alertUser };
-export { TranslatedAlert as Alert };
+export { Alert };
```
thisconnect, line 67: change to `export { Alert, alertUser };` ?

thisconnect (approved):
LGTM but please add license headers 

also added some more comments.

frontends/web/src/components/alert/Alert.tsx
shonsirsha: oops, good catch!

frontends/web/src/components/alert/Alert.tsx
shonsirsha: Yeah, it was also a bit confusing for me and also for a lot of people I believe, as storing a fn in react state isn't as "intuitive" as most we think. 🙂 

Yes sure, we could use local variable here and we can omit this `setCallback` thing :)

PR #1895, opened 2022-12-21
frontend: conditionally renders Spinner in buy/pocket

shonsirsha:
To fix the UI bug where Guide icon on the header appears twice

[Preview] Before
Screen Shot 2022-12-21 at 04 01 08

[Preview] After
Screen Shot 2022-12-21 at 03 57 39


Beerosagos (approved):
LGTM!

PR #1896, opened 2022-12-21
frontend/send: fix duplicate/no key in UTXOsClass

benma:
Fixes this error:

```
react_devtools_backend.js:4012 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `UTXOsClass`. See https://reactjs.org/link/warning-keys for more information.
    at div
    at UTXOsClass (http://localhost:8080/static/js/bundle.js:23710:5)
```

### DIFF
--- a/frontends/web/src/routes/account/send/utxos.tsx
+++ b/frontends/web/src/routes/account/send/utxos.tsx
@@ -105,7 +105,7 @@ export class UTXOsClass extends Component<Props, State> {
       return null;
     }
     return (
-      <div>
+      <div key={'utxos-' + scriptType}>
         <h2 className="subTitle">{ getScriptName(scriptType) }</h2>
         <ul className={style.utxosList}>
           { filteredUTXOs.map(utxo => (


frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -105,7 +105,7 @@ export class UTXOsClass extends Component<Props, State> {
       return null;
     }
     return (
-      <div>
+      <div key={'utxos-' + scriptType}>
```
benma, line 108: I do wonder why this is needed, after all

```
{ accountApi.allScriptTypes.map(this.renderUTXOs) }
```

the `allScriptTypes` is a constant:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/9f6a6dd5fb1e23454fae401d4fc0d7fd563b2c6c/frontends/web/src/api/account.ts#L77

Compiler should know that this list never changes? :thinking:

shonsirsha: > the allScriptTypes is a constant:
> Compiler should know that this list never changes? 🤔

Mhm, unfortunately I don't think this is the case. I don't think it'll know if allScriptTypes won't change by reading our code. Even when we're using `const` to store an array, it won't actually prevent us from changing the value of the array itself.

E.g

```js
const x = [1,2,3,4]
x[1] = 13;  // now x is [1, 13, 3, 4]
x.push(0); // now x is [1, 13, 3, 4, 0]
```

This is becuase what's stored in x is only the **reference** to the array  (the mem address where array is stored), instead of the actual **value** of the array (by ref instead of by val).

With this, react knows that the array may (or at least, it can) still be changed later on - thus needing "key". 

Now, even if the array never changes, we should still provide a "key" attribute for each element in the array because it allows React to optimise the rendering of the array. It's worth noting that even if the array itself never changes, other parts of the app may change and trigger a re-render of the list. 

Without the "key" attribute, React has no way to keep track of which elements are the same from one render to the next, so it has to re-render the entire list every time there is a change.   Hope that makes sense 🙂

shonsirsha (approved):
Untested LGTM 🚀

frontends/web/src/routes/account/send/utxos.tsx
shonsirsha: > the allScriptTypes is a constant:
> Compiler should know that this list never changes? 🤔

Mhm, unfortunately I don't think this is the case. I don't think it'll know if allScriptTypes won't change by reading our code. Even when we're using `const` to store an array, it won't actually prevent us from changing the value of the array itself.

E.g

```js
const x = [1,2,3,4]
x[1] = 13;  // now x is [1, 13, 3, 4]
x.push(0); // now x is [1, 13, 3, 4, 0]
```

This is becuase what's stored in x is only the **reference** to the array  (the mem address where array is stored), instead of the actual **value** of the array (by ref instead of by val).

With this, react knows that the array may (or at least, it can) still be changed later on - thus needing "key". 

Now, even if the array never changes, we should still provide a "key" attribute for each element in the array because it allows React to optimise the rendering of the array. It's worth noting that even if the array itself never changes, other parts of the app may change and trigger a re-render of the list. 

Without the "key" attribute, React has no way to keep track of which elements are the same from one render to the next, so it has to re-render the entire list every time there is a change.   Hope that makes sense 🙂

benma: The part about `const` I kind of knew but it's easy to forget. `const` is a misnomer :see_no_evil: 

Amazing explanation, learned a lot, especially about the re-rendering part. Thanks!

PR #1898, opened 2022-12-22
frontend: fix fontsize in dialogs regression

thisconnect:
In 484c42f59207f12f428f588935cd3628e603d83b the device settings was refactored to use View component. This had a sideeffect that --size-default CSS variable was changed on large screens and the cause of a regression that paragraph text in dialogs are rendered with a larged font.

Fixed by adding some CSS to guard against View CSS.

### DIFF
--- a/frontends/web/src/components/dialog/dialog.module.css
+++ b/frontends/web/src/components/dialog/dialog.module.css
@@ -26,6 +26,15 @@
     max-height: 100vh;
     overflow: auto;
 }
+/* guard dialog and wait-dialog from view styles */
+@media (min-width: 1200px) {
+    .header .title {
+        --size-subheader: 16px;
+    }
+    .modal .contentContainer p {
+        --size-default: 14px;
+    }
+}
 
 .modal.small {
     max-width: 340px;


thisconnect:
How to test:

- goto device settings
- open a dialog (i.e. show recovery words or change device name)
- check that fontsize looks nor unproportionally big

Here the font is too big (screenshots from current master):

frontends/web/src/components/dialog/dialog.module.css
```diff
@@ -26,6 +26,15 @@
...
     overflow: auto;
 }
+/* guard dialog and wait-dialog from view styles */
+@media (min-width: 1200px) {
+    .header .title {
+        --size-subheader: 16px;
+    }
+    .modal .contentContainer p {
+        --size-default: 16px;
+    }
```
shonsirsha, line 36: More of a design question, do we wanna keep what we have before? If so I think we should set the --size-default to be `14px`.

Original (14px):
Screen Shot 2022-12-26 at 09 24 53


This PR (16px):
Screen Shot 2022-12-26 at 09 28 59

I think 14px looks good, but we can reduce the top margin/spacing of the "You will be presented..." text. 16px looks slightly too big IMO. 🙂  But good catch regardless.

shonsirsha (approved):
lgtm with question 👍  -> tested with both 14px and 16px for `--size-default` - both works but I prefer 14px.

thisconnect:
changed to 14px https://github.com/digitalbitbox/bitbox-wallet-app/compare/510ef6b71c79188d7ba870686b39247c1fc16a0f..54dbd181309e31bc4a333a2433069bafb1e033de

PR #1901, opened 2022-12-26
frontend: style method name

shonsirsha:
to fulfil the UI requirements for the buy page

[Preview] now payment method has an icon:
Screen Shot 2022-12-26 at 18 13 07


frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
```diff
@@ -29,14 +29,22 @@
 }
 
 .paymentMethodName {
-    width: 50%;
+    color: var(--color-dark);
```
thisconnect, line 32: not yet, but soon we should not use color variables directly, see https://github.com/digitalbitbox/bitbox-wallet-app/blob/8d853d1c98aa540169f995d26f8b79ddc06f885a/frontends/web/src/style/variables.css#L23-L25

(I forgot to check in the last PR)

frontends/web/src/routes/buy/components/exchangeselectionradio.tsx
```diff
@@ -30,19 +31,27 @@ type TRadioProps = RadioProps & JSX.IntrinsicElements['input'];
...
+      </span>);
     default:
       return methodName;
     }
   };
+
   return (
     <div className={style.paymentMethodContainer}>
-      <p className={style.paymentMethodName}>{getPaymentMethodName(deal.payment)}</p>
+      {renderPaymentMethod(deal.payment)}
```
thisconnect, line 54: what do you think about changing PaymentMethod so it is a function component?
it could be the in the same module and only used internally, so not exported.

thisconnect (approved):
untested LGTM with 2 suggesstions

frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
shonsirsha: Sure. Changed all in this file but one, the `--color-skyblue` - as it's only used once in this entire app and we didn't define its "background colour" name yet.


> not yet, but soon we should not use color variables directly,

If so,  we should start thinking on defining the "names" (bg, font, status) for all these new colours that were introduced and are only used in this "buy" page. Bc atm, we're using it directly, e.g `--color-olive` or `--color-light-skyblue`,etc..

frontends/web/src/routes/buy/components/exchangeselectionradio.tsx
shonsirsha: cool

PR #1903, opened 2023-01-03
backend: whitelist pocketbitcoin's website URL

shonsirsha:
to allow opening of their website from the BitBoxApp.


benma (approved):
utACK

PR #1906, opened 2023-01-09
frontend: fix view component with guide on large screen

thisconnect:
On large screen the guide on device settings broke onto a 2nd line,
this is a regression introduced in:
- https://github.com/digitalbitbox/bitbox-wallet-app/commit/484c42f59207f12f428f588935cd3628e603d83b

Fixed by adding the old solution with 2 containers but as react
components. This does duplicate some CSS, but in the long run,
using the global CSS should be phased out.

Added overflow auto to the '.container' CSS, so that longer content
is scrollable. This should only impact the new react components
that is used in device settings and not have any side effect on
other views.

### DIFF
--- a/frontends/web/src/components/layout/guide-wrapper.tsx
+++ b/frontends/web/src/components/layout/guide-wrapper.tsx
@@ -1,5 +1,4 @@
 /**
- * Copyright 2018 Shift Devices AG
  * Copyright 2023 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/device/bitbox02/settings.tsx
```diff
@@ -61,67 +61,71 @@ export const Settings = ({ deviceID }: TProps) => {
...
-              { versionInfo && versionInfo.canGotoStartupSettings ? (
-                <GotoStartupSettings apiPrefix={apiPrefix} />
-              ) : null }
-            </Column>
-          </Grid>
-        </ViewContent>
-      </View>
-      <ManageDeviceGuide />
+      <GuideWrapper>
+        <GuidedContent>
```
thisconnect, line 65: I didn't know what else to name those container components, please let me know if you have a better suggestion.

frontends/web/src/components/layout/guide-wrapper.tsx
```diff
@@ -0,0 +1,38 @@
...
+  );
+};
+
+export const GuidedContent = ({ children }: TProps) => {
+  return (
+    <div className={styles.container}>
+      {children}
+    </div>
+  );
+};
```
thisconnect, line 38: there 2 div's are basically the same as it was before with https://github.com/digitalbitbox/bitbox-wallet-app/commit/484c42f59207f12f428f588935cd3628e603d83b#diff-21e6bfd8c5f3306cafbabf33252393a5fb8e9ddf3d23bd7b42a417d74d038e7dL64-L65

with the expection of the scrollcontainer, which is the reason that in the CSS there is an additional definition `overflow: auto`.

shonsirsha (approved):
tested LGTM

before:
Screenshot 2023-01-11 at 04 12 56

after:
Screenshot 2023-01-11 at 04 13 04

🚀

PR #1908, opened 2023-01-10
frontend/pocket: add address verification

Beerosagos:
A safe buy workflow requires a verification of the address that has been shared with the exchange. In this case pocket widget sends an email containing the address that has been shared by the user and requires a verification via `postMessage`. The address is displayed on the bb02 and should be ideally verified by the user opening the email con a second device (e.g. the phone) for better security.

### DIFF
--- a/backend/exchanges/exchanges.go
+++ b/backend/exchanges/exchanges.go
@@ -29,8 +29,13 @@ func (e ErrorCode) Error() string {
 	return string(e)
 }
 
-// ErrAddressNotFound is returned if an account is being added which already exists.
-const ErrAddressNotFound ErrorCode = "addressNotFound"
+const (
+	// ErrAddressNotFound is returned if an address provided for verification is not in the list of unused addresses.
+	ErrAddressNotFound ErrorCode = "addressNotFound"
+
+	// ErrUserAbort is returned if the user aborted the current operation.
+	ErrUserAbort ErrorCode = "userAbort"
+)
 
 // regionCodes is an array containing ISO 3166-1 alpha-2 code of all regions.
 // Source: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
--- a/backend/exchanges/exchanges.go
+++ b/backend/exchanges/exchanges.go
@@ -21,6 +21,17 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
 )
 
+// ErrorCode are errors that are represented by an error code. This helps the frontend to translate
+// error messages.
+type ErrorCode string
+
+func (e ErrorCode) Error() string {
+	return string(e)
+}
+
+// ErrAddressNotFound is returned if an account is being added which already exists.
+const ErrAddressNotFound ErrorCode = "addressNotFound"
+
 // regionCodes is an array containing ISO 3166-1 alpha-2 code of all regions.
 // Source: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
 var regionCodes = []string{
--- a/backend/exchanges/pocket.go
+++ b/backend/exchanges/pocket.go
@@ -22,6 +22,7 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/signing"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
 )
 
 const (
@@ -43,6 +44,19 @@ const (
 	// PocketName is the name of the exchange, it is unique among all the supported exchanges.
 	PocketName = "pocket"
 )
+
+// requestAddressScriptTypeMap maps from format codes specified by request-address js library
+// to our own script type codes. See https://www.npmjs.com/package/request-address
+var requestAddressScriptTypeMap = map[string]signing.ScriptType{
+	"p2pkh":  signing.ScriptTypeP2PKH,
+	"p2wpkh": signing.ScriptTypeP2WPKH,
+	"p2sh":   signing.ScriptTypeP2WPKHP2SH,
+	"p2tr":   signing.ScriptTypeP2TR,
+}
+
+// defaultScriptType is the default type used for address sharing and verification, if the
+// `format` param is not defined.
+var defaultScriptType = "p2wpkh"
 
 // PocketRegion represents informations collected by Pocket supported countries REST call.
 type PocketRegion struct {
@@ -113,13 +127,13 @@ func GetPocketSupportedRegions(httpClient *http.Client) (map[string]PocketRegion
 // 	`message` is the message that will be signed by the user with the private key linked to the address.
 //	`format` is the script type that should be used in the address derivation, as received by the widget
 //		(see https://github.com/pocketbitcoin/request-address#requestaddressv0messagescripttype).
-// 	`aoppBTCScriptTypeMap` is the map used in the AOPP flow to get the `ScriptType` object related to the `format` param.
+//		If format is empty, native segwit type is used as a fallback.
 //
 // Returned values:
 //	#1: is the first unused address corresponding to the account and the script type identified by the input values.
 //	#2: base64 encoding of the message signature, obtained using the private key linked to the address.
 //	#3: is an optional error that could be generated during the execution of the function.
-func PocketWidgetSignAddress(account accounts.Interface, message string, format string, aoppBTCScriptTypeMap map[string]signing.ScriptType) (string, string, error) {
+func PocketWidgetSignAddress(account accounts.Interface, message string, format string) (string, string, error) {
 
 	if !IsPocketSupported(account) {
 		err := fmt.Errorf("Coin not supported %s", account.Coin().Code())
@@ -128,7 +142,10 @@ func PocketWidgetSignAddress(account accounts.Interface, message string, format 
 
 	unused := account.GetUnusedReceiveAddresses()
 	// Use the format hint to get a compatible address
-	expectedScriptType, ok := aoppBTCScriptTypeMap[format]
+	if len(format) == 0 {
+		format = defaultScriptType
+	}
+	expectedScriptType, ok := requestAddressScriptTypeMap[format]
 	if !ok {
 		err := fmt.Errorf("Unknown format:  %s", format)
 		return "", "", err
@@ -150,4 +167,33 @@ func PocketWidgetSignAddress(account accounts.Interface, message string, format 
 	}
 
 	return addr.EncodeForHumans(), base64.StdEncoding.EncodeToString(sig), nil
+}
+
+// PocketWidgetVerifyAddress allows the user to verify an address for the Pocket Iframe workflow.
+// Input params:
+// 	`account` is the account from which the address is derived, and that will be linked to the Pocket order.
+// 	`address` is the address to be verified. It should be the same address previously returned by
+//		`PocketWidgetSignAddress`. Since this should be the first unused address, this function ranges
+//		among them to retrieve the ID needed for the verification.
+func PocketWidgetVerifyAddress(account accounts.Interface, address string) error {
+	if !IsPocketSupported(account) {
+		return fmt.Errorf("Coin not supported %s", account.Coin().Code())
+	}
+
+	addressLists := account.GetUnusedReceiveAddresses()
+	// iterate over the available script types to find the correct address
+	for _, list := range addressLists {
+		for _, addr := range list.Addresses {
+			if addr.EncodeForHumans() == address {
+				_, err := account.VerifyAddress(addr.ID())
+				if err != nil {
+					return err
+				}
+				return nil
+			}
+		}
+	}
+	// If this happens the address provided by pocket is not in the list of unused addresses.
+	// Reason could be that it has been used between the message signing and the verification.
+	return errp.WithStack(ErrAddressNotFound)
 }
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -1164,15 +1164,15 @@ func (handlers *Handlers) postPocketWidgetVerifyAddress(r *http.Request) (interf
 	}
 
 	type response struct {
-		Success   bool   `json:"success"`
-		Error     string `json:"error,omitempty"`
-		ErrorCode string `json:"errorCode,omitempty"`
+		Success      bool   `json:"success"`
+		ErrorMessage string `json:"errorMessage,omitempty"`
+		ErrorCode    string `json:"errorCode,omitempty"`
 	}
 
 	account, err := handlers.backend.GetAccountFromCode(request.AccountCode)
 	if err != nil {
 		handlers.log.Error(err)
-		return response{Success: false, Error: err.Error()}, nil
+		return response{Success: false, ErrorMessage: err.Error()}, nil
 	}
 
 	err = exchanges.PocketWidgetVerifyAddress(account, request.Address)
@@ -1181,7 +1181,7 @@ func (handlers *Handlers) postPocketWidgetVerifyAddress(r *http.Request) (interf
 		if errCode, ok := errp.Cause(err).(exchanges.ErrorCode); ok {
 			return response{Success: false, ErrorCode: string(errCode)}, nil
 		}
-		return response{Success: false, Error: err.Error()}, nil
+		return response{Success: false, ErrorMessage: err.Error()}, nil
 	}
 	return response{Success: true}, nil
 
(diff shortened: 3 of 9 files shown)


frontends/web/src/api/exchanges.ts
```diff
@@ -75,6 +75,15 @@ export const signAddress = (format: string, msg: string, accountCode: AccountCod
   return apiPost('exchange/pocket/sign-address', { format, msg, accountCode });
 };
 
+export type AddressVerificationResponse = {
+  status: 'ok' | 'error' | 'used-address';
+  error: string;
```
thisconnect, line 80: I am not 100% sure isn't error optional?
I believe this part https://github.com/digitalbitbox/bitbox-wallet-app/pull/1908/files#diff-b9e42f5f6cd5c1b3463424f0f375839cb238781d1c57495b03ae5e7af983e012R1169

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -96,6 +97,18 @@ export const Pocket = ({ code }: TProps) => {
...
       }
+      if (message.type === V0MessageType.VerifyAddress) {
+        setVerifying(true);
+        verifyAddress('p2wpkh', message.bitcoinAddress, code)
+          .then(response => {
+            setVerifying(false);
+            if (response.status === 'used-address') {
+              // This should not happen, unless the user receives a tx on the same address between the message signing
+              // and the address verification.
+              alertUser(t('buy.pocket.usedAddress.err1') + message.bitcoinAddress + t('buy.pocket.usedAddress.err2'));
```
thisconnect, line 108: instead of 2 different locize keys, there is a way to add parameter to 1 key, see:
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/82b3291732d0e5030de43f1217caeadb2d84f91b/frontends/web/src/locales/en/app.json#L34
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/82b3291732d0e5030de43f1217caeadb2d84f91b/frontends/web/src/routes/account/info/info.tsx#L77-L81

thisconnect (changes requested):

Beerosagos:
Thanks @thisconnect for the review! now should be better. 
I also added a second small commit to remove the alert dialog that was displayed when the user aborted the address signing (though it didn't worth making a dedicated PR for that).

PTAL :pray:

thisconnect (approved):
untested LGTM

Beerosagos:
Rebased after master got merged into pocket-staging. Lint failed but opened a separate PR to fix https://github.com/digitalbitbox/bitbox-wallet-app/pull/1919

Ping @benma :)

backend/exchanges/pocket.go
```diff
@@ -145,3 +145,47 @@ func PocketWidgetSignAddress(account accounts.Interface, message string, format
 
 	return addr.EncodeForHumans(), base64.StdEncoding.EncodeToString(sig), nil
 }
+
+// PocketWidgetVerifyAddress allows the user to verify an address for the Pocket Iframe workflow.
```
benma, line 149: please document the return value. it seems they are error codes? why not return them as `error` if that is the case?

frontends/web/src/locales/en/app.json
```diff
@@ -411,6 +411,11 @@
         "p1": "When you buy bitcoin via Pocket, you are using an external service. This service is out of scope of the BitBox02 Security Threat model and relies on the safety and security of the environment which the BitBoxApp software is running in. However we work together to improve security by using a two factor authentication mechanism to verify the address you are receiving to.",
         "title": "Security model"
       },
+      "usedAddress": "The address {{address}} has been already used, please start again.",
+      "verifyBitBox02": {
+        "dialog": "Please verify that the address you received via email matches the one displayed on your device. If possible, you should open the email on a second device for better security.",
+        "title": "Verify address on BitBox02"
```
benma, line 417: There is already `receive.verifyBitBox02` ith the same text, maybe use that instead?

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -95,14 +96,23 @@ export const Pocket = ({ code }: TProps) => {
...
-              alertUser('Message signing aborted by the user');
-            } else {
+            if (response.status !== 'abort') {
               sendAddress(response.address, response.signature);
             }
           });
       }
+      if (message.type === V0MessageType.VerifyAddress) {
+        setVerifying(true);
+        verifyAddress('p2wpkh', message.bitcoinAddress, code)
```
benma, line 106: why all this code to select the right address based on script type in the backend when it's hardcoded to p2wpkh?

backend/exchanges/pocket.go
```diff
@@ -145,3 +145,47 @@ func PocketWidgetSignAddress(account accounts.Interface, message string, format
...
+// Input params:
+// 	`account` is the account from which the address is derived, and that will be linked to the Pocket order.
+// 	`address` is the address to be verified. It should be the same address previously returned by
+//		`PocketWidgetSignAddress`. Since this should be the first unused address, this function ranges
+//		among them to retrieve the ID needed for the verification.
+//	`format` is the script type that should be used in the address derivation, as received by the widget
+//		(see https://github.com/pocketbitcoin/request-address#requestaddressv0messagescripttype).
+// 	`aoppBTCScriptTypeMap` is the map used in the AOPP flow to get the `ScriptType` object related to the `format` param.
+func PocketWidgetVerifyAddress(account accounts.Interface, address string, format string, aoppBTCScriptTypeMap map[string]signing.ScriptType) (string, error) {
```
benma, line 158: seems like the return values are error types, so I'd expect the return value to be only `error`.

You can encode the error code in the error itself.

For consistency, you can then forward it as `errorCode` to the frontend, as we do in most other places (in some places we also use the name `errCode`). If you need an additional err message, forward it as `errorMessage`  (also what we commonly use).

benma (changes requested):

backend/exchanges/pocket.go
Beerosagos: As suggested in the other comment, there was no need of return values, now it only returns error

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: Gosh! Forgot about that, now should be ok. I also defined native segwit as the default script type in case it is not defined in the request

backend/exchanges/pocket.go
Beerosagos: Right, thanks. Now should be better.

backend/exchanges/exchanges.go
```diff
@@ -21,6 +21,22 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
 )
 
+// ErrorCode are errors that are represented by an error code. This helps the frontend to translate
+// error messages.
+type ErrorCode string
```
Beerosagos, line 26: I preferred to avoid making a dedicated `error.go` file since we only have two error codes atm.

Beerosagos:
Hi! Thanks for the review. I updated the code, main changes:
- simplified error handling using error codes
- Introduced a dedicated `requestAddressScriptTypeMap` in `pocket.go` to handle the translation of the script types codes  between those used by pocket requests and our own

PTAL :pray:

Beerosagos:
> Hi! Thanks for the review. I updated the code, main changes:
> 
>     * simplified error handling using error codes
> 
>     * Introduced a dedicated `requestAddressScriptTypeMap` in `pocket.go` to handle the translation of the script types codes  between those used by pocket requests and our own
> 
> 
> PTAL pray

ping @benma

backend/exchanges/exchanges.go
```diff
@@ -21,6 +21,22 @@ import (
...
+
+func (e ErrorCode) Error() string {
+	return string(e)
+}
+
+const (
+	// ErrAddressNotFound is returned if an account is being added which already exists.
```
benma, line 33: the comment does not seem to match the name of the error. looking at its use, it seems the comment is wrong?

backend/exchanges/pocket.go
```diff
@@ -151,3 +168,48 @@ func PocketWidgetSignAddress(account accounts.Interface, message string, format
...
+	}
+
+	unused := account.GetUnusedReceiveAddresses()
+	// Use the format hint to get a compatible address
+	if len(format) == 0 {
+		format = defaultScriptType
+	}
+	expectedScriptType, ok := requestAddressScriptTypeMap[format]
+	if !ok {
+		err := fmt.Errorf("Unknown format:  %s", format)
```
benma, line 194: just return without a variable? here and in other places

frontends/web/src/api/exchanges.ts
```diff
@@ -63,18 +63,31 @@ export const getMoonpayBuyInfo = (code: string) => {
   };
 };
 
+export type ExchangeErrorCode = 'userAbort' | 'addressNotFound';
```
benma, line 66: `''` empty code missing.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -89,19 +93,35 @@ export const Pocket = ({ code }: TProps) => {
...
-              // TODO notify to the widget with a message?
-              alertUser('Message signing aborted by the user');
-            } else {
+            signing = false;
+            if (response.errorCode !== 'userAbort') {
               sendAddress(response.address, response.signature);
             }
-          });
+          })
+          .catch(() => signing = false);
```
benma, line 109: this is never executed in Qt. if you want it to you have to change `signAddress` to `reject()` upon error, which I think @thisconnect has started doing sometimes. more common in the codebase is to simply check for success inside of `then(...)`.

this and other things should be cleaned up at some point in our error handling - would be nice to unify it at all some point.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -37,12 +38,15 @@ export const Pocket = ({ code }: TProps) => {
...
   const [iframeLoaded, setIframeLoaded] = useState(false);
   const [agreedTerms, setAgreedTerms] = useState(false);
+  const [verifying, setVerifying] = useState(false);
 
   const iframeURL = useLoad(getPocketURL(code));
   const config = useLoad(getConfig);
 
   const ref = createRef<HTMLDivElement>();
   const iframeRef = createRef<HTMLIFrameElement>();
+  var addressType = '';
```
benma, line 48: this works, but using this var and passing it to the backend could be avoided if backend function simply iterated over all addresses of all script types - could be more robust too, as now it assumes a particular order of calls from the widget.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -89,19 +93,35 @@ export const Pocket = ({ code }: TProps) => {
...
+            setVerifying(false);
+            if (!response.success) {
+              if (response.errorCode === 'addressNotFound') {
+                // This should not happen, unless the user receives a tx on the same address between the message signing
+                // and the address verification.
+                alertUser(t('buy.pocket.usedAddress', { address:  message.bitcoinAddress }));
+              }
+            }
+          })
+          .catch(() => setVerifying(false));
```
benma, line 124: same comment as above - this is not executed in Qt.

benma (changes requested):

backend/exchanges/exchanges.go
Beerosagos: Sorry, mispasted line. Fixed now

frontends/web/src/api/exchanges.ts
Beerosagos: I made `errorCode` optional in `AddressSignResponse`

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: Didn't know that, thanks! I added it "just in case", now removed since it shouldn't be necessary anyway

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: Good point. Now the verification doesn't require the script type as an input

Beerosagos:
Thanks @benma, pushed an updated version. PTAL :pray:

frontends/web/src/routes/buy/pocket.tsx
thisconnect: we are using reject + `catch` in a few places, but at the moment I don't like it as the args received from catch can't be typed. I think "simply check for success inside of then(...)" is better.

frontends/web/src/api/exchanges.ts
benma: That doesn't reflect the backend accurately. `?: ...` means it can be undefined, but the backend never returns undefined (for defined struct fields). It returns the empty string `''` in this case. For pointers, `null` is returned (which is different from `undefined` :sweat_smile:).

backend/handlers/handlers.go
```diff
@@ -1155,6 +1154,39 @@ func (handlers *Handlers) getExchangePocketURL(r *http.Request) (interface{}, er
...
+		AccountCode string `json:"accountCode"`
+		Address     string `json:"address"`
+	}
+	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+		return nil, errp.WithStack(err)
+	}
+
+	type response struct {
+		Success   bool   `json:"success"`
+		Error     string `json:"error,omitempty"`
```
benma, line 1168: We should maybe just remove this: https://github.com/digitalbitbox/bitbox-wallet-app/blob/f9503f53f38f6351d6962477f9330e4ea91964ce/frontends/web/src/utils/request.ts#L50-L63 (or at least remove the `reject()` part of it).

But as it is now, I think `error` is caught and rejected only in webdev but not in Qt (see my earlier comment about the `catch` not being handled in Qt). But now, this error will not make it to the `.then(...)` clause in webdev mode because of the above.

Fyi, `error` is currently used to catch and alert errors that are returned by handler functions as in `return nil, errors.New(...))`, and to catch and alert panics, neither of which should not happen in production. The code that does that is here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/f9503f53f38f6351d6962477f9330e4ea91964ce/backend/handlers/handlers.go#L898-L917. 

So best to avoid `error` as a field name for now and use `errorMessage`, which we use commonly. Should have elaborated from the beginning in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1908/#discussion_r1072191238, sorry!

Beerosagos: Sorry, I knew this mechanism and I wanted to exploit it in my original PR, since I didn't know it wasn't working on Qt. But then I forgot about that when I read your review :D Anyway, I changed the field to `errorMessage` and added specific error handling (`alertUser` + error log) in the page.

frontends/web/src/api/exchanges.ts
Beerosagos: I am not sure about this. I think that the `omitempty` tag (see https://github.com/Beerosagos/bitbox-wallet-app/blob/address-verification/backend/handlers/handlers.go#L1168-L1169) causes the backend to avoid the error fields in the returned json. In my tests, in case of success I just get `{"success":true}`. Am I missing something?

Beerosagos:
Pushed a new update to fix error messages handling in the frontend. Not sure about the empty `ExchangeErrorCode` thing: @benma pls let me know if it is fine or if I should change it.
Thanks!

frontends/web/src/api/exchanges.ts
```diff
@@ -63,18 +63,31 @@ export const getMoonpayBuyInfo = (code: string) => {
...
+
 export type AddressSignResponse = {
-  status?: 'success' | 'abort';
+  success: boolean;
   signature: string;
   address: string;
-  error: string;
+  errorMessage?: string;
+  errorCode?: ExchangeErrorCode;
 }
```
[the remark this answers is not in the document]
Beerosagos: I think that `maybeBB02Err` is only used to map call results that go through bb02 API. It would be nice to refactor the generic endpoint communication using a standard Response struct with those three fields to be used (and extended when needed). But I guess this is a pretty broad change, that requires one or more dedicated PRs

benma (approved):

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -91,22 +92,26 @@ export const Pocket = ({ code }: TProps) => {
...
+            } else {
+              if (response.errorCode !== 'userAbort') {
+                alertUser(t('genericError'));
+                console.log('error: ' + response.errorMessage);
+              }
             }
           });
       }
-      if (message.type === V0MessageType.VerifyAddress) {
+      if (message.type === V0MessageType.VerifyAddress && !verifying) {
```
Beerosagos, line 114: Just noticed a small bug that allowed queuing several address verification. Fixed it before merging

PR #1913, opened 2023-01-12
frontend/pocket: add terms page to pocket flow

Beerosagos:
This adds a terms and conditions page similar to the one already present for moonpay. A new frontend configuration bool to hide permanently the terms screen is also added.


frontends/web/src/routes/buy/moonpay-terms.tsx
```diff
@@ -35,7 +35,7 @@ export const MoonpayTerms = ({ account, onAgreedTerms }: TProps) => {
   const name = getCryptoName(t('buy.info.crypto'), account);
 
   const handleSkipDisclaimer = (e: ChangeEvent<HTMLInputElement>) => {
-    setConfig({ frontend: { skipBuyDisclaimer: e.target.checked } });
+    setConfig({ frontend: { skipMoonpayDisclaimer: e.target.checked } });
```
Beerosagos, line 38: I wasn't really sure about leaving the old name or updating it. The update will cause the users to hide again the terms for moonpay, and to have an unused legacy configuration flag (which should not give any problem, though). Nonetheless, I felt it was better to have a coherent name.

PR #1914, opened 2023-01-12
frontend: type bitbox02 verify channel hash api 

thisconnect:
Ongoing effort to type all api calls, see also:
- backend/devices/bitbox02/handlers/handlers.go

on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1910

PR #1915, opened 2023-01-12
frontend: type bitbox02 insert sdcard api 

thisconnect:
Ongoing effort to type all api calls, see:
- backend/devices/bitbox02/handlers/handlers.go

Looks like this was wrongly displaying errorMessage instead
of response.message.

on top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1914

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2023 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


thisconnect:
but I am not sure how to trigger an error or when this would fail.

Beerosagos:
> but I am not sure how to trigger an error or when this would fail.

I think this can fail only with an error generated in the firmware and returned by the device query. The easiest way I see to test the error dialog is to generate a fake error and return it with `return maybeBB02Err(err, handlers.log), nil` in `postInsertSDCard`

Beerosagos (approved):
untested LGTM!

thisconnect:
merged in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1940

PR #1916, opened 2023-01-16
frontend: localise info content

shonsirsha:
To support multiple languages for "buy / exchange" info content, we'd need to localise them.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -310,6 +310,41 @@
       "creditCard": "Credit card",
       "fast": "Fast",
       "fee": "fee",
+      "infoContent": {
+        "moonpay": {
+          "fees": {
+            "bankTransfer": "Bank Transfer: {{fee}}%",
+            "creditDebitCard": "Credit/debit card: {{fee}}%",
+            "learnMore": "Learn more about Moonpay",
+            "title": "Fees"
+          },
+          "fullCurrenciesList": "See full list of currencies here",
+          "payment": {
+            "asteriskText": "* Not available for US residents",
+            "bankTransfer": "Bank transfer*",
+            "bankTransferDetails": {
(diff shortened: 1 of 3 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -313,7 +313,43 @@
...
+              "sepa": "SEPA and SEPA Instant (EUR transactions in SEPA countries only)",
+              "uk": "UK Faster Payments (GBP transactions in the UK only)",
+              "pix": "PIX (BR transactions in Brazil only)"
+            },
+            "asteriskText": "* Not available for US residents",
+            "learnMore": "See more details about payment methods"
+          },
+          "fees": {
+            "title": "Fees",
+            "creditDebitCard": "Credit/debit card: 4.9%",
```
Beerosagos, line 342: It could be nice to avoid the numbers and load them dynamically from the deals instead. wdyt? No strong opinion though

Beerosagos (approved):
LGTM with small nit. You'll have to run `make locize-fix` to fix the lints :)

frontends/web/src/locales/en/app.json
shonsirsha: Ah, sure would be possible. Where are we getting these values from?

Beerosagos: They are collected by `getExchangeDeals` https://github.com/digitalbitbox/bitbox-wallet-app/blob/staging-pocket/frontends/web/src/routes/buy/exchange.tsx#L68

shonsirsha: yup, just realised :)

shonsirsha: > It could be nice to avoid the numbers and load them dynamically from the deals instead. wdyt? No strong opinion though

and yes, that'd be great!

PR #1921, opened 2023-01-17
frontend/guide: update buy guide content

Beerosagos:
The guide content for the buy section is updated to cover the integration of Pocket.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -375,10 +375,12 @@
         "privacyPolicy": "MoonPay privacy policy",
         "protection": {
           "description": "The BitBoxApp does not collect any data when buying {{name}}, the incoming funds are treated like a regular transaction. MoonPay needs to collect some personal data to operate. Their Privacy Policy explains in detail how that data is handled.",
+          "descriptionGeneric": "The BitBoxApp does not collect any data when buying {{name}}, the incoming funds are treated like a regular transaction. However partner exchanges need to collect some information to operate. Please refer to their respective privacy policies to see in more detail how the data is handled.",
           "title": "Data protection"
         },
         "security": {
-          "description": "When you buy {{name}} via MoonPay, you are using an external service. This service is out of scope of the BitBox02 Security Threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
+          "description": "When you buy {{name}} via MoonPay, you are using an external service. This service is out of scope of the BitBox02 security threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
+          "descriptionGeneric": "When you buy {{name}} via a partner exchange, you are using an external service. This service is out of scope of the BitBox02 security threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
           "link": "Security threat model",
           "title": "Security model"
         },
--- a/frontends/web/src/routes/buy/guide.tsx
+++ b/frontends/web/src/routes/buy/guide.tsx
@@ -20,10 +20,24 @@ import { Guide } from '../../components/guide/guide';
 
 interface BuyGuideProps {
   name: string;
+  exchange?: 'pocket' | 'moonpay';
 }
 
-export default function BuyGuide({ name }: BuyGuideProps) {
+export default function BuyGuide({ name, exchange }: BuyGuideProps) {
   const { t } = useTranslation();
+
+  const pocketLink = {
+    text: t('buy.pocket.data.link'),
+    url: 'https://pocketbitcoin.com/policy/privacy',
+  };
+
+  const moonpayLink = {
+    text: t('buy.info.disclaimer.privacyPolicy'),
+    url: 'https://www.moonpay.com/privacy_policy',
+  };
+
+  const privacyLink = exchange === 'pocket' ? pocketLink : moonpayLink;
+
   return (
     <Guide>
       <Entry key="guide.buy.security" entry={{
@@ -31,15 +45,12 @@ export default function BuyGuide({ name }: BuyGuideProps) {
           text: t('buy.info.disclaimer.security.link'),
           url: 'https://shiftcrypto.ch/bitbox02/threat-model/',
         },
-        text: t('buy.info.disclaimer.security.description', { name }),
+        text: t('buy.info.disclaimer.security.descriptionGeneric', { name }),
         title: t('buy.info.disclaimer.security.title'),
       }} shown={true} />
       <Entry key="guide.buy.protection" entry={{
-        link: {
-          text: t('buy.info.disclaimer.privacyPolicy'),
-          url: 'https://www.moonpay.com/privacy_policy',
-        },
-        text: t('buy.info.disclaimer.protection.description', { name }),
+        link: exchange ? privacyLink : undefined,
+        text: t('buy.info.disclaimer.protection.descriptionGeneric', { name }),
         title: t('buy.info.disclaimer.protection.title'),
       }} />
       <Entry key="guide.buy.exchanges" entry={{
--- a/frontends/web/src/routes/buy/moonpay.tsx
+++ b/frontends/web/src/routes/buy/moonpay.tsx
@@ -111,7 +111,7 @@ export const Moonpay = ({ accounts, code }: TProps) => {
           )}
         </div>
       </div>
-      <Guide name={name} />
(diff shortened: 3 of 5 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -372,13 +373,12 @@
...
+          "description": "The BitBoxApp does not collect any data when buying {{name}}, the incoming funds are treated like a regular transaction. However MoonPay and Pocket need to collect some information to operate. Please refer to their respective privacy policies to see in more detail how the data is handled.",
           "title": "Data protection"
         },
         "security": {
-          "description": "When you buy {{name}} via MoonPay, you are using an external service. This service is out of scope of the BitBox02 Security Threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
+          "description": "When you buy {{name}} via MoonPay and/or Pocket, you are using an external service. This service is out of scope of the BitBox02 Security Threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
```
thisconnect, line 381: since you are here anyway, please lowercase "Security Threat model"

frontends/web/src/routes/buy/guide.tsx
```diff
@@ -36,9 +36,13 @@ export default function BuyGuide({ name }: BuyGuideProps) {
       }} shown={true} />
       <Entry key="guide.buy.protection" entry={{
         link: {
-          text: t('buy.info.disclaimer.privacyPolicy'),
+          text: t('buy.info.disclaimer.moonpayPrivacyPolicy'),
           url: 'https://www.moonpay.com/privacy_policy',
         },
+        link2: {
+          text: t('buy.pocket.data.link'),
```
thisconnect, line 43: I would expect a similar key like moonpayPrivacyPolicy, but I expect it is probably due to that this content is shared...

on the other hand iirc it is also shared for moonpay. Could you double check why this key is not `buy.info.disclaimer.pocketPrivacyPolicy`... or why the moonpay one isn't using 'buy.moonpay.data.link'  

just fyi, we shouldn't rename existing keys, so if possible it would be great to have consistent keys.

frontends/web/src/routes/buy/guide.tsx
```diff
@@ -36,9 +36,13 @@ export default function BuyGuide({ name }: BuyGuideProps) {
...
+          url: 'https://pocketbitcoin.com/policy/privacy',
```
thisconnect, line 44: nit: could add https://pocketbitcoin.com/de/policy/privacy and https://pocketbitcoin.com/it/policy/privacy (German is fine but Italian seems to not be translated yet).

moonpay is also hardcoded to EN, so feel free to keep as is.

frontends/web/src/routes/buy/moonpay-terms.tsx
```diff
@@ -94,7 +94,7 @@ export const MoonpayTerms = ({ account, onAgreedTerms }: TProps) => {
         <p>{t('buy.info.disclaimer.protection.description', { name })}</p>
         <p>
           <A className={style.link} href="https://www.moonpay.com/privacy_policy">
-            {t('buy.info.disclaimer.privacyPolicy')}
+            {t('buy.info.disclaimer.moonpayPrivacyPolicy')}
```
thisconnect, line 97: aha, related to my other comment.. since you rename anyway maybe make it consistent with the key of the pocket link?

frontends/web/src/locales/en/app.json
```diff
@@ -372,13 +373,12 @@
           },
           "title": "Payment methods and fees"
         },
-        "privacyPolicy": "MoonPay privacy policy",
         "protection": {
-          "description": "The BitBoxApp does not collect any data when buying {{name}}, the incoming funds are treated like a regular transaction. MoonPay needs to collect some personal data to operate. Their Privacy Policy explains in detail how that data is handled.",
+          "description": "The BitBoxApp does not collect any data when buying {{name}}, the incoming funds are treated like a regular transaction. However MoonPay and Pocket need to collect some information to operate. Please refer to their respective privacy policies to see in more detail how the data is handled.",
```
thisconnect, line 377: I think I am missing context, could you update the commit message with how this should work?

Is the following correct?

each exchange/broker terms has its own text, followed by a shared text (this one). which is also shared in the guide.

I am confused by that in the "own" text it's the exchange/broker name, but in the next paragraph it suddently mentioned other service that I didn't select. This would be less confusing if we just used generic term instead of the actual name.

Also what if we add 5 more exchanges broker one day, maybe the user doesn't even know about them if they are not available in different regions. Then the text lists them all?

ping @jadzeidan 

> However MoonPay, Pocket, Broker2, Exchange3, BrokerNotInMyRegion4 or P2PXchange need to collect..

frontends/web/src/components/guide/entry.tsx
```diff
@@ -57,6 +61,9 @@ export const Entry: FunctionComponent<TProps> = props => {
             {entry.link && (
               <p><A data-testid="link" href={entry.link.url}>{entry.link.text}</A></p>
             )}
+            {entry.link2 && (
+              <p><A data-testid="link" href={entry.link2.url}>{entry.link2.text}</A></p>
+            )}
```
thisconnect, line 66: what if there are 5 more exchanges/brokers?

nicer IMO would be if the exchange/broker selection view would be generic (no link). and once the user selected a service the correct one would show up.

thisconnect (changes requested):
was confused about shared text/links

frontends/web/src/locales/en/app.json
thisconnect: One more thing, updating existing text should be done in locize. but keep it here for now so we see what should change and can discuss the change.

once PR is accepted we can change locize.

Beerosagos: I changed "Security Threat model", but restored the rest of the text as it was in the beginning. Made a new `descriptionGeneric` with a generic text that avoid mentioning single exchanges, which is included in the guide.

frontends/web/src/routes/buy/guide.tsx
Beerosagos: Yes, the content is shared with the terms and conditions. Since we would need to refactor also the keys related to moonpay, but decided not to do it now, I didn't change it.

frontends/web/src/routes/buy/guide.tsx
Beerosagos: I think it is better to keep this PR simple and make a separated PR where we call localized policies for both moonpay and pocket. wdyt?

frontends/web/src/routes/buy/moonpay-terms.tsx
Beerosagos: Restored the previous name to avoid modifying keys

frontends/web/src/components/guide/entry.tsx
Beerosagos: Good point, now there is an optional parameter that allows linking the correct privacy policy, or no privacy policy at all if the exchange is not defined.

Beerosagos:
Thanks for the review! I pushed an update: 
- tried to avoid modifying locize keys or texts (exception for "Security Thread model" to lowercase, as you requested)
- added generic privacy/security text keys to be included in the guide, which do not specify any exchange name
- added an optional parameter in the guide, to handle privacy links
- Bonus: fixed a small HTML mistake in pocket terms. Is a bit OT, but hope it is fine

PTAL :pray:

frontends/web/src/locales/en/app.json
```diff
@@ -375,10 +375,12 @@
...
+          "descriptionGeneric": "The BitBoxApp does not collect any data when buying {{name}}, the incoming funds are treated like a regular transaction. However partner exchanges need to collect some information to operate. Please refer to their respective privacy policies to see in more detail how the data is handled.",
           "title": "Data protection"
         },
         "security": {
-          "description": "When you buy {{name}} via MoonPay, you are using an external service. This service is out of scope of the BitBox02 Security Threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
+          "description": "When you buy {{name}} via MoonPay, you are using an external service. This service is out of scope of the BitBox02 security threat model and relies on the safety and security of the environment which the BitBoxApp software is running in.",
```
thisconnect, line 382: thanks 👍 
I also lowercased "security thread" in locize

thisconnect (approved):
👍  LGTM tested

PR #1922, opened 2023-01-18
frontend: added extra margin for dialog modal

shonsirsha:
to centre the dialog's modal with respect to content when sidebar is opened, we needed to add an extra margin.

Extra margin's value is the width of the sidebar when it's opened.


thisconnect (approved):
LGTM

PR #1925, opened 2023-01-19
Frontend type bitbox02 restorebackup api

thisconnect:
Ongoing effort to type all api calls, see postBackupsRestore in:
- backend/devices/bitbox02/handlers/handlers.go

This was the last untyped api call in backups.tsx.

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2023 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -138,4 +138,11 @@ type TChannelHash = {
 
 export const getChannelHash = (deviceID: string): Promise<TChannelHash> => {
   return apiGet(`devices/bitbox02/${deviceID}/channel-hash`);
+};
+
+export const verifyChannelHash = (
+  deviceID: string,
+  ok: boolean,
+): Promise<void> => {
+  return apiPost(`devices/bitbox02/${deviceID}/channel-hash-verify`, ok);
 };
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -116,6 +116,16 @@ export const checkBackup = (
     });
 };
 
+export const createBackup = (deviceID: string): Promise<void> => {
+  return apiPost(`devices/bitbox02/${deviceID}/backups/create`)
+    .then((response: FailResponse | SuccessResponse) => {
+      if (!response.success) {
+        return Promise.reject(response);
+      }
+      return Promise.resolve();
+    });
+};
+
 export const upgradeDeviceFirmware = (deviceID: string): Promise<void> => {
   return apiPost(`devices/bitbox02/${deviceID}/upgrade-firmware`);
 };
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -130,3 +130,12 @@ export type TStatus = 'connected'
 export const getStatus = (deviceID: string): Promise<TStatus> => {
   return apiGet(`devices/bitbox02/${deviceID}/status`);
 };
+
+type TChannelHash = {
+  hash: string;
+  deviceVerified: boolean;
+};
+
+export const getChannelHash = (deviceID: string): Promise<TChannelHash> => {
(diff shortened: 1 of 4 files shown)


frontends/web/src/api/bitbox02.ts
```diff
@@ -126,6 +126,13 @@ export const createBackup = (deviceID: string): Promise<void> => {
     });
 };
 
+export const restoreBackup = (
+  deviceID: string,
+  selectedBackup: string,
+): Promise<FailResponse | SuccessResponse> => {
```
thisconnect, line 132: There seem to be another error that doesn't return success: false, but I am not sure

https://github.com/digitalbitbox/bitbox-wallet-app/blob/82b3291732d0e5030de43f1217caeadb2d84f91b/backend/devices/bitbox02/handlers/handlers.go#L206-L215

frontends/web/src/routes/device/bitbox02/backups.tsx
```diff
@@ -72,18 +72,16 @@ class Backups extends Component<Props, State> {
...
-        restoring: false,
-        errorText: success ? '' : this.props.t('backup.restore.error.general'),
+    restoreBackup(this.props.deviceID, this.state.selectedBackup)
+      .then(({ success }) => {
+        this.setState({
+          restoring: false,
+          errorText: success ? '' : this.props.t('backup.restore.error.general'),
+        });
+        if (this.props.backupOnAfterRestore) {
+          this.props.backupOnAfterRestore(success);
```
thisconnect, line 82: this works even if success is `undefined` instead of false, but technically the type is wrong as it expects boolean.

https://github.com/digitalbitbox/bitbox-wallet-app/blob/82b3291732d0e5030de43f1217caeadb2d84f91b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx#L315-L320

thisconnect:
my last PR in this series PTAL :)

frontends/web/src/api/bitbox02.ts
Beerosagos: Yes, you're right. If e.g. the backup ID is missing it returns `null`.

Beerosagos (commented):
Code LGTM! Is it time to test the whole thing?

frontends/web/src/api/bitbox02.ts
thisconnect: thanks I'll update the PR and rebase the others.

Beerosagos: Hey, I tested it because I wanted to be sure. Apparently, the error is translated, so that if the backup id is missing the frontend result is actually not `null`, but a struct with a error field: like `{"error":"EOF"}`.
Sorry for the mistake :pray:

thisconnect: I noticed that there is a 2nd translated error https://github.com/digitalbitbox/bitbox-wallet-app/blob/bd06af77e63dd84293dc9a3693164551cf0db660/frontends/web/src/locales/en/app.json#L122-L125

but e200 is not used for BB02 it only uses generic backup restore error

https://github.com/digitalbitbox/bitbox-wallet-app/blob/bd06af77e63dd84293dc9a3693164551cf0db660/frontends/web/src/routes/device/bitbox02/backups.tsx#L80  

in BB01 component translates it correctly:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/bd06af77e63dd84293dc9a3693164551cf0db660/frontends/web/src/routes/device/bitbox01/restore.tsx#L101

I think it makes sense to clean this up so that the api error alway return FailResponse with "success", "code" and "message" fields, (what maybeError does).

Beerosagos: Opened a separated PR to fix the endpoint: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1933

frontends/web/src/routes/device/bitbox02/backups.tsx
thisconnect: will be fixed by https://github.com/digitalbitbox/bitbox-wallet-app/pull/1933

thisconnect:
yes time to test :) here is what I was planing to test (but now I found a regression, will open another PR):

- test insert-sdcard check during setup->restore-backup and setup->create-wallet
- test create backup in manage-device->create-backup and setup->create-wallet
- test setup->restore-backup
- test restore-from-mnemonic
- delete pairing (rm appfolder.dev/bitbox02/bitbox02.json), restart app and pair again
- setup->create-wallet set-password

thisconnect:
merged in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1940

PR #1926, opened 2023-01-19
frontend: fix iframe scroll content

thisconnect:
The iframe could not be scrolled all the way to the end, about exactly the height of the headers. Turns out JavaScript read the wrong height, as it included the height of the header.

Fixed the CSS height of the iframe container, so that the script reads the correct height.

Also removed unused CSS and fixed duplicated styling in MoonPay component. For some reason MoonPay was using styles from terms instead of iframe.module.css. Now both Pocket and MoonPay should use the same CSS.


Beerosagos (approved):
tACK on webdev!

PR #1927, opened 2023-01-23
backend: remove one default btc testnet dev server

benma:
Same as btc/ltc/tltc, one dev node is sufficient for testing. This will reduce maintanence and load on the backend.

Regtest can be used to test failover behavior etc. when two servers are needed for testing.


Beerosagos (approved):
tACK :+1: 
We could also consider removing tbtc2 and tltc2 from `NewDefaultAppConfig`, wdyt? 
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/config/config.go#L177

benma:
>  We could also consider removing tbtc2 and tltc2 from NewDefaultAppConfig, wdyt?

These are production testnet servers, not devservers, so they should stay.

Beerosagos:
> These are production testnet servers, not devservers, so they should stay.

Aha! Missed that :)  Thanks for clarification

PR #1929, opened 2023-01-23
Draft: route to device-settings directly and reduce layout shift

thisconnect:
The whole app window flickered when the user navigated to the
device settings, including the sidebar. Reason: device settings
route first load: deviceswitch that checks if device is bb01, bb02
or bootloader, for bb02 next the bitbox02.tsx is loaded, that
also handles unlock, pairing, create and restore, this component
does many api requests and is most likely the reason for the
flicker as it takes some time and will eventually load the bb02
settings component.

Involving deviceswtich/bitbox02 before loading device settings
seems to be the cause for completely blank view if for example
btc nodes are misconfigured, probably due to accounts not being
initialized correctly.

This change adds a new settingsswitch component that can load the
bb01 or bb02 settings component directly without going trough
deviceswitch.

Drawback: some shared components such as manage-backup or sdcard
check have a back button, this should not leave the setup if the
component is used in setup. therefore keeping the back buttons
pointing to deviceswitch.

also further reduced flickering content with another commit:

Added appropriate skeleton component as placeholder that the layout
does not shift while loading the view.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/settings.tsx
+++ b/frontends/web/src/routes/device/bitbox02/settings.tsx
@@ -32,6 +32,7 @@ import { alertUser } from '../../../components/alert/Alert';
 import { ManageDeviceGuide } from './settings-guide';
 import { View, ViewContent } from '../../../components/view/view';
 import { Column, Grid, GuidedContent, GuideWrapper, Header, Main } from '../../../components/layout';
+import { Skeleton } from '../../../components/skeleton/skeleton';
 
 type TProps = {
     deviceID: string;
@@ -56,9 +57,6 @@ export const Settings = ({ deviceID }: TProps) => {
     route(`/passphrase/${deviceID}`);
   };
 
-  if (deviceInfo === undefined) {
-    return null;
-  }
   return (
     <Main>
       <GuideWrapper>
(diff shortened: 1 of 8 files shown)


frontends/web/src/routes/device/bitbox02/settings.tsx
```diff
@@ -101,24 +101,26 @@ export const Settings = ({ deviceID }: TProps) => {
...
+                    <SettingsButton onClick={routeToPassphrase}>
+                      { deviceInfo.mnemonicPassphraseEnabled
+                        ? t('passphrase.disable')
+                        : t('passphrase.enable')}
+                    </SettingsButton>
+                  ) : <Skeleton fontSize="var(--item-height)" /> }
                   { versionInfo && versionInfo.canGotoStartupSettings ? (
                     <GotoStartupSettings apiPrefix={apiPrefix} />
-                  ) : null }
+                  ) : <Skeleton fontSize="var(--item-height)" /> }
```
thisconnect, line 123: will add the `<Skeleton>` in a separate PR

thisconnect:
not sure if this is the right approact. need to check if there are important checks in bitbox02.tsx that would be circumvented with this change.

thisconnect:
closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1941 and https://github.com/digitalbitbox/bitbox-wallet-app/pull/1944

PR #1930, opened 2023-01-23
backend/btc: use minrelayfee if the fee estimation is smaller

benma:
In Litecoin, the default `minrelaytxfee` is 0.00001 LTC/kB. However, the fee estimation sometimes returns fee rates smaller than that:

```
estimatesmartfee 2
{
  "feerate": 0.00000999,
  "blocks": 2
}
```

If the estimated fee is used for a tx, it can't be relayed and sending a tx results in an error.

We simply use the minrelayfee if it is larger than the estimated fee rate.


Beerosagos (approved):
Couldn't generate a tx with a low enough fee to test it on testnet, but code LGTM

PR #1932, opened 2023-01-25
frontend: created language-aware download link component 

shonsirsha:
To improve our UX, it would be better to have a "language-aware" download link. E.g German language app will use the german download page.

In the case where localised download page isn't available, will be fallbacking to the "international" / "default" downlaod page.

### DIFF
--- a/frontends/web/src/components/appdownloadlink/appdownloadlink.tsx
+++ b/frontends/web/src/components/appdownloadlink/appdownloadlink.tsx
@@ -1,0 +1,38 @@
+/**
+ * Copyright 2023 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useTranslation } from 'react-i18next';
+import i18n from '../../i18n/i18n';
+import { URL_CONSTANTS } from '../../utils/url_constants';
+import A from '../anchor/anchor';
+
+export const downloadLinkByLanguage = () => {
+  const language = i18n.language;
+  switch (language) {
+  case 'de':
+    return URL_CONSTANTS.DOWNLOAD_LINK_DE;
+  default:
+    return URL_CONSTANTS.DOWNLOAD_LINK_GLOBAL;
+  }
+};
+
+const AppDownloadLink = ({ ...props }) => {
+  const { t } = useTranslation();
+
+  return <A href={downloadLinkByLanguage()} {...props}>{t('button.download')}</A>;
+};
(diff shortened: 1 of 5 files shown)


frontends/web/src/utils/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const CONSTANTS = {
+  DOWNLOAD_LINK_GLOBAL: 'https://shiftcrypto.ch/download/?source=bitboxapp',
+  DOWNLOAD_LINK_DE: 'https://shiftcrypto.ch/de/download/?source=bitboxapp'
+};
```
shonsirsha, line 20: for future improvements: could make another small PR to add other hardcoded links here so our code could be a bit neater. 🙂

thisconnect: My only nit: if the file is called `constants.ts` I would expect that various different app related constants are in there... but I am not sure what else to call it 🤔

thisconnect (approved):
untested LGTM with nit

frontends/web/src/utils/constants.ts
shonsirsha: yeah. 

We could pick one of the below:
1. rename to `URL_CONSTANTS.ts` and this will include only URLs.
2. keep the file name but change the structure of the object slightly into 
```
export const CONSTANTS = {
  URLS: {
      DOWNLOAD_LINK_GLOBAL: 'https://shiftcrypto.ch/download/?source=bitboxapp',
      DOWNLOAD_LINK_DE: 'https://shiftcrypto.ch/de/download/?source=bitboxapp'
  }
};
``` 

Regardless of 1 or 2, we'll have to make a PR to move the constants that are being used all over the app into the constant file(s).

I think the second one will reduce the need to create many different "constant" files in the future.

thisconnect: not sure, @Beerosagos wdyt?

Beerosagos: > not sure, @Beerosagos wdyt?

I don't have a strong opinion about `CONSTANTS` structure. Slight preference for option 2, but it depends on which other constants will be saved there. It would be nice to pass a language param to the webpage, which could handle an English fallback, so that we don't need any future modification to add new languages (but this is a bit OT)

thisconnect: maybe move the file to `frontends/web/src` keep name `constants.ts` and export `DOWNLOAD_URLS`, then I think it makes sense wdyt ?

thisconnect: (no strong opinion on what name to export)

shonsirsha: oops, was merged before this. New PR will address this. 

> maybe move the file to frontends/web/src keep name constants.ts and export DOWNLOAD_URLS, then I think it makes sense wdyt ?

Sure, we can then just make them not nested e.g

```
export DOWNLOAD_URLS = {...}
export SOME_OTHER_URLS = {...}
```

or we can nest it e.g `URLS:{DOWNLOAD_URLS: {...}, SOME_OTHER_URLS: {...}}`

thisconnect: yes something like that, lets clean it up next time we touch it.

PR #1935, opened 2023-01-30
frontend: added prop verticallyCentered for View component

shonsirsha:
To fulfil our UI requirements, View needs to be able to centre its content vertically when NOT on full screen (`!fullscreen`).

### DIFF
--- a/frontends/web/src/components/view/view.tsx
+++ b/frontends/web/src/components/view/view.tsx
@@ -29,6 +29,7 @@ type TViewProps = {
     minHeight?: string;
     onClose?: () => void;
     textCenter?: boolean;
+    verticallyCentered?: boolean;
     width?: string;
     withBottomBar?: boolean;
 };
@@ -41,6 +42,7 @@ type TViewProps = {
  * @param minHeight optional minimum height, useful for keeping content area same size through multiple views
  * @param onClose if a callback is provided it will render a close button that triggers the callback
  * @param textCenter centers all text content in the view
+ * @param verticallyCentered centers all text content in the view, has no effect in dialog mode
  * @param width can be used to overwrite the default width of the inner area
  * @param withBottomBar enables a footer with some logo and language switch
  */
@@ -52,11 +54,14 @@ export const View = ({
   minHeight,
   onClose,
   textCenter,
+  verticallyCentered = false,
   width,
   withBottomBar,
 }: TViewProps) => {
   const containerClasses = `${
     style[fullscreen ? 'fullscreen' : 'fill']
+  } ${
+    verticallyCentered ? style.verticallyCentered : ''
   } ${
     dialog ? style.dialog : ''
   }`;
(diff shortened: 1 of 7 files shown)


frontends/web/src/components/view/view.tsx
```diff
@@ -41,6 +42,7 @@ type TViewProps = {
  * @param minHeight optional minimum height, useful for keeping content area same size through multiple views
  * @param onClose if a callback is provided it will render a close button that triggers the callback
  * @param textCenter centers all text content in the view
+ * @param verticallyCentered centers all text content in the view
```
thisconnect, line 45: verticallyCentered make sense

the docstring text implies it should also work also for fullscreen, which would actually be nice.

I made a POC with a few other changes: https://github.com/shonsirsha/bitbox-wallet-app/pull/1  wdyt?

shonsirsha: > the docstring text implies it should also work also for fullscreen, which would actually be nice.

I don't think I can see any difference when `fullScreen` and `verticallyCentered` are both `true` as opposed just having `fullScreen` to be true (setting fullScreen automatically vertically centres it..?)

thisconnect:
afaik this doesn't create a merge conflict (I tried merging into staigng-pocket and then merged staging-pocket into master).

@shonsirsha should I do final testing then we merge this, or do you want to merge this into staging-pocket?

shonsirsha:
@thisconnect No hard preference, but I think it's fine to merge to master after you're done with the final testing..

thisconnect (approved):
tested LGTM

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -455,13 +456,14 @@ class BitBox02 extends Component<Props, State> {
...
+            verticallyCentered
             withBottomBar
-            width="600px">
+            width="690px">
             <ViewHeader title={t('button.unlock')}>
               <p>{t('bitbox02Wizard.stepConnected.unlock')}</p>
             </ViewHeader>
             <ViewContent fullWidth>
-              {attestationResult === false ? (
+              {attestationResult === true ? (
```
Beerosagos, line 466: Was this put into the commit by mistake? Looks like a regression to me cc @shonsirsha @thisconnect

shonsirsha: Yes by looking at the code this isn't right. Or at least this PR shouldn't concern anything about `attestationResult`. I'll make a PR to revert this bit. good catch!

shonsirsha: Possibly it was added when [this](https://github.com/shonsirsha/bitbox-wallet-app/pull/1) was created and this attestationResult part was only meant for testing during development. Feel free to clarify this though @thisconnect :)

PR #1936, opened 2023-01-31
frontend: fix regression on Dialog and created dialog-legacy

shonsirsha:
During the development of the new Dialog (improving dialog for mobile), we introduced a regression that affected several places such as the:

1. Factory reset device
2. send-QR

This PR fixes it.

We also will use the old dialog for all of the bb01 components and the Confirm component (which uses dialog internally).

This PR also fixes some other random bugs that are not exactly in the Dialog component but still related to it, such as:
1. WaitDialog style on mobile
2. Closing send-QR with ESC only toggles the dialog, not actually closing it.

[Preview] Factory Reset Device:

https://user-images.githubusercontent.com/28676406/215662356-0cf40823-430f-4994-9c08-a107622619a1.mov

(the WaitDialog isn't immediately closing anymore).

### DIFF
--- a/frontends/web/src/components/wait-dialog/wait-dialog.tsx
+++ b/frontends/web/src/components/wait-dialog/wait-dialog.tsx
@@ -139,7 +139,7 @@ class WaitDialog extends Component<Props, State> {
         className={style.overlay}
         ref={this.overlay}
         style={{ zIndex: 10001 }}>
-        <div className={style.modal} ref={this.modal}>
+        <div className={[style.modal, style.open].join(' ')} ref={this.modal}>
           {
             title && (
               <div className={style.header}>
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -225,14 +225,8 @@ class Send extends Component<Props, State> {
   };
 
   private handleKeyDown = (e: KeyboardEvent) => {
-    if (e.keyCode === 27) {
-      if (this.state.activeScanQR) {
-        this.toggleScanQR();
-        return;
-      }
-      if (!this.state.activeCoinControl) {
-        route(`/account/${this.props.code}`);
-      }
+    if (e.keyCode === 27 && !this.state.activeCoinControl && !this.state.activeScanQR) {
+      route(`/account/${this.props.code}`);
     }
   };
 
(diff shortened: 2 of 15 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -225,14 +225,8 @@ class Send extends Component<Props, State> {
...
-    if (e.keyCode === 27) {
-      if (this.state.activeScanQR) {
-        this.toggleScanQR();
-        return;
-      }
-      if (!this.state.activeCoinControl) {
-        route(`/account/${this.props.code}`);
-      }
+    if (e.keyCode === 27 && !this.state.activeCoinControl && !this.state.activeScanQR) {
+      route(`/account/${this.props.code}`);
```
shonsirsha, line 229: This has been existing "forever" (not regression). Did it in this PR anyway since it's small and still related to Dialog. :)

frontends/web/src/components/wait-dialog/wait-dialog.tsx
```diff
@@ -139,7 +139,7 @@ class WaitDialog extends Component<Props, State> {
         className={style.overlay}
         ref={this.overlay}
         style={{ zIndex: 10001 }}>
-        <div className={style.modal} ref={this.modal}>
+        <div className={[style.modal, style.open].join(' ')} ref={this.modal}>
```
shonsirsha, line 142: Not the regression we talked about - we didn't notice this before (or at least I didn't). 

Since wait-dialog uses CSS of dialog.css, the wait dialog won't visually show on mobile (as it has `transform: translateY(100vh);`). We need to give `transfrom: translateY(0);` so it shows up. To do it we need to add `.open` class as well. Here's a preview of it:

[Before]:

https://user-images.githubusercontent.com/28676406/215666892-812d7533-15cf-4d2d-82fa-5437ceb68023.mp4


[After]:

https://user-images.githubusercontent.com/28676406/215666922-9256684b-eb2c-4002-844a-66152d9c1217.mov



Note: In the `Dialog`, there's a minor delay while adding the `.open` so there's a visible animation. To have the same effect here in `WaitDialog`, I believe we need this to be controlled by its parent (just like `Dialog`) - but this is a bigger topic. So prob. better to do it in a different PR if we really need it.

thisconnect (approved):
tested LGTM

PR #1937, opened 2023-01-31
frontend/exchange: fix locale detection

Beerosagos:
A bug in the locale detection caused the exchange selection flow to break if an empty native locale values was retrieved. This fixes the bug exploiting `i18n.language` instead of detecting (again) the language in the page.

### DIFF
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -15,6 +15,7 @@
  */
 import 'flag-icons';
 import React, { useState, useEffect } from 'react';
+import i18n from '../../i18n/i18n';
 import { useTranslation } from 'react-i18next';
 import { Button, Select, ButtonLink } from '../../components/forms';
 import * as exchangesAPI from '../../api/exchanges';
@@ -24,8 +25,7 @@ import Guide from './guide';
 import { findAccount, getCryptoName } from '../account/utils';
 import { route } from '../../utils/route';
 import { useLoad } from '../../hooks/api';
-import { languageFromConfig } from '../../i18n/config';
-import { localeMainLanguage, getRegionNameFromLocale } from '../../i18n/utils';
+import { getRegionNameFromLocale } from '../../i18n/utils';
 import { findLowestFee, findBestDeal, getFormattedName } from './utils';
 import { ExchangeSelectionRadio } from './components/exchangeselectionradio';
 import { Spinner } from '../../components/spinner/Spinner';
@@ -35,7 +35,6 @@ import { InfoButton } from '../../components/infobutton/infobutton';
 import { InfoContent } from './components/infocontent';
 import { Globe } from '../../components/icon';
 import style from './exchange.module.css';
-import { getNativeLocale } from '../../api/nativelocale';
 import { setConfig } from '../../utils/config';
 import { getConfig } from '../../api/backend';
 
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -79,34 +76,28 @@ export const Exchange = ({ code, accounts }: TProps) => {
...
     }
 
-    const userRegion = getRegionNameFromLocale(nativeLocale);
+    const userRegion = getRegionNameFromLocale(i18n.language);
     //Region is available in the list
     const regionAvailable = !!(regionList.regions.find(region => region.code === userRegion));
     //Pre-selecting the region
     setSelectedRegion(regionAvailable ? userRegion : '');
-  }, [regionList, locale, config, nativeLocale]);
+  }, [regionList, config]);
```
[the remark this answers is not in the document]
Beerosagos: It actually complains if I add it: 
```
src/routes/buy/exchange.tsx
  Line 100:6:  React Hook useEffect has an unnecessary dependency: 'i18n.language'. Either exclude it or remove the dependency array. Outer scope values like 'i18n.language' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
```

thisconnect (approved):
tested LGTM

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -79,34 +76,28 @@ export const Exchange = ({ code, accounts }: TProps) => {
...
+    regions.sort((a, b) => a.text.localeCompare(b.text, i18n.language));
     setRegions(regions);
 
     if (config.frontend.selectedExchangeRegion) {
       setSelectedRegion(config.frontend.selectedExchangeRegion);
       return;
     }
 
-    const userRegion = getRegionNameFromLocale(nativeLocale);
+    const userRegion = getRegionNameFromLocale(i18n.language);
```
shonsirsha, line 95: Hey there! Sorry, was unaware of this PR before it gets merged. Just a question, if I'm not mistaken `i18n.language` is the language of the app (whatever the user choses), right?

if so, when getting `userRegion, won't it be misleading if we detect it solely from their language? E.g costa rica user might have "Spain" as pre-selected country due to "Spanish" being selected as their language..? Or maybe Im missing something here? :)

> A bug in the locale detection caused the exchange selection flow to break if an empty native locale values was retrieved.

I "catch" a condition when nativeLocale isn't in the list (if nativeLocale is invalid but NOT empty), it'll default to  empty string: `setSelectedRegion(regionAvailable ? userRegion : '');`, which in the UI it'll select the "select region".


Since im unaware that native locale could be empty, this wasn't "caught". So, would it be possible to "catch" this and do similar thing like the above? Im not sure what nativeLocale returned that you said it's empty (empty string maybe?). If so, could we gracefully handle it maybe like so: `const userRegion = getRegionNameFromLocale(nativeLocale || '');` and remove `nativeLocale` check at the top of `useEffect`?

_Note: I haven't experienced an empty native-locale before, so im also a bit curious on what were the steps to reproduce it.._

thisconnect: I see what you mean, maybe we should just use i18n.language or config.backend.userLang as a fallback then.

> if so, when getting `userRegion, won't it be misleading if we detect it solely from their language? E.g costa rica user might have "Spain" as pre-selected country due to "Spanish" being selected as their language..? Or maybe Im missing something here? :)

Just assume navite-locale can return anything, also empty string or weird unknown code.

> Note: I haven't experienced an empty native-locale before, so im also a bit curious on what were the steps to reproduce it..

@shonsirsha could you open a new PR and add the logic of guessing the country again, but assume it can be any string.

shonsirsha: > Just assume navite-locale can return anything, also empty string or weird unknown code.

> @shonsirsha could you open a new PR and add the logic of guessing the country again, but assume it can be any string.

I believe it was already doing this, but we just didn't anticipate that native-locale could be empty, so I believe it crashed because of the check on top of useEffect (`!nativeLocale`). I'll create a PR that will accommodate this.

PR #1938, opened 2023-02-01
frontend: use native-locale to pre-select region dropdown in exchange page.

shonsirsha:
We have preivously used native-locale but we didn't take into account that native-locale may sometimes return nothing (empty).

Another PR then was created to fix this, using `i18n.language` - which is not exactly correct.

Thus, in this PR we used native-locale back. But, we are now handling empty native-locale gracefully so it won't crash the page whenever native-locale returns empty/nothing.

### DIFF
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -34,9 +34,10 @@ import { Dialog } from '../../components/dialog/dialog';
 import { InfoButton } from '../../components/infobutton/infobutton';
 import { InfoContent } from './components/infocontent';
 import { Globe } from '../../components/icon';
-import style from './exchange.module.css';
+import { getNativeLocale } from '../../api/nativelocale';
 import { setConfig } from '../../utils/config';
 import { getConfig } from '../../api/backend';
+import style from './exchange.module.css';
 
 type TProps = {
     code: string;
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -201,7 +203,7 @@ export const Exchange = ({ code, accounts }: TProps) => {
                       },
                       ...regions]
                       }
-                      value={selectedRegion || ''}
+                      value={selectedRegion}
```
shonsirsha, line 206: Not actually doing anything here :) just a cleanup. Since selectedRegion is already always a string, that `|| ''` is useless.

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -92,12 +94,12 @@ export const Exchange = ({ code, accounts }: TProps) => {
...
     }
 
-    const userRegion = getRegionNameFromLocale(i18n.language);
+    const userRegion = getRegionNameFromLocale(nativeLocale || '');
     //Region is available in the list
     const regionAvailable = !!(regionList.regions.find(region => region.code === userRegion));
     //Pre-selecting the region
     setSelectedRegion(regionAvailable ? userRegion : '');
-  }, [regionList, config]);
+  }, [regionList, config, nativeLocale]);
```
shonsirsha, line 102: Bit of explanation:

`userRegion` will return empty string if:
1. nativeLocale is empty string
2. nativeLocale is undefined
3. nativeLocale is invalid string / locale e.g `asd` or `en-XX`.

When `userRegion` is empty, `regionAvailable` will be `false`. When it's `false`, `selectedRegion` will be empty string -> which means it'll select the placeholder "Select Region" selection.

How did it select that placeholder value when it's empty string? It's due to the code below

```
  <Select
    className={style.extraLeftSpace}
    options={[{
      text: t('buy.exchange.selectRegion'),
      disabled: true,
      value: '',
    },
    ...regions]
    }
    value={selectedRegion}
    onChange={handleChangeRegion}
    id="exchangeRegions"
  />
```

Beerosagos (approved):
tested LGTM! Thanks

PR #1940, opened 2023-02-01
Frontend type bitbox02 api

thisconnect:
I rebased https://github.com/digitalbitbox/bitbox-wallet-app/pull/1925

and tested as described here https://github.com/digitalbitbox/bitbox-wallet-app/pull/1925#issuecomment-1405058713

looking at the changes they are mostly just typescript related, I will highlight the place where it is a bit different in JavaScript, due to TS complaining.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/backups.tsx
+++ b/frontends/web/src/routes/device/bitbox02/backups.tsx
@@ -15,10 +15,10 @@
  */
 
 import { Component } from 'react';
+import { restoreBackup } from '../../../api/bitbox02';
 import Toast from '../../../components/toast/Toast';
 import { subscribe } from '../../../decorators/subscribe';
 import { translate, TranslateProps } from '../../../decorators/translate';
-import { apiPost } from '../../../utils/request';
 import { Backup, BackupsListItem } from '../components/backup';
 import backupStyle from '../components/backups.module.css';
 import { Button } from '../../../components/forms';
@@ -72,18 +72,16 @@ class Backups extends Component<Props, State> {
     if (this.props.backupOnBeforeRestore) {
       this.props.backupOnBeforeRestore(backup);
     }
-    apiPost(
-      'devices/bitbox02/' + this.props.deviceID + '/backups/restore',
-      this.state.selectedBackup).then(({ success }) => {
-      this.setState({
-        restoring: false,
-        errorText: success ? '' : this.props.t('backup.restore.error.general'),
+    restoreBackup(this.props.deviceID, this.state.selectedBackup)
+      .then(({ success }) => {
+        this.setState({
+          restoring: false,
+          errorText: success ? '' : this.props.t('backup.restore.error.general'),
+        });
+        if (this.props.backupOnAfterRestore) {
+          this.props.backupOnAfterRestore(success);
+        }
       });
-      if (this.props.backupOnAfterRestore) {
-        this.props.backupOnAfterRestore(success);
-      }
-    },
-    );
   };
 
   public render() {
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -17,7 +17,7 @@
 
 import React, { Component, FormEvent } from 'react';
 import { Backup } from '../components/backup';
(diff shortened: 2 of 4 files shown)


frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -255,13 +238,16 @@ class BitBox02 extends Component<Props, State> {
...
+          sdCardInserted: response.success,
+          waitDialog: undefined,
+        });
+        if (response.success) {
           return true;
         }
-        if (errorMessage) {
-          alertUser(errorMessage, { asDialog: false });
+        if (response.message) {
+          alertUser(response.message, { asDialog: false });
```
thisconnect, line 250: couldn't destruct the argument directly anymore as TypeScript complained that `errorMessage` is optional.

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -273,9 +259,9 @@ class BitBox02 extends Component<Props, State> {
       settingPassword: true,
       createWalletStatus: 'setPassword',
     });
-    apiPost(this.apiPrefix() + '/set-password').then(({ code, success }) => {
-      if (!success) {
-        if (code === errUserAbort) {
+    setPassword(this.props.deviceID).then((response) => {
+      if (!response.success) {
+        if (response.code === errUserAbort) {
```
thisconnect, line 264: same, can't destruct argument anymore as code is optional (only in case of success = false).

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -330,12 +316,12 @@ class BitBox02 extends Component<Props, State> {
...
-        if (!success) {
+
+      createBackup(this.props.deviceID)
+        .then(() => this.setState({ creatingBackup: false, waitDialog: undefined }))
+        .catch(() => {
           alertUser(this.props.t('bitbox02Wizard.createBackupFailed'), { asDialog: false });
-        }
-        this.setState({ creatingBackup: false, waitDialog: undefined });
-      });
+        });
```
thisconnect, line 324: using promise.reject if success is false so we can use .catch but I don't think it is better, mainly because the passed argument cannot be typed.

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -373,18 +359,12 @@ class BitBox02 extends Component<Props, State> {
...
-        this.setState({
-          appStatus: 'restoreFromMnemonic',
-        });
-      }
-      this.setState({
-        waitDialog: undefined,
-      });
-    });
+      })
+      .finally(() => this.setState({ waitDialog: undefined }));
```
thisconnect, line 367: finally is called after the promise is settled, either resoved or rejected

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

thisconnect:
PTAL

frontends/web/src/routes/device/bitbox02/backups.tsx
```diff
@@ -72,18 +72,16 @@ class Backups extends Component<Props, State> {
...
-      'devices/bitbox02/' + this.props.deviceID + '/backups/restore',
-      this.state.selectedBackup).then(({ success }) => {
-      this.setState({
-        restoring: false,
-        errorText: success ? '' : this.props.t('backup.restore.error.general'),
+    restoreBackup(this.props.deviceID, this.state.selectedBackup)
+      .then(({ success }) => {
+        this.setState({
+          restoring: false,
+          errorText: success ? '' : this.props.t('backup.restore.error.general'),
```
Beerosagos, line 79: It would be nice to rebase on https://github.com/digitalbitbox/bitbox-wallet-app/pull/1933/files after approval and maybe use the error message here instead of the generic one. But it is not localized, so I am not sure about the last part. wdyt?
cc @benma

Beerosagos (approved):
tACK! Very nice
Left just a comment about `restoreBackup`, but LGTM also as is

frontends/web/src/routes/device/bitbox02/backups.tsx
thisconnect: Ok, I'll merge this PR as is, and then we do the nice error message after #1933

PR #1941, opened 2023-02-02
frontend: reduce layout shift on settings and device settings

thisconnect:
Added missing skeleton components so that the layout does not shift while the view is loading.


frontends/web/src/routes/device/bitbox02/settings.tsx
```diff
@@ -56,9 +57,6 @@ export const Settings = ({ deviceID }: TProps) => {
     route(`/passphrase/${deviceID}`);
   };
 
-  if (deviceInfo === undefined) {
-    return null;
-  }
```
thisconnect, line 61: This would remove the flickering if the route would directly point to this component.

instead device settings loads bitbox02.tsx first does various checks and then eventually loads `device/bitbox02/settings.tsx` all that flickrs (even sidebar).

Beerosagos (approved):
tACK!

PR #1942, opened 2023-02-06
frontend: fixed minor css issues in the exchange page(s).

shonsirsha:
1. Added min-height for radio buttons container in the
exchange page to prevent the layout shifting/jumping around

2. Set `guideExists` to be `false` for `<Spinner/>` to fix
"double guide" issue.

### DIFF
--- a/frontends/web/src/routes/buy/pocket.tsx
+++ b/frontends/web/src/routes/buy/pocket.tsx
@@ -165,7 +165,7 @@ export const Pocket = ({ code }: TProps) => {
             />
           ) : (
             <div style={{ height }}>
-              {!iframeLoaded && <Spinner text={t('loading')} /> }
+              {!iframeLoaded && <Spinner guideExists={false} text={t('loading')} /> }
               <iframe
                 onLoad={() => {
                   setIframeLoaded(true);
--- a/frontends/web/src/routes/buy/exchange.module.css
+++ b/frontends/web/src/routes/buy/exchange.module.css
@@ -92,6 +92,9 @@
     margin-left: calc((var(--space-eight) + var(--space-quarter)));
 }
 
+.radioButtonsContainer {
+    min-height: 180px;
+}
 
 .title {
     font-size: 2rem;
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -213,7 +213,7 @@ export const Exchange = ({ code, accounts }: TProps) => {
                   <InfoButton onClick={() => setInfo('region')} />
                 </div>
 
-                <div>
+                <div className={style.radioButtonsContainer}>
                   {noExchangeAvailable && (
                     <p className={style.noExchangeText}>{t('buy.exchange.noExchanges')}</p>
                   )}
--- a/frontends/web/src/routes/buy/moonpay.tsx
+++ b/frontends/web/src/routes/buy/moonpay.tsx
@@ -91,7 +91,7 @@ export const Moonpay = ({ accounts, code }: TProps) => {
               />
             ) : (
               <div style={{ height }}>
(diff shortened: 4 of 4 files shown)


frontends/web/src/routes/buy/iframe.module.css
```diff
@@ -1,21 +1,17 @@
 .container {
     display: flex;
-    flex-basis: 100%;
-    flex-grow: 0;
-    flex-shrink: 1;
-    flex-direction: column;
-    overflow: hidden;
-    position: relative;
+    width: 100%;
+    height: 100vh;
```
thisconnect, line 4: please check that the fix here doens't break, see this commit https://github.com/digitalbitbox/bitbox-wallet-app/commit/a11b6d60490e54e3dd8f4ec98dcb5065505e89a1

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -57,27 +55,13 @@ export const Pocket = ({ code }: TProps) => {
...
-    if (resizeTimerID) {
-      clearTimeout(resizeTimerID);
-    }
-    resizeTimerID = setTimeout(() => {
-      if (!ref.current) {
-        return;
-      }
-      setHeight(ref.current.offsetHeight);
-    }, 200);
-  };
```
thisconnect, line 80: how does this fix the double guide?

frontends/web/src/routes/buy/iframe.module.css
shonsirsha: iframe scrolling seems to be okay.

https://user-images.githubusercontent.com/28676406/216982735-d59d8274-8750-46c7-992b-ce47e608189a.mp4

frontends/web/src/routes/buy/pocket.tsx
shonsirsha: I removed everything that has anything to do with setting the `height` in the `pocket` and `moonpay` file (I forgot to do so in Moonpay, will do it now). The reason for this is because with the new css changes, we don't use this dynamic value of `height` anymore to control the `div` height, but rather its height will always be `100%`.

thisconnect: is this related to the double guide? I still don't see what fixes the 2x guide entries.

shonsirsha: What fixed it was actually the CSS changes. With the css changes, we don't need `height` to be set from the component anymore (height for both the div and iframe). Therefore, im removing it here.

 I first removed the setHeight, which is in the `onResize`. And it seems like `onResize` is only being used to `setHeight` therefore it has no use anymore once I removed `setHeight` - thus I removed it. Does it make sense? 🙂 

ofc we don't have to remove `onResize` but again, there seems to be no use of it anymore.

shonsirsha:
Updated commit message, hope it's clearer now:)

thisconnect:
unfortunatelly the height change breaks https://github.com/digitalbitbox/bitbox-wallet-app/commit/a11b6d60490e54e3dd8f4ec98dcb5065505e89a1 scrolling to the end of the iframe in the Qt app, in webdev it is ok.

@shonsirsha could you make a separate PR for the min-height, so in the meantime we can merge the minheight fix and look for a different solution for the double guide?

also we should look into what the Guide link in the Spinner does, maybe it can be passed as child, not sure yet (haven't looked at it).

shonsirsha:
@thisconnect thanks for testing it in QT! I reverted the relevant CSS changes for `iframe.module.css` and also changes in `pocket.tsx` & `moonpay.tsx`. I now set `guideExists={false}` for the spinner in these 2 files. This should fix the double guide issue without playing around with the CSS, since it may interfere with scrolling as you said.

Could you please test this one out again in QT? Many thx. :)

thisconnect (approved):
tested LGTM

PR #1943, opened 2023-02-06
docs/BUILD: update system requirements

benma:
We long moved to Ubuntu 18.04. Newer versions are of course also supported for Ubuntu and the other OSes.


thisconnect (approved):
LGTM

PR #1944, opened 2023-02-06
frontend: fix flickering sidebar when navigating to device settings

thisconnect:
Navigating to device settings triggered the sidebar to close, this was the reason that it looked like the whole app window flickered.

Forcing the sidebar to close on will mount is not necessary as it will open/close later depending on the device state.

Test on medium screen (when hamburger is visible)
- reset device
- navigate through app
- open app with unlocked BB
- plugin uninitialized BB on waiting screen or app settings
- other tests?

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -102,13 +102,6 @@ class BitBox02 extends Component<Props, State> {
 
   private unsubscribe!: () => void;
 
-  public UNSAFE_componentWillMount() {
-    const { sidebarStatus } = panelStore.state;
-    if (['', 'forceCollapsed'].includes(sidebarStatus)) {
-      setSidebarStatus('forceHidden');
-    }
-  }
-
   public componentDidMount() {
     getVersion(this.props.deviceID).then(versionInfo => {
       this.setState({ versionInfo });


frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -102,13 +102,6 @@ class BitBox02 extends Component<Props, State> {
 
   private unsubscribe!: () => void;
 
-  public UNSAFE_componentWillMount() {
-    const { sidebarStatus } = panelStore.state;
-    if (['', 'forceCollapsed'].includes(sidebarStatus)) {
-      setSidebarStatus('forceHidden');
-    }
-  }
-
```
shonsirsha, line 111: Looks fine, but do you know why this was there in the first place?

thisconnect: I only found "Hide sidebar during setup" see https://github.com/digitalbitbox/bitbox-wallet-app/commit/ea756e04149a87e5d52be78dccfb8d8287fcdbb1

and https://github.com/digitalbitbox/bitbox-wallet-app/pull/529

I tested reseting the device and setup again, also played around with reopening the app during setup and replugin the uninitialized BB. If I read the code right `bitbox02.tsx` always sets the setSidebarStatus after it received getStatus, so it should be fine to not initialize it here.

thisconnect:
rebased PTAL, also @Beerosagos if you have a moment could you also test?

shonsirsha (approved):
tested LGTM

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
thisconnect: Ah I just realized why this maybe was needed and now is not necessary anymore:

I changed the bitbox02.tsx component a while ago to use `<View>` component with fullscreen prop. Fullscreen prop always covers the sidebar anyway (CSS). Probably this was not the case before, so the sidebar had to be programmatically hidden.

Beerosagos:
Tested on big screen with sidebar always visible, LGTM!

PR #1945, opened 2023-02-06
frontend: reword pairing code verification

thisconnect:
The pairing code prompt does not give any information about why that
the user should confirm and has no visual feedback after confirming
on the device.

There are also unused and inconsistent pairing translations.

Changed:
- added new device detected sentence to unpaired message.
- change to previously unused pairing.unpaired translation to be
  consistent with the other keys
- show a paired message after confirming on the device using unused
  pairing.paired translation
- move attestation error before the pairing messages (this should
  be the first information the user reads if it ever happens)

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -216,9 +216,9 @@
     "noPasswordMatch": "Passwords did not match, please try again.",
     "pairing": {
       "failed": "Unconfirmed pairing. Please replug your BitBox02.",
-      "paired": "You have confirmed on your device that the code matches. If this is correct, you can continue by clicking the button below.",
+      "paired": "You have confirmed the following code on your device. Please continue.",
       "title": "Verify pairing code",
-      "unpaired": "A new BitBox02 has been detected. Please verify that the following code matches what is shown on your device. If the code matches, touch underneath the check mark on your BitBox02 and then click the button below to continue."
+      "unpaired": "An unpaired BitBox02 has been detected. Please verify the pairing code matches what is shown on your BitBox02."
     },
     "restoreFromMnemonic": {
       "failed": "Restoring from recovery words failed, please try again."
@@ -260,9 +260,6 @@
       "restoreMicroSD": "Restore from microSD card",
       "restoreMnemonic": "Restore from recovery words",
       "title": "Setup your BitBox02"
-    },
-    "stepUnpaired": {
-      "verify": "Please verify the pairing code matches what is shown on your BitBox02."
     },
     "success": {
       "text": "Hooray! Your BitBox02 is now ready to use. \n\nFor further information on how to use the BitBoxApp, please use the in-app guide by clicking the question mark on the top right corner.",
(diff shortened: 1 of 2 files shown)


shonsirsha (approved):
untested LGTM. 

Unsure if you wanna move the translation (the TODO thing) in this PR or not, but regardless it'd wouldve been a small change, so i'll approve it here. :)

Also maybe good to attach a preview (video/screenshot) for stuff like this.

thisconnect:
Screen Shot 2023-02-08 at 19 37 24

thisconnect:
Screen Shot 2023-02-08 at 19 37 35

thisconnect:
Screen Shot 2023-02-13 at 09 45 18

thisconnect:
Screen Shot 2023-02-13 at 09 44 44

thisconnect:
if aborted it shows this, else it goes either to account summary or continues with the setup wizard:

Screen Shot 2023-02-13 at 10 09 26

frontends/web/src/locales/en/app.json
```diff
@@ -216,9 +216,9 @@
     "noPasswordMatch": "Passwords did not match, please try again.",
     "pairing": {
       "failed": "Unconfirmed pairing. Please replug your BitBox02.",
-      "paired": "You have confirmed on your device that the code matches. If this is correct, you can continue by clicking the button below.",
+      "paired": "Please ensure you verified the pairing code on the device and continue.",
       "title": "Verify pairing code",
-      "unpaired": "A new BitBox02 has been detected. Please verify that the following code matches what is shown on your device. If the code matches, touch underneath the check mark on your BitBox02 and then click the button below to continue."
+      "unpaired": "A new BitBox02 has been detected. Please verify the pairing code matches what is shown on your BitBox02."
```
benma, line 221: It is not necessarily a new BitBox02, can also be the same one that lost the pairing for many reasons (new compter, paired with too many other wallets, etc).

thisconnect: yes you are right, but this is an edge case, most of the time it is a new BitBox. "new" as in new to the app, first time seen, not new unboxed. I think it is better to mention something along this line. Maybe instead of "new" it would be "unseen" or "unpaired" ? @benma  any preference?

jadzeidan: "An unpaired BitBox02 has been detected..." sounds good to me.

frontends/web/src/locales/en/app.json
```diff
@@ -216,9 +216,9 @@
     "noPasswordMatch": "Passwords did not match, please try again.",
     "pairing": {
       "failed": "Unconfirmed pairing. Please replug your BitBox02.",
-      "paired": "You have confirmed on your device that the code matches. If this is correct, you can continue by clicking the button below.",
+      "paired": "Please ensure you verified the pairing code on the device and continue.",
```
jadzeidan, line 219: imo I change to "You have confirmed the following code on your device. Please continue."

Reason is because I think the instructions are not longer clear since you already confirmed on the device. The important part is that the code is still displayed.  Because if a  MITM simulating the BB02 simulated a "confirm" on the device, the BB02 will still show the pairing code (since it was not actually confirmed yet. Then it is clear the user something is wrong because the codes don't match.

frontends/web/src/locales/en/app.json
thisconnect: updated pr and locize

frontends/web/src/locales/en/app.json
thisconnect: updated pr and locize

PR #1946, opened 2023-02-07
backend/pocket: refactor `PocketUrl` to use devservers argument

Beerosagos:
This updates`PocketUrl`to return the address of the Pocket widget from staging or production environment depending on the `devservers` flag, instead of the net used, which was the previous approach.

It also adds a `servewallet-mainnet-prodservers` target in the Makefile, which runs the server on mainnet and with devservers flag disabled.

The reason for this PR is that address signing is currently disabled for the btc testnet at a firmware level, which forces using the mainnet when working on Pocket integration. This update allows choosing which Pocket environment should be used with a proper flag.

### DIFF
--- a/frontends/web/src/api/exchanges.ts
+++ b/frontends/web/src/api/exchanges.ts
@@ -88,10 +88,8 @@ export const verifyAddress = (address: string, accountCode: AccountCode): Promis
   return apiPost('exchange/pocket/verify-address', { address, accountCode });
 };
 
-export const getPocketURL = (accountCode: string) => {
-  return (): Promise<string> => {
-    return apiGet(`exchange/pocket/api-url/${accountCode}`);
-  };
+export const getPocketURL = (): Promise<string> => {
+  return apiGet('exchange/pocket/api-url');
 };
 
 export type SupportedExchanges= {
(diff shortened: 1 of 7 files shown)


frontends/web/src/api/exchanges.ts
```diff
@@ -88,9 +88,9 @@ export const verifyAddress = (address: string, accountCode: AccountCode): Promis
   return apiPost('exchange/pocket/verify-address', { address, accountCode });
 };
 
-export const getPocketURL = (accountCode: string) => {
+export const getPocketURL = () => {
   return (): Promise<string> => {
```
benma, line 92: could remove the function nesting

benma (changes requested):
Please document this new addition here:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/arguments/arguments.go#L54

Also, in another commit or PR, remove the `devmode` argument - looks unused to me.

Beerosagos:
Addressed comments and rebased, thanks!

> Also, in another commit or PR, remove the devmode argument - looks unused to me.

Will do in a separate PR on master :+1:

Beerosagos:
>     Also, in another commit or PR, remove the devmode argument - looks unused to me.

PR opened here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1952

PR #1947, opened 2023-02-07
frontend/pocket: add xpub sharing logic

Beerosagos:
This pushes the version of `request-address` js library to 0.0.6 and adds the handling of `RequestExtendedPublicKeyV0Message` requests, that allows sharing the account xpub with the Pocket widget.

NOTE: in order to test this PR, please force the staging environment for Pocket:
```
apiURL = pocketMainTestURL + "/" + locale + "/" + pocketWidgetTest
```
 in https://github.com/Beerosagos/bitbox-wallet-app/blob/xpub-sharing/backend/exchanges/pocket.go#L73

I opened a separate PR to avoid needing this workaround in the future, see https://github.com/digitalbitbox/bitbox-wallet-app/pull/1946

### DIFF
--- a/frontends/web/src/routes/buy/pocket.tsx
+++ b/frontends/web/src/routes/buy/pocket.tsx
@@ -16,10 +16,13 @@
 
 import { useTranslation } from 'react-i18next';
 import { useState, useEffect, createRef } from 'react';
-import { MessageVersion, parseMessage, serializeMessage, V0MessageType } from 'request-address';
+import { RequestAddressV0Message, MessageVersion, parseMessage, serializeMessage, V0MessageType } from 'request-address';
 import { getConfig } from '../../api/backend';
+import { getTransactionList } from '../../api/account';
 import { Dialog } from '../../components/dialog/dialog';
+import { confirmation } from '../../components/confirm/Confirm';
 import { verifyAddress, signAddress, getPocketURL } from '../../api/exchanges';
+import { getInfo } from '../../api/account';
 import { Header } from '../../components/layout';
 import { Spinner } from '../../components/spinner/Spinner';
 import { PocketTerms } from './pocket-terms';
@@ -42,11 +45,12 @@ export const Pocket = ({ code }: TProps) => {
 
   const iframeURL = useLoad(getPocketURL(code));
   const config = useLoad(getConfig);
+  const accountInfo = useLoad(getInfo(code));
 
   const ref = createRef<HTMLDivElement>();
   const iframeRef = createRef<HTMLIFrameElement>();
-  var signing = false;
-  var resizeTimerID: any = undefined;
+  let signing = false;
+  let resizeTimerID: any = undefined;
 
   const name = 'Bitcoin';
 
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -152,6 +102,113 @@ export const Pocket = ({ code }: TProps) => {
...
+    verifyAddress(address, code)
+      .then(response => {
+        setVerifying(false);
+        if (!response.success) {
+          if (response.errorCode === 'addressNotFound') {
+            // This should not happen, unless the user receives a tx on the same address between the message signing
+            // and the address verification.
+            alertUser(t('buy.pocket.usedAddress', { address:  address }));
+          } else {
+            alertUser(t('genericError'));
```
benma, line 138: if use you `unknownError` you can pass the error message to be shown to the user as well.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -152,6 +102,113 @@ export const Pocket = ({ code }: TProps) => {
...
+            console.log('error: ' + response.errorMessage);
+          }
+        }
+      });
+  };
+
+  const sendXpub = () => {
+    if (accountInfo) {
+      let bitcoinSimple = accountInfo.signingConfigurations[0].bitcoinSimple;
```
shonsirsha, line 147: Const instead of let here makes sense:

```suggestion
      const bitcoinSimple = accountInfo.signingConfigurations[0].bitcoinSimple;
```

also there are some `var`s in this file. We can change those to `let`.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -56,6 +60,8 @@ export const Pocket = ({ code }: TProps) => {
     }
   }, [config]);
 
+
+
```
shonsirsha, line 64: prob. added by mistake? 😄

shonsirsha (dismissed):
Didn't test this one out yet - just some very minor stuff looking at the code.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -152,6 +102,113 @@ export const Pocket = ({ code }: TProps) => {
...
+          extendedPublicKey: xpub,
+        });
+        current.contentWindow?.postMessage(message, '*');
+      }
+    }
+  };
+
+  const handleRequestXpub = () => {
+    getTransactionList(code as AccountCode).then(txs => {
+      if (txs.length > 0) {
```
thisconnect, line 166: I think this is fine: but getTransactionList gets all transactions (unified accounts). Only the SegWit (p2wpkh) xpub will be shared. So user could have nothing on p2wpkh account, but we still show this warning if there is coins on p2tr or p2wpkh-p2sh.
As mentioned I think it is good as is.

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -152,6 +102,113 @@ export const Pocket = ({ code }: TProps) => {
     current.contentWindow?.postMessage(message, '*');
   };
 
+  const handleRequestAddress = (message: RequestAddressV0Message) => {
+    signing = true;
+    const addressType = message.withScriptType ? String(message.withScriptType) : '';
+    const withMessageSignature = message.withMessageSignature ? String(message.withMessageSignature) : '';
```
thisconnect, line 108: `withMessageSignature` should either be falsy (undefined, null or false) or a string. So I think `String()` is not needed 🤔

frontends/web/src/routes/buy/pocket.tsx
thisconnect: also is `as AccountCode` needed?

I'll put on my todo to look into if we should type code everywhere to use AccountCode instead of string.

thisconnect (approved):
LGTM, with some small typescript nits/questions and a comment about unified account

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: Good points! Done :pray:

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -152,6 +102,113 @@ export const Pocket = ({ code }: TProps) => {
...
+        });
+        current.contentWindow?.postMessage(message, '*');
+      }
+    }
+  };
+
+  const handleRequestXpub = () => {
+    getTransactionList(code as AccountCode).then(txs => {
+      if (txs.length > 0) {
+        confirmation(t('buy.pocket.previousTransactions'), async result => {
```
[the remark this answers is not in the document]
Beerosagos: Right, thanks!

frontends/web/src/routes/buy/pocket.tsx
Beerosagos: > I think this is fine: but getTransactionList gets all transactions (unified accounts). Only the SegWit (p2wpkh) xpub will be shared. So user could have nothing on p2wpkh account, but we still show this warning if there is coins on p2tr or p2wpkh-p2sh.
As mentioned I think it is good as is.

You're right! this is something to keep in mind, but I also think it is good as is.

Beerosagos:
@shonsirsha pushed an update PTAL :pray:

thisconnect (approved):
LGTM

PR #1949, opened 2023-02-07
backend/eth/account: fix race condition on tx-proposal

Beerosagos:
A race condition between the `Initialize()` and `TxProposal()` methods was causing a deadlock where they were blocking each other in accessing `account.Synchronizer` and `account.updateLock`. This recently caused CI tests to timeout and fail from time to time.

This changes the behavior of `newTx()`, which is called by `TxProposal()`, making it returning an errror instead of waiting when the account is not yet synced. This should fix the race condition.

### DIFF
--- a/backend/accounts/errors/errors.go
+++ b/backend/accounts/errors/errors.go
@@ -39,6 +39,8 @@ var (
 	// ErrFeeTooLow is returned when the custom fee the user entered is too low to be able to
 	// broadcast the transaction.
 	ErrFeeTooLow = TxValidationError("feeTooLow")
+	// ErrAccountNotsynced is used when the account sync has not successfully finished.
+	ErrAccountNotsynced = TxValidationError("accountNotSynced")
 
 	// ErrNotAvailable is returned if data required is not available yet. Example: the headers are
 	// not synced yet, which is a prerequisite to making a timeseries of the portfolio.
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -379,7 +379,11 @@ class Send extends Component<Props, State> {
         this.setState({ proposedFee: undefined });
         if (errorCode) {
           this.unregisterEvents();
-          alertUser(errorCode, { callback: this.registerEvents });
+          if (errorCode === 'accountNotSynced') {
+            alertUser(this.props.t('send.error.accountNotSynced'), { callback: this.registerEvents });
```
Beerosagos, line 383: This is pretty ugly, but I guess that adapting the send page to calls that return instead of waiting if the account is not synced will require a bigger effort and a dedicated PR. 

Related to https://github.com/digitalbitbox/bitbox-wallet-app/pull/1859

Beerosagos:
Not sure why, but some tests are failing with this error `GitHub Actions has encountered an internal error when running your job.`, while everything is working fine on my machine..

Beerosagos:
cc @thisconnect for the frontend

frontends/web/src/routes/account/send/send.tsx
benma: I am not sure you even need to handle the error in the frontend - while the sync is going on, the app shows a "snychronizing" screen and you cannot go to the send screen, right? Maybe simpler to just remove all of that. 

Btw, in my other recent PRs that do the same as this one (return errors instead of blocking) I didn't do any frontend error handling for the same reason.

Beerosagos: Yeah, that makes sense. I removed the frontend part, PTAL

benma (approved):

PR #1950, opened 2023-02-08
frontend/exchange: make exchange region selection optional

Beerosagos:
Region selection in exchange page was mandatory in order to proceed with the buy workflow. This allows the user to explicitly avoid specifying the region.

### DIFF
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -89,11 +89,18 @@ export const Exchange = ({ code, accounts }: TProps) => {
     regions.sort((a, b) => a.text.localeCompare(b.text, i18n.language));
     setRegions(regions);
 
+    // if user had selected no region before, do not pre-select any.
+    if (config.frontend.selectedExchangeRegion === '') {
+      return;
+    }
+
     if (config.frontend.selectedExchangeRegion) {
+      // pre-select config region
       setSelectedRegion(config.frontend.selectedExchangeRegion);
       return;
     }
 
+    // user never selected a region preference, will derive it from native locale.
     const userRegion = getRegionNameFromLocale(nativeLocale || '');
     //Region is available in the list
     const regionAvailable = !!(regionList.regions.find(region => region.code === userRegion));
@@ -198,7 +205,6 @@ export const Exchange = ({ code, accounts }: TProps) => {
                       className={style.extraLeftSpace}
                       options={[{
                         text: t('buy.exchange.selectRegion'),
-                        disabled: true,
                         value: '',
                       },
                       ...regions]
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -50,9 +50,9 @@ type TOption = {
     value?: string;
 }
 
-const SelectedRegionIcon = ({ regionCode }: { regionCode: string }) => {
+const SelectedRegionIcon = ({ regionCode, noRegion }: { regionCode: string, noRegion: string }) => {
   return <span className={style.flag}>
-    {regionCode === '' ? <Globe /> : <span className={`fi fi-${regionCode}`}></span>}
+    {regionCode === noRegion ? <Globe /> : <span className={`fi fi-${regionCode}`}></span>}
   </span>;
```
thisconnect, line 56: oups, I wanted to play around if we could still keep empty string to select the globe and now ended up with an alternative approach. 

Instead of changing noRegion to dash wouldn't something like this work as well instead of this PR?

```diff
diff --git a/frontends/web/src/routes/buy/exchange.tsx b/frontends/web/src/routes/buy/exchange.tsx
index 34b85e729..80ff74167 100644
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -93,6 +93,10 @@ export const Exchange = ({ code, accounts }: TProps) => {
       setSelectedRegion(config.frontend.selectedExchangeRegion);
       return;
     }
+    // user had selected no region before, do not continue to guessing region
+    if (config.frontend.selectedExchangeRegion === '') {
+      return;
+    }
 
     const userRegion = getRegionNameFromLocale(nativeLocale || '');
     //Region is available in the list
@@ -198,7 +202,6 @@ export const Exchange = ({ code, accounts }: TProps) => {
                       className={style.extraLeftSpace}
                       options={[{
                         text: t('buy.exchange.selectRegion'),
-                        disabled: true,
                         value: '',
                       },
                       ...regions]
@@ -221,7 +224,6 @@ export const Exchange = ({ code, accounts }: TProps) => {
                   <div>
                     {!noExchangeAvailable && allExchangeDeals && allExchangeDeals.exchanges.map(exchange => exchange.supported && (<ExchangeSelectionRadio
                       key={exchange.exchangeName}
-                      disabled={!selectedRegion}
                       id={exchange.exchangeName}
                       exchangeName={exchange.exchangeName}
                       deals={exchange.deals}
```

thisconnect (commented):
defering to @shonsirsha

frontends/web/src/routes/buy/exchange.tsx
Beerosagos: Very nice! LGTM :) I pushed an updated following your approach, thanks!

thisconnect (approved):
LGTM, tested on 2 laptops one returns "" and the other "en_US" for native-locale. the "en_US" did preselect United States and remember when I selected to "Not specified"

shonsirsha (approved):
LGTM.

PR #1951, opened 2023-02-08
frontend: change bb01 banner style to warning

thisconnect:
As it has been 3 years since BitBox01 EOL and 3rd party services such as MEW silently dropped support, the banner in the BitBoxApp should be displayed in a more eye catchy warning style to encourage users moving their funds.


Beerosagos (approved):
tACK!

PR #1954, opened 2023-02-13
backend: comply with new golangci tests

Beerosagos:
Upgrading `golanci-lint` to v1.50 showed several nits in the code. With this I updated the code where possible and skipped three new tests.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -131,6 +131,9 @@ linters:
     - makezero
     - nolintlint
     - gocyclo
+    - nosnakecase
+    - interfacebloat
+    - revive
   disable-all: false
 
 issues:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ env:
   # https://docs.github.com/en/packages/guides/pushing-and-pulling-docker-images
   #
   # Keep this in sync with default in scripts/travis-ci.sh.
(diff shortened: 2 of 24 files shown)


.golangci.yml
```diff
@@ -131,6 +131,9 @@ linters:
     - makezero
     - nolintlint
     - gocyclo
+    - nosnakecase
```
Beerosagos, line 134: This generates errors like 
```
frontends/qt/server/server.go:85:26: _Ctype_int contains underscore. You should use mixedCap or MixedCap. (nosnakecase)
func backendCall(queryID C.int, s *C.char) {
```

.golangci.yml
```diff
@@ -131,6 +131,9 @@ linters:
...
+    - interfacebloat
```
Beerosagos, line 135: This generates errors like this. There are a lot of them
```
backend/handlers/handlers.go:66:14: the interface has more than 10 methods: 37 (interfacebloat)
type Backend interface {
```

.golangci.yml
```diff
@@ -131,6 +131,9 @@ linters:
...
+    - revive
```
Beerosagos, line 136: This generates errors like this: we could also add package comments, but I don't know if it is really useful and I think that we are missing a lot of them.
```
util/observable/action/action.go:15:1: package-comments: should have a package comment (revive)
package action
```

benma (approved):
nice!

PR #1957, opened 2023-02-14
backend/usb: run all hidapi API calls in a dedicated OS thread

benma:
For unknown reasons, on macOS, there are intermittent errors during hid.Device.Write():

`IOHIDLibUserClient:0x100002065 setReport failed: 0xe00002bc` In the BitBoxApp, the issue appears most frequently in the loop polling the device for a API call status in query.go/rawQueryV7.

Running all calls in the same OS thread seems to resolve the issue, even though it is not clear why that is. Maybe something in the macOS SDK relies on thread-local variables, or there is a memory-bug that does not get triggered this way.

After a long time debugging this, I am still not sure what causes it, especially because it also happens in a test program without any goroutines of this sort:

`main() { for { C.hid_write(...); C.hid_read(...) }`.

The issue disappears however if there is a `runtime.Gosched()` in every loop iteration in this test program, which is perplexing, as there are no other goroutines to yield to. Further investigation is needed to figure out what is going on under the hood - for now we apply the fix to run all calls in the same thread. `runtime.Gosched()`
is not needed in our case as the scheduler is automatically invoked
when sending to and reading from the `funcCalls` channel.

### DIFF
--- a/backend/devices/usb/hid.go
+++ b/backend/devices/usb/hid.go
@@ -17,10 +17,33 @@ package usb
 import (
 	"encoding/hex"
 	"io"
+	"runtime"
 
 	"github.com/digitalbitbox/bitbox-wallet-app/util/logging"
 	hid "github.com/digitalbitbox/usb"
 )
+
+// Run functions sent to this channel in a goroutine fixed to an OS thread. We will run all hidapi
+// API calls in that thread. This is needed for hidapi to work properly on macOS. The reason for
+// this is not entirely clear - maybe something in the macOS SDK relies on thread-local variables
+// between HID API calls. See also https://github.com/libusb/hidapi/issues/503.
+var funcCalls chan func()
+
+func init() {
+	funcCalls = make(chan func())
+	go func() {
+		runtime.LockOSThread()
+		for {
+			f := <-funcCalls
+			f()
+		}
+	}()
+}
+
+type funcCallResult[T any] struct {
+	value T
+	err   error
+}
 
 type hidDeviceInfo struct {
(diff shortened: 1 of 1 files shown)


backend/devices/usb/hid.go
```diff
@@ -66,28 +89,76 @@ func (info hidDeviceInfo) Identifier() string {
 	return hex.EncodeToString([]byte(info.DeviceInfo.Path))
 }
 
+// singleThreadedDevice runs all hidapi functions in the same OS thread. See the docs of the
+// `funcCalls` variable for more info.
+type singleThreadedDevice struct {
```
Beerosagos, line 94: Maybe mention that it implements `io.ReadWriteCloser`?

Beerosagos (commented):
Very interesting! Tested on Linux, LGTM :+1: 
Do we also need a test on Windows?

thisconnect:
works for me on macOS 11.7, also looks like this gets rid of an annoying error in `webdev`  🎉  (frame mismatch or similar)

benma:
@Beerosagos i tested on Windows, seems to work fine.

Beerosagos (approved):
LGTM!

benma:
@Beerosagos with more testing, I got at least one "timeout" where the password entry was aborted. I'll test some more on this branch and master.

benma:
> @Beerosagos with more testing, I got at least one "timeout" where the password entry was aborted. I'll test some more on this branch and master.

nvm, seems like it was a fluke. Could not reproduce.

PR #1959, opened 2023-02-14
frontend: refactor settings toggle into settingstoggle

thisconnect:
Refactored settings that use a toggle into SettingsToggle.tsx.

Cleaned up and removed CSS module imports from other components and duplicate CSS as settings toggle CSS was copied to different components that use some kind of SettingsToggle.


frontends/web/src/components/toggle/toggle.module.css
```diff
@@ -1,5 +1,6 @@
 .container {
   display: inline-block;
+  flex-shrink: 0;
```
thisconnect, line 3: This fixes an edge case when the element before the toggle was too long, it would squeeze the toggle together.

The old fix was `.setting > label` (could remove this in 2 places)

frontends/web/src/components/settingsButton/settingsToggle.tsx
```diff
@@ -0,0 +1,35 @@
...
+  children,
+  ...props
+}: Props) => {
+  return (
+    <div className={style.setting}>
+      {children}
+      <Toggle {...props} />
+    </div>
+  );
+};
```
thisconnect, line 35: DOM is slightly different and uses less nested elements, take a look at settings.tsx

frontends/web/src/routes/settings/settings.tsx
```diff
@@ -277,37 +277,30 @@ class Settings extends Component<Props, State> {
...
-                                checked={config.backend.btcUnit === 'sat'}
-                                id="satsUnit"
-                                onChange={this.handleToggleSatsUnit} />
-                            </div>
+                            </SettingsToggle>
+                            <SettingsToggle
+                              checked={config.backend.btcUnit === 'sat'}
+                              id="satsUnit"
+                              onChange={this.handleToggleSatsUnit}>
+                              {t('settings.expert.useSats')}
```
thisconnect, line 302: doens't use `<div><p className="m-none">` anymore (when not needed).

the `<p>` element was only used to apply the correct font styling, this is now done directly in the SettingsToggle.tsx component.

thisconnect (commented):
Refactored settings toggle as I need it also for dark mode toggle.

shonsirsha (approved):
tested LGTM 👍 a small question, do we have any reason why we put ids for these toggles `id=coinControl`, or `id=exptertFee`? If not, I think we could remove it 🤷‍♂️ WDYT?

thisconnect:
I think you are right, looks like it is not used. If there was  a label with a for attribute it would target the input with id attribute, maybe this is a leftover. Label now is parent of the input element so it has the same effect as for/id.

Anyway I wanted to make the least changes with the goal of having a reusable settingstoggle component and don't want to test again. So prob merge as is, let's remove this another time.

shonsirsha:
sounds good @thisconnect .

PR #1961, opened 2023-02-15
vendor: replace karalabe/usb with digitalbitbox/usb

benma:
karalabe/usb seems unmaintained. We forked it to fix a bug where a compilation on macOS SDK 12.0+ targeting SDK <12.0 resulted in a crash when running the program on <12. See
https://github.com/digitalbitbox/usb/commit/6f0e76394cf453718017d2077569e19d013a4c51 for more details.


Beerosagos (approved):
tACK on Linux

PR #1966, opened 2023-02-16
scripts: make dockerdev.sh and docker_exec.sh podman-compatible

benma:
podman is a (almost) drop-in replacment to docker that runs rootless and without a daemon. Since it runs rootless, it maps the host user to the root user inside the container. This clashes with the tricks used for docker to mount the volume with the same host user id and group id. This script removes these docker tricks when using podman.


benma:
@Beerosagos this is the same as https://github.com/digitalbitbox/bitbox02-firmware/pull/1039

Beerosagos:
I suggest updating Makefile `dockerinit` target. Otherwise, in case podman is installed, it pulls the image with docker and then tries to execute it with podman (see  https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/Makefile#L91).

benma:
> I suggest updating Makefile `dockerinit` target. Otherwise, in case podman is installed, it pulls the image with docker and then tries to execute it with podman (see https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/Makefile#L91).

Done.

Beerosagos (approved):
tACK!

PR #1968, opened 2023-02-16
frontend: fix scanning qr code with amount in sats mode

thisconnect:
The send interface shows an error when scanning a QR code with amount when in sats mode. Reason: sats cannot have decimals on L1.

Added a check to see if amount is specified and app in sats mode, in which case it tires to parse the amount and multiply with 100m.

100m is written with numeric separation for better readability https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -501,6 +501,10 @@ class Send extends Component<Props, State> {
       address = url.pathname;
       if (this.isBitcoinBased()) {
         amount = url.searchParams.get('amount') || '';
+        if (amount && this.state.btcUnit === 'sat') {
+          // convert expected amount in BTC to sat
+          amount = `${Number(amount) * 100_000_000}`;
+        }
       }
     } catch {
       address = uri;


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -501,6 +501,10 @@ class Send extends Component<Props, State> {
       address = url.pathname;
       if (this.isBitcoinBased()) {
         amount = url.searchParams.get('amount') || '';
+        if (amount && this.state.btcUnit === 'sat') {
+          // convert expected amount in BTC to sat
+          amount = `${Number(amount) * 100_000_000}`;
```
thisconnect, line 506: I considered if this needs to be in try/catch if the qr code has evil data in the amount field, but afaik Number will just return NaN, so worstcase this will be a string with 'NaN' that then is shown as "Invalid amount" error

Screen Shot 2023-02-16 at 15 16 49

benma, line 506: feels like a job for the backend :thinking: feeling  a bit uncomfortable having conversion code exist both in the backend and frontend. the backend already has all the functions etc, and will also not suffer from any sort of floating point problems that could come from `Number(...)`.

thisconnect: 👍  yep makes sense.

PR #1969, opened 2023-02-20
frontend: fix language button regression

thisconnect:
The language button didn't work in some views that have a form around the view.

regression introduced in 77fd86e40491d0af98578b2015f0fe194d609390

test:
1) create new wallet
2) on enter-name click on langauge button
3) invalid input

reason was that buttons by default are type submit, so if there is a form wrapping the whole view, clicking the button fires the onsubmit event that propagates all up the the form wrapping the view.


shonsirsha (approved):
lgtm

PR #1970, opened 2023-02-21
frontend: limit success icon min and max size

thisconnect:
Success icon is sometimes rendered too big, this seems to be a regression from View component refactoring.

Changed to limit the max size of large icons in the view component.


shonsirsha (approved):
tested LGTM.

PR #1971, opened 2023-02-22
android: run go backend on a dedicated service

Beerosagos:
Before this, the Go backend was invoked on Android throught the `GoViewModel` class. We recently discovered that this solution could cause timeout errors if the app loose the focus during a polling to the bb02, e.g. while the user is verifying an address.

The Go backend is now started and executed inside a dedicated `GoService`, bound to the `MainActivity`.
When the service starts it immediately promotes itself to foreground process, to mantain a good priority even when the app is out of focus. In order to go foreground the process needs a dedicated notification, that is set to a low priority level to be as less intrusive as possible (basically a text in the taskbar that says that the app is running).


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,90 @@
...
+
+import goserver.GoAPIInterface;
+import goserver.GoEnvironmentInterface;
+import goserver.Goserver;
+
+public class GoService extends Service {
+    // Binder given to clients
+    private final IBinder binder = new GoServiceBinder();
+
+    private Boolean started = false;
```
benma, line 22: why not lowercase `boolean`?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,90 @@
...
+                    NotificationManager.IMPORTANCE_LOW
+            );
+            channel.setDescription("BitBoxApp service notification channel");
+            NotificationManager manager = getSystemService(NotificationManager.class);
+            manager.createNotificationChannel(channel);
+        }
+
+        Notification notification = new NotificationCompat.Builder(this, channelId)
+                .setContentTitle("BitBoxApp Go backend service")
+                .setContentText("Running in the background")
```
benma, line 44: should it say 'running in the foreground' (says foreground below)?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,90 @@
...
+        startForeground(notificationId, notification);
+        Util.log("GoService onCreate completed");
+    }
+
+    @Override
+    public void onDestroy() {
+        Util.log("GoService onDestroy()");
+    }
+
+    public void startServer(String filePath, GoEnvironmentInterface goEnvironment, GoAPIInterface goAPI) {
```
benma, line 59: is there any chance this can be called concurrently? if so the `started` check is not thread-safe.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -371,10 +406,22 @@ protected void onStop() {
         Util.log("lifecycle: onStop");
     }
 
+    @Override
+    protected void onRestart() {
+        // This is here so that if the GoService gets killed while the app is in background it
```
benma, line 411: did you manage to test this?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
```diff
@@ -180,19 +179,24 @@ public void pushNotify(String msg) {
...
         super(app);
 
         this.goEnvironment = new GoEnvironment();
         this.goAPI = new GoAPI();
-
-        Goserver.serve(app.getApplicationContext().getFilesDir().getAbsolutePath(), this.goEnvironment, this.goAPI);
     }
 
     @Override
     public void onCleared() {
```
benma, line 198: can remove the whole function now.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -371,10 +406,22 @@ protected void onStop() {
...
+            bindService(intent, connection, Context.BIND_AUTO_CREATE);
+        }
+        super.onRestart();
+    }
+
     @Override
     protected void onDestroy() {
-        super.onDestroy();
         Util.log("lifecycle: onDestroy");
+        goService.stopServer();
```
benma, line 423: is onDestroy called only when the app shuts down or can it be called at any time? 

asking because i am not sure shutdown followed by serve actually works. does that scenario even happen?

benma (commented):
@Beerosagos I tested it and it seems to work fine, but I don't see any sort of tray icon or notification. Should I?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
Beerosagos: No reason, my Java is just  bit rusted (no pun intended :stuck_out_tongue: )

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
Beerosagos: Good point, I am not completely sure about that. I'll come back on this soon.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
Beerosagos: It actually complains if I remove it: `AndroidViewModel` doesn't have a default constructor apparently

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
Beerosagos: It should be called only when the app shuts down. I added this only to stop the server before it is destroyed. I made some tests and when the app is destroyed there is no service hanging, so should be fine, I guess.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
Beerosagos: I'm still not completely sure about this: it is generally correct to implement bound service's methods in a thread-safe way (see https://developer.android.com/guide/components/processes-and-threads), but in this case we have only one activity that actually binds to the service, so maybe we could avoid that.
It seems safer to handle it in a thread-safe way though: I added a lock guarding the `started` flag.

Beerosagos:
> @Beerosagos I tested it and it seems to work fine, but I don't see any sort of tray icon or notification. Should I?

That is weird, you should get a notification in the system tray similar to this one. Anyway, I don't see any problem if it doesn't show..
immagine

Beerosagos:
Pushed a new update and addressed the comments. PTAL @benma :pray:

benma:
Now I see the notification too. The square empty icon should be replaced with the BitBoxApp logo I think.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
Beerosagos: I actually think that from the user perspective (title and text are shown in the notification) it is still something going on in the background, even if technically is a foreground service. I slightly changed the title, trying to make it simpler to understand for the user. Feel free to suggest anything different, though :)

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
Beerosagos: I noticed that in case of bad connection the backend shutdown could hang on indefinitely, blocking the service destruction and giving issues when the app is started again. To avoid the issue I just removed that, since everything seems to work fine anyway.

Beerosagos:
> Now I see the notification too. The square empty icon should be replaced with the BitBoxApp logo I think.

Now we have a nice icon in the notification :D

Beerosagos:
@benma I should have addressed all the comments, PTAL :pray:

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,89 @@
...
+                    "BitBoxApp",
+                    NotificationManager.IMPORTANCE_LOW
+            );
+            channel.setDescription("BitBoxApp service notification channel");
+            NotificationManager manager = getSystemService(NotificationManager.class);
+            manager.createNotificationChannel(channel);
+        }
+
+        Notification notification = new NotificationCompat.Builder(this, channelId)
+                .setContentTitle("BitBoxApp backend service")
```
benma, line 49: I think just "BitBoxApp" is enough? "backend service" imo is too much noise and confusing for users.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,89 @@
...
+                .setContentText("Running in the background")
```
benma, line 50: I think this could lead users to question what this is doing (background activity can sound a bit scary).

I wonder how we could rephrase it. Maybe "Running in the background to ensure connection stability" (this will also helps with Electrum connection stability). Wdyt?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,89 @@
...
+                    channelId,
+                    "BitBoxApp",
+                    NotificationManager.IMPORTANCE_LOW
+            );
+            channel.setDescription("BitBoxApp service notification channel");
+            NotificationManager manager = getSystemService(NotificationManager.class);
+            manager.createNotificationChannel(channel);
+        }
+
+        Notification notification = new NotificationCompat.Builder(this, channelId)
```
benma, line 48: could you make the app open (come to foreground) when the notification is tapped? i would expect this behavior, as all notifications I have on my phone do this normally.

Beerosagos: Done :+1:

Beerosagos:
Notification  text fixed and App open on notification tap done :) PTAL @benma

benma:
@Beerosagos unfortunately I get a crash upon launch. Found this in logcat which could be the cause:

```
03-03 16:13:43.330   858   858 E AndroidRuntime: Process: ch.shiftcrypto.bitboxapp.debug, PID: 858
03-03 16:13:43.330   858   858 E AndroidRuntime: java.lang.RuntimeException: Unable to create service ch.shiftcrypto.bitboxapp.GoService: java.lang.IllegalArgumentException: ch.shiftcrypto.bitboxapp.debug: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
03-03 16:13:43.330   858   858 E AndroidRuntime: Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.app.ActivityThread.handleCreateService(ActivityThread.java:5035)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.app.ActivityThread.access$1900(ActivityThread.java:315)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2305)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.os.Looper.loopOnce(Looper.java:226)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:313)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:8751)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
03-03 16:13:43.330   858   858 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
03-03 16:13:43.330   858   858 E AndroidRuntime: Caused by: java.lang.IllegalArgumentException: ch.shiftcrypto.bitboxapp.debug: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
03-03 16:13:43.330   858   858 E AndroidRuntime: Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
```

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -0,0 +1,104 @@
...
+        Intent notificationIntent = new Intent(getApplicationContext(),MainActivity.class);
+        notificationIntent.setAction(Intent.ACTION_MAIN);
+        notificationIntent.addCategory(Intent.CATEGORY_LAUNCHER);
+        notificationIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+        PendingIntent contentIntent = PendingIntent.getActivity(
+                getApplicationContext(),
+                0,
+                notificationIntent,
+                PendingIntent.FLAG_UPDATE_CURRENT);
```
benma, line 58: Adding `| PendingIntent.FLAG_IMMUTABLE` here seems to fix the crash I commented about.

benma: Solution courtesy of ChatGPT!

Beerosagos: Great! I pushed a new commit that should fix :) Thanks

benma (approved):
tACK, well done!

PR #1975, opened 2023-02-23
frontend: use generic back button

thisconnect:
Remove custom back button text in favor of generic back button.


Beerosagos (approved):
LGTM!

PR #1976, opened 2023-02-23
frontend: rephrase sd card warning when creating a new wallet

thisconnect:
There was no exlpanation of why the microSD card should be inserted when creating a new wallet during the set wallet name step.

Moved step description next to the title followed by the warning and added minimal information about what the SD Card will be used for, i.e.  to create a backup of the wallet.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -240,7 +240,7 @@
       "nameLabel": "BitBox02 name",
       "namePlaceholder": "My BitBox02",
       "title": "Choose BitBox02 name",
-      "toastMicroSD": "Please make sure your microSD card is inserted in your BitBox02."
+      "toastMicroSD": "Please insert your microSD card into your BitBox02 which will be used to store a backup of the wallet."
     },
     "stepCreateSuccess": {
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -240,7 +240,7 @@
       "nameLabel": "BitBox02 name",
       "namePlaceholder": "My BitBox02",
       "title": "Choose BitBox02 name",
-      "toastMicroSD": "Please make sure your microSD card is inserted in your BitBox02."
+      "toastMicroSD": "Please insert your microSD card into your BitBox02 which will be used to create a backup of the wallet."
```
thisconnect, line 243: not changed in locize yet. will manually update the translation when this is reviewed and merged.

thisconnect:
current (no context what SD card is used for):

  
Screen Shot 2023-02-23 at 11 24 22

thisconnect:
with this change:

changed order so that the warning is not between title+description.

updated inser sd card message to give context.

Screen Shot 2023-02-23 at 11 22 58

frontends/web/src/locales/en/app.json
thisconnect: makes sense, changed.

shonsirsha (approved):
tested LGTM.

PR #1978, opened 2023-02-27
backend: fix external BTC value parsing when sats mode is enabled

Beerosagos:
Parsing external BTC values (e.g. amount scanned through QR scanning feature) did not properly convert BTC values to sats in case sats mode was enabled. This fixes adding a backend endpoint that allows to parse an external BTC value, converting to sats if needed.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -197,8 +197,8 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/test/register", handlers.postRegisterTestKeystoreHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/test/deregister", handlers.postDeregisterTestKeystoreHandler).Methods("POST")
 	getAPIRouter(apiRouter)("/rates", handlers.getRatesHandler).Methods("GET")
-	getAPIRouter(apiRouter)("/coins/convertToPlainFiat", handlers.getConvertToPlainFiatHandler).Methods("GET")
-	getAPIRouter(apiRouter)("/coins/convertFromFiat", handlers.getConvertFromFiatHandler).Methods("GET")
+	getAPIRouter(apiRouter)("/coins/convert-to-plain-fiat", handlers.getConvertToPlainFiatHandler).Methods("GET")
+	getAPIRouter(apiRouter)("/coins/convert-from-fiat", handlers.getConvertFromFiatHandler).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/tltc/headers/status", handlers.getHeadersStatus(coinpkg.CodeTLTC)).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/tbtc/headers/status", handlers.getHeadersStatus(coinpkg.CodeTBTC)).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/ltc/headers/status", handlers.getHeadersStatus(coinpkg.CodeLTC)).Methods("GET")
@@ -737,7 +737,7 @@ func (handlers *Handlers) getConvertToPlainFiatHandler(r *http.Request) (interfa
 
 	currentCoin, err := handlers.backend.Coin(coinpkg.Code(from))
 	if err != nil {
-		logrus.Error(err.Error())
+		handlers.log.WithError(err).Error("Could not get coin " + from)
 		return map[string]interface{}{
 			"success": false,
 		}, nil
@@ -745,7 +745,7 @@ func (handlers *Handlers) getConvertToPlainFiatHandler(r *http.Request) (interfa
 
 	coinAmount, err := currentCoin.ParseAmount(amount)
 	if err != nil {
-		logrus.Error(err.Error())
+		handlers.log.WithError(err).Error("Error parsing amount " + amount)
 		return map[string]interface{}{
 			"success": false,
 		}, nil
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -510,13 +510,31 @@ class Send extends Component<Props, State> {
...
+    const coinCode = this.getAccount()!.coinCode;
+    if (amount && coinCode === 'btc') {
+      // needed to convert to sats if sats mode is enabled.
+      parseExternalBtcAmount(amount).then(result => {
+        if (result.success) {
+          updateState['amount'] = result.amount;
+          this.setState(updateState, () => {
+            this.convertToFiat(this.state.amount);
+            this.validateAndDisplayFee(true);
+          });
```
benma, line 523: if you make parseQRResult an async function, the control flow here will become easier i think - you can await the result of `parseeExternalBtcAmount` and not duplicate these lines.

backend/handlers/handlers.go
```diff
@@ -203,6 +203,7 @@ func NewHandlers(
 	getAPIRouter(apiRouter)("/coins/ltc/headers/status", handlers.getHeadersStatus(coinpkg.CodeLTC)).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/btc/headers/status", handlers.getHeadersStatus(coinpkg.CodeBTC)).Methods("GET")
 	getAPIRouter(apiRouter)("/coins/btc/set-unit", handlers.postBtcFormatUnit).Methods("POST")
+	getAPIRouter(apiRouter)("/coins/btc/parseExternalAmount", handlers.getBTCParseExternalAmount).Methods("GET")
```
benma, line 206: why not put this into the `backend/coins/btc/handlers/` package since it's about the btc coin and not a general backend method?

backend/handlers/handlers.go
```diff
@@ -690,6 +691,43 @@ func (handlers *Handlers) getRatesHandler(_ *http.Request) (interface{}, error)
 	return handlers.backend.RatesUpdater().LatestPrice(), nil
 }
 
+func (handlers *Handlers) getBTCParseExternalAmount(r *http.Request) (interface{}, error) {
+	amount := r.URL.Query().Get("amount")
+
```
benma, line 696: maybe nice to put `type result struct { Success bool, Amount string }` here with the json tags, as we started doing in other places, so one can see the result at a glance instead of checking each return statement.

benma (changes requested):

backend/handlers/handlers.go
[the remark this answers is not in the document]
Beerosagos: Done, also noticed that coin/fiat conversion endpoints were named with the same wrong style. Refactored them in a second commit.

backend/handlers/handlers.go
```diff
@@ -690,6 +691,43 @@ func (handlers *Handlers) getRatesHandler(_ *http.Request) (interface{}, error)
...
+	btcCoin, err := handlers.backend.Coin(coinpkg.CodeBTC)
+	if err != nil {
+		logrus.Error(err.Error())
```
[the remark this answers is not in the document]
Beerosagos: Thanks, also improved logging in `getConvertToPlainFiatHandler` while I was there.

backend/handlers/handlers.go
```diff
@@ -690,6 +691,43 @@ func (handlers *Handlers) getRatesHandler(_ *http.Request) (interface{}, error)
...
+		return map[string]interface{}{
+			"success": false,
+		}, nil
+	}
+
+	var coinAmount coin.Amount
+	if btcCoin.GetFormatUnit(false) == string(coinpkg.BtcUnitDefault) {
```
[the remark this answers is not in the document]
Beerosagos: Sorry, I really can't remember what I was trying to do here. Must have been tired when I wrote this :sweat_smile:

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -510,13 +510,31 @@ class Send extends Component<Props, State> {
       sendAll: false,
       fiatAmount: ''
     } as Pick<State, keyof State>;
-    if (amount) {
-      updateState['amount'] = amount;
+
+    const coinCode = this.getAccount()!.coinCode;
+    if (amount && coinCode === 'btc') {
```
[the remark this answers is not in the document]
Beerosagos: Good catch, thanks

frontends/web/src/routes/account/send/send.tsx
Beerosagos: Right! I was a bit afraid of inadvertently modifying previous behavior, but looks much nicer now :+1:

backend/handlers/handlers.go
Beerosagos: Put here since there was already `/coins/btc/set-unit` which seems related to the same scope. Could also move them both though..

backend/handlers/handlers.go
Beerosagos: I kind of like the compactness of not declaring the result struct, but no strong opinions about that. I followed your suggestion

Beerosagos:
Thanks @benma for the review, now should be better PTAL :pray:

backend/handlers/handlers.go
benma: I actually wrote this comment and wanted to delete it. this is the right place for this after all, as it is not account-specific!

benma (approved):
utACK

PR #1979, opened 2023-02-28
frontend: fix "back btn" behaviour on country selector (buy page).

shonsirsha:
Back button in the country selector (`/buy/exchange/`) page doesn't work as expected when there's only one account. 

This is due to the back button is trying to go back to `/buy/info`, but that page redirects the user back to the `/buy/exchange/` page again. Therefore no "going back" action is actually happening for the user.

We fix this by going back 2 pages when there's only one account, and only going back once when there's more than one account. `useNavigate()` is utilised for this approach.

### DIFF
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -15,9 +15,10 @@
  */
 import 'flag-icons';
 import { useState, useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
 import i18n from '../../i18n/i18n';
 import { useTranslation } from 'react-i18next';
-import { Button, ButtonLink } from '../../components/forms';
+import { Button } from '../../components/forms';
 import * as exchangesAPI from '../../api/exchanges';
 import { IAccount } from '../../api/account';
 import { Header } from '../../components/layout';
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -175,6 +177,17 @@ export const Exchange = ({ code, accounts }: TProps) => {
   const cardFee = infoFeesDetail && infoFeesDetail.find(feeDetail => feeDetail.payment === 'card')?.fee;
   const bankTransferFee = infoFeesDetail && infoFeesDetail.find(feeDetail => feeDetail.payment === 'bank-transfer')?.fee;
 
+  const handleGoBack = () => {
+    if (accounts.length > 1) {
```
thisconnect, line 181: somehow it didn't back 🤔 , looking at this, I  think `accounts.length === 1` is what was intended.

thisconnect (changes requested):

frontends/web/src/routes/buy/exchange.tsx
shonsirsha: > somehow it didn't back 

Hm... 🤔  is it possible that for some reason you didn't have any history in that tab (maybe you went directly to the URL)? In real life scenario I assume we wouldn't have to worry about this since the user is in the app (and URL isn't accessible).

Works as intended for me:

https://user-images.githubusercontent.com/28676406/222038608-dfe1db7c-cbb9-4613-8118-b33071ea7c14.mov

> looking at this, I think accounts.length === 1 is what was intended.

If I understood it correctly, when `accounts.length === 1` , we should actually go back twice. Because going back once will bring us to `buy/info` which will auto redirect us again to the selector page (when there's only one account). This case no "back" action will be visible. 

This's essentially the bug we're trying to fix.

thisconnect: > is it possible that for some reason you didn't have any history in that tab (maybe you went directly to the URL)? 

yep maybe let me test again...

thisconnect (approved):
LGTM

PR #1981, opened 2023-03-01
frontend: convert `createbackup` into a functional component 

shonsirsha:
To follow the latest conventions and best practices of React, we'd like to convert 
our class components into functional ones.

Here, we're converting `createbackup.tsx` into a functional component.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/createbackup.tsx
+++ b/frontends/web/src/routes/device/bitbox02/createbackup.tsx
@@ -15,83 +15,66 @@
  * limitations under the License.
  */
 
-import { Component } from 'react';
-import { createBackup } from '../../../api/bitbox02';
-import { translate, TranslateProps } from '../../../decorators/translate';
-import { apiPost } from '../../../utils/request';
+import { useState } from 'react';
+import { checkBackup, createBackup as createBackupAPI } from '../../../api/bitbox02';
 import { alertUser } from '../../../components/alert/Alert';
 import { confirmation } from '../../../components/confirm/Confirm';
 import { Button } from '../../../components/forms';
 import { WaitDialog } from '../../../components/wait-dialog/wait-dialog';
+import { useTranslation } from 'react-i18next';
 
-interface CreateProps {
+type TProps = {
     deviceID: string;
 }
 
-type Props = CreateProps & TranslateProps;
+export const Create = ({ deviceID }: TProps) => {
+  const [creatingBackup, setCreatingBackup] = useState(false);
+  const [disabled, setDisabled] = useState(false);
+  const { t } = useTranslation();
 
-interface State {
-    creatingBackup: boolean;
-    disabled: boolean;
-}
-
-class Create extends Component<Props, State> {
-  public readonly state: State = {
-    creatingBackup: false,
-    disabled: false,
-  };
-
-  private createBackup = () => {
-    this.setState({ creatingBackup: true });
-    createBackup(this.props.deviceID)
-      .then(() => this.setState({ creatingBackup: false, disabled: false }))
+  const createBackup = () => {
+    setCreatingBackup(true);
+    createBackupAPI(deviceID)
+      .then(() => {
+        setCreatingBackup(false);
+        setDisabled(false);
+      })
       .catch(() => {
-        alertUser(this.props.t('backup.create.fail'));
+        setCreatingBackup(false);
+        setDisabled(false);
+        alertUser(t('backup.create.fail'));
       });
   };
 
-  private maybeCreateBackup = () => {
-    this.setState({ disabled: true });
-    apiPost('devices/bitbox02/' + this.props.deviceID + '/backups/check', {
-      silent: true,
-    }).then(check => {
-      if (check.success) {
-        confirmation(this.props.t('backup.create.alreadyExists'), result => {
-          if (result) {
-            this.createBackup();
-          } else {
-            this.setState({ disabled: false });
-          }
-        });
-        return;
-      }
-      this.createBackup();
-    });
+  const maybeCreateBackup = async () => {
+    setDisabled(true);
+    const check = await checkBackup(deviceID, true);
+    if (check) {
+      confirmation(t('backup.create.alreadyExists'), result => {
+        if (result) {
+          createBackup();
+        } else {
+          setDisabled(false);
+        }
+      });
+      return;
+    }
+    createBackup();
   };
 
-  public render() {
-    const { t } = this.props;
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/device/bitbox02/createbackup.tsx
```diff
@@ -15,83 +15,70 @@
...
+        alertUser(t('backup.create.fail'));
       });
   };
 
-  private maybeCreateBackup = () => {
-    this.setState({ disabled: true });
-    apiPost('devices/bitbox02/' + this.props.deviceID + '/backups/check', {
+  const maybeCreateBackup = () => {
+    setDisabled(true);
+    apiPost(`devices/bitbox02/${deviceID}/backups/check`, {
```
thisconnect, line 52: can you use `checkBackup` from `api/bitbox02`?

checkBackup promise is a bit different than the others, it resolves on success or rejects if there was no success.

```js

checkBackup(deviceId, true)
  .then((backupID) => console.log('success!', backupID))
  .catch(() => console.error('backup check not ok))
```

or with async await https://github.com/digitalbitbox/bitbox-wallet-app/blob/1fd8e8c63788ca9db46fa87b3a58618a699d8f4b/frontends/web/src/routes/device/bitbox02/checkbackup.tsx#L40-L41

using then/catch seemed nice at the time, but one problem is that errors in catch can't be typed, as there may be other unexpected errors. this is why I continued rewriting api calls without throwing if success is fails.

frontends/web/src/routes/device/bitbox02/createbackup.tsx
```diff
@@ -15,83 +15,70 @@
...
+    setCreatingBackup(true);
+    createBackupAPI(deviceID)
+      .then(() => {
+        setCreatingBackup(false);
+        setDisabled(false);
+      })
       .catch(() => {
-        alertUser(this.props.t('backup.create.fail'));
+        setCreatingBackup(false);
+        setDisabled(false);
```
thisconnect, line 45: this is different than before. Makes sense to me but not sure how to test, @Beerosagos can you test this?

thisconnect (changes requested):

frontends/web/src/routes/device/bitbox02/createbackup.tsx
shonsirsha: Yeah, this was actually a bug before. To test this, just click on "create backup"and cancel from the bb02 device. This should show a dismissable modal with "'backup.create.fail'" message. In master, there's an undismissable modal on top of it.

frontends/web/src/routes/device/bitbox02/createbackup.tsx
shonsirsha: Ah I see. Nice that we have `checkBackup`. Will do.

frontends/web/src/routes/device/bitbox02/createbackup.tsx
```diff
@@ -15,83 +15,67 @@
...
+    const check = await checkBackup(deviceID, true);
+    if (check) {
+      confirmation(t('backup.create.alreadyExists'), result => {
+        if (result) {
+          createBackup();
+        } else {
+          setDisabled(false);
+        }
+      });
+      return;
```
shonsirsha, line 62: Now using `checkBackup`. Just one question, I removed the last `createBackup()`. I wonder why we had it there? Seems like it doesn't have any effect..?

frontends/web/src/routes/device/bitbox02/createbackup.tsx
```diff
@@ -15,83 +15,67 @@
...
-        confirmation(this.props.t('backup.create.alreadyExists'), result => {
-          if (result) {
-            this.createBackup();
-          } else {
-            this.setState({ disabled: false });
-          }
-        });
-        return;
-      }
-      this.createBackup();
```
shonsirsha, line 68: This is the one I removed.

thisconnect (approved):
untested LGTM

frontends/web/src/routes/device/bitbox02/createbackup.tsx
```diff
@@ -15,83 +15,66 @@
...
+      confirmation(t('backup.create.alreadyExists'), result => {
+        if (result) {
+          createBackup();
+        } else {
+          setDisabled(false);
+        }
+      });
+      return;
+    }
+    createBackup();
```
thisconnect, line 62: Testing create-backup failed on empty sdCard / nothing happened in the UI, got uncaught promise error.

This is another reason why checkBackup shouldn't throw an error if it does not find a backup. One solution here for now could be to wrap the whole thing in try/catch and do the createBackup outside, or add .catch(console.error);

thisconnect: sorry i guess this wouldn't have happened if I didn't advice you to use checkBackup from api/bitbox02

thisconnect (changes requested):
unfortunatelly then checkBackup from api/bitbox02 throws an error if no backup is found. see other comments

frontends/web/src/routes/device/bitbox02/createbackup.tsx
thisconnect:
```diff
diff --git a/frontends/web/src/routes/device/bitbox02/createbackup.tsx b/frontends/web/src/routes/device/bitbox02/createbackup.tsx
index 11edd5d9d..99936b4dd 100644
--- a/frontends/web/src/routes/device/bitbox02/createbackup.tsx
+++ b/frontends/web/src/routes/device/bitbox02/createbackup.tsx
@@ -48,7 +48,7 @@ export const Create = ({ deviceID }: TProps) => {
 
   const maybeCreateBackup = async () => {
     setDisabled(true);
-    const check = await checkBackup(deviceID, true);
+    const check = await checkBackup(deviceID, true).catch(console.error);
     if (check) {
       confirmation(t('backup.create.alreadyExists'), result => {
         if (result) {
@@ -58,8 +58,9 @@ export const Create = ({ deviceID }: TProps) => {
         }
       });
       return;
+    } else {
+      createBackup();
     }
-    createBackup();
   };
 
   return (
```

thisconnect: opened a new PR with this change: https://github.com/digitalbitbox/bitbox-wallet-app/pull/1985

thisconnect:
merged in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1985

PR #1982, opened 2023-03-01
frontend: sdcard check should show microsd missing dialog

thisconnect:
It seems that the dialog component does not render if open prop is already true during component mount.

Seems like a regression in: a229726d8cd5424d502aa9610bad5962088ce7bb

Changed to stateless View/dialog component fixed the issue.


thisconnect:
closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/1983

PR #1983, opened 2023-03-01
frontend: fix dialog open on mount

thisconnect:
Currently dialog only checks on component did change if the prop is open, but ignores if the open prop is already set to true when the component mounts.

This was the reason why sdcard warning did not show when microSD card was missing on the manage-backups view.

Added on component did mount method that checks if the open prop is already set to true and if so calls this activate.

Changed the existing on props did update and removed setState, as this.activate will do setState anyway.

### DIFF
--- a/frontends/web/src/components/dialog/dialog.tsx
+++ b/frontends/web/src/components/dialog/dialog.tsx
@@ -49,13 +49,17 @@ class Dialog extends Component<Props, State> {
     renderDialog: false,
   };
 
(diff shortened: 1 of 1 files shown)


shonsirsha (approved):
tested LGTM.

This was a good catch as we never checked the `open` prop during mount, but only when it changes. This PR adds this check. 

In turn, this fixes a bug where the `open` prop is `true` during mount, such as the case of microSD card warning mentioned in this PR desc.

PR #1984, opened 2023-03-01
android: fix unwanted blocking of iframe links

Beerosagos:
Iframes integrated in the app have some external links that the user could click to open in an external browser. Before this, the view url were used to allow the links clicked in views inside the '/buy/ route. For unknown reasons this recently stopped working (the retrieved url was "about:blank"). Since it was not possible to retrive the view url in other ways, this just creates a list of allowed domains that can be pointed to.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -38,8 +38,10 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.regex.Pattern;
 
 import goserver.Goserver;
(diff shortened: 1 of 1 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -173,15 +175,22 @@ public WebResourceResponse shouldInterceptRequest(final WebView view, WebResourc
...
                 try {
-                    // Allow opening in external browser instead, for links clicked in an onramp widget.
-                    Pattern pattern = Pattern.compile("^file:///buy/.*$");
-                    if (pattern.matcher(view.getUrl()).matches()) {
-                        Util.systemOpen(getApplication(), url);
-                        return true;
+                    // Allow opening in external browser instead, for listed domains.
+                    List<Pattern> patterns = new ArrayList<>();
+                    patterns.add(Pattern.compile(".*pocketbitcoin\\.com$"));
+                    patterns.add(Pattern.compile(".*moonpay\\.com$"));
```
benma, line 187: These would also match "fakepocketbitcoin.com". Didn't check what getHost() contains, but maybe sth like "^pocketbitcoin\\.com$" and ".+\\.pocketbitcoin\\.com$" would be better?

I am also trying to remember the purpose of this whitelist here. `systemOpen` already has a whitelist, so I wonder if we could just skip any checks here (and in Qt's main.cpp). Wdyt?

Beerosagos: good point about the regex! Will fix it. Are you sure `systemOpen` has a white list? afaict this is needed because the URLs checked here do not pass through the whitelist in the backend and are directly opened (using an external browser instead of the current view, to avoid loosing the current workflow). I think we need this to support and verify links in the iframes.

benma: I think the reason for the above check was that rogue JS could not redirect anywhere (except on the buy page). The check you have now looks indeed redundant to the whitelist already in systeOpen. After considering it more, I think it's fine to skip any checks here.

benma: Nevermind, @Beerosagos made me aware that this systemOpen is not the backend's systemOpen, so this fix is appropriate!

benma (approved):

PR #1990, opened 2023-03-06
frontend: convert `CopyableInput` (`Copy.tsx`) into a functional component

shonsirsha:

Here, we're converting `copy.tsx` into a functional component.


thisconnect (approved):
tested LGTM with small change request

PR #1992, opened 2023-03-06
frontend: refactor api calls should not throw on failed result

thisconnect:
api calls that throw if success is false, have few disadvantages:

- rejecting the promise on failed success mixes failed response with real errors, so catch gets either an error or a failed response
- data passed to promise reject cannot be typed
- throwing error in the api call must be caught, either try/catch or .catch else the whole script just stops with uncaught promise error

Refcatored api calls and removed the success handling, always resolving the promise now with response data.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -325,39 +325,46 @@ class BitBox02 extends Component<Props, State> {
   };
 
   private setDeviceName = (event: FormEvent) => {
+    const { deviceID, t } = this.props;
     event.preventDefault();
     this.setState({
-      waitDialog: { title: this.props.t('bitbox02Interact.confirmName') }
-    }, () => {
-      setDeviceName(this.props.deviceID, this.state.deviceName)
-        .then(() => {
-          this.setState(
-            { waitDialog: undefined },
-            () => this.setPassword(),
-          );
-        })
-        .catch(result => {
-          if (result.message) {
-            alertUser(result.message, {
-              asDialog: false,
-              callback: () => this.setState({ waitDialog: undefined }),
-            });
-          }
-        });
+      waitDialog: { title: t('bitbox02Interact.confirmName') }
+    }, async () => {
+      try {
+        const result = await setDeviceName(deviceID, this.state.deviceName);
+        if (!result.success) {
+          alertUser(result.message || t('genericError'), {
+            asDialog: false,
+            callback: () => this.setState({ waitDialog: undefined }),
+          });
+          return;
+        }
+        this.setState(
+          { waitDialog: undefined },
+          () => this.setPassword(),
+        );
+      } catch (error) {
+        console.error(error);
+      }
     });
   };
 
-  private restoreFromMnemonic = () => {
+  private restoreFromMnemonic = async () => {
     this.setState({ waitDialog: {
       title: this.props.t('bitbox02Interact.followInstructionsMnemonicTitle'),
       text: this.props.t('bitbox02Interact.followInstructionsMnemonic'),
     } });
-    restoreFromMnemonic(this.props.deviceID)
-      .then(() => this.setState({ appStatus: 'restoreFromMnemonic' }))
-      .catch(() => {
+    try {
+      const { success } = await restoreFromMnemonic(this.props.deviceID);
+      if (!success) {
         alertUser(this.props.t('bitbox02Wizard.restoreFromMnemonic.failed'), { asDialog: false });
-      })
-      .finally(() => this.setState({ waitDialog: undefined }));
+      } else {
+        this.setState({ appStatus: 'restoreFromMnemonic' });
+      }
+      this.setState({ waitDialog: undefined });
+    } catch (error) {
+      console.error(error);
+    }
   };
 
   private handleDisclaimerCheck = (event: React.SyntheticEvent) => {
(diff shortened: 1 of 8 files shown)


frontends/web/src/routes/device/bitbox02/setdevicename.tsx
```diff
@@ -45,9 +45,15 @@ export const SetDeviceName = ({
...
+      if (!setNameResult.success) {
+        throw new Error(setNameResult.message);
+      }
+      const deviceInfoResult = await getDeviceInfo(deviceID);
+      if (!deviceInfoResult.success) {
+        throw new Error(deviceInfoResult.message);
+      }
+      setCurrentName(deviceInfoResult.deviceInfo.name);
     } catch (e) {
       alertUser('Device name could not be set');
```
thisconnect, line 58: this should be translated but better in a different PR.

frontends/web/src/routes/device/bitbox02/setdevicename.tsx
```diff
@@ -45,9 +45,15 @@ export const SetDeviceName = ({
   const updateName = async () => {
     setInProgress(true);
     try {
-      await setDeviceName(deviceID, name);
```
thisconnect, line 48: before (here) if it fails OR throws an error, it would catch and show alertuser error.

now I added success checks and manually throw new Error so that alertUser does not need to be written in 3 places, but seems a bit ugly.

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -57,39 +57,45 @@ class Passphrase extends Component<Props, State> {
...
+      }));
+      return;
+    }
+    const deviceInfoResult = await getDeviceInfo(deviceID);
+    if (deviceInfoResult.success) {
+      this.setState({
+        passphraseEnabled: deviceInfoResult.deviceInfo.mnemonicPassphraseEnabled,
+        status: 'success',
       });
+    }
```
thisconnect, line 98: funny that alertUser still shows even if before route is called.. it's because the weird global alertUser.

having alertUser before route would at least make the intention clear, but it would still be a bit confusing I guess. keeping unless somebody has a strong opinion.

thisconnect: In theory api calls such as setMnemonicPassphraseEnabled should never error, but if they would.. for example during development, it would only say uncaught promise rejection or similar... 
is better to always have in a try/catch and console.error ?

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -325,39 +325,38 @@ class BitBox02 extends Component<Props, State> {
...
-            alertUser(result.message, {
-              asDialog: false,
-              callback: () => this.setState({ waitDialog: undefined }),
-            });
-          }
+      waitDialog: { title: t('bitbox02Interact.confirmName') }
+    }, async () => {
+      const result = await setDeviceName(deviceID, this.state.deviceName);
+      if (!result.success) {
+        alertUser(result.message || t('genericError'), {
```
thisconnect, line 335: added genericError as fallback since typescript wasn't happy, as result.message is optional.

before it checked if result.message is truthy, but if missing it didn't show any error…

not tested yet

shonsirsha: tested and works fine.

frontends/web/src/routes/device/bitbox02/setdevicename.tsx
shonsirsha: Looks okay to me, but it'll still always show 'Device name could not be set' regardless of the error being thrown, right?

frontends/web/src/routes/device/bitbox02/passphrase.tsx
shonsirsha: > is better to always have in a try/catch and console.error ?

Tbh I have no strong opinion on this. Seems like our codebase isn't so consistent with it anyway (some are using try/catch, some dont). But catching the error and logging it would be okay, at least that way during development it's easier to see what error is occurring. Although not sure if that's really worth it.

frontends/web/src/routes/device/bitbox02/createbackup.tsx
```diff
@@ -35,15 +35,14 @@ export const Create = ({ deviceID }: TProps) => {
...
+        if (!result.success) {
+          alertUser(t('backup.create.fail'));
+        }
       })
-      .catch(() => {
-        setCreatingBackup(false);
-        setDisabled(false);
-        alertUser(t('backup.create.fail'));
-      });
+      .catch(console.error);
```
shonsirsha, line 45: Just a general question for all the changes, what happens if the API call resolves in an error (in this case the createBackupAPI)? Will it catch and `console.error` only? So the user won't be notified? Orr...?

shonsirsha: understood, thanks.

thisconnect:
@shonsirsha PTAL

shonsirsha (approved):
lgtm!

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -325,39 +325,46 @@ class BitBox02 extends Component<Props, State> {
...
         alertUser(this.props.t('bitbox02Wizard.restoreFromMnemonic.failed'), { asDialog: false });
-      })
-      .finally(() => this.setState({ waitDialog: undefined }));
+      } else {
+        this.setState({ appStatus: 'restoreFromMnemonic' });
+      }
+      this.setState({ waitDialog: undefined });
+    } catch (error) {
+      console.error(error);
+    }
```
benma, line 367: this is never executed as there is no reject() anymore, or not? here and in the other cases in this PR.

PR #1993, opened 2023-03-06
backend/chart: add latest rates timestamps in chart data

Beerosagos:
If one of the coin/rates pairs is older than 2hs the backend set the `chartIsUpToDate` flag to false inside the `account-summary` return struct. This adds a map to the struct that list the latest update timestamp for each coin. This will allow the frontend to show a meaningful message in the future instead of just keep loading the chart.

### DIFF
--- a/backend/chart.go
+++ b/backend/chart.go
@@ -72,6 +72,8 @@ type Chart struct {
 	FormattedTotal string `json:"formattedChartTotal"`
 	// Only valid if DataMissing is false
 	IsUpToDate bool `json:"chartIsUpToDate"`
+	// Latest rate timestamp available among all enabled coins.
+	LastTimestamp int64 `json:"lastTimestamp"`
 }
 
 func (backend *Backend) addChartData(
@@ -125,6 +127,7 @@ func (backend *Backend) ChartData() (*Chart, error) {
 		backend.log.Info("ChartDataMissing, until is zero")
 	}
 	isUpToDate := time.Since(until) < 2*time.Hour
+	lastTimestamp := until.UnixMilli()
 
 	formatBtcAsSat := util.FormatBtcAsSat(backend.Config().AppConfig().Backend.BtcUnit)
 
@@ -316,5 +319,6 @@ func (backend *Backend) ChartData() (*Chart, error) {
 		Total:          chartTotal,
 		FormattedTotal: formattedChartTotal,
 		IsUpToDate:     isUpToDate,
+		LastTimestamp:  lastTimestamp,
 	}, nil
 }
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -446,6 +446,7 @@
   },
   "chart": {
     "dataMissing": "Gathering historical data... stay tuned.",
+    "dataOldTimestamp": "Historical exchange rates updating. The chart is not displaying data after {{time}}.",
     "dataUpdating": "updating data…",
     "filter": {
       "all": "All",
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -75,6 +75,7 @@ class Chart extends Component<Props, State> {
       chartTotal: null,
       formattedChartTotal: null,
       chartIsUpToDate: false,
+      lastTimestamp: 0,
     }
   };
 
@@ -139,9 +140,9 @@ class Chart extends Component<Props, State> {
   };
 
   private createChart = () => {
-    const { data: { chartIsUpToDate, chartDataMissing } } = this.props;
+    const { data: { chartDataMissing } } = this.props;
     const darkmode = getDarkmode();
-    if (this.ref.current && this.hasData() && (chartIsUpToDate && !chartDataMissing)) {
+    if (this.ref.current && this.hasData() && !chartDataMissing) {
       if (!this.chart) {
         const chartWidth = !this.state.isMobile ? this.ref.current.offsetWidth : document.body.clientWidth;
         const chartHeight = !this.state.isMobile ? this.height : this.mobileHeight;
@@ -411,8 +412,6 @@ class Chart extends Component<Props, State> {
     });
   };
 
-
-
   private renderDate = (date: number) => {
     return new Date(date).toLocaleString(
       this.props.i18n.language,
@@ -432,7 +431,7 @@ class Chart extends Component<Props, State> {
     const {
       t,
       data: {
-        chartDataDaily,
+        lastTimestamp,
         chartDataMissing,
         chartFiat,
         chartIsUpToDate,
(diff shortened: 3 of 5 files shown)


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -103,6 +103,9 @@ class AccountsSummary extends Component<Props, State> {
 
   private getAccountSummary = () => {
     return accountApi.getSummary().then(data => {
+      for (let c in data.ratesTimestamps) {
+        console.log(c + ': ' + new Date(data.ratesTimestamps[c as accountApi.CoinCode]).toLocaleString());
```
Beerosagos, line 107: This is just to show how to handle the data, will be probably removed when the frontend starts using the timestamps to show meaningful messages.

backend/rates/history.go
```diff
@@ -215,18 +215,21 @@ func (updater *RateUpdater) HistoryEarliestTimestamp(coin, fiat string) time.Tim
 
 // HistoryLatestTimestampAll returns the latest timestamp for which there an exchange rates is
 // available for all coins. In other words: the earliest of all latest timestamps.
```
benma, line 217: Please extend the doc about the 2nd return value

benma (changes requested):
What kind of meaningful message would be shown to the user where the data added in this PR is needed? I think it would be helpful to finish this PR also from the frontend perspective (at least the wording, UX/UI/style can added later) so this becomes clear.

thisconnect:
just as examples

account summary instead of "updating" it could show: 
  - "Yen/Litecoin conversion rate is older than expected, last updated 3 hours ago."
  - "The following conversion rates are older than 2 hours: USD/Bitcoin, USD/Litecoin, USD/Ethereum"
  - "Looks like there is a problem with FIAT, the following conversion rates couldn't be fetched: USD/Bitcoin, USD/Litecoin, USD/Ethereum"

account overview and send screen, doesn't need to show which coin as the account already is using 1 coin:
- USD conversion rates is older than expected, last updated 2 hours ago
- The conversion rate for USD couldn't be fetched

cc @jadzeidan 

> What kind of meaningful message would be shown to the user where the data added in this PR is needed?

benma:
@thisconnect there is a misunderstanding - the latest rate is either present or absent, but this PR does not add any info about this that could be turned into an error msg. The latest rate is independent of the historic rates for the chart and is usually present.

If the historic rates for the chart are missing the last 2h then this PR adds info which pairs have this problem. The latest rate could still be available (historic data and the current/latest data are different endpoints).

> Yen/Litecoin conversion rate is older than expected, last updated 3 hours ago."
> "The following conversion rates are older than 2 hours: USD/Bitcoin, USD/Litecoin, USD/Ethereum"

In light of the above, this could be misleading, as the total above the chart and summary tables below the chart and all rates in the account/send pages are normally available even if the historic data is lagging by 2h.

A message like "Historic data is syncing..." in this case probably makes more sense, but don't we already show something like this?

thisconnect:
I was just brainstorming what would make sense.

> In light of the above, this could be misleading, as the total above the chart and summary tables below the chart and all rates in the account/send pages are normally available even if the historic data is lagging by 2h.

Why misleading? if latest is available but lagging 3 month behind the user should be informed. probably out of scope for this PR.

benma:
> If latest means lagging 3 month behind the user should be informed. but that is probably out of scope for this PR.

As I said  above, the latest will not lag behind, it is pretty much always available. The lag is for historic data only, but that doesn't affect the "latest" (i.e. the total above the chart, etc).

jadzeidan:
I'm a bit confused how and when this warning is displayed to the user. @thisconnect maybe you can make a screen shot?

IIUC, the warning only shows up if some data is more than 2 hours old, correct?

Beerosagos:
Rebased to solve conflicts

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -411,7 +411,29 @@ class Chart extends Component<Props, State> {
...
+          outdatedCoins.at(-1)?.toUpperCase()
+        ].join(' ')
+        : outdatedCoins[0].toUpperCase();
+
+    const oldestTimestamp = Math.min(...Object.values(ratesTimestamps));
+    const hoursAgo = Math.ceil((oldestTimestamp - Date.now()) / 3600000);
+
+    return t('chart.dataOldTimestamps', {
+      name: coinOrCoinsWithAnd,
+      timeAgo: this.relativeTime.format(hoursAgo, 'hours'),
```
thisconnect, line 434: FYI, timeAgo will now always be translated to whatever `this.props.i18n.language` is set, regarding of if there is `chart.dataOldTimestamps` translation or not. so ideally we have `chart.dataOldTimestamps` translation in every lang.

backend/chart.go
```diff
@@ -119,12 +121,21 @@ func (backend *Backend) ChartData() (*Chart, error) {
...
 	}
-	isUpToDate := time.Since(until) < 2*time.Hour
+	upToDateLimit := 2 * time.Hour
+	isUpToDate := time.Since(until) < upToDateLimit
+
+	unixTimestamps := make(map[string]int64)
+	for coin, timestamp := range timestamps {
+		if time.Since(timestamp) >= upToDateLimit {
+			// js Date is based on milliseconds.
+			unixTimestamps[coin] = timestamp.UnixMilli()
```
benma, line 136: if i read the frontend code correctly you only really display `until` there (smallest of the timestamps), so making this `unixTimestamps` map is not needed?

imo it would be much clearer to pass a `outdated { CoinsNames []string, timestamp time.Time }` struct.

frontends/web/src/locales/en/app.json
```diff
@@ -445,6 +445,7 @@
   },
   "chart": {
     "dataMissing": "Gathering historical data... stay tuned.",
+    "dataOldTimestamps": "Exchange rates for {{name}} are older than expected. Last updated {{timeAgo}}",
```
benma, line 448: "older than expected" is imo confusing to a user. what is expected?

why does the user even need to see which coin rates are older than 2h? i don't see how that is helpful. the chart stops at the earliest such time anyway, so we might as well say `"Exchange rates are out of date. Last updated {{timeAgo}"`.

frontends/web/src/locales/en/app.json
```diff
@@ -577,6 +578,7 @@
     "appVersion": "App version:"
   },
   "generic": {
+    "and": "and",
```
benma, line 581: such tricks usually don't work out in my experience. there is a good chance some languages don't function this way. For example, in Japanese there is probably a comma between all coins (no "and").

If you have to include the coins in the list (see my other comment), it is safer to do sth like  "The exchange rates of the following coins are out of date: A, B, C, D".

benma (changes requested):

backend/chart.go
Beerosagos: Yes, because we initially decided to display the last rates available for each coin/fiat pair. This is not used anymore now and can be removed if we don't plan to add this information anymore.

Beerosagos:
Code updated following the simpler approach proposed by @benma. PTAL :pray: 
cc @thisconnect

frontends/web/src/locales/en/app.json
```diff
@@ -446,6 +446,7 @@
   },
   "chart": {
     "dataMissing": "Gathering historical data... stay tuned.",
+    "dataOldTimestamp": "Historical exchange rates updating. The chart is not displaying data after {{time}}",
```
benma, line 449: period missing at the end?

benma (approved):
utACK, much better and less complicated now imo.

see nit

thisconnect:
seems macos doesn't find node-v14.15.4

PR #1995, opened 2023-03-07
frontend: cleanup gitignore

thisconnect:
- removed ignoring *.css.d.ts files, before migrating to React typings-for-css-modules-loader was used that generated these files, but now with CSS-Modules by react-scripts, this is not used anymore
- removed commented out ignore-unfinished-locales as this is in the top level .gitignore


Beerosagos (approved):
LGTM!

PR #1998, opened 2023-03-07
frontend: fix buy CTA redirect & balance loading (account dropdown)

shonsirsha:
1. Buy CTA for an empty account in the `/account/${code}` page redirects user straight into the country selector (`/buy/exchange/${code}`). We'd like it to redirect to the account selector first (`/buy/info/${code}`) so user will be able to see the account they selected. The selected account is displayed in a dropdown.

2. The said dropdown is also supposed to  display the selected account's  balance which gets fetched asyncly. The balance was fetched asyncly, but it's not yet displayed. This PR also fixes this bug; to actually display the fetched balance in the dropdown.

Preview for an empty account:

https://user-images.githubusercontent.com/28676406/223382842-aae6e8f5-e91d-4f19-95dd-5b8aa3b596ca.mov

Preview for a non-empty account:

https://user-images.githubusercontent.com/28676406/223385266-044cb8c2-7c98-45bb-b884-dd58944fb961.mov

### DIFF
--- a/frontends/web/src/components/accountselector/accountselector.tsx
+++ b/frontends/web/src/components/accountselector/accountselector.tsx
@@ -14,12 +14,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+import { FunctionComponent, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { AccountCode, IAccount } from '../../api/account';
 import { Button } from '../forms';
 import Select, { components, SingleValueProps, OptionProps, SingleValue, DropdownIndicatorProps } from 'react-select';
-import { FunctionComponent } from 'react';
 import Logo from '../icon/logo';
 import styles from './accountselector.module.css';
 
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/accountselector/accountselector.tsx
```diff
@@ -80,21 +79,24 @@ const DropdownIndicator: FunctionComponent<DropdownIndicatorProps<TOption>> = (p
...
-            value: 'choose',
-            disabled: true
-          } :
-          options.find(opt => opt.value === selected)
-        }
+        value={selected === '' ? {
+          label: t('buy.info.selectLabel'),
+          value: 'choose',
+          disabled: true
+        } : selectedAccount}
```
thisconnect, line 99: what is the difference from using defaultValue vs. value here?

thisconnect:
Nice, I tested a bit and LGTM... but  could you split those into 2 commits (keep in 1 PR)?

reason atomic commits help if for example we want to revert 1 of those changes.

frontends/web/src/components/accountselector/accountselector.tsx
shonsirsha: To put it simply, `defaultValue` is suitable for uncontrolled components. Whereas `value` is for controlled ones. Since we're using the `selectedAccount` state now, it's a controlled component. Thus we're now utilising the `value` prop.

`selectedAccount` state is used so we can get the updated balance value once it gets fetched asyncly. This previously wasn't possible, since we're only using `options.find(opt => opt.value === selected)` for the selected value and this doesn't get updated when `options` has the balance after it was fetched.

shonsirsha:
Commit has been split, thanks for the review! @thisconnect

thisconnect (approved):
LGTM

PR #2000, opened 2023-03-07
frontend/Qt: enable autoscaling for high density screens

Beerosagos:
Autoscaling was only enabled for windows, because with Qt < v5.14 it didn't look nice on Linux and Mac OS. Now that we are compiling with Qt v5.15 we can enable it on all OSs.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -165,10 +165,8 @@ public:
 
 int main(int argc, char *argv[])
 {
-// Enable auto HiDPI scaling on Windows only for now.
-#if defined(_WIN32) && QT_VERSION >= QT_VERSION_CHECK(5,6,0)
+    // Enable auto HiDPI scaling to correctly manage scale factor on bigger screens
     QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
-#endif
 
 // QT configuration parameters which change the attack surface for memory
 // corruption vulnerabilities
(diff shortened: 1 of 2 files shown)


frontends/qt/main.cpp
```diff
@@ -165,8 +165,9 @@ class WebEngineView : public QWebEngineView {
 
 int main(int argc, char *argv[])
 {
-// Enable auto HiDPI scaling on Windows only for now.
-#if defined(_WIN32) && QT_VERSION >= QT_VERSION_CHECK(5,6,0)
+// Enable auto HiDPI scaling if Qt>5.14. Otherwise do it on Windows only because it doesn't
+// look nice on Linux and MacOS.
+#if QT_VERSION >= QT_VERSION_CHECK(5,14,0) || (defined(_WIN32) && QT_VERSION >= QT_VERSION_CHECK(5,6,0))
```
Beerosagos, line 170: Not sure if it is worth making a distinction between Qt versions. wdyt?

benma: You can drop all these checks as we are on Qt 5.15 at the moment and don't support compiling this app on older Qt versions.

benma:
Please add a changelog entry

benma (changes requested):

Beerosagos:
Should be good now! PTAL

PR #2001, opened 2023-03-08
frontend: set dark mode colour scheme for account dropdown/selector

shonsirsha:
Hardcoded CSS values to display the account dropdown in dark mode. We added some new selectors in the css file, where previously weren't needed.

This is because most of react-select's default styles suit our light mode. However, it's not the case for dark mode. Thus, we need to grab ahold of more selectors to be customised for them to work with our dark mode (hence adding more selectors in the css).

Preview:
Screenshot 2023-03-13 at 09 08 05

Screenshot 2023-03-13 at 09 08 02

### DIFF
--- a/frontends/web/src/components/accountselector/accountselector.module.css
+++ b/frontends/web/src/components/accountselector/accountselector.module.css
@@ -1,6 +1,6 @@
 .balance, .balanceSingleValue{
     margin-left: auto;
-    color: var(--color-gray);
+    color: var(--color-secondary);
 }
 
 .balanceSingleValue {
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/accountselector/accountselector.module.css
```diff
@@ -20,11 +20,33 @@
...
+    background-color: var(--background-dark);
+}
+
+.select :global(.react-select__option):hover, 
+.select :global(.react-select__option--is-selected) {
+    background-color: var(--background);
+}
+
+.select :global(.react-select__option--is-disabled) span {
+    color: var(--color-gray);
```
thisconnect, line 37: please use `--color-secondary`

frontends/web/src/components/accountselector/accountselector.module.css
```diff
@@ -45,9 +67,10 @@
 }
 
 
-.valueContainer{
-    display: flex;
+.valueContainer {
     align-items: center;
+    color: var(--color-white);
```
thisconnect, line 72: please use `--color-alt`

thisconnect (changes requested):

shonsirsha:
thanks @thisconnect!

Should've read 

`/* all fonts should use these colors and not from the main colors directly */` 🙂

frontends/web/src/components/accountselector/accountselector.module.css
```diff
@@ -1,6 +1,6 @@
 .balance, .balanceSingleValue{
     margin-left: auto;
-    color: var(--color-gray);
+    color: var(--color-white);
```
thisconnect, line 3: Is `--color-white` here intentional?

shonsirsha: I would imagine so.

Otherwise, the colour difference between the text itself and the bg (since bg is gray now) would be a bit too subtle.

shonsirsha: Actually, nevermind. Color gray seems to look ok:

Screenshot 2023-03-13 at 09 06 12

I'll put it back for now,  but using `color-secondary` :)

shonsirsha:
PTAL @thisconnect  :) 🙏

thisconnect (approved):
LGTM

PR #2002, opened 2023-03-09
frontend: refactor utxo component to functional component

thisconnect:
Slightly changed how the component communicates with send.tsx,
it used a React ref that the parent (send) could clear the child.
Changed to just clear the selectedUTXOs map directly after send
or when the dialog opens (it will set the data again when it's
closed).

tested also that the UTXO list updated after sending which was
fixed in https://github.com/thisconnect/bitbox-wallet-app/commit/582ad7e63e9b8d4ab5a700d4da90d938cfa75276

### DIFF
--- a/frontends/web/src/routes/account/send/utxos.tsx
+++ b/frontends/web/src/routes/account/send/utxos.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2023 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -14,92 +15,72 @@
...
+    const unsubscribe = syncdone(accountCode, () => {
+      getUTXOs(accountCode).then(setUtxos);
     });
-  }
-
-  public clear = () => {
-    this.setState({ selectedUTXOs: {} }, () => {
-      this.props.onChange(this.state.selectedUTXOs);
-    });
-  };
```
thisconnect, line 84: this was so that the parent could clear this component, which then emptied the state. selectedUTXOs, and send it back to the parent.

changed to just clear the selectedUTXOs on the parent directly.

shonsirsha (approved):
lgtm

thisconnect:
@Beerosagos I think you also found once a bug related to coin selection, could you test this PR too if you have time?

Beerosagos:
> @Beerosagos I think you also found once a bug related to coin selection, could you test this PR too if you have time?

Yep! I have it on my TODO list :)

Beerosagos (approved):
tACK :+1:

PR #2009, opened 2023-03-13
frontend: convert `settingsitem` into functional component

shonsirsha:

Here, we're converting `settingsitem.tsx` into a functional component.

### DIFF
--- a/frontends/web/src/components/settingsButton/settingsItem.tsx
+++ b/frontends/web/src/components/settingsButton/settingsItem.tsx
@@ -1,26 +1,18 @@
-import { Component } from 'react';
+import { ReactNode } from 'react';
 import style from './settingsButton.module.css';
 
-interface SettingsItemProps {
-    optionalText?: string | null;
-    optionalIcon?: JSX.Element;
+type TProps = {
+  optionalText?: string;
+  optionalIcon?: JSX.Element;
+  children: ReactNode;
 }
+export const SettingsItem = ({ optionalText, optionalIcon, children }: TProps) => {
+  return (
+    <div className={[style.container, style.item].join(' ')}>
+      {children}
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/settingsButton/settingsItem.tsx
```diff
@@ -1,26 +1,22 @@
-import { Component } from 'react';
+import { ReactNode } from 'react';
 import style from './settingsButton.module.css';
 
-interface SettingsItemProps {
-    optionalText?: string | null;
-    optionalIcon?: JSX.Element;
+type TProps = {
+  optionalText?: string | null;
```
thisconnect, line 5: since you are here: I think the null should not be needed, can you remove it or does it explode somewhere?

frontends/web/src/components/settingsButton/settingsItem.tsx
```diff
@@ -1,26 +1,22 @@
...
 }
+export const SettingsItem = ({ optionalText, optionalIcon, children }: TProps) => {
+  return (
+    <div className={[style.container, style.item].join(' ')}>
+      {children}
+      {
+        optionalText && (
+          <span className={style.optionalText}>{optionalText}</span>
+        )
+      }
```
thisconnect, line 17: since you are here, could you remove the newlines:
`{ optionalText` or  `)}` is more consistent imo and nicer to read.

thisconnect (changes requested):

frontends/web/src/components/settingsButton/settingsItem.tsx
shonsirsha: seems to be ok..

shonsirsha:
PTAL :) @thisconnect

thisconnect (approved):
LGTM

PR #2011, opened 2023-03-13
frontend: set darkmode colour scheme for ExchangeSelectionRadio

shonsirsha:
hardcoded some CSS values in order to support darkmode for this component
(ExchangeSelectionRadio).

Used light icons for the Bank and CreditCard icons.

Preview:

Screenshot 2023-03-20 at 12 54 37


frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
```diff
@@ -129,7 +129,7 @@
 /* checked */
 
 .radio[aria-checked="true"] {
-    background-color: var(--color-light-skyblue);
+    background-color: var(--background);
```
thisconnect, line 132: in light mode `--background` is using `--color-lightgray`, but here it was `--color-light-skyblue` 🤔 

is this for focusing the background of the whole option?
do we need a new CSS variable `--background-focus` or similar?

thisconnect (changes requested):

frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
shonsirsha: > is this for focusing the background of the whole option?
Yes exactly.

> do we need a new CSS variable --background-focus or similar?
That'd be good, I think. As lightmode uses skyblue and in darkmode it obviously won't. WDYT?

thisconnect: makes sense to me

shonsirsha:
Rebased and merged conflicts solved :)

thisconnect (approved):
untested LGTM

frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
jadzeidan: I was testing it out and I think it is more intuitive for the focus background to be the lighter one. So like this:

image

shonsirsha: agree, looks better too @jadzeidan . Thanks.

PR #2013, opened 2023-03-14
draft: frontend darkmode toggle

thisconnect:
adding draft as the JS part is not React like. It "works" but doesn't update once the mediaquery changes (if system settings is changed).

Lets discuss in dm.


shonsirsha (approved):
tested lgtm. 


Note for future improvements: we should resolve the known issue in chart.tsx. 

Upon discussion, we may try implementing a hook for darkmode and use it in chart.tsx to resolve this issue. However, this requires chart to be converted into a functional component first.

thisconnect:
updated commit messages

PR #2017, opened 2023-03-17
backend/usb: only force single thread on macOS

benma:
In a35e009b we forced all USB communication to happen in a single thread because macOS otherwise fails with obscure errors (see commit msg).

For simplicity we applied this to all platforms. At least on Windows, users ar experiencing issues where the BitBox02 would timeout during password entry. It is not clear why, but since this was the only change related to USB communication, we revert it for all platforms except macOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [Unreleased]
 - Enable auto HiDPI scaling to correctly manage scale factor on high density screens
 
+## 4.36.1
+- Fix USB communication issue on Windows
+
 ## 4.36.0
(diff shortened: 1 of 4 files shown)


benma:
I applied this commit to 4.36.0 and tagged it 4.36.1 to make a hotfix for Windows only: 

- https://github.com/digitalbitbox/bitbox-wallet-app/commits/v4.36.1
- https://github.com/digitalbitbox/bitbox-wallet-app/releases/tag/v4.36.1

Beerosagos (approved):
utACK

PR #2022, opened 2023-03-21
btc: enable custom fees w/ info text if fee targets are missing

benma:
In case the fee estimations were missing for any reason (e.g. download too slow or some error), the send screen was unusable, as the transaction proposal could not be made due to missing fee. The error was not propagated to the user, and a transaction could not be made.

This patch enables custom fees as a fallback in that case, even if custom fees are otherwise disabled in the settings. It is not great (probably confusing for most users), but better than not being able to transact at all. Luckily this should not happen often - the only instance we know of so far has been with a user that used a custom Electrs node on a slow connection, going into the send screen immediately after sync.

Alternatives considered:
- try downloading the fee rates on the fly if they are missing. Didn't do this as we currently don't have custom Electrum request timeouts. In case the fee rates were missing due to the server or connection being slow, this would block loading the send screen. Doing this approach in a good way would take much more time, and since it is a rare problem, I opted for the quick fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Enable auto HiDPI scaling to correctly manage scale factor on high density screens
+- Bitcoin: enable setting a custom fee if the fee rate estimations are unavailable
 
 ## 4.36.0
 - Re-style header for a better space utilisation
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -526,9 +526,14 @@ outer:
 		feeTargets = append(feeTargets, feeTarget)
 	}
 	// If the default fee level was dropped, use the cheapest.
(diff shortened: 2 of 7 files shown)


benma:
@thisconnect one can simulate the issue by commenting this line:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/3790a033cc34cd7bf70321d8853f4e319f346646/backend/coins/btc/account.go#L425

before running `make servewallet`.

benma:
cc @Tomasvrba

jadzeidan:
Regarding the warning text, here is a alternative version which is a bit more clear and concise: `Fee rate estimations are currently unavailable. Please try again later or enter a custom fee.`

thisconnect (approved):
tested LGTM

made a small PR for styling and wording so you can fixup my commit into yours

PR #2023, opened 2023-03-21
frontend: use CSS var for legacy dialog's background

shonsirsha:
As legacy dialog is still used in some parts of the app, we need
to use proper CSS variables for the component so its styling
will be appropriate for both dark and light mode.

(`<Confirm/>`) before:
Screenshot 2023-03-21 at 10 30 45

after:
Screenshot 2023-03-21 at 10 47 30


thisconnect (approved):
LGTM

PR #2024, opened 2023-03-21
frontend: use CSS variables for account selector

shonsirsha:
We need to use proper CSS variables for the component so its styling will be appropriate for both dark and light mode.

Before:
Screenshot 2023-03-21 at 12 59 22

After:

Screenshot 2023-03-22 at 06 34 23
Screenshot 2023-03-22 at 06 34 30

### DIFF
--- a/frontends/web/src/components/accountselector/accountselector.module.css
+++ b/frontends/web/src/components/accountselector/accountselector.module.css
@@ -28,13 +28,21 @@
     background-color: var(--background-secondary);
 }
 
-.select :global(.react-select__option):hover, 
-.select :global(.react-select__option--is-selected) {
-    background-color: var(--background);
+.select :global(.react-select__option):hover {
+    background-color: var(--background-custom-select-hover);
+}
+
+.select :global(.react-select__option--is-selected),
+.select :global(.react-select__option--is-selected):hover {
+    background-color: var(--background-custom-select-selected);
 }
 
 .select :global(.react-select__option--is-disabled) span {
     color: var(--color-secondary);
+}
+
+.select :global(.react-select__option--is-disabled.react-select__option--is-selected):hover {
+    background-color: var(--background-custom-select-selected);
 }
 
 .select :global(.react-select__option--is-disabled):hover {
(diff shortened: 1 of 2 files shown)


thisconnect:
please make the  background  of the select element white `#FFF` in lightmode so all input elements have same styling.

thisconnect (changes requested):
please make the background of the select element white (`#FFF`) in light mode.

shonsirsha:
> please make the background of the select element white (#FFF) in light mode.

Did you mean when it's selected? or when hovered? or both?

Actually come to think of it, this is how it looks like on master:

Screenshot 2023-03-21 at 18 56 50


🤔 do we want to have it like this?

shonsirsha:
Changes made, screenshots updated.

thisconnect:
> Did you mean when it's selected? or when hovered? or both?

neither, I meant in normal without focus nor active.

thisconnect:
> this is how the selected and hovered state colour look like on master:
> do we want to have it like this too here?

selected and hover look good to me, but I might miss some context. I am about to open another PR first.

thisconnect:
@shonsirsha I tested sean/frontend-account-selector-css-fix in lightmode 8f4b67d2bbc00c556ad7e70f1d77967ad5998d05 

I and prefer the full blue active as in master and  slightly prefer the lightblue hover too, but I dont have a strong opinion.

I guess the lightmode UI shouldn't change and just be as on master.

shonsirsha:
> neither, I meant in normal without focus nor active.

Yup, understood :) 

> I guess the lightmode UI shouldn't change and just be as on master.

Yeah I agree. I'll make the change for the hover n selected colours so they look the same as seen on master.

shonsirsha:
Changes made. PTAL :) @thisconnect 

I had to add new colours and variables in `variables.css` as they weren't available (the blue colours you see on master came from the library). 

I think this makes sense, as these new variables will also be used here: #2025, so we should merge this PR first before that one.

Updated screenshots:

Screenshot 2023-03-22 at 06 34 23
Screenshot 2023-03-22 at 06 34 30

PR #2025, opened 2023-03-21
frontend: use CSS variables country selector

shonsirsha:
we need to use proper CSS variables for the component so its styling will be appropriate for both dark and light mode.

Before:
Screenshot 2023-03-21 at 12 56 27

After:
Screenshot 2023-03-22 at 13 42 50
Screenshot 2023-03-22 at 13 43 07


shonsirsha:
Converting to draft, will merge after #2024

PR #2026, opened 2023-03-21
frontend: add usedarkmode hook to switch assets

thisconnect:
So far darkmode changes only a CSS class on document body. Some assets or also variables in CSS can't be changed this way.

Added useDarkmode hook to use in JavaScript and JSX.


thisconnect:
hm... somehow the trick in frontends/web/src/i18n/matchmedia.mock.js doens't work anymore 🤔

shonsirsha:
This looks fine, however the hook is only reading from OS right? 

When changing theme from the theme toggle it wont be able to detect it. I wonder how we can change the colour of the SwissMadeOpenSource of settings.tsx (not device settings) on the fly (without refreshing/moving to another page) - when a user toggles the theme.

I, for example, played around with footer.tsx and changed the code like so for demonstration purposes:

```diff
export function Footer({ children }: TProps) {
++ const darkmode = useDarkmode();
  return (
    <footer className={[style.footer, 'flex flex-row flex-items-center flex-end'].join(' ')}>
      {children}
++    {darkmode ? <SwissMadeOpenSourceDark/> : <SwissMadeOpenSource/>}
      <div className="m-right-half hide-on-small">
        <Version />
      </div>
      <LanguageSwitch />
    </footer>
  );
}
```

and when `<Footer/>` is being used in the `settings.tsx`, it won't update on the fly:


https://user-images.githubusercontent.com/28676406/226801919-ca4e1203-1c15-42dc-9cf1-c45e8e060852.mov

are my concerns valid? 🤔 

(I know we wont add the swiss logo to the footer, but keep it in settings. Regardless, it's still the same problem.)

thisconnect:
> This looks fine, however the hook is only reading from OS right?

also but not exclusively, it gets mediaquery synchronously and fetches config async. If config is set it always takes it from config, see also this commit and the commit description https://github.com/digitalbitbox/bitbox-wallet-app/commit/12b5deafe9784b20a2fcb4baae1215a6a0bbfefa

> When changing theme from the theme toggle it wont be able to detect it.

true, reason is that config changes currently do not trigger a re-reneder. This is the reason that the darkmodetoggle currently sets the global variable in darkmode.tsx to detect this change.  This could be improved one day if we use something for global application state, i.e. react-context or mobx/redux.

>  I wonder how we can change the colour of the SwissMadeOpenSource of settings.tsx (not device settings) on the fly (without refreshing/moving to another page) - when a user toggles the theme.

> are my concerns valid? 🤔

yep I think you are right without context or similar settings.tsx will have a hard time to re-render. I am not sure yet. This affects all JS/JSX changes on the settings view.

All other components cannot change the darkmode, the way to do it is to refactor to functional component and then either use useDarkmode or just use the View.tsx component that already supports changing on the fly.

shonsirsha:
> also but not exclusively, it gets mediaquery synchronously and fetches config async. If config is set it always takes it from config, see also this commit and the commit description https://github.com/digitalbitbox/bitbox-wallet-app/commit/12b5deafe9784b20a2fcb4baae1215a6a0bbfefa

Yeah, I was incomplete. I meant to say it only reads the darkmode state "on the fly" for OS changes and not config :) which was my concerns yesterday when making the hook. Otherwise, we have to trigger a re-render which is by refreshing/moving to another page.
 
> All other components cannot change the darkmode, the way to do it is to refactor to functional component and then either use useDarkmode or just use the View.tsx component that already supports changing on the fly.

Yeah. I think we're mostly fine except for settings.tsx.. since that's where we can toggle the theme. One hacky way would be to refresh the whole screen when theme is changed from the toggle... but we probably don't want this at least not permanently.

Otherwise, as you've said, we have to store the darkmode up as a global state for this to work seamlessly. But idk how important (?) / pressing this is, as the only thing in `settings.tsx` that needs to be rendered conditionally according to the theme is the swiss opensource logo, right? wdyt?

thisconnect:
> I think the only thing in settings.tsx that needs to be rendered conditionally according to the theme is the swiss opensource logo, right? wdyt?

I think so, ... maybe the hamburger icon on mobile too.

I could make them show via CSS and then use the CSS class. I'll check.

@shonsirsha do you know why the CI fails?

thisconnect:
> (my app is darkmode=true in the video, thus the logo thats showing is ALWAYS the light version and its not updating)

added a workaround using CSS classes and always having the 2 versions in the DOM, see 5b118ac4a815e1d92664e575bc0f6dcc27032497

PR #2030, opened 2023-03-23
qt: fix @media prefers-color-scheme rule in Qt 5.15.2

benma:
Darkmode did not work on linux/Windows/macOS as the `@media (prefers-color-scheme: light/dark)` seemed to be ignored. Defining this env var seems to fix it. I did not investigate further about why that is.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -170,6 +170,11 @@ int main(int argc, char *argv[])
     QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
 
+    // Make `@media (prefers-color-scheme: light/dark)` CSS rules work.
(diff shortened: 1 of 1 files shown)


benma:
I tested linux win and mac and all of these have the problem and for all of them this commit fixes the issue.

thisconnect (approved):
Without user config set, I expected the app to:
- start with same setting as what OS prefers
- change on the fly when OS setting is changed

tested on macOS 11.7.x but it somehow doesn't change nor load the right one... but that shouldn't be a blocker IMO.

maybe it's not supported on older macOS.. approving anyway

thisconnect:
Update: it seems that I am using a slightly newer Qt version (from homebrew)... 🤔

PR #2034, opened 2023-03-23
frontend: fix disabled filter buttons on empty summary

thisconnect:
The chart filter buttons have custom CSS styling which didnt look correct on a new empty account summary view.

Kept lightmode as is and added custom darkmode styling for those buttons in disable state.


shonsirsha (approved):
LGTM.

PR #2035, opened 2023-03-23
frontend/android: change App style to autoswitch darkmode

Beerosagos:
App style parent is changed to `Theme.AppCompat.DayNight.DarkActionBar` as this allows to automatically enable dark mode following system setting (e.g. when on power saving).

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -1,5 +1,7 @@
 package ch.shiftcrypto.bitboxapp;
 
+import static androidx.appcompat.app.AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM;
+
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.AlertDialog;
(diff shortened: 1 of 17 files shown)


Beerosagos:
Pushed an update to change the status bar color to black. This way it will better fit both light and dark modes

benma (changes requested):

frontends/android/BitBoxApp/app/src/main/res/values/colors.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <color name="colorPrimary">#F5F5F5</color>
-    <color name="colorPrimaryDark">#E5E5E5</color>
+    <color name="colorPrimaryDark">#000000</color>
```
benma, line 4: With this, the status bar is now always black, in both dark mode and light mode. Before, it was the same color as the BitBoxApp background color, which we added in 3b1d9d78ddae1679a3d44a5102b0207ed9522bcd to make it look much better.

ChatGPT thinks you can have the color depend on dark mode like:

```
 <color name="colorPrimary">#F5F5F5</color>

<!-- For light mode -->
<color name="colorPrimaryDark">#E5E5E5</color>

<!-- For dark mode -->
<color name="colorPrimaryDark" qualifier="night">#333333</color>
```

Maybe you can experiment with this. Obviously the status bar would be wrong if the dark-mode setting in the app is not the same as the system-level dark mode setting, but it is probably better than a black status bar or a status bar that is always light.

benma (changes requested):

frontends/android/BitBoxApp/app/src/main/res/values/colors.xml
benma: that's why i wrote

  >Obviously the status bar would be wrong if the dark-mode setting in the app is not the same as the system-level dark mode setting, but it is probably better than a black status bar or a status bar that is always light.

benma: For Android, the in-app setting to toggle dark mode could maybe even toggle the Android native dark mod setting for the app, then the above trick would always work too.

https://developer.android.com/develop/ui/views/theming/darktheme

https://developer.android.com/reference/android/app/UiModeManager#setApplicationNightMode(int)

frontends/web/package-lock.json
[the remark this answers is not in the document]
Beerosagos: Done, sorry!

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -130,16 +136,47 @@ private static String getMimeType(String url) {
...
+    }
+
     @SuppressLint("HandlerLeak")
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         Util.log("lifecycle: onCreate");
 
         getSupportActionBar().hide(); // hide title bar with app name.
+        AppCompatDelegate.setDefaultNightMode(MODE_NIGHT_FOLLOW_SYSTEM);
```
benma, line 178: according to docs this is the default value anyway. was there a reason why you needed to set it?

benma (commented):
not tested yet, will test later

benma (approved):
tACK, nice!

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
Beerosagos: Right, that wasn't needed. Thanks! :)

frontends/web/src/components/darkmode/darkmode.tsx
```diff
@@ -15,10 +15,12 @@
  */
 
 import { useDarkmode } from '../../hooks/darkmode';
+import { setDarkTheme } from '../../api/darktheme';
 
 let darkmode: boolean | undefined;
 
 export const setDarkmode = (dark: boolean) => {
+  setDarkTheme(dark);
```
shonsirsha, line 23: If it's intentional to make it async, then it's fine. Otherwise, we should await this since it returns a promise. :)

Beerosagos: There is no return value and I'd proceed with the rest of the function flow even if the call would fail (which should not really happen since the only case is if there is a wrong/missing input param). I'd leave it as it is now

shonsirsha: ack. api call has no direct impact on the app - no point in awaiting :)

PR #1912, opened 2023-01-11
frontends: added css changes for dark mode

thisconnect:
Changed existing CSS and assets to hardcoded dark mode, the related JavaScript will be done in a later commit. Some assets have not yet been updated as they require some additional logic to change to dark mode.

superseds https://github.com/digitalbitbox/bitbox-wallet-app/pull/1883

PR #2036, opened 2023-03-23
frontend: fix terms background color in darkmode

thisconnect:
Text was not readable, removed hardcoded white background color.

### DIFF
--- a/frontends/web/src/routes/buy/terms.module.css
+++ b/frontends/web/src/routes/buy/terms.module.css
@@ -19,7 +19,7 @@
 }
 
 .disclaimer {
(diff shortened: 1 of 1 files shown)


shonsirsha (approved):
LGTM. 

I think we did this before but it was in [`info.module.css`](https://github.com/digitalbitbox/bitbox-wallet-app/pull/1912/files#diff-186a588740b8b9ee6d1845e8121efd268927116ff61345606bf31d500a851547). When it got moved to `terms.module.css` we used the old `disclaimer` class hence why this one was missing.

PR #2042, opened 2023-03-28
frontend: convert `reset.tsx` into a functional component

shonsirsha:
To follow the latest conventions and best practices of React, 
we'd like to convert our class components into functional ones.

Here, we're converting `reset.tsx` into a functional component.

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -32,6 +32,10 @@ type DeviceInfoResponse = SuccessResponse & {
     deviceInfo: DeviceInfo;
 };
 
+export const resetDevice = (deviceID: string): Promise<SuccessResponse | FailResponse> => {
+  return apiPost(`devices/bitbox02/${deviceID}/reset`);
+};
+
 export const getDeviceInfo = (
   deviceID: string
 ): Promise<DeviceInfo> => {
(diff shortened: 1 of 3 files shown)


frontends/web/src/api/bitbox02.ts
```diff
@@ -32,6 +32,10 @@ type DeviceInfoResponse = SuccessResponse & {
     deviceInfo: DeviceInfo;
 };
 
+export const resetDevice = (deviceID: string): Promise<SuccessResponse> => {
+  return apiPost(`devices/bitbox02/${deviceID}/reset`);
+};
```
thisconnect, line 37: should return `Promise<SuccessResponse | FailResponse>` here is the backend part  
https://github.com/digitalbitbox/bitbox-wallet-app/blob/0eef103490f729ca7d73afa78a1dfbdeaa370a70/backend/devices/bitbox02/handlers/handlers.go#L303-L309

thisconnect (changes requested):

frontends/web/src/api/bitbox02.ts
shonsirsha: Thanks! changes, made. PTAL 🙏

thisconnect (approved):
LGTM 👍 
tested 2f6a5fe3f3cb0bb74a491fa0627ead64c4030283

PR #2043, opened 2023-03-28
frontend: convert BuyInfo into a functional component

shonsirsha:

Here, we're converting `info.tsx` into a functional component.

### DIFF
--- a/frontends/web/src/routes/buy/info.tsx
+++ b/frontends/web/src/routes/buy/info.tsx
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import { Component } from 'react';
+import { useState, useEffect, useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
 import { route } from '../../utils/route';
 import { IAccount } from '../../api/account';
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -23,108 +24,100 @@ import Guide from './guide';
...
+    <Main>
+      <GuideWrapper>
+        <GuidedContent>
+          <Header title={<h2>{t('buy.info.title', { name })}</h2>} />
+          <View width="550px" verticallyCentered fullscreen={false}>
+            <ViewContent>
+              { options.length === 0 ? (
+                <div className="content narrow isVerticallyCentered">{t('accountSummary.noAccount')}</div>
+              ) : (
+                <AccountSelector title={t('buy.title', { name })} options={options} selected={selected} onChange={handleChangeAccount} onProceed={handleProceed} />
```
thisconnect, line 114: could you add newlines for each prop, that way it would be more readable.

thisconnect (approved):
tested LGTM with nit

PR #2044, opened 2023-03-28
frontend: use Context for Darkmode

shonsirsha:
Using context for Darkmode allows us to get and set the darkmode state easier across component (globally). We used our own `useDarkmode` hook to wrap the dark mode context.

Keeping `getDarkmode()` as it's still required by class components.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -37,6 +37,7 @@ import { apiPost } from './utils/request';
 import { apiWebsocket } from './utils/websocket';
 import { route, RouterWatcher } from './utils/route';
 import { Darkmode } from './components/darkmode/darkmode';
+import { DarkModeProvider } from './contexts/DarkmodeProvider';
 
  interface State {
      accounts: IAccount[];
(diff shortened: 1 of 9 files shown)


backend/chart.go
```diff
@@ -124,7 +124,7 @@ func (backend *Backend) ChartData() (*Chart, error) {
 		chartDataMissing = true
 		backend.log.Info("ChartDataMissing, until is zero")
 	}
-	isUpToDate := time.Since(until) < 2*time.Hour
+	isUpToDate := time.Since(until) < 24*time.Hour
```
thisconnect, line 127: this is unrelated (I think) remove?

shonsirsha: Ooh lol. I was testing. Will remove.

thisconnect (changes requested):
nice 👍

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }, [osPrefersDarkmode]);
+
+  useEffect(() => {
+    if (isDarkMode) {
+      document.body.classList.add('dark-mode');
+      document.body.classList.remove('light-mode');
+    } else {
+      document.body.classList.remove('dark-mode');
+      document.body.classList.add('light-mode');
+    }
```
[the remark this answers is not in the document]
shonsirsha: Ah good catch. We don't need the other file's classList changes. We just need the setting of `darkmode`. E.g

```js
let darkmode: boolean | undefined;
darkmode = isDarkMode;
```

will make the change.

shonsirsha:
Thanks for the review @thisconnect! Changes made, PTAL  🙏

frontends/web/src/app.tsx
```diff
@@ -184,28 +185,30 @@ class App extends Component<Props, State> {
...
+              />
+              <RouterWatcher onChange={this.handleRoute} />
+            </div>
+            <Alert />
+            <Confirm />
           </div>
-          <Alert />
-          <Confirm />
-        </div>
+        </DarkModeProvider>
```
thisconnect, line 211: one more context-noob question: if we wanted to move sidebar state and toggling + guide state and toggling to context, would that be another provider here? i.e. `<AppProvider>` or similar?

thisconnect (commented):
added one more comment, also added @Beerosagos and @benma to take a look since this is the first react context in the BitBoxApp

shonsirsha:
thanks for the review! @thisconnect

thisconnect:
One more to convert (if it is merged first).

benma (commented):
Is this strictly a refactoring or does it change something functionally (fix a bug or sth)?

shonsirsha:
> Is this strictly a refactoring or does it change something functionally (fix a bug or sth)?

This also fixes a bug where the theme weirdly toggles on its own when language gets changed.

shonsirsha:
> One more to convert in https://github.com/digitalbitbox/bitbox-wallet-app/pull/2049 (if it is merged first).

Nice. I think it makes sense to merge that one first, then change it to be using hook in this PR, then merge this one :) wdyt?

thisconnect:
ack. merged, please rebase and change globe icon

> Nice. I think it makes sense to merge that one first, then change it to be using hook in this PR, then merge this one :) wdyt?

shonsirsha:
> ack. merged, please rebase and change globe icon

Actually I dont think we need to make any more changes because of your PR. Since your PR is using `useDarkmode`, and `useDarkmode` is already using context here :) 

I'll still make some changes to fix the conflict due to  #2035 .

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const setAppTheme = useCallback(() => {
+    setDarkTheme(isDarkMode);
+    if (isDarkMode) {
+      document.body.classList.add('dark-mode');
+      document.body.classList.remove('light-mode');
+    } else {
+      document.body.classList.remove('dark-mode');
+      document.body.classList.add('light-mode');
+    }
+  }, [isDarkMode]);
```
shonsirsha, line 40: Here we used the new `setDarkTheme` created in #2035 . CCing @Beerosagos and @thisconnect in case wanting to take a look once more before I merge.

shonsirsha: also wrapped it in a function `setAppTheme` just so it's neater:)

thisconnect:
@shonsirsha please rebase, then CI should run properly.

PR #2046, opened 2023-03-29
frontend: convert `SkipForTesting` to functional

shonsirsha:
Converting SkipForTesting to functional component.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -318,3 +318,7 @@ export const addAccount = (coinCode: string, name: string): Promise<TAddAccount>
     name,
   });
 };
+
+export const testRegister = (pin: string): Promise<null> => {
+  return apiPost('test/register', { pin });
+};
--- a/frontends/web/src/routes/device/components/skipfortesting.tsx
+++ b/frontends/web/src/routes/device/components/skipfortesting.tsx
@@ -14,51 +14,34 @@
  * limitations under the License.
  */
 
-import React, { Component } from 'react';
+import React, { useState } from 'react';
+import { testRegister } from '../../../api/account';
 import { Button } from '../../../components/forms';
 import { PasswordSingleInput } from '../../../components/password';
-import { apiPost } from '../../../utils/request';
 
-interface SkipForTestingProps {
-    show: boolean;
-}
-
-interface SkipForTestingState {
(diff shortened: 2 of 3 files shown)


frontends/web/src/routes/device/components/skipfortesting.tsx
```diff
@@ -14,51 +14,40 @@
...
+import React, { useState } from 'react';
+import { testRegister } from '../../../api/account';
 import { Button } from '../../../components/forms';
 import { PasswordSingleInput } from '../../../components/password';
-import { apiPost } from '../../../utils/request';
 
-interface SkipForTestingProps {
+type TProps = {
     show: boolean;
 }
```
shonsirsha, line 24: I wonder if we need this at all? 🤔

thisconnect: good catch. this seems to be a left over from when it used show prop https://github.com/digitalbitbox/bitbox-wallet-app/blob/71a45aa2c6566bc1a50433a56750e7ee4f9778e2/frontends/web/src/routes/device/waiting.tsx#L60

frontends/web/src/api/account.ts
```diff
@@ -317,4 +317,8 @@ export const addAccount = (coinCode: string, name: string): Promise<TAddAccount>
     coinCode,
     name,
   });
+};
+
+export const testRegister = (pin: string): Promise<any> => {
```
thisconnect, line 322: It doesn't matter, but return type should be `Promise<null>`.

https://github.com/digitalbitbox/bitbox-wallet-app/blob/88d28497cb56f05d63e6ea465201da1934a6fc38/backend/handlers/handlers.go#L681-L692

thisconnect (approved):
One more suggestion (return type change), then it LGTM

frontends/web/src/api/account.ts
```diff
@@ -317,4 +317,8 @@ export const addAccount = (coinCode: string, name: string): Promise<TAddAccount>
     coinCode,
     name,
   });
+};
+
+export const testRegister = (pin: string): Promise<null> => {
```
benma, line 322: nit: this belongs into the backend.ts not account.ts. account.ts is for account endpoints, which all have a account `code` param. same for `addAccount` above actually.

PR #2047, opened 2023-03-30
backend/btc: do not panic when retrieving transactions

benma:
Fixes one panic TODO. All call sites deal with the error now. An error here is usually a database access for a database that was already closed (e.g. transactions endpoint called at the same time as a bitbox02 is unplugged or the account is closed for some other reason).


benma:
ping @thisconnect

PR #2048, opened 2023-03-30
backend/accounts: de-duplicate account config fields

benma:
accounts.AccountConfig and config.Account have lots of duplicate
fields. Code can be simplifed by reusing the latter in the former.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -168,7 +168,7 @@ func (backend *Backend) createAndPersistAccountConfig(
 	name string,
 	keystore keystore.Keystore,
 	activeTokens []string,
-	accountsConfig *config.AccountsConfig) (accounts.Code, error) {
+	accountsConfig *config.AccountsConfig) (accountsTypes.Code, error) {
 	rootFingerprint, err := keystore.RootFingerprint()
 	if err != nil {
 		return "", err
@@ -296,8 +296,8 @@ func (backend *Backend) CanAddAccount(coinCode coinpkg.Code, keystore keystore.K
 // determined automatically to be the increment of the highest existing account.
 // `name` is the account name, shown to the user. If empty, a default name will be set.
 func (backend *Backend) CreateAndPersistAccountConfig(
-	coinCode coinpkg.Code, name string, keystore keystore.Keystore) (accounts.Code, error) {
-	var accountCode accounts.Code
+	coinCode coinpkg.Code, name string, keystore keystore.Keystore) (accountsTypes.Code, error) {
+	var accountCode accountsTypes.Code
 	err := backend.config.ModifyAccountsConfig(func(accountsConfig *config.AccountsConfig) error {
 		nextAccountNumber, err := nextAccountNumber(coinCode, keystore, accountsConfig)
 		if err != nil {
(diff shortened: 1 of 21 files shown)


benma:
@Beerosagos sorry it's a bit long, but 95% of it are mechanical renames/replacements. I only needed to pay special attention to:

- `Active` in AccountsConfig is `Inactive` in config.Account, so needed to invert everywhere.

backend/accounts.go
```diff
@@ -634,6 +638,7 @@ func (backend *Backend) initPersistedAccounts() {
 	persistedAccounts := backend.config.AccountsConfig()
 outer:
 	for _, account := range backend.filterAccounts(&persistedAccounts, keystoreConnected) {
+		account := account
```
Beerosagos, line 641: Is this needed to make a deep copy of `account` to be passed to `createAndAddAccount`?

Beerosagos (approved):
tACK!

Small nit: maybe rename `Account.Configurations` (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/config/accounts.go#L33), since calling `account.Config().Config.Configurations` is not really intuitive

backend/accounts.go
benma: It's not a deep copy. Account is a pointer.

I did it fearing the old pitfall of Go where loop vars are reused. See https://go.dev/play/p/nX3LCI2yqLR.

Some more detail: https://medium.com/swlh/use-pointer-of-for-range-loop-variable-in-go-3d3481f7ffc9

But it seems if the var itself is a pointer there is no issue? https://go.dev/play/p/6Ej6pvOJ3Ci Can't say I understand it. One of the worst pitfalls of Go, and one that is planned to be fixed afaik.

benma:
> tACK!
> 
> Small nit: maybe rename `Account.Configurations` (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/config/accounts.go#L33), since calling `account.Config().Config.Configurations` is not really intuitive

Yeah it looks bad.

To what though? It is called `(Signing)Configurations` everywhere... I'll rename it to `SigningConfigurations` I guess. 
Maybe the second `.Config` could also be renamed, but not sure to what.

Beerosagos:
> > tACK!
> > Small nit: maybe rename `Account.Configurations` (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/config/accounts.go#L33), since calling `account.Config().Config.Configurations` is not really intuitive
> 
> Yeah it looks bad.
> 
> To what though? It is called `(Signing)Configurations` everywhere... I'll rename it to `SigningConfigurations` I guess. Maybe the second `.Config` could also be renamed, but not sure to what.

Yeah I was thinking about `SigningConfiguration` too. About the second `Config` would be nice to rename but don't know how, maybe `PersistedConfig`? Not super convinced though..

backend/accounts.go
Beerosagos: Yes, I think we could avoid that. Until we don't use the address of the iteration variable (`&account`) everything should be fine.

benma:
> 

Done the rename now. PersistedConfig sounded good in theory but in practice the lines became very long, easily going beynd the 100 char column width:

`			accountNumber, err := accountConfig.PersistedConfig.SigningConfigurations[0].AccountNumber()`

so i didn't do this one

PR #2049, opened 2023-03-30
frontend: add missing globe icon in lightmode

thisconnect:
White globe on white dropdown was not visible, added dark globe icon and conditional rendering.


shonsirsha (approved):
lgtm

PR #2051, opened 2023-03-31
frontend: remove unused reinitializeAccounts() backend API function

benma:
And associated backend handler.

### DIFF
--- a/frontends/web/src/components/rates/rates.tsx
+++ b/frontends/web/src/components/rates/rates.tsx
@@ -17,12 +17,13 @@
 
 import { Fiat, ConversionUnit, IAmount } from '../../api/account';
 import { BtcUnit } from '../../api/coins';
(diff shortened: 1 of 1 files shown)


thisconnect:
oh, why is it in api but was never used.

what about:
https://github.com/digitalbitbox/bitbox-wallet-app/blob/5925755ba092bd3d8cff2b66d737212c2b319212/frontends/web/src/components/rates/rates.tsx#L74-L94

benma:
> oh, why is it in api but was never used.
> 
> what about:
> 
> https://github.com/digitalbitbox/bitbox-wallet-app/blob/5925755ba092bd3d8cff2b66d737212c2b319212/frontends/web/src/components/rates/rates.tsx#L74-L94

oops, very good catch! i undid my commit and instead use reinitializeAccounts() in these two places!

thisconnect (approved):
LGTM 👍

PR #2052, opened 2023-03-31
backend: sort accounts in Accounts()

benma:
We have persisted accounts in accounts.json and load them into live
accounts.

We want the accounts sorted by coin, account number for the
frontend (sidebar, manage accounts). Until now we sorted the accounts
while loading them from accounts.json.

It is preferable to sort the accounts only when serving them to the
frontend - then we could easily modify the backend.accounts list on
the fly and add accounts without having to call `Reinitialize()`,
which closes all accounts and reloads all of them, just so that the
sorting is correct. When we add automatic accounts discovery, we don't
want to Reinitialize() repeatedly for every account that is found, but
just add it to the list.

The sorting function is now a bit more complicated because it needs to
also sort ETH with their ERC20 tokens explicitly (before, they were
correctly ordered implicitly by the order in which they were loaded
from accounts.json).

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -31,6 +31,7 @@ import (
 	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/observable"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/observable/action"
+	"github.com/ethereum/go-ethereum/params"
 )
 
 // hardenedKeystart is the BIP44 offset to make a keypath element hardened.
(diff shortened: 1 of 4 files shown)


Beerosagos (approved):
tACK! Very nice

Small nit: maybe update ERC20 tokens order also in the frontend, to match the sorting? https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/routes/settings/manage-accounts.tsx#L131-L142

benma:
> maybe update ERC20 tokens

hmm i think i'll skip this one. the behavior now is i think the same as before, and i don't want to spend time on this as it seems low prio and might be irrelevant if/when we support adding arbitrary erc20 tokens.

PR #2054, opened 2023-04-03
frontend: fix `ExchangeSelectionRadio`'s outline-overlap issue

shonsirsha:
Our current implementation causing the outline of the `ExchangeSelectionRadio` to be overlapping with another radio component.

To fix this, we're modifying the radio when `:focus` to have `z-index: 2` and `position: relative`.


thisconnect (approved):
LGTM 👍 

(only tested in 4.36.0 via QTWEBENGINE_REMOTE_DEBUGGING as issue is not reproducible in newer Qt versions)

PR #2056, opened 2023-04-04
frontend: remove exchanges recommendations

thisconnect:
Over time this became unmaintained, incomplete and out of date.

Removing recommended exchanges, was linked in the guide from the buy view.

Keeping some of the logos as they are also used in VASP component.


shonsirsha (approved):
LGTM.

PR #2057, opened 2023-04-04
frontend: fix globe icon in lightmode

thisconnect:
in 393344428732250e7066456a04b8d4a4cec82848 the useDarkmode hook changed, it now returns an object with isDarkMode and a function to toggle. This object was always truthy and so the globe always used light icon.


shonsirsha (approved):
LGTM!

Oops, this change shouldve been included in the context PR - but I didn't realise. Good catch.

PR #2058, opened 2023-04-04
backend/environment: detect OS theme at startup

Beerosagos:
On Android, the preferred theme is handled using media queries in the frontend. This allows to update darkmode setting on the fly if the user enable a dark theme at OS level.

On Qt this doesn't seem to work correctly, as the media queries return weird values. This update creates a dedicated `detect-dark-theme` endpoint that exploits backend environment functions to apply the correct theme at startup under Windows, Mac OS and Linux (gtk, kde, xfce4). Note that this doesn't allow on the fly updates of the theme.


frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -43,16 +45,22 @@ export const DarkModeProvider = ({ children }: TProps) => {
     getConfig()
       .then(config => {
         // use config if it exists
-        if ('darkmode' in config.frontend) {
+        if (!!config.frontend && 'darkmode' in config.frontend) {
```
Beerosagos, line 48: This missing check was causing a weird behavior due to js runtime exceptions on the first fresh start of the app, with no config file.

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -26,7 +27,8 @@ type TProps = {
 
 export const DarkModeProvider = ({ children }: TProps) => {
   const [isDarkMode, setIsDarkMode] = useState(false);
-  const osPrefersDarkmode = useMediaQuery('(prefers-color-scheme: dark)');
+  const [osPrefersDarkMode, setOsPrefersDarkMode] = useState(false);
```
shonsirsha, line 30: I know it's still on draft / probably WIP, but just reminding that I don't think we need `osPrefersDarkMode` here. Can be safely removed 🙂

Beerosagos: You mean that we could detect the theme again in `toggleDarkmode`?

shonsirsha: Sorry, i wasn't being so clear. What I meant is I guess we could remove `osPrefersDarkmode` and use the `isDarkMode` instead?

The useEffect hook that sets the isDarkMode state already checks  the user's system theme and sets the state accordingly. 

```js
          detectDarkTheme().then(isDark => {
            setOsPrefersDarkMode(isDark); 
            setIsDarkMode(isDark); // here the theme on the app startup/load already determined and set to isDarkMode
          });
```

E.g

```diff
 if (runningInAndroid()) {
          preferredDarkMode = androidPrefersDarkMode;
        } else {
-         preferredDarkMode = osPrefersDarkMode;
+         preferredDarkMode = isDarkMode;
        }
```

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -62,7 +70,13 @@ export const DarkModeProvider = ({ children }: TProps) => {
     setIsDarkMode(!isDarkMode);
```
shonsirsha, line 70: Not your changes, but I think it's fine to change it here. 

Could we change this bit to `setIsDarkMode(darkmode)` since it's already receiving the variable (`darkmode`). Just for consistency :)

Beerosagos: Done :+1:

Beerosagos:
Theme detection wasn't working properly on several Linux distros, now should be much better and ready for review.
Tested on:
- Linux Mint 20.3 (Ubuntu based, Cinnamon environment)
- Kubuntu 21.10 (KDE environment)
- Ubuntu 22.10 (Gnome environment)

@benma you previously tested the code on xfce iirc, but pls test again if you have time. Also windows haven't been tested yet. Could you test that too? :pray: 

cc @thisconnect @shonsirsha

thisconnect:
97df83dd357c6482d20d38963da657b830a49ed3 on macOS (tested with slightly newer Qt version detected system dakrmode/lightmode when there was no user config) LGTM

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -43,26 +44,37 @@ export const DarkModeProvider = ({ children }: TProps) => {
...
-      .then(config => {
-        if (osPrefersDarkmode === darkmode) {
+      .then(async config => {
+        let preferredDarkMode;
+        if (runningInAndroid()) {
+          preferredDarkMode = androidPrefersDarkMode;
+        } else {
+          preferredDarkMode = await detectDarkTheme();
+        }
+        if (preferredDarkMode === darkmode) {
```
shonsirsha, line 77: Not necessarily wrong to mix `await` and `.then()` but generally we try to stick to one convention, especially in a single function. 🙂

shonsirsha (approved):
LGTM.

Tested on mac without config: as expected, app's theme follows the system's theme. App is also able to save the theme set by the user. 🚀

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -43,26 +44,37 @@ export const DarkModeProvider = ({ children }: TProps) => {
...
         }
         // else use mode from OS
-        setIsDarkMode(osPrefersDarkmode);
-
+        if (runningInAndroid()) {
+          setIsDarkMode(androidPrefersDarkMode);
+        } else {
+          detectDarkTheme().then(isDark => {
+            setIsDarkMode(isDark);
+          });
```
benma, line 57: can be expressed shorter as `detectDarkTheme().then(setIsDarkMode)`

frontends/qt/server/server.go
```diff
@@ -91,6 +92,78 @@ func handleURI(uri *C.char) {
...
+		cmd = exec.Command("xfconf-query", "-c", "xsettings", "-p", "/Net/ThemeName")
+		out, err = cmd.Output()
+		if err == nil {
+			log.Info("xfce theme: " + string(out))
+			if matchDarkTheme(string(out)) {
+				return true
+			}
+		}
+	case "windows":
+		const regKey = `Software\Microsoft\Windows\CurrentVersion\Themes\Personalize`
```
benma, line 153: I tested - this has to be 

		const regKey = `HKCU\Software\Microsoft\Windows\CurrentVersion\Themes\Personalize`

frontends/qt/server/server.go
```diff
@@ -91,6 +92,78 @@ func handleURI(uri *C.char) {
...
+	case "darwin":
+		cmd := exec.Command("defaults", "read", "-g", "AppleInterfaceStyle")
+		out, err := cmd.Output()
+		if err == nil {
+			log.Info("MacOS theme: " + string(out))
+			if strings.TrimSpace(string(out)) == "Dark" {
+				return true
+			}
+		}
+	case "linux":
```
benma, line 112: i couldn't test this as my qubes system does not patch through this system setting to the vm's.

frontends/qt/server/server.go
```diff
@@ -91,6 +92,78 @@ func handleURI(uri *C.char) {
...
+			}
+		}
+	case "windows":
+		const regKey = `Software\Microsoft\Windows\CurrentVersion\Themes\Personalize`
+		const regName = `AppsUseLightTheme`
+		cmd := exec.Command("reg", "query", regKey, "/v", regName)
+		out, err := cmd.Output()
+		if err == nil {
+			log.Info("windows theme: " + string(out))
+			if strings.Contains(strings.TrimSpace(string(out)), "0x0") {
```
benma, line 159: same - trimspace not needed?

benma (changes requested):

frontends/qt/server/server.go
```diff
@@ -91,6 +92,78 @@ func handleURI(uri *C.char) {
...
 
+func matchDarkTheme(themeName string) bool {
+	return strings.Contains(strings.ToLower(strings.TrimSpace(themeName)), "dark")
+}
+
+// detect theme used by OS and return true if it's dark
+func detectDarkTheme() bool {
+	log := logging.Get().WithGroup("server")
+	switch myos := strings.ToLower(runtime.GOOS); myos {
+	case "darwin":
```
benma, line 103: tested works on macOS 13

benma (approved):
partly tested ack - tested windows, macOS and android.

PR #2059, opened 2023-04-05
backend/coins: improve BTC/sats amount formatting

Beerosagos:
BTC/sats amount were formatted without any kind of thousands separator, causing poor readability. This improves the formatting adding spaces as separators on both the integer and fractional part of the amount.


Beerosagos:
Converted to draft until it is not ready for a new review

Beerosagos:
Updated following an html formatting approach, only for sat amounts. PTAL :pray:

frontends/web/src/components/amount/amount.module.css
```diff
@@ -0,0 +1,3 @@
+.space {
+    margin-left: 0.5ch;
```
Beerosagos, line 2: I ended up with 0.5 because 1 seemed a bit too much imho

Beerosagos:
I also prepared the code to space the fractional part of BTC if we want to add that too. The only question is that in the past we decided to strip the last zeroes from the balance amount at the top (and I don't think that it makes sense to add spaces if we are not showing the whole fractional part). We could avoid removing the zeroes and adding spaces or leave it like in this example.

image

frontends/web/src/components/amount/amount.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import style from './amount.module.css';
+
+type TProps = {
+  amount: string;
+  unit: string;
```
benma, line 20: can't we use the account.ts:`CoinWithSat` type here?

would also be nice to rename `Coin` to `CoinUnit` and `CoinWithSat` to `CoinUnitWithSat` - was pretty confused about these type names when I looked them up now.

benma (commented):
will review the rest until tomorrow

frontends/web/src/components/amount/amount.tsx
Beerosagos: `Amount` in this PR is used for both fiat and coin amounts, so `unit` should be something like `ConversionUnit | CoinWithSAT`. What about just renaming `CoinWithSAT` to `CoinUnit`? I think it would be nice to separate coins from their units and that would mean having `unit: ConversionUnit | CoinUnit` which can be easier to understand. wdyt?

benma: makes sense, and the `Coin/MainnetCoin/TestnetCoin` ones can actually be inlined and deleted. Does this commit make sense?

https://github.com/benma/bitbox-wallet-app/commit/c839829c2e2a83d5910eb6205a7f4106056fdba5

frontends/web/src/components/amount/amount.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const blockSize = 3;
+
+    for (let i = 0; i > -len ; i -= blockSize) {
+      blocks.push(
+        <span className={i > -len ? style.space : ''}>
+          {i === 0 ? amount.slice(i - blockSize) : amount.slice(i - blockSize, i)}
+        </span>);
+    }
+    return <>{blocks.reverse()}</>;
+  };
```
benma, line 37: I found this a bit hard to understand. It is quite unusual to have a loop index start at 0 and go negative. I asked Chat to simplify and this is what came out:

```
const formatSats = (amount: string): JSX.Element => {
    const blocks: JSX.Element[] = [];
    const blockSize = 3;

    for (let i = amount.length; i > 0 ; i -= blockSize) {
      const start = Math.max(0, i - blockSize);

      blocks.push(
        <span className={start === 0 ? '' : style.space}>
          {amount.slice(start, i)}
        </span>
      );
    }

    return <>{blocks.reverse()}</>;
  };

```

imo this is easier to read and the unit tests still pass :smile: wdyt?

frontends/web/src/components/amount/amount.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+describe('Amount formatting', () => {
+  let coins;
+
+  describe('sat amounts', () => {
+    coins = ['sat', 'tsat'];
+    coins.forEach(coin => {
+      it('12345678901234 ' + coin + ' with removeBtcTrailingZeroes enabled gets spaced', () => {
+        const { getByText } = render(<Amount amount="12345678901234" unit={coin} removeBtcTrailingZeroes/>);
+        expect(getByText('12')).toBeInTheDocument();
```
benma, line 28: i removed the `.reverse()` in the code and all unit tests still passed. not sure if it is worth it to improve the tests - i checked that it is correct manually - but such a bug would be devestating.

benma (approved):
very nice!!

frontends/web/src/components/amount/amount.tsx
Beerosagos: Yeah, you're right! I included your modifications in my commit, thanks

frontends/web/src/components/amount/amount.tsx
Beerosagos: This is great! Thanks

frontends/web/src/components/amount/amount.test.tsx
Beerosagos: Good catch! I updated the test to actually check for the HTML and detect this kind of error

Beerosagos:
> I also prepared the code to space the fractional part of BTC if we want to add that too. The only question is that in the past we decided to strip the last zeroes from the balance amount at the top (and I don't think that it makes sense to add spaces if we are not showing the whole fractional part). We could avoid removing the zeroes and adding spaces or leave it like in this example.
> 
> image

I'll move this into a separate PR

PR #2061, opened 2023-04-06
frontend: pass theme to moonpay widget

thisconnect:
So that is displays correct light and dark mode.


thisconnect:
tested in webdev and qt-osx on macOS 11.7 with Qt 5.15.8

shonsirsha (approved):
LGTM.

(tested in webdev & QT 5.15.8 macOS ).

PR #2064, opened 2023-04-06
frontend: send button spacing css regression

thisconnect:
The send and back button had no more spacing. CSS regression introduced in:
- 7e49d89fb8ed97184a756b11701c21c83128abe1

Reason: on older chrome versions (before 84) gap did not work with flexbox, gap only works with dislpay: grid. See also:
- https://developer.mozilla.org/en-US/docs/Web/CSS/gap

### DIFF
--- a/frontends/web/src/components/layout/grid.module.css
+++ b/frontends/web/src/components/layout/grid.module.css
@@ -18,7 +18,10 @@
 .columnButtonsInline {
     display: flex;
     flex-direction: row-reverse;
-    gap: var(--space-half);
+}
+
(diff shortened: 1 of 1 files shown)


thisconnect:
tested qt-osx (Qt 5.15.8) on macOS 11.7

shonsirsha:
I'm not seeing this on master (HEAD: f971947edc74b92e6307f77004204e48c3d3fa5b) 🤔 

Screenshot 2023-04-11 at 11 27 01

shonsirsha (approved):
tested LGTM QT 5.15.8 macOS

PR #2065, opened 2023-04-10
frontend: convert `Status` into a functional component

shonsirsha:

Here, we're converting `status.tsx` into a functional component.


frontends/web/src/components/status/status.tsx
```diff
@@ -15,98 +15,71 @@
...
-
-export default class Status extends Component<Props, State> {
-  public readonly state: State = {
-    show: true,
-  };
-
-  public componentDidMount() {
-    this.checkConfig();
-  }
+const Status = ({ hidden, type = 'warning', dismissable, className, children }: TPRops) => {
```
thisconnect, line 37: tiny nit: do you mind adding a newline for every prop, since improves readability

frontends/web/src/components/status/status.tsx
```diff
@@ -15,98 +15,71 @@
...
-  public componentDidUpdate(prevProps: Props) {
-    if (this.props.dismissable !== prevProps.dismissable) {
-      this.checkConfig();
+  const checkConfig = useCallback(async() => {
+    if (dismissable) {
+      const config = await getConfig();
+      setShow(!config ? true : !config[dismissable]);
     }
-  }
+  }, [dismissable]);
```
thisconnect, line 45: could you change dismissable -> dismissible :) ?

code looks good to me, but I wonder why the old implementation had the componentDidUpdate. `dismissible` should not change during component livecycle, I think doing on mount would be enough. but maybe not in this commit, wdyt?

https://github.com/digitalbitbox/bitbox-wallet-app/search?q=dismissable&type=code
componentDidUpdate

shonsirsha: >  I think doing on mount would be enough. but maybe not in this commit, wdyt?


Yeah, I'd say so. I haven't exhaustively tested it yet though. 

And sure, I'll make the change for this in a new commit. :)

shonsirsha:
rebased + changes made.

Thanks for the review @thisconnect - PTAL 🙏

frontends/web/src/components/status/status.tsx
```diff
@@ -15,98 +15,77 @@
...
+}: TPRops) => {
+  const [show, setShow] = useState(true);
 
-  public componentDidUpdate(prevProps: Props) {
-    if (this.props.dismissable !== prevProps.dismissable) {
-      this.checkConfig();
+  const checkConfig = useCallback(async () => {
+    if (dismissible) {
+      const config = await getConfig();
+      setShow(!config ? true : !config[dismissible]);
```
thisconnect, line 49: they property is stored in `config.frontend[dismissible]`

best test with downgrading app version in update.go and restart `make servewallet`

```diff
diff --git a/backend/update.go b/backend/update.go
index 8b3b7b1d2..2162f72ad 100644
--- a/backend/update.go
+++ b/backend/update.go
@@ -27,7 +27,7 @@ const updateFileURL = "https://shiftcrypto.ch/updates/desktop.json"
 
 var (
        // Version of the backend as displayed to the user.
-       Version = semver.NewSemVer(4, 37, 0)
+       Version = semver.NewSemVer(4, 34, 0)
 )
 
 // UpdateFile is retrieved from the server.
```

also when dismissing the status banner, you should see an entry in appfolder.dev/config.json

shonsirsha: > they property is stored in config.frontend[dismissible]

Ups, yeah. 🤦  Shouldve been `config.frontend...`. My bad. thx!

shonsirsha: made the change and rebased.

shonsirsha:
rebased

PR #2071, opened 2023-04-13
frontend: improve spacing of exchange radio

thisconnect:
Long label after the bank/card icon broke weirdly to a new line in some languages. This change is not perfect but improves that the word break nicer when needed and not rightaway after the icon.

### DIFF
--- a/frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
+++ b/frontends/web/src/routes/buy/components/exchangeselectionradio.module.css
@@ -50,7 +50,9 @@
     font-weight: 400;
     line-height: 22px;
     margin: 0;
-    margin-left: var(--space-quarter);
+}
+.paymentMethodName img {
+    margin-right: var(--space-quarter);
 }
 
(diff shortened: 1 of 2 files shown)


thisconnect:
current:

Screen Shot 2023-04-13 at 17 56 43


with this PR:

Screen Shot 2023-04-13 at 17 09 44

shonsirsha (approved):
nice! LGTM.

Beerosagos:
Would be possible/make sense to force the badges text (e.g. "Miglior prezzo") to be on a single line?

thisconnect:
yes, it would be possible to add styling so that words in the badge never break, but then space after bank/creditcard is even smaller. Better would be if the badge wouldn't take so much space when it is breaking, but I didn't figure out a way how to do that yet and thought this change is already an improvment.

> Would be possible/make sense to force the badges text (e.g. "Miglior prezzo") to be on a single line?

Beerosagos:
LGTM too!

PR #2072, opened 2023-04-17
frontend/exchange: fix back button behavior

Beerosagos:
Back button on exchange selection page wasn't working correctly on Android. Also the logic to choose where to land when pressing back was based on the number of total accounts, which is wrong. Now it is based on the number of total accounts with at least one available exchange.

### DIFF
--- a/frontends/web/src/routes/buy/exchange.tsx
+++ b/frontends/web/src/routes/buy/exchange.tsx
@@ -15,7 +15,6 @@
  */
 import 'flag-icons';
 import { useState, useEffect } from 'react';
-import { useNavigate } from 'react-router-dom';
 import i18n from '../../i18n/i18n';
 import { useTranslation } from 'react-i18next';
 import { Button } from '../../components/forms';
@@ -27,7 +26,7 @@ import { findAccount, getCryptoName } from '../account/utils';
 import { route } from '../../utils/route';
 import { useLoad } from '../../hooks/api';
 import { getRegionNameFromLocale } from '../../i18n/utils';
-import { findLowestFee, findBestDeal, getFormattedName } from './utils';
+import { findLowestFee, findBestDeal, getFormattedName, getExchangeSupportedAccounts } from './utils';
 import { ExchangeSelectionRadio } from './components/exchangeselectionradio';
 import { Spinner } from '../../components/spinner/Spinner';
 import { Info, FrontendExchangeDealsList } from './types';
@@ -56,18 +55,23 @@ export const Exchange = ({ code, accounts }: TProps) => {
   const [regions, setRegions] = useState<TOption[]>([]);
   const [allExchangeDeals, setAllExchanges] = useState<FrontendExchangeDealsList>();
   const [info, setInfo] = useState<Info>();
+  const [supportedAccounts, setSupportedAccounts] = useState<IAccount[]>([]);
 
   const regionList = useLoad(exchangesAPI.getExchangesByRegion(code));
   const exchangeDeals = useLoad(exchangesAPI.getExchangeDeals);
   const nativeLocale = useLoad(getNativeLocale);
   const supportedExchanges = useLoad<exchangesAPI.SupportedExchanges>(exchangesAPI.getExchangeBuySupported(code));
   const config = useLoad(getConfig);
-  const navigate = useNavigate();
 
   const account = findAccount(accounts, code);
   const name = getCryptoName(t('buy.info.crypto'), account);
 
   const hasOnlyOneSupportedExchange = allExchangeDeals ? allExchangeDeals.exchanges.filter(exchange => exchange.supported).length === 1 : false;
+
+  // get the list of accounts supported by exchanges, needed to correctly handle back button.
+  useEffect(() => {
+    getExchangeSupportedAccounts(accounts).then(setSupportedAccounts);
+  }, [accounts]);
 
   // update region Select component when `regionList` or `config` gets populated.
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -40,14 +40,7 @@ export const BuyInfo = ({ code, accounts }: TProps) => {
...
     try {
-      const accountsWithFalsyValue = await Promise.all(
-        accounts.map(async (account) => {
-          const supported = await getExchangeBuySupported(account.code)();
-          return supported.exchanges.length ? account : false;
-        })
-      );
-      const supportedAccounts =
-        (accountsWithFalsyValue.filter(result => result) as IAccount[]);
+      const supportedAccounts = await getExchangeSupportedAccounts(accounts);
```
shonsirsha, line 43: nice refactor!

frontends/web/src/routes/buy/exchange.tsx
```diff
@@ -178,14 +182,13 @@ export const Exchange = ({ code, accounts }: TProps) => {
...
 
   const handleGoBack = () => {
-    if (accounts.length > 1) {
-      return navigate(-1);
+    if (supportedAccounts.length > 1) {
+      // go back to the account selection page
+      route('/buy/info');
+    } else {
+      // go back to home screen
+      route('/account-summary');
```
shonsirsha, line 190: might be slightly confusing if the user didn't come from `/account-summary`. 

E.g they might come from `settings` (before going into `/buy/info` which auto-redirects them to this page `/buy/exchange/accountId`).

Clicking on "back", they most likely would expect to go back to `settings` and not `account-summary`.  Just a nit ig.. 🤔 🤷

shonsirsha (approved):
code LGTM 👍 . Just a small comment:)

frontends/web/src/routes/buy/exchange.tsx
shonsirsha: Hm... 

I think the bit of code that you changed would do what I suggested.

```js
if (accounts.length > 1) {
      return navigate(-1);
```

Will go back to `/buy/info`

... otherwise

```js
//otherwise
navigate(-2);
```

will go back to `/buy/info` then go back to whatever page before that, which is the page that the user went before clicking on "Buy Crypto / Bitcoin".

But... if `navigate` doesn't work on Android then yeah, have to find an alternative.

shonsirsha: A little strange though that it doesn't work on Android. Did you find any bug report or people facing the same issue on android, I guess chromium (mobile ?)

Beerosagos: Good point. I could not find anything about this issue online, and I am not sure how to log that behavior on Android.
Could we try a different library maybe? Like React Navigation? https://reactnavigation.org/

shonsirsha: > Could we try a different library maybe? Like React Navigation? https://reactnavigation.org/

Might be a bit overkill if it's only for this purpose.. 🤔 

> We could also think about removing the Back button if there is only one account with available exchanges.

I initially agree w/ this, but thinking back.. could be annoying for users (especially mobile) if there's no back btn. To go out from the `/buy/exchange` user needs to click on the hamburger menu, then clicking on the desired menu. 

Especially not a very good experience since the built-in back btn on Android isn't working  if I'm not mistaken.

Let's hear thisconnect's opinion on this.

Beerosagos: > I initially agree w/ this, but thinking back.. could be annoying for users (especially mobile) if there's no back btn. To go out from the /buy/exchange user needs to click on the hamburger menu, then clicking on the desired menu.

Isn't the same in every other section of the app? I think we usually use the back button only in flows and situations where there is a clear parent page. Anyway yes, let's hear @thisconnect thoughts about this.

Beerosagos: @thisconnect we briefly discussed this yesterday, but I don't think we actually took a decision. wdyt?

frontends/web/src/api/exchanges.ts
```diff
@@ -101,3 +101,13 @@ export const getExchangeBuySupported = (code: string) => {
...
+
+export const getExchangeSupportedAccounts = async (accounts: IAccount[]): Promise<IAccount[]> => {
+  const accountsWithFalsyValue = await Promise.all(
+    accounts.map(async (account) => {
+      const supported = await getExchangeBuySupported(account.code)();
+      return supported.exchanges.length ? account : false;
+    })
+  );
+  return accountsWithFalsyValue.filter(result => result) as IAccount[];
+};
```
thisconnect, line 113: nit: all other functions in api just add types to get/post responses. This one is more like a helper function, would it make sense to have a backend endpoint that takes ` IAccount[]` ?

alternatives move to `src/routes/buy/utils` or similar?

frontends/web/src/routes/buy/exchange.tsx
thisconnect: browser back would have been an relatively simple solution, but as it doesnt' work on Android (assuming this is a limitation of android webview) I think we should only show the backbutton if it goes back to account selection. If there is only 1 account in the selection we can hide the back button for now. 

account selection doesn't have back button. account selection backbutton would go to different views, depending on where the user came from.

maybe one day we can look into a general solution for back (also android native back button) and look into why  history back doesn't work on anrdoid.

Beerosagos:
Updated hiding the back button when there is only one supported account and moving `getExchangeSupportedAccounts` into `utils.ts`. PTAL :pray:

thisconnect (approved):
LGTM

PR #2073, opened 2023-04-19
api/bitbox02: allow backups using recovery words, skipping sdcard

benma:
This adds a parameter to the createBackup endpoint to choose the sdcard backup or the recovery words backup. This can be then used later to allow skipping the sdcard backup in favor of the recovery words backup in the BitBox02 setup wizard.

Since this feature is only possible since firmware v9.13.0, one must first check if `versionInfo.canBackupWithRecoveryWords` is true before using the recovery words method.

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -68,6 +68,14 @@ export type VersionInfo = {
     currentVersion: string;
     canUpgrade: boolean;
     canGotoStartupSettings: boolean;
+    // If true, creating a backup using the mnemonic recovery words instead of the microSD card
+    // is supported in the initial setup.
+    //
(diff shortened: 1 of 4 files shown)


frontends/web/src/api/bitbox02.ts
```diff
@@ -68,6 +68,14 @@ export type VersionInfo = {
     currentVersion: string;
     canUpgrade: boolean;
     canGotoStartupSettings: boolean;
```
benma, line 70: @thisconnect how come our linter forces 2 spaces everywhere but not inside type declarations? :see_no_evil:

thisconnect: I think because eslint doesn't know typescript.

thisconnect (approved):
tested LGTM

PR #2074, opened 2023-04-19
frontend: fix regression api create-backup

thisconnect:
createBackup was still using catch in case of success false, this should have been part of 374bd5377b0021ab23f452657e1ddaf721524f63

This also fixes an issue when continuing the workflow after the user aborted is-date-correct on the device. After canceling the app showed confirm date on device but the device did not continue.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -310,10 +310,13 @@ class BitBox02 extends Component<Props, State> {
         text: this.props.t('bitbox02Interact.confirmDateText'),
       } });
       createBackup(this.props.deviceID, 'sdcard')
-        .then(() => this.setState({ creatingBackup: false, waitDialog: undefined }))
-        .catch(() => {
-          alertUser(this.props.t('bitbox02Wizard.createBackupFailed'), { asDialog: false });
-        });
+        .then(({ success }) => {
+          if (!success) {
+            alertUser(this.props.t('bitbox02Wizard.createBackupFailed'), { asDialog: false });
+          }
+          this.setState({ creatingBackup: false, waitDialog: undefined });
+        })
+        .catch(console.error);
     });
   };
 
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -310,10 +310,13 @@ class BitBox02 extends Component<Props, State> {
...
       } });
       createBackup(this.props.deviceID, 'sdcard')
-        .then(() => this.setState({ creatingBackup: false, waitDialog: undefined }))
-        .catch(() => {
-          alertUser(this.props.t('bitbox02Wizard.createBackupFailed'), { asDialog: false });
-        });
+        .then(({ success }) => {
+          if (!success) {
+            alertUser(this.props.t('bitbox02Wizard.createBackupFailed'), { asDialog: false });
+          }
```
benma, line 316: return missing here?

Beerosagos:
> This also fixes an issue when continuing the workflow after the user aborted is-date-correct on the device. After canceling the app showed confirm date on device but the device did not continue.

Tried to reproduce this on master, but could not. Which are the steps?

Also, I noticed that the endpoint returns: 
```
{
	"code": 104,
	"message": "aborted by the user",
	"success": false
}
```
Would be nice to check the error code for `errUserAbort` (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/api/bitbox02.ts#L21) and print a dedicated message in that case.

thisconnect:
when creating a new wallet cancel the "is-date-correct" prompt on the device.

without this PR it will go back to the last step.

with this PR it will show a message that it failed, then goes back to the last step.

> Tried to reproduce this on master, but could not. Which are the steps?



I agree we could show a better message, should I add a 2nd commit in this PR?

> Would be nice to check the error code for errUserAbort

Beerosagos:
> when creating a new wallet cancel the "is-date-correct" prompt on the device.
without this PR it will go back to the last step.
with this PR it will show a message that it failed, then goes back to the last step.

Ah yes now I see it!

> I agree we could show a better message, should I add a 2nd commit in this PR?

Would be great, thanks :pray:

thisconnect:
it should show user aborted error now.

Beerosagos (approved):
tACK!

PR #2076, opened 2023-04-21
backend/btc: do not panic when retrieving balance

benma:
Similar to 52da637577bc39d9b4871c03e0c2c28d58a4ff68, where the same as done for the transactions list.

Fixes one panic TODO. All call sites deal with the error now. An error here is usually a database access for a database that was already closed (e.g. transactions endpoint called at the same time as a bitbox02 is unplugged or the account is closed for some other reason).


benma:
Closing as out of date, the main problem was addressed already previously in 4427b1c2b4c797a817b339f33673b9d0f7dd9224

PR #2078, opened 2023-04-24
frontend/skipfortesting: fix bug on `registerTestingDevice`

Beerosagos:
Recent commit 8a20585 refactored `SkipForTesting` component into functional. This introduced a regression caused by the calling of `preventDefault` method of a `React.SyntheticEvent` after awaiting a post call. This on Firefox seems to cause the component to refresh before the conclusion of the call, causing a regression that blocks the use of test wallets. Moving the `preventDefault` call before the `await` fixes it.


shonsirsha (approved):
good catch!

PR #2080, opened 2023-04-25
frontend: fix default & active currencies selection

shonsirsha:
Previously, in the newly created appearance (`new-settings/appearance`) we could:

1. set a default currency that's not even in our active currencies list.

2. remove our default currency from our active currencies list. So our default currency
won't be among the active currencies, which doesn't make sense.

**These are bugs and aren't the intended behaviours.** This could be verified by the fact that in the old settings, these aren't allowed.
 
The root cause is because we needed to add the checks. After adding the checks though, it's still not fixed because we can't get the updated values of the store once these values change, except by refreshing it (which isn't what we want).

**To fix this, we unfortunately still need to use our "old" `share` decorator. It's because these values are currently stored within our own `store` implementation. We can't listen to changes in this store simply by using `useEffect` since it's not a react state.**

Thus, we need to wrap the component, in this case `activeCurrenciesDropdown` using the `share` decorator to be able to subscribe to these currencies changes and get the update immediately once the `store` values get updated.

Preview:

https://user-images.githubusercontent.com/28676406/234286169-19eb186a-2a3b-4cef-8478-93f3ae74a0f1.mov

### DIFF
--- a/frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
+++ b/frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
@@ -1,8 +1,9 @@
 import Select, { ActionMeta, MultiValue, MultiValueRemoveProps, components } from 'react-select';
 import { SettingsItemContainer } from '../settingsItemContainer/settingsItemContainer';
-import { currencies, store, selectFiat, unselectFiat } from '../../../../components/rates/rates';
+import { currencies, store, selectFiat, unselectFiat, SharedProps } from '../../../../components/rates/rates';
 import { Fiat } from '../../../../api/account';
 import styles from './defaultCurrencySetting.module.css';
+import { share } from '../../../../decorators/share';
 
 type SelectOption = {
   label: Fiat;
@@ -11,13 +12,15 @@ type SelectOption = {
 
 type TSelectProps = {
   options: SelectOption[];
-  selectedCurrencies: SelectOption[];
-}
+} & Omit<SharedProps, 'btcUnit'>
 
-const ReactSelect = ({ options, selectedCurrencies }: TSelectProps) => {
+const ReactSelect = ({ options, active, selected }: TSelectProps) => {
+  const selectedCurrencies = selected.length > 0 ? selected?.map(currency => ({ label: currency, value: currency })) : [];
+
   const MultiValueRemove = (props: MultiValueRemoveProps<SelectOption>) => {
+    const currency = props.data.value;
     return (
-      selectedCurrencies.length > 1 ?
+      currency !== active ?
         <components.MultiValueRemove {...props}>
           {'X'}
         </components.MultiValueRemove>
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
```diff
@@ -11,13 +12,15 @@ type SelectOption = {
...
+
+const ReactSelect = ({ options, active, selected }: TSelectProps) => {
+  const selectedCurrencies = selected.length > 0 ? selected?.map(currency => ({ label: currency, value: currency })) : [];
 
-const ReactSelect = ({ options, selectedCurrencies }: TSelectProps) => {
   const MultiValueRemove = (props: MultiValueRemoveProps<SelectOption>) => {
+    const currency = props.data.value;
     return (
-      selectedCurrencies.length > 1 ?
+      currency !== active ?
```
shonsirsha, line 23: removed unused check too here (`selectedCurrencies.length > 1`)

shonsirsha:
Not the best pattern anymore to use this `store` + `share` decorator cc @thisconnect , but unfort. we have to do it here as theres no other available alternatives from what I see.

IMO it makes sense to get rid of this `store` + `share` in the future - and change it w/ Context API (or smt). Don't think it makes sense to do it here though - I'd suggest to do it after this revamp is done. Lmk what you think:)

shonsirsha:
rebased:)

thisconnect:
agree, keep using share+store for fiat/rates in staging-revamp-settings and refactor to context after.

> IMO it makes sense to get rid of this store + share in the future - and change it w/ Context API (or smt). Don't think it makes sense to do it here though - I'd suggest to do it after this revamp is done. Lmk what you think:)

frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
```diff
@@ -11,13 +12,15 @@ type SelectOption = {
 
 type TSelectProps = {
   options: SelectOption[];
-  selectedCurrencies: SelectOption[];
-}
+} & Omit<SharedProps, 'btcUnit'>
```
thisconnect, line 15: I was quickly confused by this, alternative maybe make `btcUnit` optional in SharedProps?

but I am not sure, but TypeScript seems to be happy. Anyway this way seems to be good too.

thisconnect (approved):
tested LGTM 👍 

commented about SharedProps type maybe make it optional instead of using omit, but I am not sure.

frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
shonsirsha: Just didn't want to mess with "old" code and making btcUnitan optional - worried of breaking something. But seems like it's fine. I agree, making btcUnit optional will most likely be clearer at first glance. Will do this :)  Thanks!

PR #2082, opened 2023-04-26
Btc amount spacing

Beerosagos:
Current proposal: using spacing for BTC/LTC fractional part everywhere except for balance and chart, since the trailing zeroes are removed there (see pics).

image

image

immagine

### DIFF
--- a/frontends/web/src/components/amount/amount.test.tsx
+++ b/frontends/web/src/components/amount/amount.test.tsx
@@ -100,27 +100,46 @@ describe('Amount formatting', () => {
         const { container } = render(<Amount amount="10.00000000" unit={coin} removeBtcTrailingZeroes/>);
         expect(container).toHaveTextContent('10');
       });
-      it('10.12300000 ' + coin + ' with removeBtcTrailingZeroes enabled becomes 10.123', () => {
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/amount/amount.test.tsx
```diff
@@ -108,13 +108,21 @@ describe('Amount formatting', () => {
...
+        const blocks = getByTestId('amountBlocks');
+        expect(blocks.innerHTML).toBe(
+          '<span>10.00</span>' +
+          '<span class="space">000</span>' +
+          '<span class="space">000</span>');
       });
-      it('10.12300000 ' + coin + ' with removeBtcTrailingZeroes disabled stays 10.12300000', () => {
-        const { container } = render(<Amount amount="10.12300000" unit={coin}/>);
-        expect(container).toHaveTextContent('10.12300000');
+      it('10.12300000 ' + coin + ' with removeBtcTrailingZeroes disabled gets spaced', () => {
```
benma, line 119: pls add a case or modify a case where the part before the decimal point is large, e.g. 12345.67800000`

maybe also one where all decimals are used: `10.12345678`

benma (approved):
utACK

PR #2083, opened 2023-04-26
Export btc

Beerosagos:
This updates `ExportCSV` function to follow user preference for BTC
amount unit, instead of always using sats.


benma:
This could break cointracking which uses this CSV to import the data for their users.

PR #2086, opened 2023-04-27
Accounts discovery

benma:


Beerosagos:
First 3 commits looks good to me. I'm still looking into the code for the others, but with the first tests I noticed an inconsistency in the account summary view (see pic). 

What I did:
1. checkout master
2. remove appfolder and start the app in dev mode
3. go to account summary 
4. stop the app
5. checkout the PR
6. start again the app -> the transition between the existing persisted accounts and the discovery doesn't look correct

Am I doing smth wrong?

immagine

Beerosagos:
> EDIT: Tested again, properly restarting both the backend and the frontend. Everything seems to work, so I'd say that this is not a real problem.

Tested again this morning on Android, got the same weird behavior. It was not meant to be a test for this PR, just needed to compile account discovery on testnet for Android, so here is what I did:
- rebased onto #2100 to be able to compile with the latest docker image
- forced `testnet = true` here https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/android/goserver/goserver.go#L180

I don't see how these changes should break this tho. Closing the app and opening again fixed the issue. 

image

Beerosagos (changes requested):
Ok, I think I found the problem:
`backend.Accounts()` sorts the account list while locking the Read lock (see https://github.com/benma/bitbox-wallet-app/blob/accountsync2/backend/backend.go#LL429C16-L429C17), which doesn't seem correct. Moving the `sortAccounts` call inside  `backend.addAccount` seems to solve the issue. 
Locking the write lock inside `backend.Accounts` seems to cause a deadlock, and doesn't look good anyway, imho.

A couple notes: 
- with the suggested modification I could not reproduce the issue again, but I made a lot of tests removing the accounts.json file and restarting the servewallet. I noticed that sometimes the bb02 seems to take longer in deriving the keys (I can tell from the bb02 screen saver staying still instead of moving around), sometimes up to a minute or so. I thought of another race condition, but it apparently solves by itself.  It also happened once that the bb02 apparently aborted some request and needed to be unplugged and replugged (no need to restart the servewallet) to get the app working again. No idea why.
- I noticed that the number of scanned addresses is not always shown by the frontend, I didn't investigate it and neither checked if it is related to this PR or if that happens also on master.

benma:
Thanks @Beerosagos !!

I fixed the race condition and added unit tests. While adding the unit tests I discovered that adding a new account in the settings does not set the name correctly if a previously added hidden account is activated - I fixed that too.

PR #2088, opened 2023-05-01
Frontend revamp settings tabs

shonsirsha:
This PR is mainly to set-up the Tabs view that we have on Desktop, as well as the "Back" button when on mobile.

For each changes, please see the commits history.

Preview (desktop):
Screenshot 2023-04-24 at 18 44 53

Preview (mobile) _(still missing the "arrow left" / "chevron left" icon to be put on the left of the "Back" button):_
Screenshot 2023-04-24 at 18 45 29

### DIFF
--- a/frontends/web/src/routes/new-settings/components/tabs.tsx
+++ b/frontends/web/src/routes/new-settings/components/tabs.tsx
@@ -1,0 +1,79 @@
+import { ReactNode } from 'react';
+import { MobileHeader } from './mobile-header';
+import { NavLink } from 'react-router-dom';
+import { route } from '../../../utils/route';
+import { SettingsItem } from './settingsItem/settingsItem';
+import { ChevronRightDark } from '../../../components/icon';
+import { useTranslation } from 'react-i18next';
+import styles from './tabs.module.css';
+
+type TWithSettingsTabsProps = {
+  children: ReactNode
+  deviceIDs: string[]
+  hideMobileMenu?: boolean;
+  subPageTitle: string;
+}
+
+type TTab = {
+  name: string;
+  url: string;
+  hideMobileMenu?: boolean;
+}
+
+type TTabs = {
+  deviceIDs: string[];
+  hideMobileMenu?: boolean;
+}
+
+export const WithSettingsTabs = ({ children, deviceIDs, hideMobileMenu, subPageTitle }: TWithSettingsTabsProps) => {
+  return (
+    <>
+      <div className="show-on-small">
+        <MobileHeader subPageTitle={subPageTitle} />
+      </div>
+      <div className="hide-on-small">
+        <Tabs hideMobileMenu={hideMobileMenu} deviceIDs={deviceIDs} />
+      </div>
+      {children}
+    </>
+  );
+};
+
+export const Tab = ({ name, url, hideMobileMenu }: TTab) => {
+
+  if (!hideMobileMenu) {
+    // Will only be shown on mobile (index/general settings page)
+    return (
+      <div key={url} className="show-on-small">
+        <SettingsItem
+          settingName={name}
+          onClick={() => route(url)}
+          extraComponent={<ChevronRightDark/>} />
+      </div>
+    );
+  }
+
+  return (
+    <NavLink
+      className={({ isActive }) => isActive ? `${styles.active} hide-on-small` : 'hide-on-small'}
+      to={url}
+      key={url}>
+      {name}
+    </NavLink>
+  );
+};
+
+export const Tabs = ({ deviceIDs, hideMobileMenu }: TTabs) => {
+  const { t } = useTranslation();
+  return (
+    <div className={styles.container}>
+      <Tab hideMobileMenu={hideMobileMenu} name={t('settings.appearance')} url="/new-settings/appearance" />
+      <Tab hideMobileMenu={hideMobileMenu} name={'Manage accounts'} url="/new-settings/manage-accounts"/>
+      {deviceIDs.map(id => (
+        <Tab hideMobileMenu={hideMobileMenu} name={'Device settings'} key={id} url={`/device/${id}`} />
+      )) }
+      <Tab hideMobileMenu={hideMobileMenu} name={'Advanced settings'} url="/new-settings/advanced-settings" />
+      <Tab hideMobileMenu={hideMobileMenu} name={'About'} url="/new-settings/about" />
+    </div>
+  );
+};
--- a/frontends/web/src/routes/router.tsx
+++ b/frontends/web/src/routes/router.tsx
@@ -20,6 +20,7 @@ import { Passphrase } from './device/bitbox02/passphrase';
 import { Account } from './account/account';
 import { ReceiveAccountsSelector } from './accounts/select-receive';
 import { Appearance } from './new-settings/appearance';
+import { MobileSettings } from './new-settings/mobile-settings';
 
 type TAppRouterProps = {
     devices: TDevices;
@@ -110,6 +111,18 @@ export const AppRouter = ({ devices, deviceIDs, devicesKey, accounts, activeAcco
     devices={devices}
   /></InjectParams>;
 
+  const MobileSettingsEl = <InjectParams>
+    <MobileSettings
+      deviceIDs={deviceIDs}
+    />
+  </InjectParams>;
+
+  const AppearanceEl = <InjectParams>
+    <Appearance
+      deviceIDs={deviceIDs}
+    />
+  </InjectParams>;
+
   const ReceiveAccountsSelectorEl = <InjectParams><ReceiveAccountsSelector activeAccounts={activeAccounts}/></InjectParams>;
 
   return <Routes>
@@ -142,7 +155,8 @@ export const AppRouter = ({ devices, deviceIDs, devicesKey, accounts, activeAcco
         <Route path="manage-accounts" element={<ManageAccounts key={'manage-accounts'} />} />
       </Route>
       <Route path="new-settings">
-        <Route path="appearance" element={<Appearance />} />
+        <Route index element={MobileSettingsEl} />
+        <Route path="appearance" element={AppearanceEl} />
       </Route>
     </Route>
   </Routes>;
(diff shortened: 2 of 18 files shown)


shonsirsha:
Messed the commits in the old branch (PR): - so reopened this new one with the same exact changes/commits. Would you be able to re-approve? @Beerosagos 

Many thanks 🙏

frontends/web/src/routes/new-settings/components/tabs.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <SettingsItem
+          settingName={tab.name}
+          onClick={() => route(tab.url)}
+          key={tab.url}
+          extraComponent={<RightChevron />}
+        />
+      ))}
+    </>
+  );
+};
```
shonsirsha, line 73: @thisconnect Here I didn't re-style the Tabs on mobile. Instead, I'll use `SettingsItem` for this. 

My explanation:
Re-styling tabs on mobile will actually cause redundancy because we already have this same exact component called `SettingsItem`. So there's no use / doesn't make sense to re-create that component again by re-styling tabs when we can already use the already available `SettingsItem`.

However, I used the `settingsTabsDetail` data which is also used for the tabs on desktop.  

**So this way, there's only 1 source of truth of the data both for tabs displayed on desktop, and the settings on mobile.**

IMO, this is the best way to achieve the mobile settings and the desktop tabs because:

1. We're using the already created component (SettingsItem)
2. Both (desktop) Tabs and MobileSettings use the same "source"  / "data" (`settingsTabsDetail`).

WDYT? 😄

frontends/web/src/routes/new-settings/components/tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
```
shonsirsha, line 65: @thisconnect Here I didn't re-style the Tabs on mobile. Instead, I'll use `SettingsItem` for this. 

My explanation:
Re-styling tabs on mobile will actually cause redundancy because we already have this same exact component called `SettingsItem` (per design). So there's no use / doesn't make sense to re-create that component again by re-styling tabs when we can already use the already available `SettingsItem`.

However, I used the `settingsTabsDetail` data which is also used for the tabs on desktop.  

**So this way, there's only 1 source of truth of the data both for tabs displayed on desktop, and the settings on mobile.**

IMO, this is the best way to achieve the mobile settings and the desktop tabs because:

1. We're using the already created component (SettingsItem)
2. Both (desktop) Tabs and MobileSettings use the same "source"  / "data" (`settingsTabsDetail`).

WDYT? 😄

shonsirsha:
rebased:)

frontends/web/src/routes/new-settings/components/tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export type TTab = {
+  url: string;
+  name: string;
+}
+
+const settingsTabsDetail: TTab[] = [
+  { url: '/new-settings/appearance', name: i18n.t('settings.appearance') },
+  { url: '/new-settings/manage-accounts', name: 'Manage accounts' },
+  { url: '/new-settings/device-settings', name: 'Device settings' },
```
thisconnect, line 28: keep in mind, 'Device settings' should only be shown when there is a device plugged in and unlocked. 

Also the logic currently would in theory support multiple deivces. Please take the same logic

https://github.com/digitalbitbox/bitbox-wallet-app/blob/697699f6f874ef0ecafb33ef3ed33629faf0a6ec/frontends/web/src/components/sidebar/sidebar.tsx#L212-L225

thisconnect (changes requested):

frontends/web/src/routes/new-settings/components/tabs.tsx
shonsirsha: Good catch, a missing piece of info for me here. Thanks!

frontends/web/src/routes/new-settings/components/tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </div>
+          <div className="hide-on-small">
+            <Tabs settingsTabsDetail={settingsTabsDetail} />
+          </div>
+          {children}
+        </ViewContent>
+      </View>
+    </Main>)
+  ;
+};
```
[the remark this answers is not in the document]
shonsirsha: Yeah, sure. Just thought it's neater - but it looks like an anti-pattern here. Will make the change.

frontends/web/src/components/icon/assets/icons/chevron-left-light.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="8" height="12" viewBox="0 0 8 12" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M7.43753 1.41058V0.202767C7.43753 0.0980799 7.31721 0.0402674 7.23596 0.10433L0.192206 5.60589C0.132366 5.65243 0.083938 5.71203 0.050627 5.78013C0.017316 5.84824 0 5.92305 0 5.99886C0 6.07467 0.017316 6.14949 0.050627 6.21759C0.083938 6.28569 0.132366 6.34529 0.192206 6.39183L7.23596 11.8934C7.31878 11.9575 7.43753 11.8996 7.43753 11.795V10.5871C7.43753 10.5106 7.40159 10.4371 7.34221 10.3903L1.71721 5.99964L7.34221 1.60745C7.40159 1.56058 7.43753 1.48714 7.43753 1.41058Z" fill="white"/>
+</svg>
```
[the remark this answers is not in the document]
thisconnect: > Yes, the left chevron is used for back button and the right chevron is used for the settingitem. So I think they're supposed to be different.

they should be flipped 180° but **not** have different shape/color etc.

frontends/web/src/routes/router.tsx
```diff
@@ -110,6 +111,18 @@ export const AppRouter = ({ devices, deviceIDs, devicesKey, accounts, activeAcco
...
+      deviceIDs={deviceIDs}
+    />
+  </InjectParams>;
+
+  const AppearanceEl = <InjectParams>
+    <Appearance
+      deviceIDs={deviceIDs}
+    />
+  </InjectParams>;
+
```
shonsirsha, line 125: @thisconnect  wdyt of this approach of passing the deviceIDs to all of the sub-pages of the settings? 

This (deviceIDs) is actually, for now anyway, only used for the tabs whether to render device setting or not. Would be cleaner if we can somehow jsut get the deviceIDs inside the `WithSettingsTab`.

frontends/web/src/routes/new-settings/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+      // the rest of the tabs
+      // after all of the device setting tab(s)
+      ...settingsTabsDetail.slice(2)
+    ];
+
+    return settingsTabsWithDeviceSettings;
+  }
+
+  return settingsTabsDetail;
+};
```
thisconnect, line 34: ignoring how the component is built right now, I'd slightly prefer something like this:

```jsx
export const settingsTabsBuilder = (deviceIDs: string[]) => {
  return (
    <>
      <tab url="/new-settings/appearance">{t('settings.appearance')}</tab>
      <tab url="/new-settings/manage-accounts">Manage accounts</tab>
      {deviceIDs.map(id => (
        <tab url={`/device/${id}`}>Device settings</tab>
      )) }
      <tab url="/new-settings/advanced-settings">Advanced settings</tab>
      <tab url="/new-settings/about">About</tab>
    </>
  );
};
```

thisconnect: I was in a hurry a bit when reviewing before. there is also condition to render "Manage accounts" only when there is an unlocked BB02. But not a blocker can be done later.

frontends/web/src/routes/router.tsx
thisconnect: gutfeeling, the less react-router we use the easier it will be to upgrade next major version 🙈 .

I think deviceIDs should only be required for "Device settings", but probably it is also used in another tab.

frontends/web/src/routes/new-settings/components/tabs.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <ViewContent>
+      <div className="show-on-small">
+        <MobileHeader settingsTabsDetail={tabs} />
+      </div>
+      <div className="hide-on-small">
+        <Tabs settingsTabsDetail={tabs} />
+      </div>
+      {children}
+    </ViewContent>
```
thisconnect, line 35: nit: export a fragment and move ViewContent to the route.

thisconnect (changes requested):

frontends/web/src/routes/router.tsx
shonsirsha: >gutfeeling, the less react-router we use the easier it will be to upgrade next major version 🙈 .
>I think deviceIDs should only be required for "Device settings", but probably it is also used in another tab.


Atm, deviceIDs are used in the `Tabs` component (new one, following the approach you suggested). And tabs are used in each subpage(s) within its wrapper `WithSettingsTabs` which gets imported in all of the (sub-)pages. So this made us require `deviceIDs` in all of those (sub-)pages...

frontends/web/src/routes/new-settings/components/tabs.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <NavLink
+      className={({ isActive }) => isActive ? `${styles.active} hide-on-small` : 'hide-on-small'}
+      to={url}
+      key={url}>
+      {name}
+    </NavLink>
+  );
+};
```
shonsirsha, line 64: We don't have to use `isMobile` here which is nice:)

thisconnect (approved):
thank you! tested LGTM 👍

thisconnect:
since this is more like 1 change to the interface, I think you could squash the commits into 1 commit.

shonsirsha:
> 
since this is more like 1 change to the interface, I think you could squash the commits into 1 commit.


Sounds good.

PR #2089, opened 2023-05-01
frontend: single dropdown styling and refactoring

shonsirsha:
Refactored and re-styled the **single dropdown** component for the newly redesigned / revamped settings. This is so that we can re-use and already are re-using the this **single dropdown** component in this PR. 

We'll do the same for multi dropdown component (which is used for selecting active currencies) **in another PR.**

For detailed changes, please refer to the commit history.

Preview (light):
Screenshot 2023-05-01 at 17 44 48

Preview (dark):
Screenshot 2023-05-01 at 17 44 42

### DIFF
--- a/frontends/web/src/routes/new-settings/components/appearance/defaultCurrencyDropdownSetting.tsx
+++ b/frontends/web/src/routes/new-settings/components/appearance/defaultCurrencyDropdownSetting.tsx
@@ -1,32 +1,6 @@
+import { setActiveFiat, store } from '../../../../components/rates/rates';
+import { SingleDropdown } from '../singledropdown/singledropdown';
 import { SettingsItem } from '../settingsItem/settingsItem';
-import Select from 'react-select';
-import { store, setActiveFiat } from '../../../../components/rates/rates';
-import { Fiat } from '../../../../api/account';
-import styles from './defaultCurrencySetting.module.css';
-
-type SelectOption = {
-  label: Fiat;
-  value: Fiat;
-}
-
-type TSelectProps = {
-  options: SelectOption[];
-  handleChange: (fiat: Fiat) => void;
-}
-
-const ReactSelect = ({ options, handleChange }: TSelectProps) => <Select
-  className={styles.select}
-  classNamePrefix="react-select"
-  defaultValue={{ label: store.state.active, value: store.state.active }}
-  isSearchable={true}
-  onChange={(selected) => {
-    if (selected) {
-      handleChange(selected.value as Fiat);
-    }
-  }
-  }
-  options={options}
-/>;
 
 export const DefaultCurrencyDropdownSetting = () => {
   const formattedCurrencies = store.state.selected.map((currency) => ({ label: currency, value: currency }));
@@ -35,7 +9,13 @@ export const DefaultCurrencyDropdownSetting = () => {
     <SettingsItem
       settingName="Default Currency"
       secondaryText="Select your default currency."
-      extraComponent={<ReactSelect options={formattedCurrencies} handleChange={setActiveFiat}/>}
+      extraComponent={
+        <SingleDropdown
+          options={formattedCurrencies}
+          handleChange={setActiveFiat}
+          defaultValue={{ label: store.state.active, value: store.state.active }}
+        />
+      }
     />
   );
 };
--- a/frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
+++ b/frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
@@ -1,0 +1,50 @@
+.dropdown {
+    appearance: none;
+    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' fill='none' stroke='%23777' stroke-linecap='round' stroke-linejoin='round' stroke-width='2' aria-labelledby='chevron down' color='%23777'%3E%3Cdefs/%3E%3Cpolyline points='6 10 12 16 18 10'/%3E%3C/svg%3E%0A");
+    background-repeat: no-repeat;
+    font-weight: 400;
+    height: calc(var(--space-quarter) * 3);
+    padding: 0 calc(var(--space-quarter) + var(--space-eight));
+}
+
+.select {
+    width: 280px;
+}
+
+.select :global(.react-select__menu) {
+    background-color: var(--background-secondary);
+}
+
+.select :global(.react-select__option) {
+    background-color: var(--background-secondary);
+}
+
+.select :global(.react-select__option):hover {
+    background-color: var(--background-custom-select-hover);
+}
+
+.select :global(.react-select__option--is-selected),
+.select :global(.react-select__option--is-selected):hover {
+    background-color: var(--background-custom-select-selected);
+}
+
+.select :global(.react-select__control) {
+    background-color: var(--background-secondary);
+    border: 2px solid var(--color-custom-select-border);
+    border-radius: 0;
+    padding: 0 var(--space-eight);
+}
+
+.select :global(.react-select__control):hover {
+    cursor: pointer;
+}
+
+.valueContainer {
+    font-size: var(--size-default);
+    color: var(--color-default);
+    padding: 8px 0;
+}
+
+.select :global(.react-select__option--is-selected) .optionValue {
+    color: var(--color-alt);
+}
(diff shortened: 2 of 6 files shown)


thisconnect:
nice code LGTM, please squash into 1 commit as this is kind of 1 change to the UI.

ui question: should the border of the dropdown be white in darkmode? cc @jadzeidan 

dropdown on send view has no border, just the icon:

Screen Shot 2023-05-03 at 11 46 15

jadzeidan:
> nice code LGTM, please squash into 1 commit as this is kind of 1 change to the UI and rebase.

> ui question: should the border of the dropdown be white in darkmode? cc @jadzeidan

> dropdown on send view has no border, just the icon:

@thisconnect the dropdown on the send view is clear, because it is a different color to the background, so it doesn't need a boarder.

On the settings page, I think it makes sense to add a border to the dropdown but using the color `color-dark` (which would be var `--background`)

shonsirsha:
> nice code LGTM, please squash into 1 commit as this is kind of 1 change to the UI and rebase.


Rebased, change(s) made in regards to border, and squashed into 1 commit.  PTAL 🙏

jadzeidan:
> Thanks both of you!
> 
> @jadzeidan , this is how it looks like using `--background` on darkmode. Do you think it looks ok?
> 
> Screenshot 2023-05-03 at 17 00 14

Thanks, looks good to me. 

Maybe would be better if the border was the same color as the chevron. But changing the color variable would mess up the light mode version since light mode uses `--background` too (which looks good on light mode), so we would need to introduce a new variable if want to change the boarder color and keep the code clean. But for now I think the dark border looks fine.

jadzeidan:
@shonsirsha yeah that looks better. So let's use that color if you are adding a new variable anyways. :)

shonsirsha:
> @shonsirsha yeah that looks better. So let's use that color if you are adding a new variable anyways. :)

Great!  @thisconnect feel free to review (again? 😉 )

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
```diff
@@ -0,0 +1,50 @@
...
+    border-radius: 0;
+    padding: 0 var(--space-eight);
+}
+
+.select :global(.react-select__control):hover {
+    cursor: pointer;
+}
+
+.valueContainer {
+    font-size: 16px;
```
thisconnect, line 43: isn't there a CSS variable that you can use for the font-size here?

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+        if (selected) {
+          const value = (selected as SingleValueType<any>)?.value || '';
+          handleChange(value);
+        }
+      }
+      }
+      options={options}
+    />
+  );
+};
```
thisconnect, line 66: there is many `any` here, I haven't really looked into but could you check if it is possible to type it?

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
```diff
@@ -0,0 +1,50 @@
+.dropdown {
+    appearance: none;
+    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' fill='none' stroke='currentColor' stroke-linecap='round' stroke-linejoin='round' stroke-width='2' aria-labelledby='chevron down' color='%23777'%3E%3Cdefs/%3E%3Cpolyline points='6 10 12 16 18 10'/%3E%3C/svg%3E%0A");
```
thisconnect, line 3: `currentColor` only works if the SVG elements are in the DOM (afaik), but not over CSS. Could you hardcode the color?

thisconnect (changes requested):

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.tsx
shonsirsha: I'll see if we could use generics.

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
shonsirsha: Actually, we do. It's

```css
   /* wizard */
    --size-wizard-text: 16px;
```

sounds a bit specific (the naming) so I refrain from using it. Should we use this?

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
shonsirsha: It does work it seems like.. Without current colour it'll be not visible. Also tried changing it to red and the chevron changed to red, too:)

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
jadzeidan: I think 14px is good. That's the same size we have for the current currencies.

thisconnect: Looking at variables.css I think we should cleanup this up globally, not in this PR.

14px is ok, normal text input fields seem to also use 14px.

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
thisconnect: yes red is a color, but currentColor is like CSS inherit, but that doens't work if the svg is in backrgound-imag data url, only if it is in the dom.

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
shonsirsha: > Looking at variables.css I think we should cleanup this up globally, not in this PR.

Yeah, not going to modify already existing variables here thus going with `--size-default`  now (14px).  Had the impression of the text in the design being 16px not 14px..  but clearly 14px it is:)

thisconnect: > Had the impression of the text in the design being 16px not 14px.. but clearly 14px it is:)

yeah me too, I think in view.modules.css it is bigger (18px). Generally we should move to bigger font one day (soonish).

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
shonsirsha: Sorry, I understood what you meant. I was thinking the `stroke` attribute. But im not sure what you meant either by "doesn't work" here because it definitely inherits something, as leaving it blank will make the chevron invisible.
Screenshot 2023-05-08 at 12 55 05

Anyway, doesn't really matter i guess. But I went with your suggestion w/ hardcoding it with `#7777` to be safe. :)

thisconnect: gray must be some kind of fallback, I made a quick and dirty test, inline svg vs. css background with svg
both use currentColor, the parent defines green color. the inline svg inherits green. the css background doesn't, as it can't I think as it doesn't know about the context. same with external svg. 

https://jsfiddle.net/2scndvbe/

shonsirsha: Oh haha, such a neat coincidence that it fallbacks to the exact gray we're using too! :) Thanks Eni 👍

frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.tsx
thisconnect: it would be nice if it is more or less straight forward to do. else we can also merge as is.

shonsirsha:
@thisconnect  Could you give another look here? Thanks! 🙏

frontends/web/src/routes/new-settings/components/appearance/defaultCurrencyDropdownSetting.tsx
```diff
@@ -1,32 +1,7 @@
...
-  isSearchable={true}
-  onChange={(selected) => {
-    if (selected) {
-      handleChange(selected.value as Fiat);
-    }
-  }
-  }
-  options={options}
-/>;
+import { store } from '../../../../components/rates/rates';
```
thisconnect, line 4: nit:  use  same import as setActiveFiat

thisconnect (approved):
👍  untested LGTM, with small nit

PR #2090, opened 2023-05-03
frontend: fix regression sdcard-inserted-check too early

thisconnect:
This fixes 'handshake must come first' error, the sdcard check should only happen after the bitbox02 is paired.


benma (approved):
utACK

PR #2093, opened 2023-05-03
frontend: convert utils config to typescript

thisconnect:


shonsirsha (approved):
lgtm untested.

We have 1 more `setConfig` which does the same thing but doesnt return the new config and requires us to pass the current config, right?

Anyways, probably a good reason to rename either one of these lol. Also I wonder if it's possible to stick to one of these instead..

thisconnect:
yeah I know we have 2 setConfig's. the one here is saver to use becuase if it is for whatever reason taking longer and called again, it will keep a temporary config and ensure that all data is written if called multiple times.

> Anyways, probably a good reason to rename either one of these lol. Also I wonder if it's possible to stick to one of these instead

yep I agree, but this PR is only for converting to typescript.

PR #2094, opened 2023-05-04
frontend/android: update targetSdkVersion to 33

Beerosagos:
We need to update target SDK version because of this:

```
Starting August 31, 2023:

    New apps and app updates must target API level 33 to be submitted to Google Play (Wear OS must target API 30).
    Existing apps must target API level 31 or above to remain discoverable by all users on Google Play. Apps that target API level 30 or below (target API level 29 or below for Wear OS), will only be discoverable on devices running Android OS same or lower than your apps’ target API level.
```

### DIFF
--- a/frontends/android/BitBoxApp/app/build.gradle
+++ b/frontends/android/BitBoxApp/app/build.gradle
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.application'
 
 android {
-    buildToolsVersion "29.0.0"
-    compileSdkVersion 29
+    buildToolsVersion "33.0.0"
+    compileSdkVersion 33
     defaultConfig {
         applicationId "ch.shiftcrypto.bitboxapp"
         minSdkVersion 21
-        targetSdkVersion 31
+        targetSdkVersion 33
         versionCode 42
         versionName "android-4.37.0"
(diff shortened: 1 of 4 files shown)


frontends/android/BitBoxApp/app/build.gradle
```diff
@@ -1,12 +1,12 @@
 apply plugin: 'com.android.application'
 
 android {
-    buildToolsVersion "29.0.0"
-    compileSdkVersion 29
+    buildToolsVersion "33.0.0"
+    compileSdkVersion 33
```
benma, line 5: How come these had to be changed this time but not the previous times?

https://github.com/digitalbitbox/bitbox-wallet-app/commit/3d361b4891e90edd63bbf76a85fbef1905b42905
https://github.com/digitalbitbox/bitbox-wallet-app/commit/bba9423ae362f25ba60e351573bef582b8ad5db8

benma (approved):
tACK

frontends/android/BitBoxApp/app/build.gradle
Beerosagos: Android Studio was warning about having `compileSdkVersion` lower than `targetSdkVersion`. `buildToolsVersion` follows `compileSdkVersion`.

PR #2095, opened 2023-05-04
backend/handlers: add notes to utxo

Beerosagos:
This adds notes to utxo struct, to be displayed when using coin control feature.


benma (approved):
utACK

PR #2097, opened 2023-05-08
frontend: created About page for the new settings page

shonsirsha:
Preview (Desktop):
Screenshot 2023-05-08 at 09 17 12

Preview (Mobile):
Screenshot 2023-05-08 at 09 21 29

### DIFF
--- a/frontends/web/src/routes/new-settings/about.tsx
+++ b/frontends/web/src/routes/new-settings/about.tsx
@@ -1,0 +1,25 @@
+import { useTranslation } from 'react-i18next';
+import { Main, Header } from '../../components/layout';
+import { View, ViewContent } from '../../components/view/view';
+import { WithSettingsTabs } from './components/tabs';
+import { AppVersion } from './components/about/app-version-setting';
+
+type TProps = {
+  deviceIDs: string[]
+}
+
+export const About = ({ deviceIDs }: TProps) => {
+  const { t } = useTranslation();
+  return (
+    <Main>
+      <div className="hide-on-small"><Header title={<h2>{t('sidebar.settings')}</h2>} /></div>
+      <View fullscreen={false}>
+        <ViewContent>
+          <WithSettingsTabs subPageTitle={t('settings.about')} hideMobileMenu deviceIDs={deviceIDs}>
+            <AppVersion />
+          </WithSettingsTabs>
+        </ViewContent>
+      </View>
+    </Main>
+  );
+};
--- a/frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
+++ b/frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
@@ -1,0 +1,13 @@
+.version {
+    color: var(--color-secondary);
+    font-size: 14px;
+    margin: 0;
+    margin-right: 10px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.versionContainer {
+    display: flex;
+    align-items: center;
+}
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1292,6 +1292,7 @@
     "transactionDetails": "Transaction details"
   },
   "settings": {
+    "about": "About",
     "accounts": "Accounts",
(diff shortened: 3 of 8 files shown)


frontends/web/src/routes/new-settings/about.tsx
```diff
@@ -0,0 +1,25 @@
+import { useTranslation } from 'react-i18next';
+import { Main, Header } from '../../components/layout';
+import { View, ViewContent } from '../../components/view/view';
+import { WithSettingsTabs } from './components/tabs';
+import { AppVersion } from './components/about/app-version-setting';
+
+type TProps = {
+    deviceIDs: string[]
+  }
```
thisconnect, line 9: use 2 spaces, also do you know if there is a way to tell eslint to also parse typescript?

frontends/web/src/routes/new-settings/about.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export const About = ({ deviceIDs }: TProps) => {
+  const { t } = useTranslation();
+  return (
+    <Main>
+      <div className="hide-on-small"><Header title={<h2>{t('sidebar.settings')}</h2>} /></div>
+      <View fullscreen={false}>
+        <ViewContent>
+          <WithSettingsTabs subPageTitle={'Appearance'} hideMobileMenu deviceIDs={deviceIDs}>
```
thisconnect, line 18: subPageTitle should be About, please also add to en/app.json

frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
```diff
@@ -0,0 +1,16 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.versionContainer {
+    display: flex;
+    align-items: center;
+}
+
+.versionContainer > p {
```
thisconnect, line 14: nit: does it need to do direct child, or could it just be in styles.version ?

thisconnect (changes requested):

frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
```diff
@@ -0,0 +1,16 @@
+.version {
+    color: var(--color-secondary);
+    font-size: 14px;
```
thisconnect, line 3: forgot: isn't there a CSS variable that you can use for fontsize?

shonsirsha: Not exactly for 14px. We can go for 16 though, although it'd be a bit too big.. just a bit.

frontends/web/src/routes/new-settings/about.tsx
shonsirsha: > subPageTitle should be About

Lol. True

> please also add to en/app.json

There are many that are still not there, I can start adding them I guess..

frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
shonsirsha: yeah.. good catch:)

frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
thisconnect: I checked there is many hard coded 14px font-sizes.

Let's keep as is, one day we should look at font sizes globally and check if it should be rendered with bigger font on large screens.

shonsirsha:
Thanks for the review @thisconnect , PTAL 🙏

thisconnect:
LGTM

thisconnect (approved):

frontends/web/src/routes/new-settings/components/about/app-version-setting.module.css
shonsirsha: Actually we do have a 14px called `--size-default`, but this gets overriden by this class

```css
.inner p {
        --size-default: 20px;
    }
```

in `view.module.css`. So that's prob why we have a lot of 14px hardcoded, including here.

PR #2102, opened 2023-05-08
draft: frontend setup with recovery words only

thisconnect:
Add support for creating a wallet without a microSD card backup using only recovery words.

Opt-in to words is currently a toggle under create wallet, but this is subject to change and might get burried in some expert config.

Also create backup with sdcard has a warning step with a list of checkpoints. This is currently not shown for creating with words.

Also changed to go back to choice of create and restore from when user aborts the process.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -228,7 +228,8 @@
     },
     "stepBackup": {
       "beforeProceed": "Before proceeding, please read these important security considerations:",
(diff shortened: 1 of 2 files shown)


thisconnect:
probably better to cleanup https://github.com/digitalbitbox/bitbox-wallet-app/pull/2111 first

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/2137

PR #2104, opened 2023-05-10
frontend: created advanced settings tabs content for settings revamp

shonsirsha:
Preview desktop:
Screenshot 2023-05-10 at 08 46 23

Preview mobile:
Screenshot 2023-05-10 at 08 46 25


frontends/web/src/routes/new-settings/advanced-settings.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const AdvancedSettings = ({ deviceIDs, hasAccounts }: TPagePropsWithSettingsTabs) => {
+  const { t } = useTranslation();
+  const fetchedConfig = useLoad(getConfig);
+  const [config, setConfig] = useState<any>();
+  const [showRestartMessage, setShowRestartMessage] = useState(false);
+
+  useEffect(() => {
+    setConfig(fetchedConfig);
+  }, [fetchedConfig]);
```
shonsirsha, line 22: Fetching the config here **once** since it's being used in two components in the page (instead of calling the same exact thing twice from 2 diff setting item components). 

Ive been trying to minimise states at the page-level and just have them inside of each setting item to have a neater code, but IMO this one makes sense to have it here.

shonsirsha:
rebased:)

frontends/web/src/routes/new-settings/components/advanced-settings/enable-tor-proxy-setting.module.css
```diff
@@ -0,0 +1,13 @@
...
+    margin: 0;
+    margin-right: var(--space-quarter);
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.container {
+    display: flex;
+    align-items: center;
+}
```
thisconnect, line 13: is that the only component that uses text+ChevronRightDark, else maybe it's worth having an other prop for the text in SettingsItem and moving this CSS to SettingsItem.

frontends/web/src/routes/new-settings/advanced-settings.tsx
thisconnect: not important rightnow but what do you think of, instead of passing the whole config to the components, only pass the relevant part?

also I am wondering if it is really needed to receive updates via onChangeConfig and pass config down again... but no blocker. keep as is or feel free to change.

frontends/web/src/routes/new-settings/components/advanced-settings/tor-proxy-dialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const { success, errorMessage } = result;
+
+    if (success) {
+      setProxyConfig(proxy);
+    } else {
+      alertUser(errorMessage || t('account.fatalError'));
+    }
+  };
+
+  const setProxyConfig = async (proxyConfig: any) => {
```
thisconnect, line 48: config layer is not typed, but I wonder if it would make sense to add minimal type for proxyConfig in this component.

frontends/web/src/routes/new-settings/components/advanced-settings/tor-proxy-dialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+    onChangeConfig(config);
+    handleShowRestartMessage(true);
+  };
+
+  const handleToggleProxy = () => {
+    if (!config) {
+      return;
+    }
+    const proxy = { ...config.backend.proxy, useProxy: !proxyEnabled };
+    setProxyConfig(proxy);
```
thisconnect, line 62: setProxyConfig is async.. shouldn't it do await here or doesn't matter?

I am not sure, old implementation doesn't have proxyEnabled state, the toggle is directly using the config.

thisconnect (commented):
nice untested LGTM with some comments to consider.

jadzeidan:
> one more comment about UI, on dekstop there is some space between items which makes sense, but on mobile there is none.
> 
> how about adding at tiny 1-2px space on mobile as well so it helps to visually separate the settings items.
> 
> cc @jadzeidan

Sure, fine with me.

frontends/web/src/routes/new-settings/advanced-settings.tsx
```diff
@@ -0,0 +1,45 @@
...
+            <EnableCustomFeesToggleSetting config={config} />
+            <EnableTorProxySetting config={config} onChangeConfig={setConfig} handleShowRestartMessage={setShowRestartMessage} />
+            <ConnectFullNodeSetting />
+            {showRestartMessage ?
+              <InlineMessage
+                type="success"
+                align="left"
+                message={t('settings.restart')}
+                onEnd={() => setShowRestartMessage(false)}
+              /> : null}
```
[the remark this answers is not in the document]
shonsirsha: Technically, it's possible to just put it inside the `EnableTorProxySetting` and put it (visually) beforethe EnableTorProxySetting setting item. It'd look smt like this:
Screenshot 2023-05-11 at 01 40 09


WDYT?


> (old versoin showed the info box at the end which is unexpected).



I agree, wasnt easily noticable esp at first glance

frontends/web/src/routes/new-settings/components/advanced-settings/enable-tor-proxy-setting.module.css
shonsirsha: I was thinking about this. I think it makes sense to create a new prop to achieve a result similar to that.  E.g lets call it `displayedValue`. The reason is bc I wanna keep the `ExtraComponent` prop, since in some setting items there'll be text+**icon** (not just text+Chevron right).

I'll make a change for now n  feel free to lmk what you think:)

frontends/web/src/routes/new-settings/components/advanced-settings/tor-proxy-dialog.tsx
shonsirsha: > setProxyConfig is async.. shouldn't it do await here or doesn't matter?


Ups! Doesnt matter here but better to keep it consistent and have await here, good catch:)

> I am not sure, old implementation doesn't have proxyEnabled state, the toggle is directly using the config.

If im not mistaken, the toggle for proxy is still enabled  using the state (the old settings save the whole config in a state)

Screenshot 2023-05-11 at 02 04 59

shonsirsha: I'll follow the same pattern and remove `proxyEnabled` state now for the new proxy. It's just a redundancy that we don't really need since we're using config state anyways. This config state in the parent page is updated in the dialog component thru this prop `onChangeConfig ` thus we need to keep it :)

frontends/web/src/routes/new-settings/components/advanced-settings/tor-proxy-dialog.tsx
shonsirsha: > also I am wondering if it is really needed to receive updates via onChangeConfig and pass config down again... but no blocker. keep as is or feel free to change.

Keeping would be good IMO, reason: https://github.com/digitalbitbox/bitbox-wallet-app/pull/2104#discussion_r1190496329 ,


 another additional reason why we need to update this config state (that lives within the tor dialog), is bc we're also using this `config` state in the `EnableTorProxySetting` (which displays "Disabled" and "Enabled - therefore requires an updated version of config).

shonsirsha: > not important rightnow but what do you think of, instead of passing the whole config to the components, only pass the relevant part?


Yeah we can definitely find out which parts of config we're supposed to be passing from the component that's using it and just pass exactly those. Would be a bit tricky / more guesswork right now as config is not typed for now.. but yh we can do this:)

shonsirsha:
> Sure, fine with me.

Looks pretty good :) @jadzeidan 

Screenshot 2023-05-11 at 03 01 38

frontends/web/src/routes/new-settings/advanced-settings.tsx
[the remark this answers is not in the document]
thisconnect: thank you 100% better than before, I'm not sure why it ended up being at the end of the view in the first place.

frontends/web/src/routes/new-settings/components/advanced-settings/enable-tor-proxy-setting.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <TorProxyDialog
+        open={showTorProxyDialog}
+        config={config}
+        onCloseDialog={() => setShowTorProxyDialog(false)}
+        onChangeConfig={onChangeConfig}
+        handleShowRestartMessage={setShowRestartMessage}
+      />
+    </>
+  );
+};
```
thisconnect, line 55: Please add license headers to all .tsx files

thisconnect (approved):
LGTM, the tor config has a bit a weird behavior, but it was already like this in old settings. better to can improve it in a later PR.

please add license headers

PR #2115, opened 2023-05-15
frontend: refactored update component without load hoc

thisconnect:
To test this change `semver.NewSemVer` in `backend/update.go` to a lower semver number and restart servewallet.

### DIFF
--- a/frontends/web/src/components/update/update.tsx
+++ b/frontends/web/src/components/update/update.tsx
@@ -16,19 +16,19 @@
  */
 
 import { useTranslation } from 'react-i18next';
-import { load } from '../../decorators/load';
 import { runningInAndroid } from '../../utils/env';
-import { TUpdateFile } from '../../api/version';
+import { getUpdate } from '../../api/version';
 import Status from '../status/status';
 import { AppDownloadLink } from '../appdownloadlink/appdownloadlink';
+import { useLoad } from '../../hooks/api';
 
-type TProps = {
-    file: TUpdateFile | null;
-}
-
-const Update = ({ file }: TProps) => {
+export const Update = () => {
   const { t } = useTranslation();
-  return file && (
+  const file = useLoad(getUpdate);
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/update/update.tsx
```diff
@@ -41,7 +41,3 @@ const Update = ({ file }: TProps) => {
     </Status>
   );
 };
-
-const HOC = load<TProps>({ file: 'update' })(Update);
```
thisconnect, line 45: fyi: this would fetch update and write the result into a prop 'file'.
https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/decorators/load.tsx

there is a few more around which I'll try to refactor soon. load also accepts a function instead, for example here: 
https://github.com/digitalbitbox/bitbox-wallet-app/pull/2116

shonsirsha: > fyi: this would fetch update and write the result into a prop 'file'.

is this file somewhere in appfolder.dev or what is this file?

shonsirsha (approved):
tested LGTM.

🚀

shonsirsha (approved):
tested LGTM.

🚀 

EDIT: Idk how this approval got duplicated lol 😄

frontends/web/src/components/update/update.tsx
thisconnect: It's just a JSON on our servers, see the go backend endpoint code https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/update.go

shonsirsha: Oh, the 'update' endpoint will do that. I thought you were talking about the load HoC :)

thisconnect: ah, no file is just a string in which prop to put the result in.

PR #2116, opened 2023-05-15
frontend: refactor show-firmware-hash component

thisconnect:
- use typed api
- removed unnecessary div

### DIFF
--- a/frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
+++ b/frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2023 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
```diff
@@ -14,48 +15,41 @@
...
-type TProps = TToggleProps & TLoadedProps;
+type Props = {
+  deviceID: string;
+};
 
-const ToggleFWHash = ({ enabled, deviceID }: TProps) => {
+export const ToggleShowFirmwareHash = ({ deviceID }: Props) => {
   const { t } = useTranslation();
-  const [enabledState, setEnabledState] = useState<boolean>(enabled);
+  const [enabledState, setEnabledState] = useState<boolean>(false);
```
thisconnect, line 30: changed initial state to false, as on mount enabledConfig is still undefined until getShowFirmwareHash loads

frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
```diff
@@ -14,48 +15,41 @@
...
+      <Toggle
+        checked={enabledState}
+        id="togggle-show-firmware-hash"
+        onChange={handleToggle}
+        className="text-medium" />
     </div>
   );
 };
-
-const HOC = load<TLoadedProps, TToggleProps>(({ deviceID }) => ({ enabled: 'devices/bitbox02-bootloader/' + deviceID + '/show-firmware-hash-enabled' }))(ToggleFWHash);
```
thisconnect, line 60: example of load decorator hoc with function

shonsirsha (approved):
code lgtm 👍  i havent tested, but does this somehow show the toggle toggling "false" for a split second when `enabledConfig` is still `undefined`?

If so, not a big deal, but would be nice to only render the toggle when it's ready? Otherwise we can show a skeleton. This is the approach that we have on settings.

PR #2119, opened 2023-05-19
frontend: hide hamburger on large screen

thisconnect:
Hamber should not be shown on large screen as the sidebar is always visible, for example on waiting view.

CSS regression introduced in b169a1439


shonsirsha:
Ah, sorry! I was pulling from `origin` instead of `upstream` before and this update wasn't in my master 🤦‍♂️ 

LGTM.

PR #2124, opened 2023-05-23
frontend: created tabs for manage accounts page

shonsirsha:
The new settings include tabs (and respective mobile styling) for different settings page. Manage accounts is not one of them, however it makes sense to also add the tabs and its respective mobile styling here since it's accessible through the (new) settings.

Preview:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/0a32b0f6-8535-435d-a15c-4f7824876243


frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -204,53 +207,53 @@ class ManageAccounts extends Component<Props, State> {
...
-            <div className="content">
-              <div className="columnsContainer">
-                <div className="buttons m-bottom-large m-top-large">
+      <Main>
+        <div className="hide-on-small"><Header title={<h2>{t('manageAccounts.title')}</h2>} /></div>
+        <View fullscreen={false}>
+          <ViewContent>
+            <WithSettingsTabs deviceIDs={deviceIDs} hideMobileMenu hasAccounts={hasAccounts} subPageTitle={t('manageAccounts.title')}>
+              <>
+                <div className="columnsContainer">
```
thisconnect, line 220: thanks for changing to View component 👍 

since you are here... I quickly tested removing the `columnsContainer` div, it doesn't do much. It was supposed to be used in combination with `columns` + `column` iirc. It seems to add some margin-right on small screen, due to oldschool grid styles.

thisconnect (approved):
LGTM with 1 suggestion

frontends/web/src/routes/settings/manage-accounts.tsx
shonsirsha: True. I'll remove this div. Thanks!

PR #2129, opened 2023-05-29
frontend: style multi select dropdown

shonsirsha:
to improve our UI, we'd like to style our Active Currencies multi dropdown.

In this PR we also refactored it, along with the single dropdown component.

Most up-to-date preview:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/117d7325-d230-4bad-a3bc-57e4d6ce1cd7

Screenshot_2023-05-30_at_14 31 54

Screenshot_2023-05-30_at_14 31 03


shonsirsha:
There's unfortunately a known issue in react-select that I discovered while developing. When opening the multi select/dropdown, it automatically focuses on the first option. I believe this shouldn't be a default behaviour, but unfort. it is, and I still have no way to fix it that's not too hacky.

SO question: https://stackoverflow.com/questions/65510875/disable-focus-on-first-option-in-react-select (the solution is outdated as it's intended for v4, we're using v5).

Github PR that's been open for years at their end and also some discussion: https://github.com/JedWatson/react-select/pull/4080

thisconnect:
> There's unfortunately a known issue in react-select that I discovered while developing. When opening the multi select/dropdown, it automatically focuses on the first option. I believe this shouldn't be a default behaviour, but unfort. it is, and I still have no way to fix it that's not too hacky.

IMO having focus on the first item is fine. but the focus should be visible.

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+         ${search.length > 0 ? activeCurrenciesDropdownStyle.hideMultiSelect : ''}
+         `}
+      classNamePrefix="react-select"
+      isSearchable
+      isClearable={false}
+      components={{ DropdownIndicator, IndicatorSeparator: () => null, MultiValueRemove: () => null, Option }}
+      isMulti
+      closeMenuOnSelect={false}
+      hideSelectedOptions={false}
+      value={[...selectedCurrencies].reverse()}
```
shonsirsha, line 63: reverse() here so the latest selected currency will be displayed at the most front of the string.

E.g: 
1. Let's say selected currencies looks like this: **GBP, EUR**
2. User selects **AUD**
3. Now, selected looks like this: **AUD, GBP, EUR**

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.module.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/*
+displays '...' as a pseudo component of the 3rd currency and only when there's more than 3 selected.
+:nth-last-child(2) is used for the same reason as above
+*/
+.select:not(.hideMultiSelect) :global(.react-select__multi-value):nth-child(3):not(:nth-last-child(2))::after {
+    content: ' ...';
+}
+
```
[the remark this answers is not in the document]
shonsirsha: We can omit this pseudo element thing if we want to re-design the UI of that part a little bit. Such as something like this 😄 

Screenshot 2023-05-30 at 12 09 43

we ofc can remove the "X", and just have it as "badges", the way they're intended to be rendered - at least according to the lib.

shonsirsha:
changes made, demo updated in PR's desc.
PTAL 🙏 

cc @jadzeidan

frontends/web/src/components/icon/icon.tsx
```diff
@@ -52,6 +52,7 @@ import saveSVG from './assets/icons/save.svg';
 import saveLightSVG from './assets/icons/save-light.svg';
 import starSVG from './assets/icons/star.svg';
 import starInactiveSVG from './assets/icons/star-inactive.svg';
+import selectedCheckLight from './assets/icons/selected-check-light.svg';
```
thisconnect, line 55: nit: for consistency add SVG at the end as all other imported svgs end with SVG in the imoported name.

frontends/web/src/components/icon/icon.tsx
```diff
@@ -156,6 +157,8 @@ export const Save = (props: ImgProps) => (<img src={saveSVG} draggable={false} {
 export const SaveLight = (props: ImgProps) => (<img src={saveLightSVG} draggable={false} {...props} />);
 export const Star = (props: ImgProps) => (<img src={starSVG} draggable={false} {...props} />);
 export const StarInactive = (props: ImgProps) => (<img src={starInactiveSVG} draggable={false} {...props} />);
+// check component for cusotm multi-dropdown select
```
thisconnect, line 160: comment not needed IMO, this icon could also be used in a different component in the future.

frontends/web/src/components/rates/rates.tsx
```diff
@@ -34,6 +34,25 @@ export interface SharedProps {
...
+  { currency: 'HKD', displayName: 'Hong Kong dollar' },
+  { currency: 'ILS', displayName: 'Israeli new shekel' },
+  { currency: 'JPY', displayName: 'Japanese yen' },
+  { currency: 'KRW', displayName: 'South Korean won' },
+  { currency: 'NOK', displayName: 'Norwegian krone' },
+  { currency: 'RUB', displayName: 'Russian ruble' },
+  { currency: 'SEK', displayName: 'Swedish krona' },
+  { currency: 'SGD', displayName: 'Singapore dollar' },
+  { currency: 'USD', displayName: 'United States dollar' },
+  { currency: 'BTC', displayName: 'Bitcoin' }
```
thisconnect, line 54: I thought about if we should display the currency name in the native language. i.e. "Svensk krona" instead of "Swedish krona", but that does't really work for "Swiss france" with our 4 official languages 🤦 

so I guess better to have everything in English.

frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
```diff
@@ -14,77 +14,20 @@
...
-        }
-      }}
-      options={options}
-    />);
-};
+import { ActiveCurrenciesDropdown } from '../dropdowns/activecurrenciesdropdown';
 
 const ActiveCurrenciesDropdownSetting = ({ selected, active }: SharedProps) => {
-  const formattedCurrencies = currencies.map((currency) => ({ label: currency, value: currency }));
+  const formattedCurrencies = currenciesWithDisplayName.map((fiat) => ({ label: `${fiat.displayName} (${fiat.currency})`, value: fiat.currency }));
```
thisconnect, line 23: nit: this calls array map over and over but the formattedCurrencies are always the same. Maybe rates component could map once and export?

same with formattedCurrencies in defaultCurrencyDropdownSetting component (as it now always contains all currencies)

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.module.css
```diff
@@ -0,0 +1,58 @@
...
+except the last displayed component (the 3rd currency component).
+
+The reason for :nth-last-child(2) instead of just :last-child is because there's an extra
+component created by react-select after the actual last currency component. 
+
+So, we're targetting the second to last component here, 
+which is the last selected currency in the DOM.
+*/
+.select:not(.hideMultiSelect) :global(.react-select__multi-value):not(:nth-last-child(2))::after {
+    content: ', ';
```
thisconnect, line 31: looks like space doens't work it seems to collapse and be ignored in pseudo element via content.

It would look nicer with a bit of space after the comma, can you add `padding-right: 2px;`?

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.module.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/*
+displays '...' as a pseudo component of the 3rd currency and only when there's more than 3 selected.
+:nth-last-child(2) is used for the same reason as above
+*/
+.select:not(.hideMultiSelect) :global(.react-select__multi-value):nth-child(3):not(:nth-last-child(2))::after {
+    content: ' ...';
```
thisconnect, line 39: This CSS selector is a bit crazy :) but I guess adding `…` via JSX would make the component a lot more complicated. then fine I guess…

remove space as it doesn't support leading space in content property and I think space is not needed from layout perspective.

and could you use `&hellip;` tripple dot `…` 🤓  ?

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.module.css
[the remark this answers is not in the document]
thisconnect: the "badges" with close button look cool, but I guess not necessary I guess better if we keep it minimal text-only for now.

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.tsx
thisconnect: ahh that makes sense, I was wondering what the order is and why it changes before. This way it always shows the newest nice 👍

frontends/web/src/routes/new-settings/components/dropdowns/singledropdown.tsx
```diff
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 
-import { FC } from 'react';
-import Select, { components, DropdownIndicatorProps, SingleValue as SingleValueType, SingleValueProps, OptionProps } from 'react-select';
-import styles from './singledropdown.module.css';
+import Select, { components, DropdownIndicatorProps, SingleValueProps, OptionProps } from 'react-select';
+import dropdownStyles from '../dropdowns/dropdowns.module.css';
```
thisconnect, line 18: from './dropdowns.module.css';

thisconnect (changes requested):

frontends/web/src/components/rates/rates.tsx
thisconnect: Please note we have new Fiat currencies on master, maybe merge master->staging-revamp-settings first?

https://github.com/digitalbitbox/bitbox-wallet-app/pull/2141

frontends/web/src/routes/new-settings/components/dropdowns/activecurrenciesdropdown.module.css
shonsirsha: Ooh lol good catch, I didnt even notice.. 🤔

frontends/web/src/components/icon/icon.tsx
shonsirsha: oops..

frontends/web/src/routes/new-settings/components/appearance/activeCurrenciesDropdownSetting.tsx
shonsirsha: Sure I'll do this, but these exported values are probably only gonna be used in the respective dropdown(s), as it's only used for exactly the dropdown/select component (thus it needs a `label` and `value`).

shonsirsha:
@thisconnect Thanks for the review! Changes made, PTAL 🙏 

Also... yeah why not.. we can merge master to the staging branch first 🤷‍♂️ 
Here's the PR master->staging-revamp-settings: https://github.com/digitalbitbox/bitbox-wallet-app/pull/2142

thisconnect (approved):
thanks LGTM

PR #2131, opened 2023-05-31
frontend/accountsummary.tsx: refactor into functional component and add sd-card check

Beerosagos:
* Refactor AccountsSummary into functional component
* move the rendering part into a dedicated component to increase readability and modularity
* add sd-card check that was previously missing

I had to change a bit the functions declaration order in the file because of hook dependencies, sorry for that making the review a bit more complicated :pray:

### DIFF
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -15,124 +15,86 @@
  * limitations under the License.
  */
 
-import { Component, PropsWithChildren } from 'react';
-import { withTranslation, WithTranslation } from 'react-i18next';
-import { unsubscribe, UnsubscribeList } from '../../../utils/subscriptions';
+import { useState, useEffect, useRef, useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
 import * as accountApi from '../../../api/account';
-import { syncAddressesCount } from '../../../api/accountsync';
+import { unsubscribe, UnsubscribeList } from '../../../utils/subscriptions';
 import { apiWebsocket, TPayload } from '../../../utils/websocket';
+import { syncAddressesCount } from '../../../api/accountsync';
 import A from '../../../components/anchor/anchor';
 import { Header } from '../../../components/layout';
-import { Entry } from '../../../components/guide/entry';
-import { Guide } from '../../../components/guide/guide';
-import { FiatConversion } from '../../../components/rates/rates';
 import { Check } from '../../../components/icon/icon';
-import Logo from '../../../components/icon/logo';
-import Spinner from '../../../components/spinner/ascii';
-import { debug } from '../../../utils/env';
 import { Chart } from './chart';
 import { AddBuyReceiveOnEmptyBalances } from '../info/buyReceiveCTA';
-import { apiPost } from '../../../utils/request';
-import style from './accountssummary.module.css';
-import { route } from '../../../utils/route';
 import { Amount } from '../../../components/amount/amount';
 import { Skeleton } from '../../../components/skeleton/skeleton';
+import { Entry } from '../../../components/guide/entry';
+import { Guide } from '../../../components/guide/guide';
+import { debug } from '../../../utils/env';
+import { apiPost } from '../../../utils/request';
+import Logo from '../../../components/icon/logo';
+import Spinner from '../../../components/spinner/ascii';
+import { FiatConversion } from '../../../components/rates/rates';
+import { route } from '../../../utils/route';
+import style from './accountssummary.module.css';
 
-interface AccountSummaryProps {
+type TProps = {
     accounts: accountApi.IAccount[];
-}
+};
 
-export interface Balances {
+export type Balances = {
     [code: string]: accountApi.IBalance;
-}
+};
 
-interface SyncStatus {
+type SyncStatus = {
     [code: string]: number;
-}
+};
 
-interface State {
-    data?: accountApi.ISummary;
-    exported: string;
-    balances?: Balances;
-    syncStatus?: SyncStatus;
-    totalBalancePerCoin?: accountApi.ITotalBalance;
-}
-
-type Props = WithTranslation & AccountSummaryProps;
-
-interface BalanceRowProps {
+type BalanceRowProps = {
     code: accountApi.AccountCode;
     name: string;
     balance?: accountApi.IAmount;
     coinUnit: string;
     coinCode: accountApi.CoinCode;
     coinName: string;
-}
+};
 
 type TAccountCoinMap = {
     [code in accountApi.CoinCode]: accountApi.IAccount[];
 };
 
-class AccountsSummary extends Component<Props, State> {
-  private summaryReqTimerID?: number;
-  public readonly state: State = {
-    data: undefined,
-    exported: '',
-    totalBalancePerCoin: undefined,
-  };
-  private unsubscribeList: UnsubscribeList = [];
+export function AccountsSummary({
+  accounts,
+}: TProps) {
+  const { t } = useTranslation();
+  const summaryReqTimerID = useRef<number>();
+  const unsubscribeList = useRef<UnsubscribeList>([]);
+  const firstRender = useRef(true);
 
-  public componentDidMount() {
-    this.getAccountSummary();
-    this.subscribe(true);
-    this.getAccountsTotalBalance();
-  }
+  const [data, setData] = useState<accountApi.ISummary>();
+  const [totalBalancePerCoin, setTotalBalancePerCoin] = useState<accountApi.ITotalBalance>();
+  const [balances, setBalances] = useState<Balances>();
+  const [syncStatus, setSyncStatus] = useState<SyncStatus>();
+  const [exported, setExported] = useState('');
 
-  public componentWillUnmount() {
-    window.clearTimeout(this.summaryReqTimerID);
-    unsubscribe(this.unsubscribeList);
-  }
-
-  public componentDidUpdate(prevProps: Props) {
-    // accounts can be empty in webdev, this can be removed once this is migrated to FunctionalComponent
-    if (this.props.accounts.length === prevProps.accounts.length) {
-      return;
+  const getAccountSummary = () => {
+    // replace previous timer if present
+    if (summaryReqTimerID.current) {
+      window.clearTimeout(summaryReqTimerID.current);
     }
-    this.subscribe();
-  }
-
-  private subscribe = (initial?: boolean) => {
-    unsubscribe(this.unsubscribeList);
-    this.unsubscribeList.push(apiWebsocket(this.onEvent));
-    this.props.accounts.forEach(account => {
-      this.unsubscribeList.push(syncAddressesCount(account.code, this.onSyncAddressesCount));
-      this.onStatusChanged(account.code, initial);
-    });
+    accountApi.getSummary().then(setData).catch(console.error);
   };
 
-  private getAccountSummary = () => {
-    accountApi.getSummary().then(data => {
-      this.setState({ data }, () => {
-        if (this.summaryReqTimerID) {
-          return;
-        }
-        const delay = (!data || data.chartDataMissing) ? 1000 : 10000;
-        this.summaryReqTimerID = window.setTimeout(this.getAccountSummary, delay);
-      });
-    }).catch(console.error);
-  };
-
-  private async getAccountsTotalBalance() {
+  const getAccountsTotalBalance = () => {
     try {
-      const totalBalancePerCoin = await accountApi.getAccountsTotalBalance();
-      this.setState({ totalBalancePerCoin });
+      accountApi.getAccountsTotalBalance().then(setTotalBalancePerCoin);
     } catch (err) {
       console.error(err);
     }
-  }
+  };
 
-  private getAccountsPerCoin = () => {
-    return this.props.accounts.reduce((accountPerCoin, account) => {
+  const getAccountsPerCoin = () => {
+    return accounts.reduce((accountPerCoin, account) => {
       accountPerCoin[account.coinCode]
         ? accountPerCoin[account.coinCode].push(account)
         : accountPerCoin[account.coinCode] = [account];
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -15,199 +15,193 @@
...
-      }
-    }));
+  });
 
-    if (initial) {
-      return;
+  // update the timer to get a new account summary update when receiving the previous call result.
+  useEffect(() => {
+    // replace previous timer if present
+    if (summaryReqTimerID.current) {
```
Beerosagos, line 157: Needed to change a bit the timer handling, since the previous implementation didn't seem to be directly usable in a functional component. React needs the `useRef` hook to be able to update the timer in the `useEffect` (which was previously done in the callback related to the data status update).

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -0,0 +1,217 @@
...
+        ? accountPerCoin[account.coinCode].push(account)
+        : accountPerCoin[account.coinCode] = [account];
+      return accountPerCoin;
+    }, {} as TAccountCoinMap);
+  };
+
+  const balanceRow = (
+    { code, name, coinCode }: TBalanceRow,
+  ) => {
+    const balance = balances ? balances[code] : undefined;
```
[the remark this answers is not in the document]
shonsirsha: If youre wondering what's reason why we shouldn't declare a react component inside another, it's because doing so (`subTotalRow` inside `SummaryBalance`, for example) is an anti-pattern.

Read: https://stackoverflow.com/a/59776934/9786502
TL;DR: This causes whatever is inside `SummaryBalance` to get re-created every re-render of `SummaryBalance`, which becomes inefficient.

Might be interesting read: https://levelup.gitconnected.com/code-review-avoid-declaring-react-component-inside-parent-component-1768a645f523

shonsirsha (changes requested):

frontends/web/src/routes/account/summary/summarybalance.tsx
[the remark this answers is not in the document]
Beerosagos: Good to know! Thanks :)
Moved `balanceRow` and `subTotalRow` into dedicated components. Also simplified a bit `renderAccountSummary` and embedded it directly into the return JSX code.
PTAL :pray:

shonsirsha (approved):
Thanks! LGTM :) 🎉

Beerosagos:
@thisconnect do you want to have a look too?

Beerosagos:
> Just a nit but actually... would be nice if we could use async await anytime it's possible / not too hard to do smile

Ok, no problem :+1:  Why is it better tho?

shonsirsha:
>  Why is it better tho?

Better in terms of performance? Unlikely. It's just a coding style / the developers' preference. I personally feel that `async await` is more convenient than `.then()`.

As we're trying to move towards modern and best practices of React and JS (thus we're actively converting class into functional components, for example), I do believe it makes sense to also favour `async await` over `then` when it's not too big of an effort. My suggestion to use `async await` is just so we have consistency in our codebase in terms of dealing with promises (since we're already using it).  But again, as it's just a nit feel free to disregard. I give it as a nit here as we're only handling a simple promise.

TL;DR: Why `async await`? Easier to read, write, test and maintain, which was the reason why it was invented. I understand that sometimes it's a little bit tricky to write async await for example inside of `useEffect`s - so that's why I suggested to only do it when it's not "too hard".

Interesting reads: 
https://stackoverflow.com/questions/54495711/async-await-vs-then-which-is-the-best-for-performance
https://stackoverflow.com/questions/34401389/what-is-the-difference-between-javascript-promises-and-async-await

---

Aside from this PR, I guess it might still be interesting to see the differences. Let's use some examples. Let's say we're getting a user profile (API fetch) from many user IDs (another API fetch).

<details>
 <summary> Examples 🙃 🔫</summary>


We will first fetch all of the userIDs then get the single userProfile.

This is how we would do it using `.then()`

```js
const getUserProfileByID = (id) => {
 getUserAllUserIDs()
    .then(() => getUserProfile(id))
    .then((userProfile) => {
       console.log(userProfile)
       return userProfile
    })
}
```

or, some people might even do a nested `.then()`

```js
getUserProfileByID(id)
.then(userIDs => 
    getUserProfile(id)
    .then(userProfile => 
      etc...))
```

let's compare it to `async await`:

```js
const getUserProfileByID = async (id) => {
   const userIDs = await getAllUserIDs();
   const userProfile = await getUserProfile(userIDs[id]);
   console.log(userProfile);
   return userProfile;
}
```
or we can even turn it into a one-liner but that's just unnecessary 😄 

So, for me personally it's a lot easier to grasp the example with `async await` for at least 2 things:

1. They're not getting "deeper" down the line (promise chain is at least 1 level deeper to be able to access the promise / async value, and if you nest it then it's however deep you want or can get). If we somehow want to get this value outside of this closure, it's going to be a little a little bit tricky. And even if we don't wanna get it out, it's still more "mental" work to go thru all of the .thens().

2. We know that whatever is AFTER the `await` will **really** be executed after `await`. When using `.then()`, this isn't the case. This is also a big one for me. It's mentally easier for me to just see the code from top to bottom and be sure that the code in this function runs sequentially.

```js
getUser().then(result=> {
  console.log(result);
});
console.log("WOW");
```

If `getUser` is an API  fetch, "WOW" will very likely to be fired first than `result`. But due to the way it's written (using `.then()` 😌), it may not look like it.

Whereas if we do the same thing with `async await`, it's clear that "WOW" will only get fired **after** the fetch call.

```js
const result = await getUser();
console.log(result)
console.log("WOW");
```
</details>


We also have to remember that those examples don't look bad at all when using .then (both chaining n even nested) - but I wanna say that's because we only have like 2 promises. Idk which one you prefer but maybe this will give you an idea as to why I prefer `async await` (at least for now...) 😄

thisconnect:
also iirc in some situations async/await error useful stacktrace whereas promise can point to weird location that doesnt' make sense.

Beerosagos:
PR updated with `async await` calls :pray: @thisconnect let me know if you want to take a look or if I should merge :)

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -183,231 +106,144 @@ class AccountsSummary extends Component<Props, State> {
...
+                (accounts.length === Object.keys(balances || {}).length) ? (
+                  <AddBuyReceiveOnEmptyBalances balances={balances} />
+                ) : undefined
+              } />
+            <SummaryBalance
+              accounts={accounts}
+              summaryData={summaryData}
+              totalBalancePerCoin={totalBalancePerCoin}
+              balances={balances}
+              syncStatus={syncStatus}
```
thisconnect, line 230: fine for now, but it looks like all syncStatus related logic could be moved to SummaryBalance in the future.

thisconnect (approved):
nice, tested a bit with deleting headers OR accounts cache first. 

added some comments, feel free to keep as is or change as you like.

Beerosagos:
After a bit of discussion with @thisconnect I decided to try improving a bit the event handling: I pushed a new commit with some new changes:
- made a new hook that allows subscribing to a single endpoint with multiple accounts, to make easier managing account events like address count sync and status changes. 
- moved address count sync to summarybalance.tsx
This could probably be further improved (we call the endpoints repeatedly during the loading, but something could probably be avoided), but didn't want to make this pr to huge (I think the hook is already a bit borderline with the scope of the PR).

Also, despite the use of the new hook, there are several points in the code where we set a new state after an `await`. These calls could raise errors if the component gets detached during the await. Handling these cases with the UseMountedRef hook is not always easy. I wonder if it could make sense to improve this in general in the app creating another hook, dedicated to updating states in a consistent way after awaiting for a call (this is anyway smth for another PR ofc).

cc @shonsirsha

Beerosagos:
Also lmk if you think it is better to keep this PR simple and move the last commit in a new dedicated PR

frontends/web/src/api/accountsync.ts
```diff
@@ -30,6 +30,18 @@ export const syncAccountsList = (
...
+ * event to receive the progress of the address sync.
+ * Meant to be used with `useSubscribeMap`.
+ */
+export const subscribeAddressesCount = (code: AccountCode) => (
+  (cb: TSubscriptionCallback<number>) => {
+		 return subscribeEndpoint(`account/${code}/synced-addresses-count`, (
+      count: number,
+    ) => cb(count));
+  });
+
```
thisconnect, line 44: We also have `syncAddressesCount` that works slightly different (not curry function), it is a bit confusing now with both versions with `subscribeAddressesCount`. There is also other places in the app which we have curry function in combination with useSync.  I'll need some time to think about it,

thisconnect:
yes separate PR  is probably better. I'll need some time to review and test this. 

> 
Also lmk if you think it is better to keep this PR simple and move the last commit in a new dedicated PR

Beerosagos:
> would you mind making a separate commit with "- moves account sync status logic to summarybalance.tsx." before the hook refactor?
Done! Moved the hook refactor to a new draft PR #2167.

If everything looks good here, I'd merge @thisconnect

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -53,6 +58,23 @@ export function SummaryBalance ({
...
+      [code]: syncedAddressesCount,
+    }));
+  }, []);
+
+  useEffect(() => {
+    unsubscribe(unsubscribeList.current);
+    accounts.forEach(account => {
+      unsubscribeList.current.push(syncAddressesCount(account.code, onSyncAddressesCount));
+    });
+  }, [accounts, onSyncAddressesCount]);
```
thisconnect, line 76: If possible, I think it would make sense to have the subscription in BalanceRow component, maybe then you could use `useSubscribe` and remove the forEach?

also the type SyncStatus should probably not be in accountsummary.tsx

thisconnect (changes requested):

frontends/web/src/routes/account/summary/summarybalance.tsx
Beerosagos: Very good point! Thanks for the catch :) Actually, moving the logic to `BalanceRow` we don't need `SyncStatus` at all. I moved it, but I didn't use `useSubscribe` since we'd also need to modify `syncAddressesCount` which is also used in `account.tsx`. I think it is better to change that in the new PR I opened. wdyt?

frontends/web/src/routes/account/summary/balancerow.tsx
```diff
@@ -0,0 +1,99 @@
...
+  { code, name, coinCode, balance }: TProps
+) {
+  const { t } = useTranslation();
+
+  const [syncStatus, setSyncStatus] = useState<number>();
+
+  const unsubscribe = useRef<TUnsubscribe>();
+
+  const onSyncAddressesCount = (
+    _: string,
```
thisconnect, line 46: this is the code string that is returned, reason is the follwoing line
(where code can change when changing accounts)

https://github.com/digitalbitbox/bitbox-wallet-app/blob/fe9bd5c92da6d8fa1bb7de539839e8db454a2ece/frontends/web/src/routes/account/account.tsx#L122

I think fine for this PR, we can improve that in the future, maybe use curry function.

thisconnect (changes requested):

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -15,399 +15,221 @@
...
-export interface Balances {
+export type Balances = {
     [code: string]: accountApi.IBalance;
-}
+};
 
-interface SyncStatus {
+export type SyncStatus = {
     [code: string]: number;
-}
```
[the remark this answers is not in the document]
Beerosagos: How did I possibly miss this? :facepalm:

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -183,231 +106,144 @@ class AccountsSummary extends Component<Props, State> {
...
-        </td>
-        <td data-label={t('accountSummary.fiatBalance')}>
-          <strong>
-            <FiatConversion amount={balance} noAction={true} />
-          </strong>
-        </td>
-      </tr>
-    );
-  }
+  });
```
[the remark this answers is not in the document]
Beerosagos: Sorry, I did that but got lost in translation when splitting the PR! Will restore it.

frontends/web/src/routes/account/summary/balancerow.tsx
Beerosagos: yup, would be nice to remove that, but I'd do it in a future PR

frontends/web/src/routes/account/summary/balancerow.tsx
```diff
@@ -0,0 +1,99 @@
...
+    syncedAddressesCount: number,
+  ) => {
+    setSyncStatus(syncedAddressesCount);
+  };
+
+  useEffect(() => {
+    if (unsubscribe.current) {
+      unsubscribe.current();
+    }
+    unsubscribe.current = syncAddressesCount(code, onSyncAddressesCount);
```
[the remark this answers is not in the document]
Beerosagos: Good catch, thanks! But it needs to return a function, or the unsubscribe will be executed right after the subscribe. Should be fixed now :+1:

Beerosagos:
Thanks for the review @thisconnect, just pushed a new version PTAL :pray:

frontends/web/src/routes/account/summary/balancerow.tsx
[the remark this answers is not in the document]
thisconnect: ah sorry, yes correct it should return a function  thanks :)

frontends/web/src/routes/account/summary/accountssummary.tsx
thisconnect: I am thinking if the clearTimeout should be instead in the return function of this useEffect, so that it clears on unmount, but also when summaryData changes instead of here.

and then you could remove this;

https://github.com/digitalbitbox/bitbox-wallet-app/pull/2131/files?diff=split&w=0#diff-e5d06f77122fc6a06ea5e7c80d3466cd0fa4d69a252e8108de78062a5bdbea83R159

thisconnect (approved):
LGTM with one more suggestion/question

frontends/web/src/routes/account/summary/accountssummary.tsx
Beerosagos: Very nice! Done :+1:

PR #2132, opened 2023-05-31
backend: remove ext.shiftcrypto.ch and guids.shiftcrypot.ch

benma:
Not used anywhere in the app or in the banner notifications.


thisconnect (approved):
LGTM with nit

PR #2133, opened 2023-05-31
scripts/dockerdev: handle spaces in repo path

Beerosagos:
Fixes a small nit that caused `make dockerdev` to fail if there is a space in the repo path.


benma (approved):
utACK

PR #2134, opened 2023-05-31
backend/handlers: skip 0txs to mitigate address poisoning

Beerosagos:
Address Poisoning attack consists of an attacker creating empty (0 amount) txs to/from vanity addresses, hoping that the possible victim will copy/paste the wrong address from the transaction history when sending/receiving coins. This can happen in some ERC20 tokens, depending on the implementation of the `transfer` function inside the token contract (see
https://mirror.xyz/x-explore.eth/cL3d_CyNujXq8XY7ueP4omNXx_IY1EG5Dz0FD0vJ90M for details).

This hide 0 amount erc20 txs from the `transactions` endpoint, to mitigate the effects of this attack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add support for Czech Crown (CZK) and add Czech translation
 - Add support for Zloty (PLN)
 - Add advanced setup with skip microSD card or 12-word seed options
+- Hide 0 amount ERC20 transactions to partially mitigate Address Poisoning Attack
 
 ## 4.37.0
 - Bundle BitBox02 firmware version v9.14.0
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -18,6 +18,7 @@ package handlers
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -230,6 +231,10 @@ func (handlers *Handlers) getAccountTransactions(_ *http.Request) (interface{}, 
 	}
 	result.Transactions = []Transaction{}
 	for _, txInfo := range txs {
+		if txInfo.IsErc20 && big.NewInt(0).Cmp(txInfo.Amount.BigInt()) == 0 {
+			// skipping 0 amount erc20 txs to mitigate Address Poisoning attack
+			continue
+		}
 		result.Transactions = append(result.Transactions, handlers.getTxInfoJSON(txInfo, false))
 	}
 	result.Success = true
--- a/backend/coins/btc/transactions/transactions.go
+++ b/backend/coins/btc/transactions/transactions.go
@@ -492,6 +492,7 @@ func (transactions *Transactions) txInfo(
 		Size:             int64(txInfo.Tx.SerializeSize()),
 		Weight:           btcdBlockchain.GetTransactionWeight(btcutilTx),
 		CreatedTimestamp: txInfo.CreatedTimestamp,
+		IsErc20:          false,
 	}
 }
 
--- a/backend/accounts/transaction.go
+++ b/backend/accounts/transaction.go
@@ -119,8 +119,9 @@ type TransactionData struct {
 
 	// --- Fields only used for ETH follow
 
-	Gas   uint64
-	Nonce *uint64
+	Gas     uint64
+	Nonce   *uint64
+	IsErc20 bool
 }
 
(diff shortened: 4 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 - Improve sats amount readability adding thousands separator
 - Update minimum supported buntu version to 20.04+
+- Hide 0 amount ERC20 txs to partially mitigate Address Poisoning Attack
```
benma, line 6: txs=>transactions. Imo no point in abbreviating where not necessary, as there are always people that don't know some acronyms or short names.

backend/coins/btc/handlers/handlers.go
```diff
@@ -230,6 +231,10 @@ func (handlers *Handlers) getAccountTransactions(_ *http.Request) (interface{},
 	}
 	result.Transactions = []Transaction{}
 	for _, txInfo := range txs {
+		if txInfo.FeeIsDifferentUnit && big.NewInt(0).Cmp(txInfo.Amount.BigInt()) == 0 {
```
benma, line 234: `FeeIsDifferentUnit` does not imply ERC-20 token necessarily. 

I think it's better to move this logic to `eth/account.go:Transactions()` (or to wherever the transactions are populated), where you can also directly check if it is an ERC-20 token. Generally/ideally, the HTTP handlers should not contain any logic but just parse input and assemble the JSON output.

benma (changes requested):

backend/coins/btc/handlers/handlers.go
Beerosagos: As discussed, I didn't change this, to avoid modifying the persisted transactions, which could have consequences on e.g. balance computing. The code is only for frontend use anyway.

Beerosagos:
@benma pushed an update. 
As agreed I didn't change the code in handlers: If we want to look for an alternative, think the easiest could be to modify `NewOrderedTransactions` (https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/accounts/transaction.go#L157), which is called by `Transactions()`, and making it returning a deep copy of the transaction list wihout 0 amount erc20 txs. lmk if you prefer this last approach.

backend/coins/btc/handlers/handlers.go
benma: The part about `FeeIsDifferentUnit does not imply ERC-20 token necessarily.` still applies though :) Best to make a new field in the tx-info struct `IsERC20` and set/use it accordingly

backend/coins/eth/account.go
```diff
@@ -379,23 +379,44 @@ func (account *Account) update() error {
...
-		}
-		account.balance = coin.NewAmount(balance)
+		balance, err = account.coin.client.Balance(context.TODO(), account.address.Address)
+	}
+
+	if err != nil {
+		return errp.WithStack(err)
 	}
 
+	pendingAmount := pendingTxsAmount(outgoingTransactionsData, account.coin.erc20Token != nil)
```
benma, line 393: seems like an unrelated commit slipped into this PR?

benma (changes requested):

Beerosagos:
@benma Added a new `IsErc20` field. The logic is pretty much the same as `FeeIsDifferentUnit`, I wonder if it would make more sense to just rename it, wdyt?

benma:
> @benma Added a new `IsErc20` field. The logic is pretty much the same as `FeeIsDifferentUnit`, I wonder if it would make more sense to just rename it, wdyt?

No, because they are conceptually not equivalent. `FeeIsDifferentUnit` for example plays a role in the chart balance computation, and could be true for future tokens that are added which are not ERC-20.

backend/coins/btc/transactions/transactions.go
```diff
@@ -492,6 +492,7 @@ func (transactions *Transactions) txInfo(
 		Size:             int64(txInfo.Tx.SerializeSize()),
 		Weight:           btcdBlockchain.GetTransactionWeight(btcutilTx),
 		CreatedTimestamp: txInfo.CreatedTimestamp,
+		IsErc20:          false,
```
benma, line 495: can also leave it out like the other ETH fields - defaults to false anyway.

benma (approved):

benma:
I merged this so it is part of the release candidate testing.

Beerosagos:
> I merged this so it is part of the release candidate testing.

Cool, thanks!

PR #2137, opened 2023-06-05
Frontend: create wallet with advanced backup options

thisconnect:
and added support for skip microSD card and 12 words config.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -29,7 +29,7 @@ import { UpgradeButton } from './upgradebutton';
 import { Unlock } from './unlock';
 import { Pairing } from './setup/pairing';
 import { Wait } from './setup/wait';
-import { SetupOptions } from './setup/choose';
+import { SetupOptions, TWalletCreateOptions, TWalletSetupChoices } from './setup/choose';
 import { CreateWallet } from './setup/wallet-create';
 import { RestoreFromSDCard, RestoreFromMnemonic } from './setup/wallet-restore';
 import { CreateWalletSuccess, RestoreFromMnemonicSuccess, RestoreFromSDCardSuccess } from './setup/success';
@@ -44,7 +44,8 @@ interface State {
     versionInfo?: VersionInfo;
     attestation: boolean | null;
     status: '' | TStatus;
-    appStatus: 'createWallet' | 'restoreBackup' | 'restoreFromMnemonic' | '';
+    appStatus: '' | TWalletSetupChoices;
+    createOptions?: TWalletCreateOptions;
     // if true, we just pair and unlock, so we can hide some steps.
     unlockOnly: boolean;
     showWizard: boolean;
@@ -109,6 +110,13 @@ class BitBox02 extends Component<Props, State> {
     });
   };
 
+  private handleAbort = () => {
+    this.setState({
+      appStatus: '',
+      createOptions: undefined,
+    });
+  };
+
   public componentWillUnmount() {
     unsubscribe(this.unsubscribeList);
   }
@@ -117,6 +125,7 @@ class BitBox02 extends Component<Props, State> {
     const { t, deviceID } = this.props;
     const {
       attestation,
+      createOptions,
       versionInfo,
       status,
       appStatus,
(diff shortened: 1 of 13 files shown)


frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -102,7 +105,8 @@ class BitBox02 extends Component<Props, State> {
         this.setState({ unlockOnly: false });
       }
       if (status === 'seeded') {
-        this.setState({ appStatus: 'createWallet' });
+        console.log('when does this happen: seeded???????', this.state.createOptions);
```
thisconnect, line 108: Found it, this happens if the BBApp is restarted during setup. Currently it just starts the create-wallet process again. But I think if the user has choosen 12 words or skip sdcard this info is lost...

The user probably would only restart if there was some kind of issue, or if they just forgot (went to lunch)...

Maybe better to go back to setup choices?

cc @benma WDYT?

benma: this status event is fired after the bitbox02 password was defined, app restart not necessary. so you can't really go back to the setup choices here. that would rip the user out of the regular setup flow.

thisconnect: ah true 🙈

thisconnect: In this case, I suggest we just remove `this.setState({ appStatus: 'create-wallet' });` here for the following reason:

appStatus is already  'create-wallet'

restarting the app after passsword was set can then go back to setup choices, which is better for the reason I mentioned before:

>  Currently it just starts the create-wallet process again. But I think if the user has choosen 12 words or skip sdcard this info is lost

thisconnect: oups, I noticed my testing is a bit wrong, just reloading the frontend behaves different than restarting servewallet (obviously). Restarting frontend after password goes to set wallet again. Restarting servewallet goes to setup choices, (which is what I proposed anyway).

frontends/web/src/routes/device/bitbox02/setup/wallet-create.tsx
```diff
@@ -43,10 +47,12 @@ export const CreateWallet = ({
   const [status, setStatus] = useState<TCreateWalletStatus>('intro');
   const [errorText, setErrorText] = useState('');
 
+  console.log('CreateWallet', backupType, backupSeedLength);
```
benma, line 50: remove again?

benma (approved):
didn't do a code review but tested all combinations:

- sdcard, 12 words
- sdcard, 24 words
- skip sdcard, 12 words
- skip sdcard, 24 words
- 12 word feature disabled
- skip sdcard feature disabled

Everything works and I really like how you solved putting it into the advanced options! slick


Here are some small things I noticed:

1. Manage Device -> Show recovery words. 

The dialog says 'You will be presented with **24** recovery words, which form a backup of your wallet. Write them down on paper.' Should be reworded to "You will be presented with your recovery words" or similar, as it can also be 12 words.

2. if a feature is disabled it says so in red text. visually it would be better to move that red text to a new line

3. the whole label "Skip microSD card ..." and "Create 12-word ..." labels cannot be selected?! everything should be selectable so it can be copy/pasted etc.

4. please update the changelog :rocket:

frontends/web/src/routes/device/bitbox02/setup/wallet-create.tsx
thisconnect: removed

thisconnect:
thanks for testing 🙏 

changed some more text that mentioned 24 words (in passphrase ino) and will update locize before merging this PR. 

> 1. Manage Device -> Show recovery words.

also addressed feedback 2. 3. and 4.

Beerosagos (approved):
utACK! Just a small nit in a comment below. Very nice :fire:

frontends/web/src/routes/device/bitbox02/setup/choose.tsx
```diff
@@ -14,58 +14,208 @@
...
+
+export type TWalletCreateOptions = {
+  withMnemonic: boolean;
+  with12Words: boolean;
+};
 
 type Props = {
-  onSelectSetup: (option: 'create-wallet' | 'restore-sdcard' | 'restore-mnemonic') => void;
+  onSelectSetup: (
+    type: 'create-wallet' | 'restore-sdcard' | 'restore-mnemonic',
```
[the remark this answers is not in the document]
thisconnect: thanks fixed

thisconnect:
I did some more testing and couldn't reproduce the issue @Beerosagos had once.

PR #2138, opened 2023-06-06
frontend: small settings styling/spacing improvement

thisconnect:
small spacing/aligning improvement, will paste screenshots in comments.

- decrease space between active tab and underline
- decrease space between setting config label and description
- align tabs on the left with content/main-title
- center setting toggle vertically on mobile

cc @jadzeidan

### DIFF
--- a/frontends/web/src/routes/new-settings/components/settingsItem/settingsItem.module.css
+++ b/frontends/web/src/routes/new-settings/components/settingsItem/settingsItem.module.css
@@ -50,7 +50,7 @@
     color: var(--color-secondary);
     font-size: 14px;
     margin: 0;
-    margin-top: 16px;
+    margin-top: var(--space-quarter);
     overflow: hidden;
     text-overflow: ellipsis;
 }
@@ -58,12 +58,8 @@
 @media (max-width: 768px) { 
     .container {
          height: auto;
-         margin-bottom: 2px;
+         margin-bottom: 5px;
          min-width: 100%;
-    }
-
-    .rightContentContainer {
-        margin-top: var(--space-quarter);
     }
 
     .secondaryText {
--- a/frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
+++ b/frontends/web/src/routes/new-settings/components/singledropdown/singledropdown.module.css
@@ -51,6 +51,7 @@
 
 @media (max-width: 560px) {
     .select {
+        margin-top: var(--space-quarter);
         width: 100%;
     }
 }
--- a/frontends/web/src/routes/new-settings/components/tabs.module.css
+++ b/frontends/web/src/routes/new-settings/components/tabs.module.css
@@ -1,6 +1,5 @@
 .container {
     display: flex;
-    padding-left: var(--space-half);
     margin-bottom: var(--space-default);
 }
 
@@ -11,7 +10,7 @@
 }
 
 .container a.active {
-    padding-bottom: var(--space-half);
+    padding-bottom: var(--space-quarter);
     border-bottom: 2px solid var(--color-blue);
 }
 


thisconnect:
Screen Shot 2023-06-06 at 09 14 59

thisconnect:
Screen Shot 2023-06-06 at 09 20 09

thisconnect:
Screen Shot 2023-06-06 at 09 33 03

frontends/web/src/routes/new-settings/components/settingsItem/settingsItem.module.css
```diff
@@ -62,10 +62,6 @@
          min-width: 100%;
     }
 
-    .rightContentContainer {
-        margin-top: var(--space-quarter);
-    }
-
```
thisconnect, line 68: hm removing this fixes the screenshot (on mobile) that I pasted, but then it doesn't have enough spacing in another place with full width dropdowns. Will update...

shonsirsha: Sure. Seems to be okay for now. If this breaks anything it'd be a simple fix anyway.

frontends/web/src/routes/new-settings/components/settingsItem/settingsItem.module.css
```diff
@@ -50,7 +50,7 @@
     color: var(--color-secondary);
     font-size: 14px;
     margin: 0;
-    margin-top: 16px;
+    margin-top: var(--space-quarter);
```
shonsirsha, line 53: nice!

shonsirsha (approved):
tested LGTM 🤷‍♂️ thanks @thisconnect 🎆 !

On mobile the settings item do look a bit shorter (in terms of height). I think it looks good, the height before, I think, was a mistake anyway due to  https://github.com/digitalbitbox/bitbox-wallet-app/pull/2138/files#r1219088526

But, if we want to increase the height correctly on mobile, we can increase its vertical padding. Might be beneficial to increase its surface area so it's easier to tap.

but let @jadzeidan share his thoughts though 😄 im ok with either.

jadzeidan:
@shonsirsha @thisconnect thanks

I tested on mobile. Would it be possible to make the dropdowns in-line with the titles? I think it would look better that way because then all the setting options have the same height:

image

thisconnect:
problem is always that we don't know how long the label text is in different langauges (french and german are often longer), so better to keep it as is.

> I tested on mobile. Would it be possible to make the dropdowns in-line with the titles? I think it would look better that way because then all the setting options have the same height:

shonsirsha:
> I tested on mobile. Would it be possible to make the dropdowns in-line with the titles? I think it would look better that way because then all the setting options have the same height:

I'm not sure about this. It for sure is possible and actually the "collapsing" of the dropdown and title is something that's customised, but im not sure if it'll look and feel better than what we have right now.

The reason we "collapse" it, is so that more info could be displayed + more space for user to tap on the dropdown. I imagine it will also not look to pretty Iif the dropdown is so short (due to width of being only 50%).

shonsirsha:
> Also I think increasing the spacing on mobile a bit makes it look nicer. Here I tried 5px:
> accessibility should be preferred ofc.. related to this I also find these kind of spaces a bit too big on desktop and on tx overview on desktop. how about we just always set it to 5px mobile and desktop and maybe even tx overview (but not in this PR ofc)?

Aesthetically, I like the tiny spacings more than wider ones both for desktop & mobile. But this is subjective haha. I like them more because they seem to be more organised / "one group". But honestly both r okay for me, no strong opinion here :)


> Regarding the height of the options, it looks good to me. I don't think it needs to be bigger.

Sorry, I made a mistake. True. Those options arent tapable/clickable anyway! lol.

This screen is what I meant:

Screenshot 2023-06-06 at 14 16 37

But anyways probably no issues here either 🤷

jadzeidan:
> I don't really like but it's subjective and I guess it improves accessibility because it helps with clearly distinguish the items.
> accessibility should be preferred ofc.. 

@thisconnect I tested with 5px on desktop but I don't think it looks good. The intention of the design is for the options to look like individual cards, instead of a table. imo the spacing should look in proportion to the margins on the side, hence why desktop and mobile should have spacings. On desktop there is more space so we don't need to squeeze the options together as much. Decreasing the space between the options makes it looks more information dense which I think is undesirable.

image

image

> related to this I also find these kind of spaces a bit too big on desktop and on tx overview on desktop. how about we just always set it to 5px mobile and desktop and maybe even tx overview (but not in this PR ofc)?

Currently the mobile and desktop tx overview are different styles (desktop is a table and mobile is "card" style because there are no headers). So imo doesn't make sense to have the same spacing for both. A table has boarders whereas cards have bigger gaps. I prefer to use card style for both desktop and mobile but we should make that change first before changing the spacing.

thisconnect:
I see desktop tx overview  as talbe or list, not as cards of tx, anyway rabit hole. I shouldn't have brought it up here, lets discuss another time.

PR #2140, opened 2023-06-06
frontend: fix removing darkmode from config

thisconnect:
When user changes darkmode settings to the same setting that the OS prefers the config should be just removed, so it uses same as OS.

This broke in:
- e91c84a517cfe81fbdaac228ed58ef1cef8ae687

setConfig now merged with current config and ensures nothing gets lost, but this broke removing the config by omitting.

### DIFF
--- a/frontends/web/src/contexts/DarkmodeProvider.tsx
+++ b/frontends/web/src/contexts/DarkmodeProvider.tsx
@@ -76,7 +76,6 @@ export const DarkModeProvider = ({ children }: TProps) => {
           // remove darkmode from config, so it use the same mode as the OS
           const { darkmode, ...frontend } = config.frontend;
           setConfig({
-            backend: config.backend,
             frontend: {
               ...frontend,
               darkmode: undefined,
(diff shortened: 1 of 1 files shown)


frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -77,7 +77,10 @@ export const DarkModeProvider = ({ children }: TProps) => {
           const { darkmode, ...frontend } = config.frontend;
           setConfig({
             backend: config.backend,
-            frontend,
+            frontend: {
+              ...frontend,
+              darkmode: undefined,
+            },
```
thisconnect, line 83: looking at this, I think `backend: config.backend,` isn't needed anymore as setConfig will take care of it..

shonsirsha: Yes!

shonsirsha (approved):
LGTM. Thanks!!

frontends/web/src/contexts/DarkmodeProvider.tsx
thisconnect: as already discussed DM, frontend and getConfig could be removed. but the detectDarkTheme backend api seems to return false even when my OS is dark (macOS). So I was hesitant to change more as this PR is just a small regression bug fix.

PR #2141, opened 2023-06-06
Add CZK and PLN fiat currencies

thisconnect:


thisconnect:
nice I think it's working now, don't get any errors anymore. merging this.

PR #2145, opened 2023-06-07
frontend: re-add deleted locize keys

benma:
I did a locize pull before a locize push in 8dbaab3ac, accidentally removing keys.


thisconnect (approved):
thanks LGTM

PR #2151, opened 2023-06-12
frontend: using locize keys for texts in the new settings

shonsirsha:
To be able to provide translated texts to the users, we must use locize keys.

In this commit, when the exact word we want to use is already available in locize, we're using them. Otherwise, we created new entries.

Edited & deleted some keys in-place as well as they're not yet "live" on locize.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -508,11 +508,11 @@
     },
     "expert": {
       "factoryReset": {
-        "description": "Reset your device to factory settings. This deletes your wallet!",
+        "description": "Reset your device to factory settings. This deletes the wallet on your BitBox02!",
         "title": "Factory reset"
       },
       "goToStartupSettings": {
-        "description": "Enters the bootloader of the BitBox02. Can enable firmware hash from here."
+        "description": "Enter the bootloader of the BitBox02. You can enable the firmware the hash from here."
       },
       "passphrase": {
         "description": "Enable or disable the passphrase feature.",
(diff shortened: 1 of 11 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -501,18 +501,21 @@
...
     },
     "expert": {
       "factoryReset": {
-        "description": "Reset your device to factory settings. This deletes your wallet!",
+        "description": "Reset your device to factory settings. This deletes the wallet on your BitBox02!",
         "title": "Factory reset"
       },
       "goToStartupSettings": {
-        "description": "Enters the bootloader of the BitBox02. Can enable firmware hash from here."
+        "description": "Enter the bootloader of the BitBox02. You can enable the firmware the hash from here."
```
shonsirsha, line 518: As mentioned in the PR's description, since these keys are not "live" on locize yet it should be ok to edit it here.


... right..? 😄

thisconnect: correct.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -257,4 +257,4 @@ class ManageAccounts extends Component<Props, State> {
 }
 
 const HOC = translate()(ManageAccounts);
-export { HOC as ManageAccounts };
+export { HOC as ManageAccounts };
```
thisconnect, line 260: Nit: always keep newline

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -212,7 +212,7 @@ class ManageAccounts extends Component<Props, State> {
     const accountList = this.renderAccounts();
     return (
       <Main>
-        <div className="hide-on-small"><Header title={<h2>{t('manageAccounts.title')}</h2>} /></div>
+        <div className="hide-on-small"><Header title={<h2>{t('settings.title')}</h2>} /></div>
```
thisconnect, line 215: 👍

frontends/web/src/routes/new-settings/components/device-settings/firmware-setting.tsx
```diff
@@ -42,7 +42,7 @@ const FirmwareSetting = ({ deviceID, versionInfo, asButton = false }: TProps) =>
   const [dialogOpen, setDialogOpen] = useState(false);
   const [confirming, setConfirming] = useState(false);
   const canUpgrade = versionInfo.canUpgrade;
-  const secondaryText = canUpgrade ? t('deviceSettings.firmware.upgradeAvailable') : t('deviceSettings.firmware.upToDate');
+  const secondaryText = canUpgrade ? t('deviceSettings.firmware.upgradeAvailable') : t('deviceSettings.deviceInformation.firmware.upToDate');
```
thisconnect, line 45: not sure, but would it makes sense to make an exception for this one (as it is not a new key), so is stays  together next to `deviceSettings.firmware.upgradeAvailable` so that the translation is together with the other stuff in `deviceSettings.firmware`.

thisconnect (approved):
LGTM with 1 suggestion/question and a nit

frontends/web/src/routes/settings/manage-accounts.tsx
shonsirsha: 😮

frontends/web/src/routes/new-settings/components/device-settings/firmware-setting.tsx
shonsirsha: Ahh sorry... we don't need to have this change here. Just edit on locize from "... Your device ..." to "... Your firmware ...". I'll revert this. 🙏

PR #2153, opened 2023-06-13
frontend: use store.state.active in DefaultCurrencyDropdownSetting (bugfix)

shonsirsha:
When using the `store.state.active` directly from `rates` file, it won't give us the correct value. In our case, this causes invalid value for default currency after refreshing (after loading the component).

We need to use the string that contains `store.active.state` in the component, instead of inside the `rates` file.

[Preview (picking a default value and refreshing)]:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/b2d94802-5530-4051-9d4c-7b693b67edfc


thisconnect (approved):
👍  untested LGTM

PR #2155, opened 2023-06-13
frontend: added guides for relevant settings pages & modified header

shonsirsha:
There were several changes to make the guide works with our tabbed settings.

1. We now use the `Header` component's prop `title` for showing both setting page name for both desktop & mobile.

e.g "Settings" on Desktop and "Appearance" on mobile (when on `/new-settings/appearance`). Previously, to display the page name on mobile we use `WithSettingsTabs`

2. Modified `WithSettingsTab` to not receive the page title anymore due to the point above.

3. Modified `Header` to be able to hide sidebarToggler also due to point 1.

We need to be able to hide `sidebarToggler` because on an individual setting page on mobile (for example `Appearance`), we already have a back button which is inside `MobileHeader`.

So we don't need the `sidebarToggler` anymore.

Preview:

Screenshot 2023-06-13 at 17 36 29
Screenshot 2023-06-13 at 17 36 38
Screenshot 2023-06-13 at 17 36 46

### DIFF
--- a/frontends/web/src/components/layout/header.tsx
+++ b/frontends/web/src/components/layout/header.tsx
@@ -25,6 +25,7 @@ import style from './header.module.css';
 interface HeaderProps {
     title?: string | JSX.Element | JSX.Element[];
     narrow?: boolean;
+    hideSidebarToggler?: boolean;
 }
 type Props = HeaderProps & SharedPanelProps & TranslateProps;
 
@@ -38,11 +39,11 @@ class Header extends Component<Props> {
   };
 
   public render() {
-    const { t, title, narrow, children, guideExists, shown, sidebarStatus } = this.props;
+    const { t, title, narrow, children, guideExists, shown, sidebarStatus, hideSidebarToggler } = this.props;
     return (
       <div className={[style.container, sidebarStatus ? style[sidebarStatus] : ''].join(' ')}>
         <div className={[style.header, narrow ? style.narrow : ''].join(' ')}>
-          <div className={style.sidebarToggler} onClick={toggleSidebar}>
+          <div className={`${style.sidebarToggler} ${hideSidebarToggler ? style.hideSidebarToggler : ''}`} onClick={toggleSidebar}>
             <MenuDark className="show-in-lightmode" />
             <MenuLight className="show-in-darkmode" />
           </div>
--- a/frontends/web/src/routes/new-settings/about.tsx
+++ b/frontends/web/src/routes/new-settings/about.tsx
@@ -15,25 +15,50 @@
  */
 
 import { useTranslation } from 'react-i18next';
-import { Main, Header } from '../../components/layout';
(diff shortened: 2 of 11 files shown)


frontends/web/src/routes/new-settings/mobile-settings.tsx
```diff
@@ -32,19 +33,20 @@ import { TPagePropsWithSettingsTabs } from './type';
...
+  const { t } = useTranslation();
   useEffect(() => {
     if (!isMobile) {
       route('/new-settings/appearance');
     }
   }, [isMobile]);
   return (
     <Main>
-      <Header title={<h2>Settings</h2>} />
+      <Header guideExists={false} title={<h2>{t('settings.title')}</h2>} />
```
shonsirsha, line 44: This PR also takes care some bits of code that apparently still not using locize for words, for example here ("Settings).

frontends/web/src/routes/new-settings/about.tsx
```diff
@@ -15,25 +15,50 @@
...
+      <GuidedContent>
+        <Main>
+          <Header
+            hideSidebarToggler
+            title={
+              <>
+                <h2 className="hide-on-small">{t('sidebar.settings')}</h2>
+                <MobileHeader withGuide title={t('settings.about')} />
+              </>
+            } />
```
thisconnect, line 41: The component nesting looks a bit funny now, but lets keep as you have it.

I guess we should have a nicer Header component that takes care, but probably the old header needs to be refactored to not use the custom store etc.

shonsirsha:
I agree to refactor the Header into 1. It's just we didn't have this use-case before (header w/ "Back" button and hidden collapse on mobile).

frontends/web/src/components/layout/header.tsx
```diff
@@ -38,11 +39,11 @@ class Header extends Component<Props> {
   };
 
   public render() {
-    const { t, title, narrow, children, guideExists, shown, sidebarStatus } = this.props;
+    const { t, title, narrow, children, guideExists, shown, sidebarStatus, hideSidebarToggler } = this.props;
     return (
       <div className={[style.container, sidebarStatus ? style[sidebarStatus] : ''].join(' ')}>
         <div className={[style.header, narrow ? style.narrow : ''].join(' ')}>
-          <div className={style.sidebarToggler} onClick={toggleSidebar}>
+          <div className={`${style.sidebarToggler} ${hideSidebarToggler ? style.hideSidebarToggler : ''}`} onClick={toggleSidebar}>
```
shonsirsha, line 46: Just made a change to hide this via CSS so we can hide it at the correct media query / screen size. Previously we're hiding the toggler as long as `hideSidebarToggler` is `true`. This causes a bug in medium screens where the toggler gets hidden (it shouldnt). Now, toggler isn't hidden anymore in medium screens.

Before:
Screenshot 2023-06-13 at 18 21 22
(no sidebar toggler)

After:

Screenshot 2023-06-13 at 18 21 05

PR #2159, opened 2023-06-13
backend/btc: avoid some panics

benma:
When the account and db closes, DB operations subsequently fail. The panics should be handled better anyway, but in the case that the account was closed, we can just abort there and prevent a crash.


Beerosagos (approved):
utACK

PR #2163, opened 2023-06-15
Update litecoin explorers

thisconnect:


thisconnect:
There is also https://github.com/digitalbitbox/bitbox-wallet-app/blob/3018895b8302796afeaee2952b2ba5e853a04668/backend/coins/ltc/params.go#L35

Could you check if that is still working or if it can be removed?

benma (approved):
utACK

benma:
> There is also
> 
> https://github.com/digitalbitbox/bitbox-wallet-app/blob/3018895b8302796afeaee2952b2ba5e853a04668/backend/coins/ltc/params.go#L35
> 
> Could you check if that is still working or if it can be removed?

We don't need this field at all, so it does not matter.

PR #2168, opened 2023-06-19
backend/accounts: make distinction between testnet and mainnet

benma:
If there is a testnet account stored (from a testnet run) and the app is then run in mainnet, the accounts scan would also scan testnet accounts. This commit fixes it to only look at the right coins depending on testnet or mainnet.


thisconnect (approved):
tested LGTM

PR #2177, opened 2023-06-23
frontend: add success recommendation for created with mnemonic

thisconnect:
The create wallet success step had a message specific to microSD card (remove the microSD card), added a message if wallet was created with mnemonic words instead.

Screen Shot 2023-06-23 at 09 50 46


Beerosagos (approved):
tACK! :+1:

PR #2178, opened 2023-06-23
frontend: add missing cz lang in settings

thisconnect:
The new settings dropdown is missing Čeština (CZ lang), but it is shown in the language menu during setup or in the waiting view.

Changed so that both use the same list of languages and rearranged all languages types in alphabetical order.


shonsirsha (approved):

shonsirsha:
Ups! We probably only added the currency and not the language for the new settings. Thanks!

thisconnect:
they got merged at about the same time, I should have noticed during review..

PR #2180, opened 2023-06-26
frontend: fix restore from mnemonic success message

thisconnect:
The restore from mnemonic success view the last point mentioned sdcard changed to security notice about recovery words.


Beerosagos (approved):
tACK!

PR #2181, opened 2023-06-26
frontend: fix breaking word in tabs navigation

thisconnect:
There is a global style to break text in long words, this is a precausion to always show long addresses and break the address in the worst case.

But this rule breaks the tab text of the new settings view on some breakpoints (medium-ish screeen size).


thisconnect:
Screen Shot 2023-06-26 at 13 07 51

without this change

thisconnect:
Screen Shot 2023-06-26 at 13 07 41

with this fix

Beerosagos (approved):
tACK on linux Qt

PR #2183, opened 2023-06-27
frontend: attestation check icon changed and fix RedDot icon

shonsirsha:
In the manage device setting, We used the wrong attestation check icon when the check fails. The RedDot icon (used in Firmware version check) also wasn't aligning nicely before due to its size. This PR fixes it.

Before:
Screenshot 2023-06-27 at 13 27 26

After:
Screenshot 2023-06-27 at 13 24 15


Beerosagos (approved):
untested LGTM

PR #2185, opened 2023-06-27
frontend: block toggle account until coin is enabled

thisconnect:
With the 'encrypt seed in RAM' firmware it can now take slightly longer to enable an account. Clicking the toggle quickly can deactivate the account again.

https://github.com/digitalbitbox/bitbox02-firmware/pull/1056

This change sets the toggle to disabled until the enabling is complete. To keep it simple this doesn't keep the disabled state of each toggle, but just fire and forgets.

Setting the toggle to disabled quickly changes the color. In a future commit this could be changed a loading indicator  instead of chaning it to disabled color.

### DIFF
--- a/frontends/web/src/routes/settings/manage-accounts.tsx
+++ b/frontends/web/src/routes/settings/manage-accounts.tsx
@@ -61,7 +61,7 @@ class ManageAccounts extends Component<Props, State> {
   };
 
   private fetchAccounts = () => {
-    accountAPI.getAccounts().then(accounts => this.setState({ accounts }));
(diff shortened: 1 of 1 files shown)


benma:
@thisconnect toggling an account does not involve the bitbox02 at all. The reason for the lag is that all accounts are re-initialized.

benma (approved):
Thanks! I updated the commit msg as it had nothing to do with encrypted seeds etc.

Btw, changes in this PR would be an excellent fit for the `async` keyword :grin:

PR #2188, opened 2023-07-02
frontend: display root fingerprint

benma:
Display the root fingerprint in the account info and in the Manage Device section. This helps identify the current wallet for passphrase users, and could be helpful for multisig/descriptor users where the root fingerprint is often used.

The fingerprint shown in both places because:

- useful in the device settings to quickly check the wallet ID after unlock
- useful in the account info for watch-only accounts in the future

### DIFF
--- a/frontends/web/src/routes/settings/components/device-settings/root-fingerprint-setting.tsx
+++ b/frontends/web/src/routes/settings/components/device-settings/root-fingerprint-setting.tsx
@@ -1,0 +1,35 @@
+/**
+ * Copyright 2023 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useTranslation } from 'react-i18next';
+import { SettingsItem } from '../settingsItem/settingsItem';
+
+type TProps = {
+    rootFingerprint: string;
+}
+
+const RootFingerprintSetting = ({ rootFingerprint }: TProps) => {
+  const { t } = useTranslation();
+  return (
+    <SettingsItem
+      settingName={'Root fingerprint'}
+      secondaryText={t('deviceSettings.deviceInformation.rootFingerprint.description')}
+      displayedValue={rootFingerprint}
+    />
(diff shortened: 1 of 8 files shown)


benma:
Replaces https://github.com/digitalbitbox/bitbox-wallet-app/pull/2188 which was quite out of date with all the refactorings and re-designs that happened since.

In the previous PR we debated if we should put it into the account info or into the device info section. Why not both! They each have their own purpose as described in the commit msg.

My1:
will it be the same algo as the coldcard uses for its master fingerprint/whatever?
also there's another advantage, passphrase users can see they are actually on the wrong wallet -> less panic

benma:
> will it be the same algo as the coldcard uses for its master fingerprint/whatever? also there's another advantage, passphrase users can see they are actually on the wrong wallet -> less panic

It's defined in BIP32, so yes.

My1:
neat, didnt know that. that's awesome. maybe more wallets will implement this in the future.

I would suggest that when opening passphrase wallets this ideally should be shown somewhere be it in the bitboxapp or the bitbox itself in a manner that doesnt require searching it somewhere deep.

frontends/web/src/routes/settings/components/device-settings/root-fingerprint-setting.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const RootFingerprintSetting = ({ rootFingerprint }: TProps) => {
+  const { t } = useTranslation();
+  return (
+    <SettingsItem
+      settingName={'Root fingerprint'}
+      secondaryText={t('deviceSettings.deviceInformation.rootFingerprint.description')}
+      displayedValue={rootFingerprint}
+    />
```
thisconnect, line 31: Since this is quiet a long text, please add:

```diff
diff --git a/frontends/web/src/routes/settings/components/settingsItem/settingsItem.module.css b/frontends/web/src/routes/settings/components/settingsItem/settingsItem.module.css
index 89d455103..889ad050b 100644
--- a/frontends/web/src/routes/settings/components/settingsItem/settingsItem.module.css
+++ b/frontends/web/src/routes/settings/components/settingsItem/settingsItem.module.css
@@ -32,6 +32,7 @@
 .rightContentContainer {
     display: flex;
     align-items: center;
+    flex-shrink: 0;
 }
 
 .displayedValue {
```

so it doesn't break the value:

Screen Shot 2023-07-04 at 15 35 11

thisconnect (approved):
tested LGTM with small CSS suggestion

frontends/web/src/routes/settings/components/device-settings/root-fingerprint-setting.tsx
benma: Thanks, done.

benma:
Merged but PR was not closed :thinking: https://github.com/digitalbitbox/bitbox-wallet-app/commit/5c5fa98451a768bb7805453c0285417f3d966d82 closing manually

PR #2189, opened 2023-07-03
frontend: change conditions to show BuyReceiveCTA

shonsirsha:
We're now showing BuyReceiveCTA properly to avoid blank screen on account screen on testnet.

Now, BuyReceiveCTA on the account screen will always show when account is empty (instead of two conditions: empty account and supported buy).

Only when the exchange supports the buy of that certain coin, the BuyReceiveCTA will render the Buy button along with a Receive button. When it's not supported, only Receive button (and the default CTA message) will be rendered. 🙂 
 
Before (testnet, empty acc):
Screenshot 2023-07-03 at 20 50 51

After (testnet, empty acc):
Screenshot 2023-07-03 at 20 42 31

On mainnet, there should be no difference. For reference:
After (mainnet, empty acc):
Screenshot 2023-07-03 at 20 44 06

### DIFF
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -193,7 +193,6 @@ export function Account({
     && transactions.success
     && transactions.list.length === 0;
 
-  const showBuyButton = exchangeBuySupported && isAccountEmpty;
 
   const actionButtonsProps = {
     code,
@@ -232,9 +231,10 @@ export function Account({
                 {!isAccountEmpty && <ActionButtons {...actionButtonsProps} />}
               </div>
             </div>
-            {showBuyButton && (
+            {isAccountEmpty && (
               <BuyReceiveCTA
                 code={code}
+                exchangeBuySupported={exchangeBuySupported}
                 unit={balance.available.unit}
                 balanceList={[[code, balance]]}
               />
--- a/frontends/web/src/routes/account/info/buyReceiveCTA.tsx
+++ b/frontends/web/src/routes/account/info/buyReceiveCTA.tsx
@@ -13,22 +13,29 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { route } from '../../../utils/route';
-import { CoinUnit, IBalance } from '../../../api/account';
+import { CoinUnit, IAccount, IBalance } from '../../../api/account';
 import { Button } from '../../../components/forms';
 import { Balances } from '../summary/accountssummary';
-import styles from './buyReceiveCTA.module.css';
 import { isBitcoinCoin } from '../utils';
+import { getExchangeSupportedAccounts } from '../../buy/utils';
+import styles from './buyReceiveCTA.module.css';
 
 type TBuyReceiveCTAProps = {
   balanceList?: [string, IBalance][];
   code?: string;
   unit?: string;
+  exchangeBuySupported?: boolean;
 };
 
-export const BuyReceiveCTA = ({ code, unit, balanceList }: TBuyReceiveCTAProps) => {
+type TAddBuyReceiveOnEmpyBalancesProps = {
+  balances?: Balances;
+  accounts: IAccount[];
+}
+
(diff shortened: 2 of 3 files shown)


frontends/web/src/routes/account/account.tsx
```diff
@@ -232,9 +231,10 @@ export function Account({
...
               </div>
             </div>
-            {showBuyButton && (
+            {isAccountEmpty && (
               <BuyReceiveCTA
                 code={code}
+                exchangeBuySupported={exchangeBuySupported}
                 unit={balance.available.unit}
                 balanceList={[[code, balance]]}
               />
```
thisconnect, line 240: I think it could also happen in accountsummary AddBuyReceiveOnEmptyBalances in testnet with eth only accounts.

shonsirsha: Yes accountSummary has the Buy button. I imagine we also wanna hide it here too right whenever it's not applicable / (not supported)? (only makes sense I guess 😄 )
Screenshot 2023-07-04 at 19 47 48

thisconnect: the conditions in AddBuyReceiveOnEmptyBalances are a bit hard to read and I am not even 100% sure if it is correct, especially the ` code={balanceList[0][0]}`. 

```js
if (balanceList.map(entry => entry[1].available.unit).every(isBitcoinCoin)) {
    return <BuyReceiveCTA code={balanceList[0][0]} unit={'BTC'} balanceList={balanceList} />;
  }
```

Could you double check if that all makes sense?

shonsirsha:
Changes made and fixes a bug in the CTA (2nd commit). PTAL @thisconnect 🙏

frontends/web/src/routes/account/info/buyReceiveCTA.tsx
```diff
@@ -48,14 +55,21 @@ export const BuyReceiveCTA = ({ code, unit, balanceList }: TBuyReceiveCTAProps)
...
-export const AddBuyReceiveOnEmptyBalances = ({ balances }: {balances?: Balances}) => {
-  if (balances === undefined) {
+export const AddBuyReceiveOnEmptyBalances = ({ balances, accounts }: TAddBuyReceiveOnEmpyBalancesProps) => {
+
+  const [supportedAccounts, setSupportedAccounts] = useState<IAccount[]>();
+
+  useEffect(() => {
+    getExchangeSupportedAccounts(accounts).then(setSupportedAccounts);
+  }, [accounts]);
+
```
shonsirsha, line 71: This API call is used get all supported accounts. If there's more than 1 account that's supported, "Buy" button will be shown. If there's no supported accounts, it won't.

frontends/web/src/routes/account/info/buyReceiveCTA.tsx
```diff
@@ -48,22 +55,31 @@ export const BuyReceiveCTA = ({ code, unit, balanceList }: TBuyReceiveCTAProps)
...
+  if (balances === undefined || supportedAccounts === undefined) {
```
thisconnect, line 72: Looks like  supportedAccounts could also be an empty array 🤔  

shouldn't it also check `|| supportedAccounts.length > 0` or similar?

thisconnect (approved):
tested LGTM with 1 question...

also got the empty space when changing from manage-accounts to summary (quickly after adding a new account) but I think this is known issue.

frontends/web/src/routes/account/info/buyReceiveCTA.tsx
shonsirsha: Although I don't think checking .length > 0 matters though as this component won't get rendered anyway  🤔 I put an undefined check there because of the async fetch, so when it's loading it's short-circuited (thus return `null`) instead of it keeps going.

shonsirsha: Actually, I don't think we should check for `.length > 0` here because if it's === 0 then it'll still have to return the BuyReceiveCTA but `exchangeBuySupported` is `false` which we're already doing. The `undefined` here signifies that `getExchangeSupportedAccounts ` is still loading thus I short-circuit here. 🙂

thisconnect: ah yeah you are totally right. thanks 👍

PR #2195, opened 2023-07-10
backend/accounts: fix scanning of first 5 accounts

benma:
The step to look for more accounts was only invoked when the previous account was used. If one had e.g. account 4 unused, but account 5 used, this could prevent discovery of account 5.

It worked anyway in some cases due to a race condition, as maybeAddHiddenUnusedAccounts could be invoked after account 4 was scanned (e.g. a different coin's account being added after acount 4 was scanned).


Beerosagos (approved):
tACK!

PR #2196, opened 2023-07-10
frontend: using context to replace guide store

shonsirsha:
As a part of the refactoring effort, we want to move away from class based components and HoC. In this PR, we're removing the guide store (and the HoC related to it). We're using context API to fulfil the same goal.

### DIFF
--- a/frontends/web/src/components/anchor/anchor.tsx
+++ b/frontends/web/src/components/anchor/anchor.tsx
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-import { ReactNode, SyntheticEvent } from 'react';
+import { ReactNode, SyntheticEvent, useContext } from 'react';
 import { route } from '../../utils/route';
-import { hide } from '../guide/guide';
 import { debug } from '../../utils/env';
 import { apiPost } from '../../utils/request';
+import AppContext from '../../contexts/AppContext';
 import style from './anchor.module.css';
 
 type TProps = {
@@ -28,12 +28,13 @@ type TProps = {
 }
 
 const A = ({ href, icon, children, ...props }: TProps) => {
+  const { setGuideShown } = useContext(AppContext);
   return (
     <span className={style.link} onClick={(e: SyntheticEvent) => {
       e.preventDefault();
       const { hostname, origin } = new URL(href, window.location.href);
       if (origin === 'qrc:' || (debug && hostname === window.location.hostname)) {
-        hide();
+        setGuideShown(false);
         route(href);
       } else {
         apiPost('open', href).catch(console.error);
--- a/frontends/web/src/components/guide/guide.tsx
+++ b/frontends/web/src/components/guide/guide.tsx
@@ -14,86 +14,35 @@
  * limitations under the License.
  */
 
-import { ReactNode, useEffect } from 'react';
+import { ReactNode, useContext, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
-import { share } from '../../decorators/share';
-import { Store } from '../../decorators/store';
-import { translate, TranslateProps } from '../../decorators/translate';
-import { setConfig } from '../../utils/config';
-import { apiGet } from '../../utils/request';
 import A from '../anchor/anchor';
 import { CloseXWhite } from '../icon';
+import AppContext from '../../contexts/AppContext';
 import style from './guide.module.css';
 
-export type TSharedProps = {
-    shown: boolean;
-    // eslint-disable-next-line react/no-unused-prop-types
-    activeSidebar: boolean;
-    // eslint-disable-next-line react/no-unused-prop-types
-    sidebarStatus: string;
-    // eslint-disable-next-line react/no-unused-prop-types
-    guideExists: boolean;
-}
-
-export type TProps = TSharedProps & {
+export type TProps = {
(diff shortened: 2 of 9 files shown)


frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -54,6 +61,11 @@ interface SwipeAttributes {
     active?: boolean;
 }
 
+export const panelStore = new Store<SharedPanelProps>({
+  activeSidebar: false,
+  sidebarStatus: '',
+});
+
```
shonsirsha, line 68: The old store now only contains things for sidebar. We will also eventually move this into the context, but we first have to convert the components that are using this store into functional cmpts. :)

frontends/web/src/components/guide/guide.tsx
```diff
@@ -14,86 +14,46 @@
...
+
+  useEffect(() => {
+    apiGet('config').then(({ frontend }) => {
+      if (frontend && frontend.guideShown !== undefined) {
+        setGuideShown(frontend.guideShown);
+      } else {
+        setGuideShown(true);
+      }
+    });
+  }, [setGuideShown]);
```
thisconnect, line 47: since you are here, please change to `import { getConfig } from '../../utils/config';`

thisconnect: this changes the behavior, before config was only read once on application start. With this PR it gets the config now each time on component mount (if I read correctly).

If we go back to the old behavior, should this go into AppProvider?

frontends/web/src/components/anchor/anchor.tsx
```diff
@@ -28,6 +28,10 @@ type TProps = {
 }
 
 const A = ({ href, icon, children, ...props }: TProps) => {
+  const { setGuideShown } = useContext(AppContext);
+  const hide = () => {
+    setGuideShown(false);
+  };
```
thisconnect, line 34: I think it's fine to just `setGuideShown(false);` and remove hide function.

frontends/web/src/components/layout/header.tsx
```diff
@@ -15,37 +15,40 @@
...
 import { toggleSidebar } from '../sidebar/sidebar';
+import AppContext from '../../contexts/AppContext';
 import style from './header.module.css';
-
 interface HeaderProps {
     title?: string | JSX.Element | JSX.Element[];
     narrow?: boolean;
     hideSidebarToggler?: boolean;
     children?: ReactNode;
+    guideExists?: boolean;
```
thisconnect, line 31: The only place where we use Header prop `guideExists={false}` is in mobile-settings.tsx component.

Is this still needed?

Could the Header component just ask the context if there is a guide (there should be none on mobile-settings)?

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -44,6 +44,13 @@ interface SubscribedProps {
     keystores?: Array<{ type: 'hardware' | 'software' }>;
 }
 
+export type SharedPanelProps = {
+  // eslint-disable-next-line react/no-unused-prop-types
+  activeSidebar: boolean;
+  // eslint-disable-next-line react/no-unused-prop-types
+  sidebarStatus: string;
+}
```
thisconnect, line 52: I wanted to ask if you could change sidebarStatus type from string to `'' | 'forceHidden'`...

but looking at it more closely: it seems to be only used to swip sidebar in/out on mobile, but it is set only by bitbox01.tsx 🤔  and BitBox01 is not supported on mobile 🤔 

It looks like it was initially added also for bitbox02, but BB02 doesn't need it anymore since setup changed to View component.
https://github.com/digitalbitbox/bitbox-wallet-app/pull/529

maybe lets keep as is and lets try to remove sidebarStatus in the future.

thisconnect (changes requested):

frontends/web/src/contexts/AppProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ReactNode, useState } from 'react';
+import { setConfig } from '../utils/config';
+import GuideAndSidebarContext from './AppContext';
```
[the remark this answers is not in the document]
shonsirsha: Oops, it's definitely shouldve been AppContext. Thanks!

frontends/web/src/components/sidebar/sidebar.tsx
shonsirsha: True. The value of sidebarStatus is a bit unclear.

frontends/web/src/components/guide/guide.tsx
shonsirsha: Ah yeah.

> If we go back to the old behavior, should this go into AppProvider?

Absolutely right 👍

shonsirsha:
Thanks for the review @thisconnect, PTAL 🙏

thisconnect (approved):
LGTM

PR #2197, opened 2023-07-11
frontend: fix layout of manage accounts with long account names

thisconnect:
On small screen long accountnames/edit/toggle break in various different ways.

Fixed by letting the name break, so that edit and toggle are always on the same line.


thisconnect:
Screen Shot 2023-07-11 at 11 50 35

(current master)

thisconnect:
Screen Shot 2023-07-11 at 11 50 15

with this fix

PR #2198, opened 2023-07-11
Wallet connect

Tomasvrba:


backend/keystore/keystore.go
```diff
@@ -95,8 +96,15 @@ type Keystore interface {
...
+	// an uncompressed pubkey.
+	SignETHTypedMessage(chainID uint64, data []byte, keypath signing.AbsoluteKeypath) ([]byte, error)
 
 	// SignTransaction signs the given transaction proposal. Returns ErrSigningAborted if the user
 	// aborts.
 	SignTransaction(interface{}) error
+
+	// SignETHWCTransaction signs a transaction proposed by Wallet Connect. Returns ErrSigningAborted if the user
+	// aborts.
+	SignETHWCTransaction(chainID uint64, tx *types.Transaction, keypath signing.AbsoluteKeypath) ([]byte, error)
```
benma, line 109: WC => WalletConnect

backend/devices/bitbox02/keystore.go
```diff
@@ -512,5 +513,45 @@ func (keystore *keystore) SignBTCMessage(message []byte, keypath signing.Absolut
 
 // SignETHMessage implements keystore.Keystore.
 func (keystore *keystore) SignETHMessage(message []byte, keypath signing.AbsoluteKeypath) ([]byte, error) {
-	return keystore.device.ETHSignMessage(params.MainnetChainConfig.ChainID.Uint64(), keypath.ToUInt32(), message)
+	signature, err := keystore.device.ETHSignMessage(params.MainnetChainConfig.ChainID.Uint64(), keypath.ToUInt32(), message)
+	if firmware.IsErrorAbort(err) {
+		return nil, errp.WithStack(keystorePkg.ErrSigningAborted)
```
benma, line 518: i think in aopp.go you need to switch the user abort check from the old way to this new way.

benma (commented):
Thanks!

I did a quick review of the backend parts. I'll leave the frontend to @shonsirsha and @thisconnect :)

I still need to look at the signing function in more detail and do some testing.

backend/coins/eth/account.go
```diff
@@ -812,3 +814,131 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
...
+	To       string `json:"to"`
+	Data     string `json:"data"`
+	Gas      string `json:"gas,omitempty"`
+	GasPrice string `json:"gasPrice,omitempty"`
+	Value    string `json:"value,omitempty"`
+	Nonce    string `json:"nonce,omitempty"`
+}
+
+// EthSignTypedMsg signs an Ethereum Tx received from WalletConnect
+func (account *Account) EthSignWcTx(
```
benma, line 857: pls document the args

shonsirsha (approved):
approving the FE side. 🚀  Some bugs which we (tom and I) have discussed that I'll make the fix following this PR (as our wallet connect UI will change massively anyway) - but this is a good PoC to start building the FE from.

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,86 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+	if err := json.NewDecoder(r.Body).Decode(&signInput); err != nil {
+		return nil, errp.WithStack(err)
+	}
+	ethAccount, ok := handlers.account.(*eth.Account)
+	if !ok {
+		return nil, errp.New("Must be an ETH based account")
+	}
+	signedMessage, err := ethAccount.EthSignMsg(signInput)
+	if errp.Cause(err) == keystore.ErrSigningAborted {
+		return map[string]interface{}{"success": false, "aborted": true}, nil
```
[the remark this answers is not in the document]
Tomasvrba: It's not ad-hoc, it's copied from your commit :P 
https://github.com/digitalbitbox/bitbox-wallet-app/commit/cc4c1b86270b0a2f2c63080b6b122abba8d1f8c8

Should I change it there as well if I make a new struct with the "aborted" field?

backend/devices/bitbox02/keystore.go
Tomasvrba: Without digging into it in more detail, I don't really see how to make the aopp behavior the same as it has its own aopp error handling:
```
func (backend *Backend) aoppSetError(err ErrorCode) {
	backend.aopp.State = aoppStateError
	backend.aopp.ErrorCode = err
	backend.notifyAOPP()
}
```

imo aopp doesn't need to be changed as part of this PR

Tomasvrba:
@benma
Pushed !fixups which address most of the comments above (unless I commented otherwise why it won't be addressed at the moment)
@shonsirsha @thisconnect know their UI/UX FE TODOs from our meeting

Remaining TODOs for me: 
Make a new API call to the backend for hex message conversion to human readable
Filter accounts based on address so that Wallet Connect requests can be handled from anywhere in the app, not just the account page corresponding to the request

@jadzeidan Lots and lots of testing please on all EVM networks and report bugs :)

Imo this PR is more or less ready to merge into the staging branch and can be worked on further there. Maybe @shonsirsha you can take a look at the failing AppVeyor build and try to fix my poor react hook dependency arrays 😇 😇

shonsirsha:
@Tomasvrba seems like it's a linter issue (warnings). If we want to merge this soon (which makes sense) I'd probably just silence these warnings for the time being (just so our CI stops failing), as it'll mostly be reimplemented anyway. 

Otherwise, I can take a look at it later today or tomorrow! :)

shonsirsha:
Hey @Tomasvrba , unsure how you want to incorporate my changes to fix the linter warnings (so checks turn green) - but I made a PR to this branch: https://github.com/Tomasvrba/bitbox-wallet-app/pull/1, so maybe once you merge the checks will finally pass and we can merge this one?

shonsirsha:
Should we merge this? :)

thisconnect:
I think so, but please fixup the commits first.

Tomasvrba:
> ~I think so, but please fixup the commits first.~
> 
> otoh lets wait for benma's review.

Ha, too late, didn't see this xD
I pushed the pre-fixup branch here: https://github.com/digitalbitbox/bitbox-wallet-app/compare/master...Tomasvrba:bitbox-wallet-app:wc-pre-fixup :)

frontends/web/src/api/account.ts
```diff
@@ -324,3 +324,29 @@ export const addAccount = (coinCode: string, name: string): Promise<TAddAccount>
 export const testRegister = (pin: string): Promise<null> => {
   return apiPost('test/register', { pin });
 };
+
+export type TSignMessage = {
```
benma, line 328: Try changing the type to sth like this

https://github.com/digitalbitbox/bitbox-wallet-app/blob/c4375864155bffc353c11749e116d3adf2a0f9db/frontends/web/src/api/account.ts#L177

which enforces checking the success flag before accessing the fields only available on success.

same for the type below.

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,105 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+type errorResponse struct {
+	Success      bool   `json:"success"`
+	ErrorMessage string `json:"errorMessage"`
+	ErrorCode    string `json:"errorCode"`
+}
+
+func (handlers *Handlers) postEthSignMsg(r *http.Request) (interface{}, error) {
+	var signInput string
+	if err := json.NewDecoder(r.Body).Decode(&signInput); err != nil {
+		return nil, errp.WithStack(err)
```
benma, line 601: return errorResponse

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,105 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+	signature, err := ethAccount.SignMsg(signInput)
+	if errp.Cause(err) == keystore.ErrSigningAborted {
+		return signingResponse{Success: false, Aborted: true}, nil
+	}
+	if err != nil {
+		handlers.log.WithError(err).Error("Failed to sign message")
+		result := errorResponse{Success: false, ErrorMessage: err.Error()}
+		return result, nil
+	}
+	return map[string]interface{}{
```
benma, line 616: return a struct instance not a map also here

backend/coins/eth/account.go
```diff
@@ -812,3 +814,133 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
...
+			return "", "", errp.WithStack(errors.ErrInsufficientFunds)
+		}
+		account.log.WithError(err).Error("Could not estimate the gas limit.")
+		return "", "", errp.WithStack(errors.TxValidationError(err.Error()))
+	}
+
+	for _, t := range account.feeTargets() {
+		//TODO Let user choose gas price/priority
+		if t.code == accounts.FeeTargetCodeNormal {
+			if t.gasPrice.Cmp(big.NewInt(0)) <= 0 {
```
benma, line 919: i wonder why this check was needed? can gasPrice be negative?

backend/coins/eth/account.go
```diff
@@ -812,3 +814,133 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
...
+		if strings.Contains(err.Error(), etherscan.ERC20GasErr) {
+			return "", "", errp.WithStack(errors.ErrInsufficientFunds)
+		}
+		account.log.WithError(err).Error("Could not estimate the gas limit.")
+		return "", "", errp.WithStack(errors.TxValidationError(err.Error()))
+	}
+
+	for _, t := range account.feeTargets() {
+		//TODO Let user choose gas price/priority
+		if t.code == accounts.FeeTargetCodeNormal {
```
benma, line 918: if this fee target does not exist you should probably error? not sure what happens if you proceed here with a 0 gas price.

backend/coins/eth/account.go
```diff
@@ -812,3 +814,133 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
...
+	Data     string `json:"data"`
+	Gas      string `json:"gas,omitempty"`
+	GasPrice string `json:"gasPrice,omitempty"`
+	Value    string `json:"value,omitempty"`
+	Nonce    string `json:"nonce,omitempty"`
+}
+
+// EthSignTypedMsg signs an Ethereum Tx received from WalletConnect
+func (account *Account) EthSignWalletConnectTx(
+	// send: whether transaction should be broadcast after signing
```
benma, line 858: maybe document the use case for send=false - i wonder how the caller can use a signature or txhash without knowing the transaction which is constructed inside this function but not returned and not broadcast.

backend/keystore/keystore.go
```diff
@@ -87,16 +88,26 @@ type Keystore interface {
...
 	// Electrum format.
 	SignBTCMessage(message []byte, keypath signing.AbsoluteKeypath, scriptType signing.ScriptType) ([]byte, error)
+
 	// SignETHMessage signs the message using thep rivate key at the keypath. The result contains a
 	// 65 byte signature. The first 64 bytes are the secp256k1 signature in / compact format (R and
-	// S values), and the last byte is the recoverable id (recid).
+	// S values), and the last byte is the recoverable id (recid). Returns ErrSigningAborted if the user aborts.
 	SignETHMessage(message []byte, keypath signing.AbsoluteKeypath) ([]byte, error)
 
+	// ETHSignTypedMessage signs an Ethereum EIP-612 typed message. 27 is added to the recID to denote
```
benma, line 102: it's EIP-712.

you can also copy the same for above for clarity:

```
         // The result contains a
	// 65 byte signature. The first 64 bytes are the secp256k1 signature in / compact format (R and
	// S values), and the last byte is the recoverable id (recid).
```

vice versa, you can add `27 is added to the recID to denote an uncompressed pubkey.` to the doc of `SignETMessage()`, otherwise it looks like one is doing it and the other is not.

benma (changes requested):

frontends/web/src/routes/account/account.tsx
```diff
@@ -228,7 +228,13 @@ export function Account({
                 <label className="labelXLarge flex-self-start-mobile show-on-small">
                   {t('accountSummary.availableBalance')}
                 </label>
-                {!isAccountEmpty && <ActionButtons {...actionButtonsProps} />}
+                {!isAccountEmpty ?
+                  <ActionButtons {...actionButtonsProps} /> :
+                  code.includes('eth') && !code.includes('erc20') && (
```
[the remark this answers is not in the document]
shonsirsha: I think it's okay (at least for now) to allow testnets during development. Is there any downsides in allowing testnets?

<hr />

For more info, I was working on this part right here too and I agree it needs refactoring and some bugfixing. I noticed this bug, for example:

Screenshot 2023-07-31 at 15 36 33


and I'm planning to just address it on my own (in my PR, that im working on right now). I've been assuming that this PR is merely a POC of implementation of WC and we can fix minor bugs or typings in the next PR (talking about FE). Especially in the FE, **many** (if not, most) things are going to be changed.

However it seems like we're trying to fix most things right here in this PR.. so I'll try to do an additional review now. 🙏 :) I'll halt my work on WC until this one gets merged to the staging branch.

shonsirsha (changes requested):
additional reviews. Thanks.

shonsirsha:
@Tomasvrba , I made a small PR to fix some of my comments here (to the fixup branch): https://github.com/Tomasvrba/bitbox-wallet-app/pull/2 :)  I resolved some of comments above that are already fixed in that PR. Thanks! 💙

frontends/web/src/routes/account/account.tsx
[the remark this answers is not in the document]
benma: @shonsirsha i tagged you not because of testnet but so you are aware of the pitfall of using account codes / strings to identify coins.

i agree frontend stylings / UI etc can be fixed in future PRs. i still thought that the logical check here (is it ETH and not ERC?) was gonna stay, so might as well fix it here.

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,105 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+	Success bool `json:"success"`
+	Aborted bool `json:"aborted"`
+}
+
+type errorResponse struct {
+	Success      bool   `json:"success"`
+	ErrorMessage string `json:"errorMessage"`
+	ErrorCode    string `json:"errorCode"`
+}
+
```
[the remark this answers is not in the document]
Tomasvrba: Sure, I'll combine them. 
I didn't think of a user abort as an "error" per se, since nothing went wrong, but a valid signing response with a conscious choice by the user to abort

backend/coins/eth/account.go
Tomasvrba: Not sure tbh, you added it here: https://github.com/digitalbitbox/bitbox-wallet-app/commit/d5653be05589f2d14a2ab9df571fb91d244482e4

Maybe a sanity check for an invalid custom fee set by the user? Though I supposed that should be sanitized in the frontend

frontends/web/src/routes/account/walletconnect/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+  'eip155:42161' // Arbitrum One
+];
+
+export const EIP155_SIGNING_METHODS = {
+  PERSONAL_SIGN: 'personal_sign',
+  ETH_SIGN: 'eth_sign',
+  ETH_SIGN_TRANSACTION: 'eth_signTransaction',
+  ETH_SIGN_TYPED_DATA: 'eth_signTypedData',
+  ETH_SIGN_TYPED_DATA_V3: 'eth_signTypedData_v3',
+  ETH_SIGN_TYPED_DATA_V4: 'eth_signTypedData_v4',
```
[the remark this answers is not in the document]
Tomasvrba: This is a bit trickier than you'd think as with many poorly implemented ethereum things. The reality is that many dapps will request an 'eth_signTypedData' but in reality expect 'eth_signTypedData_v4' because that's basically the standard now and v4 is assumed, but not necessarily asked for specifically. 
We could probably get rid of  eth_signTypedData_v3 since I don't think anybody uses that specifically, though I would err on the side of caution, leave this as is, and assume all sign typed data requests are asking for v4 and return that

backend/coins/eth/account.go
Tomasvrba: yeah, my bad, the WC docs weren't completely clear on this, but on second look it wants the raw transaction data which can later be submitted with sendRawTransaction
changed in the current fixup

backend/coins/eth/account.go
Tomasvrba: Currently this should always exist afaik
Since ethgasstation doesn't work anymore, the fallback always sets FeeTargetCodeNormal
https://github.com/digitalbitbox/bitbox-wallet-app/blob/3fb59738e0c570bb3b5ba8817479df2c2a745405/backend/coins/eth/account.go#L725

This is to be changed after adding custom gas price to WC transactions (TODO L#917)

backend/coins/eth/account.go
```diff
@@ -812,3 +814,133 @@ func (account *Account) VerifyAddress(addressID string) (bool, error) {
...
+				return "", "", errors.ErrFeeTooLow
+			}
+			gasPrice = t.gasPrice
+		}
+	}
+
+	tx := types.NewTransaction(nonce,
+		*message.To,
+		message.Value, gasLimit, gasPrice, message.Data)
+	//TODO edit signer to match chainID proposed by wallet connect
```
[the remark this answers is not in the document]
Tomasvrba: L#934 the signer config *params.ChainConfig is made with account.coin.Net() where the ChainID on the ChainConfig will I believe only be 1 currently in the BBApp. 
Which I think will not work with other networks right now because of tx replay 

Not sure what the best approach is here, copy the *params.ChainConfig from account.coin.Net() and keep it the same except for overwriting the ChainID?

Expanded the TODO

frontends/web/src/api/account.ts
```diff
@@ -324,3 +324,29 @@ export const addAccount = (coinCode: string, name: string): Promise<TAddAccount>
...
+  return apiPost(`account/${code}/eth-sign-msg`, message);
+};
+
+export const ethSignTypedMessage = (code: AccountCode, chainId: any, data: any): Promise<TSignMessage> => {
+  return apiPost(`account/${code}/eth-sign-typed-msg`, { chainId, data });
+};
+
+export const ethSignWalletConnectTx = (code: AccountCode, send: boolean, chainId: any, tx: any): Promise<TSignWalletConnectTx> => {
+  return apiPost(`account/${code}/eth-sign-wallet-connect-tx`, { send, chainId, tx });
+};
```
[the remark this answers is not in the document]
Tomasvrba: changed the chainID
Unfortunately both data and tx come from the wallet connect requestEvent: SignClientTypes.EventArguments['session_request'] type where both are typed as `any` : https://github.com/WalletConnect/walletconnect-monorepo/blob/f1d78b6718f063f301894462bfffdbece3ce7533/packages/types/src/sign-client/client.ts#L40

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,102 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+	Signature    string `json:"signature"`
+	Aborted      bool   `json:"aborted"`
+	ErrorMessage string `json:"errorMessage"`
+	ErrorCode    string `json:"errorCode"`
+}
+
+func (handlers *Handlers) postEthSignMsg(r *http.Request) (interface{}, error) {
+	var signInput string
+	if err := json.NewDecoder(r.Body).Decode(&signInput); err != nil {
+		return signingResponse{Success: false, ErrorMessage: errp.WithStack(err).Error()}, nil
```
benma, line 598: the stack is actually not printed like this nor should it. `err.Error()` should be the same

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,102 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+		return signingResponse{Success: false, ErrorMessage: "Must be an ETH based account"}, nil
+	}
+	txHash, rawTx, err := ethAccount.EthSignWalletConnectTx(args.Send, args.ChainId, args.Tx)
+	if errp.Cause(err) == keystore.ErrSigningAborted {
+		return signingResponse{Success: false, Aborted: true}, nil
+	}
+	if err != nil {
+		handlers.log.WithError(err).Error("Failed to send transaction")
+		result := signingResponse{Success: false, ErrorMessage: err.Error()}
+		if err.Error() == errors.ErrInsufficientFunds.Error() {
```
benma, line 674: better:

if `errp.Cause(err) == errors.ErrInsufficientFunds`.

backend/coins/btc/handlers/handlers.go
```diff
@@ -580,3 +583,102 @@ func (handlers *Handlers) postSetTxNote(r *http.Request) (interface{}, error) {
...
+			result.ErrorCode = err.Error()
```
benma, line 675: is this error code even used in the frontend or anywhere?

frontends/web/src/api/account.ts
```diff
@@ -324,3 +324,29 @@ export const addAccount = (coinCode: string, name: string): Promise<TAddAccount>
 export const testRegister = (pin: string): Promise<null> => {
   return apiPost('test/register', { pin });
 };
+
+export type TSignMessage = { success: false, aborted?: boolean; errorMessage?: string; } | { success: true; signature?: string; }
```
benma, line 328: should the signature not be mandatory in case of success, i.e. `signature: string` (without the question mark)?

backend/coins/eth/account.go
benma: the check is also there for the not-custom fees. i guess let's leave it, just in case the gas price estimator breaks in unexpected ways.

backend/coins/eth/account.go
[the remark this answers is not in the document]
benma: so signing a tx currently does not work for non-ETH chains?

if so, i'd check the passed chainID and error out early if it is not equal to `account.coin.ChainID(). and worry about how to fix it in a future PR.

frontends/web/src/routes/account/walletconnect/index.tsx
```diff
@@ -0,0 +1,474 @@
...
+            return { success: false, error: result };
+          },
+          currentSession, topic, id, 'Sign Typed Data', JSON.stringify(typedData));
+        }
+        return;
+      //TODO: for the following two cases, let user set own gas price and nonce in BBApp
+      case EIP155_SIGNING_METHODS.ETH_SEND_TRANSACTION:
+        if (currentSession) {
+          handleSessionRequest(async () => {
+            const result = await ethSignWalletConnectTx(code, true, +params.chainId.split(':')[1], request.params[0]);
```
benma, line 218: that `+` business looks like JS hackery :joy: consider the explicit `parseInt()`.

what if chainID does not contain a `:`?

maybe `params.chainID.replace(/^eip155\:/, '')` is better to strip of the prefix, as then it's also clear what the format is at the same time.

frontends/web/src/routes/account/walletconnect/utils.ts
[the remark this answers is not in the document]
benma: cool. please add this comment as a comment in the code there, very helpful :smile:

benma (changes requested):
Thanks, LGTM!

Just a few nits (pick or leave at will :smile: ).

Before we can merge the CI needs to be fixed.

There is one issue with the updated mock files that I addressed here:

https://github.com/digitalbitbox/bitbox-wallet-app/pull/2216

Once that is merged you can simply rebase.

And then there is:

```
+ golangci-lint run
backend/coins/eth/account.go:819:86: Comment should end in a period (godot)
// SignMsg is used for personal_sign and eth_sign messages in BBApp via WalletConnect
                                                                                     ^
backend/coins/eth/account.go:834:83: Comment should end in a period (godot)
// SignTypedMsg signs an Ethereum EIP-612 typed message in BBApp via WalletConnect
                                                                                  ^
backend/coins/eth/account.go:846:105: Comment should end in a period (godot)
// TX proposal arguments received from Wallet Connect with Gas, GasPrice, Value and Nonce being optional
                                                                                                        ^
```

Btw you can run this in dockerdev using `golangci-lint run`.

Also please squash all commits.

Tomasvrba:
> Thanks, LGTM!
> 
> Just a few nits (pick or leave at will 😄 ).

Thanks, will get all of that done sometime this week 👍

shonsirsha (approved):
approving for FE side

benma:
@Tomasvrba https://github.com/digitalbitbox/bitbox-wallet-app/pull/2216 is now merged. If you rebase onto staging-walletconnect again that CI issue should be fixed

backend/coins/btc/handlers/handlers.go
Tomasvrba: Good catch, it isn't right now and the ErrorMessage should be sufficient in any case since it contains the same err.Error(), removed.

frontends/web/src/routes/account/walletconnect/index.tsx
Tomasvrba: That's no hackery :D
Standard Unary Plus: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus
imo completely sufficient for this usecase :) 
https://i.stack.imgur.com/LLrgj.png

chainID will always be in the 'eip155:' format, but I will replace it with the regex 👍

Tomasvrba:
>  so signing a tx currently does not work for non-ETH chains?

> if so, i'd check the passed chainID and error out early if it is not equal to `account.coin.ChainID(). and worry about how to fix it > in a future PR.

Right, yeah, added an error to catch that and a TODO to fix. 
Duh, this is actually a little more complicated since we obviously can't broadcast L2 TXs from the backend using geth, we will need RPCs for each corresponding EVM network

benma (approved):

PR #2202, opened 2023-07-13
Frontend fix manage account

thisconnect:
Tokens did not break correctly to their own section.

Regression introduced in https://github.com/digitalbitbox/bitbox-wallet-app/commit/87b820d99dedf36050e89d7740987b03f405e2b7

This is only visible with make servewallet-mainnet as tokens are
only shown in mainnet.

also fixed className support on toggles see 2nd commit

### DIFF
--- a/frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
+++ b/frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
@@ -48,8 +48,7 @@ export const ToggleShowFirmwareHash = ({ deviceID }: Props) => {
       <Toggle
         checked={enabledState}
         id="togggle-show-firmware-hash"
-        onChange={handleToggle}
-        className="text-medium" />
+        onChange={handleToggle} />
     </div>
   );
 };
--- a/frontends/web/src/components/toggle/toggle.tsx
+++ b/frontends/web/src/components/toggle/toggle.tsx
@@ -1,19 +1,15 @@
-import { ChangeEvent } from 'react';
 import style from './toggle.module.css';
 
-export type TToggleProps = {
-  checked: boolean;
-  className?: string;
-  disabled?: boolean;
-  id?: string;
-  name?: string;
-  onChange: (event: ChangeEvent<HTMLInputElement>) => void;
-  value?: string;
-};
+export type TToggleProps = JSX.IntrinsicElements['input']
 
-export const Toggle = (props: TToggleProps) => {
+export const Toggle = (
+  {
+    className = '',
+    ...props
+  }: TToggleProps
+) => {
   return (
-    <label className={style.container}>
+    <label className={`${style.container} ${className}`}>
       <input
         type="checkbox"
         {...props} />
--- a/frontends/web/src/components/toggle/toggle.tsx
+++ b/frontends/web/src/components/toggle/toggle.tsx
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2023 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


thisconnect:
the input element is hidden and only used with :checked pseudo selector

Screen Shot 2023-07-13 at 18 24 44

frontends/web/src/components/devices/bitbox02bootloader/toggleshowfirmwarehash.tsx
```diff
@@ -48,8 +48,7 @@ export const ToggleShowFirmwareHash = ({ deviceID }: Props) => {
       <Toggle
         checked={enabledState}
         id="togggle-show-firmware-hash"
-        onChange={handleToggle}
-        className="text-medium" />
+        onChange={handleToggle} />
     </div>
   );
 };
```
thisconnect, line 54: `text-medium` did nothing here

Screen Shot 2023-07-13 at 18 24 44

thisconnect:
rebased

frontends/web/src/components/toggle/toggle.tsx
```diff
@@ -1,19 +1,15 @@
...
-export type TToggleProps = {
-  checked: boolean;
-  className?: string;
-  disabled?: boolean;
-  id?: string;
-  name?: string;
-  onChange: (event: ChangeEvent<HTMLInputElement>) => void;
-  value?: string;
-};
+export type TToggleProps = JSX.IntrinsicElements['input']
```
shonsirsha, line 19: nice!

frontends/web/src/routes/settings/manage-accounts.module.css
```diff
@@ -21,18 +21,24 @@
...
     display: flex;
+    flex-basis: calc(100% - 140px);
     flex-grow: 1;
 }
 
 .editBtn {
     background: transparent;
     border: none;
     color: var(--color-primary);
+    flex-basis: 60px;
```
shonsirsha, line 32: quick question, is this intended? The spacing between "Edit" and the toggle is not the same anymore now (bigger).

Before (master):
Screenshot 2023-07-17 at 17 33 17


After:
Screenshot 2023-07-17 at 17 32 45

On mobile (>=375px width), without setting the basis still looks fine:
Screenshot 2023-07-17 at 17 38 05


Do we need to set flex-basis here? 🤔

shonsirsha (approved):
Approving with a question.

> Regression introduced in https://github.com/digitalbitbox/bitbox-wallet-app/commit/87b820d99dedf36050e89d7740987b03f405e2b7

> This is only visible with make servewallet-mainnet as tokens are
only shown in mainnet.

Makes sense, thanks for noticing! 👍

frontends/web/src/routes/settings/manage-accounts.module.css
thisconnect: I did change it to flex wrap so the tokens could break into its own section

```
┌────────────────────────────────┬──────┬──────┐
│                                │      │      │
├────────────────────────────────┴──────┴──────┤
│                                              │
│                                              │
│                                              │
│                                              │
└──────────────────────────────────────────────┘
```

but now I noticed that with different translations the text "Edit" in German "Bearbeiten" breaks :( this could be solved by increasing the flex-basis 60px but doesn't feel right.. i.e. with different os the font-rendering can take more space..

I'll try to make the edit column growing...

thisconnect: I pushed a fix, please test in mainnet in German and use long account name so that the account name breaks onto 2 lines. I upped to 180px safety, as iirc on Linux font rendering may use more space.

shonsirsha (approved):
> I pushed a fix, please test in mainnet in German and use long account name so that the account name breaks onto 2 lines. I upped to 180px safety, as iirc on Linux font rendering may use more space.

LGTM! My mistake, I was only considering "English" when testing.

Screenshot 2023-07-18 at 09 51 06

Screenshot 2023-07-18 at 09 51 13


Just a note, I think another option (if for some reason super long name still messes up with our "responsive" approach in certain language) we can just redact the account name either at the end or in the middle using ellipsis (...). On mobile we can also alter their text size and the icon size. 

This way we will always have the same "toggle container / box" height, which makes our UI more aesthetically pleasing while not eliminating important information.

PR #2207, opened 2023-07-20
frontend: refactored appupgraderequired view

thisconnect:
Changed AppUpgradeRequired component to use View component, also changed download link to use Button and added ViewFooter to display language menu.

Another noticeable change is that the view is now fullscreen without
showing the sidebar that contains settings.

Test by lowering semver number for lowestSupportedFirmwareVersion and lowestNonSupportedFirmwareVersion in:
- bitbox02-api-go/api/firmware/device.go

### DIFF
--- a/frontends/web/src/components/appupgraderequired.tsx
+++ b/frontends/web/src/components/appupgraderequired.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2023 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


thisconnect:
Screen Shot 2023-07-20 at 10 00 10

(new with this PR)

thisconnect:
Screen Shot 2023-07-20 at 10 05 17

(without this change, current master)

frontends/web/src/components/appupgraderequired.tsx
```diff
@@ -15,27 +15,26 @@
...
-          </div>
-        </div>
-      </div>
-    </div>
+    <Main>
+      <Header />
+      <View
+        fullscreen
+        verticallyCentered
+        width="840px"
```
shonsirsha, line 31: May be better that one day we have these (and media query values) to be using variables instead of hardcoded.

shonsirsha (approved):
LGTM

frontends/web/src/components/appupgraderequired.tsx
thisconnect: yes makes sense, currently it's ok I think, there is a `min-width: 100%` so this prop is more like desired width but can go smaller.

Maybe like this: `style={{ "--content-width-large": "840px", "--content-width-small": "640px" }}`?

thisconnect:
Updated comment and commit message.

PR #2210, opened 2023-07-20
Bitsurance first commits

Beerosagos:
This integrates the first part of the Bitsurance service and includes:
- logic to compute the bitsurance id from an account xpub
- logic to fetch the up-to-date insured ids list form bitsurance server and look for one or more given accounts
- a new bitsurance page that allows triggering the insurance status check of the enabled accounts manually
- an automatic check of the insurance status of already insured accounts, and notification in case of insurance expiration.
More details are available in the commit messages :)

What is not included yet (and will in future PRs):
- backend TEST
- switch backend insurance check towards a per-account endpoint instead of downloading a full list 
- showing only native segwit address on receive page
- alert if an insured account has non-p2wpkh UTXOs
- integration of an onboarding widget by Bitsurance
- Bitsurance T&C page
- Guide texts
- a lot of css


backend/accounts.go
```diff
@@ -188,6 +190,63 @@ func (backend *Backend) SupportedCoins(keystore keystore.Keystore) []coinpkg.Cod
 	return availableCoins
 }
 
+// LookupInsuredAccounts checks the insurance status of one or more accounts. If the input
+// account code is not empty, the check is made exclusively on that account, otherwise it is done
+// on all the BTC/TBTC, active, not hidden accounts.
+// In case of updates of the insurance status of any account, the new status is persisted and the
+// accounts are reinitialized.
+// It returns a slice of insured account codes.
+func (backend *Backend) LookupInsuredAccounts(accountCode string) ([]types.Code, error) {
```
benma, line 199: can this be `accountCode types.Code`? 

I also wonder why it is a string here and immediately converted to the `Code` type:

```
func (backend *Backend) GetAccountFromCode(code string) (accounts.Interface, error) {
	acctCode := accountsTypes.Code(code)
```

backend/accounts.go
```diff
@@ -188,6 +190,63 @@ func (backend *Backend) SupportedCoins(keystore keystore.Keystore) []coinpkg.Cod
...
+		if accountConfig.Insured != insured {
+			backend.log.Infof("Account [%s] insurance status changed to %v", code, insured)
+			err := backend.config.ModifyAccountsConfig(func(_ *config.AccountsConfig) error {
+				accountConfig.Insured = insured
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+			backend.ReinitializeAccounts()
```
benma, line 241: why is this needed? maybe `backend.emitAccountsStatusChanged()` is enough?

either way, better to do it once after the loop instead of in every loop iteration.

backend/accounts.go
```diff
@@ -188,6 +190,63 @@ func (backend *Backend) SupportedCoins(keystore keystore.Keystore) []coinpkg.Cod
...
+	// if any account insurance status changed, persist the change and reinitialize the accounts.
+	for code, insured := range bitsuranceAccounts {
+		accountConfig := backend.Config().AccountsConfig().Lookup(code)
+		if accountConfig == nil {
+			return nil, errp.Newf("Could not find account %s", code)
+		}
+		if accountConfig.Insured != insured {
+			backend.log.Infof("Account [%s] insurance status changed to %v", code, insured)
+			err := backend.config.ModifyAccountsConfig(func(_ *config.AccountsConfig) error {
+				accountConfig.Insured = insured
```
benma, line 235: feels flaky to change it this way and not use the actual parameter. 

imo it is clearer and safer to wrap the whole loop in `ModifyAccountsConfig()` and use the param there to look up the accontConfig, to read and modify the value.

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+
+	res, err := httpClient.Do(req)
+	if err != nil {
+		return nil, errp.WithStack(err)
+	}
+	defer res.Body.Close() //nolint:errcheck
+	if res.StatusCode != http.StatusOK {
+		return nil, errp.Newf("%s - bad response code %d", endpoint, res.StatusCode)
+	}
+	const max = 81920
```
benma, line 58: isn't this way to small of a limit?

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+	return strings.Split(string(responseBody), "\n"), nil
+}
+
+// findInsuredAccounts compares `accountIds` and `insuredIds` and returns a slice
+// with the matching elements.
+//
+// `accountIds` is an ordered slice of strings, containing the hashes of the
+// requested wallet accounts xpubs
+// `insuredIds` is another ordered slice of strings, containing the hashes of
+// all the insured xpubs retrieved from the Bitsurance server.
```
benma, line 72: "containing the IDs" is maybe clearer

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+
+// BitsuranceGetId returns the BitsuranceId of a given account.
+// The id is computed hashing with sha256 the P2WPKH xpub of the account, concatenated with a fixed salt.
+// If a P2WPKH xpub can't be found in the account, empty string is returned.
+func BitsuranceGetId(account accounts.Interface) string {
+	for _, signingConf := range account.Info().SigningConfigurations {
+		bitcoinScriptType := signingConf.BitcoinSimple
+		if bitcoinScriptType != nil && bitcoinScriptType.ScriptType == signing.ScriptTypeP2WPKH {
+			h := sha256.New()
+			h.Write([]byte(bitcoinScriptType.KeyInfo.ExtendedPublicKey.String() + xpubSalt))
```
benma, line 112: can just do on one line: `hash := sha256.Sum256(...)` then `hex.EncodeToString(hash[:])`

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+
+	bitsuranceIds, err := fetchBitsuranceIds(httpClient)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, account := range accounts {
+		code := account.Config().Config.Code
+		accountId := BitsuranceGetId(account)
+		idCodeMap[accountId] = code
```
benma, line 141: you probably want to skip this if accountId is empty, otherwise there could be clashes with multiple empty ones, and also potentially flag ETH accounts as insured if the bitsurance ID list contains a non-filtered empty line.

backend/handlers/handlers.go
```diff
@@ -1058,6 +1069,29 @@ func (handlers *Handlers) getExchangesByRegion(r *http.Request) interface{} {
...
+		return errorResult{Error: err.Error()}
+	}
+
+	return insuredAccounts
+}
+
+func (handlers *Handlers) getBitsuranceURL(r *http.Request) interface{} {
+	lang := handlers.backend.Config().AppConfig().Backend.UserLanguage
+	if len(lang) == 0 {
+		// userLanguace config is empty if the set locale matches the system locale, so we have
```
benma, line 1088: typo

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+			return hex.EncodeToString(h.Sum(nil))
+		}
+	}
+	return ""
+
+}
+
+// BitsuranceURL reutrns the url for the Bitsurance widget for a given locale.
+func BitsuranceURL(lang, bitsuranceId string) string {
+	return widgetURL + lang + "/?id=" + bitsuranceId
```
benma, line 122: i guess bitsurance falls back to English if the lang is unknown to them?

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+			h := sha256.New()
+			h.Write([]byte(bitcoinScriptType.KeyInfo.ExtendedPublicKey.String() + xpubSalt))
+			return hex.EncodeToString(h.Sum(nil))
+		}
+	}
+	return ""
+
+}
+
+// BitsuranceURL reutrns the url for the Bitsurance widget for a given locale.
```
benma, line 120: typo

backend/bitsurance/bitsurance.go
```diff
@@ -0,0 +1,156 @@
...
+func BitsuranceGetId(account accounts.Interface) string {
+	for _, signingConf := range account.Info().SigningConfigurations {
+		bitcoinScriptType := signingConf.BitcoinSimple
+		if bitcoinScriptType != nil && bitcoinScriptType.ScriptType == signing.ScriptTypeP2WPKH {
+			h := sha256.New()
+			h.Write([]byte(bitcoinScriptType.KeyInfo.ExtendedPublicKey.String() + xpubSalt))
+			return hex.EncodeToString(h.Sum(nil))
+		}
+	}
+	return ""
```
benma, line 116: Why return "" and not an error? seems safer to return an error and force checking it at the call-site (that check is currently missing there)

backend/handlers/handlers.go
```diff
@@ -226,6 +229,8 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/exchange/pocket/api-url", handlers.getExchangePocketURL).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/pocket/sign-address", handlers.postPocketWidgetSignAddress).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/exchange/pocket/verify-address", handlers.postPocketWidgetVerifyAddress).Methods("POST")
+	getAPIRouterNoError(apiRouter)("/bitsurance/lookup", handlers.getBitsuranceLookup).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/bitsurance/url/{code}", handlers.getBitsuranceURL).Methods("GET")
```
benma, line 233: why is it part of the URL here and a GET param in the lookup endpoint? should ideally be consistent

benma (changes requested):
done with the first commit, haven't checked/tested the 2nd commit yet.

in the first commit, I am missing unit tests - would be great to add unit tests for basically all exported functions. maybe ChatGPT can write them quickly :)

frontends/web/src/locales/en/app.json
```diff
@@ -1450,6 +1455,7 @@
     "addAccount": "Add account",
     "buy": "Buy crypto",
     "device": "Manage device",
+    "insurance": "Insurance",
```
benma, line 1458: same as above in bisruance.title => if it makes sense add it only once so it does not need to be translated into all languages multiple times.

benma (changes requested):

backend/handlers/handlers.go
```diff
@@ -1058,6 +1069,29 @@ func (handlers *Handlers) getExchangesByRegion(r *http.Request) interface{} {
...
+	type errorResult struct {
+		Error string `json:"error"`
+	}
+	insuredAccounts, err := handlers.backend.LookupInsuredAccounts(r.URL.Query().Get("code"))
+	if err != nil {
+		handlers.log.Error(err)
+		return errorResult{Error: err.Error()}
+	}
+
+	return insuredAccounts
```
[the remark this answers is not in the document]
Beerosagos: Did this way because most of the GET endpoint actually just returns data if everything goes well. But yeah, a response struct is much better, will do it like that. Would be nice to refactor the other endpoints as well in future PRs

backend/accounts.go
Beerosagos: Good point. Added a new commit to fix that.

backend/accounts.go
Beerosagos: Now should be better

backend/bitsurance/bitsurance.go
Beerosagos: Probably yes, but we'll probably change this part calling a per-account endpoint instead of downloading the full list. I didn't change it as the whole function will probably be removed of heavily changed in the next PR

backend/bitsurance/bitsurance.go
Beerosagos: nice! thanks

backend/bitsurance/bitsurance.go
Beerosagos: Added error handling

frontends/web/src/locales/en/app.json
Beerosagos: left it in the sidebar (for coherence with other pages) and removed the bitsurance.title one.

frontends/web/src/routes/account/account.tsx
```diff
@@ -58,12 +60,28 @@ export function Account({
...
   const [stateCode, setStateCode] = useState<string>();
   const supportedExchanges = useLoad<SupportedExchanges>(getExchangeBuySupported(code), [code]);
 
+  const account = accounts && accounts.find(acct => acct.code === code);
+
+  const maybeCheckBitsuranceStatus = useCallback(async () => {
+    setInsured(false);
+    if (account?.bitsuranceId) {
+      const insuredAccounts = await bitsuranceLookup(code);
+      if (!insuredAccounts?.includes(code)) {
```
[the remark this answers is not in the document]
Beerosagos: Just felt safer :D removed

frontends/web/src/api/bitsurance.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { apiGet } from '../utils/request';
+import { AccountCode } from './account';
+
+export const getBitsuranceURL = (code: AccountCode): Promise<string> => {
+  return apiGet(`bitsurance/url/${code}`);
+};
+
+export const bitsuranceLookup = (code?: AccountCode): Promise<AccountCode[]> => {
```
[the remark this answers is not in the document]
Beerosagos: fixed

frontends/web/src/routes/account/account.tsx
```diff
@@ -58,12 +60,28 @@ export function Account({
...
   const [transactions, setTransactions] = useState<accountApi.TTransactions>();
   const [usesProxy, setUsesProxy] = useState<boolean>();
+  const [insured, setInsured] = useState<boolean>(false);
   const [stateCode, setStateCode] = useState<string>();
   const supportedExchanges = useLoad<SupportedExchanges>(getExchangeBuySupported(code), [code]);
 
+  const account = accounts && accounts.find(acct => acct.code === code);
+
+  const maybeCheckBitsuranceStatus = useCallback(async () => {
+    setInsured(false);
```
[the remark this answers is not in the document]
Beerosagos: good point, now should be better

benma (approved):
LGTM

PR #2213, opened 2023-07-26
created ReceiverAddressInput component for `send`

shonsirsha:
as a part of the refactoring effort of `send.tsx`, we're creating a new component ReceiverAddressInput as its own component to be used in the send screen.

### DIFF
--- a/frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
+++ b/frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
@@ -1,0 +1,147 @@
+import { ChangeEvent, SyntheticEvent, useContext, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { debug } from '../../../../../utils/env';
+import { getReceiveAddressList } from '../../../../../api/account';
+import DarkModeContext from '../../../../../contexts/DarkmodeContext';
+import { Input } from '../../../../../components/forms';
+import qrcodeIconDark from '../../../../../assets/icons/qrcode-dark.png';
+import qrcodeIconLight from '../../../../../assets/icons/qrcode-light.png';
+import style from '../../send.module.css';
+import { ScanQRDialog } from '../dialogs/scan-qr-dialog';
+import { BrowserQRCodeReader } from '@zxing/library';
+import { alertUser } from '../../../../../components/alert/Alert';
+
+type TToggleScanQRButtonProps = {
+    onClick: () => void;
+}
+
+type TReceiverAddressInputProps = {
+    accountCode?: string;
+    addressError?: string;
+    onClickSendToSelfButton: (e: SyntheticEvent, receiveAddress: string) => void;
+    onInputChange: (value: string) => void;
+    recipientAddress: string;
+    activeScanQR: boolean;
+    parseQRResult: (uri: string) => void;
+    onChangeActiveScanQR: (activeScanQR: boolean) => void
+}
+
+const ScanQRButton = ({ onClick }: TToggleScanQRButtonProps) => {
+  const { isDarkMode } = useContext(DarkModeContext);
+  return (
+    <button onClick={onClick} className={style.qrButton}>
+      <img src={isDarkMode ? qrcodeIconLight : qrcodeIconDark} />
+    </button>);
+};
+
+export const ReceiverAddressInput = ({
+  accountCode,
+  addressError,
+  onInputChange,
+  onClickSendToSelfButton,
+  recipientAddress,
+  activeScanQR,
+  parseQRResult,
+  onChangeActiveScanQR
+}: TReceiverAddressInputProps) => {
+  const { t } = useTranslation();
+  const [hasCamera, setHasCamera] = useState(false);
+  const qrCodeReader = useRef<BrowserQRCodeReader>();
+
+  useEffect(() => {
+    import('../../../../../components/qrcode/qrreader')
+      .then(({ BrowserQRCodeReader }) => {
+        if (!qrCodeReader.current) {
+          qrCodeReader.current = new BrowserQRCodeReader();
+        }
+
+        qrCodeReader.current.getVideoInputDevices()
+          .then(videoInputDevices => {
+            setHasCamera(videoInputDevices.length > 0);
+          });
+      })
+      .catch(console.error);
+
+    return () => {
+      if (qrCodeReader.current) {
+        qrCodeReader.current.reset();
+      }
+    };
+  }, []);
+
+  useEffect(() => {
+    if (activeScanQR) {
+      if (qrCodeReader.current) {
+        qrCodeReader.current.decodeFromInputVideoDevice(undefined, 'video').then(result => {
+          onChangeActiveScanQR(false);
+
+          parseQRResult(result.getText());
+          if (qrCodeReader.current) {
+            qrCodeReader.current.reset(); // release camera
+          }
+        })
+          .catch((error) => {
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -0,0 +1,78 @@
...
+    onClickSendToSelfButton: (e: SyntheticEvent, receiveAddress: string) => void;
+    onInputChange: (e: SyntheticEvent) => void;
+    recipientAddress: string;
+}
+
+const ScanQRButton = ({ onClick }: TToggleScanQRButtonProps) => {
+  const { isDarkMode } = useContext(DarkModeContext);
+  return (
+    <button onClick={onClick} className={style.qrButton}>
+      <img src={isDarkMode ? qrcodeIconLight : qrcodeIconDark} />
```
shonsirsha, line 29: Once we have the svg versions of this icon (light&dark), we can do the usual and import from `icon.tsx` (instead of putting .png into `img` directly like what we're doing here).

frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -0,0 +1,78 @@
...
+import style from '../../send.module.css';
+
+type TToggleScanQRButtonProps = {
+    onClick: () => void;
+}
+
+type TReceiverAddressInputProps = {
+    accountCode?: string;
+    addressError?: string;
+    hasCamera: boolean;
```
thisconnect, line 18: could the hasCamera related code be moved into this component?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -651,27 +643,15 @@ class Send extends Component<Props, State> {
...
-                        </button>
-                      )}
-                    </Input>
+                    <ReceiverAddressInput
+                      accountCode={this.getAccount()?.code}
+                      addressError={addressError}
+                      hasCamera={hasCamera}
+                      onClickScanQRButton={this.toggleScanQR}
+                      onClickSendToSelfButton={this.sendToSelf}
+                      onInputChange={this.handleFormChange}
```
thisconnect, line 652: handleFormChange uses `id` of the event.target to update a send.tsx state. With this change the `id` is now hidden in the child. 

I believe it could set the state directly without using handleFormChange/event.target.id etc. wdyt?

https://github.com/digitalbitbox/bitbox-wallet-app/blob/c4375864155bffc353c11749e116d3adf2a0f9db/frontends/web/src/routes/account/send/send.tsx#L398

thisconnect (changes requested):

frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
shonsirsha: I thought about it, but `hasCamera` depends on the `videoInputDevices.length` which comes from `qrCodeReader.getVideoInputDevices()`. The `this.qrCodeReader` is used in another component (at least for `<ScanQRDialog />`. Idk if that's possible

But, how would you do it?

shonsirsha: Ofc we can pass `qrCodeReader.getVideoInputDevices()` result into a prop of `<ReceiverAddressInput />` directly and evaluate the result inside the `<ReceiverAddressInput />` instead (wheter `hasCamera` or not). Is this what you're thinking?

frontends/web/src/routes/account/send/send.tsx
shonsirsha: If I see the `<ReceiveAddressInput />` correctly and the code block that you highlighted, when a user types in the receive address input (not clicking the send all checkmark) it'll modify the `recipientAddress` state.

>  believe it could set the state directly without using handleFormChange/event.target.id etc. wdyt?

Not 100% sure I understood this one. Did you mean we can make a new method in the parent that does `
```js
this.setState({recipientAddress},() => { 
   this.validateAndDisplayFee(true); 
 }));
``` 

in the parent (`send.tsx`) instead of using the general  handleFormChange? Yeah sure we can do that, but why if I may know? It's essentially the same thing.. maybe easier to read I guess (instead of guessing target.id we know it's `recipientAddress`)?

frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
thisconnect: sorry I didn't look closely, but it feels like ScanQRDialog could all be move  into this component.

frontends/web/src/routes/account/send/send.tsx
thisconnect: yes, I think explicitly setting the state with
```js
this.setState({recipientAddress},() => { 
   this.validateAndDisplayFee(true); 
 }));
 ```
is much better than using `handleFormChange` and depending on DOM element id. Especially now that id is buried in `<ReceiverAddressInput` is more difficult to read and understand. I think in the long term we should try to remove `handleFormChange`

File: frontends/web/src/routes/account/send/send.tsx [hunk 4]
```diff
@@ -221,7 +197,7 @@ class Send extends Component<Props, State> {
   };
 
   private handleKeyDown = (e: KeyboardEvent) => {
-    if (e.keyCode === 27 && !this.state.activeCoinControl && !this.state.activeScanQR) {
+    if (e.keyCode === 27 && !this.state.activeCoinControl) {
```
shonsirsha, line 200: I tried combining ScanQRDialog and ReceiverAddressInput @thisconnect . Unfortunately this isn't 100% working "smoothly". See this bit of code that I'm highlighting.

We somehow still need the parent (send page) to know if Dialog is opened, so the firing of "escape" can work correctly in the send screen; which is "escape" should close qrdialog when qrdialog is opened, otherwise it closes the whole page. The current (broken) behaviour will close the whole page regardless of the dialog's condition.

To fix this, I'll put back `activeScanQR` in the parent for now and pass that state and its setter to `ReceiverAddressInput` - please lmk what you think though. I don't think this is a very clean fix... but a simple enough change for now.

This might be an indication that we may need to refactor/rethink the way our Dialog works.

shonsirsha: Another hacky fix is we can try to check if the actual dom is there e.g `document.querySelector()` - but I'd classify this as a hack.

shonsirsha:
feel free to review too! @Beerosagos :) 🙏 thanks

Beerosagos (approved):
tACK, just a possible nit on a comment. Very nice!

frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -0,0 +1,149 @@
...
+    } catch (e) {
+      console.error(e);
+    }
+  };
+
+
+  const toggleScanQR = () => {
+    if (activeScanQR) {
+      if (qrCodeReader.current) {
+        // release camera; invokes the catch function below.
```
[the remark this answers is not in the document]
shonsirsha: Ups, yeah. Sorry I missed this comment! The catch is here now: https://github.com/digitalbitbox/bitbox-wallet-app/blob/146703d76be587c90e93ca2980ff8a329523c403/frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx#L83-L88. I'll update the wording of the comment in the code :). Thanks for noticing!

PR #2214, opened 2023-07-26
Add NGN and ZAR fiat currencies

Zero-1729:
This PR adds support for the Nigerian Naira (NGN) and South African Rand (ZAR).

CoinGecko supports both currencies, see https://api.coingecko.com/api/v3/simple/supported_vs_currencies.


benma:
Thanks for your contribution! Unfortunately we cannot merge this for the time being as we first need to make optimizations to our currency conversion backend, which is feeling the load of all the currency pairs currently supported!

It's not clear when we will get to this either, as it's quite low prio at the moment.

PR #2216, opened 2023-08-07
backend: fix aopp test after updating moq

benma:
Regenerating the keystore mock using an updated moq newly adds locks to the struct, so the linter complained with:

```
ackend/aopp_test.go:386:10: copylocks: assignment copies lock value to ks2: github.com/digitalbitbox/bitbox-wallet-app/backend/keystore/mocks.KeystoreMock contains sync.RWMutex (govet)
                ks2 := ks
                       ^
backend/aopp_test.go:422:10: copylocks: assignment copies lock value to ks2: github.com/digitalbitbox/bitbox-wallet-app/backend/keystore/mocks.KeystoreMock contains sync.RWMutex (govet)
                ks2 := ks
                       ^
```

This commit updates the mock file and fixes the test to avoid the linting error.


Beerosagos (approved):
tACK

PR #2219, opened 2023-08-08
frontend: using `word-break: break-word` for sidebarLabel

shonsirsha:
To break word (account name) appropriately, we need to use the `break-word` value for `word-break` css property for `.sidebarLabel`.

Previously, the sidebarLabel is able to break world "appropriately" when account name is either short or long **with** spaces. **Long names without spaces will cause the coin logo to shift to the left, which isn't ideal.**

Before:
Screenshot 2023-08-08 at 12 48 12

After:
Screenshot 2023-08-08 at 12 48 18
Screenshot 2023-08-08 at 12 53 40

### DIFF
--- a/frontends/web/src/components/sidebar/sidebar.module.css
+++ b/frontends/web/src/components/sidebar/sidebar.module.css
@@ -172,9 +172,11 @@ a.sidebarActive .single img,
     line-height: 1;
     flex: 1;
     padding-top: 0;
+    padding-right: var(--space-default);
     font-size: var(--size-default);
     font-weight: 400;
     transition: all 0.2s ease;
+    word-break: break-word;
 }
 
 .activeGroup {


Beerosagos:
Nice! That seems to work fine, but I'd suggest to add a bit of padding to the sidebar to avoid having the words breaking up at the border, wdyt?

immagine

shonsirsha:
@Beerosagos Thanks! I agree, added a bit of padding (space-default). Lmk what you think :) 

Screenshot 2023-08-08 at 15 01 22
Screenshot 2023-08-08 at 15 01 11

Beerosagos:
Thanks @shonsirsha! The padding looks nice, but now my *very* long account names are overflowing from top and bottom. But maybe this is fine as they are really longer then one could expect.. wdyt?

immagine

shonsirsha:
@Beerosagos , Yeah.. I think the reason for this is because we have a fixed height applied: `(--sidebar-item-height)` - so this would've happened even without the padding adjustment (just try to make the name even longer and it'll do this too). 

I'd say the current changes are already an improvement and this "fixed height" is a separate issue, plus as you said it's such an edge case anyway as the name is super long at that point... So idk if we should change the fixed height here to accommodate this (edge) case.

PR #2221, opened 2023-08-14
frontend: created CoinInput for send screen

shonsirsha:
as a part of the refactoring effort of `send.tsx`, we're creating a new component CoinInput as its own component to be used in the send screen.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import React, { Component } from 'react';
+import { Component } from 'react';
 import * as accountApi from '../../../api/account';
 import { syncdone } from '../../../api/accountsync';
 import { BtcUnit, parseExternalBtcAmount } from '../../../api/coins';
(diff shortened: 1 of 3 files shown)


thisconnect (approved):
untested LGTM with suggestion

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -358,27 +359,6 @@ class Send extends Component<Props, State> {
...
-      this.convertToFiat(value);
-    }
-    this.setState(prevState => ({
-      ...prevState,
-      [target.id]: value,
-    }), () => {
-      this.validateAndDisplayFee(true);
-    });
-  };
-
```
shonsirsha, line 381: @thisconnect In regards to your comment, the whole function can actually be removed totally. This was previously used in `sendToSelf`, but it serves no purpose as the `value` when it gets called from `sendToSelf` will be undefined (it's a label).

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -418,9 +398,10 @@ class Send extends Component<Props, State> {
     }
   };
 
-  private sendToSelf = (event: React.SyntheticEvent, recipientAddress: string) => {
-    this.setState({ recipientAddress });
-    this.handleFormChange(event);
+  private sendToSelf = (recipientAddress: string) => {
+    this.setState({ recipientAddress }, () => {
+      this.validateAndDisplayFee(true);
+    });
```
shonsirsha, line 404: `sendToSelf` only needs to set `recipientAddress` to be our own address like this. We previously called `handleFormChange` but I believe there's no point in doing so.

shonsirsha:
Rebased and made a small change to remove unused fn, PTAL 🙏

frontends/web/src/routes/account/send/send.tsx
thisconnect: very nice, 👍 

actually, the sendToSelf function is now the same as onReceiverAddressInputChange, I think you can remove sendToSelf and use onReceiverAddressInputChange.

thisconnect (changes requested):

shonsirsha:
Thanks! good catch. done! PTAL @thisconnect

thisconnect (approved):
LGTM :thumbsup:

PR #2223, opened 2023-08-16
Merge master into staging-walletconnect

shonsirsha:
Merging `master` into `staging-walletconnect` so the staging branch has the most up-to-date version of master.

Also, some components like `<ScanQRDialog />` that exists in the current master will be used in `staging-walletconnect` (it doesnt exist in the staging branch yet).


Beerosagos (approved):
ntACK!

PR #2227, opened 2023-08-21

Beerosagos:
```
go get github.com/digitalbitbox/bitbox02-api-go@f9cc9858
go mod tidy
go mod vendor
```

EDIT: This update brings a `btcd` version bump that breaks the tx hash backward compatibility (see https://github.com/btcsuite/btcd/pull/2025). I added a second commit to update `btcd/chaincfg/chainhash` dep and include my fix commit. 

```
go get github.com/btcsuite/btcd/chaincfg/chainhash@f12a0b3
go mod tidy
go mod vendor
```


Beerosagos:
ping @benma

benma (changes requested):
The commit msg says

     go get github.com/digitalbitbox/bitbox02-api-go@f9cc9858

That should be `73e3f60`, the current master - which one did you actually run? 

Also, +30k lines added, wow... I guess not much we can do about that.

Beerosagos:
> That should be `73e3f60`, the current master - which one did you actually run?

I'm pretty sure I did @f9cc9858 as in the commit message, which was the current master when I started the PR. The result should be the same, exception made for the 'eip712-uint' commit that would not be included until the next update. Can that wait or do you prefer to update to the newest commit?

> Also, +30k lines added, wow... I guess not much we can do about that.

Yeah, I'm afraid so

thisconnect:
> Also, +30k lines added, wow... I guess not much we can do about that.

one of the vendors worked a lot?

benma (approved):
> I'm pretty sure I did @f9cc9858 as in the commit message, which was the current master when I started the PR. The result should be the same, exception made for the 'eip712-uint' commit that would not be included until the next update. Can that wait or do you prefer to update to the newest commit?

It also contains https://github.com/digitalbitbox/bitbox02-api-go/commit/06483fc190d1473f419621b2a0061990fe16e5e7. I guess it does not matter since you also manually updated it here, so LGTM.

It would be good to add a unit test with a fixture for the tx deserialize/serialize issue that we almost missed.

Beerosagos:
> It would be good to add a unit test with a fixture for the tx deserialize/serialize issue that we almost missed.

Good point! I'll note that and make a new PR

PR #2228, opened 2023-08-22
frontend: changed qrcode icon in send screen

shonsirsha:
Using svg qr code icons to improve the quality of the icon.

[Preview] Light mode:
Screenshot 2023-08-22 at 10 30 28

[Preview] Dark mode:
Screenshot 2023-08-22 at 10 30 17


thisconnect:
nice, btw. I noticed that the clickable area is a bit small, do you see a straight-forward way to change the CSS so that the clickable area includes the surrounding space?

shonsirsha:
>  nice, btw. I noticed that the clickable area is a bit small, do you see a straight-forward way to change the CSS so that the clickable area includes the surrounding space?

Hm, i tried 2x-ing the button size (not the icon) and adjusted the positioning.  Lmk what you think! :) 

p.s: we can also make it 3x and fit the whole height of the input, but maybe it's too big.

thisconnect (approved):
LGTM 👍

PR #2229, opened 2023-08-23
frontend: increase displayedValue setting item left padding

shonsirsha:
to prevent displayedValue text and the setting description text collapsing with one another, we adjusted the left padding of displayedValue text.

Before:
Screenshot 2023-08-23 at 13 53 34

After:
Screenshot 2023-08-23 at 13 53 45


shonsirsha:
please lmk if we want the space to be even bigger. :)

thisconnect (approved):
LGTM

PR #2230, opened 2023-08-25
backend: add wallet connect endpoint get account code from address

Tomasvrba:
Adds an endpoint to get an Ethereum account code and name based on address:

request: { address: ethHexAddress }
response: { success: bool, code: accountCode, name: accountName }

this is so that wallet connect requests can be listened to and acted on from anywhere in the app, not just from the corresponding account page


benma (changes requested):

Tomasvrba:
Should be all fixed @benma thanks for the review
will fixup and rebase on https://github.com/digitalbitbox/bitbox-wallet-app/pull/2231 once merged and approved

Tomasvrba:
> > Should be all fixed @benma thanks for the review will squash and rebase on #2231 once merged and approved
> 
> merged, pls rebase. i see at least one missing docstring where the linter will complain :)

done
🤦‍♂️ forgot to add it to the new function in the fixup :D

backend/coins/eth/account.go
```diff
@@ -961,3 +966,16 @@ func (account *Account) EthSignWalletConnectTx(
 	}
 	return "0x" + hex.EncodeToString(txHash[:]), "0x" + hex.EncodeToString(rawTx), nil
 }
+
+// Returns the account's single Ethereum address.
+func (account *Account) Address() *Address {
+	if !account.isInitialized() {
+		return &Address{}
```
benma, line 973: i know the only current use works out with this, but it's more robust and standard to return an error. potentially future use cases could fail silently or in hard to debug ways by getting a "valid" but wrong address.

benma (changes requested):

backend/coins/eth/account.go
Tomasvrba: yeah, that was mostly laziness on my part since I didn't want to rewrite
`if ethcommon.HexToAddress(address).Hex() == ethAccount.Address().Hex() && !eth.IsERC20(ethAccount)` to get the address first and check for errors :D

Tomasvrba: It's the same in: 
```
func (account *Account) GetUnusedReceiveAddresses() []accounts.AddressList {
	if !account.isInitialized() {
		return nil
	}
	return []accounts.AddressList{{
		Addresses: []accounts.Address{account.address},
	}}
}
``` :P

benma: I thought that was the reason ;)

GetUnusedReceiveAddresses should be converted to you are right :sweat_smile:

PR #2231, opened 2023-08-25
golangci-lint: enable check that exported items have comments

benma:
This used to be included in golint or go vet, but was moved to the `revive` linter.

https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#exported

It is good practice to document all exported items.


benma:
Merged to unblock https://github.com/digitalbitbox/bitbox-wallet-app/pull/2230

PR #2232, opened 2023-08-27
merge master to staging-walletconnect

shonsirsha:
There's a new hook for QR code scanner  in `master` that we can re-use in the `staging-walletconnect`  for the QR reading feature of the WC address. Merging master mainly so we can use that hook (and to keep other stuff updated, ofc)  🙂


thisconnect:
@shonsirsha was there any conflicts that I should take a closer look at?

thisconnect (approved):
LGTM with question

shonsirsha:
>  @shonsirsha was there any conflicts that I should take a closer look at?

Thanks for reviewing! And nope @thisconnect , all good. 👍

PR #2235, opened 2023-08-29
wallet-connect: add util function to decode eth messages in frontend

Tomasvrba:
Also displays the decoded message in the current POC frontend modal

### DIFF
--- a/frontends/web/src/routes/account/walletconnect/index.tsx
+++ b/frontends/web/src/routes/account/walletconnect/index.tsx
@@ -25,9 +25,8 @@ import { Button, Input } from '../../../components/forms';
 import { PairingTypes, SessionTypes, SignClientTypes } from '@walletconnect/types';
 import { buildApprovedNamespaces, getSdkError } from '@walletconnect/utils';
 import { Dialog, DialogButtons } from '../../../components/dialog/dialog';
-import useInitialization, { SUPPORTED_CHAINS, pair, web3wallet } from './utils';
+import useInitialization, { SUPPORTED_CHAINS, EIP155_SIGNING_METHODS, decodeEthMessage, pair, web3wallet } from './utils';
 import React from 'react';
-import { EIP155_SIGNING_METHODS } from './utils';
 import { alertUser } from '../../../components/alert/Alert';
 import { ethSignMessage, ethSignTypedMessage, ethSignWalletConnectTx } from '../../../api/account';
 
@@ -151,6 +150,7 @@ export const WalletConnect = ({
       const { topic, params, id } = requestEvent;
       const { request } = params;
       let message: string;
+      let decodedMessage;
       const activeSessions = Object.values(web3wallet?.getActiveSessions());
       const currentSession = activeSessions.find((session) => session.topic === topic);
 
@@ -161,6 +161,7 @@ export const WalletConnect = ({
         * while PERSONAL_SIGN gives them as [message, address]
         */
         message = request.params[1];
+        decodedMessage = decodeEthMessage(message);
         if (currentSession) {
           handleSessionRequest(async () => {
             const result = await ethSignMessage(code, message);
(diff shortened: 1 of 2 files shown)


shonsirsha:
> I added the api call to the frontend as well now ...
@Tomasvrba thanks, will be handy :)

frontends/web/src/api/backend.ts
```diff
@@ -31,6 +31,8 @@ export interface ISuccess {
     errorCode?: string;
 }
 
+export type TDecodedMessage = {success: boolean, decodedMessage: string, errorMessage: string}
```
benma, line 34: again here, why not `{ success: true; decodedMessage: string } | { success: false; errorMessage: string }`. typing more explicitly prevents one from accessing the wrong fields.

frontends/web/src/routes/account/walletconnect/index.tsx
```diff
@@ -161,6 +172,8 @@ export const WalletConnect = ({
         * while PERSONAL_SIGN gives them as [message, address]
         */
         message = request.params[1];
+        decodedMessage = await decodeMessage(message);
+        console.log(decodeMessage);
```
benma, line 176: remove log

benma (changes requested):

backend/handlers/handlers.go
```diff
@@ -1235,3 +1238,20 @@ func (handlers *Handlers) postAOPPApproveHandler(r *http.Request) interface{} {
 	handlers.backend.AOPPApprove()
 	return nil
 }
+
+func (handlers *Handlers) postEthDecodeMsg(r *http.Request) interface{} {
+	var hexMessage string
+	type response struct {
+		Success      bool   `json:"success"`
+		Message      string `json:"decodedMessage"`
```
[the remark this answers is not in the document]
Tomasvrba: I knew you wouldn't like that :D

thisconnect:
@Tomasvrba there seems to be a conflict in `backend/handlers/handlers.go` please rebase

frontends/web/src/api/backend.ts
```diff
@@ -70,3 +72,7 @@ export const getDefaultConfig = (): Promise<any> => {
 export const socksProxyCheck = (proxyAddress: string): Promise<ISuccess> => {
   return apiPost('socksproxy/check', proxyAddress);
 };
+
+export const ethDecodeMessage = (message: string): Promise<TDecodedMessage> => {
```
benma, line 76: sorry for the late suggestion, but why not simply make this a pure JS function? Untested from ChatGPT:

```
function ethDecodeHex(message: string): string {
  let str = '';
  for (let i = 0; i < message.length; i += 2) {
    str += String.fromCharCode(parseInt(message.substr(i, 2), 16));
  }
  return decodeURIComponent(escape(str));
}
```

would that work? if so, it seems a lot simpler compared to this current PR

Tomasvrba: I'll test out how it works with JS and I can change it. 
The reason for doing it in the backend was that you mentioned it when we first talked about it and the idea was that it would be easier/more robust just to do a simple api call and do the decoding in go

benma: I meant that versus pulling in more NPM deps, as we want to keep the NPM deps to a minimum. If the functionality is so simple like above I see no problem in adding the code to the frontend instead.

Tomasvrba: escape() is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape

Did it slightly differently (with GPTs help :D) and added some validation and error checking, all in the frontend now

benma:
Seems like this needs a rebase for some reason - otherwise utACK :)

We should also test this in Qt and Android just in case the WebViews there do not support something, but should be ok,

PR #2236, opened 2023-08-29
frontend: created Dashboard screen for wallet connect

shonsirsha:
The dashboard screen contains all connected WC sessions of all accounts in the bitbox (to dApps), as well as a button to go to the connect screen.

It also contains some general information of the currently opened account such as the account name and its receive/wallet address.

Preview:
Screenshot 2023-09-02 at 12 58 05
Screenshot 2023-09-02 at 12 58 10

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -16,6 +16,7 @@
 
 import { apiGet, apiPost } from '../utils/request';
 import { ChartData } from '../routes/account/summary/chart';
+import { SuccessResponse } from './response';
 
 export type CoinCode = 'btc' | 'tbtc' | 'ltc' | 'tltc' | 'eth' | 'goeth';
 
@@ -63,6 +64,15 @@ export interface ITotalBalance {
 
 export const getAccountsTotalBalance = (): Promise<ITotalBalance> => {
   return apiGet('accounts/total-balance');
+};
+
+type TETHAddressByAccountCode = SuccessResponse & {
+  code: AccountCode;
+  name: string;
+}
+
+export const getETHAddressByAccountCode = (address: string): Promise<TETHAddressByAccountCode> => {
+  return apiPost('accounts/eth-account-code', { address });
 };
 
 export interface IStatus {
@@ -349,3 +359,4 @@ export const ethSignTypedMessage = (code: AccountCode, chainId: number, data: an
 export const ethSignWalletConnectTx = (code: AccountCode, send: boolean, chainId: number, tx: any): Promise<TSignWalletConnectTx> => {
   return apiPost(`account/${code}/eth-sign-wallet-connect-tx`, { send, chainId, tx });
 };
+
--- a/frontends/web/src/routes/account/walletconnect/components/session-card/session-card.module.css
+++ b/frontends/web/src/routes/account/walletconnect/components/session-card/session-card.module.css
@@ -1,0 +1,100 @@
+.container {
+    background-color: var(--background-session-card);
+    display: flex;
+    width: 100%;
+    padding: calc(var(--space-half) + var(--space-quarter));
+}
+
+.container p {
+    color: var(--color-default);
+    font-size: 16px;
+    line-height: 1.9;
+    margin: 0;
+}
+
+.textDataContainer {
+    display: flex;
+    flex-direction: column;
+    gap: calc(var(--space-quarter) + var(--space-half));
+}
+
+.receiveAddress {
+    color: var(--color-secondary);
+}
+
+.accountName {
+    font-weight: bold;
+    width: 240px;
+    padding-right: 16px;
+}
+
+.dAppMetadataAndIconContainer {
+    align-items: center;
+    display: flex;
+}
+
+.dAppMetadataAndIconContainer img {
+    width: var(--item-height);
+    height: var(--item-height);
+    margin-left: calc(var(--space-quarter) + var(--space-half));
+}
+
+.dAppNameAndUrlContainer {
+    display: flex;
+    flex-direction: column;
+    gap: calc(var(--space-eight) * 2);
+    width: 240px;
+}
+
+.accountNameAndWalletContainer {
+    display: flex;
+}
+
+.buttonAndIconContainer {
+    align-items: flex-end;
+    display: flex;
+    margin-left: auto;
+}
+
+.buttonDisconnect {
+    height: unset;
+    padding: calc(var(--space-eight) + var(--space-quarter)) calc(var(--space-default) + var(--space-default));
+}
+
+@media (max-width: 768px) { 
+    .container {
+        padding: calc(var(--space-quarter) * 2);
+    }
+
+    .container p { 
+        font-size: var(--size-default);
+    }
+
+    .textDataContainer {
+        gap: var(--space-quarter);
+        width: 50%;
+    }
+
+    .dAppNameAndUrlContainer,
+    .accountName {
+        width: 100%;
+    }
+
+    .buttonAndIconContainer {
+        flex-direction: column;
+        justify-content: flex-end;
+    }
+
+    .buttonDisconnect {
+        font-size: var(--size-default);
+        height: var(--item-height-small);
+        padding: var(--space-eight) calc(var(--space-quarter) + var(--space-half));
+    }
+
+    .buttonAndIconContainer img {
+        height: var(--item-height-small);
+        width: var(--item-height-small);
+        margin-top: var(--space-default);
+        margin-bottom: var(--space-default);
+    }
+}
--- a/frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
+++ b/frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2023 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CoreTypes } from '@walletconnect/types';
+import { Button } from '../../../../../components/forms';
+import { useTranslation } from 'react-i18next';
+import { useLoad } from '../../../../../hooks/api';
+import { getETHAddressByAccountCode } from '../../../../../api/account';
+import { truncateAddress } from '../../utils';
+import styles from './session-card.module.css';
+
+type TTextDataProps = {
+    accountName: string;
+    receiveAddress: string;
+    dAppName: string;
+    dAppUrl: string;
+    iconUrl?: string;
+}
+
+type TWCSessionCardProps = {
+  metadata: CoreTypes.Metadata;
+  onDisconnect: () => void;
+  receiveAddress: string;
+}
+
+const TextData = ({ accountName, receiveAddress, dAppName, dAppUrl, iconUrl }: TTextDataProps) => {
+  return <div className={styles.textDataContainer}>
+
+    <div className={styles.accountNameAndWalletContainer}>
+      <p className={styles.accountName}>{accountName}</p>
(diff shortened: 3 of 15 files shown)


frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+
+export const WCSessionCard = ({ metadata, receiveAddress, onDisconnect }: TWCSessionCardProps) => {
+  const { t } = useTranslation();
+  const { name, url, icons } = metadata;
+  return (
+    <div className={styles.container}>
+      <TextData
+        accountName="Account name here (TODO)"
```
shonsirsha, line 64: Todo after #2235  gets merged (to get the account name by receive address).

shonsirsha:
Ready for review again :) screenshots updated 

Some things to be tested/QAed:

1. Verify that wallet connect CTA in the account page routes us to the newly created dashboard page `/wallet-connect/dashboard`
2. Verify that clicking on the `New connection` btn routes us to the connect page `/wallet-connect/connect`
3. Verify that in the connect page, clicking on the `Cancel` btn brings us back to the dashboard page
4. Verify that in the connect page, after clicking "Approve" when connecting to a dApp, both "Reject" and "Approve" btn are in a loading state which is in **disabled state** (so user can't "spam click" during this 2-4 seconds loading time).
5. Verify that clicking `Done` btn after a successful pairing also brings us back to the dashboard page
6. Verify that all sessions are shown in the dashboard page.
7. Verify that both disconnecting via dApp and by clicking the `Remove` btn in each session works (actually removes the session from the page).
8. Verify that the sessions displayed in the dashboad page show the correct information (correct address, account name, and the dApp metadata info).

cc @jadzeidan in case you want to test 😄

frontends/web/src/api/account.ts
```diff
@@ -65,6 +65,16 @@ export const getAccountsTotalBalance = (): Promise<ITotalBalance> => {
   return apiGet('accounts/total-balance');
 };
 
+type TETHAddressByAccountCode = {
+  success: boolean;
+  code: AccountCode;
+  name: string;
+}
```
thisconnect, line 72: I assume you only get code and name if success is true, so you could hardcode to success: true or remove and do `type TETHAddressByAccountCode =  SuccessResponse & {`, this way TypeScript can error if success is false and you assume there is code or name.

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.module.css
```diff
@@ -0,0 +1,99 @@
...
+        height: var(--item-height-small);
+        padding: var(--space-eight) calc(var(--space-quarter) + var(--space-half));
+    }
+
+    .container .buttonAndIconContainer img {
+        height: var(--item-height-small);
+        width: var(--item-height-small);
+        margin-top: var(--space-default);
+        margin-bottom: var(--space-default);
+    }
```
thisconnect, line 98: Is the additional CSS specifity needed or could you remove all `.container` classes except for the container styles itself?

i.e. `.container .buttonAndIconContainer button` -> `.buttonAndIconContainer button` ?

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <div className={styles.container}>
+      <TextData
+        accountName={accountName}
+        receiveAddress={truncatedAddress}
+        dAppName={name}
+        dAppUrl={url}
+        iconUrl={icons[0]}
+      />
```
thisconnect, line 76: These `icons[0]` still confuse me a bit, could you check if there is ever more than 1 icon?

also is that absolute URL pointing to random urls? We should check if this image can load in Qt app, as I think we might block arbitrary domains.

If they all come from the same domain we could white list the domain. If they come from random domains we should look into other solutions. I just make a note here so we don't forget, doesn't have to be in this PR, but would be nice to find out.

frontends/web/src/routes/account/walletconnect/dashboard.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </div>
+            <Button onClick={() => route(`/account/${code}/wallet-connect/connect`)} primary>{t('walletConnect.dashboard.newConnection')}</Button>
+          </div>
+          <hr className={styles.separator} />
+          {hasSession &&
+            <div className={styles.sessionCardsContainer}>
+              <p>{t('walletConnect.dashboard.allSessions')}</p>
+              {sessions.map(s =>
+                <WCSessionCard
+                  key={s.topic}
```
thisconnect, line 97: could you change `s` to `session`, and is `session.topic` unique?

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.module.css
thisconnect: actually in the example of button I think it would be better to just have a class to pass to `<Button className={style.buttonDisconnect}` so that the CSS doesn't depend on the child components DOM.

i.e. in theory if we ever change Button component internally depending on `.buttonAndIconContainer button` could break at some point and it would be hard to notice.

frontends/web/src/routes/account/walletconnect/dashboard.tsx
```diff
@@ -0,0 +1,113 @@
...
+      web3wallet?.on('session_delete', updateSessions);
+      return () => {
+        web3wallet?.off('session_delete', updateSessions);
+      };
+    }
+  }, [isInitialized]);
+
+  const updateSessions = () => {
+    setSessions(Object.values(web3wallet?.getActiveSessions()));
+  };
```
thisconnect, line 59: could you move `updateSessions` to the top, so it's defined before useEffect where it is used for the first technically it prob doesn't matter but mentally it helps me read the code.

frontends/web/src/routes/account/walletconnect/utils.ts
```diff
@@ -118,4 +118,16 @@ export async function pair(params: { uri: string }) {
     throw new Error('Please use a new URI!');
   }
   await web3wallet?.core.pairing.pair({ uri });
-}
\ No newline at end of file
+}
+
+export const getAddressFromEIPString = (address: string) => {
+  const parts = address.split(':');
+  return parts.length > 2 ? parts[2] : '';
```
thisconnect, line 125: do those EIP strings always have a colon `:`?

thisconnect (changes requested):

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.module.css
shonsirsha: Thanks for these comments. I was unsure how we wanna tackle this. I was trying to reduce the usage of classnames and rather target it semantically. 

But your comment regarding the breaking when we restructure divs is valid - I guess this is a  "give or take" situation. I'll reduce the complexity by just resorting to classnames as you suggested.

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
shonsirsha: Well, we can only know the data structure because the icons are coming from the dApp (probably theres even a possibility, albeit small, that theres no icon??  but anyway i expect "production" dApps to have it).

Looking at the [documentation](https://docs.walletconnect.com/2.0/specs/clients/core/pairing/data-structures), it **will always be an array**. So it's safe to say we can try to get `icons[0]` at all times. To be extra careful we can even have a placeholder icon and use it if `icons[0]` is undef. Feel free to suggest! :)

>also is that absolute URL pointing to random urls?

Good question. Yes they're pointing to different origins. Regarding our blocking/whitelist I'm unsure how we can tackle this.. 🤔

frontends/web/src/routes/account/walletconnect/dashboard.tsx
shonsirsha: sure! and yes topic is unique.

frontends/web/src/routes/account/walletconnect/utils.ts
shonsirsha: In this case, yes I assume so. Assumption comes from here (we "write" it this way):

https://github.com/digitalbitbox/bitbox-wallet-app/blob/87c1987c8247aee791e5a1f731e762ac206aed3e/frontends/web/src/routes/account/walletconnect/components/incoming-pairing/incoming-pairing.tsx#L61 , or in Tom's PoC:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/87c1987c8247aee791e5a1f731e762ac206aed3e/frontends/web/src/routes/account/walletconnect/index.tsx#L269
(chain is for example `eip155:1`)


Just to confirm, @Tomasvrba, could you verify that the address is indeed always formatted the same; which is after the second colon (e.g: `eip155:1:walletAddress`), please?  thanks. 🙏 

_note: can also use regex instead if we prefer that._

shonsirsha:
Thanks for reviewing! @thisconnect. Changes made, PTAL 🙏

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { getETHAddressByAccountCode } from '../../../../../api/account';
+import { truncateAddress } from '../../utils';
+import styles from './session-card.module.css';
+
+type TTextDataProps = {
+    accountName: string;
+    receiveAddress: string;
+    dAppName: string;
+    dAppUrl: string;
+    iconUrl: string;
```
thisconnect, line 30: `iconUrl: string | undefined;` icons[0] could be undefined

thisconnect (approved):
untested LGTM 👍  with last nit, cc @jadzeidan  for testing https://github.com/digitalbitbox/bitbox-wallet-app/pull/2236#issuecomment-1698949829

thisconnect:
would it be possible to give the url a bit more space?

Screen Shot 2023-08-30 at 19 45 23

Screen Shot 2023-08-30 at 19 46 32

frontends/web/src/routes/account/walletconnect/utils.ts
Tomasvrba: That's correct, see: https://docs.walletconnect.com/2.0/specs/clients/sign/namespaces#25-session-namespaces-may-contain-multiple-accounts-for-one-chain

frontends/web/src/routes/account/walletconnect/components/session-card/session-card.tsx
Tomasvrba: > These `icons[0]` still confuse me a bit, could you check if there is ever more than 1 icon?
> 
> also is that absolute URL pointing to random urls? We should check if this image can load in Qt app, as I think we might block arbitrary domains.
> 
> If they all come from the same domain we could white list the domain. If they come from random domains we should look into other solutions. I just make a note here so we don't forget, doesn't have to be in this PR, but would be nice to find out.

The URLs can be anything since they are afaik hosted on the Dapp's domain. I believe the last time we discussed this, we said if there was an issue with Qt, we could blanket whitelist all requests coming from the /wallet-connect route in the BBApp, or whatever the final route will end up where we fetch the icons

Tomasvrba:
Looking great :)

frontends/web/src/routes/account/walletconnect/utils.ts
thisconnect: > can also use regex instead if we prefer that.

I think it is fine, no need for regex.

shonsirsha:
> would it be possible to give the url a bit more space?


Thanks! Updated some text spacings (see updated preview above) and the `iconUrl` typing :)

thisconnect:
tested https://github.com/digitalbitbox/bitbox-wallet-app/pull/2236#issuecomment-1698949829 a bit, LGTM 👍

jadzeidan:
The account names aren't showing up for me, is this a known issue?

image

thisconnect:
I think we only know ETH address and need  https://github.com/digitalbitbox/bitbox-wallet-app/pull/2235 for getting account names  (and account code) 

> The account names aren't showing up for me, is this a known issue?

shonsirsha:
> The account names aren't showing up for me, is this a known issue?

Hm, should show up. The PR is actually #2230 , not #2235 - sorry my comments above were a typo. cc @thisconnect

thisconnect:
I just assumed 😇  lol they are there on my screenshots

>  do the names not show up in yours too?

thisconnect:
LGTM, @shonsirsha please merge.

PR #2237, opened 2023-08-29
frontend: type notify user

thisconnect:
Converting api post call to trigger native os notification.


shonsirsha (approved):
untested LGTM. Thanks 🚀 👍

PR #2238, opened 2023-08-30
backend/bitsurance: switch insurance status endpoint

Beerosagos:
Previously we fetched a list of all the insured accounts using the `allAccountHashes` endpoint. This solution is not much scalable and forces Bitsurance to expose a list of all the active accounts. This commit exploits the `accountDetails` endpoint instead, and makes the app do a single endpoint call for each account to be verified.


backend/bitsurance/bitsurance.go
```diff
@@ -35,70 +33,38 @@ const (
...
-		default:
-			// If the element in insuredIds is smaller, move to the next element
-			j++
-		}
+// If an accountId is not retrieved at all, the endpoint return a 404 http code.
+func bitsuranceCheckId(httpClient *http.Client, accountId string) (bool, error) {
+	endpoint := apiURL + "accountDetails/" + accountId
+	details := accountDetails{}
+	code, err := util.APIGet(httpClient, endpoint, apiKey, 1024, &details)
+	if *code == http.StatusNotFound {
```
benma, line 57: do this check after `err != nil` check or this can panic

backend/bitsurance/bitsurance_test.go
```diff
@@ -0,0 +1,48 @@
+package bitsurance
```
benma, line 1: license header missing

backend/util/util.go
```diff
@@ -23,33 +23,41 @@ import (
...
-func APIGet(httpClient *http.Client, endpoint string, result interface{}) error {
+// APIGet performs a HTTP Get call to a given endpoint and unmarshal the result populating a given object.
+// Input params:
+// - `httpClient` which is used to perform the call
+// - `endpoint` the url for the endpoint to fetch
+// - `apikey` if not empty it is used as X-API-KEY header value
+// - `maxSize` indicates the max expected response size. If it exceeds the function returns an error
+// - `result` object that should be used to unmarshal the response body
+// Returns a pointer to the error code (if available) and possibly an error.
+func APIGet(httpClient *http.Client, endpoint string, apiKey string, maxSize int64, result interface{}) (*int, error) {
```
benma, line 34: I think it's more natural/idiomatic to return `int` not `*int` and just return `0` instead of `nil`.

benma (changes requested):

backend/bitsurance/bitsurance.go
Beerosagos: nice catch! Thanks

backend/util/util.go
Beerosagos: noted, thanks

Beerosagos:
PR updated, @benma PTAL :pray:

backend/util/util.go
```diff
@@ -23,33 +23,41 @@ import (
...
-// populating a given object.
-func APIGet(httpClient *http.Client, endpoint string, result interface{}) error {
+// APIGet performs a HTTP Get call to a given endpoint and unmarshal the result populating a given object.
+// Input params:
+// - `httpClient` which is used to perform the call
+// - `endpoint` the url for the endpoint to fetch
+// - `apikey` if not empty it is used as X-API-KEY header value
+// - `maxSize` indicates the max expected response size. If it exceeds the function returns an error
+// - `result` object that should be used to unmarshal the response body
+// Returns a pointer to the error code (if available) and possibly an error.
```
benma, line 33: "Returns the error code" (not pointer)

benma (approved):
LGTM except for a small typo in the docstring

PR #2240, opened 2023-09-04
frontend: created WC QR code reading functionality

shonsirsha:
To improve WC user experience especially on mobile, we allow users to connect to dApps using  QR code provided by WC.

We use our existing qr code reader hook to read the QR codes.

Also fixed / adjusted some CSS, as well as moved `scan-qr-dialog` component from `send` to `components` directory as it's now technically a shared component (it's used here in WC qr code scanning desktop view, and still used in the send screen).

Previews: 

Screenshot 2023-09-04 at 12 52 53

Screenshot 2023-09-04 at 17 17 22

### DIFF
--- a/frontends/web/src/components/scanqrdialog/scan-qr-dialog.tsx
+++ b/frontends/web/src/components/scanqrdialog/scan-qr-dialog.tsx
@@ -15,21 +15,36 @@
  */
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { Dialog, DialogButtons } from '../../../../../components/dialog/dialog';
-import { Button } from '../../../../../components/forms';
-import { SpinnerAnimation } from '../../../../../components/spinner/SpinnerAnimation';
-import style from '../../send.module.css';
+import { Dialog, DialogButtons } from '../dialog/dialog';
+import { Button } from '../forms';
+import { SpinnerAnimation } from '../spinner/SpinnerAnimation';
+import style from './scan-qr-dialog.module.css';
 
 type TProps = {
   activeScanQR: boolean;
   onToggleScanQR: () => void;
 }
 
+export const QRVideo = () => {
+  const [videoLoading, setVideoLoading] = useState(true);
+  return (<>
+    {videoLoading &&
+        <div className={style.spinnerAnimationContainer}>
+          <SpinnerAnimation />
+        </div>
+    }
+    <video
+      id="video"
+      height={300 /* fix height to avoid ugly resize effect after open */}
+      className={style.qrVideo}
+      onLoadedData={() => setVideoLoading(false)}
+    />
+  </>);
+};
+
 export const ScanQRDialog = ({ activeScanQR, onToggleScanQR }: TProps) => {
   const { t } = useTranslation();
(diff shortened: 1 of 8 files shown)


frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -81,12 +85,20 @@ export const useQRCodeScanner = ({
...
+            // navigates from WC connect screen to another page
+            // (this error is 'classified' as NotFoundException).
+            if (error.toString().includes('NotFoundException') && isMobile) {
+              return;
+            }
+            // Otherwise, alert as normal.
             alertUser(error.message || error);
           }
+        }).finally(() => {
           onChangeActiveScanQR(false);
```
thisconnect, line 98: remove the other `onChangeActiveScanQR(false);` in the first "then"?

frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -81,12 +85,20 @@ export const useQRCodeScanner = ({
...
           if (error) {
+            // Preventing error alerting when user on mobile
+            // navigates from WC connect screen to another page
+            // (this error is 'classified' as NotFoundException).
+            if (error.toString().includes('NotFoundException') && isMobile) {
+              return;
+            }
+            // Otherwise, alert as normal.
             alertUser(error.message || error);
           }
```
thisconnect, line 96: nit: would it be a bit cleaner to have a onError callback instead of passing isMobile ?

thisconnect (approved):
nice, I added a few small nits.

One point: I tried to scan he QRCode with the camera on my laptop (bit older now) from the QRCode displayed on my phone.. On the video it looked sharp but some of the pixels are rather small... I couldn't get the thing to scan neither in webdev not QTapp.

maybe the WC2 QRcode is not supported by our qrscanner library or maybe my cam is just too low res?

@jadzeidan can you try?

other than that LGTM

frontends/web/src/hooks/qrcodescanner.ts
shonsirsha: Good idea but unfort doesn't work well because we need to call this before parseQRResult. Not 100% sure but I think since parseQRreuslt also uses alertUser / dialog, this "clashes" with the currently opened QR scanner dialog and causing the "with id 'video' not found" error. I'll put this back to the `catch` block.

frontends/web/src/components/scanqrdialog/scan-qr-dialog.tsx
```diff
@@ -37,19 +58,11 @@ export const ScanQRDialog = ({ activeScanQR, onToggleScanQR }: TProps) => {
...
-      <video
-        id="video"
-        width={400}
-        height={300 /* fix height to avoid ugly resize effect after open */}
-        className={style.qrVideo}
+
+      <QRVideo
         onLoadedData={() => setVideoLoading(false)}
+        videoLoading={videoLoading}
       />
```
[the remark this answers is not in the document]
shonsirsha: Yeah good catch, we can put this  state into the QRVideo!

PR #2243, opened 2023-09-07
frontend: use qr-scanner

thisconnect:
Changing library to scan QRCodes from @xzing/library to qr-scanner which is a more modern port of the same library, see
- https://github.com/nimiq/qr-scanner#readme

The original @xzing/library project is in maintenance mode, with no active development, see:
- https://github.com/zxing-js/library

Another problem was that at least our version of @xzing/library could not scan modern WC V2 Qr Codes.

The new library also has a highlightScanRegion feature to render an outline around the scan region on the video stream.

Unfortunatelly the dynamic import did not the same way as with the old library, but it seems this only slightly increases the build size (see below). The issue was a combination of lazy loading the library, still depending on types from it and passing a react DOM ref, which made everything complicated.

Dynamic loading could probably work on a higher parent component, i.e. the whole receive-address or even the full view Send route.

make buildweb before (master 897dd97d2)

    644.14 kB  build/static/js/main.06fd36e9.js
    87.22 kB   build/static/js/459.4bf01830.chunk.js
    35.56 kB   build/static/css/main.9773ffd0.css
    157 B      build/static/js/909.925590b9.chunk.js

make buildweb with this commit

    649.18 kB  build/static/js/main.d274d59b.js
    35.59 kB   build/static/css/main.65886e68.css
    10.32 kB   build/static/js/684.891a2f0a.chunk.js

Other changes:
- increased the size of the QrScanner video popup on desktop


thisconnect:
It works for me in Qt app on macOS 11.7, but we should also test on other platforms Android, Window, Linux

benma:
@thisconnect tested on Android, works very well!

thisconnect:
@shonsirsha PTAL

frontends/web/src/hooks/qrcodescanner.test.ts
```diff
@@ -1,75 +0,0 @@
...
-  it('should determine if a camera is available', async () => {
-    const qrCodeReaderRef = { current: mockedQRCodeReaderInstance() } as unknown as MutableRefObject<BrowserQRCodeReader | undefined>;
-    const props = {
-      qrCodeReaderRef,
-      activeScanQR: false,
-      onChangeActiveScanQR: jest.fn(),
-      parseQRResult: jest.fn()
-    };
-
-    const { result, waitForNextUpdate } = renderHook(() => useQRCodeScanner(props));
```
thisconnect, line 45: @shonsirsha to get this test back it would need some virtual HTMLVideoElement I think.
  
Screen Shot 2023-09-18 at 18 40 34

frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -14,79 +14,63 @@
...
+          scanner?.stop();
+          onResult(result);
+        }, {
+          onDecodeError: onError,
+          highlightScanRegion: true,
+          highlightCodeOutline: true,
         })
-        .catch(error => {
-          if (error) {
-            alertUser(error.message || error);
```
thisconnect, line 84: would be nice to have the error back, i.e if camera has no permission.. 

but not any error for example the old "Video stream has ended before any code could be detected." doens't have to be shown.

frontends/web/src/hooks/qrcodescanner.test.ts
thisconnect: Yes please take the tests over 👍 . thank you 🙏

thisconnect: but if not straight forward or if it is too annoying we can also drop.

frontends/web/src/hooks/qrcodescanner.ts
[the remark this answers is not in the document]
thisconnect: added a301e4c7714a9cf41192d9c1f72facf7f6b18ce4 with a message when camera access hasn't been granted yet.

It works for me in the Qt app, but if I manually revoke permission in macOS system settings I sometimes get false positive. In other words I think it works for the first time but once changed it seems somhow to still remember old permission.

frontends/web/package.json
```diff
@@ -15,10 +15,10 @@
     "typescript": "^4.4.2"
   },
   "dependencies": {
-    "@zxing/library": "0.18.3",
     "flag-icons": "^6.6.6",
     "i18next": "21.6.16",
     "lightweight-charts": "3.8.0",
+    "qr-scanner": "^1.4.2",
```
thisconnect, line 21: I'll remove the caret, so we will have to manually upgrade to new version as described here https://github.com/digitalbitbox/bitbox-wallet-app#npm-dependencies

frontends/web/src/hooks/permissions.ts
```diff
@@ -0,0 +1,43 @@
...
+      // Type '"camera"' is not assignable to type 'PermissionName'.ts(2322)
+      .query({ name: deviceName } as unknown as PermissionDescriptor)
+      .then((permissionStatus) => {
+        setPermissionState(permissionStatus.state);
+        console.log(`camera permission status is ${permissionStatus.state}`);
+        permissionStatus.onchange = () => {
+          setPermissionState(permissionStatus.state);
+          console.log(
+            `camera permission status has changed to ${permissionStatus.state}`,
+          );
```
thisconnect, line 37: I kept the console logs for now for testing.

please note that starting the BBApp from the cli for example `QTWEBENGINE_REMOTE_DEBUGGING=8765 ./frontends/qt/build/osx/BitBox.app/Contents/MacOS/BitBox -testnet`, will use the Terminal.app permissions and not BitBox.app permissions.

thisconnect:
@shonsirsha PTAL

frontends/web/src/routes/account/send/components/inputs/scan-qr-video.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useRef } from 'react';
+import { useQRScanner } from '../../../../../hooks/qrcodescanner';
+import { SpinnerAnimation } from '../../../../../components/spinner/SpinnerAnimation';
+import style from '../../send.module.css';
+import { useDevicePermission } from '../../../../../hooks/permissions';
+import { useTranslation } from 'react-i18next';
```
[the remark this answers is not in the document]
thisconnect: done

thisconnect:
@shonsirsha I added 1 more commit (which I would fix into the last commit), basically it removes the event listener again and only updates the state if mounted. @shonsirsha sorry PTAL one more time.

PR #2244, opened 2023-09-14
ci: use latest nodejs 18.x lts version for macos

thisconnect:
This uses latest nodejs LTS 18.x version used in macOS GitHub ci.

Linux ci does not have specific nodejs version set but installs
nodejs in scripts/docker_install.sh

Changelog:
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#18.17.0
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#18.17.1


thisconnect:
updated commit message.

PR #2245, opened 2023-09-16
testing: run tests with -race flag

NicolaLS:
Add the -race flag for go test in the ci.sh script to detect race conditions.

Fix failing test by closing the channel and document the `quitChan` in the `Account` struct.

**note for reviewers:**

In case there are tests that we know have races but it does not matter or it's intentional we can use the [build tag](https://pkg.go.dev/go/build#hdr-Build_Constraints) `!race` to exclude that test. In this case we would need to add a script to build/run these tests without the `-race` flag.

I can drop the documentation commit if you want as it's a bit unnecessary.

Feel free to close this if you think adding this will be to annoying/slowing dev. down.

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -89,6 +89,8 @@ type Account struct {
 	// if not nil, SendTx() will sign and send this transaction. Set by TxProposal().
 	activeTxProposal *TxProposal
 
+	// quitChan is used to send a quit signal to the accounts long running routines that
+	// should listen to it.
 	quitChan chan struct{}
 
 	log *logrus.Entry
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -71,6 +71,8 @@ type Account struct {
 	initialized     bool
 	initializedLock locker.Locker
 
+	closed bool
+
 	// enqueueUpdateCh is used to invoke an account update outside of the regular poll update
 	// interval.
 	enqueueUpdateCh chan struct{}
@@ -137,7 +139,7 @@ func (account *Account) FilesFolder() string {
 
 func (account *Account) isClosed() bool {
 	defer account.initializedLock.RLock()()
-	return account.quitChan == nil
+	return account.closed
 }
 
 func (account *Account) isInitialized() bool {
@@ -156,7 +158,7 @@ func (account *Account) Initialize() error {
 	}
 
 	defer account.initializedLock.Lock()()
-	if account.quitChan == nil {
+	if account.closed {
 		return errp.New("Initialize: account was closed, init only works once.")
 	}
 	if account.initialized {
@@ -447,7 +449,7 @@ func (account *Account) Close() {
 		account.log.Info("Closed DB")
 	}
 	close(account.quitChan)
-	account.quitChan = nil
+	account.closed = true
 	account.Config().OnEvent(accountsTypes.EventStatusChanged)
 }
 
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -17,9 +17,12 @@
 set -e
 set -x
 
+# Set go-langs data race detector options
+export GORACE="halt_on_error=1"
+
 # This script has to be called from the project root directory.
 go build -mod=vendor ./...
-go test -mod=vendor ./... -count=1 -v
+go test -race -mod=vendor ./... -count=1 -v
 golangci-lint run
 
 npm --prefix=frontends/web install # needed to install dev dependencies.


backend/coins/eth/account.go
```diff
@@ -438,6 +440,7 @@ func (account *Account) FatalError() bool {
 // Close implements accounts.Interface.
 func (account *Account) Close() {
 	defer account.initializedLock.Lock()()
+	account.quitChan <- struct{}{}
```
NicolaLS, line 443: this seems unnecessary because we close the channel later anyway, but the problem is the `account.quitChan = nil` as the channel might be read at the same time. I guess moving the `close(account.quitChan)` up here and leaving the `..= nil` where it is, makes more sense

NicolaLS: actually just removing the `account.quitChan = nil` would be the best solution imo. I don't think it affects how the struct is being garbage collected anyway

benma: as you state, this is the right solution to the problem. this is blocking operation until some goroutine reads it. close(...) broadcasts to all listening goroutines instead.

use a RWMutex on quitChan instead to resolve the race condition.

benma (changes requested):
thanks!

the race check is not deterministic, so it can happen that CI will pass now but will randomly fail in future PRs unrelated to the PRs. let's see how it goes!

backend/coins/eth/account.go
NicolaLS: I used the `locker.Lock` for the convenient `defer` locking/closing.

backend/coins/eth/account.go
```diff
@@ -207,6 +212,7 @@ func (account *Account) Initialize() error {
 }
 
 func (account *Account) poll(initDone func()) {
+	defer account.quitChanLock.RLock()()
```
NicolaLS, line 215: is it ok to RLock once in the start or do we need to do it for both selects ?

benma: should do it for every access to .quitChan. easiest to add a helper function for that which acquires the lock and returns the chan. if you do it like now at the top, it looks like a deadlock? the read-lock is never released, and Close() needs to acquire the lock to be able to close the channel.

backend/coins/eth/account.go
```diff
@@ -447,6 +453,7 @@ func (account *Account) Close() {
 		account.log.Info("Closed DB")
 	}
 	close(account.quitChan)
```
benma, line 455: this should be be guarded too by the lock.

NicolaLS: oh right

backend/coins/eth/account.go
NicolaLS: I thought this change would be straight forward but writing it out gave me a bunch of problems. 1. the polling gets extremely messy with re-locking/unlocking everywhere, a possible fix could be [this](https://gist.github.com/NicolaLS/c85c0eca8afced7dd8b26b88c5c28fb1) And 2. Either I am missing something or it is not really possible to use a RWMutex for this because of the `account.quitChan = nil` basically there is no guarantee that the polling will stop (read the channel) in the short time were it is unlocked after closing:
```
	quitChan, unlock := account.quitChan.WChan()
	close(*quitChan)
	unlock()
	quitChan, _ = account.quitChan.WChan()
	*quitChan = nil
	unlock()
``` 
I might be missing something here but maybe it's best to refactor the account closed checking to not use the channel that is used to quit the polling `account.quitChan == nil` should not be used to check that @benma ?

NicolaLS: note: I created a ```
type quitChan struct {
	mutex   sync.RWMutex
	channel chan struct{}
}
``` with a method to lock it and return the channel and the unlock fn, but I did not push this yet as I can't get the tests to pass..

Comment (benma) [reply backend/coins/eth/account.go:215]: I meant helper method that looks sth like this:

```
func (account *Account) getQuichChan() <-chan struct{} {
    defer account.quitChanLock.RLock()()
    return account.quitChan
}
```

then in select:

`case <-account.getQuitChan():` etc.

Comment (NicolaLS) [reply backend/coins/eth/account.go]: I changed it like that, PTAL. I think this is not correct though the channel gets RUnlocked as soon as the method returns. Because it is being locked swiftly it might help with synchronization but the channel is **not**  locked when we read it later (?).

File: backend/coins/eth/account.go [hunk 4]
```diff
@@ -446,8 +454,12 @@ func (account *Account) Close() {
 		}
 		account.log.Info("Closed DB")
 	}
+	unlock := account.quitChanLock.Lock()
 	close(account.quitChan)
+	unlock()
+	unlock = account.quitChanLock.Lock()
 	account.quitChan = nil
+	unlock()
```
NicolaLS, line 462: This looks really wrong..but we have to let readers notice the channel closed before we set it to `nil` this is why I lock -> close -> unlock -> lock -> nil.

backend/coins/eth/account.go
```diff
@@ -438,6 +445,7 @@ func (account *Account) FatalError() bool {
 // Close implements accounts.Interface.
 func (account *Account) Close() {
 	defer account.initializedLock.Lock()()
```
benma, line 447: Double lock looks weird.

in btc/account.go, initializedLock is for both `.initialized` and `.closed`. I think the same could make sense here.

NicolaLS: makes sense, I also changed the name from `accountClosed` to `closed`.
I guess `initializedLock` should maybe be renamed to `stateLock` or something like this, but not in this PR.

scripts/ci.sh
```diff
@@ -17,9 +17,12 @@
 set -e
 set -x
 
+# Set go-langs data race detector options
+export GORACE="halt_on_error=1"
```
benma, line 21: would be good to know the reason for this either as a comment or in the commit msg, because it is not the default and it seems normal to continue running other unit tests to gather all problems.

benma (commented):
LGTM! thanks for your patience with this one. just left one more question

scripts/ci.sh
NicolaLS: I thought it would be reasonable to stop as soon as this fails but if you think it's better to just run all tests even if there is a race detection fail let me know I can remove it. Otherwise I'm not sure what to comment the `halt_on_error` is pretty self documenting right ?

NicolaLS: I think if this fails its hard to see other failures scrolling through the logs in GH. Or you just see that test `xyz` failed, then fix it and push again, just to learn that you did not scroll up enough to see that a race detection failed too, so you'll have to go back and fix/push again.

benma (approved):

PR #2247, opened 2023-09-18
backend: add PreviousOutputs to TxProposal

NicolaLS:
Refactor the TxProposal to contain PreviousOutputs which are all Outputs selected in coin selection as Inputs for the transaction.

I noticed we pass all spendable outputs as `previousOutputs` [here](https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/transaction.go#L240-L244) this is not a performance problem since it is a map but I felt like it would make more sense that it is available in the TxProposal since it knows which Outputs it used for its Transaction Inputs.

**motivation**: drafting CPFP I refactored the `SpendableOutputs` to also return unconfirmed outputs and it made me wonder if it is correct to pass all outputs to the `signTransaction` function.

**note**: maybe its not good that `maketx` depends on `transactions` now..but we need the methods defined on `SpendableOutput`. An alternative would be to just duplicate the SpendableOutput in maketx. 

This might be kind of a unnecessary refactor, I'm just trying to do something with the codebase, so feel free to close.

### DIFF
--- a/backend/coins/btc/maketx/maketx.go
+++ b/backend/coins/btc/maketx/maketx.go
@@ -23,11 +23,20 @@ import (
 	"github.com/btcsuite/btcd/wire"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/errors"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/addresses"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/transactions"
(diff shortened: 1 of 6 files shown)


NicolaLS:
side note: ik that it is important that the wallet generally does not consider unconfirmed (with not allInputsOurs) outputs. I'll submit a draft PR soon regarding the changes necessary for CPFP where you'll see how I refactored/changed the `SpendableOutputs`

PR #730, opened 2019-12-10
Erc20

Tomasvrba:

PR #2252, opened 2023-09-20
Remove sai token

NicolaLS:
Removes the SAI token from frontend and backend.

I found the files like this:
`grep -rwiIl 'sai' ./frontends/web/src/ ./backend/`

Also checked that everything that was added in #730 [commit](https://github.com/digitalbitbox/bitbox-wallet-app/pull/730/commits/e5f2fc035f5b05954622f7021390c1c738cac541) was removed.


benma (changes requested):
Thanks!

Could you also make a config migration to remove it from accounts.json? Currently this remains in the ETH accounts if this token was previously active:

```
 "activeTokens": [
                "eth-erc20-sai0x89d2"
            ]
```

Similar to [here](https://github.com/digitalbitbox/bitbox-wallet-app/blob/a8b58dd8cf97463d4d1f54bdbc3378536f36aca7/backend/config/config.go#L262), but for the accounts config not the app config.

backend/config/config.go
```diff
@@ -438,3 +438,20 @@ func migrateUserLanguage(appconf *AppConfig) {
...
+// migrateActiveTokens removes tokens from AccountsConfig
+func migrateActiveTokens(accountsConf *AccountsConfig) {
+	for _, account := range accountsConf.Accounts {
+		// copy account so it does not take a reference to the loop scoped account
+		ac := account
+		if ac.CoinCode != coin.CodeETH {
+			continue
+		}
+
+		err := ac.SetTokenActive("eth-erc20-sai0x89d", false)
```
NicolaLS, line 451: I never really worked with migrations (when I did they were created automatically) it feels wrong to hard code it like this but I guess this is what we need to do here ? @benma and in case there are other tokens added and then removed they would also need to be removed here ?

benma: Well it's not a database with a schema or its own query language like SQL, it is just a json file. Maybe we can check if there are some tools overlaying a schema and automatic migrations, but that would be very low prio - what we have is working well enough.

I think this is the first token to be removed.

backend/config/config.go
```diff
@@ -438,3 +438,20 @@ func migrateUserLanguage(appconf *AppConfig) {
 		delete(frontconf, "userLanguage")
 	}
 }
+
+// migrateActiveTokens removes tokens from AccountsConfig.
+func migrateActiveTokens(accountsConf *AccountsConfig) {
+	for _, account := range accountsConf.Accounts {
+		// copy account so it does not take a reference to the loop scoped account.
+		ac := account
```
benma, line 446: that should not be needed here right? the account var is not accessed after the loop iteration.

backend/config/config.go
```diff
@@ -438,3 +438,20 @@ func migrateUserLanguage(appconf *AppConfig) {
 		delete(frontconf, "userLanguage")
 	}
 }
+
+// migrateActiveTokens removes tokens from AccountsConfig.
+func migrateActiveTokens(accountsConf *AccountsConfig) {
```
benma, line 443: looks like you are not calling this function anywhere yet.

benma (changes requested):

backend/config/config.go
NicolaLS: True, I thought accessing an attribute on it could lead to problems but that would just be the case if it was in a go routine.

backend/config/config.go
NicolaLS: oh....yes

backend/config/config.go
```diff
@@ -256,10 +256,12 @@ func NewConfig(appConfigFilename string, accountsConfigFilename string) (*Config
...
 	appconf := config.appConfig
+	accountconf := config.accountsConfig
 	migrateFiatList(&appconf)
 	migrateFiatCode(&appconf)
 	migrateElectrumX(&appconf)
 	migrateUserLanguage(&appconf)
+	migrateActiveTokens(&accountconf)
 	if err := config.SetAppConfig(appconf); err != nil {
 		return nil, errp.WithStack(err)
 	}
```
benma, line 267: There is SetAppConfig() but no corresponding SetAccontsConfig(), so it is confusing what is happening and if the migration is actually saved. It does not look like the migration is actually saved right now.

You probably want to use `config.ModifyAccountsConfig(...)` instead and put `migrateActiveTokens()` inside the callback of that function.

benma (changes requested):

backend/config/config.go
NicolaLS: Sorry no idea how I missed that, I remember looking and thinking _ah its persisted here_ but that was only the appconf..

benma (approved):
Thanks, LGTM. Could you please squash all commits? One way to think of commits is in logical units - if you wanted to revert the change, you'd want to revert backend/frontend/migration, so it makes sense for them to be in one commit.

backend/config/config.go
```diff
@@ -438,3 +441,20 @@ func migrateUserLanguage(appconf *AppConfig) {
 		delete(frontconf, "userLanguage")
 	}
 }
+
+// migrateActiveTokens removes tokens from AccountsConfig.
+func migrateActiveTokens(accountsConf *AccountsConfig) error {
+	for _, account := range accountsConf.Accounts {
+		// copy account so it does not take a reference to the loop scoped account.
```
benma, line 448: remove this comment, it is a leftover from before when you copied the var

benma (changes requested):

benma (approved):
utACK

PR #2253, opened 2023-09-21
frontend: revert / put back WC POC

shonsirsha:
frontend: revert / put back WC POC :)


Tomasvrba:
I can't merge this anyway, so since you said adding the signing back using the new implementation is almost done, let's just focus on merging that asap

PR #2254, opened 2023-09-21
backend: add unit tests for eth address>account code matching

Tomasvrba:


backend/coins/eth/account_test.go
```diff
@@ -158,3 +159,40 @@ func TestTxProposal(t *testing.T) {
...
+	acct := newAccount(t)
+	defer acct.Close()
+	acct.Synchronizer.WaitSynchronized()
+
+	// Test invalid Ethereum address
+	t.Run("Invalid Ethereum address", func(t *testing.T) {
+		matches, err := acct.MatchesAddress("invalid_address")
+		require.Error(t, err)
+		require.Equal(t, matches, false)
+		assert.Contains(t, err.Error(), "invalidAddress")
```
benma, line 173: can check here that it is actually `errors.ErrInvalidAddress` 

    require.Equal(s.T(), errors.ErrInvalidAddress, errp.Cause(err))

benma (changes requested):
thx!

Tomasvrba:
Not sure why CI failed because of locize, there aren't any changes that should affect that here and running make locize-fix didn't change any files

thisconnect:
I think we merged a formatting issue and probably due to nodejs not running properly. nodejs should be fixed on master and will be merged into staging-walletconnect

If I run locize fix I get changes, what output did you get, did locize even run?

instead of installing locize globally you can also do `npx locize-cli format frontends/web/src/locales --format json`

> Not sure why CI failed because of locize, there aren't any changes that should affect that here and running make locize-fix didn't change any files

benma:
@Tomasvrba maybe a rebase will fix it? seems like it is fixed in the staging branch.

Tomasvrba:
> @Tomasvrba maybe a rebase will fix it? seems like it is fixed in the staging branch.

should be all fixed up now

PR #2257, opened 2023-09-24
frontend: created approve & reject WC signing request ability

shonsirsha:
Users are now able to receive signing request events sent from dapp connected through WC. Upon receiving the signing requests, users can reject or approve those requests in **anywhere (in any page)** within the BBapp.

Which then to be continued in the bb02 device where the actual signing takes place.

[Preview]:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/fb1650c3-7b16-4cbc-bda4-fc165ea184a9


Tomasvrba:
Awesome, thanks @shonsirsha!
I'll test tomorrow

frontends/web/src/components/icon/icon.tsx
```diff
@@ -19,6 +19,7 @@ import alert from './assets/icons/alert-triangle.svg';
 import BB02StylizedDark from '../../assets/device/bitbox02-stylized-reflection-dark.png';
 import BB02StylizedLight from '../../assets/device/bitbox02-stylized-reflection-light.png';
 import info from './assets/icons/info.svg';
+import arbitrumSVG from '../icon/assets/arbitrum.svg';
```
thisconnect, line 22: `'../icon/assets/` -> `'./assets/`, also for bscSVG  and the other icons here.

frontends/web/src/utils/walletconnect-eth-sign-handlers.ts
```diff
@@ -0,0 +1,179 @@
...
+  const requestParams = params.request.params;
+  const accountAddress = requestParams[0];
+  const data = requestParams[1];
+  const typedData = JSON.parse(data);
+  const { accountName, accountCode } = await fetchAccountNameAndAddress(accountAddress);
+  const apiCaller = async () => {
+    // If the typed data to be signed includes its own chainId, we use that, otherwise use the id in the params
+    const chainId = typedData.domain.chainId ?
+      +typedData.domain.chainId :
+      +params.chainId.replace(/^eip155:/, '');
```
thisconnect, line 115: Is the  `+` to convert to Number? In other places in the codebase we use `Number('123')`, same in `ethSignOrSendTransactionHandler`

Tomasvrba: that's probably from my code in the POC, I like the unary plus :)))

thisconnect: please use `Number()` as we use in other places in the codebase, and is less mental overhead when reading the code. unary plus and Number apparently behaves the same.

frontends/web/src/api/account.ts
```diff
@@ -71,7 +71,7 @@ type TETHAddressByAccountCode = SuccessResponse & {
   name: string;
 }
 
-export const getETHAddressByAccountCode = (address: string): Promise<TETHAddressByAccountCode> => {
+export const getEthAccountCodeAndNameByAddress = (address: string): Promise<TETHAddressByAccountCode> => {
```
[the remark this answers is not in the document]
Tomasvrba: The type still has the old name in the Promise: TETHAddressByAccountCode

frontends/web/src/components/wallet-connect/incoming-signing-request-dialog.tsx
```diff
@@ -0,0 +1,145 @@
...
+              <li className={styles.item}>
+                <p className={styles.label}>{t('transaction.details.type')}</p>
+                <p className={styles.itemText}>{method}</p>
+              </li>
+
+              {signingData &&
+            (
+              <li className={styles.item}>
+                <p className={styles.label}>{t('walletConnect.signingRequest.data')}</p>
+                <textarea className={styles.textarea} readOnly value={signingData.toString()} />
```
Tomasvrba, line 118: This can be a TODO for a later PR as well, but in the case of signing Typed Data, can we prettify the data field so it has JSON formatting (new lines and indents) so it's more readable and expand the input field + modal so the entire text is visible?

thisconnect: does it have complicated nested data? I guess `JSON.stringify(data, null, 2);` could work.

Tomasvrba: for me, whatever works to make it a bit more readable and not too scary to the average user

Tomasvrba:
>     6. Verify that all of the above also works with multiple account connected (when we have more than just 1 session).

Everything works for me, nice!

For #6 I would just like to add a future TODO (I can do this as well, depends who gets to it first) and that's to implement the 'accountChanged' event

multiple accounts work right now when connected to different apps, but we cannot connect multiple accounts to the same dapp

should maybe give the user a selection of accounts which they want to pair with a dapp like in the example wallet: https://react-wallet.walletconnect.com/

thisconnect (approved):
LGTM with one last question

frontends/web/src/utils/walletconnect-eth-sign-handlers.ts
```diff
@@ -0,0 +1,195 @@
...
+  const data = requestParams[1];
+  let typedData: any;
+  const { accountName, accountCode } = await fetchAccountNameAndAddress(accountAddress);
+
+  try {
+    typedData = JSON.parse(data);
+  } catch (e) {
+    console.error('Failed to parse JSON', e);
+    typedData = undefined;
+    // If JSON parsing fails, typedData will be undefined.
```
thisconnect, line 123: thanks! 

one alternative could be to just assign the string to this variable `typedData = data` and then it signs that string, but maybe this never happens.

cc @Tomasvrba WDYT?

Tomasvrba: Maybe do it like I did with the hex data decoding here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/5ccc03aae817aea0fde54199bf1dbb91f208eecc/frontends/web/src/utils/walletconnect-eth-sign-handlers.ts#L80-L82

ie. return when the parsing/decoding fails? There's probably a reason why it failed so we shouldn't send it to the BB to sign. Or if it was just a glitch in the app, the user can try again, but better be safe imo

thisconnect:
macOS pipeline fails due to nodejs, should be fixed with https://github.com/digitalbitbox/bitbox-wallet-app/commit/f010a5c474270c23747ec90b35ccc3c0f14ba4a2 and ok once we merge master->staging-walletconnect

thisconnect:
LGTM

PR #2258, opened 2023-09-25
frontend: make clean work cross platform

thisconnect:
make buildweb is required to install all frontend dependencies, and get started with using make webdev. Unfortunatelly it fails on windows as Powershell errors with

    process_begin: CreateProcess(NULL, rm -rf frontends/web/build, ...) failed.
    make (e=2): The system cannot find the file specified.
    make: *** [Makefile:55: buildweb] Error 2

This adds a propoer clean task to the frontend directory, without using rimraf or other dependency, just calling node directly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ servewallet-mainnet-prodservers:
 buildweb:
 	node --version
 	npm --version
-	rm -rf ${WEBROOT}/build
+	cd ${WEBROOT} && ${MAKE} clean
 	cd ${WEBROOT} && npm ci --ignore-scripts
 	cd ${WEBROOT} && npm run build
 webdev:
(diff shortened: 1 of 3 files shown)


benma (commented):
I don't think the Makefile is supposed to work in Powershell generally speaking. 

On Windows, you can/should use bash, e.g. git-bash, according to the [doc](https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/docs/BUILD.md#windows). 

There are more `rm -rf` lines and probably many other things that won't work in Powersell.

Would it be better to simply document better that one should use git-bash on Windows?

thisconnect:
yes there are other `rm -rf` but buildweb would be nice to get to work so that webdev can work. the others are more for building the Qt app.

> There are more rm -rf lines and probably many other things that won't work in Powersell.

I'll check if git-bash works out of the box... 

> Would it be better to simply document better that one should use git-bash on Windows?

thisconnect:
git-bash works, closing this.

PR #2260, opened 2023-09-26
frontend: merge master to staging-walletconnect

shonsirsha:
frontend: merge master to staging-walletconnect

also integrated the newly created qr code reader components & hooks for WC.


thisconnect (approved):
untestedLGTM, one small nit:

could you re-create the lock file (dep tree) with `cd frontends/web && rm package-lock.json && npm i --ignore-scripts`. When I do that I get slightly newer dependencies.

PR #2262, opened 2023-09-27
refactor Confrim.tsx

NicolaLS:
As @thisconnect said in #1625:
_"Deprecate Confirm component, exporting a function and updating
the state of the global confirm instance is a bit an antipattern."_

I renamed `Confrim.tsx` to `Confrim-Legacy.tsx` to make this clear and to introduce a new version of the component (called `Confirm.tsx`).

The new component will be useful to avoid verbosity/bloat when we need a simple confirm dialog with e.g a message as `show-recovery-words-settings.tsx` needed for example, because without using the `Legacy-Confirm` we would need to use the bare `Dialog`.

note: I made this change as a setup to add the **warning** (_do not show your recovery words..._) to the dialog. This was not possible in the Legacy-Confirm but now you can pass a message but also custom children (`childrenBefore`, `childrenAfter`)


thisconnect:
also in https://github.com/digitalbitbox/bitbox-wallet-app/pull/1625

> Just use a dialog in the component instead of calling the global confirm instance. We already do that in most places.

so if I understand this change is only needed to show another component in the confirm dialog?

if so, can't you just use Dialog instead of Confirm?

NicolaLS:
> also in #1625
> 
> > Just use a dialog in the component instead of calling the global confirm instance. We already do that in most places.
> 
> so if I understand this change is only needed to show another component in the confirm dialog?
> 
> if so, can't you just use Dialog instead of Confirm?

essentially **yes**. but I also thought it might be good to have a confirm that does not use `DialogLegacy` but still does the same thing (showing a dialog, parsing a message into `SimpleMarkup`s)

NicolaLS:
Closing as this refactor is not necessary

thisconnect:
👍 

> essentially yes. but I also thought it might be good to have a confirm that does not use DialogLegacy but still does the same thing (showing a dialog, parsing a message into SimpleMarkups).

We already have: WaitDialog, Dialog, DialogLegacy, View[type=dialor], Confirm ! 🤯 over time we should reduce them.

PR #2263, opened 2023-09-27
Show recovery words warning

NicolaLS:
Show a warning in the app before showing the recovery words (mnemonic) on the device:
_Never share your recovery words with anyone. Your recovery words give full access to your wallet. If someone is asking you for your recovery words, it's a scammer, do not share them!_

~~Requires: #2262 to be merged-~~

~~Please point me in the right direction on how to style this, if I understood correctly it should not be part of the `description` in the `app.json` but another key (`warning`) this means it could also use a strong tag anywhere but we would need to use `SimpleMarkup` instead of `Message`.~~

~~note: **don't look at the 'files changed' diff as it contains commits from the confrim dialog refactor, select the last commit to see the relevant changes**~~

### DIFF
--- a/frontends/web/src/routes/settings/components/device-settings/show-recovery-words-setting.tsx
+++ b/frontends/web/src/routes/settings/components/device-settings/show-recovery-words-setting.tsx
@@ -19,9 +19,11 @@ import { useTranslation } from 'react-i18next';
 import { SettingsItem } from '../settingsItem/settingsItem';
 import { ChevronRightDark } from '../../../../components/icon';
 import { WaitDialog } from '../../../../components/wait-dialog/wait-dialog';
-import { MultilineMarkup } from '../../../../utils/markup';
+import { MultilineMarkup, SimpleMarkup } from '../../../../utils/markup';
 import { showMnemonic } from '../../../../api/bitbox02';
-import { confirmation } from '../../../../components/confirm/Confirm';
+import { Message } from '../../../../components/message/message';
+import { Dialog, DialogButtons } from '../../../../components/dialog/dialog';
+import { Button } from '../../../../components/forms';
 
 type TProps = {
   deviceID: string;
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/settings/components/device-settings/show-recovery-words-setting.tsx
```diff
@@ -34,25 +36,40 @@ type TDialog = {
...
       <SettingsItem
         settingName={t('backup.showMnemonic.title')}
         secondaryText={t('deviceSettings.backups.showRecoveryWords.description')}
         extraComponent={<ChevronRightDark />}
-        onClick={handleShowMnemonic}
+        onClick={() => setShowDialog(true)}
       />
       <ShowMnemonicWaitDialog inProgress={inProgress} />
+
+      <Dialog title={t('backup.showMnemonic.title')} open={showDialog} onClose={() => setShowDialog(false)}>
```
NicolaLS, line 58: the `onClose` does not seem necessary but otherwise you can't open the dialogue again after exiting with `esc` key instead of the cancel button.

frontends/web/src/routes/settings/components/device-settings/show-recovery-words-setting.tsx
```diff
@@ -66,6 +83,10 @@ const ShowMnemonicWaitDialog = ({ inProgress }: TDialog) => {
 
   return (
     <WaitDialog title={t('backup.showMnemonic.title')}>
+      <Message type="warning">{
+        <SimpleMarkup tagName="span" markup={t('backup.showMnemonic.warning')}/>
+      }
```
thisconnect, line 88: nit: there should be no need for wrapping `<SimpleMarku` in brackets `{`

thisconnect (changes requested):

thisconnect (approved):
thanks! tested LGTM 👍

PR #2264, opened 2023-09-28
frontend/guide/entry: fix repeated key=''

benma:
Multiple empty lines in an entry result in a duplicate key. We use the index instead. This is not a problem as the order of the lines in the entry never change.

### DIFF
--- a/frontends/web/src/components/guide/entry.tsx
+++ b/frontends/web/src/components/guide/entry.tsx
@@ -54,7 +54,7 @@ export const Entry: FunctionComponent<TProps> = props => {
       <div className={[style.entryContent, shown ? style.expanded : ''].join(' ')}>
         {shown ? (
           <div className="flex-1">
-            {entry.text.trim().split('\n').map(p => <p key={p}>{p}</p>)}
+            {entry.text.trim().split('\n').map((p, idx) => <p key={idx}>{p}</p>)}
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/guide/entry.tsx
```diff
@@ -54,7 +54,7 @@ export const Entry: FunctionComponent<TProps> = props => {
       <div className={[style.entryContent, shown ? style.expanded : ''].join(' ')}>
         {shown ? (
           <div className="flex-1">
-            {entry.text.trim().split('\n').map(p => <p key={p}>{p}</p>)}
+            {entry.text.trim().split('\n').map((p, idx) => <p key={idx}>{p}</p>)}
```
thisconnect, line 57: > Multiple empty lines in an entry result in a duplicate key

this would only if the lines have same content correct?

benma: yeah see commit msg 

> Multiple empty lines in an entry result in a duplicate key.

thisconnect (approved):
LGTM

frontends/web/src/components/guide/entry.tsx
thisconnect: thanks for highlighting, I somehow didn't see the word "empty"

PR #2265, opened 2023-09-28
defer some keystore checks

benma:


Beerosagos (approved):
tACK!

Do you think it would make sense to use error codes to allow translating the errors?

benma:
> tACK!
> 
> Do you think it would make sense to use error codes to allow translating the errors?

Thanks! I considered and will update the commit msg - I think the overhead is not worth it as one error will not be shown (xpub only on btc is filtered by the frontend already) and signing messages is supported by the BB02 and we don't consider BB01 anymore (EOL).

PR #2268, opened 2023-10-01
frontend: optimise wc loading and web3wallet initialisation

shonsirsha:
Every time a user visits `/wallet-connect/dashboard`, the app initialises the `web3wallet` object, achieved through the use of `Core` and `Web3Wallet`, which are lazy-loaded. The function is called `initializeWeb3Wallet`

Only on a successful dapp via WC connection, `hasUsedWalletConnect` is set to config. This condition essentially used to check if we should call `initializeWeb3Wallet` on app start (via provider), or on dashboard first visit.

### DIFF
--- a/frontends/web/src/contexts/WCWeb3WalletProvider.tsx
+++ b/frontends/web/src/contexts/WCWeb3WalletProvider.tsx
@@ -16,9 +16,10 @@
 
 import { ReactNode, useEffect, useState } from 'react';
 import { WCWeb3WalletContext } from './WCWeb3WalletContext';
-import { IWeb3Wallet, Web3Wallet } from '@walletconnect/web3wallet';
-import { Core } from '@walletconnect/core';
+import { IWeb3Wallet } from '@walletconnect/web3wallet';
 import { getTopicFromURI, pairingHasEverBeenRejected } from '../utils/walletconnect';
+import { useLoad } from '../hooks/api';
+import { getConfig, setConfig } from '../utils/config';
 
 type TProps = {
     children: ReactNode;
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/account/walletconnect/dashboard.tsx
```diff
@@ -37,14 +39,25 @@ type TProps = {
...
 
   const updateSessions = useCallback(() => {
     const activeSessions = Object.values(web3wallet?.getActiveSessions() || []);
     setSessions(activeSessions);
   }, [web3wallet]);
 
+  useEffect(() => {
+    if (config && config.frontend && !config.frontend.hasUsedWalletConnect) {
+      setConfig({ frontend: { hasUsedWalletConnect: true } }).then(() => {
+        navigate(0); // reloads the page
```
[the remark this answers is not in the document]
thisconnect: This works in webdev, but the Qt app errors. Reason (I think) is that on-load Qt finds `index.html` and the `location.href` is something like `qrc:/index.html`, but on-reload the location is something like `qrc:/account/v0-12345678-eth-0/wallet-connect/dashboard'`.

Screen Shot 2023-10-02 at 12 56 31

just thinking out loud: could the part that loads the library in the provider be moved to a function "load" and also exposed by the provider so that "load" can be called from here?

I guess that could work, but I have no idea if and how the web3wallet would be updated.

If that works somehow, I am thinking if the setConfig part should be moved to "load" function instead of here.

frontends/web/src/contexts/WCWeb3WalletProvider.tsx
```diff
@@ -27,56 +28,69 @@ type TProps = {
...
-          setIsWalletInitialized(true);
-        } catch (err: unknown) {
-          console.log('Error for initializing', err);
+      const wallet = await Web3Wallet.init({
+        core,
+        metadata: {
+          name: 'BitBoxApp',
+          description: 'BitBoxApp',
+          url: 'https://bitbox.swiss',
+          icons: ['https://bitbox.swiss/assets/images/logos/dbb-logo.png']
```
thisconnect, line 49: @Tomasvrba what should we use for name and description? and isn't this also configured somewhere in walletconnect account?

I am currently thinking it should be 'BitBox' and the description should maybe mention that BitBox is a hardware wallet.

Also should this be translated or hardcoded to english language?

Tomasvrba: BitBox as name is fine imo
and maybe something like "BitBox02 hardware wallet" as description

The thing in WC cloud is what is displayed in the explorer/when users selects a wallet to connect
The info entered here is what will be shown in the dapp during/after pairing

English is fine imo since there isn't really much text there

thisconnect: ack makes sense to me.

thisconnect (approved):
tested LGTM 👍 
 
please update the commit message. with that config.json is only updated once a successful connection is created.

frontends/web/src/contexts/WCWeb3WalletProvider.tsx
thisconnect: @shonsirsha feel free to change to "BitBox" and "BitBox02 hardware wallet" in this PR, or we can also do later in a different PR.. up to you.

PR #2270, opened 2023-10-02
backend: fix SendAll bitcoin

NicolaLS:
The TxProposal refactor #2247 broke the SendAll funtionallity because it did not add the PreviousOutputs in the NewTxSpendAll but only in the NewTx. This commit fixes this by also adding it into the NewTxSpendAll function.


backend/coins/btc/maketx/maketx.go
```diff
@@ -142,12 +142,16 @@ func NewTxSpendAll(
 ) (*TxProposal, error) {
 	selectedOutPoints := []wire.OutPoint{}
 	inputs := []*wire.TxIn{}
+	previousOutputs := make(PreviousOutputs, len(selectedOutPoints))
```
benma, line 145: len(selectedOutpoints) is `0`. did you mean `len(spendableOutputs)`?

benma (changes requested):

backend/coins/btc/maketx/maketx.go
NicolaLS: oh yes..

PR #2274, opened 2023-10-03
frontend: fix slow camera detection

thisconnect:
On macOS M1/M2 the whole send screen takes a long time to load and QrScanner icon is even slower. This is an indication that qr-scanner hasCamera takes very long to test if there are any available cameras.

https://github.com/nimiq/qr-scanner/blob/abcfe1bce2703721408d8ce7ebde94a359998506/src/qr-scanner.ts#L13-L54

Unfortunatelly it looks like that enumerateDevices is for some reason very slow.

     await navigator.mediaDevices.enumerateDevices();

The app used to check if there are any camera devices and only shows the camera icon if some cameras have been detected.

This change removes the check and now always shows the Qr code scanner icon.

superseds https://github.com/digitalbitbox/bitbox-wallet-app/pull/2266


thisconnect:
@shonsirsha can you test this in Qt app on macOS with camera?

@jadzeidan can you test this on a machine without camera?

thisconnect:
@benma PTAL

benma (changes requested):
Send screen opens fast now. Clicking the cam icon, it again starts lagging/freezing but it works in the end. 

Small issue: while it is loading it says 'please grant access to the camera' even though access has been given already, it is simply waiting to load the camera stream. The message could be confusing in that moment.

There is a bug though: Opening the 'Scan QR code' dialog a second time does not work - the dialog opens with the loading dots loading forever.

thisconnect:
@benma thanks for testing 🙏 

> Send screen opens fast now. Clicking the cam icon, it again starts lagging/freezing but it works in the end.

works fast on apple intel with macOS 11.x  I'll try to test it on a different intel with newer macOS to see if macOS version matters.


> Small issue: while it is loading it says 'please grant access to the camera' even though access has been given already, it is simply waiting to load the camera stream. The message could be confusing in that moment.

fixed in newest commit.


> There is a bug though: Opening the 'Scan QR code' dialog a second time does not work - the dialog opens with the loading dots loading forever.

I could not reproduce with this branch nor with master (2c667fd98b8f1261c1d7767e983fb7258e8f11aa)

thisconnect:
rebased

benma (approved):
Tested now. Loading send screen is fast. Clicking camera is slow until it is ready, but then it works smoothly. The previous bug of opening the cam dialog twice disappeared, it works now.

The slow cam loading is already present on the last release, so it is not a regression with the new qr code library.

PR #2275, opened 2023-10-04
frontend/account: add alert if insured account has uncovered utxos

Beerosagos:
Bitsurance service only covers p2wpkh utxos. This adds a check in the account page that warns the user in case the account is insured but there are uncovered utxos that need to be moved.

### DIFF
--- a/frontends/web/src/routes/account/account.tsx
+++ b/frontends/web/src/routes/account/account.tsx
@@ -41,6 +41,10 @@ import { Insured } from './components/insuredtag';
 import { AccountGuide } from './guide';
 import { BuyReceiveCTA } from './info/buyReceiveCTA';
 import { isBitcoinBased } from './utils';
+import { getScriptName } from './utils';
+import { MultilineMarkup } from '../../utils/markup';
+import { Dialog } from '../../components/dialog/dialog';
+import A from '../../components/anchor/anchor';
 
 type Props = {
   accounts: accountApi.IAccount[];
@@ -61,10 +65,22 @@ export function Account({
   const [transactions, setTransactions] = useState<accountApi.TTransactions>();
   const [usesProxy, setUsesProxy] = useState<boolean>();
   const [insured, setInsured] = useState<boolean>(false);
+  const [uncoveredFunds, setUncoveredFunds] = useState<string[]>([]);
   const [stateCode, setStateCode] = useState<string>();
   const supportedExchanges = useLoad<SupportedExchanges>(getExchangeBuySupported(code), [code]);
 
   const account = accounts && accounts.find(acct => acct.code === code);
+
+  const checkUncoveredUTXOs = useCallback(async () => {
+    const uncoveredScripts: accountApi.ScriptType[] = [];
+    const utxos = await accountApi.getUTXOs(code);
+    utxos.forEach((utxo) => {
+      if (utxo.scriptType !== 'p2wpkh' && !uncoveredScripts.includes(utxo.scriptType)) {
+        uncoveredScripts.push(utxo.scriptType);
+      }
+    });
+    setUncoveredFunds(uncoveredScripts.map(getScriptName));
+  }, [code]);
 
   const maybeCheckBitsuranceStatus = useCallback(async () => {
     if (account?.bitsuranceId) {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/account.tsx
```diff
@@ -61,11 +65,23 @@ export function Account({
...
+    let uncoveredScripts: accountApi.ScriptType[] = [];
+    const utxos = await accountApi.getUTXOs(code);
+    utxos.forEach((utxo) => {
+      if (utxo.scriptType !== 'p2wpkh' && !uncoveredScripts.includes(utxo.scriptType)) {
+        uncoveredScripts.push(utxo.scriptType);
+      }
+    });
+    setUncoveredFunds(uncoveredScripts.map(script => getScriptName(script)));
+  }, [code]);
+
```
benma, line 84: This is "business logic" creeping into the frontend, not sure what I think of it. I know it is a bit easier to do it likes this sometimes, but ideally the frontend is solely for visualization of data delivered by the backend. 

What do you think of moving this to the backend w/ a separate API endpoint? I feel bad for asking because I see that it is more effort - how important is it that we stay consistent and keep the logic in the backend?

frontends/web/src/routes/account/account.tsx
```diff
@@ -222,6 +239,13 @@ export function Account({
         <Status hidden={!hasCard} type="warning">
           {t('warning.sdcard')}
         </Status>
+        <Dialog open={insured && uncoveredFunds.length !== 0} medium title={t('account.warning')} onClose={() => setUncoveredFunds([])}>
+          <MultilineMarkup tagName="p" markup={t('account.uncoveredFunds', {
+            name: account?.name,
```
benma, line 244: why the question mark? `account` should be defined, right?

frontends/web/src/routes/account/account.tsx
```diff
@@ -61,11 +65,23 @@ export function Account({
...
 
+  const checkUncoveredUTXOs = useCallback(async () => {
+    let uncoveredScripts: accountApi.ScriptType[] = [];
+    const utxos = await accountApi.getUTXOs(code);
+    utxos.forEach((utxo) => {
+      if (utxo.scriptType !== 'p2wpkh' && !uncoveredScripts.includes(utxo.scriptType)) {
+        uncoveredScripts.push(utxo.scriptType);
+      }
+    });
+    setUncoveredFunds(uncoveredScripts.map(script => getScriptName(script)));
```
benma, line 82: `.map(getScriptName)` should work as well

benma (approved):
utACK

frontends/web/src/routes/account/account.tsx
```diff
@@ -222,6 +239,13 @@ export function Account({
         <Status hidden={!hasCard} type="warning">
           {t('warning.sdcard')}
         </Status>
+        <Dialog open={insured && uncoveredFunds.length !== 0} medium title={t('account.warning')} onClose={() => setUncoveredFunds([])}>
+          <MultilineMarkup tagName="p" markup={t('account.uncoveredFunds', {
+            name: account?.name,
+            uncovered: uncoveredFunds,
+          })}/>
+          <A href="#">{t('account.uncoveredFundsLink')}</A>
```
thisconnect, line 247: small nit: using anchor with `href="#"` is usually indication that we should use a button element.

in BBApp iirc we use `<A>` component to open links in default browser.

Beerosagos: yeah, this should be a link, we just don't have the guide page yet! Maybe I can add a //TODO comment to fix before merging into master

frontends/web/src/routes/account/account.tsx
Beerosagos: Keeping as is, since this is consistent with #2239. Will keep in mind to move all the logic to the backend in the future, unless we find a good reason for not doing it

frontends/web/src/routes/account/account.tsx
Beerosagos: Yeah, I was just being paranoid :D

frontends/web/src/routes/account/account.tsx
Beerosagos: cool! thanks

PR #2283, opened 2023-10-06
frontend: move SkipForTesting to sidebar

benma:
In watch-only, accounts can be loaded even if no keystore is connected. The debug "skip for testing" software keystore unlock would not be shown, as the app would always route to the portfolio.

We simply move it to the sidebar where it is always accessible.

### DIFF
--- a/frontends/web/src/routes/device/bitbox01/bitbox01.module.css
+++ b/frontends/web/src/routes/device/bitbox01/bitbox01.module.css
@@ -136,14 +136,6 @@
     margin-bottom: 0;
 }
 
-.testingContainer {
-    margin-top: var(--space-default);
-}
-
-.testingContainer > * {
-    margin: 0 auto;
-}
-
 @media screen and (max-width: 640px) {
     .qrcodeContainer {
         width: 50%;
--- a/frontends/web/src/routes/device/components/skipfortesting.tsx
+++ b/frontends/web/src/routes/device/components/skipfortesting.tsx
@@ -16,32 +16,45 @@
 
 import React, { useState } from 'react';
 import { testRegister } from '../../../api/account';
+import { getTesting } from '../../../api/backend';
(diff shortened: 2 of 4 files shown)


frontends/web/src/routes/device/components/skipfortesting.tsx
```diff
@@ -16,32 +16,45 @@
...
-        Skip for Testing
-      </Button>
-    </form>
+    <>
+      <button onClick={() => setDialog(true)}>{title}</button>
+      <Dialog open={dialog} title={title} onClose={() => setDialog(false)}>
+        <form onSubmit={registerTestingDevice}>
+          <PasswordSingleInput
+            type="password"
+            autoFocus
```
benma, line 47: @thisconnect when i open the dialog the input is not auto-focused. any idea why?

thisconnect: lol, the checkbox has focus, this is probably because we have some old weird logic in the dialog that tries to focus the first input element. This was made when we always had focus issues in preact. In react I noticed almost no more focus issues. It's probably time to refactor dialog.tsx. I guess we can just remove the old logic, here: https://github.com/digitalbitbox/bitbox-wallet-app/blob/2c667fd98b8f1261c1d7767e983fb7258e8f11aa/frontends/web/src/components/dialog/dialog.tsx#L83-L101

frontends/web/src/routes/device/bitbox01/bitbox01.module.css
```diff
@@ -136,14 +136,6 @@
...
 }
 
-.testingContainer {
-    margin-top: var(--space-default);
-}
-
-.testingContainer > * {
-    margin: 0 auto;
-}
-
```
thisconnect, line 146: 👍

thisconnect (approved):
tested LGTM 👍

PR #2285, opened 2023-10-07
frontend: type newtxs subscription api

thisconnect:
This types the subscription to listen for incoming transactions.

It is used to trigger native OS notification, test with Qt app.


thisconnect:
rebased

thisconnect:
> What's up with the test again... 😄

it's driving me nuts  🤦

PR #2289, opened 2023-10-09
frontend: improve/add err handling for WC connect attempt & pairing approval

shonsirsha:
alerts user using localised error messages for common errors during
an attempt to connect to a dapp and when approving a pairing request.

Otherwise, alert them using the native/default error message
from WC library.

In both cases, errors get logged.

### DIFF
--- a/frontends/web/src/contexts/WCWeb3WalletProvider.tsx
+++ b/frontends/web/src/contexts/WCWeb3WalletProvider.tsx
@@ -15,6 +15,7 @@
  */
 
 import { ReactNode, useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import { WCWeb3WalletContext } from './WCWeb3WalletContext';
 import { IWeb3Wallet } from '@walletconnect/web3wallet';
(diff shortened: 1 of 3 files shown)


frontends/web/src/contexts/WCWeb3WalletProvider.tsx
```diff
@@ -15,6 +15,7 @@
  */
 
 import { ReactNode, useEffect, useState } from 'react';
+import { t } from 'i18next';
```
thisconnect, line 18: in other places we use the `useTranslation` hook 😇

frontends/web/src/locales/en/app.json
```diff
@@ -1570,6 +1570,7 @@
       "successfullySigned": "Request succesfully signed",
       "walletConnectRequest": "WalletConnect request"
     },
+    "useNewUri": "Please use a new URI",
```
Beerosagos, line 1573: I think it would be nice to have a more descriptive error, something like "This URI has been already used to attempt a connection. Please use a new URI"
Not a strong opinion, though :)

Beerosagos (approved):
LGTM :fire: just a small nit, if you agree

PR #2290, opened 2023-10-09
frontend: adds loading state on dapp connect attempt using WC

shonsirsha:
uses loading state to disable buttons and text input when attempting to connect to a dapp using WC. Especially noticeable for slow connections

Previews below are created using a throttled network (which is very slow), available from chrome dev tools.

[Preview before, connect, cancel buttons, and textfield are interactable. Textfield also gets reset prematurely]:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/2ab4ad3f-010c-4822-bfdb-9df56f5a072b

[Preview after]:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/2e81b6fe-0cd0-4c29-b721-de07d32c87a5


thisconnect:
can you also disable the qr scanner icon while it is loading?

PR #2291, opened 2023-10-10
frontend/hooks: fix permissions hook android bug

Beerosagos:
Recent commit 48eebe9 introduced a new hook to check for camera permissions and inform the user about that, that was causing a frontend exception on Android. This fixes the bug.


thisconnect (approved):
thank you. untested LGTM 👍

PR #2293, opened 2023-10-10
backend/config: persist keystore with keystore name

benma:
We want to display the name of the keystore (e.g. BitBox02 device name) in the sidebar above the accounts derived from that keystore. This is useful for watch-only mode when someone uses multiple keystores.

The BitBox02 keystore name for now is simply the device name. For passphrase users that means that multiple keystores will have the same name. We will not give this any special consideration for now. The user can label the individual accounts, which is enough to disambiguate.


Beerosagos (approved):
LGTM :+1:

PR #2295, opened 2023-10-11
frontend: fix camera on linux

thisconnect:
Regression introduced in 48eebe98e81fa1c026c45e3b9ed1b1f602aae435

That commit added a please-grant-camera-permission warning, in the same component that contains the video element. It uses a new hook to check for camera devices and conditionally renders the warning.

Conditionally showing the warning based on the hook re-renders the component and seems to be the cause for the video to error with:

    DOMException: The play() request was interrupted by a new load request.

This commit is an approach to lift the warning up to the parent component. It seems to work on macOS, but the warning does not go away.
Could be that the event listener on permissionstate object is not working.

### DIFF
--- a/frontends/web/src/hooks/permissions.ts
+++ b/frontends/web/src/hooks/permissions.ts
@@ -41,7 +41,7 @@ export const useDevicePermission = (deviceName: TExperimentalDeviceName) => {
         permissionObject.current = permissionStatus;
         handlePermissionChange();
       });
-  });
+  }, [handlePermissionChange, deviceName]);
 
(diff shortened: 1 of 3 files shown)


thisconnect:
I'll keep this PR open and revert [the ](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2243/commits/48eebe98e81fa1c026c45e3b9ed1b1f602aae435)

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/2296

PR #2301, opened 2023-10-16
frontend/qt: whitelist frontend requests for wallet-connect

Beerosagos:
To proceed with the Walletconnect workflow, we need to allow some external requests originated by the frontend WC library.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -109,10 +109,12 @@ public:
             return;
         }
 
+        auto currentUrl = mainPage->requestedUrl().toString();
+        auto requestedUrl = info.requestUrl().toString();
+
         // We treat the onramp page specially because we need to allow onramp
         // widgets to load in an iframe as well as let them open external links
         // in a browser.
-        auto currentUrl = mainPage->requestedUrl().toString();
         bool onBuyPage = currentUrl.contains(QRegularExpression("^qrc:/buy/.*$"));
         if (onBuyPage) {
             if (info.firstPartyUrl().toString() == info.requestUrl().toString()) {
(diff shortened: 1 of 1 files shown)


thisconnect:
we have 8 other ref's in use:
- Confirm component to store the most recent callback that should be exectued after confirm prompt (yuck)
- Copy component to keep a ref to the textarea element
- AddAccount has a ref to an input element only to focus it
- Receive to keep an array of matching script types 🤔 
- ScanQrVideo to keep a ref to video element
- AccountSummary to keep a timeout integer in order to use it with clearTimeout later
- SetDeviceName a ref to input field for form validation

frontends/qt/main.cpp
```diff
@@ -124,6 +127,21 @@ class RequestInterceptor : public QWebEngineUrlRequestInterceptor {
             return;
         }
 
+        // All the requests originated in the wallet-connect section are allowed, as they are needed to
+        // load the Dapp logos and it is not easy to filter out non-images requests.
+        bool onWCPage = currentUrl.contains(QRegularExpression("^qrc:/account/[^\/]+/wallet-connect/.*$"));
+        if (onWCPage) {
+          info.block(false);
```
benma, line 134: are the ` info.block(false);` actually needed? not sure what they do.

frontends/qt/main.cpp
```diff
@@ -124,6 +127,21 @@ class RequestInterceptor : public QWebEngineUrlRequestInterceptor {
...
+          return;
+        }
+
+        // Needed for the wallet connect workflow.
+        bool VerifyWCRequest = requestedUrl.contains(QRegularExpression("^https://verify.walletconnect.com/.*$"));
```
benma, line 139: use `\.` for dots, otherwise verify-walletconnect.com etc. is also accepted

benma (changes requested):

frontends/qt/main.cpp
Beerosagos: Right, they seem to be useless

frontends/qt/main.cpp
Beerosagos: ouch! good catch, thanks

Beerosagos:
Pushed a new update PTAL @benma :pray:

benma (approved):
utACK

PR #2302, opened 2023-10-16
ci: use latest nodejs 18.x lts version for macos

thisconnect:
This uses latest nodejs LTS 18.x version used in macOS GitHub ci.

Before that we set specifc node version but as soon as the version is not available anymore ci fails. Changed to just request major version.

Linux ci does not have specific nodejs version set but installs nodejs in scripts/docker_install.sh


thisconnect:
I dropped both minor and patch version. but if you prefer I guess we could keep the minor version.

PR #2305, opened 2023-10-19
frontend/android: block unexpected external requests

Beerosagos:
Commit 3b62772 introduced a base url to be used in external requests to comply with Moonpay frame-ancestors content security policy. It somehow broke the update of the android webview location and made impossible to block external requests on the base of the page that originated them.

This commit implements a workaround that updates the location after navigation related events and blocks unexpected requests as already done in Qt.

Please note:
- I'm opening this PR towards staging-walletconnect since it will probably be soon merged into master. This way we avoid being forced to merge master into that branch before and then adding the rules to allow the external requests for dapps.
- Reverting commit 3b62772 solves the navigation issue in a cleaner way, but I could not find a solution to the Moonpay iframe block, this is why I used this workaround.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -42,6 +42,7 @@ import androidx.lifecycle.Observer;
 import androidx.lifecycle.ViewModelProviders;
 
 import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -69,6 +70,8 @@ public class MainActivity extends AppCompatActivity {
     private PermissionRequest webViewpermissionRequest;
 
     GoService goService;
+
+    private String location = "";
 
     // Connection to bind with GoService
     private ServiceConnection connection = new ServiceConnection() {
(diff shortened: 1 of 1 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -226,20 +229,58 @@ public void run() {
...
+                            if (pattern.matcher(location).matches()) {
+                                return super.shouldInterceptRequest(view, request);
+                            }
+                        }
+
+                        String domain = request.getUrl().getHost();
+                        if (domain != null) {
+                            // allow if domain is listed
+                            patterns = new ArrayList<>();
+                            patterns.add(Pattern.compile("^(.*\\.)?verify\\.walletconnect\\.com$"));
```
benma, line 271: in Qt the pattern looks a bit different https://github.com/digitalbitbox/bitbox-wallet-app/pull/2301/files#diff-c5a5346e0c9e1d112c810b5097daea72abb6bd88406cefb73d52de55c9836f2cR137 - `https://verify\...`. Why is it different here?

benma (approved):
utACK w/ a question

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
Beerosagos: No reason, I didn't notice that in Qt we were not allowing subdomains. I updated the code here to have the same behavior

benma (approved):
LGTM

PR #2308, opened 2023-10-24
frontend: group accounts by keystore in sidebar and account settings

benma:
Group shows the keystore name.

The BitBox02 keystore name for now is simply the device name. For passphrase users that means that multiple keystores will have the same name. We will not give this any special consideration for now. The user can label the individual accounts, which is enough to disambiguate.

### DIFF
--- a/backend/config/accounts.go
+++ b/backend/config/accounts.go
@@ -104,16 +104,25 @@ func (cfg AccountsConfig) Lookup(code accountsTypes.Code) *Account {
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


Beerosagos (approved):
tACK with a single keystore active. Is there a way to test with more than one?

Beerosagos:
> tACK with a single keystore active. Is there a way to test with more than one?

Tested setting to true https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/accounts.go#L753

LGTM

PR #2307, opened 2023-10-24
frontend: type keystores

thisconnect:

PR #2311, opened 2023-10-25
Keystore event fix

Beerosagos:
Fixes an event transmission lag that caused an unexpected error (discovered in #2307) and renames a misleading backend method. See commit messages for details.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -931,6 +931,7 @@ func (backend *Backend) uninitAccounts() {
 			backend.onAccountUninit(account)
 		}
 		account.Close()
+		backend.emitAccountsStatusChanged()
 	}
 	backend.accounts = []accounts.Interface{}
 }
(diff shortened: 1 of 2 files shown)


backend/accounts.go
```diff
@@ -931,6 +931,7 @@ func (backend *Backend) uninitAccounts() {
 			backend.onAccountUninit(account)
 		}
 		account.Close()
+		backend.emitAccountsListChanged()
```
benma, line 934: at this stage `backend.accounts` has not changed at all though, so why fire an event here?

benma (commented):
> n keystore deregister the accounts linked to the keystore were closed
and sent statusChanged events to the frontend + a final `accounts` event
to notify the change in the accounts number. The lag between the status
and the accounts events caused the execution of calls towards endpoint
of already closed accounts, which generated unexpected errors.

I have a hard time understanding this. Which `statusChanged` event exactly?

Beerosagos:
> I have a hard time understanding this. Which `statusChanged` event exactly?
This https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/coins/btc/account.go#L457

PR #2319, opened 2023-10-27
frontend: full screen qr-video for mobile

NicolaLS:
The video element in te ScanQRVideo component did not have a mimetype in the uri of the poster attribute which caused Qt to display a warning. Using `data:image/svg+xml` fixes this. However this makes the video fit into the size of the placeholder svg.

TODO: 
- [ ] either scale the svg placeholders size, or change the qr videos content box if possible
- [ ] there might be a vertical/horizontal scroll on mobile

~~Using the mimetype `image/svg+xml` caused the camera view to be distorted/sqished. Using `image` has the desired effect.
Also added `utf8` to be more explicit, and removed unnecessary encoding (`%22`, `%20`)~~

### DIFF
--- a/frontends/web/src/routes/account/send/components/inputs/scan-qr-video.module.css
+++ b/frontends/web/src/routes/account/send/components/inputs/scan-qr-video.module.css
@@ -9,9 +9,18 @@
 
 .qrVideo {
   width: 100%;
+  object-fit: cover;
   margin-bottom: var(--spacing-default);
 }
 .qrVideo + :global(.scan-region-highlight) svg {
   stroke: var(--color-darkyellow) !important;
   stroke-width: 3 !important;
+}
+
+@media (max-width: 768px) {
+  .qrVideo {
+    height: 100vh;
+    object-fit: cover;
+    margin-bottom: 0;
+  }
 }
--- a/frontends/web/src/routes/account/send/components/inputs/scan-qr-video.tsx
+++ b/frontends/web/src/routes/account/send/components/inputs/scan-qr-video.tsx
@@ -44,7 +44,7 @@ export const ScanQRVideo = ({
       <video
         className={style.qrVideo}
         ref={videoRef}
-        poster="%3Csvg%20xmlns=%22http://www.w3.org/2000/svg%22%20width=%2264%22%20height=%2248%22%3E%3C/svg%3E"
+        poster="data:image/svg+xml,%3Csvg%20xmlns=%22http://www.w3.org/2000/svg%22%3E%20width=%2264%22%20height=%2248%22%3C/svg%3E"
       />
     </>
   );
--- a/frontends/web/src/components/dialog/dialog.module.css
+++ b/frontends/web/src/components/dialog/dialog.module.css
@@ -248,6 +248,26 @@
 
 @media (max-width: 768px) {
 
+    .header.fullscreenOnMobile {
+        position: fixed;
+        top: 0;
+        left: 0;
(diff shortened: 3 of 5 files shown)


frontends/web/src/routes/account/send/components/inputs/scan-qr-video.tsx
```diff
@@ -44,7 +44,7 @@ export const ScanQRVideo = ({
       <video
         className={style.qrVideo}
         ref={videoRef}
-        poster="%3Csvg%20xmlns=%22http://www.w3.org/2000/svg%22%20width=%2264%22%20height=%2248%22%3E%3C/svg%3E"
+        poster="data:image,%3Csvg xmlns='http://www.w3.org/2000/svg' width='64' height='48'%3E%3C/svg%3E"
```
thisconnect, line 47: nit: @NicolaLS I think `data:image` is fine, but I don't know how convert the original svg to your version. Can we not just use encodeURI or encodeURIComponent?

NicolaLS: Yes, I'm sorry the svg is actually wrong...`encodeURIComponent` is what we want. With the correct uri in the poster attribute the video is only 48 high as the default `object-fit: fill` t[his css property ](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit)is also what caused the distortion with the incorrect uri.

It looks like `object-fit: none` also does not do what we want:

Screenshot_20231101_131658_BitBoxApp DEBUG

NicolaLS:
@thisconnect 
1. The svg uri was wrong, I amended the commit and used `encodeURIComponent` on this svg `<svg xmlns="http://www.w3.org/2000/svg" width="64" height="48"></svg>`
2. I did not figure out what is causing a vertical scroll without adding [this](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2319/files#diff-0662a9a68e92c57af80a8c93ec88d616038889d3a5042d0b3a61f7cbba2563baR27) hack, please let me know what you think..(I tried using `max-height: 100vh` on the video and its container without success.
3. It looks good on mobile now, when you scale the qt application to a mobile size the `scan-highlight` does not get rescaled and it looks bad, same for `webdev`
4. For some reason the whole dialog/video scan is bigger on qt-linux now..does not look bad and maybe it is also a side effect from actually using the correct mimetype

The design can probably be improved too, this is how it looks 
Screenshot_20231101_213325_Permission controller

Screenshot_20231101_213341_BitBoxApp DEBUG
on mobile:

NicolaLS:
rebased @thisconnect PTAL

frontends/web/src/routes/account/send/components/inputs/scan-qr-video.module.css
```diff
@@ -9,9 +9,18 @@
...
 }
 .qrVideo + :global(.scan-region-highlight) svg {
   stroke: var(--color-darkyellow) !important;
   stroke-width: 3 !important;
+}
+
+@media (max-width: 768px) {
+  .qrVideo {
+    height: 100vh;
+    object-fit: cover;
```
thisconnect, line 23: you can remove`object-fit: cover;` here, it is already defined on line 12

thisconnect (changes requested):
nice 👍  with one nit. After that I'd like to do some more testing.

thisconnect:
superseded by https://github.com/digitalbitbox/bitbox-wallet-app/pull/2610

PR #2320, opened 2023-10-30
frontend: fix overflowing WC CTA text when balance is very long

shonsirsha:
improving the account page UI by collapsing the buttons when there's not enough space.

This is probably quite an edge case, and the changes in this PR shouldn't affect anything else except for: **the account page that's compatible with WC on desktop.**

Before:
99gnmm

After:
xx9927

Mobile (no difference):

x92


thisconnect (approved):
tested LGTM 👍

PR #2322, opened 2023-10-30
frontend: fix route for `onReceiveCTA`

shonsirsha:
fixing a small bug, regression happened here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/2316/commits/3e2018d67774a83325c226ae6e3c7ba20b3d14bf

balanceList is of type `IBalance[]`, which is:

```js
IBalance {
    hasAvailable: boolean;
    available: IAmount;
    hasIncoming: boolean;
    incoming: IAmount;
}
```

thus `/account/balanceList[0]/receive` is an invalid route.

### DIFF
--- a/frontends/web/src/routes/account/info/buyReceiveCTA.tsx
+++ b/frontends/web/src/routes/account/info/buyReceiveCTA.tsx
@@ -45,7 +45,7 @@ export const BuyReceiveCTA = ({ code, unit, balanceList, exchangeBuySupported = 
         route('accounts/select-receive');
         return;
       }
-      route(`/account/${balanceList[0]}/receive`);
+      route(`/account/${code}/receive`);
     }
   };
 


frontends/web/src/routes/account/info/buyReceiveCTA.tsx
```diff
@@ -45,7 +45,7 @@ export const BuyReceiveCTA = ({ code, unit, balanceList, exchangeBuySupported =
         route('accounts/select-receive');
         return;
       }
-      route(`/account/${balanceList[0]}/receive`);
+      route(`/account/${code}/receive`);
```
thisconnect, line 48: We could have same issue in other places.
I think we never want `undefined` in template literals could that be an eslint or ts rule somehow?

PR #2326, opened 2023-10-31
backend: fix eip1559 signTransaction bug return raw tx with sig

Tomasvrba:


Tomasvrba:
testing eth_signTransaction with https://react-app.walletconnect.com/ works now
@shonsirsha
sorry, bug was my bad, the backend was returning the unsigned tx :O 🤦‍♂️

PR #2327, opened 2023-10-31
Add watchonly setting to accounts & handle watchonly account loading/unloading

benma:
Enabling the watchonly setting on an account leaves the account loaded when the keystore is removed, and the account is loaded at app launch too.

Watch-only is opt-in per account for now - in the future we might switch to enabling watch-only by default.

**Not done in this PR:** handling the case when the user performs an action on a watchonly account that requires the keystore, like verifying an address or transaction. This will follow in another PR.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -487,6 +487,46 @@ func (backend *Backend) RenameAccount(accountCode accountsTypes.Code, name strin
 	return nil
 }
 
+// AccountSetWatch sets the account's persisted watch flag to `watch`. Set to `true` if the account
+// should be loaded even if its keystore is not connected.
+// If `watch` is set to `false`, the account is unloaded and the frontend notified.
+func (backend *Backend) AccountSetWatch(accountCode accountsTypes.Code, watch bool) error {
+	err := backend.config.ModifyAccountsConfig(func(accountsConfig *config.AccountsConfig) error {
+		acct := accountsConfig.Lookup(accountCode)
+		if acct == nil {
+			return errp.Newf("Could not find account %s", accountCode)
+		}
+		acct.Watch = &watch
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	defer backend.accountsAndKeystoreLock.Lock()()
+
+	// ETH tokens inherit the Watch-flag from the parent ETH account.
+	// If the watch status of an ETH account was changed, we update it for its tokens as well.
+	//
+	// This ensures that removing a keystore after setting an ETH account including tokens to
+	// watchonly results in the account *and* the tokens remaining loaded.
+	if acct := backend.accounts.lookup(accountCode); acct != nil {
+		if acct.Config().Config.CoinCode == coinpkg.CodeETH {
+			for _, erc20TokenCode := range acct.Config().Config.ActiveTokens {
+				erc20AccountCode := Erc20AccountCode(accountCode, erc20TokenCode)
+				if tokenAcct := backend.accounts.lookup(erc20AccountCode); tokenAcct != nil {
+					tokenAcct.Config().Config.Watch = &watch
+				}
+			}
+		}
+	}
+
+	if !watch {
+		backend.initAccounts(false)
+		backend.emitAccountsStatusChanged()
+	}
+	return nil
+}
+
 // addAccount adds the given account to the backend.
 // The accountsAndKeystoreLock must be held when calling this function.
 func (backend *Backend) addAccount(account accounts.Interface) {
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -502,6 +502,10 @@ func (backend *Backend) addAccount(account accounts.Interface) {
 
 // The accountsAndKeystoreLock must be held when calling this function.
 func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *config.Account) {
+	if backend.accounts.lookup(persistedConfig.Code) != nil {
+		// Do not create/load account if it is already loaded.
+		return
+	}
 	var account accounts.Interface
 	accountConfig := &accounts.AccountConfig{
 		Config:      persistedConfig,
@@ -933,9 +937,10 @@ func (backend *Backend) updatePersistedAccounts(
 }
 
 // The accountsAndKeystoreLock must be held when calling this function.
-func (backend *Backend) initAccounts() {
+// if force is true, all accounts are uninitialized first, even if they are watch-only.
+func (backend *Backend) initAccounts(force bool) {
 	// Since initAccounts replaces all previous accounts, we need to properly close them first.
-	backend.uninitAccounts()
+	backend.uninitAccounts(force)
 
 	backend.initPersistedAccounts()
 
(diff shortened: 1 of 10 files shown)


backend/accounts.go
```diff
@@ -478,6 +487,46 @@ func (backend *Backend) RenameAccount(accountCode accountsTypes.Code, name strin
 	return nil
 }
 
+// AccountSetWatch sets the account's persisted watch flag to `watch`. Set to `true` if the account
+// should be loaded even if its keystore is not connected.
+// If `watch` is set to `false`, the account is unloaded and the frontend notified.
+func (backend *Backend) AccountSetWatch(accountCode accountsTypes.Code, watch bool) error {
```
benma, line 493: not unit tested yet, i plan on adding one at some point

backend/accounts.go
```diff
@@ -478,6 +487,46 @@ func (backend *Backend) RenameAccount(accountCode accountsTypes.Code, name strin
...
+	defer backend.accountsAndKeystoreLock.Lock()()
+
+	// ETH tokens inherit the Watch-flag from the parent ETH account.
+	// If the watch status of an ETH account was changed, we update it for its tokens as well.
+	//
+	// This ensures that removing a keystore after setting an ETH account including tokens to
+	// watchonly results in the account *and* the tokens remaining loaded.
+	if acct := backend.accounts.lookup(accountCode); acct != nil {
+		if acct.Config().Config.CoinCode == coinpkg.CodeETH {
+			for _, erc20TokenCode := range acct.Config().Config.ActiveTokens {
```
benma, line 514: @jadzeidan fyi the erc20-as-their-own-accounts model is really quite bad for the code :see_no_evil: special cases everywhere. it would be very nice (from the code point of view) if we could bundle the tokens into the parent ETH account. could also reduce the confusion about that the fees for an erc20 token tx are coming from a "different" account.

thisconnect:
@Beerosagos PTAL :)

Beerosagos:
> @Beerosagos PTAL :)

doing it rn :)

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -203,7 +203,7 @@ class Sidebar extends Component<Props> {
           {
             accountsByKeystore.map(keystore => (<React.Fragment key={keystore.keystore.rootFingerprint}>
               <div className={style.sidebarHeaderContainer}>
-                <span className={style.sidebarHeader} hidden={!keystores?.length}>
+                <span className={style.sidebarHeader} hidden={!keystore.accounts.length}>
                   {t('sidebar.accounts')} - {keystore.keystore.name}
```
Beerosagos, line 207: Just a nit: I find the `ACCOUNT - <keystore name>` notation a bit confusing, as it seems to suggest that  <keystore name> is the name of an account. I'd prefer `<keystore name> - ACCOUNT` or even removing the ACCOUNT part completely (maybe replacing it with the root fingerprint? smth like `<keystore name> [fingerprint]`. 
Anyway, this can be discussed and possibly changed in a future PR as well.

cc @jadzeidan

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -101,6 +101,16 @@ class ManageAccounts extends Component<Props, State> {
               this.toggleAccount(account.code, !active)
                 .then(() => event.target.disabled = false);
             }} />
+          Watchonly:
```
Beerosagos, line 104: Should this be localized?

Beerosagos (commented):
I'd like to test a bit more, but I'll start sharing a few comments

frontends/web/src/routes/settings/manage-accounts.tsx
benma: this is just a placeholder for the frontend guys to clean up - i don't expect the final version to contain this word.

backend/config/accounts.go
```diff
@@ -36,7 +36,15 @@ type Account struct {
...
 	// and not be shown in 'Manage accounts', because the account is unused (has no transaction
 	// history). This is used to facilitate automatic discovery of used accounts.
-	HiddenBecauseUnused   bool                   `json:"hiddenBecauseUnused"`
+	HiddenBecauseUnused bool `json:"hiddenBecauseUnused"`
+	// Watch indicates if the account should be loaded even if its keystore is not connected.
+	//
+	// If false, the account is only displayed if the keystore is connected. If true, it is loaded
+	// and displayed when the app launches.
+	//
+	// If nil, it is considered false, but will be set to true the first time the keystore is
```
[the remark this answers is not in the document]
benma: it the first version of this code it was, when we wanted to enabling it by default. we changed to opt-in. will fix the comment

backend/accounts.go
```diff
@@ -758,39 +758,54 @@ func (backend *Backend) persistETHAccountConfig(
 
 // The accountsAndKeystoreLock must be held when calling this function.
 func (backend *Backend) initPersistedAccounts() {
-	if backend.keystore == nil {
-		return
-	}
 	// Only load accounts which belong to connected keystores.
```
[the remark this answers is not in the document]
benma: done

backend/config/accounts.go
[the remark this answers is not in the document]
benma: done

backend/accounts.go
```diff
@@ -758,39 +758,54 @@ func (backend *Backend) persistETHAccountConfig(
...
-			for _, cfg := range account.SigningConfigurations {
-				if !backend.keystore.SupportsAccount(coin, cfg.ScriptType()) {
-					continue outer
+
+		// Watch-only accounts are loaded regardless, and if later e.g. a BitBox02 BTC-only is
+		// inserted with the same seed as a Multi, we will need to catch that mismatch when the
+		// keystore will be used to e.g. display an Ethereum address etc.
+		//
+		// If the keystore is connected and the account is not marked as watch-only, it does not
+		// make sense to load it.
```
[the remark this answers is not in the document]
benma: > you mean that it doesn't make sense to load an eth account when connecting a btc-only keystore?

yeah - unless that account was marked as watch-only before (which can happen if one uses a multi edition the same seed first). if it is marked watch-only, it should be shown regardless.

frontends/web/src/components/sidebar/sidebar.tsx
benma: For clarity, this is the title of the group header above the accounts (it says ACCOUNTS, not ACCOUNT).

Anyway, I agree the "ACCOUNTS" part could be dropped.

About the fingeprint: we had discussed that fingeprints are too technical/noisy to show by default. that discussion was in relation to  disambiguate keystores from the same wallet but different passphrases, but decided that users can just rename the individual accounts.

benma:
@Beerosagos thanks for the nice review! I addressed all comments, PTAL

backend/accounts.go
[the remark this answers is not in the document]
Beerosagos: Thanks! I still find the last two lines a bit confusing, I'd either remove them (as the new comment you added is very clear, IMO) or change to something like ` If the connected keystore doesn't support the account, and it is not marked as watch-only, it does not make sense to load it.` 

This is just a nit tho, feel free to keep as is.

frontends/web/src/components/sidebar/sidebar.tsx
Beerosagos: Ah then it is just in the Italian translation, I guess. Plural english words are often transformed into their singular version: in Italian we have 'ACCOUNT' instead of accounts.

Beerosagos (approved):
tACK! Very nice :fire:

backend/accounts.go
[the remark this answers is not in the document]
benma: I fully agree, these two lines are very confusing lol. I removed them.

PR #2331, opened 2023-11-02
eip1559: add ethereum eip1559 type 2 TXs to app backend

Tomasvrba:


Tomasvrba:
This whole check can probably be removed since I changed BBApp signing to only do 1559 Transaction, but I left it for now as a sanity check:
https://github.com/digitalbitbox/bitbox-wallet-app/pull/2331/files#diff-3dc4499a5e2ffebff22d3f2c24392c2d001f4add386b892ac3394c4f042a6c03R457-R484

Tomasvrba:
CI depends on: https://github.com/digitalbitbox/bitbox02-api-go/pull/88

backend/coins/eth/account.go
```diff
@@ -599,15 +599,23 @@ func (account *Account) newTx(args *accounts.TxProposalArgs) (*TxProposal, error
 			}
 		}
 	}
-	tx := types.NewTransaction(account.nextNonce,
-		*message.To,
-		message.Value, gasLimit, suggestedGasPrice, message.Data)
+	txData := &types.DynamicFeeTx{
```
benma, line 602: I think you can't just switch to this because it does not work with users that did not upgrade the BitBox02, right?

Can only use this if the firmware version is a at least 9.16.0 (or whatever version it ends up being released in)

Tomasvrba: yeah, correct, thought of that too, how do you want to handle this if not a mandatory fw upgrade for 1559?

benma: use eip1559 only if fw is at least 9.16.0, otherwise use legacy.

Tomasvrba:
@benma squashed and force pushed the new changes since it's been a while since you reviewed this first anyway. Should be all good to go, tested both legacy and eip1559 on mainnet

backend/coins/eth/account.go
```diff
@@ -610,15 +605,41 @@ func (account *Account) newTx(args *accounts.TxProposalArgs) (*TxProposal, error
 			}
 		}
 	}
-	tx := types.NewTransaction(account.nextNonce,
-		*message.To,
-		message.Value, gasLimit, suggestedGasPrice, message.Data)
+
+	keystore, err := account.Config().ConnectKeystore()
```
benma, line 609: hmmmm I think this is called when modifying anything in the Send form, e.g. when pasting a recipient address, changing the amount, etc.

In watch-only mode, this means as soon as you do anything in the Send form, a 'Connect your BitBox02' prompt will appear, right? that would be different to how it was before and how it is in bitcoin, where the prompt only happens when one clicks to sign/send.

Tomasvrba: hmm, yeah, good catch, thoughts how to handle it? 
seems strange to have the send screen available in watch only mode? Maybe the send button should only be enabled when the device is connected? 
Or do you want to handle it here in the send screen instead?

benma: As discussed in private chat - we will prompt to connect the keystore when going to the send scree as you suggested, but only for ETH accounts. This is implemented here:

https://github.com/digitalbitbox/bitbox-wallet-app/pull/2456/

benma (changes requested):
please add two changelog entries (EIP1559, new fee estimator)

Tomasvrba:
addressed review

backend/coins/eth/account.go
```diff
@@ -735,9 +738,10 @@ func (account *Account) FeeTargets() ([]accounts.FeeTarget, accounts.FeeTargetCo
 	return feeTargets, accounts.DefaultFeeTarget
 }
 
-// gasPrice returns the currently suggested gas price for the given fee target, or a custom gas
-// price if the fee target is `FeeTargetCodeCustom`.
-func (account *Account) gasPrice(args *accounts.TxProposalArgs) (*big.Int, *big.Int, error) {
+// gasFees returns the currently suggested maxFeePerGas and maxPriorityFee for the given fee target, or a custom fee
+// if the fee target is `FeeTargetCodeCustom`. The custom fee sets both maxFeePerGas and maxPriorityFee to the same value.
+// TODO: The UI should have and advanced setting to allow the user to set maxFeePerGas and maxPriorityFee separately.
```
benma, line 743: what do Rabby and MEW do with custom fees? Two separate fields? sounds like very difficult UX.

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -603,24 +603,33 @@ func (etherScan *EtherScan) FeeTargets(ctx context.Context) ([]*ethtypes.FeeTarg
 	if err != nil {
 		return nil, err
 	}
-	baseFeeWei := int64(baseFee * 1e9)
-	// Conversion from 10x Gwei to Wei.
+	baseFeeWei := big.NewInt(int64(baseFee * 1e9))
```
benma, line 606: idk if it is a problem but would `new(big.Int).Mul(big.NewInt(int64(baseFee)), factor)` be safer?

backend/coins/eth/rpcclient/rpcclient.go
```diff
@@ -51,6 +51,8 @@ type Interface interface {
 	// SuggestGasPrice retrieves the currently suggested gas price to allow a timely
 	// execution of a transaction.
 	SuggestGasPrice(ctx context.Context) (*big.Int, error)
+	// FeeTargets returns EIP-1559 compatible priorities (maxFeePerGas + baseFee)
+	// If not be reachable fallback to only one priority, estimated by the eth_gasPrice RPC endpoint.
```
benma, line 55: Remove this line - the fallback is not implemented by the rpcclient (i.e. etherscan).

benma (changes requested):

backend/coins/eth/rpcclient/rpcclient.go
benma: i just realized that this interface is only for RPC methods, and etherscan can provide the eth full node RPC methods, or any other full node can. 

FeeTargets() is not an RPC method but solely a etherscan feature. 

As we only use Etherscan right now for RPC, this did not become obvious :P

We can just leave it like this for now, too much hassle to restructure the code to make it "proper" when this RPC abstraction is not even used at the moment. Could consider just deleting the interface (but not in this PR).

backend/coins/eth/account.go
Tomasvrba: From what I saw, MEW doesn't let you set a custom fee at all, and Rabby does the same as this PR currently does, just one custom fee field which sets both maxFeePerGas and maxPriorityFeePerGas to that same custom fee value. I guess it's only MetaMask really from the most commonly used wallets that lets you set the fees separately in advanced mode.

I'll delete the TODO and we can revisit this if there's ever demand for separate custom fee fields

backend/coins/eth/etherscan/etherscan.go
Tomasvrba: can't hurt for consistency I suppose, I'll change it

backend/coins/eth/rpcclient/rpcclient.go
Tomasvrba: yeah, I know, I didn't like adding it to the interface and under `// ----- RPC node proxy methods follow` when it's not an RPC but a custom etherscan feature, I meant to leave a comment about that somewhere but I guess I didn't

I'll leave it but make a comment there for future reference

benma (approved):
LGTM - well done, and thanks for your patience :)

Please squash and then it's time to finally merge :tada:

Tomasvrba:
> LGTM - well done, and thanks for your patience :)
> 
> Please squash and then it's time to finally merge 🎉

rebased and squashed

PR #2332, opened 2023-11-03
Specify 'platform' parameter to run docker build on Apple Silicon

rootzoll:
With the current `dockerdev.sh` I was not able to get the `amd64` docker image started on my M2 arm based Mac. By adding the `--platform "linux/amd64"` parameter I was able to enforce virtualization with docker and get it running. So I recommed to add it for future developers using a Mac arm based laptop.

On other machines this additional parameter should not change anything - but I was not able to test it.


benma (approved):
thanks!

PR #2333, opened 2023-11-05
Custom block explorer

NicolaLS:
This PR implements enabling a custom block explorer for the bitcoin config, since it is implemented as an interface it will be easy to use it for other coin configs too.

**Non-Whitelisted URLs**
I decided to approach this problem by implementing a confirmation dialog (thanks to [this](https://daveteu.medium.com/react-custom-confirmation-box-458cceba3f7b) ) we might want to use this approach instead of the deprecated `alertUser` and `confirmation` but this should be done in another PR.

Let me know if you think there is a better approach (e.g selecting from a predefined whitelisted list, or handling the warning differently)

Should I add a commit that adds popular bitcoin blockexplorers like mempool.space, and others ? This would give the user still complete freedom to choose but users that stick with common explorers don't see the warning.

**Restart the app**
I wasted a lot of time on trying to understand how to get the frontend to reload the `account` by calling the `/accounts` endpoint again..I could not figure it out so I added the _restart the app_ message to the setting that. We should think about making it easy for the settings to trigger config/account refreshing but this can also be another PR.

**using bitcoinConfig instead of explorerConfig**
I noticed that I need to [spread out](https://github.com/digitalbitbox/bitbox-wallet-app/compare/master...NicolaLS:bitbox-wallet-app:custom-block-explorer?expand=1#diff-7a4636f83354958dfaf38414670500261eaa14fdad5b3499925067c7cc0e67a8R59) the bitcionConfig because the `setConfig()` function does not _merge/update_ the config but overwrites it for nested config as is the case in `backend.btc.XX` `backend.tbtc.XXX` maybe we should refactor this function so we can be less verbose elsewhere.

### DIFF
--- a/backend/config/blockexplorer.go
+++ b/backend/config/blockexplorer.go
@@ -1,0 +1,44 @@
+package config
+
+import "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
+
+// BlockExplorer interface that different coin configurations can implement.
+// It defines the Enabled() method which returns true if the user enabled the
+// option to use a custom blockexplorer in the settings. The ExplorerURL() method
+// returns the configured URL or a defined default as a string.
+type BlockExplorer interface {
+	Enabled()
+	ExplorerURL()
+}
+
+// The BlockExplorerConfig struct is used by the Backend configuration.
+type BlockExplorerConfig struct {
+	UseExplorer bool `json:"useCustomBlockExplorer"`
+	// Thies field is filled with a default when the config is created
+	// the default needs to be the placeholder value specified in the frontend.
+	ExplorerURL string `json:"explorerURL"`
+}
+
+// Implement custom blockexplorer for Bitcoin
+
+const defaultTestnet = "https://blockstream.info/testnet/tx/"
+const defaultBitcoin = "https://blockstream.info/tx/"
+
+func (btcCfg btcCoinConfig) Enabled() bool {
+	return btcCfg.BlockExplorer.UseExplorer
+}
+
+func (btcCfg btcCoinConfig) ExplorerURL() string {
+	if btcCfg.Enabled() {
+		return btcCfg.BlockExplorer.ExplorerURL
+	}
+	url := ""
+	switch btcCfg.Code {
+	case coin.CodeBTC:
+		url = defaultBitcoin
+	case coin.CodeTBTC:
+		url = defaultTestnet
+
+	}
+	return url
+}
--- a/backend/config/config.go
+++ b/backend/config/config.go
@@ -41,7 +41,9 @@ func (s *ServerInfo) String() string {
 
 // btcCoinConfig holds configurations specific to a btc-based coin.
 type btcCoinConfig struct {
-	ElectrumServers []*ServerInfo `json:"electrumServers"`
+	Code            coin.Code
+	ElectrumServers []*ServerInfo       `json:"electrumServers"`
+	BlockExplorer   BlockExplorerConfig `json:"explorer"`
 }
 
 // ETHTransactionsSource  where to get Ethereum transactions from. See the list of consts
@@ -153,6 +155,7 @@ func NewDefaultAppConfig() AppConfig {
 			DeprecatedEthereumActive: true,
 
 			BTC: btcCoinConfig{
+				Code: coin.CodeBTC,
 				ElectrumServers: []*ServerInfo{
 					{
 						Server:  "btc1.shiftcrypto.io:443",
@@ -165,8 +168,13 @@ func NewDefaultAppConfig() AppConfig {
 						PEMCert: shiftRootCA,
 					},
 				},
+				BlockExplorer: BlockExplorerConfig{
+					UseExplorer: false,
+					ExplorerURL: "https://blockstream.info/tx/",
+				},
 			},
 			TBTC: btcCoinConfig{
+				Code: coin.CodeTBTC,
 				ElectrumServers: []*ServerInfo{
 					{
 						Server:  "tbtc1.shiftcrypto.io:443",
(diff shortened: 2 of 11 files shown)


backend/config/blockexplorer.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type BlockExplorerConfig struct {
+	UseExplorer bool `json:"useCustomBlockExplorer"`
+	// Thies field is filled with a default when the config is created
+	// the default needs to be the placeholder value specified in the frontend.
+	ExplorerURL string `json:"explorerURL"`
+}
+
+// Implement custom blockexplorer for Bitcoin
```
NicolaLS, line 17: maybe `blockexplorer.go` should only define the interface and the implementation should live in the respective coin directory ? e.g. `coins/btc/blockexplorer.go` ?

backend/config/config.go
```diff
@@ -165,8 +168,13 @@ func NewDefaultAppConfig() AppConfig {
 						PEMCert: shiftRootCA,
 					},
 				},
+				BlockExplorer: BlockExplorerConfig{
+					UseExplorer: false,
+					ExplorerURL: "https://blockstream.info/tx/",
```
NicolaLS, line 173: can be empty too as the coin implementation _should_ define a default. but I guess it is fine to also have the defaults here ? (hmm might be annoying for maintainability in case this is refactored later)

frontends/web/src/components/anchor/anchor.tsx
```diff
@@ -39,25 +41,49 @@ type TProps = {
...
 }: TProps) => {
+  const { t } = useTranslation();
+  const [ getConfirmation, Confirmation ]: TUseConfirm = useConfirm();
+
+  const onReject = async (reason: string) => {
+    if (reason.indexOf('Blocked') === -1) {
+      return;
+    }
+
+    // TODO: i18n title, message
```
NicolaLS, line 65: remove this (done)

frontends/web/src/routes/settings/components/advanced-settings/explorer-dialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const handleSetExplorerButton = async () => {
+    if (!bitcoinConfig?.explorer || explorerURL === undefined) {
+      // FIXME: this looks like we avoid handeling a problem here
+      return;
+    }
+    const explorer = bitcoinConfig.explorer;
+    explorer.explorerURL = explorerURL.trim();
+
+    // NOTE: I guess its good to pingthe server to see if it works..is socksProxyCheck ok for that ?
```
NicolaLS, line 55: I actually just left this in here after copy/pasting from `enable-proxy` component, check if this is correct and remove comment (it does work)

frontends/web/src/routes/settings/components/advanced-settings/explorer-dialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+  useEffect(() => {
+    if (bitcoinConfig?.explorer) {
+      setExplorerURL(bitcoinConfig.explorer.explorerURL);
+    }
+  }, [bitcoinConfig]);
+
+
+  const handleSetExplorerButton = async () => {
+    if (!bitcoinConfig?.explorer || explorerURL === undefined) {
+      // FIXME: this looks like we avoid handeling a problem here
```
NicolaLS, line 49: remove this, it is done in other components too should not be a problem

frontends/web/src/utils/request.ts
```diff
@@ -20,7 +20,6 @@ import { alertUser } from '../components/alert/Alert';
 import { call } from './transport-qt';
 import { androidCall } from './transport-android';
 import { runningInAndroid, runningInQtWebEngine } from './env';
-
```
NicolaLS, line 23: fix this

NicolaLS (commented):
some comments

NicolaLS:
actually the confirm dialog does not work on qt linux (probably android too) which is why this is a draft now

frontends/web/src/components/anchor/anchor.tsx
```diff
@@ -39,25 +41,48 @@ type TProps = {
...
+  const [ getConfirmation, Confirmation ]: TUseConfirm = useConfirm();
+
+  const onReject = async (reason: string) => {
+    if (reason.indexOf('Blocked') === -1) {
+      return;
+    }
+
+    const status = await getConfirmation(t('blockedOpen.title'), t('blockedOpen.message'));
+    if (status) {
+      window.open(href, '_blank');
```
thisconnect, line 67: This does and should not work in Qt/WebView. This is what `import { open } from 'api/system.ts';` is for,

NicolaLS: ah yes, I though so but still wanted to give it a shoot :) However I could not test this because of the issue with the dialog not even opening/showing.

Is there the possibility to open a link without `api/system.tx open` ? If not then my whole approach does not really work anyway...I wanted to use `open` and then if it [fails](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2333/files#diff-a0758e6db02536ac8e041b50367493f59198a892c667b35d768d2cfc7da9b497R61) because it was blocked the user should confirm opening it anyway in the dialog.

NicolaLS:
In case the `api/system.tx open` is the only way to open links right now using a confirm dialog and then opening the link anyway does not work. However it would still be interesting to understand why the `alertUser` and the confirmation dialog don't show up at all in qt/android. (I tested it without my changes where it should (and does in web) show the `alertUser` but nothing happens on qt)

frontends/web/src/components/anchor/anchor.tsx
NicolaLS: (just checked it looks like we need the backend to open any url)

NicolaLS:
superseded by PR 2579

PR #2334, opened 2023-11-06
Bitsurance dashboard implementation

Beerosagos:
Updates the backend service to load bitsurance contract details for each insured account and implements a dashboard page in the frontend. The dashboard page is only about content, it needs several aesthetic improvements which will be done in a future PR.


frontends/web/src/routes/account/account.tsx
```diff
@@ -84,17 +84,20 @@ export function Account({
...
-    if (account?.bitsuranceId) {
+    if (account?.bitsuranceStatus) {
       const insuredAccounts = await bitsuranceLookup(code);
       if (!insuredAccounts.success) {
         alertUser(insuredAccounts.errorMessage || t('genericError'));
         return;
       }
-      if (!insuredAccounts.accountCodes.includes(code)) {
+      let bitsuranceAccount = insuredAccounts.bitsuranceAccounts[0];
+      let active = bitsuranceAccount.status === 'active';
```
thisconnect, line 94: I'd always prefer `const` if the intention is that the variable never changes after assignment.

frontends/web/src/routes/account/components/insuredtag.tsx
```diff
@@ -17,17 +17,20 @@
...
 export const Insured = () => {
   const { t } = useTranslation();
   return (
-    <div className={`${style.insured}`}>
-      <img draggable={false} src={shieldIcon} alt={t('sidebar.insurance')} />
-      {t('account.insured')}
-    </div>
+    <Link className={`${style.insured}`} to="/bitsurance/dashboard">
+      <div >
+        <img draggable={false} src={shieldIcon} alt={t('sidebar.insurance')} />
```
thisconnect, line 30: Move to icon component?

thisconnect (changes requested):
nice, I haven't tested yet, but planing to later.

frontends/web/src/api/account.ts
```diff
@@ -50,7 +50,7 @@ export interface IAccount {
     isToken: boolean;
     activeTokens?: IActiveToken[];
     blockExplorerTxPrefix: string;
-    bitsuranceId?: string;
+    bitsuranceStatus?: string;
```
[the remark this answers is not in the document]
Beerosagos: Right, thanks

frontends/web/src/components/icon/assets/icons/yellow-dot.svg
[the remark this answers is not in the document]
Beerosagos: They came from the MVP, but I modified them to follow those you suggested

frontends/web/src/routes/account/account.tsx
Beerosagos: done :+1:

Beerosagos:
I realized I need to modify a bit the behavior for canceled contract, putting back in draft while I work on an update

Beerosagos:
> I realized I need to modify a bit the behavior for canceled contract, putting back in draft while I work on an update

Fixed, ready for review again :+1:

backend/bitsurance/bitsurance.go
```diff
@@ -35,38 +37,41 @@ const (
 	widgetTestURL = "https://test.bitsurance.eu/?wallet=bitbox&version=" + widgetVersion + "&lang="
 )
 
-const statusActive = "active"
+type DetailStatus string
```
benma, line 40: exported things should have a docstring - I thought our CI would complain about it :smile: 

would be great to have a description here and for all the instances below about what the states mean and when they occur.

backend/bitsurance/bitsurance.go
```diff
@@ -35,38 +37,41 @@ const (
...
-	if code == http.StatusNotFound {
-		return false, nil
-	}
-
-	if details.Status == statusActive {
-		return true, nil
-	}
-	return false, nil
+	ratAmount := new(big.Rat).SetInt64(int64(account.Details.MaxCoverage))
+	account.Details.MaxCoverageFormatted = coin.FormatAsCurrency(ratAmount, account.Details.Currency, false)
```
benma, line 73: seems like the currency field for now is unused except if it is BTC? maybe it is safer to convert the param to `isBTC bool` then, otherwise things might silently break if we start using the currency parameter for other things (e.g. we don't know exactly if the bitsurance Currency field maps perfectly to our own list of fiat currencies).

backend/handlers/handlers.go
```diff
@@ -1079,17 +1070,24 @@ func (handlers *Handlers) getExchangesByRegion(r *http.Request) interface{} {
 
 func (handlers *Handlers) getBitsuranceLookup(r *http.Request) interface{} {
```
benma, line 1071: generally state-changing HTTP calls should be done as POST requests for clarity/consistency.

i'd switch it to be a POST call instead.

frontends/web/src/routes/account/account.tsx
```diff
@@ -84,17 +84,22 @@ export function Account({
   }, [code]);
 
   const maybeCheckBitsuranceStatus = useCallback(async () => {
-    if (account?.bitsuranceId) {
-      const insuredAccounts = await bitsuranceLookup(code);
+    if (account?.bitsuranceStatus) {
+      // we set the `persistCancel` flag to avoid notifying
+      // again in case of insurance deactivation.
```
benma, line 89: not sure I understand this comment, could you elaborate?

persistCancel = true leads to `backend.emitAccountsStatusChanged()` right? or do you mean some other notification?

frontends/web/src/routes/bitsurance/bitsurance.tsx
```diff
@@ -44,11 +44,13 @@ export const Bitsurance = ({ accounts }: TProps) => {
   const amount = '100.000€';
 
   useEffect(() => {
-    setInsuredAccounts(accounts.filter(({ bitsuranceId }) => bitsuranceId));
+    if (accounts.filter(({ bitsuranceStatus }) => bitsuranceStatus).length) {
```
benma, line 47: `accounts.some()` does the same I think

benma (changes requested):

backend/bitsurance/bitsurance.go
Beerosagos: It used to complain, don't know what happened :joy: 

I accidentally erased the comment explaining the possible status, I'll add it again.

frontends/web/src/routes/account/account.tsx
Beerosagos: I mean the frontend notification dialog that is shown on the account page when the account insurance status changes from 'active' to something else (https://github.com/Beerosagos/bitbox-wallet-app/blob/bitsurance-dashboard-rebased/frontends/web/src/routes/account/account.tsx#L99)

backend/bitsurance/bitsurance.go
Beerosagos: Done

backend/bitsurance/bitsurance.go
Beerosagos: Done, I made a dedicated commit and took the chance to refactor a bit, renaming a few confusing variables.

frontends/web/src/routes/bitsurance/bitsurance.tsx
Beerosagos: TIL :open_mouth:

Beerosagos:
Pushed an update to address the comments (except the one about the post call, where I propose an alternative approach). I also fixed a couple things that came up while testing: 
- added the balance to the account selection list
- route to / when the user detaches the bitbox while on bitsurance page.

frontends/web/src/routes/bitsurance/dashboard.tsx
```diff
@@ -0,0 +1,156 @@
...
+    case 'waitpayment':
+      return <YellowDot/>;
+    case 'refused':
+      return <RedDot/>;
+    case 'inactive':
+      return <OrangeDot/>;
+    case 'canceled':
+      return <RedDot/>;
+    }
+  };
```
thisconnect, line 63: small nit: This is pure function and could be defined before `export const BitsuranceDashboard`, maybe rename to AccountStatusIcon so you can use it in JSX, `<AccountStatusIcon status={account.status}/>`

thisconnect (changes requested):

Beerosagos:
@benma just pushed an update. Didn't address @thisconnect considerations yet (WIP), but in the new code the bitsurance lookup call is now a POST, as you suggested. If the lookup function fetches a new insurance status, it is immediately persisted in the account config, without special cases for deactivated (inactive or canceled) contracts. When a contract is deactivated, the backend saves the related account code inside a dedicated field of the frontend config. This way, the frontend can check for that value, alert the user about the contract deactivation and then remove it from the config, to avoid duplicated alerts.

backend/accounts.go
```diff
@@ -228,19 +228,36 @@ func (backend *Backend) LookupInsuredAccounts(accountCode accountsTypes.Code) ([
...
-				return errp.Newf("Could not find account %s", code)
+				return errp.Newf("Could not find account %s", bitsuranceAccount.AccountCode)
 			}
-			if accountConfig.Insured != insured {
-				accountConfig.Insured = insured
-				backend.log.Infof("Account [%s] insurance status changed to %v", code, insured)
+			if accountConfig.InsuranceStatus != bitsuranceStatus {
+				backend.log.Infof("Account [%s] insurance status changed to %v", bitsuranceAccount.AccountCode, bitsuranceStatus)
+				canceled := bitsuranceStatus == string(bitsurance.CanceledStatus) || bitsuranceStatus == string(bitsurance.InactiveStatus)
+				if canceled {
```
benma, line 240: just an idea: couldn't you return a flag from this function in this case, and in the handler return that result to the frontend, and then let the frontend react to it by doing a notification?

to avoid having this frontend stuff here deep in the backend.

or is this function also called in other places independently from the handler that still needs to notify the user?

benma (approved):

backend/accounts.go
benma: another idea:

send an event to the frontend and have the frontend react to that? `backend.Notify(observable.Event{...})`

Beerosagos: The function is called both on the dashboard and on the account page, but the notification is only displayed on the account page. This solution allows notifying the user even if the app is closed before the alert is actually displayed. As an alternative, we could subscribe to the event and persist the config from the frontend, but I'm not sure that it is going to simplify much the code.

frontends/web/src/routes/buy/info.tsx
```diff
@@ -45,7 +44,7 @@ export const BuyInfo = ({ code, accounts }: TProps) => {
       const options =
         supportedAccounts.map(({ name, code, coinCode }) => ({ label: `${name}`, value: code, coinCode, disabled: false }));
       setOptions(options);
-      getBalances(options);
+      setOptionBalances(options, setOptions);
```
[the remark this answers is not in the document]
Beerosagos: Yeah, you're definitely right. I updated the function to just take in input an options array and return an updated array with balances. I also removed the double call to the state setter in `buy/info.tsx` and `bitsurance/account.tsx`, as I don't see a big advantage in populating it twice just to avoid waiting for the balance calls to return.

frontends/web/src/api/account.ts
```diff
@@ -16,6 +16,7 @@
 
 import { apiGet, apiPost } from '../utils/request';
 import { ChartData } from '../routes/account/summary/chart';
+import { TDetailStatus } from './bitsurance';
```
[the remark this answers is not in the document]
Beerosagos: Done

frontends/web/src/routes/bitsurance/account.tsx
```diff
@@ -39,15 +41,26 @@ export const BitsuranceAccount = ({ code, accounts }: TProps) => {
...
+    if (!response.success) {
+      alertUser(response.errorMessage);
+      return;
+    }
+    let insuredAccountsCodes = response.bitsuranceAccounts.map(account => account.status && account.status !== 'canceled' ? account.code : null);
     const options = accounts
-      .map(({ name, code, coinCode, bitsuranceId }) => (
-        { label: name, value: code, coinCode, disabled: false, insured: !!bitsuranceId }
+      .map(({ name, code, coinCode }) => (
+        { label: name, value: code, coinCode, disabled: false, insured: insuredAccountsCodes.includes(code) }
```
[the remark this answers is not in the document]
Beerosagos: In the previous implementation the cancel or inactive status weren't immediately persisted in the account config, so we needed to fetch the status to be sure. Now this is not needed anymore, as each status change is immediately persisted by the backend.

frontends/web/src/routes/bitsurance/bitsurance.tsx
```diff
@@ -57,15 +59,20 @@ export const Bitsurance = ({ accounts }: TProps) => {
       alertUser(response.errorMessage);
       return;
     }
-    setInsuredAccounts(accounts.filter(({ code }) => response.accountCodes.includes(code)));
+    let insuredAccountsCodes = response.bitsuranceAccounts.map(account => account.status ? account.code : null);
+    let insured = accounts.filter(({ code }) => insuredAccountsCodes.includes(code));
```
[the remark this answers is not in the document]
Beerosagos: Done :)

frontends/web/src/routes/bitsurance/dashboard.tsx
Beerosagos: Done!

frontends/web/src/routes/bitsurance/dashboard.tsx
```diff
@@ -0,0 +1,160 @@
...
+                      { balances ? (
+                        <>
+                          <Amount
+                            amount={balances[account.code]?.available.amount}
+                            unit={balances[account.code]?.available.unit}
+                            removeBtcTrailingZeroes/>
+                          {' ' + balances[account.code]?.available.unit}
+                        </>
+                      ) : <Skeleton/>}
+                      <p>{}</p>
```
thisconnect, line 133: remove?

frontends/web/src/routes/bitsurance/account.tsx
```diff
@@ -39,15 +41,26 @@ export const BitsuranceAccount = ({ code, accounts }: TProps) => {
...
 
-  // check supported accounts
-  useEffect(() => {
+  const detect = useCallback(async () => {
+    const response = await bitsuranceLookup();
+    if (!response.success) {
+      alertUser(response.errorMessage);
+      return;
+    }
+    let insuredAccountsCodes = response.bitsuranceAccounts.map(account => account.status && account.status !== 'canceled' ? account.code : null);
```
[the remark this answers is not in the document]
thisconnect: thanks, I'll keep this unresolved for future me and others. Or maybe this is worth adding a comment in the 2 places?

thisconnect (changes requested):

frontends/web/src/routes/bitsurance/account.tsx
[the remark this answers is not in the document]
Beerosagos: Added a comment in account.tsx to explain why 'canceled' status are included

backend/bitsurance/bitsurance.go
```diff
@@ -35,38 +37,49 @@ const (
...
+)
+
+type AccountDetails struct {
+	AccountCode types.Code   `json:"code"`
+	Status      DetailStatus `json:"status"`
+	Details     struct {
+		MaxCoverage          int    `json:"maxCoverage"`
+		MaxCoverageFormatted string `json:"maxCoverageFormatted"`
+		Currency             string `json:"currency"`
+		Support              string `json:"support"`
```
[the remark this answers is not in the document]
Beerosagos: It is the link to the proper support page, added a comment to explain

frontends/web/src/api/bitsurance.ts
```diff
@@ -14,18 +14,31 @@
...
 };
 export const getBitsuranceURL = (): Promise<string> => {
   return apiGet('bitsurance/url');
 };
 
-export const bitsuranceLookup = (code?: AccountCode): Promise<TInsuredAccounts> => {
-  return apiGet(`bitsurance/lookup?code=${code || ''}`);
+export const bitsuranceLookup = (code: AccountCode = ''): Promise<TInsuredAccounts> => {
+  return apiPost('bitsurance/lookup', { code });
 };
```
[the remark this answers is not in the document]
Beerosagos: Yeah, if code is specified it checks only that account, otherwise it'll check all the available btc accounts. Added a comment to explain

frontends/web/src/routes/bitsurance/account.tsx
```diff
@@ -39,15 +41,25 @@ export const BitsuranceAccount = ({ code, accounts }: TProps) => {
...
+    if (!response.success) {
+      alertUser(response.errorMessage);
+      return;
+    }
     const options = accounts
-      .map(({ name, code, coinCode, bitsuranceId }) => (
-        { label: name, value: code, coinCode, disabled: false, insured: !!bitsuranceId }
+      .map(({ name, code, coinCode, bitsuranceStatus }) => (
+        { label: name, value: code, coinCode, disabled: false, insured: bitsuranceStatus && bitsuranceStatus !== 'canceled' }
       )).filter(account => account.coinCode === 'btc' && !account.insured);
```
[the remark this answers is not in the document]
Beerosagos: Good point, done!

frontends/web/src/routes/bitsurance/dashboard.tsx
Beerosagos: done :+1:

thisconnect (approved):
LGTM 👍 
thank you

PR #2338, opened 2023-11-06
ci: use macos 12

thisconnect:
CI currently always fails on macOS, seems due to homebrew not supporting macOS 11 anymore.


Beerosagos (approved):
utACK

thisconnect:
We could also use macos-13, but I'll merge as it is so that the CI's don't run for hours.

Screen Shot 2023-11-07 at 10 40 55

PR #2341, opened 2023-11-08
merge master into staging-watchonly

thisconnect:
Fixed conflict in backend/config/accounts.go


benma (approved):
Thanks :heart:  I verified I get the same result by merging & resolving.

PR #2344, opened 2023-11-08
frontend: type system open api

thisconnect:
Part of typing all api calls.


benma (approved):
utACK :heart:  (sorry I was not assigned but I just opened for fun and it's a small change :smile: )

PR #2345, opened 2023-11-08
frontend: created `NoteInput` component for send screen

shonsirsha:
As a part of the refactoring effort of `send.tsx`, we're creating a new component `NoteInput` as its own component to be used in the send screen.


thisconnect (approved):
untested LGTM 👍

PR #2352, opened 2023-11-13
frontend: better error message for invalid pairing uri WC

shonsirsha:
To improve UX, we'd like to make the error message clearer when user is pairing dapp via WC using invalid URI.

[Preview before]: 

xxx

[Preview after]:

qwedas


thisconnect (approved):
LGTM 
Related: The connect button could be disabled as long as the input field is empty.

PR #2353, opened 2023-11-13
created `ButtonsGroup` component for send screen

shonsirsha:
As a part of the refactoring of `send.tsx` component, we'd like to extract `ButtonsGroup` into its own component to be used in the send component.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -25,8 +25,8 @@ import { getDeviceInfo } from '../../../api/bitbox01';
 import { alertUser } from '../../../components/alert/Alert';
 import { Balance } from '../../../components/balance/balance';
 import { HideAmountsButton } from '../../../components/hideamountsbutton/hideamountsbutton';
-import { Button, ButtonLink } from '../../../components/forms';
-import { Column, ColumnButtons, Grid, GuideWrapper, GuidedContent, Header, Main } from '../../../components/layout';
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -654,21 +657,11 @@ class Send extends Component<Props, State> {
...
-                        {t('button.back')}
-                      </ButtonLink>
-                    </ColumnButtons>
+                    <ButtonsGroup
+                      onSendButtonClick={this.send}
+                      accountCode={code}
+                      isSendButtonDisabled={sendButtonDisabled}
+                    />
                   </Column>
                 </Grid>
```
thisconnect, line 666: This change hides await the 2 buttons, but does it need to be in its own re-usable component? I am a bit afraid that continuing this way the code base is getting fragmented. I slightly prefer having `<ColumnButtons>` and `<Column>` be used in the same render. Do you have strong opinion?

thisconnect: I slightly prefer to keep as is and close this PR so it is easier to understand the code.

thisconnect: >  Main idea here is just to slightly improve readability in send.tsx, which is as we know; very long. And breaking up into different components helps in this case. It's less about making them a reusable component.

IMO the render/JSX part is fine, the beast is in the business logic of send.tsx. Looking forward to the service improvements.

PR #2354, opened 2023-11-14
frontend/bitsurance: update bitsurange external page link

Beerosagos:


frontends/web/src/routes/bitsurance/bitsurance.tsx
```diff
@@ -62,6 +65,14 @@ export const Bitsurance = ({ accounts }: TProps) => {
     setScanLoading(false);
   };
 
+  const getBitsurancePageLink = (): string => {
+    const selectedLanguage = defaultLanguages[getSelectedIndex(defaultLanguages, i18n)];
```
thisconnect, line 69: In other places `i18n.language` is used, isn't this the same?

https://github.com/digitalbitbox/bitbox-wallet-app/blob/b3f968702a030a55eca5eadb1582de1ae8848b5a/frontends/web/src/components/appdownloadlink/appdownloadlink.tsx#L24

thisconnect (approved):
LGTM with nit

frontends/web/src/routes/bitsurance/bitsurance.tsx
Beerosagos: AFAIK `i18n.language` could contain subtags (e.g. en-US). Using `getSelectedIndex' should be more robuts (see https://github.com/Beerosagos/bitbox-wallet-app/blob/886b61375af9f33bce36455b0dd5509c156cdc4c/frontends/web/src/utils/language.ts#L4)

thisconnect: @Beerosagos sorry I didn't have time and it is still open in one of my many tabs. I wanted to look into why this is different and why we use `i18n.language` in 15 other places.

thisconnect (approved):
LGTM with nit

frontends/web/src/routes/bitsurance/bitsurance.tsx
thisconnect: You're absolutely right, I changed my OS to German and `i18n.language` is 'de-CH' which causes some links to open in the wrong language, for example in the waiting guide.

So we shouldn't use `i18n.language` directly anymore, except internally internally language

Lets keep as is now and consider using https://www.i18next.com/overview/api#resolvedlanguage in the future.

thisconnect: PR https://github.com/digitalbitbox/bitbox-wallet-app/pull/2380

I noticed there are a few places that use Intl formatting, there we should keep the region info. I wrote some details in the commit message

PR #2358, opened 2023-11-15
frontend: make locize-pull

benma:
passphrase

making the 'no longer access your coins' part bold on @My1's suggestion


thisconnect (approved):
LGTM

PR #2359, opened 2023-11-16
frontend/status: fix config.frontend null bug

Beerosagos:
config.frontend could be null, raising a js exception. This fixes it.


frontends/web/src/components/status/status.tsx
```diff
@@ -45,7 +45,7 @@ export const Status = ({
   const checkConfig = useCallback(async () => {
     if (dismissible) {
       const config = await getConfig();
-      setShow(!config ? true : !config.frontend[dismissible]);
+      setShow(!config?.frontend ? true : !config.frontend[dismissible]);
```
thisconnect, line 48: In many other places we check only for config and if defined assume that there is `config.frontend`. How  can this happen / could you add steps to reproduce?

alternatively, couldn't the backend ensure that there is always a frontend property with at least an empty object?

Beerosagos: I noticed an error caused by this a few days ago, but could not reproduce it anymore. I opened the PR anyway as it was quick, but yeah if there are other places with the same check it could make sense to address it backend side

Beerosagos: Didn't have time yet, but I'll look into the backed part to see if we can fix it there

Beerosagos:
Pushed a new commit to ensure that `config.frontend` is always not null. 
@benma do you see any issue with this?

PR #2362, opened 2023-11-19
frontend: created bitsurance terms component

shonsirsha:
also moved all terms component (`pocket-terms`, `moonpay-terms`, and `bitsurance-terms`) into a dedicated 'terms' directory.

[Preview (bitsurance-terms)]: 

eqwdas


frontends/web/src/components/terms/bitsurance-terms.tsx
```diff
@@ -0,0 +1,64 @@
...
+import style from './terms.module.css';
+
+type TProps = {
+  onAgreedTerms: () => void;
+}
+
+export const BitsuranceTerms = ({ onAgreedTerms }: TProps) => {
+  const { t } = useTranslation();
+  const handleSkipDisclaimer = (e: ChangeEvent<HTMLInputElement>) => {
+    setConfig({ frontend: { skipPocketDisclaimer: e.target.checked } });
```
Beerosagos, line 31: `skipPocketDiscalimer` should be updated :)

frontends/web/src/components/terms/bitsurance-terms.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <div className={style.disclaimerContainer}>
+      <div className={style.disclaimer}>
+        <p>{t('bitsurance.terms.text1')}</p>
+        <p>{t('bitsurance.terms.text2')}</p>
+        <p>{t('bitsurance.terms.text3')}</p>
+        <p>{t('bitsurance.terms.text4')}</p>
+        <p>
+          {t('bitsurance.terms.text5')}
+          {' '}
+          <A href="https://bitsurance.eu/en/dataprotection">
```
Beerosagos, line 44: Please use `www.bitsurance.eu` as it is the whitelisted domain in the backend

Beerosagos (changes requested):
Found a couple issues, but also the page is not checking the config set to skip the disclaimer on load. Could you add it please?

frontends/web/src/components/terms/bitsurance-terms.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <div className={style.disclaimerContainer}>
+      <div className={style.disclaimer}>
+        <p>{t('bitsurance.terms.text1')}</p>
+        <p>{t('bitsurance.terms.text2')}</p>
+        <p>{t('bitsurance.terms.text3')}</p>
+        <p>{t('bitsurance.terms.text4')}</p>
+        <p>
+          {t('bitsurance.terms.text5')}
+          {' '}
+          <A href="https://www.bitsurance.eu/en/dataprotection">
```
thisconnect, line 44: Link to https://www.bitsurance.eu/datenschutz/ would be nice to have too.

thisconnect (changes requested):

frontends/web/src/components/terms/bitsurance-terms.tsx
Beerosagos: Good point, actually! Thanks

thisconnect: We used to put it in locize but it got messy quickly and afaik we started putting having links in JSX directly with language switch. I guess German-first is ok in this case (their english page is not ready yet).

thisconnect (approved):
LGTM

PR #2363, opened 2023-11-19
frontend: changed WC supported chains to only ETH & ETH goerli

shonsirsha:
also removed unused chain logos (.svgs). 🙂


thisconnect (approved):
LGTM 👍

PR #2366, opened 2023-11-20
frontend/send: fix bug in send form

Beerosagos:
Calling `validateAndDisplayFee` without waiting for `setState` to complete was causing the actual transaction proposal amount possibly not to be updated to the last inserted digit.


thisconnect (approved):
nice catch, untested LGTM 👍

PR #2368, opened 2023-11-22
Android system auth

Beerosagos:
Watch-only can benefit from a feature that allows the user to authenticate when opening the bitboxapp, to protect the privacy. This commit creates an Auth component in the frontend that can trigger an environment-based authentication.

On Qt and Webdev the authentication succeds automatically, while on Android it requires the user to complete the system biometric authentication every time the app is resumed (.i.e. at startup and when the app comes to foreground after being put in the background).

The authentication check can be requested by the frontend using the `auth` endpoint.

Authentication messages from backend to frontend are sent via three new notifications:
- 'auth-required': which triggers a new `authenticate` call from the frontend Auth component
- 'auth-ok': which means authentication completed successfully
- 'auth-err': which means authentication failed or has been canceled by the user.

There is also a `trigger-auth` endpoint that can be used by frontend components to trigger a new authentication flow (making the backend send a new `auth-required` notification).

NOTES: 
- the authentication screen is just a draft and will need some styling in a future PR

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -116,6 +116,7 @@ type authEventType string
 
 const (
 	authRequired authEventType = "auth-required"
+	authForced   authEventType = "auth-forced"
 	authOk       authEventType = "auth-ok"
 	authErr      authEventType = "auth-err"
 )
@@ -309,6 +310,22 @@ func (backend *Backend) Config() *config.Config {
 	return backend.config
 }
 
+// Authenticate executes a system authentication if
+// the authentication config flag is enabled or if the
+// `force` input flag is enabled (as a consequence of an
+// 'auth/auth-forced' notification).
+// Otherwise, the authentication is automatically assumed as
+// successful.
+func (backend *Backend) Authenticate(force bool) {
+	backend.log.Info("Auth requested")
+	if backend.config.AppConfig().Backend.Authentication || force {
+		backend.environment.Auth()
+	} else {
+		backend.AuthResult(true)
+	}
+}
+
+// TriggerAuth triggers an auth-required notification.
 func (backend *Backend) TriggerAuth() {
 	backend.Notify(observable.Event{
 		Subject: "auth",
@@ -319,6 +336,27 @@ func (backend *Backend) TriggerAuth() {
 	})
 }
 
+// ForceAuth triggers an auth-forced notification
+// followed by an auth-required notification.
+func (backend *Backend) ForceAuth() {
+	backend.Notify(observable.Event{
+		Subject: "auth",
+		Action:  action.Replace,
+		Object: authEventObject{
+			Typ: authForced,
+		},
+	})
+	backend.Notify(observable.Event{
+		Subject: "auth",
+		Action:  action.Replace,
+		Object: authEventObject{
+			Typ: authRequired,
+		},
+	})
+}
+
+// AuthResult triggers an auth-ok or auth-err notification
+// depending on the input value.
 func (backend *Backend) AuthResult(ok bool) {
 	backend.log.Infof("Auth result: %v", ok)
 	typ := authErr
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -117,6 +117,7 @@ type authEventType string
 const (
 	authRequired authEventType = "auth-required"
 	authForced   authEventType = "auth-forced"
+	authCanceled authEventType = "auth-canceled"
 	authOk       authEventType = "auth-ok"
 	authErr      authEventType = "auth-err"
 )
@@ -332,6 +333,17 @@ func (backend *Backend) TriggerAuth() {
 		Action:  action.Replace,
 		Object: authEventObject{
 			Typ: authRequired,
+		},
+	})
+}
+
+// CancelAuth triggers an auth-canceled notification.
+func (backend *Backend) CancelAuth() {
+	backend.Notify(observable.Event{
+		Subject: "auth",
+		Action:  action.Replace,
+		Object: authEventObject{
+			Typ: authCanceled,
 		},
 	})
 }
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -112,6 +112,18 @@ type AccountEvent struct {
 	Data string             `json:"data"`
 }
 
+type authEventType string
+
+const (
+	authRequired authEventType = "auth-required"
+	authOk       authEventType = "auth-ok"
+	authErr      authEventType = "auth-err"
+)
+
+type authEventObject struct {
+	Typ authEventType `json:"typ"`
+}
+
 // Environment represents functionality where the implementation depends on the environment the app
 // runs in, e.g. Qt5/Mobile/webdev.
 type Environment interface {
@@ -143,6 +155,7 @@ type Environment interface {
 	SetDarkTheme(bool)
 	// DetectDarkTheme returns true if the dark theme is enabled at OS level.
 	DetectDarkTheme() bool
+	Auth()
 }
 
 // Backend ties everything together and is the main starting point to use the BitBox wallet library.
@@ -294,6 +307,31 @@ func (backend *Backend) notifyNewTxs(account accounts.Interface) {
 // Config returns the app config.
 func (backend *Backend) Config() *config.Config {
 	return backend.config
+}
+
+func (backend *Backend) TriggerAuth() {
+	backend.Notify(observable.Event{
+		Subject: "auth",
+		Action:  action.Replace,
+		Object: authEventObject{
+			Typ: authRequired,
+		},
+	})
+}
+
+func (backend *Backend) AuthResult(ok bool) {
+	backend.log.Infof("Auth result: %v", ok)
+	typ := authErr
+	if ok {
+		typ = authOk
+	}
+	backend.Notify(observable.Event{
+		Subject: "auth",
+		Action:  action.Replace,
+		Object: authEventObject{
+			Typ: typ,
+		},
+	})
 }
 
 // DefaultAppConfig returns the default app config.
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -101,6 +101,9 @@ type Backend interface {
 	AOPPCancel()
 	AOPPApprove()
 	AOPPChooseAccount(code accountsTypes.Code)
+	Authenticate(force bool)
+	TriggerAuth()
+	ForceAuth()
 	GetAccountFromCode(code string) (accounts.Interface, error)
 	HTTPClient() *http.Client
 	CancelConnectKeystore()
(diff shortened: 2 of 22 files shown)


Beerosagos:
Setting it back in draft as I'm considering a few changes

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -358,12 +358,39 @@ public void onChanged(Boolean isDarkTheme) {
             }
         });
 
+        goViewModel.getRequestAuth().observe(this, new Observer<Boolean>() {
+            @Override
+            public void onChanged(Boolean requestAuth) {
+                Util.log("requested Auth: " + requestAuth);
+                if (requestAuth) {
+                    goViewModel.setRequestAuth(false);
```
Beerosagos, line 366: I don't like much using a MutableLiveData to trigger the requestAuth, but I could not find a nicer way to do it.

frontends/web/src/components/auth/authrequired.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const subscriptions = [
+      subscribeAuthRequired(performAuth),
+      subscribeAuthErr(performAuth),
+      subscribeAuthOk(() => setAuthRequired(false)),
+    ];
+    return () => unsubscribe(subscriptions);
+  }, []);
+
+  return (
+    <div className={[style.auth, authRequired ? style.active : style.hidden].join(' ')}>
```
benma, line 46: instead of `style.hidden` you could just return null?

`if (!autRequired) { return null; }`

to not put anything into the DOM.

backend/backend.go
```diff
@@ -296,6 +297,39 @@ func (backend *Backend) Config() *config.Config {
...
+		backend.environment.Auth()
+	} else {
+		backend.AuthResult(true)
+	}
+}
+
+// TriggerAuth triggers an auth-required notification.
+func (backend *Backend) TriggerAuth() {
+	backend.Notify(observable.Event{
+		Subject: "auth-required",
```
benma, line 314: imo it would be clearer to use the "auth" subject for all three events and use the Data field instead to send additional info which will be captured in the frontend subscription callback param. See for example

https://github.com/digitalbitbox/bitbox-wallet-app/blob/6b8ec3cc4b6a1037ffa93e19fb92d9cf5e731ff9/frontends/web/src/api/backend.ts#L79-L102 and https://github.com/digitalbitbox/bitbox-wallet-app/blob/6b8ec3cc4b6a1037ffa93e19fb92d9cf5e731ff9/backend/accounts.go#L571-L578

backend/handlers/handlers.go
```diff
@@ -193,6 +195,8 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/update", handlers.getUpdateHandler).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/banners/{key}", handlers.getBannersHandler).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/using-mobile-data", handlers.getUsingMobileDataHandler).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/authenticate", handlers.getAuthenticateHandler).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/trigger-auth", handlers.getTriggerAuthHandler).Methods("GET")
```
benma, line 199: these feel more like they should be POST - GET is for retrieving data, POST is to change state.

frontends/android/goserver/goserver.go
```diff
@@ -209,3 +211,11 @@ func Serve(dataDir string, environment GoEnvironmentInterface, goAPI GoAPIInterf
...
 	bridgecommon.Shutdown()
 }
+
+func TriggerAuth() {
+	bridgecommon.TriggerAuth()
+}
+
+func AuthResult(ok bool) {
+	bridgecommon.AuthResult(ok)
+}
```
benma, line 221: docstrings missing (I know it is a bit pedantic :joy: but imo it is great for code hygiene to always document exported functions, even if they are trivial).

we need to fix that linter :P

frontends/web/src/routes/settings/advanced-settings.tsx
```diff
@@ -82,6 +87,9 @@ export const AdvancedSettings = ({ deviceIDs, hasAccounts }: TPagePropsWithSetti
               >
                 <EnableCustomFeesToggleSetting frontendConfig={frontendConfig} onChangeConfig={setConfig} />
                 <EnableCoinControlSetting frontendConfig={frontendConfig} onChangeConfig={setConfig} />
+                { runningInAndroid() && (
```
benma, line 90: i would move the `runningInAndroid()` check inside the `EnableAuthSetting` component so it is self-contained, but no big deal.

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+type TProps = {
+  backendConfig?: TBackendConfig;
+  onChangeConfig: Dispatch<TConfig>;
+}
+
+export const EnableAuthSetting = ({ backendConfig, onChangeConfig }: TProps) => {
+  const { t } = useTranslation();
+
+  const handleToggleFee = async (e: ChangeEvent<HTMLInputElement>) => {
```
benma, line 33: wrong name :P

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
```diff
@@ -0,0 +1,69 @@
...
+      updateConfig(true);
+    }
+  };
+
+  const updateConfig = async (auth: boolean) => {
+    const config = await setConfig({
+      backend: { authentication: auth },
+    }) as TConfig;
+    onChangeConfig(config);
+
```
benma, line 51: remove empty line?

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
```diff
@@ -0,0 +1,69 @@
...
+type TProps = {
+  backendConfig?: TBackendConfig;
+  onChangeConfig: Dispatch<TConfig>;
+}
+
+export const EnableAuthSetting = ({ backendConfig, onChangeConfig }: TProps) => {
+  const { t } = useTranslation();
+
+  const handleToggleFee = async (e: ChangeEvent<HTMLInputElement>) => {
+    // force a last auth before disabling it
```
benma, line 34: shouldn't we also do an auth for enabling it, to test that it works? say the auth thing does not work for some reason on some systems or Android versions - if we enabled this the user might not be able to unlock it ever again.

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (!e.target.checked) {
+      const unsubscribe = subscribeAuthOk(() => {
+        updateConfig(false);
+        unsubscribe();
```
benma, line 38: i am impressed that typescript allows this, seeing that `unsubscribe` is not defined yet syntactically speaking :exploding_head:

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
```diff
@@ -0,0 +1,69 @@
...
+      triggerAuth();
+    } else {
+      updateConfig(true);
+    }
+  };
+
+  const updateConfig = async (auth: boolean) => {
+    const config = await setConfig({
+      backend: { authentication: auth },
+    }) as TConfig;
```
benma, line 49: @thisconnect I wondered the other day why `getConfig()` does not return `TConfig` so that we don't have to put this `as TConfig` thing everywhere. is there a reason?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
benma: i guess the reason is so you can do the stuff here in MainActivity to have access to `MainActivity.this`? or is there a different reason?

this line specifically to reset the requestAuth var so the next one triggers this callback again does look a bit fishy :joy: 

i had a chat with ChatGPT: https://chat.openai.com/share/f7b28f24-d71f-4502-9fa9-0b79b0e7cc20

it commented on the fact that a device rotation etc. will trigger a onStart() again and observe here again, which means it could be called multiple times for each auth-request? did you test what happens after you rotate the screen multiple times:

- what happens to an active auth request?
- what happens to a auth request triggered after some screen rotations?

benma: sholdn't you set this back to `false` after auth has failed or succeeded, so that if multiple auth requests happen in parallel, only one is processed at once?

benma (changes requested):
very nice! haven't tested it yet since testing android is quite difficult for me, but will do it once before we merge this.

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
thisconnect: I forgot that we have `TConfig` :P

The plan is iirc that we move config stuff to a react context and then type it correctly.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
Beerosagos: > i guess the reason is so you can do the stuff here in MainActivity to have access to MainActivity.this? or is there a different reason?

Yes, exactly.

I tested device rotations, but those do not seem to trigger a new onStart. I never experienced issues in that sense. Anyway, I spent a lot of time trying to improve this part to make it both more readable and stable. I experienced some nasty bugs, but now it seems good enough to me.

backend/backend.go
Beerosagos: Good point, done.

frontends/android/goserver/goserver.go
Beerosagos: Eheh sorry, added

frontends/web/src/routes/settings/advanced-settings.tsx
Beerosagos: Done

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
Beerosagos: Ops :D

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
Beerosagos: Good point. I did it, but since the backend automatically skipped the auth if the config was set to false, I had to introduce a new concept of forced auth, which is executed without considering the authentication config flag.

Beerosagos:
@benma thanks for the review. I pushed a new update, PTAL :pray:

frontends/web/src/locales/en/app.json
```diff
@@ -1059,6 +1062,10 @@
       }
     },
     "advancedSettings": {
+      "authentication": {
+        "description": "Requires authentication to use the app.",
+        "title": "Enable authentication"
```
benma, line 1067: phoenix calls this "Screen lock" and description "Lock access to the app with Android's screen lock/fingerprint". I think these fit better, as "authentication" is not very clear imo.

Beerosagos:
@benma Added a new commit that introduces an auth-canceled notification.

benma (approved):
tACK, most excellent!

PR #2374, opened 2023-11-28
frontend: created a toggle for watch-only for each account

shonsirsha:
this toggle is placed inside the 'Edit' dialog in the manage accounts page. The toggle won't be shown if the user disables watch-only in the settings/appearance page.

Preview (when watchonly is disabled from settings/appearance page) - stays the same:
xas

Preview (when watchonly is enabled from settings/appearance page):

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/366dd608-e45a-4ea7-8de4-5ea215b555de

Preview (when BB02 is unplugged and watchonly is enabled from settings/appearance page):

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/0a327a0a-f8eb-4d63-979b-b7272c2ccda5

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -493,6 +493,7 @@ func copyBool(b *bool) *bool {
 // AccountSetWatch sets the account's persisted watch flag to `watch`. Set to `true` if the account
 // should be loaded even if its keystore is not connected.
 // If `watch` is set to `false`, the account is unloaded and the frontend notified.
+// If `watch` is set to `true`, the account is loaded (if the global watchonly flag is enabled) and the frontend notified.
 func (backend *Backend) AccountSetWatch(filter func(*config.Account) bool, watch *bool) error {
 	err := backend.config.ModifyAccountsConfig(func(accountsConfig *config.AccountsConfig) error {
 		for _, acct := range accountsConfig.Accounts {
@@ -524,10 +525,8 @@ func (backend *Backend) AccountSetWatch(filter func(*config.Account) bool, watch
 		}
 	}
 
-	if watch == nil || !*watch {
-		backend.initAccounts(false)
-		backend.emitAccountsStatusChanged()
-	}
+	backend.initAccounts(false)
+	backend.emitAccountsStatusChanged()
 	return nil
 }
 


frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -138,6 +126,36 @@ class ManageAccounts extends Component<Props, State> {
...
+        <div className="flex">
+          <EyeOpenedDark width={18} height={18} />
+          <p className={style.watchOnlyTitle}>Hide account</p>
+        </div>
+        <Toggle
+          checked={!account.watch}
+          className={style.toggle}
+          id={account.code}
+          onChange={async (event) => {
+            event.target.disabled = true;
```
thisconnect, line 142: during testing I could not really see that this disabled the toggle.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -273,36 +291,48 @@ class ManageAccounts extends Component<Props, State> {
...
+                                    primary
+                                    type="submit">
+                                    {t('button.update')}
+                                  </Button>
+                                </DialogButtons>
+                              </form>
+                            </Dialog>
+                          </div>
+                        );
+                      })
```
thisconnect, line 334: I tested with a 2nd wallet and got into a situation when it showed 2 dialogs 🤔  

Screen Shot 2023-11-28 at 15 37 37

thisconnect: Looks like the reason is that I have 2 items in `accountsByKeystore` and Dialog open condition is just if `editAccountCode` is truthy. This is the reason it opens a dialog for every keystore.

thisconnect (changes requested):
I think I found a bug, but that was probably already there. I only tested this branch commit fa794c6cc3ad7fff3474ec67111d7bb01e0a8d90

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -264,46 +287,51 @@ class ManageAccounts extends Component<Props, State> {
...
                         <Message type="error" hidden={!editErrorMessage}>
                           {editErrorMessage}
                         </Message>
                         <Input
-                          onInput={e => this.setState({ editAccountNewName: e.target.value })}
-                          value={editAccountNewName} />
+                          onInput={e => this.setState({ currentlyEditedAccount: { ...currentlyEditedAccount, name: e.target.value } })}
+                          value={currentlyEditedAccount.name}
+                        />
+                        {watchonly &&
```
thisconnect, line 319: is the fragment needed here or would that work?

```suggestion
                        {watchonly && this.renderWatchOnlyToggle()}
```

thisconnect (approved):
nice LGTM with suggestion

benma:
I haven't tested this PR, but in the video, it does not actually show the account disappearing, which would happen if you clicked that button on an account while in watch-only (bitbox02/keystore not connected). what happens to the dialog in that case?

No strong opinion, but should the prompt not be **before** performing the action of hiding the account, in case the user did not actually want the account to disappear immediately? (that was why I suggested a prompt initially)

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -264,46 +287,51 @@ class ManageAccounts extends Component<Props, State> {
...
+                          {this.renderAccounts(keystore.accounts)}
                         </div>
-                      </React.Fragment>))
-                    }
-                    { accounts.length === 0 ? t('manageAccounts.noAccounts') : null }
-
+                      </React.Fragment>
+                    ))
+                  }
+                  {currentlyEditedAccount && (
```
thisconnect, line 306: just one more question: if the Dialog is conditionally rendered it wont animate on small screen correct?

thisconnect:
LGTM

PR #2375, opened 2023-11-28
frontend: fix empty element

thisconnect:
Unlocking a device renders an empty main element. This is only visible in some views (i.e. settings) as a large empty space on top of the view.

Reason is that the setup wizard is now always mounted since watch-only as the device can be unlocked at any time.


thisconnect:
Screen Shot 2023-11-28 at 15 18 59

thisconnect:
@shonsirsha and if you are already on settings and then plugin a device and unlock? Yesterday I could reproduce it everytime. Note if you navigate once it might not happen.

PR #2376, opened 2023-11-29
frontend/qt: directly build shared library, remove windows-cross

benma:
frontend/qt: directly build shared library, remove windows-cross

- We never used the windows-cross target and don't plan on doing
so. We don't know if it works and the rest of the app is not setup for
crosscompiling.

- We drop windows-legacy and start using the windows target, which
directly creates a DLL instead of first creating an static archive and
then converting it to a DLL. It is not clear why the workaround
existed. I vaguely remember that maybe the Go/cgo tooling did not work
properly for building shared libraries on Win back then, but I could
be wrong. The DLL's size is newly 16592896 bytes, before it was
17601090 bytes (saving ~1MB).

- The `windows` target includes some hardening flags from
harderning.mk.inc, which should be good for security. I did not
study/go over them again one by one, but the app seems to be
working. All flags seem to be supported, as unknown flags throw a
compile/link error.

- We remove most libs in `WLIBS`. They were needed in the past when
converting the .a to the .DLL, and were copied to the new `windows`
target when it was made without them being necessary there (at least
not necessary anymore in the current Go version). Go/cgo already
include what is needed when linking. At least `hid` should never have
been necessary as we statically build and link hidapi. Nothing changes
about which DLLs we deploy, so nothing should break. The only
exception is `libssp` which is needed for the stack protector.


benma:
Closing, this actually did not work as we need the `.lib` built from the `server.def` file to link the DLL properly into the BitBox.exe :exploding_head:

PR #2377, opened 2023-11-29
frontend/qt: link breez to the Windows build

benma:
In Windows, DLLs in the same folderas the binary referencing are
automatically found, so we just need to copy the DLL to the same
folder as libserver.dll (which references libbreez).

Builds on https://github.com/digitalbitbox/bitbox-wallet-app/pull/2373/


Beerosagos (approved):
tACK!

Beerosagos (approved):
tACK

Beerosagos (approved):
tACK

PR #2378, opened 2023-11-29
frontend/qt: deploy and link breez to the macOS build

benma:
Similar to the commit that links the breez lib in linux, libserver.so
ends up having an absolute path to libbreez_sdk_bindings becaus of the
rpath linker flag set here:

```
cgo darwin,amd64 LDFLAGS: -Wl,-rpath,${SRCDIR}/lib/darwin-amd64 -L${SRCDIR}/lib/darwin-amd64
```

(https://github.com/breez/breez-sdk-go/blob/7b566724daae0fa539f622b301fe840a3e5cbd59/breez_sdk/cgo.go)

`SRCDIR` is an absolute path and it ends up in the libserver.so
binary (check with `otool -l libserver.so`).

By adding libserver.so to the executables to rewire by macdepoyqt, we
fix its rpath and path to libbreez.

Builds on #2377

### DIFF
--- a/breeztest/breeztest_disabled.go
+++ b/breeztest/breeztest_disabled.go
@@ -1,3 +1,3 @@
-//go:build !linux && !windows
+//go:build !linux && !windows && !darwin
 
 package breeztest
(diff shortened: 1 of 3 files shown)


Beerosagos (approved):
tACK

PR #2381, opened 2023-11-29
frontend: add warning when disabling watch-only from settings

shonsirsha:
Displays a warning/confirmation dialog when disabling watch-only from settings page. Text is subject to change.

Preview: 

ow2


thisconnect (approved):
tested LGTM

PR #2390, opened 2023-11-30
frontend: minor auth view ui improvements

thisconnect:
Using ViewHeader and ViewButtons to position the title and button as we already do in other views.

Had to also add an empty ViewContent so that the layout worked correctly, (we don't have any content yet in this view).


thisconnect:
Screen Shot 2023-11-30 at 10 54 00

Beerosagos (approved):
Looks great! tACK

PR #2391, opened 2023-11-30
android: hide window contents in app switcher when screen lock is active

benma:
When the screen lock setting is active, we want to also hide the window contents in the app switcher, otherwise the contents can be seen there circumventing the screen lock.

Downside: the same setting also disables screenshots. This is why we only enable it when screen lock is enabled, and not always.


Beerosagos (approved):
tACK!

PR #2393, opened 2023-11-30
frontend: bring back download button on android

thisconnect:
The download button in the update banner was removed in:
- aa8a016b322c84e5ad81f92b137c103e3fa8fa38

Reason given was: because Android users usually auto-update their app through a store. This should be added again with a download button that opens up the store where the app has been downloaded.

Adding it back for the following reasons:
- the update banner should have an actionable item and not just information that there is a newer version
- the download page on the website will be improved and will try to show the platform specific download button instead of all downloads
- linking to play store would probably be ok for most users, but some users install the apk directly from GitHub or own build, in these cases it makes sense to just link to the downloads page

### DIFF
--- a/frontends/web/src/components/update/update.tsx
+++ b/frontends/web/src/components/update/update.tsx
@@ -16,7 +16,6 @@
  */
 
 import { useTranslation } from 'react-i18next';
-import { runningInAndroid } from '../../utils/env';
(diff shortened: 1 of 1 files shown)


thisconnect:
I've asked multiple Android users and nobody ever saw an update banner..

If there is a need we can bring it back.

PR #2394, opened 2023-12-01
frontend: show actual account name in sidebar

thisconnect:
The sidebar header used to display ACCOUNTS (uppercase) but with watch-only this now includes the wallet name (user input).

Removed uppercase styling so that the BitBoxApp shows the exact name that the user has choosen. The user can decide to use a name that is all uppercase. The app should just display the name as choosen and not transform to uppercase.


thisconnect:
Screen Shot 2023-12-01 at 09 25 56

thisconnect:
Screen Shot 2023-12-01 at 09 25 28

with this change:

Screen Shot 2023-12-01 at 09 25 45

PR #2398, opened 2023-12-04
frontend: invert watch account option

thisconnect:
The BitBoxApp already has an enable option in the manage-accounts view. Adding a similar 'hide' option in the edit-account dialog is confusing.

This commit changes the wording and flips the toggle, from hide account to watch account. The verb 'watch' is more naturally associable with the feature watch-only.


benma (approved):
tACK. for concept review i defer to @jadzeidan

PR #2400, opened 2023-12-05
backend: avoid deadlock

benma:
backend.AccountSetWatch holds the accounts config lock while it is modifying the config, and while it holds the config lock it tries to acquire the accountsAndKeystoreLock in the filter function. This can deadlock if the accountsAndKeystoreLock is held already.


Beerosagos (approved):
utACK! Very nice

PR #2401, opened 2023-12-05
ci: remove macOS caching

benma:
Recently, half of the PRs fail CI because macOS cache downloading gets
stuck at 99.9%.

We simply remove it, it does not seem to have a performance
downside. Downloading from brew seems to take more or less the same
amount of time as downloading the cache. It may even be faster - CI
with this commit ran in 6min while a previous one with the cache ran
in 7min.


Beerosagos (approved):
LGTM!

PR #2402, opened 2023-12-05
frontend/sidebar: add fingerprint to accounts group to disambiguate

benma:
If multiple keystores have the same name (commonly when you add keystores from one seed using multiple different passphrases), we append the fingerprint to disambiguate the groups.

keystores


thisconnect (approved):
tested LGTM 👍

PR #2404, opened 2023-12-05
frontend: fix route after passphrase device settings changed

thisconnect:
After enabling the passphrase option (watch-only enabled) the app says "Please replug the BitBox02 now", but the app just keeps showing that view.

Without watch-only the unlock device view shows as expected.

Reason was that the /passphrase route was not under /settings and had some special condition which including testing accounts. But with watch-only there might be always accounts, so this failed.

Moved the /passphrase router under /settings/device-settings, so that it is covered by another condition in app.tsx routing logic. With this there is no need to check for /passphrase directly, as it is now under /settings.

Alternative would have been to explicitly test for /passphrase.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -139,7 +139,6 @@ class App extends Component<Props, State> {
       currentURL.startsWith('/account-summary')
       || currentURL.startsWith('/add-account')
       || currentURL.startsWith('/settings/manage-accounts')
-      || currentURL.startsWith('/passphrase')
     )) {
       route('/', true);
       return;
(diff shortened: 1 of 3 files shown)


thisconnect:
I just tested enabling/disabling passphrase once with watch-only and once without watch-only.

thisconnect:
lint failed with some websocket error 🤔

shonsirsha (approved):
Code LGTM. :)

1 Question:

> Moved the /passphrase router under /settings/device-settings, so that it is covered by another condition in app.tsx routing logic. 

Is it this logic? 

https://github.com/digitalbitbox/bitbox-wallet-app/blob/bd1d060b86b19db5f409c000d8253ffe312f861d/frontends/web/src/app.tsx#L146-L151

thisconnect:
> Is it this logic?

exactly.

thisconnect:
ty @shonsirsha 🙏 

last week @benma also wanted to take a look, PTAL!

PR #2405, opened 2023-12-05
frontend: use normal font-size in account edit dialog

thisconnect:
Whenever possible always use regular font-size. In the edit account dialog there is a small description about watch-only this can just use normal font-size.


Beerosagos (approved):
LGTM

PR #2406, opened 2023-12-05
backend: convert global watchonly setting to a keystore setting

benma:
Before, if the global watchonly setting was active, all currently loaded accounts become watchonly, as well as all future accounts, even from different keystores (ther BitBox02s, same BitBox02 with different passphrases, etc.).

This could be confusing. We instead make the global watchonly setting a keystore setting instead.


backend/backend.go
```diff
@@ -886,12 +886,16 @@ func (backend *Backend) CancelConnectKeystore() {
...
-// SetWatchonly sets the app config's watchonly flag to `watchonly`.
-// When enabling watchonly, all currently loaded accounts are turned into watchonly accounts.
-// When disabling watchonly, all persisted accounts's watchonly status is reset.
-func (backend *Backend) SetWatchonly(watchonly bool) error {
-	err := backend.config.ModifyAppConfig(func(config *config.AppConfig) error {
-		config.Backend.Watchonly = watchonly
+// SetWatchonly sets the keystore's watchonly flag to `watchonly`.
+// When enabling watchonly, all currently loaded accounts of that keystore are turned into watchonly accounts.
+// When disabling watchonly, all the watchonly status of all of the keystore's persisted accounts is reset.
+func (backend *Backend) SetWatchonly(rootFingerprint []byte, watchonly bool) error {
```
Beerosagos, line 892: Later in this function, when you call `backend.AccountSetWatch` to set or reset the watch-only flag of the accounts, I think you should filter for keystore, otherwise the setting propagates to all the loaded accounts. This causes e.g. the removal of a watched device's accounts when disabling the watch-only for another device.

backend/accounts_test.go
Beerosagos, line 1: Could be useful to add a test to check that setting the watch-only flag for a keystore doesn't affect accounts belonging to other keystores.

Beerosagos (changes requested):
There is a bug about disabling watchonly for a connected wallet that disables it also for other wallets' accounts. Apart from that looks good!

A couple notes, that is probably better to handle in a next frontend PR (cc @thisconnect):
- accounts are not reloaded inside the manage account view when a keystore is registered/deregistered or device watch-only is enabled/disabled
- I think it would be nice to have the connected keystore's accounts at the top of the sidebar and in the manage account view

backend/accounts_test.go
benma: Yeah the first round of tests was just a first round, I planned to add more later.

backend/backend.go
benma: Great  catch, thanks. Will fix!

backend/accounts_test.go
benma: Fixed and added tests

Beerosagos (approved):
partially tested ACK :+1:

PR #2409, opened 2023-12-06
frontend: only show closebutton in a watchonly receive dialog

thisconnect:
With watch-only the verify receive address can show a connect the correct wallet dialog, introduced in 61a9322a5a1877f518d50e8dfe68e0a9e09953ad

This introduced a minor regression that shows a close button when the dialog should be blocking.


benma (approved):
utACK

PR #2411, opened 2023-12-06
Frontend watch-only guide

thisconnect:
- separated manage-account-guide and add-account-guide
- changed to named export
- add guide entry about watch wallet to manage-account-guide


shonsirsha (approved):
LGTM 🚀

PR #2418, opened 2023-12-08
frontend/web: compile using relative base path instead of absolute

benma:
The default value for `base` is `/`, resulting in

```
<script type="module" crossorigin src="/assets/index-98aa38f3.js"></script>
<link rel="stylesheet" href="/assets/index-f26e81d9.css">
```

in `frontends/web/build/index.html`. For embedded use, the vite docs recommend using a relative path:

```
<script type="module" crossorigin src="./assets/index-98aa38f3.js"></script>
<link rel="stylesheet" href="./assets/index-f26e81d9.css">
```

For Qt and Android, this does not seem to make a difference. On iOS however, the way that local assets are looked up, it has to be relative.


thisconnect (approved):
LGTM 👍 

Lightly tested qt-osx and webdev

Beerosagos (approved):
utACK

PR #2419, opened 2023-12-10
Fix transport types

benma:


benma:
CI fail is a data race likely fixed in https://github.com/digitalbitbox/bitbox-wallet-app/pull/2416/

frontends/web/src/utils/transport-android.ts
```diff
@@ -15,20 +15,20 @@
...
+import { TMsgCallback, TPayload, TQueryPromiseMap } from './transport-common';
 import { runningInAndroid } from './env';
 
 let queryID: number = 0;
 const queryPromises: TQueryPromiseMap = {};
-const currentListeners: Function[] = [];
+const currentListeners: TMsgCallback[] = [];
 
-export function androidCall(query: string) {
+export function androidCall(query: string): Promise<any> {
```
thisconnect, line 25: Before TS somehow automatically used `Promise<unknown>` which is in theory stricter than `Promise<any>`.

Screen Shot 2023-12-24 at 09 37 00

How about using `Promise<unknown>` here? Although it doesn't matter much I think as we set the correct return type in api/ files. I did a quick test and weblint didn't complain using `Promise<unknown>`.  I slightly prefer unkown here.

frontends/web/src/utils/transport-android.ts
```diff
@@ -15,20 +15,20 @@
...
   return new Promise((resolve, reject) => {
     if (runningInAndroid()) {
       if (typeof window.onAndroidCallResponse === 'undefined') {
         window.onAndroidCallResponse = (
           queryID: number,
-          response: string,
+          response: any,
```
thisconnect, line 31: same question as before

> Isn't the response always a string?

frontends/web/src/utils/transport-qt.ts
```diff
@@ -51,7 +51,7 @@ async function initTransport() {
   return webChannel;
 }
 
-export function call(query: string) {
+export function call(query: string): Promise<any> {
```
thisconnect, line 54: Also here, removing it shows me `Promise<unknown>` which should be stricter.

Screen Shot 2023-12-24 at 10 27 32

frontends/web/src/globals.d.ts
```diff
@@ -14,13 +14,15 @@
...
+
 export declare global {
     interface Window {
         qt?: { webChannelTransport: unknown; };
         android?: {
             call: (queryID: number, query: string) => void;
         }
-        onAndroidCallResponse?: (queryID: number, response: string) => void;
-        onAndroidPushNotification?: (msg: string) => void;
+        onAndroidCallResponse?: (queryID: number, response: any) => void;
```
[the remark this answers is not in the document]
benma: No, the string is something is a JSON like `"{ success: true, ... }"` - it is the response of the backend handlers which is always JSON.

frontends/web/src/utils/transport-android.ts
benma: same answer :P

frontends/web/src/utils/transport-android.ts
benma: Sure. It also matches the type in `TQueryPromiseMap`

https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/frontends/web/src/utils/transport-common.ts#L53

frontends/web/src/utils/transport-qt.ts
benma: Done, but strange that `apiGet()` and `apiPost()`, which return `call(...)`, have `Promise<any>` and it does not seem to work if changed `Promise<unknown>`.

frontends/web/src/globals.d.ts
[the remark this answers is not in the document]
thisconnect: you are wright 👍 . I found also where it is converted using JSON.parse:

- https://github.com/digitalbitbox/bitbox-wallet-app/blob/6ee01421939d3fca39d12b73d40b4c7daad392a2/frontends/web/src/utils/transport-websocket.ts#L42
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/6ee01421939d3fca39d12b73d40b4c7daad392a2/frontends/web/src/utils/qwebchannel.js#L79

frontends/web/src/utils/transport-qt.ts
thisconnect: true. I think we could return `Promise<unknown>` for apiGet and apiPost, but then we need to rewrite all api modules and type the return `as Promise<Type>`

```diff
diff --git a/frontends/web/src/api/account.ts b/frontends/web/src/api/account.ts
index 6323214db..1029e95b2 100644
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -80,8 +80,8 @@ type TEthAccountCodeAndNameByAddress = SuccessResponse & {
   name: string;
 }
 
-export const getEthAccountCodeAndNameByAddress = (address: string): Promise<TEthAccountCodeAndNameByAddress> => {
-  return apiPost('accounts/eth-account-code', { address });
+export const getEthAccountCodeAndNameByAddress = (address: string) => {
+  return apiPost('accounts/eth-account-code', { address }) as Promise<TEthAccountCodeAndNameByAddress>;
 };
 
 export interface IStatus {
@@ -91,8 +91,8 @@ export interface IStatus {
     offlineError: string | null;
 }
 
-export const getStatus = (code: AccountCode): Promise<IStatus> => {
-  return apiGet(`account/${code}/status`);
+export const getStatus = (code: AccountCode) => {
+  return apiGet(`account/${code}/status`) as Promise<IStatus>;
 };
 ```

But I am not sure if this improves something.

thisconnect (approved):
LGTM thanks 👍

frontends/web/src/utils/transport-qt.ts
benma: that would look like a step backwards actually

PR #2421, opened 2023-12-11
frontend: extracted txProposal error handling into a function

shonsirsha:
To reduce the amount of logic and as a part of the refactoring of `send.tsx`, we create a new service file to place more logic-heavy codes outside of the `send` component.

This commit moves a lengthy switch case into the service and called it `txProposalErrorHandling`.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -38,6 +38,7 @@ import { route } from '../../../utils/route';
 import { UnsubscribeList, unsubscribe } from '../../../utils/subscriptions';
 import { ConfirmingWaitDialog } from './components/dialogs/confirm-wait-dialog';
 import { SendGuide } from './send-guide';
+import { TProposalError, txProposalErrorHandling } from './services';
 import { MessageWaitDialog } from './components/dialogs/message-wait-dialog';
 import { ReceiverAddressInput } from './components/inputs/receiver-address-input';
 import { CoinInput } from './components/inputs/coin-input';
@@ -60,7 +61,7 @@ interface SignProgress {
 
 type Props = SendProps & TranslateProps;
 
-interface State {
+export type State = {
     account?: accountApi.IAccount;
     balance?: accountApi.IBalance;
     proposedFee?: accountApi.IAmount;
@@ -78,9 +79,9 @@ interface State {
     isSent: boolean;
     isAborted: boolean;
     isUpdatingProposal: boolean;
-    addressError?: string;
-    amountError?: string;
-    feeError?: string;
+    addressError?: TProposalError['addressError'];
+    amountError?: TProposalError['amountError'];
+    feeError?: TProposalError['feeError'];
     paired?: boolean;
     noMobileChannelError?: boolean;
     signProgress?: SignProgress;
@@ -339,32 +340,8 @@ class Send extends Component<Props, State> {
         this.convertToFiat(result.amount.amount);
       }
     } else {
-      const errorCode = result.errorCode;
-      switch (errorCode) {
-      case 'invalidAddress':
-        this.setState({ addressError: this.props.t('send.error.invalidAddress') });
-        break;
-      case 'invalidAmount':
-      case 'insufficientFunds':
-        this.setState({
-          amountError: this.props.t(`send.error.${errorCode}`),
-          proposedFee: undefined,
-        });
-        break;
-      case 'feeTooLow':
-        this.setState({ feeError: this.props.t('send.error.feeTooLow') });
-        break;
-      case 'feesNotAvailable':
-        this.setState({ feeError: this.props.t('send.error.feesNotAvailable') });
-        break;
-      default:
-        this.setState({ proposedFee: undefined });
-        if (errorCode) {
-          this.unregisterEvents();
-          alertUser(errorCode, { callback: this.registerEvents });
-        }
-      }
-      this.setState({ isUpdatingProposal: false });
+      const errorHandling = txProposalErrorHandling(this.registerEvents, this.unregisterEvents, result.errorCode);
+      this.setState({ ...errorHandling, isUpdatingProposal: false });
     }
   };
 
--- a/frontends/web/src/routes/account/send/services.ts
+++ b/frontends/web/src/routes/account/send/services.ts
@@ -1,0 +1,45 @@
+/**
+ * Copyright 2018 Shift Devices AG
+ * Copyright 2023 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


frontends/web/src/routes/account/send/services.ts
```diff
@@ -0,0 +1,51 @@
...
+    errorHandling = { addressError: t('send.error.invalidAddress') };
+    break;
+  case 'invalidAmount':
+  case 'insufficientFunds':
+    errorHandling = { amountError: t(`send.error.${errorCode}`), proposedFee: undefined };
+    break;
+  case 'feeTooLow':
+  case 'feesNotAvailable':
+    errorHandling = { feeError: t(`send.error.${errorCode}`) };
+    break;
```
thisconnect, line 43: Looks like you could just `return` the object instead and drop errorHandling variable.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -338,33 +339,9 @@ class Send extends Component<Props, State> {
...
-        this.setState({ feeError: this.props.t('send.error.feesNotAvailable') });
-        break;
-      default:
-        this.setState({ proposedFee: undefined });
-        if (errorCode) {
-          this.unregisterEvents();
-          alertUser(errorCode, { callback: this.registerEvents });
-        }
-      }
-      this.setState({ isUpdatingProposal: false });
```
thisconnect, line 367: Before this change if there is no result.errorCode it did still run switch/default here.

But with this change there is `else if (result.errorCode)` that only runs if errorCode is truthy. Is that intentional change?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -338,33 +339,9 @@ class Send extends Component<Props, State> {
...
+    } else if (result.errorCode) {
+      const errorHandling = txProposalErrorHandling(this.registerEvents, this.unregisterEvents, result.errorCode);
+      this.setState({ ...errorHandling, isUpdatingProposal: false });
     }
```
thisconnect, line 345: If I read correctly now `this.setState({ isUpdatingProposal: false });` only runs if result.errorCode is truthy. Before this change the state was always updated after the switch.

thisconnect (changes requested):

frontends/web/src/routes/account/send/services.ts
shonsirsha: Ah true, `proposedFee` isn't even an error anyway. So saving it in a variable called errorHandling isn't exactly accurate either. Ty!

frontends/web/src/routes/account/send/send.tsx
shonsirsha: I guess it's the same question with [the other one](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2421/files/4f8053f1e227d809b0c954212ebd74f12d61f0b7#r1427876867). Indeed I should've kept the else, not else if! 

Thanks:)

shonsirsha:
Thanks sm for the review @thisconnect ! Changes made & rebased. PTAL 🙏

frontends/web/src/routes/account/send/services.ts
```diff
@@ -0,0 +1,44 @@
...
+  case 'insufficientFunds':
+    return { amountError: t(`send.error.${errorCode}`), proposedFee: undefined };
+  case 'feeTooLow':
+  case 'feesNotAvailable':
+    return { feeError: t(`send.error.${errorCode}`) };
+  default:
+    if (errorCode) {
+      unregisterEvents();
+      alertUser(errorCode, { callback: registerEvents });
+    }
```
thisconnect, line 42: add a `return { proposedFee: undefined };` after the condition, then it looks to me that it should behave the same as before this refactoring.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -38,6 +38,7 @@ import { route } from '../../../utils/route';
 import { UnsubscribeList, unsubscribe } from '../../../utils/subscriptions';
 import { ConfirmingWaitDialog } from './components/dialogs/confirm-wait-dialog';
 import { SendGuide } from './send-guide';
+import { TProposalError, txProposalErrorHandling } from './services';
```
thisconnect, line 41: nit: are you planing to add more functions to services.ts, else maybe name it proposals.ts ?

thisconnect (changes requested):

frontends/web/src/routes/account/send/services.ts
shonsirsha: Oh good catch, ty!

frontends/web/src/routes/account/send/send.tsx
shonsirsha: yes, there'll be more functions.

frontends/web/src/routes/account/send/services.ts
shonsirsha: actually, looking at it again, returning the proposedFee after the if condition seems to be a bit of a different flow than the original one.

Original: first set the state, then if statement.

Refactor: if statement then set the state.

To me the difference looks harmless. Wanna make sure by testing it but unsure how one can reach the `default` block (with `errorCode`). WDYT? Any concerns?

shonsirsha:
@thisconnect thanks again! Rebased, made the change and added some tests for `txProposalErrorHandling `. PTAL.

frontends/web/src/routes/account/send/services.ts
thisconnect: true, the order is a bit different, but it looks good to me.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -339,32 +340,8 @@ class Send extends Component<Props, State> {
...
-      default:
-        this.setState({ proposedFee: undefined });
-        if (errorCode) {
-          this.unregisterEvents();
-          alertUser(errorCode, { callback: this.registerEvents });
-        }
-      }
-      this.setState({ isUpdatingProposal: false });
+      const errorHandling = txProposalErrorHandling(this.registerEvents, this.unregisterEvents, result.errorCode);
+      this.setState({ ...errorHandling, isUpdatingProposal: false });
```
thisconnect, line 344: I was thinking what it would take to not need to pass `this.registerEvents`, `this.unregisterEvents`. I guess if `txProposalErrorHandling` would return a promise, it could resolve or we could reject on default+errorCode and then move the alertUser out of the promise...

I'm not sure, maybe for a future refactor?

thisconnect (approved):
thanks for adding the tests 👍 

LGTM

frontends/web/src/routes/account/send/send.tsx
shonsirsha: Yes let's do it as a future refactor!

PR #2422, opened 2023-12-11
frontend: add `connected` to TKeystore

benma:
Helpful for watch-only, so the user can see which one is connected.


thisconnect (approved):
nice thanks, untested LGTM

PR #2423, opened 2023-12-11
Limit receive

Beerosagos:
Insured accounts should only receive to native segwit addresses as they are the only addresses supported by Bitsurance. This limitation was already in the code, but in the frontend part. This commit moves the limitation code to the backend, as this is where the logic should be.

### DIFF
--- a/backend/coins/btc/account_test.go
+++ b/backend/coins/btc/account_test.go
@@ -16,6 +16,7 @@
 package btc_test
 
 import (
+	"crypto/sha256"
 	"math/big"
 	"os"
 	"testing"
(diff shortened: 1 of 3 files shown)


backend/coins/btc/account_test.go
```diff
@@ -52,18 +52,32 @@ func TestAccount(t *testing.T) {
...
 	require.NoError(t, err)
 
-	signingConfigurations := signing.Configurations{signing.NewBitcoinConfiguration(
-		signing.ScriptTypeP2WPKHP2SH,
-		[]byte{1, 2, 3, 4},
-		keypath,
-		xpub)}
+	natSegKeypath, err := signing.NewAbsoluteKeypath("m/84'/1'/0'")
+	require.NoError(t, err)
+	natSegXpub, err := hdkeychain.NewMaster(make([]byte, 32), net)
```
benma, line 64: maybe a differnet xpub to the other one?

backend/coins/btc/account_test.go
```diff
@@ -95,4 +109,39 @@ func TestAccount(t *testing.T) {
...
 	require.Equal(t, []*btc.SpendableOutput{}, account.SpendableOutputs())
+
+	// check the number of available addresses for native and wrapped segwit.
+	require.Equal(t, len(account.GetUnusedReceiveAddresses()[0].Addresses), 20)
+	require.Equal(t, *account.GetUnusedReceiveAddresses()[0].ScriptType, signing.ScriptTypeP2WPKHP2SH)
+	require.Equal(t, len(account.GetUnusedReceiveAddresses()[1].Addresses), 20)
+	require.Equal(t, *account.GetUnusedReceiveAddresses()[1].ScriptType, signing.ScriptTypeP2WPKH)
+
+	// Create a new insured account.
+	account2 := btc.NewAccount(
```
benma, line 120: nit: seems like a separate unit test `TestInsuredAccount(...)` would be a bit cleaner/easier

backend/coins/btc/account_test.go
```diff
@@ -95,4 +109,39 @@ func TestAccount(t *testing.T) {
...
+			GetSaveFilename: func(suggestedFilename string) string { return suggestedFilename },
+		},
+		coin, nil,
+		logging.Get().WithGroup("account_test"),
+	)
+	require.NoError(t, account2.Initialize())
+	require.True(t, account2.Synced())
+
+	// verify that the wrapped segwit addresses are not available.
+	require.Equal(t, len(account2.GetUnusedReceiveAddresses()[0].Addresses), 0)
```
benma, line 141: why even provide it at all, why does `GetUnusedReceiveAddresses()` not return only one slice element for native segwit?

Beerosagos: Yeah, you're right, it was a bit lazy as it was before. Fixed.

Beerosagos:
@benma pushed a new update, PTAL :pray:

benma (approved):
tACK

small typo in commit title `accout`

PR #2424, opened 2023-12-11
frontend: migrate rates store to context API

shonsirsha:
We'd like to move away from our self-made `store` HoC & decorator
and use React's Context API instead.

This commit migrates rates store to context API. We created RatesProvider to provide "global" rates data/state & methods for the whole app and we can get them from anywhere within the app using RatesContext.

Also renamed these method & states/variables from the store for
better consistency & clarity:
1. `setActiveFiat()` -> `updateDefaultFiat()`
2. `active` -> `defaultCurrency`
3. `selected` -> `activeCurrencies`

### DIFF
--- a/frontends/web/src/components/rates/rates.tsx
+++ b/frontends/web/src/components/rates/rates.tsx
@@ -14,27 +14,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+import { useContext } from 'react';
 import { Fiat, ConversionUnit, IAmount } from '../../api/account';
-import { BtcUnit } from '../../api/coins';
-import { reinitializeAccounts } from '../../api/backend';
-import { share } from '../../decorators/share';
-import { Store } from '../../decorators/store';
-import { setConfig } from '../../utils/config';
+import { RatesContext } from '../../contexts/RatesContext';
 import { Amount } from '../../components/amount/amount';
-import { equal } from '../../utils/equal';
-import { apiGet } from '../../utils/request';
 import style from './rates.module.css';
 
-export interface SharedProps {
-    active: Fiat;
-    // eslint-disable-next-line react/no-unused-prop-types
-    selected: Fiat[];
-    btcUnit?: BtcUnit;
-    alwaysShowAmounts?: boolean;
-}
-
-export type FiatWithDisplayName = {
+type FiatWithDisplayName = {
   currency: Fiat,
   displayName: string
 }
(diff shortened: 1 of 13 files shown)


frontends/web/src/routes/account/send/send-wrapper.tsx
```diff
@@ -0,0 +1,41 @@
+/**
```
shonsirsha, line 1: This file is a wrapper for our `send` component **and it's a temporary solution / file**.

Reason: `send` needs something from the context store, but since it's still not a functional cmpt yet, we use this wrapper to get it for us (`useContext(RatesContext`) which passes it down to  our actual `send` component as props.

frontends/web/src/routes/settings/components/appearance/activeCurrenciesDropdownSetting.tsx
```diff
@@ -14,14 +14,16 @@
  * limitations under the License.
  */
 
-import { store, SharedProps, formattedCurrencies } from '../../../../components/rates/rates';
+import { formattedCurrencies } from '../../../../components/rates/rates';
 import { SettingsItem } from '../settingsItem/settingsItem';
-import { share } from '../../../../decorators/share';
 import { ActiveCurrenciesDropdown } from '../dropdowns/activecurrenciesdropdown';
 import { useTranslation } from 'react-i18next';
+import { useContext } from 'react';
```
thisconnect, line 21: since you are here, please move imports from react first, then npm such as 'react-i18nex' then ours.

frontends/web/src/contexts/RatesContext.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createContext } from 'react';
+import { Fiat } from '../api/account';
+import { BtcUnit } from '../api/coins';
+
+type RatesContextProps = {
+    defaultFiat: Fiat;
```
thisconnect, line 22: It is usually Fiat, but we also have the option to select 'btc' which is not Fiat just a currency, for that reason I slightly prefer to keep the name `defaultCurrency` instead of `defaultFiat`. But I know we already have this a bit wrong in the Fiat type.

thisconnect (commented):
very nice 🎉 

added some comments, also please rebase

frontends/web/src/contexts/RatesContext.tsx
shonsirsha: Hm, I see your point since BTC is not a fiat. Anyways, I could change this to defaultCurrency, same like selectedFiats to selectedCurrencies.

shonsirsha: Actually I'll change selectedFiats to `activeCurrencies` as this is what the UI displays anyway:

Screenshot 2023-12-15 at 17 04 59

shonsirsha:
Thanks for the review! @thisconnect Changes made, PTAL 🙏

frontends/web/src/contexts/RatesContext.tsx
thisconnect: nice thanks 👍

frontends/web/src/components/rates/rates.tsx
```diff
@@ -141,38 +65,37 @@ type TProvidedProps = {
...
+}: TProvidedProps) {
+
+  const { rotateFiat, defaultCurrency, btcUnit } = useContext(RatesContext);
 
   let formattedAmount = <>{'---'}</>;
   let isAvailable = false;
 
-  var activeUnit: ConversionUnit = active;
-  if (active === 'BTC' && btcUnit === 'sat') {
+  var activeUnit: ConversionUnit = defaultCurrency;
```
thisconnect, line 88: since you are here anyway please change `var` -> `let`

frontends/web/src/components/rates/rates.tsx
```diff
@@ -141,38 +65,37 @@ type TProvidedProps = {
...
+  if (defaultCurrency === 'BTC' && btcUnit === 'sat') {
     activeUnit = 'sat';
   }
 
   // amount.conversions[active] can be empty in recent transactions.
```
thisconnect, line 93: please update the comment `active` -> `defaultCurrency`

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -115,7 +115,7 @@ class Send extends Component<Props, State> {
     isAborted: false,
     isUpdatingProposal: false,
     noMobileChannelError: false,
-    fiatUnit: fiat.state.active,
+    fiatUnit: this.props.activeCurrency,
     coinControl: false,
     btcUnit : 'default',
```
thisconnect, line 120: question: should `btcUnit` come from context in the future (maybe once send is functional component)?

frontends/web/src/routes/account/send/send-wrapper.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const SendWrapper = ({ accounts, code, deviceIDs, devices }: TSendProps) => {
+  const { defaultCurrency } = useContext(RatesContext);
+  return (
+    <Send
+      accounts={accounts}
+      code={code}
+      devices={devices}
+      deviceIDs={deviceIDs}
+      activeCurrency={defaultCurrency}
+    />
```
thisconnect, line 39: just a comment: I guess here it could pass btcUnit, so that send doesn't have to fetch it from config.

but thinking about it, it's probably better if we wait and refactor send to functional component first and then remove the wrapper and use btcUnit from context.

thisconnect (commented):
very nice 👍 

I added some tiny nits and will do some testing now.

frontends/web/src/contexts/RatesProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+      value={{
+        defaultCurrency,
+        activeCurrencies,
+        btcUnit,
+        rotateFiat,
+        selectFiat,
+        updateDefaultFiat,
+        updateRatesConfig,
+        unselectFiat
+      }}
```
thisconnect, line 99: let's keep as is for now, but in the future I think at some point in the future we should rename the function names from Fiat to Currency.

frontends/web/src/routes/account/send/send.tsx
shonsirsha: good point, I think we could use context. Just a bit strange that we use config here and context on `rates.tsx`. Which is the same as the old implementation (config on send but store on `rates`).

We set the value of `btcUnit` in context from config too, so essentially it's the same thing..

Do you think that's intentional? 🤔

frontends/web/src/routes/account/send/send-wrapper.tsx
shonsirsha: oh just read this after I wrote my comment above. Yeah seems like its the sme thing and we could totally pass it down here. I agree on just waiting for finish of func. component first.

shonsirsha:
Thx for the review @thisconnect! Changes made & rebased. Happy testing!

frontends/web/src/routes/account/send/send.tsx
thisconnect: > Do you think using config on send and context/store on rates is intentional? 🤔

I believe btcUnit was added later, but send component was not updated to use it from the custom store.

Lets keep as is and once we have send as functional change it so it doesn't need to get the config, but it can take if from context.

thisconnect (approved):
tested LGTM 👍 

Also tested in qt-osx, and waited a while until the conversion rate updated (took a few minutes)

PR #2425, opened 2023-12-11
frontend: update manage-accounts ui with remember account

thisconnect:
- chanage from watch wallet to remember wallet
- removed outdated translations and replaced with rememberWallet
- moved enable/disable account to edit dialog
- comment code that allowed to enable watching one account, if many users request this we might bring it back
- changed manage-accounts ui to use Grid/Column component

### DIFF
--- a/frontends/web/src/routes/settings/components/manage-accounts/watchonlySetting.tsx
+++ b/frontends/web/src/routes/settings/components/manage-accounts/watchonlySetting.tsx
@@ -17,13 +17,13 @@
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Toggle } from '../../../../components/toggle/toggle';
-import { SettingsItem } from '../settingsItem/settingsItem';
 import * as backendAPI from '../../../../api/backend';
 import * as accountAPI from '../../../../api/account';
 import { useLoad } from '../../../../hooks/api';
 import { getConfig } from '../../../../utils/config';
 import { Dialog, DialogButtons } from '../../../../components/dialog/dialog';
-import { Button } from '../../../../components/forms';
+import { Button, Label } from '../../../../components/forms';
+import style from './watchonlySettings.module.css';
 
 type Props = {
   keystore: accountAPI.TKeystore;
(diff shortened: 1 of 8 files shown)


frontends/web/src/routes/settings/components/manage-accounts/watchonlySetting.tsx
```diff
@@ -74,32 +74,25 @@ export const WatchonlySetting = ({ keystore }: Props) => {
...
-                    checked={watchonly}
-                    disabled={disabled}
-                    onChange={toggleWatchonly}
-                  />
-                ) :
-                null
-            }
-          </>
-        }
-      />
```
thisconnect, line 102: `<SettingsItem` is not needed here.

frontends/web/src/routes/settings/manage-accounts.module.css
```diff
@@ -21,20 +38,38 @@
 .acccountLink {
     align-items: center;
     display: flex;
-    flex-basis: calc(100% - 180px);
-    flex-grow: 0;
+    flex-basis: calc(100% - 100px);
```
thisconnect, line 41: should give more space for long account names on small screen, I'll add to the commit message.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -123,41 +115,41 @@ class ManageAccounts extends Component<Props, State> {
...
+    //         }}
+    //       />
+    //     </div>
+    //     <p className={style.watchOnlyNote}>{t('manageAccounts.watchAccountDescription')}</p>
+    //     {
+    //       !currentlyEditedAccount.watch && <div className={style.watchOnlyAccountHidden}>
+    //         <p>{t('manageAccounts.accountHidden')}</p>
+    //       </div>
+    //     }
+    //   </div>);
```
thisconnect, line 152: we discussed keeping the code and maybe bring it back if many users need this.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -258,6 +250,10 @@ class ManageAccounts extends Component<Props, State> {
           }
           return;
         }
+        const account = accounts.find(({ code }) => currentlyEditedAccount.code === code);
+        if (currentlyEditedAccount.active !== account?.active) {
+          this.toggleAccount(currentlyEditedAccount.code, currentlyEditedAccount.active);
+        }
```
thisconnect, line 256: I don't like this, but now that the enable/disable account toggle is in the dialog, it should not fire right away but only after the user clicked the dialog update button.

frontends/web/src/routes/settings/manage-accounts.tsx
```diff
@@ -315,6 +315,28 @@ class ManageAccounts extends Component<Props, State> {
...
+                            checked={currentlyEditedAccount.active}
+                            className={style.toggle}
+                            id={currentlyEditedAccount.code}
+                            onChange={(event) => {
+                              event.target.disabled = true;
+                              this.setState({
+                                currentlyEditedAccount: {
+                                  ...currentlyEditedAccount,
+                                  active: event.target.checked,
+                                }
```
thisconnect, line 335: similar to how the name is temporary stored in currentlyEditedAccount, we can do the same here until the dialog is either closed or the user clicked the update button in the dialog.

frontends/web/src/style/layout.css
```diff
@@ -212,67 +212,6 @@
...
-.box .buttons:not(.ignore) > *:not(:first-child) {
-    margin-top: var(--space-quarter);
-}
-
-.box .buttons:not(.ignore) button,
-.box .buttons:not(.ignore) button,
-.box .buttons:not(.ignore) a {
-    width: 100%;
-}
-
```
thisconnect, line 275: not needed anymore now that we use Grid / Column

thisconnect:
General note: currently the manage-accounts view is not subscribed to to accounts changes. And might not update, I noticed a few times. I haven't looked into adding a subscription for accounts yet, but that should probably be in a different PR.

thisconnect:
Screen Shot 2023-12-11 at 18 56 26

thisconnect:
Screen Shot 2023-12-11 at 18 55 57

thisconnect:
rebased

Beerosagos (approved):
LGTM with a few nits

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: > this way we also call renameAccount when the name has not changed

true, as said I don't like it :)

reason it's messy it was 2 separate UI elements that are now combined into 1 dialog and updated together. 

One alternative could be to check first if either name or toggle changed and then only call the ones needed,.. but this make the function complicated. So I prefer to keep the logic for now. 

I change the function to updateAccount.

frontends/web/src/routes/settings/manage-accounts.tsx
thisconnect: commented

PR #2426, opened 2023-12-11
frontend: improve connected keystore styling 

thisconnect:
Added small connected icon as a badge so it is easy to see.

Did following modifications to the badge:
- badge style can now have transparent colors which look the same
  in light mode but now also supports dark mode due to transparent
  color working on both backgrounds
- made children optional
- added CSS class is children are missing
- fixed a bug when adding className prop, before it replaced all
  classes and broke the design

Also removed 'Accounts' and added 'Account' in the sidebar.


thisconnect:
Screen Shot 2023-12-11 at 21 22 20

thisconnect:
Screen Shot 2023-12-11 at 21 22 49

thisconnect:
Screen Shot 2023-12-11 at 21 23 43

frontends/web/src/components/badge/badge.tsx
```diff
@@ -14,22 +14,33 @@
...
 import style from './badge.module.css';
 
 type TBadgeStyles = 'success' | 'warning'; // TODO: not yet implemented 'info' | 'danger'
 
 type TProps = {
-  children: ReactNode;
+  children?: ReactNode;
   type?: TBadgeStyles;
-}
+  variation?: 'icon-only';
```
shonsirsha, line 25: Instead of being a prop, do you think it makes sense if we derive this from the `children` prop instead?

E.g, if children is undef, then it's icon-only. Otherwise, we shouldnt apply icon-only class.

Bc applying `variation=icon-only` w/ `children` is never be the case if i'm not mistaken.

frontends/web/src/components/badge/badge.tsx
```diff
@@ -14,22 +14,33 @@
...
+  withIcon?: (props: { className: string }) => ReactElement;
```
shonsirsha, line 26: what about renaming it to just `icon`? `withIcon` sounds a bit `boolean`-y

shonsirsha (approved):
tested lgtm!
Just some nits and questions, but not a merge blocker.

frontends/web/src/style/variables.css
```diff
@@ -2,8 +2,8 @@
     /* colors */
     --color-darkgreen: #4f9879;
     --color-green: #7aba7a;
-    --color-lightgreen: #aecfc1;
-    --color-palegreen: #dceae4;
+    --color-lightgreen: rgba(79, 152, 121, .33); /* #aecfc1; */
+    --color-palegreen: rgba(79, 152, 121, .2); /* #dceae4; */
```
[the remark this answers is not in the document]
thisconnect: Yes exactly, see commit message

> - badge style can now have transparent colors which look the same
  in light mode but now also supports dark mode due to transparent
  color working on both backgrounds

frontends/web/src/components/badge/badge.tsx
thisconnect: ack, will change this.

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -203,30 +204,38 @@ class Sidebar extends Component<Props> {
...
               <div className={style.sidebarHeaderContainer}>
                 <span className={style.sidebarHeader} hidden={!keystore.accounts.length}>
-                  {t('sidebar.accounts')} - {keystore.keystore.name} {keystore.keystore.connected ? '- CONNECTED' : null}
+                  {`${t('sidebar.account')} - ${keystore.keystore.name}  `}
+                  {keystore.keystore.connected ? (
+                    <Badge
+                      className={style.activeKeystoreIcon}
+                      icon={props => <USBSuccess {...props} />}
+                      type="success"
+                      title={t('sidebar.keystoreConnected')} />
```
thisconnect, line 224: Actually we might need this badge in other places (manage accounts), I'll rename the key to something not sidebar specific.

frontends/web/src/components/badge/badge.tsx
```diff
@@ -14,22 +14,31 @@
...
+  icon,
   type = 'success',
+  ...props
 }: TProps) => {
+  const withChildrenStyle = children !== undefined ? style.withChildren : '';
+  const iconOnlyStyle = (children === undefined && icon) ? style.iconOnly : '';
   return (
-    <span className={`${style.badge} ${style[type]}`}>
+    <span
+      className={`${style.badge} ${style[type]} ${withChildrenStyle} ${iconOnlyStyle} ${className}`}
```
shonsirsha, line 39: To prevent adding an unwated `.undefined` class:

```suggestion
      className={`${style.badge} ${style[type]} ${withChildrenStyle} ${iconOnlyStyle} ${className || ''}`}
```

frontends/web/src/components/badge/badge.tsx
```diff
@@ -14,22 +14,31 @@
...
+      {...props}>
+      {icon && icon({ className: style.badgeIcon })}
```
shonsirsha, line 41: If it's not too difficult, could we add a tiny space like 4px between the icon & text (when theres one)?

IMO looks a bit neater with it:

with extra space:

Screenshot 2023-12-12 at 14 04 54

without:

Screenshot 2023-12-12 at 14 04 16

shonsirsha (approved):
👍 🙏

frontends/web/src/components/badge/badge.tsx
thisconnect: > IMO looks a bit neater with

It has a small space, the main reason I didn't want to add margin-left is that in the case when there is a long name and the icon breaks on a 2nd line... but now that we don't have the "Account" aanymore I think it makes sense to add it again. I will update the PR.

shonsirsha: tysm!

thisconnect: Screen Shot 2023-12-12 at 14 59 48

and in cases where it shows root fingerprint

Screen Shot 2023-12-12 at 15 00 11

thisconnect:
squashed into 1 commit. (the others didn't make sense anymore)

PR #2428, opened 2023-12-12
frontend/account: fix bitsurance badge on uninsured accounts

Beerosagos:
Bitsurance badge was not correctly removed navigating from an insured account to an uninsured one. This fixes the issue.


thisconnect:
> Bitsurance badge was not correctly removed navigating from an insured account to an uninsured one. This fixes the issue.

nit: how about moving this into commit message? 😇

thisconnect (approved):
untested LGTM with nit

PR #2429, opened 2023-12-12
frontend: subscribe to accounts in manage-accounts view

thisconnect:
Without watch-only manage-accounts view could load accounts once and did not need to update the state anymore. This changed as with watch-only the accounts should update, i.e. when the device is unplugged but not remembered.

App component already subscribed to accounts, so this commit only passes the accounts to manage-accounts component, and changes accounts from state to props.


Beerosagos (approved):
LGTM!

PR #2431, opened 2023-12-13
backend: fix regression related to updating persisted accounts

benma:
It was supposed to only be applied to the accounts that belong to the current keystore, but 52bf446c7e1d09ac95c7c4d61e9f784fe028e1c1 added a regression and it was applied to all persisted accounts.

The `setWatch()` thing can be removed - it was there before when the watchonly setting was global and accounts would need to be marked as `Watch` when the keystore is connected. After we moved it to be a keystore setting, it can only be enabled after connecting the keystore anyway in 'Manage accounts'.

### DIFF
--- a/backend/accounts_test.go
+++ b/backend/accounts_test.go
@@ -1523,4 +1523,48 @@ func TestWatchonly(t *testing.T) {
 		checkShownAccountsLen(t, b, 3, 6)
 	})
 
+	// Adding new accounts after the keytore has been connected: new account is watched if the
+	// keystore is already watched.
+	t.Run("", func(t *testing.T) {
+		b := newBackend(t, testnetDisabled, regtestDisabled)
+		defer b.Close()
+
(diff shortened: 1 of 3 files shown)


backend/accounts_test.go
```diff
@@ -1523,4 +1523,49 @@ func TestWatchonly(t *testing.T) {
...
+
+		rootFingerprint, err := ks.RootFingerprint()
+		require.NoError(t, err)
+
+		b.registerKeystore(ks)
+		checkShownAccountsLen(t, b, 3, 3)
+		require.NoError(t, b.SetWatchonly(rootFingerprint, true))
+
+		// An account may already have been addded as part of autodiscover, so we add two.
+
```
Beerosagos, line 1542: nit: remove empty line?

Beerosagos (approved):
partially tested ACK with nit

PR #2432, opened 2023-12-13
backend/keystore: mark owned addresses to improve BitBox verification

Beerosagos:
In recent 9.15.0 BitBox02 firmware release we introduced a specific screen to verify owned addresses in send-to-self transactions. This commit modifies the transaction data to exploit this new feature.

Note: This seems to work nicely with native segwit, but using taproot or wrapped segwit the firmware returns an 'invalid input' error.


Beerosagos:
Closing, #2451 was merged instead

PR #2433, opened 2023-12-13
frontend: show which wallet is connected on manage-accounts

thisconnect:
So the user can see which wallet is connected.


thisconnect:
Screen Shot 2023-12-13 at 20 38 30

or further down:

Screen Shot 2023-12-13 at 20 38 38

Beerosagos (approved):
tACK

PR #2434, opened 2023-12-14
frontend: make locize-pull

benma:


frontends/web/src/locales/en/app.json
```diff
@@ -537,7 +537,7 @@
     },
     "expert": {
       "factoryReset": {
-        "description": "Reset your device to factory settings. This deletes the wallet from your BitBox02!",
+        "description": "Reset your device to factory defaults - this will erase your BitBox02 wallet!",
```
thisconnect, line 540: Nit:

I am not sure what this change was about. Before it was more clear IMO that the wallet is only erased from the device. Now it could be misinterpreted that the wallet will be gone forever.
Also it doesn't really reset to factory defaults. Factory default has no firmware.

thisconnect (approved):
LGTM with content nit

frontends/web/src/locales/en/app.json
benma: I agree, but I think there has already been some bike-shedding on this :sweat_smile:

benma: I reverted it

frontends/web/src/locales/de/app.json
```diff
@@ -136,8 +140,9 @@
...
-      "title": "Recovery-Wörter anzeigen"
+      "description": "Dir werden 24 Recovery-Wörter angezeigt, die ein Backup deiner Wallet sind. Schreibe diese auf ein Stück Papier. \n\n<strong>Bewahre diese Wörter nicht digital auf und mache kein Foto von ihnen.</strong>\n\n<strong>Sprich die Wörter nicht laut aus.</strong>\n\n<strong>Dieses Backup ist nicht Passwort geschützt.</strong>\n\nNachdem du die Wörter aufgeschrieben hast wirst du gefragt werden sie nochmals zu bestätigen.\n",
+      "title": "Recovery-Wörter anzeigen",
+      "warning": "<strong>Teile deine Wiederherstellungswörter niemals mit irgendjemandem.</strong> Deine Wiederherstellungswörter ermöglichen vollen Zugriff zu deinem Wallet . Wenn dich jemand nach deinen Wiederherstellungswörtern fragt, handelt es sich um einen Betrüger. Gebe diese nicht weiter!"
```
benma, line 145: @jstrnbrg first `Recovery-Wörter` then `Wiederherstellungswörter`? Please check that it is consistent everywhere.

And then: the English source does not say `24` (cause it could also be `12`).

frontends/web/src/locales/de/app.json
```diff
@@ -485,11 +491,16 @@
     "infoWhenPaired": "Zuerst auf deinem gepaarten Mobiltelefon und dann auf deiner BitBox"
   },
   "confirmOnDevice": "Bitte bestätige auf deinem Gerät.",
+  "connectKeystore": {
+    "promptNoName": "Bitte verbinde deine BitBox02 um fortzufahren",
+    "promptWithName": "Bitte verbinde deine BitBox02 mit dem Namen „{{name}}\" um fortzufahren"
```
jstrnbrg, line 496: @benma is this syntax valid?

PR #2435, opened 2023-12-14
frontend/utils: sort getAccountsByKeystore map

Beerosagos:
The input accounts list seems to be possibly ordered differently in different places in the code (e.g. sidebar vs manage account page). This ensures that the list of accounts grouped by keystore always has a consistent order, and also keeps keystores with same name (e.g. create using different passphrases) to be close to each other in the list.

### DIFF
--- a/frontends/web/src/routes/account/utils.ts
+++ b/frontends/web/src/routes/account/utils.ts
@@ -112,7 +112,13 @@ export function getAccountsByKeystore(accounts: IAccount[]): TAccountsByKeystore
     }
     acc[key].accounts.push(account);
     return acc;
-  }, {} as Record<string, TAccountsByKeystore>));
+  }, {} as Record<string, TAccountsByKeystore>))
+    // sort the output array as we noticed that somehow the input accounts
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/account/utils.ts
```diff
@@ -102,7 +102,13 @@ export type TAccountsByKeystore = {
 
 // Returns the accounts grouped by the keystore fingerprint.
 export function getAccountsByKeystore(accounts: IAccount[]): TAccountsByKeystore[] {
-  return Object.values(accounts.reduce((acc, account) => {
+  const sortedAccounts = accounts
```
benma, line 105: why don't you sort the `TAccountsByKestore[]` result instead, if the keystore's are out of order? (i assume the accounts inside the keystores are ordered properly).

then you don't have to worry about deep copies either, and not owrry about if the sort is a stable sort and retains the order of the accounts.

Beerosagos:
@benma rebased and updated after your comment, PTAL :pray:

benma (approved):
utACK

PR #2439, opened 2023-12-18
Frontend fix sidebar connected icon

thisconnect:
### prevent jumping icon when device is unplugged

The connected icon is slightly bigger than the small text fontsize.
This caused a tiny jump, when the device is disconnected.

Fixed by having the icon always in place and just turn on and off
the visibility.

### fix centered badge icon

The icon in the badge was not perfectly centered. Reason is
that the icon has an uneven width (9px) so it rounds the
position to be 1px off.

Changed so that the icon is now a max of 10px wide. Ideally
all icons in the future that are used in badges should have
an even width.

### DIFF
--- a/frontends/web/src/components/badge/badge.module.css
+++ b/frontends/web/src/components/badge/badge.module.css
@@ -18,7 +18,7 @@
 }
 
 .badgeIcon {
-  max-width: 9px;
+  max-width: 10px;
 }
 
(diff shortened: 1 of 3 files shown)


thisconnect:
Screen Shot 2023-12-18 at 10 58 22

the icon is not slightly bigger, the badge has still same size.

PR #2440, opened 2023-12-18
frontend: fix routing when removing account

benma:
If you are inside one account that is removed when the device/keystore is disconnected, it always routes to `/`, even if there are other watchonly accounts. This commit fixes it to route to the portfolio if there is at least one account.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -83,8 +83,7 @@ class App extends Component<Props, State> {
         }));
       }),
       syncAccountsList(accounts => {
-        const oldAccountsLen = this.state.accounts.length;
-        this.setState({ accounts }, () => this.maybeRoute(oldAccountsLen));
+        this.setState({ accounts }, () => this.maybeRoute());
       }),
       syncDeviceList((devices) => {
         this.setStateWithDeviceList({ devices });
@@ -94,7 +93,6 @@ class App extends Component<Props, State> {
 
   private setStateWithDeviceList(newState: Partial<State>) {
     const oldDeviceIDList = Object.keys(this.state.devices);
-    const oldAccountsLen = this.state.accounts.length;
     this.setState(currentState => ({ ...currentState, ...newState }), () => {
       const newDeviceIDList: string[] = Object.keys(this.state.devices);
       // If a device is newly connected, we route to the settings.
@@ -114,7 +112,7 @@ class App extends Component<Props, State> {
           return;
         }
       }
-      this.maybeRoute(oldAccountsLen);
+      this.maybeRoute();
     });
   }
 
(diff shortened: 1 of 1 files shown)


frontends/web/src/app.tsx
```diff
@@ -149,13 +147,13 @@ class App extends Component<Props, State> {
       route('/', true);
       return;
     }
-    // if on an account that isnt registered route to /
+    // if on an account that isn't registered route to /
     if (inAccounts && !accounts.some(account => currentURL.startsWith('/account/' + account.code))) {
       route('/', true);
```
benma, line 152: @thisconnect this is executed first if you are in an account which is removed when you unplug.

frontends/web/src/app.tsx
```diff
@@ -149,13 +147,13 @@ class App extends Component<Props, State> {
...
       return;
     }
-    // if on index page and we go from zero accounts to at least 1 account, route to /account-summary
-    if (isIndex && oldAccountsLen === 0 && accounts.length) {
+    // if on index page and have at least 1 account, route to /account-summary
+    if (isIndex && accounts.length) {
```
benma, line 156: @thisconnect this change looks a little risky :P we should test with the routing test table to see if this breaks something else

thisconnect: I'm on it, fyi the routing tests only cover none-watchonly situations.

thisconnect:
ok it took me a while to reproduce the issue:

- there needs to be a remembered wallet
- the current visible account has to be one with the connected device but is not watched/remembered

with that when I unplug it goes to waiting view.

@strmci and me did the none-watch routing testing, no regression.

thisconnect:
@benma would you mind updating the commit message?

> If you are inside one account that is removed when the device/keystore
is disconnected, it always routes to `/`, 

I believe it doesn't "always" route to `/` only if this is true:

"the current visible account has to be one with the connected device but is not watched/remembered"

also please change  `/` to waiting view as slash is not obvious which view it is.

thisconnect (approved):
nice 👍  
tested LGTM with small commit message nit.

tested none-watchonly routing for regressions. We don't yet have routing tests for watch-only.

benma:
> ok it took me a while to reproduce the issue:
> 
>     * there needs to be a remembered wallet
> 
>     * the current visible account has to be one with the connected device but is not watched/remembered
> 
> 
> with that when I unplug it goes to waiting view.
> 
> @strmci and me did the none-watch routing testing, no regression.

Sorry! I thought this description was clear :(

```
If you are inside one account that is removed when the device/keystore is disconnected, it always routes to /, even if there are other watchonly accounts.
```

> I believe it doesn't "always" route to / only if this is true:
> "the current visible account has to be one with the connected device but is not watched/remembered"

This is what the message says though, just with different words (`If you are inside one account that is removed when the device/keystore is disconnected` means this).

I added the info that `/` is the waiting view to the commit msg.

PR #2441, opened 2023-12-18
frontend/summary,account: sdcard check false positive

benma:
If one inserts a BitBox02 while in the account summary or account, `useSDCard(devices);` returns true even if there is no sdcard, because:

1. It is called before the BitBox02 has established the communication (paired channel)
2. The checkSDCard endpoints returns an error
3. The `useSDCard()` assumes the result is `boolean` always, so it interprets the non-zero error as `true`, which leads to the warning banner being shown even if the sdcard is not inserted.

This is a quick fix to not show the banner in this case by returning `false` instead of an error.

The proper fix is postponed, and involes rewriting `useSDCard` to listen on the `statusChanged` event of the BitBox02 to only chekc the sdcard once the device is paired (or do something similar in the backend and make an `sdcard` event).


Beerosagos (approved):
utACK

PR #2443, opened 2023-12-18
android: bump versionCode

benma:
Submitted the app bundle for 46 already, but not gonna release it due to a bug we found and fixed in 8c190e94e.


thisconnect (approved):
LGTM

PR #2444, opened 2023-12-19
frontend: fix enable account toggle label text

thisconnect:
Fixes label text of enable/disable account toggle.


shonsirsha (approved):
tested LGTM.

PR #2447, opened 2023-12-19
frontend: remove 'Back' text on MobileHeader

shonsirsha:
the "Back" text sometimes crashed into the page's title/header text causing unclarity in the UI, so we'd like to remove it.

Also adjusted the position of the header text so it centres better.

Before:
Screenshot 2023-12-19 at 12 42 51

After:
Screenshot 2023-12-19 at 12 46 33

### DIFF
--- a/frontends/web/src/routes/settings/components/mobile-header.module.css
+++ b/frontends/web/src/routes/settings/components/mobile-header.module.css
@@ -38,7 +38,7 @@
 }
 
 .withGuide .headerText {
-    left: calc(50% + 37px);
+    left: calc(50% + 22px);
 }
 
 @media (max-width: 768px) {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/settings/components/mobile-header.module.css
```diff
@@ -38,7 +38,7 @@
 }
 
 .withGuide .headerText {
-    left: calc(50% + 37px);
+    left: calc(50% + 22px);
```
shonsirsha, line 41: 22 is the width of the back btn without the margin (which is invisible). This is to visually centre the header text.

shonsirsha:
Closing in favour of https://github.com/digitalbitbox/bitbox-wallet-app/pull/2476

PR #2450, opened 2023-12-20
frontend: apply `displayedValue` class conditionally to rm space

shonsirsha:
Unuseful space was added due to always having `.displayedValue` class regardless when there's displayed value or not for `SettingsItem` (visible on mobile). We should only add it if there's a displayed value.

Before:
CKGM

After:

jkaem

After for stuff with "displayed value" (no change):
x


thisconnect (approved):
untested LGTM 👍

PR #2451, opened 2023-12-20
backend/bitbox02: mark recipient address as ours if they are ours

benma:
This will make the BitBox02 show 'This BitBox02' before the address verification when making a self-send transaction.

There is a annoying edge case that is fixed by moving the `firmware.BTCSignNeedsPrevTxs()` check below processing the outputs:

- Previous transactions are not needed if all inputs are Taproot.
- The firmware determines this by checking if all provided script configs are Taproot - if it is, it does not require the previous transcations, otherwise it does.
- Unfortunately, change outputs and now also recpient outputs can add a script config for its own type. For changes it didn't change anything as we only added Taproot changes if all inputs were Taproot.
- This means that if you send-to-self using only Taproot inputs but send to a non-taproot (native segwit) recipient address of the same account, the non-taproot script config is added to the list and the BB02 will require previous transactions, even though they are not technically needed.

We just live with this edge case (send from taproot to a native segwit recipient address of the same account) - it works, but just performs more work by sending the previous transactions.

In all other cases it works as expected.

### DIFF
--- a/backend/devices/bitbox02/keystore.go
+++ b/backend/devices/bitbox02/keystore.go
@@ -24,6 +24,7 @@ import (
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/blockchain"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/types"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/btc/util"
 	coinpkg "github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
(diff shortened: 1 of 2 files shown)


backend/devices/bitbox02/keystore.go
```diff
@@ -396,16 +364,21 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
 		default:
 			return errp.Newf("unsupported output type: %v", address)
 		}
-		changeAddress := btcProposedTx.TXProposal.ChangeAddress
-		isChange := changeAddress != nil && bytes.Equal(
-			changeAddress.PubkeyScript(),
-			txOut.PkScript,
-		)
+		outputAddress := btcProposedTx.GetAddress(blockchain.NewScriptHashHex(txOut.PkScript))
+		isChange := outputAddress != nil && outputAddress.IsChange()
```
benma, line 368: could actually also use `btcProposedTx.TXProposal.ChangeAddress` to check if it is change, maybe that would be better, but this way we could delete `ChangeAddress` from the struct :shrug: either is okay I think.

benma: switched back to using ChangeAddress for simplicity

PR #2453, opened 2023-12-20
backend/accounts_test: fix race condition

benma:
The `go backend.maybeAddHiddenunusedAccounts()` call in `registerKeystore()` could execute before or after the first `CreateAndPersistAccountConfig()` in that Watchonly unit test. If it happened after, then there would already be two accounts and we would fail at

    require.Nil(t, b.Config().AccountsConfig().Lookup(expectedNewAccountCode2))


benma:
@Beerosagos 

I am not very happy with the solution in this commit, as it is a bit arbitrary that `registerKeystore()` adds one batch of discovered accounts and not more than one, but I am not sure how to structure the autodiscover code so that unit tests become simpler either.  I got ourselves into a bit of a mess here :innocent: will need to think a bit harder about how to make all of this work nicer and easier to understand.

Beerosagos (approved):
LGTM

PR #2456, opened 2023-12-22
frontend/account: prompt keystore for ETH before going to send screen

benma:
BitBox02 fw <9.15.0 can only sign legacy ETH
transactions (EIP155). After 9.15.0, it can also sign EIP1559 transactions. The app decides which one to create in the send screen depending on which keystore is used, so we need to already have it connected when entering the send screen.

We do this only for Ethereum, as for Bitcoin/Litecoin, this is not needed.

Alternaties dismissed:
- Prompt keystore on first send form modification - bad UX
- Force fw upgrade when sending ETH and only do EIP1559 - bad UX

### DIFF
--- a/frontends/web/src/routes/account/actionButtons.tsx
+++ b/frontends/web/src/routes/account/actionButtons.tsx
@@ -14,29 +14,47 @@
  * limitations under the License.
  */
 
+import { MouseEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
+import { useNavigate } from 'react-router';
 import { WalletConnectLight } from '../../components/icon';
 import { useMediaQuery } from '../../hooks/mediaquery';
-import { IAccount } from '../../api/account';
+import { connectKeystore, AccountCode, CoinCode, IAccount } from '../../api/account';
 import { isEthereumBased } from './utils';
 import style from './account.module.css';
 
 type TProps = {
-    canSend?: boolean;
-    code: string;
-    exchangeBuySupported?: boolean;
-    account: IAccount;
+  canSend?: boolean;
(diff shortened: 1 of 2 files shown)


thisconnect (approved):
tested LGTM

only tiny behavior nit is, when the user accidentally unlocks the wrong device or passphrase, the view changes in the background to account-summary.

frontends/web/src/routes/account/actionButtons.tsx
```diff
@@ -14,29 +14,51 @@
...
+      // Proceed to the send screen.
+      navigate(sendLink);
+    }
+  };
+
   return (
     <div className={`${style.actionsContainer} ${walletConnectEnabled ? style.withWalletConnect : ''}`}>
       {canSend ? (
-        <Link key="sendLink" to={`/account/${code}/send`} className={style.send}>
+        <Link key="sendLink" to={sendLink} className={style.send} onClick={maybeRouteSend}>
```
thisconnect, line 61: small note: if I read the logic in maybeRouteSend correctly, the normal Link isn't used at all. Also for Bitcoin it uses navigation function in maybeRouteSend.

You could change the `Link` component to `span` here and remove `to` prop, to simplify.

Alternatively: you could have a condition in `onClick` prop: `onClick={isEthereumBased(coinCode) ? maybeRouteSend}`

benma: If I do a `span` then the style is different, as the others are `a`. E.g. the mouse pointer is different. 

I went with your second option :+1:

thisconnect: nice 👍

thisconnect (approved):
tested LGTM

PR #2458, opened 2023-12-25
frontend: remove old scanner zxing/library

thisconnect:
In the following commit qr-scanner was added, but @zxing/library was not removed: 35cf802331e6e285361cb3fa06c3d85cfe1e3977


shonsirsha (approved):
LGTM, thanks for this! 🙏

PR #2460, opened 2023-12-25
frontend: type address signing

thisconnect:
This is the last component that used the subscribe decorator.

React class componentDidUpdate is not necessary anymore, this was called when the props changed and provided prevProps. Changed to state / subscription, which means we don't have access to prevProps which is rewritten using currentState and setState callback. This part of the code is responsible for managing the account to use.


frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,47 +38,55 @@ const Banner = ({ children }: TProps) => (
...
+        )
+      ) {
+        shouldUpdateAccountCodeDefault = true;
+      }
+      return { aopp };
+    }, () => { // callback when state did update
+      if (shouldUpdateAccountCodeDefault) {
+        this.setAccountCodeDefault();
+      }
+    });
```
thisconnect, line 85: This should work the same. 😇

thisconnect:
I'll add a 2nd commit to remove `decorators/subscribe.ts`.

thisconnect:
There is a few things I would simplify in this code, and ofc rewrite to functional.. but I thougt it's better to keep the changes in this PR minimal.

Also with this we can remove subscribe decorator.

shonsirsha (approved):
LGTM!

PR #2461, opened 2023-12-26
frontend: default `show` as `false` for status component

shonsirsha:
defaulting `show` to `true` may cause flickering if the status
is `dismissible` and has previously been dismissed.

This is due to the status getting rendered (since `show` is true) by
default, several seconds/ms before `checkConfig` finishes.

In other words, the status gets rendered for a very short time
before it disappears after `checkConfig` finishes, giving the
appearance of a flicker.

Before:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/62f0cb88-d211-422d-84e3-36d454fb1021

After:

https://github.com/digitalbitbox/bitbox-wallet-app/assets/28676406/6553aabb-dbb2-4588-a010-9b0d098a7d48

### DIFF
--- a/frontends/web/src/components/status/status.tsx
+++ b/frontends/web/src/components/status/status.tsx
@@ -40,7 +40,7 @@ export const Status = ({
   className,
   children,
 }: TPRops) => {
-  const [show, setShow] = useState(true);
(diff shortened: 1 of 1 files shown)


thisconnect (approved):
tested LGTM 👍

PR #2462, opened 2023-12-27
frontend: use AccountCode everywhere

benma:
For clarity, we should use the named type instead of `string`.


thisconnect (approved):
LGTM with small nit

PR #2463, opened 2023-12-28
 frontend: convert bitbox02bootloader.tsx to a function component 

benma:


frontends/web/src/api/bitbox02bootloader.ts
```diff
@@ -41,7 +42,7 @@ export type TVersionInfo = {
 
 export const getVersionInfo = (
   deviceID: string,
-): Promise<TStatus> => {
+): Promise<TVersionInfo> => {
```
thisconnect, line 45: 👍  (looks like this was not used anywhere before)

frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.tsx
```diff
@@ -14,165 +15,109 @@
...
     );
   }
-}
-
-const loadHOC = load<LoadedProps, BitBox02BootloaderProps & TranslateProps>(({ deviceID }) => ({ versionInfo: 'devices/bitbox02-bootloader/' + deviceID + '/version-info' }))(BitBox02Bootloader);
-const HOC = translate()(loadHOC);
-export { HOC as BitBox02Bootloader };
+  return (
+    <CenteredContent>
+      {getDarkmode() ? <BitBox02Inverted /> : <BitBox02 />}
```
thisconnect, line 116: nit: with functional component we could now use darkmode hook.

```js
import { useDarkmode } from '../hooks/darkmode';

```

thisconnect (approved):
untested LGTM 🎉

frontends/web/src/api/bitbox02bootloader.ts
benma: it was me in the previous commit in this same PR :facepalm: fixed in the right commit :)

frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.tsx
benma: great! done

PR #2467, opened 2023-12-28
frontend: convert bitbox02.tsx to a functional component

benma:
The status does not need to be synced, as the status does not change from or to `initialized` while the user is in the Manage Device screen.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bitbox02.tsx
@@ -15,59 +15,27 @@
  * limitations under the License.
  */
 
-import { Component } from 'react';
-import { getStatus, TStatus } from '../../../api/bitbox02';
+import { getStatus } from '../../../api/bitbox02';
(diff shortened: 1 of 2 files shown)


thisconnect (approved):
thank you LGTM 🎉

frontends/web/src/routes/device/bitbox02/bitbox02.tsx
```diff
@@ -15,59 +15,27 @@
...
-  constructor(props: Props) {
-    super(props);
-    this.state = {
-      status: '',
-    };
-  }
+export const BitBox02 = ({ deviceID, deviceIDs, hasAccounts }: TProps) => {
+  const status = useSync(
+    () => getStatus(deviceID),
+    cb => statusChanged(deviceID, () => {
```
benma, line 32: @thisconnect found a way to use `useSync` anyway, which looks better as it contains the mounted ref stuff too. PTAL.

thisconnect (approved):
LGTM

PR #2468, opened 2023-12-28
backend/bitbox02: add /invoke-bip85 handler

benma:
To invoke the BIP-85 workflow on the device (only available from v9.16.0).


thisconnect (approved):
untested LGTM thanks

PR #2469, opened 2023-12-28
frontend/bootloader: use replace instead of reload in getStatus

benma:
And also load the status when the component loads, otherwise the view is wrong if the user exists and re-enters the view (until a new push notification arrives).

### DIFF
--- a/frontends/web/src/api/bitbox02bootloader.ts
+++ b/frontends/web/src/api/bitbox02bootloader.ts
@@ -31,6 +31,14 @@ export const getStatus = (
   return apiGet(`devices/bitbox02-bootloader/${deviceID}/status`);
 };
 
+export const syncStatus = (deviceID: string) => {
+  return (
+    cb: TSubscriptionCallback<TStatus>
+  ) => {
+    return subscribeEndpoint(`devices/bitbox02-bootloader/${deviceID}/status`, cb);
+  };
+};
+
 export type TVersionInfo = {
   // Indicates whether the device has any firmware already installed on it.
   // It is considered "erased" if there's no firmware, and it also happens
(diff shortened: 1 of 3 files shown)


frontends/web/src/api/bitbox02bootloader.ts
```diff
@@ -31,6 +31,14 @@ export const getStatus = (
   return apiGet(`devices/bitbox02-bootloader/${deviceID}/status`);
 };
 
+export const syncStatus = (deviceID: string) => {
+  return (
+    cb: TSubscriptionCallback<TStatus>
+  ) => {
+    return subscribeEndpoint(`devices/bitbox02-bootloader/${deviceID}/status`, cb);
+  };
```
benma, line 39: the get/sync pair used in `useSync()` is most often at the same endpoint, isn't it? sounds like we need another hook that only takes the endpoint and the object type and automatically calls `apiGet()` and `subscribeEndpoint()` :grin:

thisconnect: yeah it must subscribe at the same endpoint otherwise this wont reload 

https://github.com/digitalbitbox/bitbox-wallet-app/blob/56d4eab5dcd46fec73ec2f3675a7bb00c945b698/frontends/web/src/api/subscribe.ts#L40

there is only a few using action.reload, maybe we can migrate those to action.replace and get rid of reload all together in the future?

Screen Shot 2023-12-29 at 13 38 31

thisconnect (approved):
untested LGTM 👍

frontends/web/src/api/bitbox02bootloader.ts
benma: In theory that should be possible, but not sure if worth it? The only reason to do it would be that it would be faster, right? But I doubt it would be noticeable.

PR #2470, opened 2023-12-29
frontend: convert BB02 wizard.tsx to a functional component

benma:
The waitDialog state was unused so I removed it.

Instead of resetting `status` to `''`, instead I reset `showWizard`, which should work the same way to hide the dialog. The reason is that there is no easy way to set the `status` state manually, and it also seems cleaner.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/wizard.tsx
+++ b/frontends/web/src/routes/device/bitbox02/wizard.tsx
@@ -14,211 +14,155 @@
  * limitations under the License.
  */
 
-import { Component } from 'react';
-import { getStatus, getVersion, VersionInfo, verifyAttestation, TStatus } from '../../../api/bitbox02';
+import { useEffect, useState } from 'react';
+import { useNavigate } from 'react-router';
+import { useLoad, useSync } from '../../../hooks/api';
+import { getStatus, getVersion, verifyAttestation } from '../../../api/bitbox02';
 import { attestationCheckDone, statusChanged } from '../../../api/devicessync';
-import { UnsubscribeList, unsubscribe } from '../../../utils/subscriptions';
-import { route } from '../../../utils/route';
 import { AppUpgradeRequired } from '../../../components/appupgraderequired';
 import { FirmwareUpgradeRequired } from '../upgrade-firmware-required';
 import { Main } from '../../../components/layout';
 import { Unlock } from './unlock';
 import { Pairing } from './setup/pairing';
-import { Wait } from './setup/wait';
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/device/bitbox02/wizard.tsx
```diff
@@ -14,211 +14,154 @@
...
-    this.setState({ status: '' });
+export const Wizard = ({ deviceID }: TProps) => {
+  const versionInfo = useLoad(() => getVersion(deviceID));
+  const attestation = useSync(
+    () => verifyAttestation(deviceID),
+    cb => attestationCheckDone(deviceID, () => {
+      verifyAttestation(deviceID).then(cb);
+    })
+  );
+  const [appStatus, setAppStatus] = useState<'' | TWalletSetupChoices>('');
```
thisconnect, line 44: Since you are here, wdyt about renaming appStatus to something else, i.e. setupChoice or similar?

thisconnect (approved):
very nice, untested LGTM 👍 

@strmci if you have time could you do the routing tests with this branch?

frontends/web/src/routes/device/bitbox02/wizard.tsx
```diff
@@ -14,211 +14,154 @@
...
  */
 
-import { Component } from 'react';
-import { getStatus, getVersion, VersionInfo, verifyAttestation, TStatus } from '../../../api/bitbox02';
+import { useEffect, useState } from 'react';
+import { useLoad, useSync } from '../../../hooks/api';
+import { getStatus, getVersion, verifyAttestation } from '../../../api/bitbox02';
 import { attestationCheckDone, statusChanged } from '../../../api/devicessync';
-import { UnsubscribeList, unsubscribe } from '../../../utils/subscriptions';
 import { route } from '../../../utils/route';
```
thisconnect, line 21: with this change you could use useNavigation hook instead.

benma: Done

benma:
> very nice, untested LGTM 👍

@thisconnect I'd prefer it if you did a small bit of testing :smile:

frontends/web/src/routes/device/bitbox02/wizard.tsx
benma: `appStatus` is named on contrast to `status` which is the device status.

no strong opinion, but i don't feel like spending time on it :see_no_evil:

thisconnect:
tested LGTM 👍

PR #2474, opened 2024-01-01
frontend: deactivate hide amounts if disabled from settings page

shonsirsha:
previously, we deactivate hide amounts only via the "hide/show amounts" button (shown on the top of some pages).

Due to this, amounts will still be hidden even if we disable the settings as a whole - given we didn't disable the feature via the "hide/show amounts" btn.

This commit fixes it by always setting `setHideAmounts(false)`
as well as the config value `hideAmounts: false` if we disable
the feature via settings.

### DIFF
--- a/frontends/web/src/contexts/AppContext.tsx
+++ b/frontends/web/src/contexts/AppContext.tsx
@@ -28,6 +28,7 @@ type AppContextProps = {
     setGuideExists: Dispatch<SetStateAction<boolean>>;
     setGuideShown: Dispatch<SetStateAction<boolean>>;
     setSidebarStatus: Dispatch<SetStateAction<TSidebarStatus>>;
+    setHideAmounts: Dispatch<SetStateAction<boolean>>;
     toggleGuide: () => void;
     toggleHideAmounts: () => void;
     toggleSidebar: () => void;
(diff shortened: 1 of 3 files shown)


thisconnect:
This works for the currently running app, but doesn't update the config.json (I think), the config.json still has "hideAmounts" true. So when opening the app again it will hide the amounts.

```
    "frontend": {
        "allowHideAmounts": false,
        "guideShown": false,
        "hideAmounts": true
    }
```

tested this branch f1700eb57e584eec6a41775cf5558d3209d8b3de

1) hide amounts
2) disable settings
3) restart app

amounts are still hidden.

shonsirsha:
> This works for the currently running app, but doesn't update the config.json, the config.json still has "hideAmounts" true. So when opening the app again it will hide the amounts.
> 
> 
> 
> ```
> 
>     "frontend": {
> 
>         "allowHideAmounts": false,
> 
>         "guideShown": false,
> 
>         "hideAmounts": true
> 
>     }
> 
> ```
> 
> 
> 
> tested this branch f1700eb57e584eec6a41775cf5558d3209d8b3de
> 
> 
> 
> 1) hide amounts
> 
> 2) disable settings
> 
> 3) restart app
> 
> 
> 
> amounts are still hidden.
> 
> 
> 
> 

Oh good find.. we need to also set the config to false. Thx!

shonsirsha:
PTAL @thisconnect  🙏

thisconnect:
please rebase

thisconnect (approved):
thanks 👍  tested LGTM

PR #2483, opened 2024-01-08
frontend: type 2 subscription fns signProgresss and signConfirm 

shonsirsha:
type 2 subscription fns `signProgresss` and `signConfirm`

On top of that did a small refactor creating a reusable `updateBalance` fn.

### DIFF
--- a/frontends/web/src/api/devicessync.ts
+++ b/frontends/web/src/api/devicessync.ts
@@ -77,3 +77,30 @@ export const attestationCheckDone = (
   });
   return unsubscribe;
 };
+
+export type TSignProgress = {
+  steps: number;
+  step: number;
+}
+
+export const signProgress = (
+  cb: (progress: TSignProgress) => void
+): TUnsubscribe => {
+  const unsubscribe = subscribeLegacy('signProgress', event => {
+    if ('type' in event && event.type === 'device' && event.data === 'signProgress') {
+      cb(event.meta);
+    }
+  });
+  return unsubscribe;
+};
+
+export const signConfirm = (
+  cb: () => void
+): TUnsubscribe => {
+  const unsubscribe = subscribeLegacy('signConfirm', event => {
+    if ('type' in event && event.type === 'device' && event.data === 'signConfirm') {
+      cb();
+    }
+  });
+  return unsubscribe;
+};
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -30,20 +30,20 @@ import { Column, ColumnButtons, Grid, GuideWrapper, GuidedContent, Header, Main 
 import { Status } from '../../../components/status/status';
 import { translate, TranslateProps } from '../../../decorators/translate';
 import { apiGet, apiPost } from '../../../utils/request';
-import { apiWebsocket } from '../../../utils/websocket';
-import { isBitcoinBased, findAccount } from '../utils';
 import { FeeTargets } from './feetargets';
-import { TSelectedUTXOs, UTXOs } from './utxos';
 import { route } from '../../../utils/route';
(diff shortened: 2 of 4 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -158,28 +158,10 @@ class Send extends Component<Props, State> {
...
-              this.setState({ signProgress: meta, signConfirm: false });
-              break;
-            case 'signConfirm':
-              this.setState({ signConfirm: true });
-              break;
-            }
-            break;
-          }
-        }
-      }),
```
thisconnect, line 177: Can you type these 2 subscriptions in api/devicessync?

It should be similar to channelHashChanged in api/devicessync.ts

thisconnect (changes requested):

shonsirsha:
Made the change, PTAL :) updaated commit msg and rebased as well.

frontends/web/src/api/devicessync.ts
```diff
@@ -77,3 +79,24 @@ export const attestationCheckDone = (
   });
   return unsubscribe;
 };
+
+
+export const signProgress = (
+  cb: (meta: TSignProgress) => void
+): TUnsubscribe => {
+  return apiWebsocket(payload => {
```
thisconnect, line 87: just use subscribeLegacy like the other functions in this module do.

have a look at frontends/web/src/utils/event-legacy.ts there it uses apiWebsocket internally

frontends/web/src/api/devicessync.ts
```diff
@@ -77,3 +79,24 @@ export const attestationCheckDone = (
...
+    if ('type' in payload && payload.type === 'device' && payload.data === 'signProgress') {
+      cb(payload.meta);
+    }
+  });
+};
+
+export const signConfirm = (
+  cb: (signConfirm: boolean) => void
+): TUnsubscribe => {
+  return apiWebsocket(payload => {
```
thisconnect, line 97: same, change to subscribeLegacy

frontends/web/src/api/devicessync.ts
```diff
@@ -77,3 +79,24 @@ export const attestationCheckDone = (
...
+    if ('type' in payload && payload.type === 'device' && payload.data === 'signProgress') {
+      cb(payload.meta);
+    }
+  });
+};
+
+export const signConfirm = (
+  cb: (signConfirm: boolean) => void
```
thisconnect, line 95: I think the backend doesn't return anything here, but maybe check if there is something in the event object, or check this handler in the backend go implementation.

so I thin you can just use `() => void` to use this as a callback.

and then change

```js
  signConfirm(() =>
    this.setState({ signConfirm: true })
  ),
```

frontends/web/src/routes/account/send/services.ts
```diff
@@ -18,6 +18,11 @@
 import { alertUser } from '../../../components/alert/Alert';
 import { i18n } from '../../../i18n/i18n';
 
+export type TSignProgress = {
+  steps: number;
+  step: number;
+}
```
thisconnect, line 24: this type is coming from the backend api and should be in api/devicessync.ts

thisconnect (changes requested):

frontends/web/src/routes/account/send/services.ts
shonsirsha: Makes sense 👍

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -158,28 +158,15 @@ class Send extends Component<Props, State> {
...
+      signProgress((meta) =>
```
thisconnect, line 161: nit: meta is event internal.. Maybe name it progress or similar?

```suggestion
      signProgress((progress) =>
```

thisconnect (approved):
untested LGTM with a small nit

PR #2484, opened 2024-01-09
fix: buy page title shows "Buy Bitcoin" if only has btc account

shonsirsha:
also fixed the active state / class  of the sidebar when user is in specific account's buy page (`/buy/exchange/accountCode`).

Preview:

Shows "Buy Bitcoin" as title when only BTC accounts are loaded:

btcttl

:Shows Sidebar active when on specific account's buy page:

sdb

### DIFF
--- a/frontends/web/src/components/sidebar/sidebar.tsx
+++ b/frontends/web/src/components/sidebar/sidebar.tsx
@@ -75,6 +75,7 @@ const Sidebar = ({
   accounts,
 }: SidebarProps) => {
   const { t } = useTranslation();
+  const { pathname } = useLocation();
   const { activeSidebar, sidebarStatus, toggleSidebar } = useContext(AppContext);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -209,7 +211,7 @@ const Sidebar = ({
         { accounts.length ? (
           <div key="buy" className={style.sidebarItem}>
             <NavLink
-              className={({ isActive }) => isActive ? style.sidebarActive : ''}
+              className={({ isActive }) => isActive || userInSpecificAccountBuyPage ? style.sidebarActive : ''}
```
benma, line 214: how does this `isActive` work, where is it coming from?

shonsirsha: from [React Router's NavLink](https://reactrouter.com/en/main/components/nav-link). 

`isActive` is true if current route matches the `to` prop of `NavLink`

shonsirsha, line 214: At this point we don't actually need the `isActive` here, as `userInSpecificAccountBuyPage` checks for all urls prefixed by `/buy`. 

Keeping for consistency,  but lmk if we wanna remove it.

Beerosagos: Not sure which is best, actually. It's ok to leave it, I guess.

Beerosagos (approved):
tACK! Thanks

PR #2485, opened 2024-01-09
Handle keystores in Bitsurance workflow

Beerosagos:
Refactor message signing code and add keystore handling on the Bitsurance workflow


backend/backend.go
```diff
@@ -930,3 +933,75 @@ func (backend *Backend) SetWatchonly(rootFingerprint []byte, watchonly bool) err
...
+//	`format` is the script type that should be used in the address derivation, as received by the widget
+//		(see https://github.com/pocketbitcoin/request-address#requestaddressv0messagescripttype).
+//		If format is empty, native segwit type is used as a fallback.
+//
+// Returned values:
+//
+//	#1: is the first unused address corresponding to the account and the script type identified by the input values.
+//	#2: base64 encoding of the message signature, obtained using the private key linked to the address.
+//	#3: is an optional error that could be generated during the execution of the function.
+func (backend *Backend) SignBTCAddress(account accounts.Interface, message string, format string) (string, string, error) {
```
benma, line 964: `backend` is not used at all, so this does not need to be a `Backend` method.

frontends/web/src/api/backend.ts
```diff
@@ -32,6 +32,8 @@ export interface ISuccess {
     errorCode?: string;
 }
 
+export type BackendErrorCode = 'userAbort' | 'addressNotFound';
```
benma: i wonder if it's better to have one enum of all backend error codes or scope them to the specific handlers. i think i prefer the latter a bit, as then it's very clear which function needs to handle which errors.

frontends/web/src/api/coins.ts
```diff
@@ -46,3 +46,16 @@ export type TAmount = {
...
+export type AddressSignResponse = {
+  success: boolean;
+  signature: string;
+  address: string;
+  errorMessage?: string;
+  errorCode?: BackendErrorCode;
+}
+
+export const signAddress = (format: string, msg: string, accountCode: AccountCode): Promise<AddressSignResponse> => {
+  return apiPost('coins/btc/sign-address', { format, msg, accountCode });
```
benma, line 59: i wonder why the sign function and endpoints are not added to btc/account.go and btc/handlers.go, since it's all BTC specific.

benma (changes requested):

frontends/web/src/api/coins.ts
```diff
@@ -46,3 +46,16 @@ export type TAmount = {
...
 };
+
+export type AddressSignResponse = {
+  success: boolean;
+  signature: string;
+  address: string;
+  errorMessage?: string;
+  errorCode?: BackendErrorCode;
+}
+
```
thisconnect, line 57: Please change to

```ts
export type AddressSignResponse = {
  success: true;
  signature: string;
  address: string;
} | {
  success: false;
  errorMessage?: string;
  errorCode: BackendErrorCode;
}
```

this is just an example, please adjust the optional fields as the backend returns those. Reason this way we can tell TypeScript that there is always a errorCode if success is false. Or that there is a signature if success true. etc.

backend/backend.go
Beerosagos: Uh, right! Changed and moved to btc/account.go as suggested in the other comment

frontends/web/src/api/coins.ts
Beerosagos: Good point, done.

frontends/web/src/api/coins.ts
Beerosagos: Done, thanks!

Beerosagos:
Pushed a new update, PTAL :pray:

backend/coins/btc/account.go
```diff
@@ -829,3 +831,75 @@ func (account *Account) VerifyExtendedPublicKey(signingConfigIndex int) (bool, e
...
+		format = defaultScriptType
+	}
+	expectedScriptType, ok := requestAddressScriptTypeMap[format]
+	if !ok {
+		err := fmt.Errorf("Unknown format:  %s", format)
+		return "", "", err
+	}
+	signingConfigIdx := account.Config().Config.SigningConfigurations.FindScriptType(expectedScriptType)
+	if signingConfigIdx == -1 {
+		err := fmt.Errorf("Unknown format: %s", format)
```
benma, line 890: use `errp.Newf()` instead, which adds a stack trace.

PR #2486, opened 2024-01-10
backend/account: hide non-segwit xpubs for insured accounts

Beerosagos:
Bitsurance contract only covers funds on native segwit addresses. We already hide non-segwit addresses from the receive screen, and now we also hide non-segwit xpubs from the account info page.


benma (approved):
utACK good catch!

PR #2487, opened 2024-01-11
frontend/keystore: create keystore context to handle keystore requests

Beerosagos:
With the current implementation, when the wrong keystore is connected after a keystore request, the app is not able to automatically detect the connection of the correct keystore. The only option for the user is to cancel the 'connect keystore' popup and repeat the action to create a new keystore connection request.  This is counterintuitive and inefficient. 

This commit introduces a frontend keystore context that allows 
- to keep track of keystore connection requests in a centralized way
- to easily create keystore connection requests in the frontend, providing a callback function to be called in case of success
- to automatically detect and handle the connection of the correct keystore after a wrong one

As a first use example, the keystore context is used in the buy/info.tsx page. All the other occurrences of `connectKeystore` will be replaced with the new method in the next commits.


frontends/web/src/routes/buy/info.tsx
```diff
@@ -76,7 +80,6 @@ export const BuyInfo = ({ code, accounts }: TProps) => {
     maybeProceed();
   }, [maybeProceed, options]);
```
thisconnect, line 81: I find this `maybeProceed` a bit confusing, in the other places I assume we don't need a redirectAttempted.current ref because it's only called once in onSuccess?

thisconnect (commented):
I haven't tested yet, but added some frontend review comments. Looks very nice.

frontends/web/src/api/account.ts
```diff
@@ -342,7 +342,14 @@ export const testRegister = (pin: string): Promise<null> => {
   return apiPost('test/register', { pin });
 };
 
-export const connectKeystore = (code: AccountCode): Promise<{ success: boolean; }> => {
+export type TResultConnectKeystore = {
+  success: boolean;
+  errorCode?: 'wrongKeystore';
+};
```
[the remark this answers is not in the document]
Beerosagos: Good point. I improved the typing, but I could not make the errorCode not optional, as it may happen that the backend only returns success=false without any error code. E.g. when a user cancels the current keystore request

thisconnect:
@Beerosagos would it also make sense to use requestKeystore for when WalletConnect needs it?

frontends/web/src/api/account.ts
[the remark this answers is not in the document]
thisconnect: We don't need it now, but maybe the backend should return 'canceled' or similar?

Beerosagos:
> @Beerosagos would it also make sense to use requestKeystore for when WalletConnect needs it?

I think so. I had a look but it seems a bit more complicated than in other places and I think we'll need to add an `onCancel` handler to the requestKeystore, to send the transaction reject in case the user aborts on the 'connect keystore' popup. I'd leave that for a separate PR

Beerosagos:
> I think so. I had a look but it seems a bit more complicated than in other places and I think we'll need to add an `onCancel` handler to the requestKeystore, to send the transaction reject in case the user aborts on the 'connect keystore' popup. I'd leave that for a separate PR

@thisconnect I drafted another PR based on this one, if you want to have a look in advance: https://github.com/digitalbitbox/bitbox-wallet-app/pull/2489/commits/92d87d82352a255a3edfa356f3184cf02cf9c16e

frontends/web/src/api/keystores.ts
```diff
@@ -31,3 +31,34 @@ export const subscribeKeystores = (
 export const getKeystores = (): Promise<TKeystores> => {
   return apiGet('keystores');
 };
+
+export type TSyncConnectKeystore = null | {
+  typ: 'connect';
```
thisconnect, line 36: isn't it `type` instead of `typ` ?

thisconnect:
> I had a look but it seems a bit more complicated than in other places and I think we'll need to add an onCancel handler to the requestKeystore, to send the transaction reject in case the user aborts on the 'connect keystore' popup. I'd leave that for a separate PR

so we need canceled event? see my comment here: https://github.com/digitalbitbox/bitbox-wallet-app/pull/2487#discussion_r1452581257

frontends/web/src/api/keystores.ts
Beerosagos: Nope, I think this is done on purpose to avoid collisions with protected keywords. It is defined like this in the backend https://github.com/digitalbitbox/bitbox-wallet-app/blob/master/backend/accounts.go#L579-L584

frontends/web/src/api/account.ts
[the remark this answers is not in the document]
Beerosagos: I added the 'userAbort' errorCode in the next PR I drafted (see my other comment https://github.com/digitalbitbox/bitbox-wallet-app/pull/2487#issuecomment-1893403206). Anyway, the backend function can still returns unexpected errors without a specific error code. The only way to always have a defined error code is to wrap all the unexpected errors around a 'genericError' error code, but I don't know if that would be actually useful.

Beerosagos:
> > I had a look but it seems a bit more complicated than in other places and I think we'll need to add an onCancel handler to the requestKeystore, to send the transaction reject in case the user aborts on the 'connect keystore' popup. I'd leave that for a separate PR
> 
> so we need canceled event? see my comment here: [#2487 (comment)](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2487#discussion_r1452581257)

No, we already have it, we just need to correctly handle the cancellation with another optional callback in the context provider.

thisconnect:
@shonsirsha has more experience with context if you have time could check this PR?

frontends/web/src/contexts/KeystoreProvider.tsx
```diff
@@ -0,0 +1,92 @@
...
+    // keystore event, and a new call to `connect` when there is a change in the keystores.
+    // If a `cancelConnectKeystore` is invoked, `connectKeystore` will return with success = false
+    // and without errorCode.
+    const connectResult = await connectKeystore(accountCode);
+    if (!connectResult.success) {
+      if (connectResult?.errorCode === 'wrongKeystore') {
+        if (keystoreWait?.accountCode !== accountCode) {
+          setKeystoreWait({ accountCode, onSuccess });
+        }
+      }
```
shonsirsha, line 60: nit: instead of nesting, would it be possible to have them as just 1 if stmt? 🤔 

```js
const hasWrongKeystore = a && b && c;
if (hasWrongKeystore)
```

frontends/web/src/routes/buy/info.tsx
shonsirsha: I'm not certain if it helps, but have you tried wrapping `requestKeystore` to be inside useCallback?

shonsirsha (commented):
Context API-wise looks totally fine to me :) just left a small nit.

thisconnect (approved):
tested Receive workflow and buy account selector in webdev, with 3 wallets on 2 different devices (one with 2 passphrases) Works as expected. 

tested  LGTM 👍 

I see @shonsirsha has a nit and a recommendation. I think the nit would be nice as it improves readability. The refactoring recommendation is probably better to do in a separate PR later.

PR #2489, opened 2024-01-16
 frontend/walletconnect: adopt requestKeystore in walletconnect 

Beerosagos:


frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -101,43 +103,47 @@ export const Pocket = ({ code }: TProps) => {
...
-          sendAddress(response.address, response.signature);
-        } else {
-          if (response.errorCode !== 'userAbort') {
-            alertUser(t('unknownError', { errorMessage: response.errorMessage }));
-            console.log('error: ' + response.errorMessage);
+    requestKeystore(code, async () => {
+      signing = true;
+      const addressType = message.withScriptType ? String(message.withScriptType) : '';
+      const withMessageSignature = message.withMessageSignature ? message.withMessageSignature : '';
+      signAddress(
```
benma, line 110: the callback is marked `async`, so you should use `await` here instead of `...then()`. this or remove the `async` keysword as it is only needed if you want to make use of `await`

frontends/web/src/routes/buy/pocket.tsx
```diff
@@ -101,43 +103,47 @@ export const Pocket = ({ code }: TProps) => {
...
   };
 
   const handleVerifyAddress = (address: string) => {
-    setVerifying(true);
-    verifyAddress(address, code)
-      .then(response => {
-        setVerifying(false);
-        if (!response.success) {
-          if (response.errorCode === 'addressNotFound') {
+    requestKeystore(code, async () => {
```
benma, line 130: same

frontends/web/src/contexts/KeystoreContext.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createContext } from 'react';
+import { AccountCode } from '../api/account';
+
+type KeystoreContextProps = {
+    requestKeystore: (accountCode: AccountCode, onSuccess?: () => void, onCancel?: () => void) => Promise<void>;
```
benma, line 21: since now the backend does half the work and the frontend does half the work in managing this, it could be useful to add some docstrings in these two functions about how to use them and what they do exactly in each case.

also useful to document what happens if one does *not* call this function - the backend calls `keystore, err := account.Config().ConnectKeystore()` anyway which will generally work, but without the benefits introduced in this PR.

benma (changes requested):

PR #2490, opened 2024-01-16
Rotated Bitsurance API key

rootzoll:
In preperation for the GoLive I did a last API key rotation. The new key also works with test & production server and should now stay constant after release. Its a public published key anyway, but now its prepared for GoLive. 

Not part of this PR but as a reminder ... the `backend.DevServers()` can now give back `false` to signal to work against the production API for next release.

### DIFF
--- a/backend/bitsurance/bitsurance.go
+++ b/backend/bitsurance/bitsurance.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	apiKey        = "265bfd773038c9ad9da7047e107babba0269bc3d31952172d9b10335e8a9d8e9"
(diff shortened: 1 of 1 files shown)


Beerosagos:
Thanks @rootzoll ! Can we just keep the last commit ([d3c4f32](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2490/commits/d3c4f3270761243bf44b76d0338a390f61439a11)), or there is a reason to keep also the others? 

About the `backend.DevServers()` it returns false if the app is compiled for production, true otherwise, so we should not need any change there.

rootzoll:
@Beerosagos I cleaned up the commits ... let me know if you need anything else.

PR #2491, opened 2024-01-16
frontend: refactor get keystore by accounts

thisconnect:
- removed sort from getAccountsByKeystore (this was not really necessary anymore after share decorator was removed from sidebar in a8be91a1d3d007cbb17ab2ca29e554e2335532d9)
- moved the sort to the parent component to after the accounts subscription (so we don't sort on every render)
- added unit test for getAccountsByKeystore
- added separate sortAccounts function + tests
- added missing 'lastConnected' key in TKeystore

### DIFF
--- a/frontends/web/src/routes/account/utils.test.ts
+++ b/frontends/web/src/routes/account/utils.test.ts
@@ -1,0 +1,131 @@
+/**
+ * Copyright 2023 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { CoinCode, IAccount } from '../../api/account';
+import { getAccountsByKeystore } from './utils';
+
+describe('utils/getAccountsByKeystore', () => {
+
+  it('should run and return an empty array', () => {
+    const result = getAccountsByKeystore([]);
+    expect(result).toBeTypeOf('object');
+    expect(result.length).toBe(0);
+  });
+
+  it('should return a new empty array', () => {
+    const accounts: IAccount[] = [];
+    const result = getAccountsByKeystore(accounts);
+    expect(accounts).not.toBe(result);
+  });
+
+  it('should return one keystore entry with 2 accounts', () => {
+    const accounts = [
+      {
+        active: true,
+        blockExplorerTxPrefix: 'https://blockstream.info/testnet/tx/',
+        code: 'v0-123de678-tbtc-0',
+        coinCode: 'tbtc' as CoinCode,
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/account/utils.ts
```diff
@@ -100,6 +100,18 @@ export type TAccountsByKeystore = {
...
 };
 
+export const sortAccounts = (accounts: IAccount[]) => {
+  return Array.from(accounts).sort(
+    (ac1, ac2) => ac1.keystore.name.localeCompare(ac2.keystore.name)
+  );
+  // TODO: thinking we should use keystore.fingerprint account.code
+  /*
+  return Array.from(accounts).sort((ac1, ac2) => {
+    return `${ac1.keystore.fingerprint}${ac1.code}`.localeCompare(`${ac2.keystore.fingerprint}${ac2.code}`)
```
thisconnect, line 110: I haven't tested it yet, but I think this should fix it.

WDYT about combining 
keystore.fingerprint + account.code ?

thisconnect: actually the account.code already inclues the fingerprint, so it's not strictly needed.. should I keep as this is the intention?

thisconnect:
Initially I wanted to refactor the actual function, but this version is better :)

I am not sure how to squash the commits.. I'll keep those as they are for now for review.

frontends/web/src/routes/account/utils.ts
```diff
@@ -100,6 +100,12 @@ export type TAccountsByKeystore = {
   accounts: IAccount[];
 };
 
+export const sortAccounts = (accounts: IAccount[]): IAccount[] => {
+  return Array.from(accounts).sort((ac1, ac2) => {
+    return `${ac1.keystore.rootFingerprint}${ac1.code}`.localeCompare(`${ac2.keystore.rootFingerprint}${ac2.code}`);
```
thisconnect, line 105: @benma should we rely on that Account.code will always include rootFingerprint in the future?

In that case this could be simplified by just: `ac1.code.localeCompare(ac2.code})`

frontends/web/src/routes/account/utils.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    expect(result[0].keystore.rootFingerprint).toBe('w1');
+    expect(result[1].keystore.rootFingerprint).toBe('w1');
+    expect(result[2].keystore.rootFingerprint).toBe('w1');
+    expect(result[3].keystore.rootFingerprint).toBe('w2');
+    expect(result[4].keystore.rootFingerprint).toBe('w2');
+    expect(result[5].keystore.rootFingerprint).toBe('w2');
+    expect(result[6].keystore.rootFingerprint).toBe('w2');
+  });
+
+  it('should return a new sorted array, this time all keystores have the same name', () => {
```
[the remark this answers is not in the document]
Beerosagos: Yeah, sorting by account name looks like a better option!

frontends/web/src/routes/account/utils.ts
Beerosagos: I think we can avoid adding the fingerprint explicitly, but no strong opinion

Beerosagos (approved):
tACK

frontends/web/src/routes/account/utils.ts
thisconnect: I'll keep as is, we should assume it will change.

thisconnect:
I squashed the failing test + the next commit that solved it.

Beerosagos:
> I squashed the failing test + the next commit that solved it.

LGTM :+1:

PR #2493, opened 2024-01-17
backend/keystore: detect new keystores after a wrong one

Beerosagos:
With the current implementation, when the wrong keystore is connected after a keystore request, the app is not able to automatically detect the connection of the correct keystore. The only option for the user is to cancel the 'connect keystore' popup and repeat the action to create a new keystore connection request. This is counterintuitive and inefficient.

This adds a `connectkeystore.retryCallback` that allows waiting for a new keystore to be connected and checking if that is the expected one.

This PR is a better solution than #2487, as it solves the same issue while being easier and completely transparent to the frontend. It also makes #2489 unnecessary.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -660,7 +660,7 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
 			type data struct {
 				Type         string `json:"typ"`
 				KeystoreName string `json:"keystoreName"`
-				ErrorCode    string `json:"errorCode"`
+				ErrorCode    string `json:"errorCode,omitempty"`
 				ErrorMessage string `json:"errorMessage"`
 			}
 			accountRootFingerprint, err := persistedConfig.SigningConfigurations.RootFingerprint()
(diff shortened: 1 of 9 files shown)


backend/accounts.go
```diff
@@ -591,25 +591,54 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
...
+							ErrorCode:    "wrongKeystore",
+							ErrorMessage: err.Error(),
+						},
+					})
+					c := make(chan bool)
+					// retryCallback is called when the current keystore is deregistered or when
+					// CancelConnectKeystore() is called.
+					// In the first case it allows to make a new connection attempt, in the last one
+					// it'll make this function return ErrUserAbort.
+					backend.connectKeystore.retryCallback = func(retry bool) {
```
benma, line 626: should add a method `setRetryCallback()` which acquires the lock and call that here, otherwise there could be a race condition.

backend/accounts.go
```diff
@@ -591,25 +591,54 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
...
+						c <- retry
+					}
+					retry := <-c
```
benma, line 629: Could be good to `select { case retry:=<- r: ...; case time.After(time.Minute): ... }` so that this surely finishes at some point.

benma (commented):
still need to test it, just some initial comments.

benma (changes requested):
Bugs I had during testing:

1.

I went to verify a receive address.
Unlock with wrong keystore, the correct error was shown.

Then click 'cancel' and click again to verify: it starts with the same error state 'Wrong  wallet connected', but I expect that it would show the normal 'Please connect ...' prompt.

2. already exists on master

If you click to verify address, and then exit via `Esc`, there are two bugs:

2a) we exit the whole receive screen instead of just the keystore prompt dialog
2b) when clicking again to verify address, nothing happens - probably cause the previous connect keystore thing is still running somehow.

i guess the keystore connect dialog needs to call cancel on dialog close/exit.

frontends/web/src/components/keystoreconnectprompt.tsx
```diff
@@ -51,7 +52,7 @@ export function KeystoreConnectPrompt() {
           <SkipForTesting />
```
benma, line 52: would be good to add that also to the `wrongKeystore` error state, so one could try again unlocking also in devmode.

benma (approved):
tACK - bug 1 was not a bug, i forgot to detach the test keystore.

bug 2 can be handled separately

backend/accounts.go
Beerosagos: Good catch, thanks! Done

backend/accounts.go
Beerosagos: Done. I also improved a bit the general handling of timeout events.

frontends/web/src/components/keystoreconnectprompt.tsx
Beerosagos: Done

Beerosagos:
Thanks for the review!  I should have addressed all the points.

About bug2: it is also present on master, I'll make a separate PR for that

backend/accounts.go
Beerosagos: Thinking about it again, since that code is only executed when a `connect` call returns `wrongKeystore`, and the `connect` call returns something different when a new `connect` is executed, that should be only executed by a single thread at a time. Anyway, using a proper lock is definitely better and safer.

benma (approved):

PR #2497, opened 2024-01-22
frontend: created and used GroupedAccountSelector

shonsirsha:
to improve our UX when multiple keystore is connected, we'd like to add a grouping on the account selector.

This account selector is used on the info buy page (`/buy/info`) as well as the receive page (`/accounts/select-receive`) and bitsurance.

Screenshot 2024-01-22 at 11 40 10

;l

d2

d3

### DIFF
--- a/frontends/web/src/components/groupedaccountselector/groupedaccountselector.tsx
+++ b/frontends/web/src/components/groupedaccountselector/groupedaccountselector.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 Shift Crypto AG
+ * Copyright 2023-2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,15 +13,25 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { FunctionComponent, useContext, useEffect, useState } from 'react';
+import { useContext, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import Select, { components, SingleValueProps, OptionProps, SingleValue, DropdownIndicatorProps } from 'react-select';
-import { AccountCode, IAccount, getBalance } from '../../api/account';
+import Select, { components, SingleValueProps, OptionProps, SingleValue, DropdownIndicatorProps, GroupProps, GroupHeadingProps as ReactSelectGroupHeadingProps } from 'react-select';
+import { AccountCode, IAccount } from '../../api/account';
 import { Button } from '../forms';
 import Logo from '../icon/logo';
 import { AppContext } from '../../contexts/AppContext';
-import styles from './accountselector.module.css';
+import { USBSuccess } from '../icon';
+import { Badge } from '../badge/badge';
 import { InsuredShield } from '../../routes/account/components/insuredtag';
+import { getAccountsByKeystore } from '../../routes/account/utils';
+import { createGroupedOptions, getBalancesForGroupedAccountSelector } from './services';
+import styles from './groupedaccountselector.module.css';
+
+export type TGroupedOption = {
+  label: string;
+  connected: boolean;
+  options: TOption[];
+}
 
 export type TOption = {
   label: string;
@@ -35,20 +45,13 @@ export type TOption = {
 type TAccountSelector = {
   title: string;
   disabled?: boolean;
-  options: TOption[];
   selected?: string;
   onChange: (value: string) => void;
   onProceed: () => void;
+  accounts: IAccount[]
 }
 
-export const setOptionBalances = async (options: TOption[]): Promise<TOption[]> => {
-  return await Promise.all(options.map(async (option) => {
-    const balance = await getBalance(option.value);
-    return { ...option, balance: `${balance.available.amount} ${balance.available.unit}` };
-  }));
-};
-
-const SelectSingleValue: FunctionComponent<SingleValueProps<TOption>> = (props) => {
+const SelectSingleValue = (props: SingleValueProps<TOption>) => {
   const { hideAmounts } = useContext(AppContext);
   const { label, coinCode, balance, insured } = props.data;
   return (
@@ -65,7 +68,7 @@ const SelectSingleValue: FunctionComponent<SingleValueProps<TOption>> = (props) 
   );
 };
 
-const SelectOption: FunctionComponent<OptionProps<TOption>> = (props) => {
+const SelectOption = (props: OptionProps<TOption>) => {
   const { hideAmounts } = useContext(AppContext);
   const { label, coinCode, balance, insured } = props.data;
 
@@ -81,7 +84,7 @@ const SelectOption: FunctionComponent<OptionProps<TOption>> = (props) => {
   );
 };
 
-const DropdownIndicator: FunctionComponent<DropdownIndicatorProps<TOption>> = (props) => {
+const DropdownIndicator = (props: DropdownIndicatorProps<TOption>) => {
   return (
     <components.DropdownIndicator {...props}>
       <div className={styles.dropdown} />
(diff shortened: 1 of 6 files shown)


frontends/web/src/routes/buy/info.tsx
```diff
@@ -34,45 +35,50 @@ type TProps = {
...
     }
 
   }, [accounts]);
 
+  useEffect(() => {
+    checkSupportedCoins();
+  }, [checkSupportedCoins]);
+
+
   const maybeProceed = useCallback(async () => {
```
shonsirsha, line 68: I wanted to refactor the GroupedAccountSelector to just accept an IAccount object and do all the object / group constructing in that component. 

However, we need the group options outside of that component sometimes like here.

I think it still can be refactored though, I could work on this next. My idea is so that we can just pass an `IAccount[]` and "it just works" (automatically grouped by keystore like we have) without having to do all the grouping in the parent component like currently.

frontends/web/src/routes/accounts/select-receive.tsx
```diff
@@ -16,27 +16,33 @@
...
   useEffect(() => {
-    const options = activeAccounts.map(account => ({ label: account.name, value: account.code, disabled: false, coinCode: account.coinCode } as TOption));
+    const accountNames = getAccountsByKeystore(activeAccounts);
     //setting options without balance
-    setOptions(options);
+    const groupedOpts: TGroupedOption[] = accountNames.map(grouped => ({
+      label: `${grouped.keystore.name} ${isAmbiguiousName(grouped.keystore.name, accountNames) ? `(${grouped.keystore.rootFingerprint})` : ''}`,
+      connected: grouped.keystore.connected,
+      options: grouped.accounts.map((account) => ({ label: account.name, value: account.code, coinCode: account.coinCode, disabled: false })) as TOption[]
+    }));
```
thisconnect, line 42: how about destructing accounts and keystores?

```ts
    const groupedOpts: TGroupedOption[] = accountNames.map(({ accounts, keystore }) => ({
      label: `${keystore.name} ${isAmbiguiousName(keystore.name, accountNames) ? `(${keystore.rootFingerprint})` : ''}`,
      connected: keystore.connected,
      options: accounts.map((account) => ({ label: account.name, value: account.code, coinCode: account.coinCode, disabled: false })) as TOption[]
    }));
```

frontends/web/src/routes/bitsurance/account.tsx
```diff
@@ -48,17 +50,22 @@ export const BitsuranceAccount = ({ code, accounts }: TProps) => {
...
-           || account.bitsuranceStatus === 'canceled'
-           || account.bitsuranceStatus === 'refused'))
-      .map(({ name, code, coinCode }) => (
-        { label: name, value: code, coinCode, disabled: false }
-      ));
-    setBtcAccounts(await setOptionBalances(options));
+
+    // btc accounts that have never been insured, or with a canceled
+    // insurance contract, can be used to make a new contract.
+    const btcAccountz = accounts.filter(account => account.coinCode === 'btc' &&
```
thisconnect, line 56: typo btcAccounts :)

frontends/web/src/routes/bitsurance/account.tsx
```diff
@@ -48,17 +50,22 @@ export const BitsuranceAccount = ({ code, accounts }: TProps) => {
...
+        (!account.bitsuranceStatus
+         || account.bitsuranceStatus === 'canceled'
+        || account.bitsuranceStatus === 'refused'));
+
+    const groupedBTCAccounts = getAccountsByKeystore(btcAccountz);
+    const groupedOpts: TGroupedOption[] = groupedBTCAccounts.map(grouped => ({
+      label: `${grouped.keystore.name} ${isAmbiguiousName(grouped.keystore.name, groupedBTCAccounts) ? `(${grouped.keystore.rootFingerprint})` : ''}`,
+      connected: grouped.keystore.connected,
+      options: grouped.accounts.map((account) => ({ label: account.name, value: account.code, coinCode: account.coinCode, disabled: false })) as TOption[]
+    }));
```
thisconnect, line 66: is this map callback function the same as in select-receive and buy-info, can you move it to a shared function?

frontends/web/src/routes/bitsurance/account.tsx
```diff
@@ -97,16 +104,17 @@ export const BitsuranceAccount = ({ code, accounts }: TProps) => {
...
                 <div>{t('bitsuranceAccount.noAccount')}</div>
               ) : (
-                <AccountSelector
+                <GroupedAccountSelector
                   title={t('bitsuranceAccount.select')}
-                  options={btcAccounts}
                   disabled={disabled}
+                  options={groupedOptions}
                   selected={selected}
                   onChange={handleChangeAccount}
```
thisconnect, line 115: here you didn't rewrite onChange, but in the 2 other similar components you have  `onChange={(selected) => setSelected(selected)}` and `onChange={setCode}`.

thisconnect (changes requested):

frontends/web/src/components/groupedaccountselector/groupedaccountselector.tsx
```diff
@@ -65,7 +66,7 @@ const SelectSingleValue: FunctionComponent<SingleValueProps<TOption>> = (props)
   );
 };
 
-const SelectOption: FunctionComponent<OptionProps<TOption>> = (props) => {
+const SelectOption = (props: OptionProps<TOption, false>) => {
```
thisconnect, line 69: why is this needed?

thisconnect:
When going to `/buy/info` In testnet (`make servewallet`) the keystore prompts directly. 

It works normally in  mainnet ( with `make servewallet-mainnet`).

it doesn't happen for me in master.

frontends/web/src/routes/bitsurance/account.tsx
shonsirsha: oups, good point. this one comes from the Bitsurance staging merge,  I'll just change this to setState. Same with the () => setSelected, will change to setSelected. Ty!

frontends/web/src/components/groupedaccountselector/groupedaccountselector.tsx
shonsirsha: I only wanted to remove the `FunctionComponent` as it's not needed (easier to read without it, right?). I'll omit the `false` as thats the default value anyway (for `isMulti` - for multi select in react-select).

frontends/web/src/routes/bitsurance/account.tsx
shonsirsha: Yeah, it's related to [this](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2497#discussion_r1461681393). I'll refactor.

frontends/web/src/routes/accounts/select-receive.tsx
shonsirsha: tbh I prefer it this way. To me it's clearer which `name` it is for example if we explicitly do `keystore.name` than just `name` knowing in this block we also have `account.name`. Also clearer for me personally to see that `rootFingerPrint` for ex, comes from keystore by doing`keystore.`. 

 But if you strongly prefer destructured, we could do that too ofc :)

shonsirsha:
> When going to `/buy/info` In testnet (`make servewallet`) the keystore prompts directly.
> 
> It works normally in mainnet ( with `make servewallet-mainnet`).
> 
> it doesn't happen for me in master.

I don't think I can reproduce this. @Beerosagos, would you be able to check?

shonsirsha:
@thisconnect thanks for the review. I refactored to achieve [this](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2497#discussion_r1461681393), which should account for your [comment](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2497#discussion_r1462232630). I think it looks a lot cleaner now, and easier to use as we can just plug-in an `IAccount[]` into the `GroupedAccountSelector`. PTAL :)

thisconnect:
> When going to /buy/info In testnet (make servewallet) the keystore prompts directly.

thanks @shonsirsha and @Beerosagos for looking into this. looks like it is correct behavior if there is only 1 account that is supported by the exchanges (i.e. in testnet only moonpay sells TBTC).

but weird that i had it as I always had 3 watch-only wallets with at least TBTC 1 account each. maybe it was loading weirdly?

anyway I can't reproduce the issue anymore since this branch was rebased. soo all good 👍

thisconnect (commented):
LGTM with small nits

I'll do some more testing.

thisconnect (approved):
LGTM

PR #2498, opened 2024-01-22
frontend: downgrade qrscanner from 1.4.2 to 1.3.0

thisconnect:
Some Android versions seem to crash with qr-scanner@1.4.2

Older verison does not have:
- native BarcodeDetector
- highlight region

### DIFF
--- a/frontends/web/package.json
+++ b/frontends/web/package.json
@@ -26,7 +26,7 @@
     "flag-icons": "^6.6.6",
     "i18next": "21.6.16",
     "lightweight-charts": "4.1.1",
-    "qr-scanner": "1.4.2",
+    "qr-scanner": "1.3.0",
     "react": "18.2.0",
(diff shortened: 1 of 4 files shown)


frontends/web/package.json
```diff
@@ -26,7 +26,7 @@
     "flag-icons": "^6.6.6",
     "i18next": "21.6.16",
     "lightweight-charts": "4.1.1",
-    "qr-scanner": "1.4.2",
+    "qr-scanner": "1.3.0",
```
benma, line 29: Add a comment here to be careful with upgrading this one? Otherwise a regression will come easily.

Also, which Android versions don't work, and how did you figure it out?

thisconnect: I don't know how to add comments in JSON 🙈 

For now this is only a draft for testing.

I'll update the commit message.

PR #2503, opened 2024-01-23
frontends: add first prototype of the iOS app

benma:
This is quick prototype of making the app run on iOS.

It was produced with heavy help of ChatGPT.

When setting it up, there was a choice between SwiftUI and UIKit. I
chose SwiftUI because it is newer, but have not checked what this
means for iOS version compatibility - it may shut out some old
versions.

The app runs in the simulator. It implements some but not all of the
backend environment interface functions, some tested and some not -
we'll need to go thorugh all of them in due time.

It works basically the same way as the Android app: a WebView loads
the web bundle, the mobileserver is built using gomobile and linked
in, and they communicate with each other in-process (without a HTTP server).

For now the Mobileserver.xcframework library is referenced but not
embeded and not signed, as some Info.plist is missing - to be fixed
later. When the framework has the needed Info.plist file, we can
toggle it from 'Do not embed' to 'Embed & Sign' and it should in
theory work for deployment.

BitBox02 (USB) support is not part of this commit.


benma:
> Tested and works pretty nicely!
> 
> One question: are we sure we want to merge directly into master updating also the build.md?

I think so - imo it does not disturb other developments and making a staging branch for it seems like needless overhead.

PR #2504, opened 2024-01-24
frontend/bitsurance: fix bug in dashboard

Beerosagos:
Account's balance was fetched without checking if the account was active. This caused unexpected errors during the component reload after disconnecting a wallet.


thisconnect (approved):
thanks LGTM 👍

PR #2508, opened 2024-01-24
frontend: fix restore-mnemonic error in webdev

thisconnect:
This fixes a failed-error at the end of restore-from-mnemonic workflow. Although there was a fail meassage everything worked normally after (new keystore showed up in sidebar etc). Could only reproduce in webdev. but not in 4.40.0 release or qt-osx build form master.

It seems to happen when restoreFromMnemonic is called multiple times. This can happend due to state changes in parent wizard component.

restoreFromMnemonic is a long-poll request that starts the workflow and stays open until the whole process on the device is completed. Maybe the backend could be improved as well and keep track if the workflow was already initiated.
Also this might timeout, we should test.

The frontend fix is to keep a reference if the api was called. This could probably be written as hook.

Also used useCallback hooks in parent component, so optimize the effects so that onAbort does cause the effect to exectute.

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
+++ b/frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { restoreFromMnemonic, errUserAbort } from '../../../../api/bitbox02';
 import { alertUser } from '../../../../components/alert/Alert';
@@ -72,22 +72,34 @@ export const RestoreFromMnemonic = ({
   onAbort,
 }: Props) => {
   const { t } = useTranslation();
+  const [errorCode, setErrorCode] = useState<number>();
+  const inProgress = useRef<string>();
 
   useEffect(() => {
-    restoreFromMnemonic(deviceID)
-      .then(result => {
-        if (!result.success) {
-          const errorText = result.code === errUserAbort
-            ? t('bitbox02Wizard.restoreFromMnemonic.e104')
-            : t('bitbox02Wizard.restoreFromMnemonic.failed');
-          alertUser(errorText, {
-            asDialog: false,
-            callback: () => onAbort(),
-          });
-        }
-      })
-      .catch(console.error);
-  }, [deviceID, onAbort, t]);
+    if (inProgress.current !== deviceID) {
+      // remember for which device the workflow already has been initialized
+      inProgress.current = deviceID;
+      restoreFromMnemonic(deviceID)
+        .then(result => {
+          if (!result.success) {
+            setErrorCode(result.code);
+          }
+        })
+        .catch(console.error);
+    }
+  }, [deviceID]);
+
+  useEffect(() => {
+    if (errorCode) {
+      const errorText = errorCode === errUserAbort
+        ? t('bitbox02Wizard.restoreFromMnemonic.e104')
+        : t('bitbox02Wizard.restoreFromMnemonic.failed');
+      alertUser(errorText, {
+        asDialog: false,
+        callback: () => onAbort(),
+      });
+    }
+  }, [errorCode, onAbort, t]);
 
   return (
     <Wait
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
```diff
@@ -72,22 +72,34 @@ export const RestoreFromMnemonic = ({
...
+        .then(result => {
+          if (!result.success) {
+            setErrorCode(result.code);
+          }
+        })
+        .catch(console.error);
+    }
+  }, [deviceID]);
+
+  useEffect(() => {
```
benma, line 92: why a separate useEffect and not just handle it above like before?

frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
```diff
@@ -72,22 +72,34 @@ export const RestoreFromMnemonic = ({
...
-      })
-      .catch(console.error);
-  }, [deviceID, onAbort, t]);
+    if (inProgress.current !== deviceID) {
+      // remember for which device the workflow already has been initialized
+      inProgress.current = deviceID;
+      restoreFromMnemonic(deviceID)
+        .then(result => {
+          if (!result.success) {
+            setErrorCode(result.code);
```
benma, line 85: don't you have to reset inProgress.current? Maybe if you abort or an error happens, you cannot start it again?

benma:
> It seems to happen when restoreFromMnemonic is called multiple times

when/how could it be called multiple times?

frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
thisconnect: There is no way to start the logic from within this component.

The workflow starts right-aways when this View is mounted.

When the workflow ends it will either load a different view (controlled in parent wizard logic) or show an error + onAbort callback that calls a function from wizard.

does my explanation make sense more or less?

frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
thisconnect: I noticed that onAbort callback changed and caused the effect with restoreFromMnemonic to execute again. So I tried to make an effect with as little depenencies as possible. and one for error handling.

thisconnect:
Additionally (or instead) of this change

> Maybe the backend could be improved as well and keep track if the workflow was already initiated.

In which case I imagine the backend should not start the workflow again.. but use the last connection to reply once workflow is done.

these are all problem related to long-poll. I checked if create-wallet, restore-sdcard could have similar issue but didn't notice.

benma:
What's the error exactly? I can imagine that if you call the restore workflow twice, it will simply be called a 2nd time after the first one has finished, but the 2nd must fail because the bb02 is already initialized. but still the question is how it could be invoked twice

thisconnect:
> when/how could it be called multiple times?

as written in commit message: "This can happened due to state changes in parent wizard component."

but steps to reproduce is:

1) open devtools network tab
2) click on "Restore from recovery words" in webdev
3) see 3 requests to devices/bitbox02/${deviceID}/restore-from-mnemonic (expected is 1)

Once RestoreFromMnemonic is mounted each subscriptions or useLoad in the wizard can trigger a re-render. I didn't find exactly which state change caused it. At this point I just noticed that the api replied instantly with success: false on the other 2 requests. Then I put effort into making the component work nicely with this long-polling requests that is fired when mounted.


sidenote: 
With the onAbort in a useCallback, I could reduce the times restoreFromMnemonic was called from 3 to 2. 
This is not really needed as with the separate effect we can keep track if the api was called. But I thought it is cleaner to keep onAbort in useCallback hook.

thisconnect:
> 
What's the error exactly? I can imagine that if you call the restore workflow twice, it will simply be called a 2nd time after the first one has finished, but the 2nd must fail because the bb02 is already initialized. but still the question is how it could be invoked twice

success false.

Please note that this is long-polling request. This means the backend doesn't respond until the workflow has ended.
This request could stay open for 20 minutes. once the workflow is ended the backend replies to this long-poll request with success true or false.

when calling this same api many times at the same time, I think only the first one will long-poll, but the backend won't re-use the 2nd or 3rd for long-polling, but imediatelly reply with success false.. (as far as I tested).

But this all doesn't explain why the bug doesn't appear in Qt 🤔

thisconnect:
closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/2514

PR #2510, opened 2024-01-25
frontend: update the bitsurance guide links

thisconnect:
Just small fix, as we already use /redirects/ on the website


Beerosagos (approved):
LGTM

PR #2511, opened 2024-01-25
frontend/bitsurance: check keystore when skipping account selection

Beerosagos:
When there is only one insurable account, the Bitsurance workflow skips the account selection step. Doing it, it was missing to request the keystore connection in case it was disconnected. This fixes the issue.


shonsirsha (approved):
LGTM 🫡

PR #2512, opened 2024-01-25
frontend/rates: fix bug in rotate fiat

Beerosagos:
`rotateFiat` function of the RatesProvider was missing to update the app config with the new default fiat selected. This fixes the issue.


shonsirsha (approved):
Good catch! LGTM.

PR #2513, opened 2024-01-25
frontend/account-summary: group account-summary balance by wallet

Beerosagos:
With the introduction of the watch-only feature, accounts belonging to different keystores can be shown in the account-summary page. This refactors both the frontend and the backend logic to show the accounts grouped by wallet.


frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -55,73 +63,87 @@ export function SummaryBalance ({
...
+                    <strong>
+                      <Amount amount={totalBalance.total} unit={totalBalance.fiatUnit}/>
+                    </strong>
+                    {' '}
+                    <span className={style.coinUnit}>
+                      {totalBalance.fiatUnit}
+                    </span>
+                  </>
+                ) : (<Skeleton />) }
               </td>
```
shonsirsha, line 142: **nit:** this has always been there, but I generally find this component to be a bit hard to read due to the number of conditional renders.

I think we could make the component easier to read by extracting these components into its own. E.g, here we could have `<TotalBalance ... />` component - also applies to the one above (lets say `TotalCoinBalance` or smt).

Just a nit, we could also do it another time ofc :)

shonsirsha (approved):
Approving for FE. FE-wise looks good. tested LGTM. Thanks for finishing off this feature 👍

frontends/web/src/routes/account/summary/summarybalance.tsx
Beerosagos: Good point. Done! I only did it for the TotalBalance, as the coin total looks already good to me with the `SubTotalRow` component. Did you have something different in mind?

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -55,73 +77,79 @@ export function SummaryBalance ({
...
+                  balanceRows.push(
+                    <SubTotalRow
+                      key={account.coinCode}
+                      coinCode={account.coinCode}
+                      coinName={account.coinName}
+                      balance={totalBalancePerCoin && totalBalancePerCoin[coinCode]}
+                    />);
+                }
+                return balanceRows;
+              })
```
shonsirsha, line 130: @Beerosagos I was referring to this whole condition. Doing balanceRows.push inside the whole component looks a bit atypical, to me. I'd personally take it out into a new function, maybe like `renderBalanceRows(coins)` or smt like that.

But if it's clear to you then no big deal tbh :smile: :+1:

backend/accounts.go
```diff
@@ -58,6 +58,10 @@ const accountsHardLimit = 5
 // AccountsList is an accounts.Interface slice which implements a lookup method.
 type AccountsList []accounts.Interface
 
+// KeystoresAccountsListMap is a map where keys are keystores' fingerprints and values are
```
benma, line 61: the keys are Keystore object pointers, not fingerprints. I think it would be better to make the key the actual fingerprint (as bytes if possible, otherwise as hex string), as it's the only thing needed, and then one does not have to worry about pointers (multiple different pointers could point to the same keystore object, for example).

backend/accounts.go
```diff
@@ -58,6 +58,10 @@ const accountsHardLimit = 5
...
+// AccountsLists of accounts belonging to each keystore.
+type KeystoresAccountsListMap map[*config.Keystore]AccountsList
```
benma, line 63: why is this in accounts.go but `AccountsByKeystore()` which uses it is in backend.go? seems like they should both be in accounts.go.

backend/backend.go
```diff
@@ -540,6 +544,97 @@ func (backend *Backend) Accounts() AccountsList {
 	return backend.accounts
 }
 
+// AccountsByKeystore returns a map of the current accounts of the backend, grouped
+// by keystore.
+func (backend *Backend) AccountsByKeystore() (KeystoresAccountsListMap, error) {
+	accountsByKeystore := KeystoresAccountsListMap{}
+	defer backend.accountsAndKeystoreLock.RLock()()
```
benma, line 551: maybe move it to the first line, looks a bit strange on the second :P

backend/backend.go
```diff
@@ -540,6 +544,97 @@ func (backend *Backend) Accounts() AccountsList {
 	return backend.accounts
 }
 
+// AccountsByKeystore returns a map of the current accounts of the backend, grouped
+// by keystore.
+func (backend *Backend) AccountsByKeystore() (KeystoresAccountsListMap, error) {
```
benma, line 549: unit test would be nice

backend/backend.go
```diff
@@ -540,6 +544,97 @@ func (backend *Backend) Accounts() AccountsList {
...
+		keystore, err := backend.Config().AccountsConfig().LookupKeystore(rootFingerprint)
+		if err != nil {
+			return nil, err
+		}
+		accountsByKeystore[keystore] = append(accountsByKeystore[keystore], account)
+	}
+	return accountsByKeystore, nil
+}
+
+// KeystoreTotalAmount represents the total balance amount of the accounts belongings to a keystore.
```
benma, line 567: s/belongings/belonging.

backend/backend.go
```diff
@@ -540,6 +544,97 @@ func (backend *Backend) Accounts() AccountsList {
...
+		fiatUnit = "sat"
+	}
+
+	formatBtcAsSat := util.FormatBtcAsSat(backend.Config().AppConfig().Backend.BtcUnit)
+
+	accountsByKeystore, err := backend.AccountsByKeystore()
+	if err != nil {
+		return nil, err
+	}
+	for keystore, accountList := range accountsByKeystore {
```
benma, line 591: a lot of duplication here with chart.go. The total computed in chart.go is basically the sum of the keystore totals here right? seems like code re-use could be possible here, or am i missing something?

no need to do it in this PR, but it would be nice to de-duplicate.

backend/backend.go
```diff
@@ -540,6 +544,97 @@ func (backend *Backend) Accounts() AccountsList {
...
+// KeystoreTotalAmount represents the total balance amount of the accounts belongings to a keystore.
+type KeystoreTotalAmount = struct {
+	// FiatUnit is the fiat unit of the total amount
+	FiatUnit string `json:"fiatUnit"`
+	// Total formatted for frontend visualization
+	Total string `json:"total"`
+}
+
+// AccountsTotalBalanceByKeystore returns a map of accounts' total balances across coins, grouped by keystore.
+func (backend *Backend) AccountsTotalBalanceByKeystore() (map[string]KeystoreTotalAmount, error) {
```
benma, line 576: maybe also move to accounts.go. unit test would be nice here too :P

backend/backend.go
benma, line 549: i wonder if this could replace the `getAccountsByKeystore()` frontend function which basically does the same.

benma (approved):
tACK, looks nice! comments could be addressed also in future PRs.

Beerosagos:
I'll just merge to avoid blocking the next RC, but I'll address all the remaining comments in a new PR. Thanks both @benma @shonsirsha :pray:

backend/accounts.go
Beerosagos: Addressed in #2609

backend/accounts.go
Beerosagos: Addressed in #2609

backend/backend.go
Beerosagos: Addressed in #2609

backend/backend.go
Beerosagos: Addressed in #2609

backend/backend.go
Beerosagos: Addressed in #2609

backend/backend.go
Beerosagos: Partially addressed in #2609 (unit test will come in a separate PR)

backend/backend.go
Beerosagos: I had a look but I think that requires a bit of rework and deserves a separate PR, since the accounts are fetched in app.tsx and then passed as properties to the child components. Maybe something for @shonsirsha ? :)

PR #2509, opened 2024-01-25
frontend: validate devicename input field in setup

thisconnect:
We already have a backend validation, but that only shows an invalid input message after the user clicked continue.

Better experience is to give live feedback and information about what is wrong.

It should only allow: printable ascii chars, with max length of 30 including spaces, but not starting or ending with (that is just trimmed, without showing an error to the user).

PR #2514, opened 2024-01-26
Frontend restore mnemonic with name

thisconnect:
On top of https://github.com/digitalbitbox/bitbox-wallet-app/pull/2509

Restoring from mnemonic did not allow to set a device name so it
was always 'My BitBox'.

Added set device name step before continuing with restoreMnemonic.

This also fixes a bug in webdev that was caused by calling
restoreMnemonic on mount, ending up with multiple api calls and
a fail message after the workflow was done.

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -264,7 +264,7 @@
       "unlock": "Enter BitBox02 password to unlock."
     },
     "stepCreate": {
-      "description": "This name is used as the device name and for the backup.",
+      "description": "This name is used as the device name and for backups.",
       "error": {
         "genericMessage": "Use letters, numbers, basic symbols, spaces. Max 30 characters.",
         "invalidChars": "Name contains invalid characters: {{invalidChars}}.",
--- a/frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
+++ b/frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
@@ -14,14 +14,15 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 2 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -264,7 +264,12 @@
       "unlock": "Enter BitBox02 password to unlock."
     },
     "stepCreate": {
-      "description": "This name is used as the device name and for the backup.",
+      "description": "This name is used as the device name and for backups.",
```
thisconnect, line 267: I changed this to get feedback in this review.

This message is currently used when setting the device name before creating a new wallet OR with this pr when restoring from mnemonic.

Phrasing: "the backup" implies that there will be always a sdcard backup. But when restoring from mnemonic, no sdcard is involved. But it is not wrong as the user still could make a backup on a sdcard in the future.

Ok to change it so it works for both create-new-wallet and restore-from-mnemonic ?

let me know then I'll update locize directly.

with this wording:

Screen Shot 2024-01-26 at 09 19 55

frontends/web/src/routes/device/bitbox02/setup/wallet-restore.tsx
```diff
@@ -72,12 +73,13 @@ export const RestoreFromMnemonic = ({
   onAbort,
 }: Props) => {
   const { t } = useTranslation();
+  const [status, setStatus] = useState<'intro' | 'setName' | 'restoreMnemonic'>('intro');
 
-  useEffect(() => {
-    restoreFromMnemonic(deviceID)
+  const restoreMnemonic = () => {
+    bitbox02.restoreFromMnemonic(deviceID)
       .then(result => {
```
benma, line 80: why no `async/await` :sob:

benma (approved):
tACK, very niiice :tada:

thisconnect:
rebased.

frontends/web/src/locales/en/app.json
thisconnect: changed and updated locize.

PR #2515, opened 2024-01-26
frontend: sort accounts in manage-accounts

thisconnect:
Only the active accounts were sorted before, but manage-accounts has a list of all account including inactive accounts which had different order.

Changed so that all accounts are sorted the same way.


Beerosagos (approved):
utACK

PR #2516, opened 2024-01-29
frontend/settings: remove dead code form manage account

Beerosagos:
In previous commit 4fb2f72 we commented out the code to enable/disable watch-only at an account level. Since we didn't receive any request about that now we are completely removing it from the code base.


thisconnect (approved):
LGTM

PR #2523, opened 2024-01-31
backend/btc: update SignBTCAddress and add tests

Beerosagos:
Removes an unnecessary check in the function and adds a unit test function.

This was suggested in the review of #2485

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -838,26 +838,12 @@ func (account *Account) VerifyExtendedPublicKey(signingConfigIndex int) (bool, e
 	return false, nil
 }
 
-// requestAddressScriptTypeMap maps from format codes specified by request-address js library
-// to our own script type codes. See https://www.npmjs.com/package/request-address
-var requestAddressScriptTypeMap = map[string]signing.ScriptType{
-	"p2pkh":  signing.ScriptTypeP2PKH,
-	"p2wpkh": signing.ScriptTypeP2WPKH,
-	"p2sh":   signing.ScriptTypeP2WPKHP2SH,
-	"p2tr":   signing.ScriptTypeP2TR,
-}
-
-// defaultScriptType is the default type used for address sharing and verification, if the
-// `format` param is not defined.
-var defaultScriptType = "p2wpkh"
-
 // SignBTCAddress returns an unused address and makes the user sign a message to prove ownership.
 // Input params:
 //
 //	`account` is the account from which the address is derived.
 //	`message` is the message that will be signed by the user with the private key linked to the address.
-//	`format` is the script type that should be used in the address derivation, as received by the widget
-//		(see https://github.com/pocketbitcoin/request-address#requestaddressv0messagescripttype).
+//	`format` is the script type that should be used in the address derivation.
 //		If format is empty, native segwit type is used as a fallback.
 //
 // Returned values:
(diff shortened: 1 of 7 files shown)


backend/coins/btc/account.go
```diff
@@ -867,10 +866,6 @@ var defaultScriptType = "p2wpkh"
 //	#2: base64 encoding of the message signature, obtained using the private key linked to the address.
 //	#3: is an optional error that could be generated during the execution of the function.
 func SignBTCAddress(account accounts.Interface, message string, format string) (string, string, error) {
-	if !exchanges.IsPocketSupported(account) {
```
Beerosagos, line 870: This is not needed anymore, as we previously moved the endpoint that calls the function from handlers/handlers.go to coins/btc/handlers.go, so it will always be called on btc accounts. Also this is not exclusively used for Pocket anymore

Beerosagos:
ping @benma :innocent:

backend/coins/btc/account.go
benma: the function signature is still closely tied to pocket or the request-address NPM package:

```
`format` is the script type that should be used in the address derivation, as received by the widget
//		(see https://github.com/pocketbitcoin/request-address#requestaddressv0messagescripttype).
//		If format is empty, native segwit type is used as a fallback.
```

Probably too much overhead now, but maybe better if this function took `signing.ScriptType` directly and the conversion happened at the callsite. I'd just worry a bit that in the future there will be other uses of this function unrelated to the NPM package, in which case the stringly-typed format is not so great.

backend/coins/btc/account_test.go
```diff
@@ -193,3 +184,17 @@ func TestInsuredAccountAddresses(t *testing.T) {
 	require.Equal(t, *account2.GetUnusedReceiveAddresses()[0].ScriptType, signing.ScriptTypeP2WPKH)
 
 }
+
+func TestSignAddress(t *testing.T) {
+	account := mockAccount(t, nil)
+	require.NoError(t, account.Initialize())
+	// pt2r is not an available script type in the mocked account.
+	_, _, err := btc.SignBTCAddress(account, "Hello there", "p2tr")
+	require.NotNil(t, err)
```
benma, line 193: `Error` is the method for checking errors

backend/coins/btc/account_test.go
```diff
@@ -193,3 +184,17 @@ func TestInsuredAccountAddresses(t *testing.T) {
...
+	address, signature, err := btc.SignBTCAddress(account, "Hello there", "p2wpkh")
+	require.Nil(t, err)
```
benma, line 195: `NoError()`

backend/coins/btc/account_test.go
```diff
@@ -193,3 +184,17 @@ func TestInsuredAccountAddresses(t *testing.T) {
...
+	require.NotNil(t, address)
```
benma, line 196: address is a string, strange to compare for `nil`. probably `require.NotEmpty()` is better.

backend/coins/btc/account_test.go
```diff
@@ -193,3 +184,17 @@ func TestInsuredAccountAddresses(t *testing.T) {
...
+	require.Equal(t, signature, base64.StdEncoding.EncodeToString([]byte("signature")))
```
benma, line 197: in Equal(), the first value is the expected and the 2nd the one tested for, so you should reverse the two.

benma (changes requested):

Beerosagos:
rebased

backend/coins/btc/account.go
Beerosagos: Yeah, you're right. Since I was there, I added a second commit to refactor it. Now the frontend has an util that converts from the request-address library script type to the script type used in the rest of the app, and the backend doesn't depend on it anymore.

Beerosagos:
Pushed a new update. @benma PTAL :pray:

Beerosagos:
ping @benma

benma (approved):
LGTM

PR #2524, opened 2024-02-01
frontend: replace BrowserRouter with HashRouter

benma:
Using the BrowserRouter is needlessly complicated, we don't need the browser to make a new "request" when we route to a different page. Getting this to work is especially difficult in iOS, where the file path to index.html is loaded in the webview, and routing to e.g. `/account-summary' fails because the absolute filepath `/account-summary` does not exist and is outside the app folder sandbox.

Using the hash router instead, where the app path is tracked behind the `#` (e.g. index.html#/account-summary) is more appropriate and easier to handle.


benma:
This is a nice change but it destroys allowing requests to third parties based on the current url (e.g. can load moonpay when on buy page, etc.), because with the hash router, the Android webview does not register url changes anymore.

The blocking of external requests can probably be circumvented by potential malware easily enough, so the benefit of doing this blocking while everything on some pages is questionable. 

We could simply remove the blocking and proceed with this PR.

We could also abandon this PR, as I figured out how to fix routing on iOS without changing to the HashRouter.

thisconnect:
> We could also abandon this PR, as I figured out how to fix routing on iOS without changing to the HashRouter.

👍 closing

PR #2525, opened 2024-02-01
frontend/account-summary: fix regression during address scanning

Beerosagos:
Recent commit d8c6ba0 refactored the account summary section to group accounts by keystore. The commit also introduced a small regression, hiding the account rendering until the balance of each account is ready, while the previous behavior was to show the accounts displaying the scanned addresses.

This fixes the regression.


shonsirsha (approved):
good catch. tested LGTM.

PR #2526, opened 2024-02-02
frontend: use unique keys for sidebar items

thisconnect:
With watch-only items in the sidebar are now groupped by keystore, but there is a rare issue that React gets confused and renders one account at the top of the sidebar.

This change explicitly sets keys for all items in the sidebar, and wrapps keystores and the associated  account in a separate container. This should theoretically fix the rare rendering bug.


Beerosagos (approved):
utACK

PR #2531, opened 2024-02-05
frontend: fix line height / alignment on plus icon in the insure account btn

shonsirsha:
Before:
ewdsa

After:
wdsa


Beerosagos (approved):
tACK

PR #2532, opened 2024-02-05
backend: replace BIP69 with random sorting for tx inputs/outputs

strmci:
This commit refactors the sorting mechanism for transaction inputs and outputs. It replaces the existing BIP69 lexicographical sorting with a randomized sorting approach.

BIP69, or "Lexicographic Order of Transaction Inputs and Outputs," was initially proposed to enhance privacy and mitigate blockchain fingerprinting. However, due to its limited adoption and effectiveness, it has been replaced by random shuffling of inputs and outputs, offering improved privacy benefits.

### DIFF
--- a/backend/coins/btc/maketx/maketx.go
+++ b/backend/coins/btc/maketx/maketx.go
@@ -15,10 +15,13 @@
 package maketx
 
 import (
+	"crypto/rand"
+	"encoding/binary"
+	mrand "math/rand"
 	"sort"
+	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
-	"github.com/btcsuite/btcd/btcutil/txsort"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/errors"
@@ -168,7 +171,10 @@ func NewTxSpendAll(
 		TxOut:    []*wire.TxOut{output},
 		LockTime: 0,
 	}
-	txsort.InPlaceSort(unsignedTransaction)
+
+	secureRand := mrand.New(mrand.NewSource(secureSeed()))
+	shuffleTxInputsAndOutputs(unsignedTransaction, secureRand)
+
 	log.WithField("fee", maxRequiredFee).Debug("Preparing transaction to spend all outputs")
 
 	setRBF(coin, unsignedTransaction)
@@ -249,7 +255,10 @@ func NewTx(
 		} else {
 			changeAddress = nil
 		}
-		txsort.InPlaceSort(unsignedTransaction)
+
+		secureRand := mrand.New(mrand.NewSource(secureSeed()))
+		shuffleTxInputsAndOutputs(unsignedTransaction, secureRand)
+
 		log.WithField("fee", finalFee).Debug("Preparing transaction")
 
 		setRBF(coin, unsignedTransaction)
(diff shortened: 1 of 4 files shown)


backend/coins/btc/maketx/maketx.go
```diff
@@ -168,7 +168,11 @@ func NewTxSpendAll(
 		TxOut:    []*wire.TxOut{output},
 		LockTime: 0,
 	}
-	txsort.InPlaceSort(unsignedTransaction)
+
+	// Shuffle inputs and outputs
+	ShuffleTxInputs(unsignedTransaction)
```
Beerosagos, line 173: small nit: looks like you don't really need two separate functions, right?

strmci: I agree, having just one function is sufficient, I've made the update, PTAL

backend/coins/btc/maketx/maketx.go
```diff
@@ -263,3 +267,16 @@ func NewTx(
 		}, nil
 	}
 }
+
+// ShuffleTxInputsAndOutputs shuffles both the TxIn and TxOut slices of a wire.MsgTx.
+func ShuffleTxInputsAndOutputs(tx *wire.MsgTx) {
```
benma, line 272: can unexport this `shuffleTxInputsAndOutputs` as it is only used locally.

benma (commented):
Would be great if you could explain the reason for this in the commit msg

Beerosagos (approved):
tACK!

Would be nice to have a test, but I don't think there is a good way to check that the ins and outs are actually shuffled. wdyt @benma ?

benma:
> tACK!
> 
> Would be nice to have a test, but I don't think there is a good way to check that the ins and outs are actually shuffled. wdyt @benma ?

test would be nice, and you can do it by using a local `rand.Rand` object which is initialized with a constant seed in tests.

strmci:
@Beerosagos @benma  I made some updates:
- updated the `ShuffleTxInputsAndOutputs` so it uses the `rand.Rand` object and added unit test
- kept it exported as it is used in the test file
- updated changelog file
PTAL

backend/coins/btc/maketx/maketx.go
```diff
@@ -168,7 +171,10 @@ func NewTxSpendAll(
 		TxOut:    []*wire.TxOut{output},
 		LockTime: 0,
 	}
-	txsort.InPlaceSort(unsignedTransaction)
+
+	secureRand := mrand.New(mrand.NewSource(secureSeed()))
```
Beerosagos, line 175: I'm not 100% sure, but I think that it would be better to seed the generator using a random int64 obtained from the default rand generator.  @benma wdyt?

backend/coins/btc/maketx/maketx_test.go
```diff
@@ -312,3 +313,40 @@ func (s *newTxSuite) TestNewTxCoinSelection() {
...
+		{Value: 3000000, PkScript: []byte{}},
+	}
+
+	// Create a new transaction and add inputs and outputs
+	tx := wire.NewMsgTx(wire.TxVersion)
+	tx.TxIn = txIns
+	tx.TxOut = txOuts
+	// Shuffle with constant seed
+	testRand := rand.New(rand.NewSource(1000))
+	maketx.ShuffleTxInputsAndOutputs(tx, testRand)
```
benma, line 336: change the first line from `package maketx_test` to `package maketx`, then you can access the unexported stuff. this way, you can unexport `ShuffleTxInputsAndOutputs()` again (it should not be exported as it is only used internally).

benma (changes requested):
utACK, with a small nit about unexporting the function again

benma:
> also please add a note to CHANGELOG.md 😄

@strmci also this ^

strmci:
@benma  

- as we discussed, I created a separate file `maketx_internal_test.go` to keep the function unexported
- CHANGELOG.md was already updated

PTAL

benma (approved):
LGTM

PR #2533, opened 2024-02-05
backend/handlers: fix user abor error handling

Beerosagos:
ETH sign related endpoints were incorrectly handling user abort error generated in case the user aborted the connect keystore popup. This commit fixes the issue.


benma (approved):
utACK

PR #2535, opened 2024-02-06
frontend: ensure keystore connected in aopp workflow

thisconnect:
With remember wallet the user could choose an account that is not connected, this ensures that the correct keystore is used.


thisconnect:
closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/2538

PR #2537, opened 2024-02-06
backend: use new URLs for update/banners

benma:
The desktop.json/banners.json files are moved from the bitbox.swiss website to a dedicated domain to serve files to the BitBoxApp.

The old URLs will do a 301 redirect to the new ones - old BitBoxApp versions will automatically follow the redirects.


thisconnect (approved):
curl'ed LGTM

PR #2538, opened 2024-02-06
backend/aopp: fix bug in aopp signing flow

Beerosagos:
Since watch-only feature introduction, AOPP doesn't correctly handle the available accounts list, making it possible to request signing a message for an account, while having a different wallet connected. In that case there is a mismatch between the addresses displayed by the bbapp and the wallet. This also causes the signing flow to fail because of the mismatch between the address and the signature.

This commit gives a poor man's fix to the issue, filtering the available accounts and including only those belonging to the currently connected keystore.

### DIFF
--- a/backend/aopp.go
+++ b/backend/aopp.go
@@ -179,6 +179,16 @@ func (backend *Backend) aoppKeystoreRegistered() {
 	var accounts []account
 	var filteredDueToScriptType bool
 	for _, acct := range backend.accounts {
+		accountFingerprint, err := acct.Config().Config.SigningConfigurations.RootFingerprint()
+		if err != nil {
+			backend.log.WithError(err).Error("Account rootfingerprint not available")
+			backend.aoppSetError(errAOPPUnknown)
+			return
+		}
+
+		if err := compareRootFingerprint(backend.keystore, accountFingerprint); err != nil {
+			continue
+		}
 		if acct.Config().Config.Inactive || acct.Config().Config.HiddenBecauseUnused {
 			continue
 		}
--- a/backend/aopp_test.go
+++ b/backend/aopp_test.go
@@ -37,6 +37,7 @@ const dummyMsg = "message to be signed"
 const dummySignature = "signature"
 
 var rootFingerprint = []byte{0x55, 0x055, 0x55, 0x55}
+var rootFingerprint2 = []byte{0x66, 0x066, 0x66, 0x66}
 
 func defaultParams() url.Values {
 	params := url.Values{}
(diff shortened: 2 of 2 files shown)


backend/aopp.go
```diff
@@ -179,6 +179,15 @@ func (backend *Backend) aoppKeystoreRegistered() {
 	var accounts []account
 	var filteredDueToScriptType bool
 	for _, acct := range backend.accounts {
+		var accountFingerprint, err = acct.Config().Config.SigningConfigurations.RootFingerprint()
```
benma, line 182: strange syntax, why not the usual `accountFingerprint, err := ...`?

backend/aopp.go
```diff
@@ -179,6 +179,15 @@ func (backend *Backend) aoppKeystoreRegistered() {
...
+		if err != nil {
+			backend.log.WithError(err).Error("Account rootfingerprint not available")
+			backend.aoppSetError(errAOPPUnknown)
+		}
```
benma, line 186: missing `return`

benma (changes requested):

backend/aopp.go
Beerosagos: lol, copy+paste went bad :see_no_evil:

Beerosagos:
@benma updated, PTAL :pray:

benma (approved):
update unit tests? :smiley_cat:

backend/aopp_test.go
```diff
@@ -305,6 +305,22 @@ func TestAOPPSuccess(t *testing.T) {
...
 	})
+	//  Keystore watch-only is enabled, but the keystore is still required to sign, as its accounts are not available
+	// in the AOPP flow.
+	t.Run("watch-only", func(t *testing.T) {
+		b := newBackend(t, testnetDisabled, regtestDisabled)
+		defer b.Close()
+		params := defaultParams()
+		b.registerKeystore(makeKeystore(t, scriptTypeRef(signing.ScriptTypeP2WPKH), keystoreHelper.ExtendedPublicKey))
+		fingerprint, err := b.keystore.RootFingerprint()
+		require.Nil(t, err)
```
benma, line 316: `require.NoError(t, err)`, but you could also just the global `rootFingerprint` which is used everywhere in these tests

benma (approved):
Thanks. The unit tests are also missing the case where the accounts are filtered to the connected keystore fingerprint.

Not a blocker for mering/releasing.

Beerosagos:
Thanks, changed the test to explicitly cover the aopp change, verifying the root fingerprint of the AOPP accounts

PR #2539, opened 2024-02-06
backend: adjust chart display to start from zero after first tx

strmci:
This commit modifies the chart display after the initial transaction. Previously, the chart would begin at the first fiat value of the transaction,  resulting in only one data point visible in the middle of the chart. With this update, the chart now starts at zero, providing a better view of the data.


Beerosagos (approved):
tACK!

thisconnect:
@Beerosagos can this be merged?

Beerosagos:
> @Beerosagos can this be merged?

Yep!

PR #2542, opened 2024-02-07
frontend: fix disabled chart filters after user empty wallet

strmci:
This commit addresses the problem of disabled chart filters in the portfolio view when users empty their wallet. They were unable to change the chart view once they emptied their wallet.

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -135,6 +135,10 @@ class Chart extends Component<Props, State> {
     return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length > 0;
   };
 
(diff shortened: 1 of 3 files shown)


thisconnect:
I sent everything away, and the filter are active 👍 

but the chart threw an error after I clicked on "week" and the app showed white screen. 🤔 

Screen Shot 2024-02-07 at 15 31 39

thisconnect:
Here is my chart data, for some reason chartDataHourly is an empty array, this seems to be the reason it errors.

but I think there is something else wrong: [chart-data.txt](https://github.com/digitalbitbox/bitbox-wallet-app/files/14195312/chart-data.txt)

thisconnect (approved):
tested LGTM thanks 👍

PR #2543, opened 2024-02-07
frontend: do not sort accounts

benma:
The backend has a fairly involved accounts sorting function already in `backend/accounts.go:sortAccounts`. The current frontend sorting function sorts lexicographically by fingerprint and account code, which is a different ordering than before, but has two more problems:

- account number 10 comes before account number 2, which is wrong
- account codes cannot be relied upon to have the current structure, they could be random IDs, in which case the accounts would be ordered randomly

The original sorting in the frontend was added in
937a023bfa16e9de69ce52da4b1b7df5c10333a1 to work around a React bug where accounts were ordered differently in the sidebar vs manage accounts, but since moving to React functional components in both places, this is not needed anymore.


thisconnect (approved):
LGTM

PR #2544, opened 2024-02-07
experiment] frontend: show spinner after unlock, before accounts are loaded

benma:
After unlock, especially the first time, it can take many seconds for the initial set of accounts to be loaded. Until then, the app shows nothing in the sidebar and 'Please connect your device to get started'.

The user is easily confused, it seems as something went wrong.

We can probably speed up this loading with firmware and app changes, but this is a quick improvement that shows a spinner until the accounts become available.

### DIFF
--- a/frontends/web/src/hooks/backend.ts
+++ b/frontends/web/src/hooks/backend.ts
@@ -1,0 +1,30 @@
+/**
+ * Copyright 2024 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect, useState } from 'react';
+import { TKeystores, subscribeKeystores, getKeystores } from '../api/keystores';
+
+export function useKeystores(): TKeystores | undefined {
+  const [keystores, setKeystores] = useState<TKeystores>();
+  useEffect(() => {
+    getKeystores().then(keystores => {
+      setKeystores(keystores);
+    });
+    // this passes the unsubscribe function directly the return function of useEffect, used when the component unmounts.
+    return subscribeKeystores(setKeystores);
+  }, []);
+  return keystores;
+}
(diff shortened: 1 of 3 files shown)


frontends/web/src/hooks/backend.ts
```diff
@@ -0,0 +1,30 @@
...
+  const [keystores, setKeystores] = useState<TKeystores>();
+  useEffect(() => {
+    getKeystores().then(keystores => {
+      setKeystores(keystores);
+    });
+    // this passes the unsubscribe function directly the return function of useEffect, used when the component unmounts.
+    return subscribeKeystores(setKeystores);
+  }, []);
+  return keystores;
+}
```
thisconnect, line 30: rename to `hooks/keystores`?

frontends/web/src/routes/device/waiting.tsx
```diff
@@ -16,17 +16,29 @@
...
 import { Footer, Header } from '../../components/layout';
 import style from './bitbox01/bitbox01.module.css';
 
 export const Waiting = () => {
   const { t } = useTranslation();
   const { isDarkMode } = useDarkmode();
-
+  const keystores = useKeystores();
+  const devices = useSync(getDeviceList, syncDeviceList);
+  console.log('LOL', keystores, devices);
```
thisconnect, line 36: console.log

thisconnect (changes requested):
nice 2 small nits

frontends/web/src/routes/device/waiting.tsx
```diff
@@ -27,6 +32,13 @@ export const Waiting = () => {
   const { t } = useTranslation();
   const { isDarkMode } = useDarkmode();
 
+  const keystores = useKeystores();
+  const devices = useSync(getDeviceList, syncDeviceList);
```
benma, line 36: i was hoping only the keystore check would be enough but somehow the frontend gets the keystore event way too late (looks like the backend/frontend is overloaded with work or handling events during that time, and we have no way to prioritize certain events at the moment), so i added the devices check too, which necessarily comes earlier (already before the unlock). software keystore

Beerosagos (commented):
It is a nice improvement, but the fact that the loading screen disappears after loading the first accounts, but other accounts can still be discovered and added to the sidebar looks a bit buggy to me.
Would it be possible to fire an event when the account discovery stops looking for new accounts and use that to hide the loading screen?

thisconnect:
> the fact that the loading screen disappears after loading the first accounts, but other accounts can still be discovered and added to the sidebar looks a bit buggy to me.

Generally I think it is better to show and render all data that is available, and if 1 account is already found, list it incrementally. I think it's better than having a loading screen for seconds/minutes and nothing happens.

@Beerosagos what if there was a small loading icon so that it is clear that it is still searching? (but also that would need some kind of discovery-is-finished event)

Beerosagos:
> @Beerosagos what if there was a small loading icon so that it is clear that it is still searching? (but also that would need some kind of discovery-is-finished event)

That would be really nice, imo

benma:
Closing in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/2683

PR #2550, opened 2024-02-12
frontend: preselect backup when there's only 1 & show message when there's more

shonsirsha:
Preselect backup when there's only 1, and show message of "Please select backup to continue" when there are more than 1 backups available.

[Preview]:

ONLY_1

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/setup/restore.tsx
+++ b/frontends/web/src/routes/device/bitbox02/setup/restore.tsx
@@ -48,7 +48,8 @@ export const RestoreFromSDCardBackup = ({
           showRestore={true}
           showRadio={true}
           onSelectBackup={onSelectBackup}
-          onRestoreBackup={onRestoreBackup}>
+          onRestoreBackup={onRestoreBackup}
+        >
           <Button
             secondary
             onClick={onBack}>
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -141,6 +141,7 @@
       },
       "restoring": "Restoring backup…",
       "selectedBackup": "<strong>{{backupName}}</strong> created on {{createdDateTime}} will be restored.",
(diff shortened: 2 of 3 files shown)


frontends/web/src/routes/device/bitbox02/setup/restore.tsx
```diff
@@ -43,12 +45,15 @@ export const RestoreFromSDCardBackup = ({
       width="700px">
       <ViewHeader title={t('backup.restore.confirmTitle')} />
       <ViewContent>
+        {showSubtitleText ? <p className="m-none">{t('backup.restore.subtitle')}</p> : null}
```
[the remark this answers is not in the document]
thisconnect: An alternative could be that this is rendered by BackupsV2 component itself, returning a  document fragment that conditionally includes this paragraph. 

Condition could be showRadio === true, as in this case the user is expected to selected one, that could be simpler WDYT? No strong opinion just wanted to discuss.

frontends/web/src/routes/device/bitbox02/setup/restore.tsx
```diff
@@ -48,7 +48,8 @@ export const RestoreFromSDCardBackup = ({
           showRestore={true}
           showRadio={true}
           onSelectBackup={onSelectBackup}
-          onRestoreBackup={onRestoreBackup}>
+          onRestoreBackup={onRestoreBackup}
+        >
```
thisconnect, line 52: Nit: revert this change?

thisconnect (approved):
tested f30a1ea9d965ff4f6345e765d2ca63deb678d497 LGTM 👍 
with small nit.

frontends/web/src/routes/device/bitbox02/setup/restore.tsx
thisconnect: ack 👍

benma:
A CHANGELOG entry would be nice :smile:

PR #2552, opened 2024-02-13
handlers: drop "Handler" suffix in method names and `_` var name

benma:
Some had the Handler suffix, some didn't. in btc/handlers, none have that suffix. This commit makes it consistent.

Same for the `_ *http.Request` change, where some have the `_` and some don't.


Beerosagos (approved):
utACK

PR #2557, opened 2024-02-14
frontend: split lightning send workflow into smaller components

thisconnect:
Having one big component with multiple conditional variable child components is convenient but can also lead to more spaghetti code.

Split Send into separate components, similar as we did in setup.

### DIFF
--- a/frontends/web/src/routes/lightning/send/send.tsx
+++ b/frontends/web/src/routes/lightning/send/send.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2023 Shift Crypto AG
+ * Copyright 2023-2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,7 @@ import * as accountApi from '../../../api/account';
 import { Column, Grid, GuideWrapper, GuidedContent, Header, Main } from '../../../components/layout';
 import { View, ViewButtons, ViewContent } from '../../../components/view/view';
 import { Button } from '../../../components/forms';
-import { InputType, InputTypeVariant, SdkError, getParseInput, postSendPayment } from '../../../api/lightning';
+import { InputType, InputTypeVariant, LnInvoice, SdkError, getParseInput, postSendPayment } from '../../../api/lightning';
 import { SimpleMarkup } from '../../../utils/markup';
 import { route } from '../../../utils/route';
 import { toSat } from '../../../utils/conversion';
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/lightning/send/send.tsx
Beerosagos, line 1: Not so important, but would be nice to show an error in case of sent timeout, instead of `setTimeout(() => route('/lightning'), 5000);`

frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -46,7 +46,138 @@ const SendingSpinner = () => {
...
+          {/* <Button primary onClick={parseInput} disabled={busy}>
+          {t('button.send')}
+        </Button> */}
+          <Button secondary onClick={onBack}>
+            {t('button.back')}
+          </Button>
+        </ViewButtons>
+      </View>
+    );
+  case 'confirm':
```
Beerosagos, line 144: Not sure if we should check for the available balance (and possibly block the send) here, or in the backend endpoint. wdyt?

frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -46,7 +46,138 @@ const SendingSpinner = () => {
...
+}: SendWorkflowProps) => {
+  const { t } = useTranslation();
+  switch (step) {
+  case 'select-invoice':
+    return (
+      <View fitContent>
+        <ViewContent textAlign="center">
+          <Grid col="1">
+            <Column>
+              {/* this flickers quickly, as there is 'SdkError: Generic: Breez SDK error: Unrecognized input type' when logging rawInputError */}
```
Beerosagos, line 124: @dangeross what does this comment mean?

frontends/web/src/routes/lightning/send/send.tsx
Beerosagos: Probably backend is better

Beerosagos (approved):
partially tested ack, with a couple nits

frontends/web/src/routes/lightning/send/send.tsx
dangeross: I think @thisconnect added this comment. The `rawInputError` is the result of calling `parseInput()` (called by onResult of the ScanQRVideo component) with an invalid input type. It's looking for BOLT11, LN-address, LNURL etc. Could be it's called with an empty string?

frontends/web/src/routes/lightning/send/send.tsx
thisconnect: This PR is only to split the components into smaller one, but I agree we should look into how this can be improved in the future.

frontends/web/src/routes/lightning/send/send.tsx
thisconnect: I think it's unrelated to this PR.

frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -46,7 +46,138 @@ const SendingSpinner = () => {
...
+              {rawInputError && <Status type="warning">{rawInputError}</Status>}
+              <ScanQRVideo onResult={onInput} />
+              {/* Note: unfortunatelly we probably can't read from HTML5 clipboard api directly in Qt/Andoird WebView */}
```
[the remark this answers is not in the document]
thisconnect: fixed

frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -46,7 +46,138 @@ const SendingSpinner = () => {
...
+              <Button transparent onClick={() => console.log('TODO: implement paste')}>
+                {t('lightning.send.rawInput.label')}
+              </Button>
+            </Column>
+          </Grid>
+        </ViewContent>
+        <ViewButtons reverseRow>
+          {/* <Button primary onClick={parseInput} disabled={busy}>
```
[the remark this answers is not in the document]
thisconnect: removed

frontends/web/src/routes/lightning/send/send.tsx
thisconnect: I have to check this again and would like to keep the comment for now.

Beerosagos (approved):
LGTM

frontends/web/src/routes/lightning/send/send.tsx
Beerosagos: Yeah, especially if we want to do it in the backend. I'll make a new issue

frontends/web/src/routes/lightning/send/send.tsx
Beerosagos: oh ok! Sorry @dangeross

PR #2558, opened 2024-02-15
frontend: move lightning as first keystore item in the sidebar

thisconnect:
This commit moves Lightning item in the sidebar to be the first item of each wallet (keystore).

Alternative would be before the wallets, after the 'My portfolio' item, this would be suitable if only one lightning wallet should be supported.


Beerosagos (approved):
tACK

PR #2560, opened 2024-02-15
backend/lightning: refactor lightning configuration

Beerosagos:
To ensure future compatibility in case we'll support multiple lightning accounts, we need to refactor the breez cache folder name and the lightning config struct, which is persisted as a json config file.

### DIFF
--- a/backend/config/lightning.go
+++ b/backend/config/lightning.go
@@ -15,17 +15,35 @@
 
 package config
 
+import (
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/types"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/jsonp"
+)
+
+// LightningAccountConfig is the configuration of a single Lightning account.
+type LightningAccountConfig struct {
+	// Mnemonic is the wallet node generated from the device entropy.
+	Mnemonic string `json:"mnemonic"`
+	// RootFingerprint is fingerprint of the keystore that generated the entropy.
+	RootFingerprint jsonp.HexBytes `json:"rootFingerprint"`
+	// Code is the code of the lightning account.
+	Code types.Code `json:"code"`
+	// Number is the lightning account incremental number.
+	Number uint16 `json:"number"`
+}
+
 // LightningConfig holds information related to the lightning config.
 type LightningConfig struct {
 	// Inactive is true if lightning has not yet been setup.
 	Inactive bool `json:"inactive"`
-	// Mnemonic is the wallet node generated from the device entropy.
-	Mnemonic string `json:"mnemonic"`
+	// Accounts is an array of existing lightning accounts configurations.
+	Accounts []*LightningAccountConfig `json:"accounts"`
 }
 
 // newDefaultAccountsConfig returns the default accounts config.
 func newDefaultLightningConfig() LightningConfig {
 	return LightningConfig{
 		Inactive: true,
+		Accounts: []*LightningAccountConfig{},
 	}
 }
--- a/backend/config/lightning.go
+++ b/backend/config/lightning.go
@@ -29,13 +29,11 @@ type LightningAccountConfig struct {
 	// Code is the code of the lightning account.
 	Code types.Code `json:"code"`
 	// Number is the lightning account incremental number.
-	Number uint16 `json:"number"`
+	Number uint16 `json:"num"`
 }
 
 // LightningConfig holds information related to the lightning config.
 type LightningConfig struct {
-	// Inactive is true if lightning has not yet been setup.
-	Inactive bool `json:"inactive"`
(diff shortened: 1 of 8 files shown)


backend/lightning/lightning.go
```diff
@@ -72,14 +75,24 @@ func (lightning *Lightning) Activate() error {
...
 	}
 
+	lightningAccount := config.LightningAccountConfig{
+		Mnemonic:        entropyMnemonic,
+		RootFingerprint: fingerprint,
+		Code:            types.Code(strings.Join([]string{"v0-", hex.EncodeToString(fingerprint), "-ln-0"}, "")),
+	}
 	lightningConfig.Inactive = false
-	lightningConfig.Mnemonic = entropyMnemonic
+	lightningConfig.Accounts = append(lightningConfig.Accounts, &lightningAccount)
```
benma, line 95: shouldn't we check if the account already exists?

backend/config/lightning.go
```diff
@@ -15,17 +15,33 @@
...
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/types"
+	"github.com/digitalbitbox/bitbox-wallet-app/util/jsonp"
+)
+
+// LightningAccountConfig is the configuration of a single Lightning account.
+type LightningAccountConfig struct {
+	// Mnemonic is the wallet node generated from the device entropy.
+	Mnemonic string `json:"mnemonic"`
+	// RootFingerprint is fingerprint of the keystore that generated the entropy.
+	RootFingerprint jsonp.HexBytes `json:"rootFingerprint"`
```
benma, line 28: the account number could be added, but then again it could also be added later if we actually have multiple

benma (dismissed):
utACK, but see comment

backend/config/lightning.go
Beerosagos: I was thinking about having it embedded in the code in the format `v0-<fingerprint>-ln-<account-number>`. Did you have something different in mind?

backend/lightning/lightning.go
Beerosagos: Activate should be called only once, when we enable lightning in the settings, and the config is cleaned up on `Deactivate()`, so it should not be possible to have any account already saved here.

benma: Sure, but I think it can't hurt to do a check anyway, because it's not easy to see that this property holds and will hold - plus sometimes there are bugs or race conditions etc.

Similar check exists here for normal accounts:

https://github.com/digitalbitbox/bitbox-wallet-app/blob/b02e97fd6492bf8163289e307c607ec6a03227f9/backend/accounts.go#L654-L657

backend/config/lightning.go
benma: i think the account code should never be parsed for information, it is simply an ID and should be treated like it's random. the account codes have already changed once (they were in fact random) in the past and they might change again, in which case worrying about all the places where the ID carries information would be hard.

anyway, maybe it's good to add the AccountNumber field (set to 0) already now to avoid the temptation of parsing the account code in the future :joy:

Beerosagos: Yeah that makes sense :) Done

backend/lightning/lightning.go
Beerosagos: Added a check to verify that the accounts list is empty on `Activate()`

Beerosagos:
@benma I think we have some rule on the branch that automatically dismisses a review when pushing a new update, PTAL :innocent:

backend/config/lightning.go
```diff
@@ -15,17 +15,35 @@
...
+	// Code is the code of the lightning account.
+	Code types.Code `json:"code"`
+	// Number is the lightning account incremental number.
+	Number uint16 `json:"number"`
+}
+
 // LightningConfig holds information related to the lightning config.
 type LightningConfig struct {
 	// Inactive is true if lightning has not yet been setup.
 	Inactive bool `json:"inactive"`
```
benma, line 38: is this flag still needed, now that the the code is setup for allowing more than one account?

backend/lightning/lightning.go
```diff
@@ -62,6 +65,10 @@ func (lightning *Lightning) Activate() error {
 		return errp.New("Lightning node already active")
 	}
 
+	if len(lightningConfig.Accounts) > 0 {
+		return errp.New("Lightning accounts already configured")
+	}
+
```
benma, line 71: i just saw now that the check above kind of does the same, but then i wonder if `lightnignConfig.Inactive` still makes sense after this commit? (see other comment)

benma (dismissed):

Beerosagos:
@benma rebased and added a second commit to remove the `Inactive` flag and change the frontend accordingly. PTAL

PR #2567, opened 2024-02-21
Fetch breez and greenlight credentials 

Beerosagos:
Based on #2566

@benma with this solution we'll fetch the credentials each time the app starts. Should we cache them for better privacy?

### DIFF
--- a/backend/lightning/lightning.go
+++ b/backend/lightning/lightning.go
@@ -17,6 +17,7 @@ package lightning
 
 import (
 	"encoding/hex"
+	"net/http"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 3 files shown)


backend/lightning/lightning.go
```diff
@@ -155,34 +165,57 @@ func (lightning *Lightning) connect() {
...
+
+		breezApiKey, err := fetchAuthData(lightning.httpClient, breezApiKeyUrl)
+		if err != nil {
+			lightning.log.WithError(err).Error("Breez api key fetch failed")
 			return
 		}
 
-		config := breez_sdk.DefaultConfig(breez_sdk.EnvironmentTypeProduction, "", nodeConfig)
+		// fetched key could have an unwanted newline, we'll just trim it for safety.
+		trimmedKey := strings.TrimSuffix(string(*breezApiKey), "\n")
```
benma, line 212: imo just call `strings.TrimSpace()`, it does not hurt and will prevent failure if some other whitespace is accidentally present

backend/lightning/lightning.go
```diff
@@ -203,3 +236,28 @@ func (lightning *Lightning) setLightningConfig(config config.LightningConfig) er
 
 	return nil
 }
+
+func fetchAuthData(httpClient *http.Client, endpoint string) (*[]uint8, error) {
```
benma, line 240: no reason to return a pointer, just return `[]uint8`. at call site you could add a sanity check that the result is not empty.

i feel like this function body (http request, limited size, StatusOK checkm ...) appears many times in the codebase. can we just have a helper function in `util/`?

benma (commented):
>  with this solution we'll fetch the credentials each time the app starts. 

Is it when the app starts? The code is in `connect()`, which I assume does not happen until the lightning wallet is initialized? the files are not fetched if someone didn't setup up the LN wallet right?

About caching: I wouldn't worry about it for now.

backend/lightning/lightning.go
```diff
@@ -155,34 +165,60 @@ func (lightning *Lightning) connect() {
...
 		if err := os.MkdirAll(workingDir, 0700); err != nil {
-			lightning.log.WithError(err).Warn("Error creating working directory")
+			lightning.log.WithError(err).Error("Error creating working directory")
 			return
 		}
 
-		config := breez_sdk.DefaultConfig(breez_sdk.EnvironmentTypeProduction, "", nodeConfig)
+		statusCode, breezApiKey, err := util.HTTPGet(lightning.httpClient, breezApiKeyUrl, "", int64(4096))
+		if err != nil {
+			lightning.log.WithError(err).Errorf("Breez api key fetch failed, errorCode: %v", statusCode)
```
benma, line 210: i think the status code is already in the error message of `err`, isn't it? no need to repeat it. same above

benma (approved):
utACK

PR #2568, opened 2024-02-21
Frontend qr scanner fullscreen III

thisconnect:


frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -30,6 +30,7 @@ interface Props {
     disabledClose?: boolean;
     children: React.ReactNode;
     open: boolean;
+    fullscreenOnMobile?: boolean;
```
thisconnect, line 33: note to myself: in last commit we try with View component, so if that goes well this can be undone

thisconnect:
Once ok I'll squash the last commits into the 2nd commit.

thisconnect:
Screen Shot 2024-03-05 at 08 53 13

and small screen:

Screen Shot 2024-03-05 at 08 53 35

thisconnect:
Screen Shot 2024-03-05 at 08 51 37

and small screen:

Screen Shot 2024-03-05 at 08 52 01

thisconnect:
unfortunatelly somehow it doesn't load the video in webdev :/

thisconnect:
@shonsirsha and @NicolaLS  I pushed 2 commits, with that it should also work in webdev (was only onchain send that didn't work).

PTAL

NicolaLS:
tested it on my Samsung S10e android 12 phone. looks very nice, could not find any issues :)

frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -121,10 +121,11 @@ const SendWorkflow = ({
   switch (step) {
   case 'select-invoice':
     return (
-      <View fitContent>
+      <View textCenter width="660px">
+        <ViewHeader title="Scan lightning invoice" />
```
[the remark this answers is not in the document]
thisconnect: not yet there are many strings that are not yet translated.
 like to get the english correct first.

frontends/web/src/components/view/view.module.css
```diff
@@ -75,7 +75,7 @@
     }
     .inner.fit {
         padding-bottom: var(--space-half);
-        padding-top: 2vh;
+        padding-top: 0; /* TODO: was 2vh test passphrase */
```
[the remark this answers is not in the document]
thisconnect: that impacts other views. and my next step is to test other views that use this.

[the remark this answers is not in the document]
thisconnect: ok tested a bit more, this is also used in CreateWalletSuccess and AOPP success, and this moves the content a bit to the top on mobile. But I noticed a bug both views have verticallyCentered prop, but the content it not vertically centered on mobile. 

I will keep this change herer and make a fix on master.

frontends/web/src/components/view/view.module.css
```diff
@@ -248,8 +248,11 @@
...
 }
 .dialog .buttons {
-    padding-left: var(--space-default);
-    padding-right: var(--space-default);
+    padding-left: var(--space-half);
+    padding-right: var(--space-half);
+}
+.fit .buttons {
+    margin-top: var(--space-half);
 }
```
thisconnect, line 256: So far this was only used in Alert component, but with this change we also have dialog + fitcontent and this introduced a small padding but in normal Alert.

Will push a fix.

frontends/web/src/routes/account/walletconnect/components/header/header.module.css
```diff
@@ -1,7 +1,8 @@
 .headerContainer {
+    align-items: center;
     display: flex;
     flex-direction: column;
-    align-items: center;
+    margin-bottom: var(--space-half);
```
thisconnect, line 5: this is only needed on WC on mobile, when scanning or on success view. I'll move this further down inside the media query, so it only applies on small screen.

frontends/web/src/routes/account/walletconnect/connect.module.css
```diff
@@ -19,6 +19,9 @@
 
 
 @media (max-width: 768px) { 
+    .contentContainer {
+        margin-top: var(--space-half);
+    }
```
thisconnect, line 24: oups this is a leftover, removing connect.module.css as this PR refactors connect.tsx to not use this CSS module.

thisconnect:
rebased and added commit to remove un-used CSS file that was used for scan video before.

PR #2570, opened 2024-02-21
backend/lightning: make ligthning SDK connection synchronous

Beerosagos:
The previous asynchronous connection made difficult to properly handle errors during the activation and also caused the redirect to the lightning account to happen before the sdk completed connecting. This was causing an error being shown for a few seconds right after the activation.

The activation flow may now take a few seconds to complete, but the frontend shows a nice loader so that should not be a big issue.

### DIFF
--- a/backend/lightning/lightning.go
+++ b/backend/lightning/lightning.go
@@ -108,14 +108,18 @@ func (lightning *Lightning) Activate() error {
 		return err
 	}
 
-	go lightning.connect(true)
+	if err = lightning.connect(true); err != nil {
+		return err
+	}
 
 	return nil
 }
 
 // Connect needs to be called before any requests are made.
 func (lightning *Lightning) Connect() {
-	go lightning.connect(false)
+	if err := lightning.connect(false); err != nil {
+		lightning.log.WithError(err).Warn("BreezSDK: Error connecting SDK")
+	}
 }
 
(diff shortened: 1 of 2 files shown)


backend/lightning/lightning.go
```diff
@@ -108,14 +108,18 @@ func (lightning *Lightning) Activate() error {
...
-	go lightning.connect(true)
+	if err = lightning.connect(true); err != nil {
+		return err
+	}
 
 	return nil
 }
 
 // Connect needs to be called before any requests are made.
 func (lightning *Lightning) Connect() {
```
benma, line 119: I see this is called on app launch in backend.Start(). Isn't this overkill as many will not have an active LN account?

benma (approved):
~~utACK~~ got a comment after all about doing async in backend.Start() anyway

backend/lightning/lightning.go
benma: Also, if it takes a while, this would block/delay the app launch, right?

benma (commented):
(see other comments)

backend/lightning/lightning.go
Beerosagos: It should be really fast, as the node has been already started. Anyway, here the `go ...` doesn't hurt. I'll do it :+1:

PR #2574, opened 2024-02-26
Lightning fiat

Beerosagos:
This is a first PR to start adding Fiat conversion support for lightning. 
Fiat conversions for LN send/receive will be addressed in a separate PR.

### DIFF
--- a/backend/lightning/handlers.go
+++ b/backend/lightning/handlers.go
@@ -20,6 +20,8 @@ import (
 	"net/http"
 
 	"github.com/breez/breez-sdk-go/breez_sdk"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
 )
 
 // PostLightningActivateNode handles the POST request to activate the lightning node.
@@ -42,7 +44,7 @@ func (lightning *Lightning) PostLightningDeactivateNode(r *http.Request) interfa
 	return responseDto{Success: true}
 }
 
-// GetNodeInfo handles the GET request retrieve the node info.
+// GetNodeInfo handles the GET request to retrieve the node info.
 func (lightning *Lightning) GetNodeInfo(_ *http.Request) interface{} {
 	if lightning.sdkService == nil {
 		return responseDto{Success: false, ErrorMessage: "BreezServices not initialized"}
@@ -54,6 +56,32 @@ func (lightning *Lightning) GetNodeInfo(_ *http.Request) interface{} {
 	}
 
 	return responseDto{Success: true, Data: toNodeStateDto(nodeState)}
+}
+
+// GetBalance handles the GET request to retrieve the node balance and its fiat conversions.
+func (lightning *Lightning) GetBalance(_ *http.Request) interface{} {
+	balance, err := lightning.Balance()
+	if err != nil {
+		return responseDto{Success: false, ErrorMessage: err.Error()}
+	}
+
+	formatAsSats := lightning.backendConfig.AppConfig().Backend.BtcUnit == coin.BtcUnitSats
+	btcCoin := lightning.btcCoin
+
+	formattedAvailableAmount := coin.FormattedAmount{
+		Amount:      btcCoin.FormatAmount(balance.Available(), false),
+		Unit:        btcCoin.GetFormatUnit(false),
+		Conversions: coin.Conversions(balance.Available(), btcCoin, false, lightning.ratesUpdater, formatAsSats),
+	}
+
+	return responseDto{
+		Success: true,
+		Data: accounts.FormattedAccountBalance{
+			HasAvailable: balance.Available().BigInt().Sign() > 0,
+			Available:    formattedAvailableAmount,
+			HasIncoming:  false,
+			Incoming:     coin.FormattedAmount{},
+		}}
 }
 
 // GetListPayments handles the GET request to list payments.
--- a/backend/lightning/lightning.go
+++ b/backend/lightning/lightning.go
@@ -23,9 +23,12 @@ import (
 	"strings"
 
 	"github.com/breez/breez-sdk-go/breez_sdk"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/accounts/types"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/coins/coin"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/config"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/keystore"
+	"github.com/digitalbitbox/bitbox-wallet-app/backend/rates"
 	"github.com/digitalbitbox/bitbox-wallet-app/backend/util"
 	"github.com/digitalbitbox/bitbox-wallet-app/util/errp"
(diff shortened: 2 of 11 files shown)


frontends/web/src/api/lightning.ts
```diff
@@ -282,17 +282,17 @@ export type MetadataItem = {
...
-  connectedPeers: string[];
-  inboundLiquidityMsats: number;
+  // onchainBalanceMsat: number;
+  // utxos: UnspentTransactionOutput[];
+  // maxPayableMsat: number;
+  // maxReceivableMsat: number;
+  // maxSinglePaymentAmountMsat: number;
+  // maxChanReserveMsats: number;
+  // connectedPeers: string[];
+  // inboundLiquidityMsats: number;
```
Beerosagos, line 295: All this fields are not currently used in the FE. I guess it will be useful to get all the node info in the future, but atm this is dead code. We should review the whole FE API and delete the unnecessary code.

Beerosagos:
Requested reviews from both @benma and @thisconnect for backend/frontend :innocent: :pray: thanks a lot

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -217,7 +216,7 @@ export function AccountsSummary({ accounts, devices }: TProps) {
                   totalBalancePerCoin={ balancePerCoin ? balancePerCoin[keystore.rootFingerprint] : undefined}
                   totalBalance={ accountsTotalBalance ? accountsTotalBalance[keystore.rootFingerprint] : undefined}
                   balances={balances}
-                  lightningNodeState={(lightningConfig.accounts.length && lightningConfig.accounts[0].rootFingerprint === keystore.rootFingerprint) ? nodeState : undefined}
+                  lightningBalance={(lightningConfig.accounts.length && lightningConfig.accounts[0].rootFingerprint === keystore.rootFingerprint) ? lightningBalance : undefined}
```
thisconnect, line 219: This component doesn't really need `lightningConfig` and `lightningBalance` as those are only used in SummaryBalance component, wdyt about moving as much s possible into SummaryBalance?

you'll probably have to pass `keystore.rootFingerprint` as well somehow for the condition.

frontends/web/src/api/lightning.ts
thisconnect: IMO the type should always contain all data that is returned by the backend, even if we don't use it.

There seems to be useful information in here that we should consider showing somehow. i.e. maxPayableMsat or maxReceivableMsat, inboundLiquidityMsats.

Could you keep as is for now and remove/refactor later?

thisconnect (commented):
tested a bit, looks very nice. added some comments.

backend/lightning/lightning.go
```diff
@@ -154,13 +167,51 @@ func (lightning *Lightning) Deactivate() error {
 	return nil
 }
 
+// AssertActive returns an error if the lightning service not has been activated.
+func (lightning *Lightning) AssertActive() error {
```
benma, line 171: assert is the wrong word, as assert in all contexts I know (in Python, C, ...), it crashes/panics/exists.

`CheckActive` or `ValidateActive` would be better.

backend/lightning/lightning.go
```diff
@@ -154,13 +167,51 @@ func (lightning *Lightning) Deactivate() error {
...
+	lightningConfig := lightning.backendConfig.LightningConfig()
+	if len(lightningConfig.Accounts) == 0 || lightning.sdkService == nil {
+		return errp.New("Lightning not initialized")
+	}
+	return nil
+}
+
+// Balance returns the balance of the lightning account, together with its conversions to fiat.
+func (lightning *Lightning) Balance() (interface{}, error) {
```
benma, line 180: don't return `interface{}` but a concrete type

backend/lightning/lightning.go
```diff
@@ -154,13 +167,51 @@ func (lightning *Lightning) Deactivate() error {
...
+	if err := lightning.AssertActive(); err != nil {
+		return coin.FormattedAmount{}, err
```
benma, line 182: return nil, err

backend/lightning/lightning.go
```diff
@@ -154,13 +167,51 @@ func (lightning *Lightning) Deactivate() error {
...
+	}
+
+	nodeInfo, err := lightning.sdkService.NodeInfo()
+	if err != nil {
+		return coin.FormattedAmount{}, err
```
benma, line 187: return nil, err

backend/lightning/lightning.go
```diff
@@ -154,13 +167,51 @@ func (lightning *Lightning) Deactivate() error {
...
+	}
+
+	amount := coin.NewAmount(big.NewInt(int64(nodeInfo.ChannelsBalanceMsat / 1000)))
```
benma, line 190: there is `coin.NewAmountFromInt64()`

backend/lightning/lightning.go
benma: this function should really return `*accounts.Balance`, like btc/account.go and eth/account.go, and the handler should convert to `AccountBalance`.

Reason: AccountBalance is frontend focused, and it's the handler's job to convert input/output from/to the frontend.

backend/accounts/account.go
```diff
@@ -32,6 +32,14 @@ type AddressList struct {
 	Addresses  []Address
 }
 
+// AccountBalance includes the total and incoming balance of an account.
+type AccountBalance struct {
```
benma, line 36: there is already a `Balance` struct (in balance.go). making this ultra confusing.

Maybe move this to balance.go too and name it `FormattedAccountBalance`.

benma (changes requested):

backend/lightning/handlers.go
```diff
@@ -56,6 +56,15 @@ func (lightning *Lightning) GetNodeInfo(_ *http.Request) interface{} {
 	return responseDto{Success: true, Data: toNodeStateDto(nodeState)}
 }
 
+// GetBalance handles the GET request to retrieve the node balance and its fiat conversions.
+func (lightning *Lightning) GetBalance(_ *http.Request) interface{} {
+	balance, err := lightning.Balance()
+	if err != nil {
+		return responseDto{Success: false, ErrorMessage: err.Error()}
```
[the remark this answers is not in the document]
Beerosagos: It's Data Transfer Object. It's quite used as an abbreviation, afaik.

frontends/web/src/routes/account/summary/accountssummary.tsx
Beerosagos: yeah, I moved the LN logic inside the component and refactored a bit to only pass it the keystore struct, instead of various fields belonging to it.

backend/lightning/lightning.go
Beerosagos: Agree :+1: done.

backend/lightning/lightning.go
Beerosagos: Nice! thanks

Beerosagos:
Rebased and addressed comments. @benma @thisconnect PTAL :pray:

backend/lightning/lightning.go
benma: Only now I see that the handlers are methods on the same struct :O but the split was still good, as we will hopefully refactor this later and make a clean split :smile:

benma (approved):
Thanks, utACK

thisconnect (approved):
tested LGTM 👍

PR #2576, opened 2024-02-26
frontend: fix transaction list for lightning

shonsirsha:
following our already existing transaction list layout, on mobile, the amount collapses (flex column) to the second row. This keeps the amount visible if it's long or when the description / note is long.

Also removes unused CSS classes.

_screen size: 350px_
[Preview], before:
_(long note)_
992S

ODAS

[Preview], after: 
asx

dxa

### DIFF
--- a/frontends/web/src/routes/lightning/components/payment.module.css
+++ b/frontends/web/src/routes/lightning/components/payment.module.css
@@ -12,18 +12,9 @@
     font-weight: 400;
 }
 
-.date {
-
-}
-
 .description {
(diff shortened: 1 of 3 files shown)


thisconnect (approved):
nice I tested a bit. Mobile looks good to me. 

there is some funny spacings sometimes on medium size screen.

IMO feel free to merge as is and we do the medium screen optimization later or you can do also in this PR up to you.

Screen Shot 2024-02-26 at 13 42 47

PR #2579, opened 2024-02-26
Block explorer selection

NicolaLS:
Allow users to configure their preferred block explorer (from predefined selection) in the settings.

Block explorers need to be added to the `AvailableExplorers` struct and whitelisted in the backend. The frontend learns all available options by calling the `/available-explorers` endpoint. Adding options will only need a change in the backend the frontend will always show all available selections.

notes:
- I chose to keep it simply by statically defining the `AvailableExplorers` of course we could also implement an interface instead that each coin can implement, this would be more elegant but I did not see the need for that.
- I really don't like the implementation of the disable/enable "save" button (compare the json configs) see here. I don't know how to do it differently though..
- I had to fetch the `initialConfig` in a different place than the `config` because it would always update alongside config if they were initialized in the same place `loadConfig` see here. I don't understand why it behaves like this, I just fetch the config again to initialize `initialConfig` maybe there is a better way avoiding two get requests...

Right now the frontend uses the `/supported-coins` endpoint to decide how many/which selection to show in the setting. This requires a _keystore_ to be connected which is weird UX. We can fix this later by introducing a separate check for a multicoin, or bitcoin only user/wallet in the config.

Please let me know if you have suggestions for more block explorers, I just added the most popular. In case they deviate from `host.com/tx/xxx` (e.g. `host.com?tx=xx`) there needs to be more refactoring in the backend, maybe then it would make sense to define an interface rather than a static stuct for the `AvailableExplorers`.


frontends/web/src/routes/account/account.tsx
```diff
@@ -123,8 +124,17 @@ export function Account({
...
   useEffect(() => {
     maybeCheckBitsuranceStatus();
-    getConfig().then(({ backend }) => setUsesProxy(backend.proxy.useProxy));
-  }, [maybeCheckBitsuranceStatus]);
+    getConfig().then(({ backend }) => {
+      setUsesProxy(backend.proxy.useProxy);
+      if (account) {
+        if (backend[account.coinCode]) {
+          setBlockExplorerTxPrefix(backend[account.coinCode].blockExplorerTxPrefix);
+        } else {
```
NicolaLS, line 132: this is redundant as we check `blockExplorerTxPrefix` undefined on line 337. Should I remove the else branch ?

backend/backend.go
```diff
@@ -73,13 +73,21 @@ var fixedURLWhitelist = []string{
...
+	// Block explorers (ltc).
 	"https://sochain.com/tx/LTCTEST/",
 	"https://blockchair.com/litecoin/transaction/",
+	// Block explorers (eth).
 	"https://etherscan.io/tx/",
 	"https://goerli.etherscan.io/tx/",
+	"https://sepolia.etherscan.io/tx/",
+	"https://ethplorer.io/tx/",
+	"https://goerli.ethplorer.io/tx/",
+	"https://sepolia.ethplorer.io/tx/",
```
benma, line 90: instead of adding them all here, it's better to add another loop in SystemOpen() that goes through all explorers, so that one only has to modify the list of explorers when making changes (adding or removing explorers).

backend/config/blockexplorer.go
```diff
@@ -0,0 +1,88 @@
+package config
+
+// BlockExplorer defines a selectable block explorer.
+type BlockExplorer struct {
+	Name string `json:"name"`
+	Url  string `json:"url"`
```
benma, line 6: document the fields as well. Especially that the Url is a prefix and the txID will be appended.

backend/config/config.go
```diff
@@ -59,6 +60,7 @@ const (
 // ethCoinConfig holds configurations for ethereum coins.
 type ethCoinConfig struct {
 	DeprecatedActiveERC20Tokens []string `json:"activeERC20Tokens"`
+	BlockExplorerTxPrefix       string   `json:"blockExplorerTxPrefix"`
```
benma, line 63: i wonder if it's maybe better to add a separate field to the backend config `blockExplorers`, so it's all concentrated in one location :thinking:

frontends/web/src/locales/en/app.json
```diff
@@ -1013,6 +1013,27 @@
         "title": "Why is there a network fee?"
       }
     },
+    "settings-block-explorer": {
```
benma, line 1016: settingsBlockExplorer, all entries should be camelCase (most are, but there are some misses like right below with settings-electrum :see_no_evil: )

frontends/web/src/locales/en/app.json
```diff
@@ -1013,6 +1013,27 @@
...
+      "instructions": {
+        "link": {
+          "text": "Guide on how to use a block explorer"
+        },
+        "text": "For a full tutorial, please visit our guide:",
+        "title": "How do I use a block explorer?"
+      },
+      "options": {
+        "text": "The BitBoxApp features a protection mechanism designed to prevent the opening of arbitrary links. This serves as a security measure against malicious links. It cannot be bypassed on an individual basis; rather, its suspension would be required to allow users to configure their own URLs.",
```
benma, line 1025: imo this is super confusing to users, especially the last part. i'd stop at the first part

"The BitBoxApp features a protection mechanism designed to prevent the opening of arbitrary links. This serves as a security measure against malicious links."

frontends/web/src/routes/settings/select-explorer.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const save = async () => {
+    setSaveDisabled(true);
+    await setConfig(config);
+    initialConfig.current = await getConfig();
+  };
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const coins = await backendAPI.getSupportedCoins();
+      const allExplorerSelection = await apiGet('available-explorers');
```
benma, line 84: make an API call wrapper in api/backend.ts, and a definition of the return type.

benma (commented):
thanks! 

partial review, couldn't finish yet.

backend/config/config.go
NicolaLS: that makes sense.

backend/config/config.go
```diff
@@ -76,12 +86,16 @@ type Backend struct {
...
-	LTC  btcCoinConfig `json:"ltc"`
-	TLTC btcCoinConfig `json:"tltc"`
-	ETH  ethCoinConfig `json:"eth"`
+	BTC    btcCoinConfig `json:"btc"`
+	TBTC   btcCoinConfig `json:"tbtc"`
+	RBTC   btcCoinConfig `json:"rbtc"`
+	LTC    btcCoinConfig `json:"ltc"`
+	TLTC   btcCoinConfig `json:"tltc"`
+	ETH    ethCoinConfig `json:"eth"`
+	GOETH  ethCoinConfig `json:"goeth"`
```
NicolaLS, line 95: I guess I can remove `GOETH` and `SEPETH` again, as I only had to add them for the nested explorer url which is now in `BlockExplorers`

NicolaLS:
addressed your comments @benma , please resolve the threads if satisfied. also noticed I can remove `GOETH` and `SEPETH` from the config again, will wait for more feedback and do it in the next push.

NicolaLS:
I added another commit that uses accounts instead of supported-coins, but now there will not be any changes/additions anymore.

NicolaLS:
force pushed rebase for clean diff

NicolaLS:
Please let me know if I should add something to the `CHANGELOG.md`. Everything should be ready, the change requests are addressed but I leave it to @benma to resolve them. @thisconnect could you review the frontend?

PR #2580, opened 2024-02-27
frontend/lightning: add fiat conversions to send/receive flows

Beerosagos:
This also adds a getBTCSatsAmount backend endpoint that allows to convert a sat amount into an Amount object (FE: IAmount, BE: FormattedAmount) with the fiat conversions. This is not the best solution, as it would be much better to heavily refactor the invoices parsing, removing all the FE types that depends on breez and making the enpoints return a simpler struct containing only the used data (including the fiat conversions).

This is a workaround that allows to show fiat conversions in the frontend while we take our time to decide how to refactor this part.

### DIFF
--- a/frontends/web/src/routes/lightning/receive/receive.tsx
+++ b/frontends/web/src/routes/lightning/receive/receive.tsx
@@ -40,13 +40,16 @@ import { unsubscribe } from '../../../utils/subscriptions';
 import { Spinner } from '../../../components/spinner/Spinner';
 import { Checked, Copy, EditActive } from '../../../components/icon';
 import styles from './receive.module.css';
+import { FiatConversion } from '../../../components/rates/rates';
+import { getBtcSatsAmount } from '../../../api/coins';
+import { IAmount } from '../../../api/account';
 
 type TStep = 'select-amount' | 'wait' | 'invoice' | 'success';
 
 export function Receive() {
   const { t } = useTranslation();
-  const [amountSats, setAmountSats] = useState<number>(0);
-  const [amountSatsText, setAmountSatsText] = useState<string>('');
+  const [inputSatsText, setInputSatsText] = useState<string>('');
+  const [invoiceAmount, setInvoiceAmount] = useState<IAmount>();
   const [description, setDescription] = useState<string>('');
   const [openChannelFeeResponse, setOpenChannelFeeResponse] = useState<OpenChannelFeeResponse>();
   const [receivePaymentResponse, setReceivePaymentResponse] = useState<ReceivePaymentResponse>();
@@ -65,7 +68,7 @@ export function Receive() {
       setStep('select-amount');
       setReceiveError(undefined);
       if (step === 'success') {
-        setAmountSatsText('');
+        setInputSatsText('');
       }
       break;
     }
@@ -73,7 +76,7 @@ export function Receive() {
 
   const onAmountSatsChange = (event: ChangeEvent<HTMLInputElement>) => {
     const target = event.target as HTMLInputElement;
-    setAmountSatsText(target.value);
+    setInputSatsText(target.value);
   };
 
   const onDescriptionChange = (event: ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/lightning/receive/receive.tsx
```diff
@@ -91,18 +95,28 @@ export function Receive() {
   }, [onPaymentsChange]);
 
   useEffect(() => {
-    setAmountSats(+amountSatsText);
-  }, [amountSatsText]);
+    setInputSats(+inputSatsText);
```
thisconnect, line 98: please use `Number()` instead.. or maybe better change to inputSatsNumber using number and `target.valueAsNumber` since it is an input with type number.

frontends/web/src/routes/lightning/receive/receive.tsx
```diff
@@ -40,13 +40,17 @@ import { unsubscribe } from '../../../utils/subscriptions';
...
+import { FiatConversion } from '../../../components/rates/rates';
+import { getBtcSatsAmount } from '../../../api/coins';
+import { IAmount } from '../../../api/account';
 
 type TStep = 'select-amount' | 'wait' | 'invoice' | 'success';
 
 export function Receive() {
   const { t } = useTranslation();
-  const [amountSats, setAmountSats] = useState<number>(0);
-  const [amountSatsText, setAmountSatsText] = useState<string>('');
```
thisconnect, line 49: In send I refactored this, as we don't need 2 states holding the same data once a number and once as string...

We should refactor this at some point here as well.

frontends/web/src/routes/lightning/receive/receive.tsx
thisconnect: I left this comment and then realized that the state in this component should be simplified.. 
feel free to ignore.

frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -79,7 +77,7 @@ const InvoiceInput = ({ invoice }: InvoiceInputProps) => {
         </div>
       )}
     </>
-  );
+  ) : <Skeleton />);
```
thisconnect, line 80: In general the Skeleton component should not replace the whole content, but only the parts that can't be rendered.

Ideally only Amount and FiatConversion are replaced with a Skeleton, but all the other div's etc. can already be shown.

frontends/web/src/routes/lightning/receive/receive.tsx
thisconnect: feel free to ignore expect for the type conversion, nit: please use `Number()`

benma (approved):
utACK, but please see comment about `number` for sats

frontends/web/src/routes/lightning/receive/receive.tsx
benma: why does `inputSats` have to be a state var? seems like it's derived from `inputSatsText` only, so one could always just use `inputSatsText`? I am probably missing something.

thisconnect: no you don't miss anything. I did refactor this in send a while ago, we should refactor this in receive as well.

Beerosagos: Refactored keeping the text version as state and converting with `Number()` when needed :+1:

frontends/web/src/routes/lightning/receive/receive.tsx
Beerosagos: Done

frontends/web/src/routes/lightning/send/send.tsx
Beerosagos: Done

frontends/web/src/routes/lightning/receive/receive.tsx
```diff
@@ -40,13 +40,17 @@ import { unsubscribe } from '../../../utils/subscriptions';
...
+  const [inputSats, setInputSats] = useState<number>(0);
```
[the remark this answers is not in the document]
Beerosagos: Done :+1:

benma (approved):
utACK again, thx!

PR #2581, opened 2024-02-27
backend/lightning: check available balance before sending LN payment

Beerosagos:
This adds a backend check on the available balance, to avoid sending a LN payment wihtout sufficient funds.

### DIFF
--- a/backend/lightning/handlers.go
+++ b/backend/lightning/handlers.go
@@ -176,6 +176,29 @@ func (lightning *Lightning) PostSendPayment(r *http.Request) interface{} {
 		return responseDto{Success: false, ErrorMessage: err.Error()}
 	}
 
+	invoice, err := breez_sdk.ParseInvoice(jsonBody.Bolt11)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


backend/lightning/handlers.go
```diff
@@ -176,6 +176,24 @@ func (lightning *Lightning) PostSendPayment(r *http.Request) interface{} {
...
+	nodeState, err := lightning.sdkService.NodeInfo()
+	if err != nil {
+		return responseDto{Success: false, ErrorMessage: err.Error()}
+	}
+
+	amount := invoice.AmountMsat
+	if jsonBody.AmountMsat != nil {
+		amount = jsonBody.AmountMsat
+	}
+	if *amount > nodeState.ChannelsBalanceMsat {
```
benma, line 193: can amount be nil? maybe best to check to avoid a panic

benma (approved):
utACK

backend/lightning/handlers.go
Beerosagos: Should not be possible, but added just in case

PR #2583, opened 2024-02-27
android: add lnalpha build target

benma:
So an alpha release of the BitBoxApp for Android does not interfere with existing installations of the BitBoxApp.


Beerosagos (approved):
utACK

PR #2591, opened 2024-02-28
frontend/lightning: fix invoice copy button

Beerosagos:
LN invoice copy button wasn't working on Qt, this fixes it.

co-authored: thisconnect


thisconnect (approved):
LGTM

PR #2598, opened 2024-02-29
bitbox02/bootloader: log monotonic version

benma:
Sometimes users get 'invalid firmware' and go into the bootloader screen without the button to install/upgrade `CanUpgrade=false`. This should never happen if the user is using the latest app release. This may help debug the problem should it appear again.


Beerosagos (approved):
utACK

PR #2599, opened 2024-02-29
qt: log rendering mode

benma:
If a user manually has to set this, it would be helpul to see this msg
to know that the right var was set.


Beerosagos (approved):
utACK!

Maybe it could be useful to log the value in case it is invalid. wdyt?

benma:
I thought about it but it's C++, was too annoying to assemble the string :sweat_smile: 

I want to make this a command line argument anyway, as setting the env var is horrible from a UX point of view (CLI flag is also not great, but a lot better)

PR #2600, opened 2024-02-29
qt: delete custom libserver.h, use auto-generated one

benma:
Not sure why we copied the contents of libserver.h initially, maybe we needed to modify the declarations a bit. Anyway, it seems to work by directly using the generated file.

This makes sure the changes to CGO in upgraded versions (the preamble) are adopted automaticaly, and adding/changing functions in server.go works automatically.


frontends/qt/BitBox.pro
```diff
@@ -56,7 +56,8 @@ SOURCES += \
         main.cpp \
         filedialog.cpp
 
-HEADERS += libserver.h webclass.h filedialog.h
+# server/libserver.h is generated by building server.go, see the Makefiles in server/.
+HEADERS += server/libserver.h webclass.h filedialog.h
```
benma, line 60: @Beerosagos the file is not committed - not sure if we should commit it or not :thinking: if committed, upgrading Go might change the file and we'd need to make a PR to update it.

Beerosagos: Yeah I agree, I don't see the advantage of committing it

PR #2601, opened 2024-03-01
frontend/conversions: show BTC as active currency in sats mode

strmci:
In BTC accounts with sats mode enabled, displaying two amounts in sats is redundant. This commit addresses this by ensuring that when sats mode is enabled for BTC accounts, the active currency is displayed as BTC. BTC is now displayed in this setting as the active currency on the send page and during the transaction review process.


strmci:
superseded by https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2682, closing.

PR #2603, opened 2024-03-04
frontend/lightning: fix bug in handling receive invoice error

Beerosagos:
The LN receive workflow was failing to show errors returned by the backend because the loading screen wasn't hidden in case of errors.

### DIFF
--- a/frontends/web/src/routes/lightning/receive/receive.tsx
+++ b/frontends/web/src/routes/lightning/receive/receive.tsx
@@ -142,6 +142,7 @@ export function Receive() {
       setReceivePaymentResponse(receivePaymentResponse);
       setStep('invoice');
     } catch (e) {
+      setStep('select-amount');
       if (e instanceof SdkError) {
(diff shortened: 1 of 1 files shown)


thisconnect (approved):
untested LGTM

Beerosagos:
Actually, I realized that it is better to set the step right after the catch, otherwise the wait screen will hang in case of other errors :grimacing: PTAL :pray:

thisconnect:
@Beerosagos what is the best way to test this?

Beerosagos:
> @Beerosagos what is the best way to test this?

`<Button primary onClick={receivePayment} disabled={disableConfirm && false}>` at receive.tsx:188

And then try to receive on a new lightning wallet with a low amount (less than open channel fee)

thisconnect (approved):
thanks tested 3165a08de4397e8fa16bb092d3e309c8dae60868 LGTM 👍

PR #2605, opened 2024-03-05
backend: update coin control outputs sorting

strmci:
Currently, the list of all UTXOs in the coin control selection is ordered by amount. Users sometimes re-use addresses and are not aware that there might be other UTXOs they didn't want or want to use on the same address as their selected output. Grouping the UTXOs by address helps the user to detect if there are other UTXOs.

This commit updates the sorting of outputs in the coin control dialog and is now grouped by address from the biggest to the lowest amount.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -44,7 +44,6 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/chaincfg"
-	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


strmci:
@thisconnect thanks, I updated the commit message, PTAL

thisconnect:
Before this change the sorting was outputs by amounts. Now with this change it is outputs grouped by address then sorted by amounts.

For users that didn't re-use addresses nothing should change.

For users that did reuse address the list should change so that all UTXO's of the same address are shown together.

I think it would be great if there is a banner on top that shows "Warning: some addresses were re-used" or similar (banner of shame).

backend/coins/btc/account.go
```diff
@@ -773,19 +772,55 @@ func (account *Account) CanVerifyAddresses() (bool, bool, error) {
...
+		sum     int64
+	}, 0, len(grouped))
+
+	// Calculate sums of values for each group and store in the sums slice
+	for address, outputs := range grouped {
+		var sum int64
+		for _, output := range outputs {
+			sum += output.TxOut.Value
+		}
+		sums = append(sums, struct {
```
Beerosagos, line 798: nit: I don't like it much defining the struct twice. Maybe it is better to define it once with a proper private name outside the function.

Beerosagos (approved):
LGTM apart for a small nit. I agree that we definitely need a proper warning banner in case of address reuse, and maybe also highlighting somehow the reused addresses (e.g. bold font, or smth else).

backend/coins/btc/account.go
strmci: I added a new type `addressOutputsSum` and used it in the function.

strmci:
>  I agree that we definitely need a proper warning banner in case of address reuse, and maybe also highlighting somehow the reused addresses (e.g. bold font, or smth else).

Would you like to add it here, or should we do it in a separate PR?

thisconnect:
I think separate PR makes sense or at least separate commit. For example in the case we need to undo one of the 2 changes.

strmci:
PR to add address reuse warning: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2724

PR #2607, opened 2024-03-06
backend/handlers: fix bug in etherscan.ERC20GasErr handling

Beerosagos:
The error string was not correctly checked in handlers.go, which lead the frontend to show a generic error instead of the specific one.


benma (approved):
utACK

PR #2608, opened 2024-03-06
backend: add deletion of breez cache folder

strmci:
The breez cache folder is no longer needed after disabling the lightning wallet and may consume unnecessary disk space.


Beerosagos (approved):
tACK :+1:

PR #2614, opened 2024-03-08
frontend: update redirects when lightning wallet active

strmci:
When the lightning wallet is enabled and the device disconnects (with no watch-only accounts present), the application should redirect to My Portfolio and not to "Please connect your device" as the lightning wallet is active.

This PR also updates the initial state of lightningConfig within the LightningProvider component to use an empty array for the `accounts` property.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -43,6 +43,7 @@ import { Darkmode } from './components/darkmode/darkmode';
 import { AuthRequired } from './components/auth/authrequired';
 import { WCSigningRequest } from './components/wallet-connect/incoming-signing-request';
 import { Providers } from './contexts/providers';
+import { getLightningConfig, subscribeLightningConfig } from './api/lightning';
 
 export const App = () => {
   const { t } = useTranslation();
@@ -52,7 +53,7 @@ export const App = () => {
   const devices = useDefault(useSync(getDeviceList, syncDeviceList), {});
 
   const prevDevices = usePrevious(devices);
-
+  const lightningConfig = useDefault(useSync(getLightningConfig, subscribeLightningConfig), { accounts: [] },);
   useEffect(() => {
     return syncNewTxs((meta) => {
       notifyUser(t('notification.newTxs', {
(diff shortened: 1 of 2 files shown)


Beerosagos (changes requested):
Clicking on the BitBoxApp title on top of the sidebar brings you to the account summary, which doesn't load correctly. I guess we should fix the account summary (harder, but would be nice), or at least avoid redirecting there.

image

frontends/web/src/app.tsx
```diff
@@ -52,7 +53,7 @@ export const App = () => {
   const devices = useDefault(useSync(getDeviceList, syncDeviceList), {});
 
   const prevDevices = usePrevious(devices);
-
+  const lightningConfig = useDefault(useSync(getLightningConfig, subscribeLightningConfig), { accounts: {} as TLightningAccountConfig[] },);
```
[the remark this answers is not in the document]
strmci: I updated it to use an array in both files.

frontends/web/src/app.tsx
```diff
@@ -69,12 +70,12 @@ export const App = () => {
 
     // QT and Android start their apps in '/index.html' and '/android_asset/web/index.html' respectively
     // This re-routes them to '/' so we have a simpler uri structure
-    if (isIndex && currentURL !== '/' && (!accounts || accounts.length === 0)) {
+    if (isIndex && currentURL !== '/' && (!accounts || accounts.length === 0) && lightningConfig.accounts.length === 0) {
```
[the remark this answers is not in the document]
strmci: I created it to check both mainnet and lightning accounts.

strmci:
@Beerosagos updated and rebased to the newest SDK, PTAL
as we discussed I will fix the My portfolio issue in my other PRs

frontends/web/src/app.tsx
```diff
@@ -94,7 +96,7 @@ export const App = () => {
       return;
     }
     // if on index page and have at least 1 account, route to /account-summary
-    if (isIndex && accounts.length) {
+    if (isIndex && !noAccounts) {
```
Beerosagos, line 99: Also the next if conditions can exploit `noAccounts`

Beerosagos: I mean
```
    if (accounts.length === 0 && currentURL.startsWith('/buy/')) {
      navigate('/');
      return;
    }
    // if on the /bitsurance/ view and there are no accounts view route to /
    if (accounts.length === 0 && currentURL.startsWith('/bitsurance/')) {
      navigate('/');
      return;
    }
```

frontends/web/src/app.tsx
```diff
@@ -66,15 +67,16 @@ export const App = () => {
...
     // QT and Android start their apps in '/index.html' and '/android_asset/web/index.html' respectively
     // This re-routes them to '/' so we have a simpler uri structure
-    if (isIndex && currentURL !== '/' && (!accounts || accounts.length === 0)) {
+    if (isIndex && currentURL !== '/' && noAccounts) {
       navigate('/');
       return;
     }
     // if no accounts are registered on specified views route to /
-    if (accounts.length === 0 && (
+    if (noAccounts && (
```
Beerosagos, line 79: 'add-account' and 'manage-accounts' pages should not depend on lightning accounts. I think that we need two different `noAccounts` conditions.

frontends/web/src/app.tsx
strmci: I didn't add it there as these conditions (bitsurance and buy) are related to mainnet accounts.

frontends/web/src/app.tsx
thisconnect: please consider doing some of this refactoring if possible  on master.

strmci:
@Beerosagos as we discussed, I only created `noLightningAccounts` and haven't changed the conditions based on `accounts.length` for now.

Beerosagos (approved):
partially tested ACK

PR #2615, opened 2024-03-08
Update deps

benma:


go.mod
```diff
@@ -19,8 +19,8 @@ require (
 	github.com/skip2/go-qrcode v0.0.0-20170812161211-695fc75a0973
 	github.com/stretchr/testify v1.9.0
 	go.etcd.io/bbolt v1.3.9
-	golang.org/x/crypto v0.14.0
-	golang.org/x/net v0.17.0
+	golang.org/x/crypto v0.21.0
+	golang.org/x/net v0.21.0
```
Beerosagos, line 23: Should we add x/net in the commit description too? Or was it updated automatically?

Beerosagos (approved):
utACK with a small question :)

go.mod
benma: Damn merged before seeing this :see_no_evil: It was somehow done automatically by updating crypto :thinking:

Beerosagos: Not a big deal, anyway :)

PR #2623, opened 2024-03-11
backend: add lightning balance to total balance

strmci:
This commit fixes calculation of totals in my portfolio when the lightning wallet is active.

### DIFF
--- a/backend/accounts_test.go
+++ b/backend/accounts_test.go
@@ -17,6 +17,7 @@ package backend
 import (
 	"encoding/hex"
 	"fmt"
+	"math/big"
 	"net/http"
 	"testing"
 	"time"
@@ -1505,4 +1506,20 @@ func TestAccountsTotalBalanceByKeystore(t *testing.T) {
 
 	require.NotNil(t, totalBalance[hex.EncodeToString(ks2Fingerprint)])
 	require.Equal(t, "0.13", totalBalance[hex.EncodeToString(ks2Fingerprint)].Total)
+}
+
+func TestConvertBtcAmountToFiat(t *testing.T) {
+	b := newBackend(t, testnetDisabled, regtestDisabled)
+	defer b.Close()
+	b.ratesUpdater = rates.MockRateUpdater()
+	defer b.ratesUpdater.Stop()
+
+	// convert 1 BTC to USD
+	amount := coinpkg.NewAmountFromInt64(100000000)
+	expectedValue := big.NewRat(21, 1)
+
+	converted, err := b.convertBtcAmountToFiat(amount, rates.USD.String())
+	require.NoError(t, err)
+
+	require.Equal(t, expectedValue, converted)
 }
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -249,6 +249,31 @@ func (backend *Backend) accountFiatBalance(account accounts.Interface, fiat stri
 	return fiatValue, nil
 }
 
+// Converts bitcoin amount to fiat.
+func (backend *Backend) convertBtcAmountToFiat(amount coin.Amount, fiat string) (*big.Rat, error) {
+	btcCoin, err := backend.Coin(coinpkg.CodeBTC)
+	if err != nil {
+		return nil, err
+	}
+	coinDecimals := new(big.Int).Exp(
+		big.NewInt(10),
+		big.NewInt(int64(btcCoin.Decimals(false))),
+		nil,
+	)
+	price, err := backend.RatesUpdater().LatestPriceForPair(btcCoin.Unit(false), fiat)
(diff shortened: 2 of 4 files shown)


backend/backend.go
```diff
@@ -642,6 +642,17 @@ func (backend *Backend) AccountsTotalBalanceByKeystore() (map[string]KeystoreTot
 			)
 			currentTotal.Add(currentTotal, fiatValue)
 		}
+		// Add lightning balance if it is enabled.
+		if len(backend.Config().LightningConfig().Accounts) > 0 {
```
Beerosagos, line 646: Could you please add a `lightningConfig.LightningEnabled()` method that returns `len(backend.Config().LightningConfig().Accounts) > 0` inside config/lightning.go, and use it to check if lightning is enabled? this way we avoid duplicating the check around the code. We may want to update it in the future.

backend/backend.go
```diff
@@ -650,6 +661,33 @@ func (backend *Backend) AccountsTotalBalanceByKeystore() (map[string]KeystoreTot
 	return totalAmounts, nil
 }
 
+// updateTotalWithLightningBalance adds lightning balance to current fiat total.
+func (backend *Backend) updateTotalWithLightningBalance(currentTotal *big.Rat) (*big.Rat, error) {
```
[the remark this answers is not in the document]
Beerosagos: Also, it would be nice to have a unit test for this! :innocent:

backend/backend.go
```diff
@@ -650,6 +661,33 @@ func (backend *Backend) AccountsTotalBalanceByKeystore() (map[string]KeystoreTot
...
+	fiat := backend.Config().AppConfig().Backend.MainFiat
+	coinDecimals := new(big.Int).Exp(
+		big.NewInt(10),
+		big.NewInt(int64(8)),
```
Beerosagos, line 669: You can retrieve the BTC coin with `backend.Coin(coinpkg.CodeBTC)` and use it to retrieve the values for decimals and "BTC" unit.

Beerosagos (changes requested):

backend/backend.go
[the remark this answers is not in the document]
strmci: Created new function `convertAmountToFiat` with a unit test that is used in `AccountsTotalBalanceByKeystore` and `ChartData`

backend/backend.go
strmci: that will be better, `LightningEnabled` added

backend/backend.go
strmci: I added the `backend.Coin(coinpkg.CodeBTC)` to the `convertAmountToFiat` function, now the function is only for BTC amounts, shall we make it work for other coins?

Beerosagos: We could make it work with any coin, or rename it to `convertBtcAmountToFiat`. Maybe keep it only on BTC is good for now, we can always extend. Also, generic backend refactoring are not in the scope of the branch, and makes merging the master more difficult

backend/accounts_test.go
```diff
@@ -1506,3 +1507,17 @@ func TestAccountsTotalBalanceByKeystore(t *testing.T) {
...
+	b := newBackend(t, testnetDisabled, regtestDisabled)
+	defer b.Close()
+	b.ratesUpdater = rates.MockRateUpdater()
+	defer b.ratesUpdater.Stop()
+	// convert 1 BTC to USD
+	amount := coinpkg.NewAmountFromInt64(100000000)
+	converted, err := b.convertAmountToFiat(amount, "USD")
+	expectedValue := big.NewRat(21, 1)
+
+	require.NoError(t, err)
```
Beerosagos, line 1521: nit: this line should be right below b.convertAmountToFiat for better readability

backend/accounts_test.go
```diff
@@ -1506,3 +1507,17 @@ func TestAccountsTotalBalanceByKeystore(t *testing.T) {
...
 }
+
+func TestConvertAmountToFiat(t *testing.T) {
+	b := newBackend(t, testnetDisabled, regtestDisabled)
+	defer b.Close()
+	b.ratesUpdater = rates.MockRateUpdater()
+	defer b.ratesUpdater.Stop()
+	// convert 1 BTC to USD
+	amount := coinpkg.NewAmountFromInt64(100000000)
+	converted, err := b.convertAmountToFiat(amount, "USD")
```
Beerosagos, line 1518: nit: "USD" -> rates.USD.String()

Beerosagos (changes requested):

strmci:
@Beerosagos I renamed the function to `convertBtcAmountToFiat` and made all the suggested changes, PTAL

PR #2627, opened 2024-03-12
backend: handle connection error during lightning activation

strmci:
If an error occurs during the lightning connection process, it is returned, but the lightning configuration is not cleared. This leads to the appearance of a lightning wallet in the app when, in reality, it doesn't exist. This commit addresses this issue by deactivating the lightning configuration if any error occurs during this step.

### DIFF
--- a/backend/lightning/lightning.go
+++ b/backend/lightning/lightning.go
@@ -121,6 +121,9 @@ func (lightning *Lightning) Activate() error {
 	}
 
 	if err = lightning.connect(true); err != nil {
+		if deactivateErr := lightning.Deactivate(); deactivateErr != nil {
+			lightning.log.Error(deactivateErr)
+		}
 		return err
 	}
 


backend/lightning/lightning.go
```diff
@@ -117,11 +117,11 @@ func (lightning *Lightning) Activate() error {
 	lightningConfig.Accounts = append(lightningConfig.Accounts, &lightningAccount)
 
 	if err = lightning.setLightningConfig(lightningConfig); err != nil {
-		return err
+		return handleLightningSetupError(lightning, err)
```
Beerosagos, line 120: If `setLightningConfig` fails, calling `handleLightningSetupError` doesn't do anything, right? 
I think we can skip it here and move the content of `handleLightningSetupError` directly inside the if that checks `lightning.connect(true)` err value.

Beerosagos (changes requested):
Nice! Just a small nit

backend/lightning/lightning.go
strmci: I agree. I've moved the `handleLightningSetupError` logic only to `lightning.connect(true)` if
@Beerosagos PTAL

Beerosagos (approved):
tACK! Nice :+1:

PR #2628, opened 2024-03-13
frontend: fix continue btn disabled on choosing device name setup

shonsirsha:
Fixed a regression in: 4633efb486632425943ef11f35dcda440250aeb4

Also added a new device name error state 'tooShort'. Error will only show when user has started typing and deletes everything. Won't show on initial (empty) name.


thisconnect (approved):
tested 7f973f64cc9b73e6a30b2c86b61ca4f51ef37255 LGTM 👍

PR #2630, opened 2024-03-13
frontend: fix horizontal scrollbar checklist on mobile

thisconnect:
This fixes having horizontal scrollbar on mobile on the checklist step of the setup.


thisconnect:
I could only reprocue it in webdev but not in qt-osx. Could happen on Android too.

Screen Shot 2024-03-13 at 21 06 57

PR #2631, opened 2024-03-13
frontend: reduce container spacing on mobile

thisconnect:
With this ViewBottons are nicely positioned on mobile.

One change affects large screen that fixes a small misalingment of
aopp success buttons.

Buttons container on large screen should have no padding if inside
a View component with fitContent prop, else they are misaligned
with the content.

Button padding is only needed for small screen and defined later
in this change.

I tested the following views:
- send scan qr code dialog
- AOPP
- AppUpgradeRequired
- AuthRequired
- walletconnect (tested only without active pairing)
- ReceiveAccountsSelector accounts/select-receive
- party tested bitsurance
- pairing
- passphrase

### DIFF
--- a/frontends/web/src/components/view/view.module.css
+++ b/frontends/web/src/components/view/view.module.css
@@ -77,7 +77,7 @@
         min-height: auto !important;
     }
     .inner.fit {
-        padding-bottom: var(--space-half);
+        padding-bottom: 0;
         padding-top: 0;
     }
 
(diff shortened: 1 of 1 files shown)


thisconnect:
@shonsirsha I did a lot fo testing and am pretty confident about this PR, but if you could have a quick look at walletconnect (mostly mobile) that would be nice as I only tested without active pairing.

There is one more FE view issue that I am aware of, but didn't want to do in this PR. The issue is that the buttons in ViewButtons not always use whole available width on mobile. But I will open a separate PR for this.

PR #2632, opened 2024-03-13
frontend: center appupgraderequired view

thisconnect:
App-upgrade-required component looks better if fully centered, as it is just a message and a download button.


thisconnect:
current / master

Screen Shot 2024-03-13 at 20 56 35

Screen Shot 2024-03-13 at 20 56 41


with this commit
    
Screen Shot 2024-03-13 at 20 55 41

Screen Shot 2024-03-13 at 20 56 59

PR #2634, opened 2024-03-14
frontend: fixed weblint by sorting locales 

thisconnect:
Manually ran:

    npx locize-cli format ./frontends/web/src/locales --format json


benma:
fyi, `make locize-fix` does the same and maybe more, and is checked by CI, so this is always the right command to fix the locale files.

benma (approved):

thisconnect:
I can't install it globally, yeah it's the same module.

Do you know why the ci didn't run on @sutterseba PR?

PR #2636, opened 2024-03-18
Update Breez SDK to v0.3.8

dangeross:
**SDK v0.3.6:** This update fixes various VLS signer issues updating to v0.11, mitigates connectivity issues to Greenlight at startup and improves GRPC connection performance to chain and breez services

**SDK v0.3.8:** This update bypasses the Greenlight signer loop initialization and lowers the TCP keepalive


Beerosagos (approved):
tACK! It doesn't always work in a reliable way, but it seems to be due to the GL part

PR #2642, opened 2024-03-21
backend/usb: update hid identifier

Beerosagos:
Whith this update the device identifier changes when the bitbox reboots into bootloader and forces the app to detect the reboot, that was sometime missed.


benma (approved):
utACK

PR #2643, opened 2024-03-21
frontend: add device password recommendation 

sutterseba:
Adds a small note below the password demo video which links to the knowledge base entry on how to choose a good device password.

Some users are confused when it comes to the device password and may choose overly long device password. The knowledge base entry provides valuable context and enables the user to make an informed decision.

To avoid confusion with BIP-39 passphrases, "password" has been rephrased to "device password".

**Preview:**

Screenshot 2024-04-29 at 13 38 07

Screenshot 2024-04-29 at 13 38 21

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/setup/password.tsx
+++ b/frontends/web/src/routes/device/bitbox02/setup/password.tsx
@@ -15,16 +15,29 @@
  */
 
 import { useTranslation } from 'react-i18next';
+import { i18n } from '../../../../i18n/i18n';
 import { View, ViewContent, ViewHeader } from '../../../../components/view/view';
 import { Backup } from '../../../../api/backup';
 import { PasswordEntry } from '../components/password-entry/password-entry';
 import { Status } from '../../../../components/status/status';
 import { MultilineMarkup } from '../../../../utils/markup';
 import { convertDateToLocaleString } from '../../../../utils/date';
+import { A } from '../../../../components/anchor/anchor';
+import { Info } from '../../../../components/icon';
+import style from './password.module.css';
 
 type Props = {
   errorText: string | undefined;
 }
+
+const getSupportLink = () => {
+  switch (i18n.resolvedLanguage) {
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/device/bitbox02/setup/password.tsx
```diff
@@ -15,17 +15,30 @@
...
 type Props = {
   errorText: string | undefined;
 }
 
+const getSupportLink = () => {
+  switch (i18n.resolvedLanguage) {
+  case 'de':
+    return 'https://shiftcrypto.support/help/de-de/3-basics/84-wie-lange-sollte-mein-geratepasswort-sein';
+  default:
+    return 'https://shiftcrypto.support/help/en-us/3-basics/84-wie-lange-sollte-mein-geratepasswort-sein';
```
thisconnect, line 38: nit: if you change from german to english the uri still contains the german verision. (silly knowledge base tool).

please use 
https://shiftcrypto.support/help/en-us/3-basics/84-how-long-should-my-device-password-be

thisconnect (changes requested):

thisconnect:
could you paste a screenshot of SetPasswordWithBackup ?

sutterseba:
SetPasswordWithBackup:

Screenshot 2024-03-22 at 09 42 29

thisconnect:
I think we should close this as discussed in private. Reaons:
- we don't to add links/buttons to leave the workflow while setting up their wallet
- coming up with real random is really difficult for humans
- recommending something new may lead to uncertainty for existings users.. "is my BB now insecure because I used a pwd length of only 4"
- what is the equivalent of 5 random upp/lower/numbers in lowercase only etc.
- also imo we don't explain fully, i.e. after reading the KB the user currently cannot make an educated choice, i.e what does it mean to not set a password at all, what does a weak password protect against

for these reasons I think we should close this for now and rethink if/how/where to inform

PR #2645, opened 2024-03-22
frontend: fix resizing chart

thisconnect:
Sometimes the chart did not resize properly, leaving half of the
area empty.

Removed setTimeout hack and just call chart.timeScale().fitContent()
seem to fix this.

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 Shift Crypto AG
+ * Copyright 2023-2024 Shift Crypto AG
  *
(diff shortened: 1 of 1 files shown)


thisconnect:
Please also test on linux cc @strmci

strmci:
tested with qt-linux, looks good

thisconnect:
> On macOS qt: seems like when resizing from a small screen to a bigger one the chart overflows.

nice catch. Looks like it happens when there is a single resize event that changes the window and sidebar visiblity.

rebased..

thisconnect:
> On macOS qt: seems like when resizing from a small screen to a bigger one the chart overflows.

Should be fixed now, if there was only a single resize event when clicking on maximize window the isMobile wasn't properly take into account.

PR #2647, opened 2024-03-26
frontend: better typing for Input component

shonsirsha:
Used `HTMLProps<HTMLInputElement>` for Input component props to remove unnecessary code & improve typing

### DIFF
--- a/frontends/web/src/components/forms/input.tsx
+++ b/frontends/web/src/components/forms/input.tsx
@@ -15,39 +15,21 @@
  * limitations under the License.
  */
 
-import { forwardRef } from 'react';
+import { ChangeEvent, HTMLProps, forwardRef } from 'react';
 import styles from './input.module.css';
 
-type withValue = {
-  value: string | number;
-  defaultValue?: string | number;
-} | {
-  defaultValue: string | number;
-}
 
-export type Props = withValue & {
+
+export type Props = {
     align?: 'left' | 'right';
-    autoFocus?: boolean;
     children?: React.ReactNode;
     className?: string;
-    disabled?: boolean;
     error?: string | object;
-    id?: string;
-    label?: string;
-    min?: string;
-    name?: string;
-    onInput?: (e: any) => void;
-    onPaste?: (e: any) => void;
-    pattern?: string;
-    placeholder?: string;
-    readOnly?: boolean;
-    step?: string;
-    title?: string;
+    onInput?: (e: ChangeEvent<HTMLInputElement>) => void;
     transparent?: boolean;
-    type?: 'text' | 'password' | 'number';
-    maxLength?: number;
     labelSection?: JSX.Element | undefined;
-}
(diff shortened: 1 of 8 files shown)


frontends/web/src/components/forms/input.tsx
```diff
@@ -15,39 +15,20 @@
...
-    readOnly?: boolean;
-    step?: string;
-    title?: string;
+    onInput?: (e: ChangeEvent<HTMLInputElement>) => void;
     transparent?: boolean;
-    type?: 'text' | 'password' | 'number';
-    maxLength?: number;
     labelSection?: JSX.Element | undefined;
-}
+} & Omit<HTMLProps<HTMLInputElement>, 'onInput'>
```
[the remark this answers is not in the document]
thisconnect: > however I'm unsure if there's a specific reason that made us to use onInput instead of onChange

afaik onInput triggers on every change, and onChange fires when the element looses focus. So we prefer onInput for validation etc.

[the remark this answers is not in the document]
thisconnect: oh good to know.. in this case I think our Input component should have `& Omit<HTMLProps<HTMLInputElement>, 'onInput'>` so that we can use onInput and onChange...


found a link to an explanation in your SO link https://github.com/facebook/react/issues/9567#issuecomment-298326902

thisconnect:
very nice, somewhat related:

frontends/web/src/routes/account/send/utxos.tsx uses checkboxes, it looks to me that they could be refactored the same way

```diff
-  const handleUTXOChange = (event: React.SyntheticEvent) => {
-    const target = event.target as HTMLInputElement;
+  const handleUTXOChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const target = event.target;
```

frontends/web/src/components/forms/input.tsx
```diff
@@ -15,39 +15,21 @@
...
-import { forwardRef } from 'react';
+import { ChangeEvent, HTMLProps, forwardRef } from 'react';
 import styles from './input.module.css';
 
-type withValue = {
-  value: string | number;
-  defaultValue?: string | number;
-} | {
-  defaultValue: string | number;
-}
```
thisconnect, line 26: question: this forced us to either set defaultValue OR value + optional defaultValue, but now it's gone. Was this in the way somehow?

thisconnect (approved):
LGTM with last question

frontends/web/src/components/forms/input.tsx
thisconnect: > IMO it's a given that input will need to have a value

fine with me, but then can/should we make value mandatory?

thisconnect: ack lets keep as is for now.

thisconnect (approved):
untested LGTM

PR #2652, opened 2024-03-28
frontend/android: fix app close and re-open crash

Beerosagos:
On various Android versions, when closing the app through the app manager (i.e. swiping out the app from the list you get by pressing the squared button on the device) the backend service was still running. This was causing a crash the next time the app was opened, as an already running backend was detected.

The nice fix for this would be to call Mobileserver.shutdown() when destroying the GoService, but this is not easy, as the server shutdown is only drafted, at the moment.

This commit explicitly kills the app (the same way as it happens when clicking the back button) to fix the issue.


benma (approved):
tACK, thanks! :tada:

PR #2659, opened 2024-04-05
android: add com.google.android.gms.version to AndroidManifest.xml

NicolaLS:
Edited PR #2646 

A user reported a crash when accessing the camera for QR code scanning. The crash report said that this tag was required.

I could not find much info about why accessing the camera would require Google Play Services. It seems however that simply adding the tag should not prevent the app to be installed on devices without Google Play Services.

Edit: the app did not build because the resource `@integer/google_play_services_version` was not defined/did not exist which indicates that google play services was not properly integrated/recognized by the build system.

Adding google play services to the dependencies in `build.gradle` fixes the problem and the app builds. Two GrapheneOS users confirmed that this build fixes the issue for them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add specific titles to guides replacing the generic "Guide" previously used on all pages
 - Android: enable transactions export feature
 - Format amounts using localized decimal and group separator
+- Fix BitBoxApp crash on GrapheneOS and other phones without Google Play Services when scanning QR codes.
 
 ## 4.42.0
 - Preselect backup when there's only one backup available
--- a/frontends/android/BitBoxApp/app/build.gradle
+++ b/frontends/android/BitBoxApp/app/build.gradle
@@ -37,6 +37,11 @@ dependencies {
     implementation "android.arch.lifecycle:extensions:1.1.1"
     implementation "android.arch.lifecycle:viewmodel:1.1.1"
     implementation project(path: ':mobileserver')
+    // The frontend depends on qr-scanner (https://github.com/nimiq/qr-scanner)
+    // which uses the Shape Detection API BarcodeDetector.  We use Google Play Services
+    // to interact with the native barcode scanner. See (2):
+    // https://developers.google.com/ml-kit/vision/barcode-scanning/android
+    implementation 'com.google.android.gms:play-services-mlkit-barcode-scanning:18.3.0'
 
 
     // Fix Duplicate class
--- a/frontends/android/BitBoxApp/app/src/main/AndroidManifest.xml
+++ b/frontends/android/BitBoxApp/app/src/main/AndroidManifest.xml
@@ -34,6 +34,9 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
+        <!-- This tag might be required when interacting with Google Play Services, see https://developers.google.com/android/guides/setup. -->
+        <!-- We interact with Google Play services through the JS QR code scanning library, which invokes the GPS Barcode scanner if available. -->
+        <meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" />
         <!-- launchMode: Makes sure that onIntent/onResume is reliably called on usb attached/detached events -->
         <!-- configChanges: Makes sure onCreate is not called on those changes, which prevents an ugly reload -->
         <activity


NicolaLS:
I think that _requesting/checking availability_ of the native qr scanner was what broke on GrapheneOS because the system does not like that the app even tried interacting with google play services. However I **think** that we never used the native barcode scanner even on other platforms because we did not properly integrate it.

I'm confused on the integration thought [qr-scanner](https://github.com/nimiq/qr-scanner) is using [this](https://developer.chrome.com/docs/capabilities/shape-detection#os-support) and it says _Barcode detection is available on macOS, ChromeOS, and Android. [Google Play Services](https://play.google.com/store/apps/details?id=com.google.android.gms) are required on Android._ but **which ones**??

The only dependency I found that made sense was [ml kit barcode scanning.](https://developers.google.com/ml-kit/vision/barcode-scanning/android)

**important** I'm worried that adding `mlkit-barcode-scanning` only solves the problem that the app did not build. But if its not the correct service the app will continue not using the native barcode scanner as it is doing now (assumption).

benma:
>  think that we never used the native barcode scanner

Yeah we don't, we use some qr code scanning JS lib. The question is then why on GrapheneOS the Play Service thing is needed/invoked.

NicolaLS:
> > think that we never used the native barcode scanner
> 
> Yeah we don't, we use some qr code scanning JS lib. The question is then why on GrapheneOS the Play Service thing is needed/invoked.

I know, what I meant was _I think our qr-scanner lib is never using the native barcode scanner_ it only requests the google play service to check whether it is available (which graphene did not like) (assumtion) but since we never added it as a dependency/integrated it properly (see [here](https://developers.google.com/ml-kit/vision/barcode-scanning/android#try-it-out)) it always used the "worker" from the library (assumtion)

NicolaLS:
> > > think that we never used the native barcode scanner
> > 
> > 
> > Yeah we don't, we use some qr code scanning JS lib. The question is then why on GrapheneOS the Play Service thing is needed/invoked.
> 
> I know, what I meant was _I think our qr-scanner lib is never using the native barcode scanner_ it only requests the google play service to check whether it is available (which graphene did not like) (assumtion) but since we never added it as a dependency/integrated it properly (see [here](https://developers.google.com/ml-kit/vision/barcode-scanning/android#try-it-out)) it always used the "worker" from the library (assumtion)

[here](https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L634) qr-scanner checks if the native barcode scanner is available and uses it's own (less performant?) implementation if not.

in case we don't need the native barcode scanner and want to avoid even requesting/checking its availability we could fork it and disable it before it even checks:
https://github.com/nimiq/qr-scanner/commit/87d5618c46fab5a6a63d34f2c108e742bf5548be

(users confirmed that this works too)

thisconnect:
> I think https://github.com/digitalbitbox/bitbox-wallet-app/pull/2660 is better.

why?

thisconnect:
tested LGTM

reproduced the issue on my Android with 4.41.0

tested apk from this PR (BitBoxApp-macos-4d2fc910e822c9da2ca29ca4dbb8d84c417a9696.zip) works as expected.

NicolaLS:
> > I think #2660 is better.
> 
> why?

I think we don't care that much for APK size and with the other solution we don't need to use Google Play Services. Also it is more performant apparently, but I can not find the issues that I've read that in.

Also see:
https://developers.google.com/ml-kit/tips/installation-paths#how_to_choose_between_bundled_and_unbundled

benma:
@NicolaLS since this is using the unbundled option, the module is downloaded from google play services, but it still works on GrapheneOS. So the lack of play services was not the issue? What am I missing?

I initially thought the other PR was better (bundled), but if this works as a fix, then this could be better after all? Smaller APK size, and the module can be installed if missing directly after app install, according to [this](https://developers.google.com/ml-kit/vision/barcode-scanning/android), by using

```
<application ...>
      ...
      <meta-data
          android:name="com.google.mlkit.vision.DEPENDENCIES"
          android:value="barcode" >
      <!-- To use multiple models: android:value="barcode,model2,model3" -->
</application>
```

Let's try that?

NicolaLS:
> @NicolaLS since this is using the unbundled option, the module is downloaded from google play services, but it still works on GrapheneOS. So the lack of play services was not the issue? What am I missing?
> 
> I initially thought the other PR was better (bundled), but if this works as a fix, then this could be better after all? Smaller APK size, and the module can be installed if missing directly after app install, according to [this](https://developers.google.com/ml-kit/vision/barcode-scanning/android), by using
> 
> ```
> <application ...>
>       ...
>       <meta-data
>           android:name="com.google.mlkit.vision.DEPENDENCIES"
>           android:value="barcode" >
>       
> </application>
> ```
> 
> Let's try that?

I'll send the APK to a GrapheneOS tester and open a new PR.

NicolaLS:
>So the lack of play services was not the issue? What am I missing?

It is hard to understand this from user reports..some where saying they saw an error message about _native code debugging_ then we got the crash report that indicated the missing declaration in the manifest. I don't know which of those users were using Google Play Services in a Sandbox or had it disabled completely.

I think this could be a plausible explanation:
```javascript       
            // These are true since the feature is supported by WebView
            && 'BarcodeDetector' in window
            && BarcodeDetector.getSupportedFormats
```
```javascript
            // I think this is what crashes the app on Graphene Google Play Services are not declared in the Manifest.
            && (await BarcodeDetector.getSupportedFormats()).includes('qr_code');
```

On Graphene without GP Services but with a proprer Manifest, I think qr-scanner will just fall back to the JS implementation. Either `&& (await BarcodeDetector.getSupportedFormats()).includes('qr_code');` is `false` or it catches some exception [here](https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L565).

NicolaLS:
> On Graphene without GP Services but with a proprer Manifest, I think qr-scanner will just fall back to the JS implementation. Either `&& (await BarcodeDetector.getSupportedFormats()).includes('qr_code');` is `false` or it catches some exception [here](https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L565).

On the first glimpse it looks like it does not make sense that `&& (await BarcodeDetector.getSupportedFormats()).includes('qr_code');` would be false if it is available as an interface but GP Services was disabled but this is how the API works:

_Purely checking for the existence of the constructors to feature detect the Shape Detection API doesn't suffice. The presence of an interface doesn't tell you whether the underlying platform supports the feature. This is working [as intended](https://crbug.com/920961). It's why we recommend a defensive programming approach by doing feature detection like this:_

See [feature detection.](https://developer.chrome.com/docs/capabilities/shape-detection#featuredetection)

frontends/android/BitBoxApp/app/src/main/AndroidManifest.xml
```diff
@@ -34,6 +34,9 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
+        <!-- Fixes crash when accessing the camera on some devices -->
+        <!-- ```A required meta-data tag in your app's AndroidManifest.xml does not exist.  You must have the following declaration within the <application> element:     <meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" /> ``` -->
```
benma, line 38: Replace this comment with:

> This tag is needed when interacting with Google Play Services, see https://developers.google.com/android/guides/setup. We interact with Google Play services through the JS QR code scanning library, which invokes the GPS Barcode scanner if available.

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
--
+- Fix BitBoxApp crash on GrapheneOS by declaring Google Play Services in the AndroidManifest.xmls, and adding ML Kit barcode scanning as a dependency in build.gradle.
```
benma, line 4: this changelog is user facing, so it should be light on tech details. Just "Fix BitBoxApp crash on GrapheneOS and other phones without Google Play Services when scanning QR codes" or similar will do.

benma (changes requested):

frontends/android/BitBoxApp/app/build.gradle
```diff
@@ -37,6 +37,7 @@ dependencies {
     implementation "android.arch.lifecycle:extensions:1.1.1"
     implementation "android.arch.lifecycle:viewmodel:1.1.1"
     implementation project(path: ':mobileserver')
+    implementation 'com.google.android.gms:play-services-mlkit-barcode-scanning:18.3.0'
```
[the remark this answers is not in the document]
benma: Pls address this too

CHANGELOG.md
NicolaLS: thanks.

frontends/android/BitBoxApp/app/src/main/AndroidManifest.xml
```diff
@@ -34,6 +34,9 @@
...
+        <meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" />
```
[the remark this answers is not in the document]
NicolaLS: (context) we decided to just keep it since the APK from this CI build works (tested by @thisconnect).

thisconnect:
> The meta-data declaration in the AndroidManifest might not be required
to fix the issue.

this is in the commit message, do you want me to test again without the meta-data?

NicolaLS:
> > The meta-data declaration in the AndroidManifest might not be required
> > to fix the issue.
> 
> this is in the commit message, do you want me to test again without the meta-data?

@thisconnect we said we'll just leave it in here since it works and the meta data declaration does not work. in case you have time it would be of course nice to verity that it is indeed not required. but as discussed this should be ready to merge, now.

frontends/android/BitBoxApp/app/src/main/AndroidManifest.xml
```diff
@@ -34,6 +34,9 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
+        <!-- This tag might be required when interacting with Google Play Services, see https://developers.google.com/android/guides/setup. -->
```
benma, line 37: my suggestion was `This tag is needed` not `This tag might be required`, as I had the impression it was needed. Not sure why you changed it :) anyway

benma (approved):

benma:
This issue has been going on long enough, let's not spend more time with testing minor tweaks. Merged.

PR #2660, opened 2024-04-07
android: bundle ML Kit barcode scanner with app

NicolaLS:
Bundle the ML Kit barcode scanner with the app to avoid dynamically downloading the model via Google Play Services. Model is statically linked to the app at build time. See:
https://developers.google.com/ml-kit/vision/barcode-scanning/android#try-it-out

Alternative to #2659 

Pro's: avoid using google play services entirely (also performance improvement)
Con's: app size increases more

`make android` results in a 49.3MB on this branch and 35.1MB on #2659 I guess for production this can be optimized but no idea what the process is there. the developer resources say a 2.4MB increase is to be expected.


NicolaLS:
One GrapheneOS tester confirmed the app does not crash and the QR-Scanner works with and without Google Play Services.

benma:
Thanks! This looks like the right solution, well done!

> Bundle the ML Kit barcode scanner with the app to avoid dynamically
downloading the model via Google Play Services.

What does it mean: "dynamically downloading"? 

Could you try to explain very clearly what is going on, so future devs know what this dep is about exactly?

1. Does this change anything for non-GrapeheneOS users? Does this change the QR code scanner that is used for everybody?
2. In https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L634, `useBarcodeDetector` was true on GrapheneOS but using it failed? Is this value true on non-GrapheneOS apps? In [BarcodeDetector](https://developer.mozilla.org/en-US/docs/Web/API/BarcodeDetector) it says this API is only available in HTTPS, which the BitBoxApp WebView is not :thinking:

@thisconnect please test, you know most about our QR code scanning and how it is supposed to behave/look like.

frontends/android/BitBoxApp/app/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation "android.arch.lifecycle:extensions:1.1.1"
     implementation "android.arch.lifecycle:viewmodel:1.1.1"
     implementation project(path: ':mobileserver')
-
+    implementation 'com.google.mlkit:barcode-scanning:17.2.0'
```
benma, line 40: Please add a comment that the qr-code scanning JS lib uses this as the QR code scanner.

NicolaLS:
> Thanks! This looks like the right solution, well done!
> 
> > Bundle the ML Kit barcode scanner with the app to avoid dynamically
> > downloading the model via Google Play Services.
> 
> What does it mean: "dynamically downloading"?
> 
> Could you try to explain very clearly what is going on, so future devs know what this dep is about exactly?
> 
>     1. Does this change anything for non-GrapeheneOS users? Does this change the QR code scanner that is used for everybody?
> 
>     2. In https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L634, `useBarcodeDetector` was true on GrapheneOS but using it failed? Is this value true on non-GrapheneOS apps? In [BarcodeDetector](https://developer.mozilla.org/en-US/docs/Web/API/BarcodeDetector) it says this API is only available in HTTPS, which the BitBoxApp WebView is not 🤔
> 
> 
> @thisconnect please test, you know most about our QR code scanning and how it is supposed to behave/look like.

Thanks :) So tbh. I did not really know what was going on but I did some research.

1. qr-scanner performs feature detection to decide whether to use the native barcode detector or a js implementation.
- what's the native barcode scanner? [see this](https://developer.mozilla.org/en-US/docs/Web/API/BarcodeDetector).
- but that API is only available in HTTPS? No the statement is a bit misleading ([see this](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts)):
_Locally-delivered resources such as those with `http://127.0.0.1` URLs, `http://localhost` and `http://*.localhost` URLs (e.g. `http://dev.whatever.localhost/`), and `file://` URLs are also considered to have been delivered securely._
- the api is fully supported by `WebView` ([see this](https://developer.mozilla.org/en-US/docs/Web/API/BarcodeDetector#browser_compatibility))

2. So why does using the API interact with Google Play Services / requires ML Kit?
https://docs.google.com/document/d/1QeCDBOoxkElAB0x7ZpM3VN3TQjS1ub1mejevd2Ik1gQ/edit#heading=h.7dzxyu38gyk5
The ressource they link is deprecated and was replaced by ML Kit:
https://developers.google.com/ml-kit

-There are different [installation paths](https://developers.google.com/ml-kit/tips/installation-paths#key_differences_between_installation_options) this commit chooses to bundle the barcode scanner model with the app, [see how to choose between bundled and unbundled](https://developers.google.com/ml-kit/tips/installation-paths#how_to_choose_between_bundled_and_unbundled).

**Conclusion**:
>What does it mean: "dynamically downloading"?

Tbh. I don't know how Google Play Services work in detail, but you can use the dependency through them which means google play will download the model the first time it is requested and then manages its version in contrast to statically link it with the app which means nothing will be downloaded after the app was installed.

Actually [this](https://developers.google.com/ml-kit/tips/installation-paths#download-models) would be another alternative. Like this Google Play Services will download it automatically after the app is installed. If done as in #2659 then the model is actually downloaded the [first time it is used ](https://developers.google.com/android/guides/module-install-apis)(if I understand correctly) thus _dynamically downloaded_.

>Does this change anything for non-GrapeheneOS users? Does this change the QR code scanner that is used for everybody?

Android/GrapheneOS:

Yes Google Play Services will no longer manage the barcode scanner. It will not be downloaded when the native barcode scanner is requested the first time. Actually maybe non-Graphene users might not even used the native scanner as we did not declare it properly in the manifest / installed it via. google play services. It could be possible since qr-scanner would [silently recover](https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L565) from an _not implemented/unavailable_ error by falling back to the js implementation.

I could see how GrapheneOS had an issue with the app even trying and crashed but other Android systems just recovering to the JS implementation. But...no idea how to find out if that is true.

Other apps (qt/mac..):

No this _should not_ change anything. The native barcode scanner for the specific system will be used as described [here](https://docs.google.com/document/d/1QeCDBOoxkElAB0x7ZpM3VN3TQjS1ub1mejevd2Ik1gQ/edit#heading=h.bwvon0qcjrmr) but nothing changes.

>In https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L634, useBarcodeDetector was true on GrapheneOS but using it failed? Is this value true on non-GrapheneOS apps?

I'm 95% certain it is true since the app did not crash for an graphene user when I used a forked qr-scanner dependency that had the [_disableBarcodeDetector](https://github.com/nimiq/qr-scanner/blob/master/src/qr-scanner.ts#L631C47-L631C70) to true.

benma:
Took another look at https://github.com/digitalbitbox/bitbox-wallet-app/pull/2659, which thanks to your links and explanations I understand better. Thinking now that the unbundled option may be better if it works, which it seems it does according to @thisconnect' tests.

benma:
Closing this in favor of https://github.com/digitalbitbox/bitbox-wallet-app/pull/2659

PR #2661, opened 2024-04-09
backend/mobileserver: add another GH issue in comment

benma:
I just tried building with gomobile using modules+vendoring and it still didn't work. Added a relevant issue.


Beerosagos (approved):
LGTM

PR #2663, opened 2024-04-10
frontend/buy: fix bug in country selector

Beerosagos:
The CountrySelect component was crashing if no region was selected. This adds a check to fix the bug.


benma (approved):
tACK

PR #2668, opened 2024-04-11
docs/BUILD: document notarytool to notarize (altool is deprecated)

benma:
Apple decomissioned the previous commands to notarize an app, they do not work anymore.

Following
https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool, `notarytool` should now be used.


Beerosagos (approved):
utACK

PR #2669, opened 2024-04-11
Mempool space

Beerosagos:
Bitcoin Core fee estimation is not very accurate. With this change, meempool.space api will be used as a preferred estimation method per BTC fees. The fee estimations are fetched through a Shift mirror server.

### DIFF
--- a/backend/accounts/feetarget.go
+++ b/backend/accounts/feetarget.go
@@ -16,6 +16,7 @@ package accounts
 
 import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
+	"github.com/btcsuite/btcd/btcutil"
 )
 
 // FeeTarget interface has priority codes.
@@ -39,6 +40,10 @@ func NewFeeTargetCode(code string) (FeeTargetCode, error) {
 	case string(FeeTargetCodeNormal):
 	case string(FeeTargetCodeHigh):
 	case string(FeeTargetCodeCustom):
+	case string(FeeTargetCodeMempoolFastest):
+	case string(FeeTargetCodeMempoolHalfHour):
+	case string(FeeTargetCodeMempoolHour):
+	case string(FeeTargetCodeMempoolEconomy):
 	default:
 		return "", errp.WithStack(errp.Newf("Unrecognized fee target code %s", code))
 	}
@@ -58,10 +63,51 @@ const (
 	// FeeTargetCodeHigh is the high priority fee target.
 	FeeTargetCodeHigh FeeTargetCode = "high"
 
+	// FeeTargetCodeMempoolFastest is the mempool highest priority fee target.
+	FeeTargetCodeMempoolFastest FeeTargetCode = "mFastest"
+
+	// FeeTargetCodeMempoolHalfHour is the mempool half hour fee target.
+	FeeTargetCodeMempoolHalfHour FeeTargetCode = "mHalfHour"
+
+	// FeeTargetCodeMempoolHour is the mempool hour fee target.
+	FeeTargetCodeMempoolHour FeeTargetCode = "mHour"
+
+	// FeeTargetCodeMempoolEconomy is the mempool economy fee target.
+	FeeTargetCodeMempoolEconomy FeeTargetCode = "mEconomy"
+
 	// FeeTargetCodeCustom means that the actual feerate is supplied separately instead of being
 	// estimated automatically.
 	FeeTargetCodeCustom FeeTargetCode = "custom"
 
+	// DefaultMempoolFeeTarget is the default fee target for mempool fees.
+	DefaultMempoolFeeTarget = FeeTargetCodeMempoolHalfHour
+
 	// DefaultFeeTarget is the default fee target.
 	DefaultFeeTarget = FeeTargetCodeNormal
 )
+
+// MempoolSpaceFees contains mempool.space recommended fees API response
+// (https://mempool.space/docs/api/rest#get-recommended-fees)
+type MempoolSpaceFees struct {
+	FastestFee  int64 `json:"fastestFee"`
+	HalfHourFee int64 `json:"halfHourFee"`
+	HourFee     int64 `json:"hourFee"`
+	EconomyFee  int64 `json:"economyFee"`
+	MinimumFee  int64 `json:"minimumFee"`
+}
+
+// GetFeeRate returns the btcutil.Amount of the fee for the given FeeTargetCode.
+func (fees MempoolSpaceFees) GetFeeRate(code FeeTargetCode) btcutil.Amount {
+	var feeRatePerByte int64
+	switch code {
+	case FeeTargetCodeMempoolFastest:
+		feeRatePerByte = fees.FastestFee
+	case FeeTargetCodeMempoolHalfHour:
+		feeRatePerByte = fees.HalfHourFee
+	case FeeTargetCodeMempoolHour:
+		feeRatePerByte = fees.HourFee
+	case FeeTargetCodeMempoolEconomy:
+		feeRatePerByte = fees.EconomyFee
+	}
+	return btcutil.Amount(feeRatePerByte * 1000)
+}
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -18,6 +18,7 @@ package btc
 import (
 	"encoding/base64"
 	"fmt"
+	"net/http"
 	"os"
 	"path"
 	"sort"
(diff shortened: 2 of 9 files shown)


backend/coins/btc/account.go
```diff
@@ -139,13 +148,16 @@ func NewAccount(
...
 
 		// feeTargets must be sorted by ascending priority.
+		// current block waiting is set to be compatible with mempool.space estimation
+		// (see https://mempool.space/docs/api/rest#get-recommended-fees for details).
 		feeTargets: []*FeeTarget{
 			{blocks: 24, code: accounts.FeeTargetCodeEconomy},
-			{blocks: 12, code: accounts.FeeTargetCodeLow},
-			{blocks: 6, code: accounts.FeeTargetCodeNormal},
-			{blocks: 2, code: accounts.FeeTargetCodeHigh},
+			{blocks: 6, code: accounts.FeeTargetCodeLow},
```
Beerosagos, line 155: We'll need to update `send.feeTarget.description.*` frontend translations.

benma (commented):
I noticed another issue:

The current (old) estimator is using Bitcoin Core's `estimatesmartfee` call, which only takes into consideration past blocks. So fees are updated only when a new block is found. This is why in the app, the fees are updated also when a new block is found.

mempool.space fees take the mempool into consideration, so updating only when a block is found migt not be not sufficient to get accurate fees.

In general it was maybe not a good idea to download the fee estimates even if the user is not going to make a transaction, just viewing the account.

I propose we switch the logic to download the fees when the user needs it, i.e. when they go to the send screen and the fee targets are to be loaded, i.e. at the same time (or part of) the `account/${code}/fee-targets` API call. This would work for both the old and the new fee estimators.

Beerosagos:
rebased ^

frontends/web/src/locales/en/app.json
```diff
@@ -1560,6 +1560,10 @@
         "low": "2 hours (12 blocks)",
         "low_eth": "5 minutes or less",
         "low_ltc": "30 minutes (12 blocks)",
+        "mEconomy": "-",
```
Beerosagos, line 1563: mempool.space doesn't give a time estimation for the economy fee. It this ok?

Beerosagos:
Pushed a new update:
- removed the fee caching 
- created different fee targets for mempool.space
- removed the deletion of duplicated fees from `account.FeeTargets()`
- addressed the other comments

@benma PTAL :pray: 

I feel like we could refactor a bit more, moving the fee handling part from the account to the coin. But I think it's better to do it in a separate PR, to keep things simple.

backend/accounts/feetarget.go
```diff
@@ -58,10 +59,51 @@ const (
...
+type MempoolSpaceFees struct {
+	FastestFee  int64 `json:"fastestFee"`
+	HalfHourFee int64 `json:"halfHourFee"`
+	HourFee     int64 `json:"hourFee"`
+	EconomyFee  int64 `json:"economyFee"`
+	MinimumFee  int64 `json:"minimumFee"`
+}
+
+// GetFeeAmount returns the btcutil.Amount of the fee for the given FeeTargetCode.
+func (fees MempoolSpaceFees) GetFeeAmount(code FeeTargetCode) btcutil.Amount {
```
benma, line 96: Better name could be `GetFeeRate()`, as this does not return the fee, but the fee rate (sats/vB).

backend/coins/btc/account.go
```diff
@@ -476,26 +475,72 @@ func (account *Account) Notifier() accounts.Notifier {
...
+// The minimum relay fee is used as a last resource fallback in case also Bitcoin Core is
+// unavailable.
+func (account *Account) feeTargets() []*FeeTarget {
+	// for mainnet BTC we fetch mempool.space fees, as they should be more reliable.
+	mempoolFees := &accounts.MempoolSpaceFees{}
+	if account.coin.Code() == coin.CodeBTC {
+		_, err := util.APIGet(account.httpClient, mempoolSpaceMirror, "", 1000, mempoolFees)
+		if err != nil {
+			mempoolFees = nil
+			account.log.WithError(err).Error("Fecthing fees from " + mempoolSpaceMirror + " failed")
```
benma, line 490: fyi there is also `Errorf`: `.Errorf("Fetching fees from %s failed", mempoolSpaceMirror)`

also: type in `Fetching`

backend/coins/btc/account.go
```diff
@@ -476,26 +475,72 @@ func (account *Account) Notifier() accounts.Notifier {
...
+
+		if mempoolFees != nil {
+			feeRatePerKb = mempoolFees.GetFeeAmount(feeTarget.code)
+		} else {
+			// If mempool.space fees are not available, we fallback on Bitcoin Core estimation.
+			// If even that one is not available, we just offer the min relay fee.
+			if account.coin.Code() == coin.CodeBTC {
 				account.log.WithField("fee-target", feeTarget.blocks).
-					Warning("Fee could not be estimated. Taking the minimum relay fee instead")
+					Debug("Fee server not available. Trying to estimate.")
```
benma, line 528: Isn't this log redundant with the one above (`Fecthing fees from " + mempoolSpaceMirror + " failed")`)?

backend/coins/btc/account.go
```diff
@@ -476,26 +475,72 @@ func (account *Account) Notifier() accounts.Notifier {
...
 	}
-	for _, feeTarget := range account.feeTargets {
-		feeRatePerKb, err := account.coin.Blockchain().EstimateFee(feeTarget.blocks)
-		if err != nil {
-			if account.coin.Code() != coin.CodeTLTC {
+
+	for _, feeTarget := range feeTargets {
+		var feeRatePerKb btcutil.Amount
+
+		if mempoolFees != nil {
```
benma, line 521: this condition is true always even for TBTC or LTC as you initialize it to ` mempoolFees := &accounts.MempoolSpaceFees{}`.

backend/coins/btc/account.go
```diff
@@ -476,26 +475,72 @@ func (account *Account) Notifier() accounts.Notifier {
...
-func (account *Account) updateFeeTargets() {
-	defer account.feeTargetsLock.Lock()()
+// feeTargets fetches the available fees. For mainnet BTC it uses mempool.space estimation.
+//
+// For the other coins or in case mempool.space is not available it fallbacks on Bitcoin Core.
+// The minimum relay fee is used as a last resource fallback in case also Bitcoin Core is
+// unavailable.
+func (account *Account) feeTargets() []*FeeTarget {
+	// for mainnet BTC we fetch mempool.space fees, as they should be more reliable.
+	mempoolFees := &accounts.MempoolSpaceFees{}
```
benma, line 485: init to `nil`, set to nonnil inside the condition below - see comment below about the bug this causes otherwise.

benma (changes requested):
I run into this error:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/4ae57210793e7cf6eb25d7a2213e51e6f05292cc/backend/accounts/feetarget.go#L43

Because the mempool space codes are not covered.

backend/accounts/feetarget.go
Beerosagos: Ack, I wasn't really satisfied about the name either

backend/coins/btc/account.go
Beerosagos: Removed :+1:

backend/coins/btc/account.go
Beerosagos: Shoot! I don't know how I missed this during the testing. Great catch :pray:

Beerosagos:
rebased

Beerosagos:
@benma Thanks for the review and for catching the bugs. I think I only checked that the fees were displayed properly in the interface and I didn't test an actual signing, so I missed the error you got. 

Now it should be good.

benma:
> @benma Thanks for the review and for catching the bugs. I think I only checked that the fees were displayed properly in the interface and I didn't test an actual signing, 

I also didn't do any signing - the fees dropdown is disabled until you add an address/amount, and at that point it already failed.

benma (approved):
Thanks.

When I go to the send-screen, the fee-targets endpoint is called many times, which means there are many requests to the server. I fixed this in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2730.

tACK

Beerosagos:
> I also didn't do any signing - the fees dropdown is disabled until you add an address/amount, and at that point it already failed.

I was curious and checked: it actually worked with just the amount until you add an address, because it needs both the address and the amount to make the TX proposal (which was the failing part). This is how I didn't notice the error :D

> When I go to the send-screen, the fee-targets endpoint is called many times, which means there are many requests to the server. I fixed this in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2730.

Awesome :heart:  Thanks a lot

PR #2671, opened 2024-04-16
frontends/android: implement transactions export

Beerosagos:
Transactions export feature was disabled for Android environment, this implements it.

Android app storage space is sandboxed: the transactions csv is persisted inside the app dedicated storage and can be shared by any third-party app available on the system.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
@@ -18,6 +18,7 @@ import android.os.Message;
 import androidx.lifecycle.AndroidViewModel;
 import androidx.lifecycle.MutableLiveData;
 
+import java.io.File;
 import java.util.Locale;
 
 import mobileserver.GoAPIInterface;
@@ -123,6 +124,12 @@ public class GoViewModel extends AndroidViewModel {
         public void auth() {
             Util.log("Auth requested from backend");
             requestAuth();
+        }
+
+        @Override
+        public String getSaveFilename(String fileName)  {
+            File folder = getApplication().getApplicationContext().getExternalFilesDir(null);
+            return new File(folder, fileName).getAbsolutePath();
         }
 
         public void onAuthSettingChanged(boolean enabled) {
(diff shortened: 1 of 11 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -126,22 +126,7 @@ public void onReceive(Context context, Intent intent) {
          }
     };
 
-    private static String getMimeType(String url) {
```
Beerosagos, line 129: Moved in Util.java

util/config/appdir.go
```diff
@@ -84,7 +84,8 @@ func AppDir() string {
 // DownloadsDir returns the absolute path to the Downloads folder in the home folder.
 func DownloadsDir() (string, error) {
 	if runtime.GOOS == "android" {
-		return "", errp.New("android not yet supported")
+		// Android apps are sandboxed, we don't need to specify a folder.
+		return "", nil
```
benma, line 88: a bit confusing to use DownloadsDir() on Android if it's not the actual DownloadsDir(). not sure if doing sth else here is worth it though

benma (commented):
tested works :tada:  but probably the cache folder is preferable.

benma (approved):
tACK

Beerosagos:
rebased

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoViewModel.java
```diff
@@ -125,6 +126,12 @@ public void auth() {
             requestAuth();
         }
 
+        @Override
+        public String getSaveFilename(String fileName)  {
+            File folder = getApplication().getApplicationContext().getExternalFilesDir(null);
+            return new File(folder, fileName).getAbsolutePath();
```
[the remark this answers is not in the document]
Beerosagos: I'd keep it as is, there should be no big difference between the cache and the external files folders at a practical level. Both are protected areas, owned by the app and that cannot be accessed without the app explicitly sharing it. Also it is deleted when the app is uninstalled. The only difference I see is that saving the file in cache allows the user to delete it from memory by cleaning the bbapp cache without uninstalling it.

util/config/appdir.go
Beerosagos: Renamed to ExportsDir()

PR #2672, opened 2024-04-17
nodejs: update node to latest lts version 20.x

thisconnect:


thisconnect:
I merged and pushed first to my fork, for some reason it is stuck there, but I think it might be due to the fork
https://github.com/thisconnect/bitbox-wallet-app

PR #2673, opened 2024-04-18
changelog: guide title changelog

shonsirsha:
previously added in c37e7f76d8e62e00a89277b424559db76b887596 . Was wrongly added to 4.42.0 changelog, should be under Unreleased instead.


thisconnect (approved):
LGTM 👍

PR #2674, opened 2024-04-18
frontend: fix bip-85 graphic too small

thisconnect:
This change fixes the graphic about BIP-85 child keys that did not use the whole available width in the Qt app.


thisconnect:
Screen Shot 2024-04-18 at 10 56 18

thisconnect:
Screen Shot 2024-04-18 at 10 56 05

PR #2678, opened 2024-04-20
frontend: type convert coin from and to currency

thisconnect:
Part of typing all api calls.

Note: in the future we should call Fiat just currencies as it can contain Fiat as well as Bitcoin.

### DIFF
--- a/frontends/web/src/api/coins.ts
+++ b/frontends/web/src/api/coins.ts
@@ -15,8 +15,8 @@
  */
 
 import { subscribeEndpoint, TSubscriptionCallback } from './subscribe';
-import { CoinCode } from './account';
-import { ISuccess } from './backend';
+import type { CoinCode, Fiat } from './account';
+import type { ISuccess } from './backend';
 import { apiPost, apiGet } from '../utils/request';
 
 export type BtcUnit = 'default' | 'sat';
@@ -45,4 +45,42 @@ export type TAmount = {
 
 export const parseExternalBtcAmount = (amount: string): Promise<TAmount> => {
   return apiGet(`coins/btc/parse-external-amount?amount=${amount}`);
+};
+
+type TConvertCurrency = {
+  amount: string;
+  coinCode: CoinCode;
+  fiatUnit: Fiat;
+};
+
+type TConvertFromCurrencyResponse = {
+  success: true;
+  amount: string;
+} | {
+  success: false;
+  errMsg: string; // TODO: backend should return useful errorMessage
+};
+
+export const convertFromCurrency = ({
+  amount,
+  coinCode,
+  fiatUnit,
+}: TConvertCurrency): Promise<TConvertFromCurrencyResponse> => {
+  return apiGet(`coins/convert-from-fiat?from=${fiatUnit}&to=${coinCode}&amount=${amount}`);
+};
+
+type TConvertToCurrencyResponse = {
+  success: true;
+  fiatAmount: string;
+} | {
+  success: false;
+  // errMsg: string; // TODO: backend should return useful errorMessage
+};
+
+export const convertToCurrency = ({
+  amount,
+  coinCode,
+  fiatUnit,
+}: TConvertCurrency): Promise<TConvertToCurrencyResponse> => {
+  return apiGet(`coins/convert-to-plain-fiat?from=${coinCode}&to=${fiatUnit}&amount=${amount}`);
 };
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -247,18 +247,23 @@ class Send extends Component<Props, State> {
     }
   };
 
-  private txInput = (): accountApi.TTxInput => ({
-    address: this.state.recipientAddress,
-    amount: this.state.amount,
-    feeTarget: this.state.feeTarget || 'economy',
-    customFee: this.state.customFee,
-    sendAll: this.state.sendAll ? 'yes' : 'no',
-    selectedUTXOs: Object.keys(this.selectedUTXOs),
-  });
-
-  private sendDisabled = () => {
-    const txInput = this.txInput();
-    return !txInput.address || this.state.feeTarget === undefined || (txInput.sendAll === 'no' && !txInput.amount) || (this.state.feeTarget === 'custom' && !this.state.customFee);
+  private getValidTxInputData = (): Required<accountApi.TTxInput> | false => {
+    if (
+      !this.state.recipientAddress
+      || this.state.feeTarget === undefined
+      || (!this.state.sendAll && !this.state.amount)
+      || (this.state.feeTarget === 'custom' && !this.state.customFee)
+    ) {
+      return false;
+    }
+    return {
+      address: this.state.recipientAddress,
+      amount: this.state.amount,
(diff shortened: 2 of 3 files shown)


frontends/web/src/api/coins.ts
```diff
@@ -46,3 +46,41 @@ export type TAmount = {
...
+}: TConvertCurrency): Promise<TConvertFromCurrencyResponse> => {
+  return apiGet(`coins/convert-from-fiat?from=${fiatUnit}&to=${coinCode}&amount=${amount}`);
+};
+
+type TConvertToCurrencyResponse = {
+  success: true;
+  fiatAmount: string;
+} | {
+  success: false;
+  // errMsg: string; // TODO: backend should return useful errorMessage
```
thisconnect, line 77: backend currently doesn't return errorMessage https://github.com/digitalbitbox/bitbox-wallet-app/blob/43232f453dbe4a39a384f424515ecd9ae939b9ec/backend/handlers/handlers.go#L970-L972

frontends/web/src/api/coins.ts
```diff
@@ -46,3 +46,41 @@ export type TAmount = {
...
+  coinCode: CoinCode;
+  fiatUnit: Fiat;
+};
+
+type TConvertFromCurrencyResponse = {
+  success: true;
+  amount: string;
+} | {
+  success: false;
+  errMsg: string; // TODO: backend should return useful errorMessage
```
thisconnect, line 61: backend currently may return 'internal error' or 'invalid amount', it should proabably return errorMessage and errorCode instead.

 https://github.com/digitalbitbox/bitbox-wallet-app/blob/43232f453dbe4a39a384f424515ecd9ae939b9ec/backend/handlers/handlers.go#L995-L998

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -248,10 +247,10 @@ class Send extends Component<Props, State> {
     }
   };
 
-  private txInput = () => ({
+  private txInput = (): accountApi.TTxInput => ({
     address: this.state.recipientAddress,
     amount: this.state.amount,
-    feeTarget: this.state.feeTarget || '',
+    feeTarget: this.state.feeTarget || 'economy',
```
thisconnect, line 253: I'm not sure about this but ideally feeTarget is not an empty string..

thisconnect: refactored in the next commit so this fallback is not necessary.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -270,17 +270,24 @@ class Send extends Component<Props, State> {
       amountError: undefined,
       feeError: undefined,
     });
-    if (this.sendDisabled()) {
+    if (this.sendDisabled() || this.state.feeTarget === undefined) {
```
shonsirsha, line 273: Won't checking `feeTarget === undefined` be redundant here? As it's already evaluated the same way in `sendDisabled`?

shonsirsha (approved):
LGTM.. left 1 small question & comment.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -140,17 +138,17 @@ class Send extends Component<Props, State> {
...
             const account = this.getAccount();
             const paired = mobileChannel && pairing;
             const noMobileChannelError = pairing && !mobileChannel && account && isBitcoinBased(account.coinCode);
             this.setState(prevState => ({ ...prevState, paired, noMobileChannelError }));
           });
-      });
+      }).catch(console.error);
     }
-    apiGet('config').then(config => {
+    getConfig().then(config => {
```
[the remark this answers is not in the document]
thisconnect: componentDidMount cannot be an async function afaik.

frontends/web/src/routes/account/send/send.tsx
thisconnect: absolutely, but TypeScript ignores that feeTarget has been already checked in the sendDisabled function and complains, I'll check if I find an elegant solution.

thisconnect: Found an elegant solution and typescript found a bug, fixed in separate commit.

PTAL

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -270,30 +271,36 @@ class Send extends Component<Props, State> {
...
+    this.proposeTimeout = setTimeout(async () => {
+      const proposePromise = accountApi.proposeTx(this.getAccount()!.code, txInput);
+      // keep this as the last known proposal
+      this.lastProposal = proposePromise;
+      try {
+        const result = await proposePromise;
+        // continue only if this is the most recent proposal
+        if (proposePromise === this.lastProposal) {
+          this.txProposal(updateFiat, result);
+        }
```
shonsirsha, line 293: Always been like this, so no probs. Just for my understanding: would it also be ok to await `proposePromise` inside this `if` stmt?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -249,18 +245,23 @@ class Send extends Component<Props, State> {
...
+      return false;
+    }
+    return {
+      address: this.state.recipientAddress,
+      amount: this.state.amount,
+      feeTarget: this.state.feeTarget,
+      customFee: this.state.customFee,
+      sendAll: (this.state.sendAll ? 'yes' : 'no'),
+      selectedUTXOs: Object.keys(this.selectedUTXOs),
+    };
```
shonsirsha, line 264: Maybe I'm missing smt, but why are we returning this object exactly? Why not just return `true`? 🤔  And if so, change fn name to like `isTxInputDataValid` or something.

shonsirsha:
Thanks for addressing my comments.

frontends/web/src/routes/account/send/send.tsx
thisconnect: if we return true we run into this problem again https://github.com/digitalbitbox/bitbox-wallet-app/pull/2678#discussion_r1587746718

Reason: typescript doesn't pass type information it knows from inside a fn to the scope where it was called. So a condition like this `if (fn() {` would again need to have another check so it can ensure that feetarget is not undefined.

does that make sense?

frontends/web/src/routes/account/send/send.tsx
thisconnect: I haven't tested, but the purpose here is to do this check when we get a result. This is in the edge case that if the first promise takes longer than the latest promise. We don't want the promise that is the slowest. But we want the result of the promise that was called last.

does that make sense?

frontends/web/src/routes/account/send/send.tsx
shonsirsha: Ah, okay. Thanks!

shonsirsha (approved):
LGTM

frontends/web/src/api/coins.ts
thisconnect: As we have other inconsistencies like this, I propose to keep the TODO comment and maybe one day take a look at all those error vs. errorMsg vs. errorMessage etc. consistent.

cc @Beerosagos ack?

PR #2683, opened 2024-04-22
frontend: show spinner after unlock, before accounts are loaded

shonsirsha:
After unlock, especially the first time, it can take many seconds for the initial set of accounts to be loaded. Until then, the app shows nothing in the sidebar and 'Please connect your device to get started'.

The user is easily confused, it seems as something went wrong.

We can probably speed up this loading with firmware and app changes, but this is a quick improvement that shows a spinner until the accounts become available.

[Preview]:

tiuwerhdks

### DIFF
--- a/frontends/web/src/routes/device/waiting.tsx
+++ b/frontends/web/src/routes/device/waiting.tsx
@@ -16,9 +16,14 @@
 
 import { useTranslation } from 'react-i18next';
 import { i18n } from '../../i18n/i18n';
+import { getDeviceList } from '../../api/devices';
+import { syncDeviceList } from '../../api/devicessync';
+import { useSync } from '../..//hooks/api';
+import { useKeystores } from '../../hooks/backend';
 import { useDarkmode } from '../../hooks/darkmode';
 import { Entry } from '../../components/guide/entry';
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/device/waiting.tsx
```diff
@@ -27,6 +32,15 @@ export const Waiting = () => {
   const { t } = useTranslation();
   const { isDarkMode } = useDarkmode();
 
+  const keystores = useKeystores();
+  const devices = useSync(getDeviceList, syncDeviceList);
+  const hasAccounts = (keystores !== undefined && keystores.length) || (devices !== undefined && Object.keys(devices).length);
```
benma, line 37: the name of this var is confusing I think. In this moment there are no accounts :P 

how about `loading = ...`?

benma (approved):
utACK

frontends/web/src/routes/device/waiting.tsx
shonsirsha: > In this moment there are no accounts

Ah fair enough - I thought because keystore & devices are truthy it means we have an account, but not exactly at all (especially we're actually loading for accounts here).

Thanks.

PR #2686, opened 2024-04-24
frontend: run tests in ci in single thread

thisconnect:
CI sometimes still hangs, trying to run without threads

- https://github.com/vitest-dev/vitest/issues/2008
- https://github.com/vitest-dev/vitest/issues/3077

In earlier vitest versions the optinon was --no-threads, but this changed to --pool=forks in:
- https://github.com/vitest-dev/vitest/releases/tag/v1.0.0-beta.0
- https://github.com/vitest-dev/vitest/blob/main/docs/guide/migration.md#pools-are-standardized-4172

i# with '#' will be ignored, and an empty message aborts the commit.


thisconnect:
pipeline where CI was hanging https://github.com/digitalbitbox/bitbox-wallet-app/pull/2678/commits

thisconnect:
I almost thought it was just me, but now it happened again

- https://github.com/digitalbitbox/bitbox-wallet-app/pull/2690
- https://github.com/digitalbitbox/bitbox-wallet-app/actions/runs/8845678474/job/24290059454?pr=2690

shonsirsha (approved):
LGTM. Thanks for the findings.

I think we could also add this in the vite config instead (if we prefer to have everything vite related 'centralised' in that config file). Anyways, it's fine for me like this.

thisconnect:
> I think we could also add this in the vite config instead (if we prefer to have everything vite related 'centralised' in that config file). Anyways, it's fine for me like this.

Not sure, maybe `make webtest` locally can use multiple threads or do you prefer if it runs with the same config  as in CI?

shonsirsha:
I think it's good to actually have it the same as in the CI. @thisconnect

PR #2688, opened 2024-04-24
qt: fix unknown escape sequence warning

benma:
`\/` in the regular expressions is not a valid escape sequence in C++ strings. It still somehow worked, but it's good to fix the warning just in case.

We use raw string literals to avoid any escaping:
https://en.wikipedia.org/wiki/C%2B%2B11#New_string_literals


Beerosagos (approved):
utACK! Thanks :pray:

PR #2695, opened 2024-05-02
frontend: fix account summary spacing on small screen

thisconnect:
- on some smaller breakpoints but before it centeres the total balance was touching the left screen, added some padding
- filter buttons was not centered correctly

Screenshot 2024-05-02 at 11 57 54


strmci (approved):
tACK

PR #2696, opened 2024-05-02
frontend: use typed apis

thisconnect:
The api calls used in i18n/config.ts are already available.


shonsirsha (approved):
untested LGTM 👍 thx

PR #2699, opened 2024-05-03
frontend: trying responsive font size

thisconnect:
Andoird has the option to make text bigger which breaks the layout in the BitBoxApp.

This is a test to see if we can use the rem mode trick.


thisconnect:
ok got some feedback from @shonsirsha 

my next step is to cleanup master and convert all `rem` to `px` first. As currently 1rem = 16px, but with this change it will be 1rem = 10px.

PR #2700, opened 2024-05-06
backend/bitbox02bootloader: allow re-installing firmware

benma:
Problem: a new device ships with the latest firmware monotonic version. User sees 'Install' because the device is erased (no firmware on it), but during first install of firmware, unplugs. Upon replug, the device will show 'Invalid firmware' and boot into bootloader again. Since the monotonic version of the firmware to be installed is the same as the version stored on the device, and the device is not erased (half of the firmware is on it), the user sees no button to upgrade/instll anymore.

This commit fixes this by still showing the upgrade button if the versions are the same. The downside is that users see this even if the firmware is fully installed. To fix this, we'd need a new bootloader API call to check if the firmware is bootable/valid, so this is a workaround.

### DIFF
--- a/backend/devices/bitbox02bootloader/device.go
+++ b/backend/devices/bitbox02bootloader/device.go
@@ -223,8 +223,9 @@ func (device *Device) UpgradeFirmware() error {
 
 // VersionInfo contains version information about the upgrade.
 type VersionInfo struct {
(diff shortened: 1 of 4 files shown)


benma:
Replaced by https://github.com/digitalbitbox/bitbox-wallet-app/pull/2711

PR #2701, opened 2024-05-06
frontend/android: check if webview can go back or close the app

thisconnect:
This should replace native Android back behavior with either going
back in the webview history or closing the app.

In theory the react-router pushes history entries to the webview
navigation history, as soon as the user clicks a Link or navigate
is called. The navigate function can also be called with replace
option so that it does not add to the history.

This is using the native goBack method but it seems that the app
only goes back to account summary. In theory this should behave
the same as native browser back. It is not clear why it jumps to
the beginning.

So this could probably be improved, but seems already useful with
just going back to the account summary.

On the account summary it cannot go back any further so it will
trigger the old do-you-want-to-quit alert.

Replaced deprecated onBackPressed override.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -422,7 +422,7 @@ public class MainActivity extends AppCompatActivity {
 
         try {
             String data = readRawText(getAssets().open("web/index.html"));
-            vw.loadDataWithBaseURL(BASE_URL, data, null, null, BASE_URL);
+            vw.loadDataWithBaseURL(BASE_URL, data, null, null, "file:///android_asset/web/index.html");
         } catch (IOException e) {
             e.printStackTrace();
         }
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -371,7 +371,7 @@ public class MainActivity extends AppCompatActivity {
             public void handleOnBackPressed() {
 
                 if (vw.canGoBack()) {
-                    vw.goBack();
+                    vw.goBackOrForward(-1);
                 } else {
                     // To avoid unexpected behaviour, we prompt users and force the app process
                     // to exit which helps with preserving phone's resources by shutting down
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -44,6 +44,7 @@ import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
 import androidx.lifecycle.Observer;
 import androidx.lifecycle.ViewModelProviders;
+import androidx.activity.OnBackPressedCallback;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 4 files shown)


thisconnect:
> A way to unit test this stuff would be very nice, to document the navigation behavior clearly.

To be explicit, you want to test `vw.canGoBack()` and `vw.goBack()` and or the whole thing around it?

I guess that would be very nice. I have no idea what that takes, would it test with a real webview with our production bundled.js and simulating user clicking in react UI and then simulating webview.goBack() and testing if the URI is correct?

Sounds to me more like integration / e2e test, rather than unit test? As far as I see we don't have any java test or maybe I am blind.

Also I am afraid that I have no idea how to setup something like this.

> I also wonder how robust this is - can a change like https://github.com/digitalbitbox/bitbox-wallet-app/pull/2524 change this behavior? Again some sort of automated tests would be very very good to have for this.

I don't see any reason why it shouldn't work, but I will test this branch on top of #2524 .  I believe worst case `vw.canGoBack()` does not return true and then it calls the old code (alert + quitting the app). 

> When/where does a navigation add to the navigation history?

This is a good question. There is `history.pushState` https://developer.mozilla.org/en-US/docs/Web/API/History/pushState to add entries to the browser history. And it looks like our router somehow adds history entries, but I am not sure why it doesn't correctly go back when calling `vw.goBack();`.
Afaik react-router does window.history.pushState, but I will try to check.

thisconnect:
rebased

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -615,39 +657,4 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
...
-    // Note that this is different from users tapping on "home" button or switching
-    // to another app and then back, in which case no extra goroutines are created.
-    //
-    // A proper fix is to make the backend process run in a separate system thread.
-    // Until such solution is implemented, forced app exit seems most appropriate.
-    //
-    // See the following for details about task and activity stacks:
-    // https://developer.android.com/guide/components/activities/tasks-and-back-stack
-    @Override
-    public void onBackPressed() {
```
thisconnect, line 636: > API level 33.
Use [OnBackInvokedCallback](https://developer.android.com/reference/android/window/OnBackInvokedCallback) or androidx.activity.OnBackPressedCallback to handle back navigation instead.


https://developer.android.com/reference/android/app/Activity#onBackPressed()

thisconnect:
> can a change like https://github.com/digitalbitbox/bitbox-wallet-app/pull/2524 change this behavior?

here is a POC with this commit on top of #2524 it seems to behave the same.
https://github.com/digitalbitbox/bitbox-wallet-app/pull/2705

benma:
> > A way to unit test this stuff would be very nice, to document the navigation behavior clearly.
> 
> To be explicit, you want to test `vw.canGoBack()` and `vw.goBack()` and or the whole thing around it?
> 
> I guess that would be very nice. I have no idea what that takes, would it test with a real webview with our production bundled.js and simulating user clicking in react UI and then simulating webview.goBack() and testing if the URI is correct?
> 
> Sounds to me more like integration / e2e test, rather than unit test? As far as I see we don't have any java test or maybe I am blind.
> 
> Also I am afraid that I have no idea how to setup something like this.
> 
> > I also wonder how robust this is - can a change like #2524 change this behavior? Again some sort of automated tests would be very very good to have for this.
> 
> I don't see any reason why it shouldn't work, but I will test this branch on top of #2524 . I believe worst case `vw.canGoBack()` does not return true and then it calls the old code (alert + quitting the app).
> 
> > When/where does a navigation add to the navigation history?
> 
> This is a good question. There is `history.pushState` https://developer.mozilla.org/en-US/docs/Web/API/History/pushState to add entries to the browser history. And it looks like our router somehow adds history entries, but I am not sure why it doesn't correctly go back when calling `vw.goBack();`. Afaik react-router does window.history.pushState, but I will try to check.

The tests don't have to be inside Android/Java, you can assume it works like a Chromium browser. 

It would just be very good if we had a good list of tests (manual tests in the worst case, automated would be best) that show that history-back goes to the expected place. Worse than a back button that closes the app is one that does not work as expected :smile:

thisconnect:
> It would just be very good if we had a good list of tests (manual tests in the worst case, automated would be best) that show that history-back goes to the expected place. Worse than a back button that closes the app is one that does not work as expected 😄

unfortunately it already does not work as expected. I expect that it does history back, but it always just goes back to account summary. I am not sure why, and am not sure how to remote debug right-now.

In webdev history back works as expected.. It would be very nice if both behave the same, need to investigate what exactly is happending in the webview when we call goBack() (or the js version `vw.evaluateJavascript("history.back();", null);`)

thisconnect:
testing if initializing with correct historyUrl changes the goBack behavior.

thisconnect:
closing in favor of https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2819

PR #2703, opened 2024-05-07
frontend: update 'My portfolio' with lightning account

strmci:
This commit adds functionality to display Lightning account
in 'My portfolio' in cases where it is active but not associated
with a connected or remembered wallet, or when all mainnet
accounts in the connected wallet are disabled.

The total coins table is updated to include the Lightning account
Hide chart, percentage change, and filters in my portfolio if
only the lightning account is displayed.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -216,6 +216,7 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/testing", handlers.getTesting).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/account-add", handlers.postAddAccount).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/keystores", handlers.getKeystores).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/keystore-name", handlers.getKeystoreName).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/accounts", handlers.getAccounts).Methods("GET")
 	getAPIRouter(apiRouter)("/accounts/balance", handlers.getAccountsBalance).Methods("GET")
 	getAPIRouter(apiRouter)("/accounts/coins-balance", handlers.getCoinsTotalBalance).Methods("GET")
@@ -603,6 +604,30 @@ func (handlers *Handlers) getKeystores(*http.Request) interface{} {
 	return keystores
 }
 
+func (handlers *Handlers) getKeystoreName(r *http.Request) interface{} {
+	type response struct {
+		Success      bool   `json:"success"`
+		KeystoreName string `json:"keystoreName,omitempty"`
+	}
+
+	rootFingerprint := r.URL.Query().Get("rootFingerprint")
+
+	hexFingerprint, err := hex.DecodeString(rootFingerprint)
+	if err != nil {
+		return response{Success: false}
+	}
+
+	keystore, err := handlers.backend.Config().AccountsConfig().LookupKeystore(jsonp.HexBytes(hexFingerprint))
+	if err != nil {
+		return response{Success: false}
+	}
+
+	return response{
+		Success:      true,
+		KeystoreName: keystore.Name,
+	}
+}
+
 func (handlers *Handlers) getAccounts(*http.Request) interface{} {
 	persistedAccounts := handlers.backend.Config().AccountsConfig()
 
@@ -807,6 +832,21 @@ func (handlers *Handlers) getCoinsTotalBalance(_ *http.Request) (interface{}, er
 		} else {
 			totalCoinsBalances[coinCode] = amount.BigInt()
 			sortedCoins = append(sortedCoins, coinCode)
+		}
+	}
+
+	if handlers.backend.Config().LightningConfig().LightningEnabled() {
+		lightningBalance, err := handlers.backend.Lightning().Balance()
+		if err != nil {
+			return nil, err
+		}
+		availableBalance := lightningBalance.Available().BigInt()
+
+		if bitcoinBalance, exists := totalCoinsBalances[coin.CodeBTC]; exists {
+			bitcoinBalance.Add(bitcoinBalance, availableBalance)
+		} else {
+			totalCoinsBalances[coin.CodeBTC] = availableBalance
+			sortedCoins = append([]coin.Code{coin.CodeBTC}, sortedCoins...)
 		}
 	}
 
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -440,6 +440,18 @@ export const connectKeystore = (code: AccountCode): Promise<{ success: boolean; 
   return apiPost(`account/${code}/connect-keystore`);
 };
 
+export type TResponseKeystoreName = {
+  success: true,
+  keystoreName: string;
+} | {
+  success: false;
+}
+
+export const getKeystoreName = (rootFingerprint: string): Promise<TResponseKeystoreName> => {
+  return apiGet(`keystore-name?rootFingerprint=${rootFingerprint}`);
+};
+
+
 export type TSignMessage = { success: false, aborted?: boolean; errorMessage?: string; } | { success: true; signature: string; }
 
 export type TSignWalletConnectTx = {
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -27,7 +27,7 @@ import { Status } from '@/components/status/status';
 import { GuideWrapper, GuidedContent, Header, Main } from '@/components/layout';
 import { View } from '@/components/view/view';
 import { Chart } from './chart';
-import { SummaryBalance } from './summarybalance';
+import { LightningBalance, SummaryBalance } from './summarybalance';
 import { CoinBalance } from './coinbalance';
 import { AddBuyReceiveOnEmptyBalances } from '@/routes/account/info/buyReceiveCTA';
 import { Entry } from '@/components/guide/entry';
@@ -36,6 +36,8 @@ import { HideAmountsButton } from '@/components/hideamountsbutton/hideamountsbut
 import { AppContext } from '@/contexts/AppContext';
 import { getAccountsByKeystore, isAmbiguousName } from '@/routes/account/utils';
 import { RatesContext } from '@/contexts/RatesContext';
+import { useLightning } from '@/hooks/lightning';
+import { getLightningBalance, subscribeNodeState } from '../../../api/lightning';
 import { ContentWrapper } from '@/components/contentwrapper/contentwrapper';
 import { GlobalBanners } from '@/components/globalbanners/globalbanners';
 
@@ -65,8 +67,78 @@ export const AccountsSummary = ({
   const [accountsTotalBalance, setAccountsTotalBalance] = useState<accountApi.TAccountsTotalBalance>();
   const [coinsTotalBalance, setCoinsTotalBalance] = useState<accountApi.TCoinsTotalBalance>();
   const [balances, setBalances] = useState<Balances>();
+  const { lightningConfig } = useLightning();
+  const [lightningBalance, setLightningBalance] = useState<accountApi.IBalance>();
+  const [lightningKeystoreName, setLightningKeystoreName] = useState('');
 
   const hasCard = useSDCard(devices);
+
+  const fetchLightningBalance = useCallback(async () => {
+    try {
+      const balance = await getLightningBalance();
+      if (mounted.current) {
+        setLightningBalance(balance);
+      }
+    } catch (err) {
+      console.error(err);
+    }
+  }, [mounted]);
+
+  const fetchLightningKeystoreName = useCallback(async () => {
+    if (!lightningConfig.accounts[0]) {
+      setLightningKeystoreName('');
+      return;
+    }
+    try {
+      const response = await accountApi.getKeystoreName(lightningConfig.accounts[0].rootFingerprint);
+      if (mounted.current) {
+        setLightningKeystoreName(response.success ? response.keystoreName : '');
+      }
+    } catch (err) {
+      console.error(err);
+    }
+  }, [lightningConfig.accounts, mounted]);
+
+  const getCoinsTotalBalance = useCallback(async () => {
+    try {
+      const coinBalance = await accountApi.getCoinsTotalBalance();
+      if (!mounted.current) {
+        return;
+      }
+      setCoinsTotalBalance(coinBalance);
+    } catch (err) {
+      console.error(err);
+    }
+  }, [mounted]);
+
+  // if there is an active lightning account subscribe to any node state changes.
+  useEffect(() => {
+    const lightningAccounts = lightningConfig.accounts;
+    if (lightningAccounts.length) {
+      fetchLightningBalance();
+      fetchLightningKeystoreName();
+      const subscriptions = [
+        subscribeNodeState(async () => {
+          await fetchLightningBalance();
+          await getCoinsTotalBalance();
+        })
+      ];
+      return () => unsubscribe(subscriptions);
+    }
+  }, [fetchLightningBalance, getCoinsTotalBalance, fetchLightningKeystoreName, lightningConfig]);
+
+  // lightning account exists but is not from any connected or remembered keystores
+  const hasLightningFromOtherKeystore = (
+    lightningConfig.accounts.length !== 0
+    && (
+      !accountsByKeystore.some(({ keystore }) => {
+        return keystore.rootFingerprint === lightningConfig.accounts[0].rootFingerprint;
+      })
+    )
+  );
+  const allKeystores = (lightningKeystoreName && hasLightningFromOtherKeystore)
+    ? [...accountsByKeystore, { keystore: { name: lightningKeystoreName } }]
+    : accountsByKeystore;
 
   const getAccountSummary = useCallback(async () => {
     // replace previous timer if present
@@ -136,18 +208,6 @@ export const AccountsSummary = ({
     },
     [mounted]
   );
-
-  const getCoinsTotalBalance = useCallback(async () => {
-    try {
-      const coinBalance = await accountApi.getCoinsTotalBalance();
-      if (!mounted.current) {
-        return;
-      }
-      setCoinsTotalBalance(coinBalance);
-    } catch (err) {
-      console.error(err);
-    }
-  }, [mounted]);
 
   const update = useCallback(
     (code: accountApi.AccountCode) => {
@@ -221,23 +281,41 @@ export const AccountsSummary = ({
                 accounts.length && accounts.length <= Object.keys(balances || {}).length ? (
                   <AddBuyReceiveOnEmptyBalances accounts={accounts} balances={balances} />
                 ) : undefined
-              } />
-            {accountsByKeystore.length > 1 && (
+              }
+              hideChartDetails={hasLightningFromOtherKeystore && allKeystores.length == 1}
+            />
+            {allKeystores.length > 1 && (
               <CoinBalance
                 summaryData={summaryData}
                 coinsBalances={coinsTotalBalance}
+              />
+            )}
+            {hasLightningFromOtherKeystore && (
+              <LightningBalance
+                lightningBalance={lightningBalance}
+                lightningAccountKeystoreName={lightningKeystoreName}
+                keystoreDisambiguatorName={
+                  isAmbiguousName(lightningKeystoreName, allKeystores)
+                    ? lightningConfig.accounts[0].rootFingerprint
+                    : undefined
+                }
               />
             )}
             {accountsByKeystore &&
               (accountsByKeystore.map(({ keystore, accounts }) =>
                 <SummaryBalance
                   key={keystore.rootFingerprint}
-                  keystoreDisambiguatorName={isAmbiguousName(keystore.name, accountsByKeystore) ? keystore.rootFingerprint : undefined}
+                  keystoreDisambiguatorName={
+                    isAmbiguousName(keystore.name, allKeystores)
+                      ? keystore.rootFingerprint
+                      : undefined
+                  }
                   accountsKeystore={keystore}
                   accounts={accounts}
                   totalBalancePerCoin={balancePerCoin ? balancePerCoin[keystore.rootFingerprint] : undefined}
                   totalBalance={accountsTotalBalance ? accountsTotalBalance[keystore.rootFingerprint] : undefined}
                   balances={balances}
+                  lightningBalance={ (lightningConfig.accounts.length && lightningConfig.accounts[0].rootFingerprint === keystore.rootFingerprint) ? lightningBalance : undefined}
                 />
               ))}
           </View>
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -26,12 +26,14 @@ import { Filters } from './filters';
 import { getDarkmode } from '@/components/darkmode/darkmode';
 import { DefaultCurrencyRotator } from '@/components/rates/rates';
 import { AppContext } from '@/contexts/AppContext';
+import { TChartFiltersProps } from './types';
 import styles from './chart.module.css';
 
 type TProps = {
   data?: TSummary;
   noDataPlaceholder?: JSX.Element;
   hideAmounts?: boolean;
+  hideChartDetails: boolean;
 };
 
 const defaultData: Readonly<TSummary> = {
@@ -88,7 +90,8 @@ const renderDate = (
 export const Chart = ({
   data = defaultData,
   noDataPlaceholder,
-  hideAmounts = false
+  hideAmounts = false,
+  hideChartDetails = false
 }: TProps) => {
   const height: number = 300;
   const mobileHeight: number = 150;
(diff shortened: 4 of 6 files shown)


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -60,9 +61,28 @@ export function AccountsSummary({ accounts, devices }: TProps) {
...
+      } else {
+        const foundLightning = accountsByKeystore.some((accountsKeystore) => {
+          return accountsKeystore.keystore.rootFingerprint === lightningConfig.accounts[0].rootFingerprint;
+        });
+        return !foundLightning;
+      }
+    } else {
+      return false;
+    }
+  };
```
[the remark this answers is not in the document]
strmci: nice, I just had to change the `rootFingerprint` condition to 
` return keystore.rootFingerprint !== lightningConfig.accounts[0].rootFingerprint;`
and it works fine, thanks

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -213,13 +228,16 @@ export function AccountsSummary({ accounts, devices }: TProps) {
...
+            {showTotalCoins && (
               <CoinBalance
                 accounts={accounts}
                 summaryData={summaryData}
                 coinsBalances={coinsTotalBalance}
               />
             )}
+            {hasLightningFromOtherKeystore && (
+              <LightningBalance/>
+            )}
```
thisconnect, line 240: is my understanding correct, that if the Lightning account is from a connected keystore it will be included together with normal accounts?

strmci: yes, it will be included together with normal accounts

strmci: it will be included with normal accounts also when it is from a remembered keystore

frontends/web/src/routes/account/summary/coinbalance.tsx
```diff
@@ -20,6 +20,8 @@ import { SubTotalCoinRow } from './subtotalrow';
 import { Amount } from '../../../components/amount/amount';
 import { Skeleton } from '../../../components/skeleton/skeleton';
 import style from './accountssummary.module.css';
+import { useLightning } from '../../../hooks/lightning';
+import { CoinCode } from '../../../api/account';
```
thisconnect, line 24: please always import styles at the end.

frontends/web/src/routes/account/summary/coinbalance.tsx
```diff
@@ -50,6 +56,22 @@ export function CoinBalance ({
   const accountsPerCoin = getAccountsPerCoin();
   const coins = Object.keys(accountsPerCoin) as accountApi.CoinCode[];
 
+  let coinsWithLightning: TCoinMap = {} as TCoinMap;
```
thisconnect, line 59: If the intention is to re-assign the variable later use `let` else better use `const`.

frontends/web/src/routes/account/summary/coinbalance.tsx
```diff
@@ -70,21 +92,18 @@ export function CoinBalance ({
...
+                return (
+                  <SubTotalCoinRow
+                    key={coinCode}
+                    coinCode={coinCode as CoinCode}
+                    coinName={coinName}
+                    balance={coinsBalances && coinsBalances[coinCode]}
+                  />
+                );
+              }
+              )) : null}
```
thisconnect, line 106: this looks incorrect:


Screenshot 2024-05-10 at 09 19 18

frontends/web/src/routes/account/summary/coinbalance.tsx
```diff
@@ -70,21 +92,18 @@ export function CoinBalance ({
...
-                      coinCode={account.coinCode}
-                      coinName={account.coinName}
-                      balance={coinsBalances && coinsBalances[coinCode]}
-                    />
-                  );
-                }
-                return null;
-              })) : null}
+            { (Object.keys(coinsWithLightning).length > 0) ? (
+              Object.entries(coinsWithLightning).map(([coinCode, coinName]) => {
```
thisconnect, line 96: When reading this I could understand what the intention was (iterative over coins with at least 1 account or more), but now I find it hard to understand what the intention is.

thisconnect (changes requested):

frontends/web/src/routes/account/summary/coinbalance.tsx
strmci: it is correct, you have Bitcoin (in lightning) and also some bitcoin testnet coins.

but it could be sorted better as:
Bitcoin
Bitcoin Testnet
...

frontends/web/src/routes/account/summary/coinbalance.tsx
strmci: the intention is to iterate over `coinsWithLightning` but only if there are any keys in it so it won't iterate if it is empty

frontends/web/src/routes/account/summary/coinbalance.tsx
strmci: Updated.

frontends/web/src/routes/account/summary/coinbalance.tsx
strmci: I wasn't sure here, updated.

frontends/web/src/routes/account/summary/coinbalance.tsx
```diff
@@ -50,6 +56,22 @@ export function CoinBalance ({
...
+  });
+
+  const { lightningConfig } = useLightning();
+  if (lightningConfig.accounts.length > 0) {
+    const bitcoinCode = 'btc';
+    const bitcoinName = 'Bitcoin';
+    if (!coinsWithLightning[bitcoinCode]) {
+      coinsWithLightning[bitcoinCode] = bitcoinName;
+    }
+  }
```
thisconnect, line 74: coinsWithLightning is confusing me, but maybe it's only the name?

I find it hard to read. It creates a map here, and then later turns the map into Object.keys and Object.entries. We already receave another map in this component coinsBalances and have a helped function to create another map accountsPerCoin.. feel like this all could be simplified, but I am not sure.

Couldn't you just make a `const totals = [];` array instead of the map and push coins/total in there somehow? 

Also we already have `coinsBalances` map 🤔

strmci: Yeah, the name may be confusing.

eg. when you have only a Litecoin account enabled the `accountsPerCoin` will only contain this account, so I created `coinsWithLightning`to add Bitcoin there (when Lightning is active and if `accountsPerCoin` doesn't include it yet)

`coinsBalances` can't be used as they contain just balances, not active accounts. If the amounts are 0 on the accounts, we want to show them. It also doesn't include `coinName`.

strmci: yeah, you are right, it can be simplified to use only `coinsBalances`. I have to include the coin name there.

strmci: We miss the coin name in `coinsBalances`...

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -60,9 +61,23 @@ export function AccountsSummary({ accounts, devices }: TProps) {
...
   const [balances, setBalances] = useState<Balances>();
+  const { lightningConfig } = useLightning();
 
   const hasCard = useSDCard(devices);
 
+  // lightning account exists but is not from any connected or remembered keystores
+  const hasLightningFromOtherKeystore = (
+    lightningConfig.accounts.length !== 0
+    && (
+      accountsByKeystore.length === 0
```
Beerosagos, line 72: I think that you can avoid checking the length, it should be handled by the some condition anyway, right?

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -60,9 +61,23 @@ export function AccountsSummary({ accounts, devices }: TProps) {
...
+      || accountsByKeystore.some(({ keystore }) => {
+        return keystore.rootFingerprint !== lightningConfig.accounts[0].rootFingerprint;
+      })
+    )
+  );
+
+  const showTotalCoins = accountsByKeystore.length > 1 || (hasLightningFromOtherKeystore && accountsByKeystore.length === 1);
```
Beerosagos, line 79: Nit: what about doing smth like

`let keystores = accountsByKeystore.length;
if (hasLightningFromOtherKeystore) {
    keystores += 1;
}`
  
  and check for keystores > 1 below?
  
  I feel like it would be more readable

frontends/web/src/routes/account/summary/coinbalance.tsx
Beerosagos: I agree, this is kind of confusing to me, too. Also, I think that if some info is missing, it should come from the backend in a simple and straight-forward way, as a general approach.

Beerosagos:
Here I found the keystore (c6d..) repeated twice, during my test

image

Beerosagos (changes requested):

frontends/web/src/routes/account/summary/coinbalance.tsx
thisconnect: this OR simplifies the logic and should come first: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2714

thisconnect (changes requested):

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -166,3 +166,51 @@ export function SummaryBalance({
...
+    fetchLightningBalance();
+    const subscriptions = [subscribeNodeState(fetchLightningBalance)];
+    return () => unsubscribe(subscriptions);
+  }, [fetchLightningBalance, lightningConfig]);
+
+  return (
+    <div>
+      <div className={style.accountName}>
+        <p>
+          {lightningConfig.accounts[0].keystoreName} ({lightningConfig.accounts[0].rootFingerprint})
```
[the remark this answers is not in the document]
strmci: please try to disable and enable the lightning wallet on this branch again

frontends/web/src/routes/account/summary/coinbalance.tsx
thisconnect: ok there is some nice stuff in my PR, but the coinName and order of coins/tokens shouldn't be job of the frontend. So I will close the PR, but maybe take some TS improvements and make a separate PR.

frontends/web/src/routes/account/summary/accountssummary.tsx
strmci: Yes, this check is redundant here, removed.

frontends/web/src/routes/account/summary/accountssummary.tsx
strmci: I updated this to use `keystores`

strmci:
@Beerosagos 
I cherry-picked the total coins refactor from master and updated this PR.

> Here I found the keystore (c6d..) repeated twice, during my test
Also, this should be resolved now.

PTAL

thisconnect:
Maybe merge master->staging-ln before this?

Beerosagos:
> There was some refactoring on master, maybe consider merge master->staging-ln before this?

Yes, I agree :+1:

Beerosagos:
merge master > staging-ln PR: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2782

strmci:
@Beerosagos rebased.
PTAL

backend/config/lightning.go
```diff
@@ -26,6 +26,8 @@ type LightningAccountConfig struct {
 	Mnemonic string `json:"mnemonic"`
 	// RootFingerprint is fingerprint of the keystore that generated the entropy.
 	RootFingerprint jsonp.HexBytes `json:"rootFingerprint"`
+	// KeyStoreName is name of the keystore.
```
Beerosagos, line 29: I realized that storing the keystore name could lead to issues when the device name is changed by the user.
I didn't look deeply into this, but I think that the proper way of doing it could be to get the persisted keystore info and retrieve the name there (see `backend.Config().AccountsConfig()`), but we may need to actively persist the current keystore when enabling the LN account).

cc @benma wdyt?

Beerosagos (changes requested):
Left a couple comments.

Also, I noticed that if the only active account is the lightning one, then the chart in the account summary is empty. Not sure what we want to do with this, the easier is probably to hide it, at least for now. wdyt @thisconnect ?

thisconnect:
> Also, I noticed that if the only active account is the lightning one, then the chart in the account summary is empty. Not sure what we want to do with this, the easier is probably to hide it, at least for now. wdyt @thisconnect ?

ack to hiding the chart for now if there is only lightning. But it would be nice in a future PR to have the chart also is there if there is only the lightining account.

strmci:
@Beerosagos 
- as discussed, moved the addition of the lightning balance after the accounts iteration and added Bitcoin to the beginning of `sortedCoins` if it is not already there
- hid the chart in my portfolio if only the lightning account is active
- updated to use `Config().AccountsConfig().LookupKeystore` to get the lightning account keystore name
  
Currently, we use the `isAmbiguousName` function to determine the table title for watch-only accounts. We could apply a similar approach for lightning accounts from other keystores. For now, if there’s an active lightning account from another keystore, the table title will be displayed as "keystoreName (rootFingerprint)".

PTAL

backend/handlers/handlers.go
```diff
@@ -1543,10 +1559,16 @@ func (handlers *Handlers) getLightningConfigHandler(_ *http.Request) interface{}
 	lightningAccounts := handlers.backend.Config().LightningConfig().Accounts
 	accounts := make([]*lightningAccountConfigWithoutMnemonic, len(lightningAccounts))
 	for i, account := range lightningAccounts {
+		keystoreName := ""
+		lightningKeystore, err := handlers.backend.Config().AccountsConfig().LookupKeystore(account.RootFingerprint)
+		if err == nil {
+			keystoreName = lightningKeystore.Name
```
Beerosagos, line 1565: This is an interesting solution, but I think it has a defect: if the keystore device is renamed, the config in the frontend is not updated bc it is served by the provider. That means the keystore name won't be updated until the app is restarted. I think it would be better to add a backend endpoint `keystore-name` that given a rootfingerprint looks for a keystore in the config and returns the name. 

There is also a more latent bug, which depends on the watch-only implementation: when the device is renamed, the keystore name in the config file is not updated until the device is disconnected and reconnected again. But this deserves its own PR as it is out of scope here, imo. 

cc @benma

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -158,3 +158,55 @@ export const SummaryBalance = ({
     </div>
   );
 };
+
+
+export function LightningBalance() {
```
Beerosagos, line 163: There is some code duplication here, I think that it would be nice to move the logic to subscribe and fetch the LN balance inside accountsummary.tsx and pass it to the LightningBalance and SummaryBalance components. wdyt?

Beerosagos (changes requested):

Beerosagos:
Also, I noticed that when the chart is hidden, the `week/month/..` filter buttons are still visible (disabled). They should be hidden too, I think.

Beerosagos:
> Currently, we use the `isAmbiguousName` function to determine the table title for watch-only accounts. We could apply a similar approach for lightning account from other keystore. For now, if there’s an active lightning account from another keystore, the table title will be displayed as "keystoreName (rootFingerprint)".

Yeah, I agree :+1:

strmci:
@Beerosagos
- moved the logic to subscribe and fetch the LN balance inside accountsummary.tsx and pass it to the LightningBalance and SummaryBalance components
- created a new endpoint to get the keystore name from the root fingerprint
- refactored the `isAmbiguousName` function, so it is easier to use with lightning account (already in master: https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/661f819cda87ea5e7087a5b9f746e6f60addbc48)
  and updated how it is used in the account summary
- hid the filter buttons together with the chart and percentage difference when only a lightning account is active

PTAL

frontends/web/src/api/account.ts
```diff
@@ -440,6 +440,15 @@ export const connectKeystore = (code: AccountCode): Promise<{ success: boolean;
   return apiPost(`account/${code}/connect-keystore`);
 };
 
+export type TResponseKeystoreName = null | {
```
Beerosagos, line 443: why `null | `? I think you can remove it

backend/handlers/handlers.go
```diff
@@ -216,6 +216,7 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/testing", handlers.getTesting).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/account-add", handlers.postAddAccount).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/keystores", handlers.getKeystores).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/keystore-name", handlers.getKeystoreName).Methods("POST")
```
Beerosagos, line 219: Why POST method? We usually use POST for calls that change data or that actively do something. In this case, since you are getting data instead of modifying it you should use a GET

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -65,9 +68,61 @@ export const AccountsSummary = ({
...
+  const fetchLightningBalance = useCallback(async () => {
+    setLightningBalance(await getLightningBalance());
+  }, []);
+
+  // if there is an active lightning account derived from this keystore, subscribe to any node state changes.
+  useEffect(() => {
+    const lightningAccounts = lightningConfig.accounts;
+    if (lightningAccounts.length) {
+      fetchLightningBalance();
+      const subscriptions = [subscribeNodeState(fetchLightningBalance)];
```
Beerosagos, line 86: When there is a node state change you should also fetch the coinsTotalBalance again, otherwise you miss the update there

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -65,9 +68,61 @@ export const AccountsSummary = ({
...
+    && (
+      !accountsByKeystore.some(({ keystore }) => {
+        return keystore.rootFingerprint === lightningConfig.accounts[0].rootFingerprint;
+      })
+    )
+  );
+  let keystores = accountsByKeystore.length;
+  let lightningAccountKeystoreName = '';
+  let keystoreDisambiguatorName;
+  if (hasLightningFromOtherKeystore) {
```
Beerosagos, line 109: This whole `if` condition is not very readable, IMO. Can you try to simplify a bit? Maybe it could be useful to return an empty string instead of `null` in the `lightningKeystoreName` useLoad?

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -48,34 +44,25 @@ type TProps = {
   totalBalance?: accountApi.TAccountTotalBalance;
   balances?: Balances;
   keystoreDisambiguatorName?: string;
+  lightningBalance: accountApi.IBalance | undefined;
```
Beerosagos, line 47: @thisconnect can this be simplified like this? same below in `TLightningProps`

```suggestion
  lightningBalance?: accountApi.IBalance;
```

Beerosagos (changes requested):

frontends/web/src/routes/account/summary/summarybalance.tsx
thisconnect: yes but with subtile difference. 

with this: `lightningBalance?: accountApi.IBalance;`  we tell TypeScript that the prop can be omitted.

with `lightningBalance: accountApi.IBalance | undefined;` the prop must be used but the value can be `undefined`

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -65,9 +68,73 @@ export const AccountsSummary = ({
...
+  );
+  let keystores = accountsByKeystore.length;
+  let keystoreDisambiguatorName;
+  if (hasLightningFromOtherKeystore) {
+    keystores += 1;
+    keystoreDisambiguatorName = isAmbiguousName(
+      lightningKeystoreName,
+      lightningKeystoreName
+        ? [...accountsByKeystore, { keystore: { name: lightningKeystoreName } }]
+        : accountsByKeystore
```
Beerosagos, line 132: You are computing this several times: here and for each `SummaryBalance`. You could simplify and make it more efficient by reusing it, right?

frontends/web/src/routes/account/summary/filters.tsx
```diff
@@ -23,12 +23,16 @@ export const Filters = ({
...
+  showFiltersChartPercentageDiff,
   onDisplayWeek,
   onDisplayMonth,
   onDisplayYear,
   onDisplayAll
 }: TChartFiltersProps) => {
   const { t } = useTranslation();
+  if (!showFiltersChartPercentageDiff) {
+    return null;
+  }
```
Beerosagos, line 35: Since this property completely hides the whole component, is it really necessary? I think you could avoid creating it and just decide if including the filters at all in the parent component

Beerosagos (changes requested):
Left some new comments. @strmci, sorry for nitpicking a bit, but this part of the code is already quite messy and I think it is important to keep the changes as simple as possible

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -65,9 +68,73 @@ export const AccountsSummary = ({
...
+      return () => unsubscribe(subscriptions);
+    }
+  }, [fetchLightningBalance, getCoinsTotalBalance, lightningConfig]);
+
+  const lightningKeystoreName = useLoad(() => {
+    const account = lightningConfig.accounts[0];
+    return account
+      ? accountApi.keystoreName(account.rootFingerprint).then(response => response.keystoreName ?? '')
+      : Promise.resolve('');
+  }, [lightningConfig]) || '';
```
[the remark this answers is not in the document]
thisconnect: I agree it is a bit hard to read.

frontends/web/src/api/account.ts
```diff
@@ -440,6 +440,16 @@ export const connectKeystore = (code: AccountCode): Promise<{ success: boolean;
   return apiPost(`account/${code}/connect-keystore`);
 };
 
+export type TResponseKeystoreName = {
+  success: boolean;
+  keystoreName?: string;
+}
+
+export const keystoreName = (rootFingerprint: string): Promise<TResponseKeystoreName> => {
```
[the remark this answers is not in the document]
thisconnect: since you are here, ideally always return Fail | Success, that way keystoreName does not have to be optional since we know it is always a string in case of success. Also see other response types.

strmci:
@Beerosagos I made some updates:

- replaced `useLoad` for getting the lightning keystore name with `fetchLightningKeystoreName`
- removed unnecessary condition from  `getKeystoreName`
- added a check for `mounted.current` before calling `setLightningBalance`
- removed `showFiltersChartPercentageDiff` from `chartFiltersProps` and `filters.tsx` and updated conditions in `chart.tsx`
- reusing the `{ keystore: { name: lightningKeystoreName } } ` object
- comment and formatting updates

PTAL

frontends/web/src/routes/account/summary/accountssummary.tsx
strmci: I see, ok I will update it

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -487,9 +488,10 @@ export const Chart = ({
 
   const hasDifference = difference && Number.isFinite(difference);
   const disableFilters = !hasData || chartDataMissing;
+  const showFiltersChartPercentageDiff: boolean = chartDataMissing || hasData || !!noDataPlaceholder;
```
Beerosagos, line 491: I'm not sure if it wouldn't be better to add a new boolean prop to the Chart object, smth like "hideChartDetails" that we can set if the lightning account is the only one available, and that can be used instead of `showFiltersChartPercentageDiffs`. The reason is that it looks more maintainable to me, as it is clear what it does and why and it's easier to realize when it's not useful anymore and can be removed. But not a strong opinion, we can also leave it as is. cc @thisconnect wdyt?

Beerosagos (changes requested):

frontends/web/src/routes/account/summary/accountssummary.tsx
strmci: I created a new variable `allKeystores` which allowed me to get rid of `keystores` var and simplify the usage of `isAmbiguousName`

frontends/web/src/routes/account/summary/chart.tsx
strmci: @Beerosagos I updated to use `hideChartDetails` as a `Chart` boolean prop.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -516,50 +519,56 @@ export const Chart = ({
...
-          hidden={!toolTipVisible || isMobile}>
-          {toolTipValue !== undefined ? (
-            <span>
-              <h2 className={styles.toolTipValue}>
-                <Amount amount={toolTipValue} unit={chartFiat} />
-                <span className={styles.toolTipUnit}>{chartFiat}</span>
-              </h2>
-              <span className={styles.toolTipTime}>
-                {renderDate(toolTipTime * 1000, i18n.language, source)}
+      {(!hideChartDetails) ? (
```
Beerosagos, line 540: Nit: You could also do `{hideChartDetails || (...)}`

Beerosagos (approved):
tACK! :partying_face: 
(with a last small nit)

frontends/web/src/routes/account/summary/chart.tsx
strmci: Updated

PR #2706, opened 2024-05-09
frontend: convert manage-backups switch to functional component

thisconnect:
This is a container for the manage-backup route and a switch to render backup component for BB01/BB02. This is not used for setup restore-from-backup, but only on the route when the user accesses manage backups from settings.

The class methods were mostly used to render different parts of the component. Moved this to separate components in the same file.

Also changed to named export instead of default export.


frontends/web/src/routes/device/manage-backups/manage-backups.tsx
```diff
@@ -0,0 +1,140 @@
...
+export const ManageBackups = ({
+  deviceID,
+  devices,
+}: TProps) => {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+
+  if (!devices[deviceID]) {
+    navigate('/');
+    return null;
```
NicolaLS, line 43: nit: instead of `return null;` you could `return;` more concise and we seem to do it like in other places where we use `navigate(..)` as well.

frontends/web/src/routes/router.tsx
```diff
@@ -151,6 +151,7 @@ export const AppRouter = ({ devices, deviceIDs, devicesKey, accounts, activeAcco
 
   const ManageBackupsEl = <InjectParams><ManageBackups
     key={devicesKey('manage-backups')}
+    deviceID=""
```
NicolaLS, line 154: this confused me a bit, maybe add a comment that says the `InjectParams` will override/correctly set it? But maybe its not a good idea, to add too many comments.

we could define a `const TO_INJECT = ""` with a comment, and then always use that in such situations? no strong opinion, feel free to ignore this.

NicolaLS (commented):
tested LGTM.

frontends/web/src/routes/device/manage-backups/manage-backups.tsx
```diff
@@ -0,0 +1,140 @@
...
+          deviceID={deviceID}
+          showCreate={true}
+          showRestore={false}
+          showRadio={false}>
+          <BackButton deviceID={deviceID} />
+        </BackupsV2>
+      </SDCardCheck>
+    );
+  default:
+    return;
```
NicolaLS, line 97: nit: this could be `return null` to express that this component will be rendered but nothing is aborted?

frontends/web/src/routes/device/manage-backups/manage-backups.tsx
NicolaLS: I think `return null` makes sense to be more explicit when we want to say "don't render this" e.g. [here](https://github.com/digitalbitbox/bitbox-wallet-app/pull/2706/files#diff-00e572a4a466c40766a05fbb1474a35a0efa21a505387b30425c9130b5f9dbc0R138). But not when _aborting_/_navigating_.

thisconnect: I don't think navigate call is synchronous. If there is no device `!devices[deviceID]` this component should render null.

frontends/web/src/routes/device/manage-backups/manage-backups.tsx
thisconnect: fixed

frontends/web/src/routes/router.tsx
thisconnect: Indeed this is confusing. I changed this to how other route components are initialized (with `deviceID={null})`.

In the future I think we should take a look if we can simplify router.tsx and get rid of `InjectParams`.

thisconnect:
fixed up changes. PTAL also please test once in Qt app.

NicolaLS:
@thisconnect LGTM, tested the AppImage and `webdev` on linux.

shonsirsha (approved):
tested LGTM

PR #2707, opened 2024-05-09
frontend: exposing amount localization hook using context

shonsirsha:
useLocalizedPunctuation may get noisy (log/warnings) if native
locale is empty or invalid. This is due to the hook
being used in many places and thus it runs indpendently
of one another.

Using context API, we can expose this globally and every
consumer will get the same instance of useLocalizedPunctuation.
This reduces the calls of useLocalizedPunctuation which results
in fixing the warning noises.


frontends/web/src/contexts/LocalizationContext.ts
```diff
@@ -0,0 +1,11 @@
...
+
+type DarkModeContextProps = {
+    decimal: string;
+    group: string;
+  }
+
+
+// Create a Context for the localization values
+const LocalizationContext = createContext<DarkModeContextProps>({}as DarkModeContextProps);
+export default LocalizationContext;
```
thisconnect, line 11: DarkModeContextProps -> LocalizationContextProps

nit: I think it's better to prefer named exports if possible

frontends/web/src/contexts/LocalizationProvider.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const LocalizationProvider = ({ children }: TProps) => {
+  const { decimal, group } = useLocalizedPunctuation('');
+
+  return (
+    <LocalizationContext.Provider value={{ decimal, group }}>
+      {children}
+    </LocalizationContext.Provider>
+  );
+};
```
thisconnect, line 17: nit: most of the codebase uses lower-case filenames, but there are already a few files in context that contain uppercase letters... maybe start using lowercase here?

thisconnect (changes requested):

shonsirsha:
Thanks for the quick review! Ready for review now. PTAL @thisconnect

frontends/web/src/components/amount/amount.tsx
```diff
@@ -17,8 +17,8 @@
 import { useContext } from 'react';
 import { AppContext } from '../../contexts/AppContext';
 import { CoinUnit, ConversionUnit } from './../../api/account';
-import { useLocalizedPunctuation } from '../../hooks/localized';
 import style from './amount.module.css';
+import { AmountLocalizationContext } from '../../contexts/amount-localization-context';
```
thisconnect, line 21: how about naming it just LocalizationContext and maybe in the future we move some other Intl, stuff in there?

Here are a few other Intl calles we have and potentially could go int there one day: 
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/365dfd450cedf2f4e47845516f40ad47c4cefd51/frontends/web/src/components/headerssync/headerssync.tsx#L50
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/365dfd450cedf2f4e47845516f40ad47c4cefd51/frontends/web/src/routes/buy/components/countryselect.tsx#L82
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/365dfd450cedf2f4e47845516f40ad47c4cefd51/frontends/web/src/components/transactions/transaction.tsx#L64
- https://github.com/digitalbitbox/bitbox-wallet-app/blob/365dfd450cedf2f4e47845516f40ad47c4cefd51/frontends/web/src/i18n/utils.ts#L39

frontends/web/src/contexts/localization-provider.tsx
```diff
@@ -0,0 +1,17 @@
+import { ReactNode } from 'react';
+import { useLocalizedPunctuation } from '../hooks/localized';
+import { AmountLocalizationContext } from './amount-localization-context';
+
+type TProps = {
+    children: ReactNode;
+  }
+
+export const AmountLocalizationProvider = ({ children }: TProps) => {
+  const { decimal, group } = useLocalizedPunctuation('');
```
thisconnect, line 10: rightnow it is hardcoded to empty string, but this should `nativeLocale` 
```js
  const { nativeLocale } = useContext(AppContext);
```

frontends/web/src/contexts/amount-localization-context.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext } from 'react';
+
+type AmountLocalizationContextProps = {
+    decimal: string;
+    group: string;
+  }
```
thisconnect, line 6: some day we should check again if eslint can do type spacing maybe this https://eslint.style/rules/ts/type-annotation-spacing

thisconnect (changes requested):

frontends/web/src/contexts/localization-provider.tsx
shonsirsha: Oops. Sorry. Good catch.

frontends/web/src/components/amount/amount.tsx
shonsirsha: Yeah, not opposed to naming it Localization.

shonsirsha:
Thanks @thisconnect , PTAL

shonsirsha:
Eh seems like some tests are still failing. Will fix these.

frontends/web/src/hooks/localized.test.ts
```diff
@@ -0,0 +1,162 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { renderHook } from '@testing-library/react';
+import { useLocalizedPunctuation } from './localized';
+import { describe, expect, it } from 'vitest';
+
+
+describe('useLocalizedPunctuation', () => {
```
shonsirsha, line 22: Created this unit test and discarded `amount.localized.test`.

shonsirsha (commented):
@thisconnect , just fixed the tests. PTAL 🙏

thisconnect (approved):
tested 0ec1372c8762f6baa56877d88ddf3271c99cd5c8 LGTM

since, there is 2 console.warn one `frontends/web/src/hooks/localized.ts` and another  `frontends/web/src/utils/localize.ts` how about remove one of them?

Both log the same warning, maybe remove the console.warn in localize.ts wdyt?

PR #2709, opened 2024-05-09
frontend: convert connected.tsx component

NicolaLS:
Convert the class component in frontends/web/src/connected.tsx into a functional component.

Tested on `webdev` (normal operation and when killing the backend) and on AppImage (only normal operation, I don't know how to kill the backend when using the AppImage).

### DIFF
--- a/frontends/web/src/connected.tsx
+++ b/frontends/web/src/connected.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


frontends/web/src/connected.tsx
```diff
@@ -14,46 +14,28 @@
...
-  }
+export const ConnectedApp = ({ children }: TProps) => {
+  const [connected, setConnected] = useState(true);
 
-  public componentWillUnmount() {
-    this.unsubscribe();
-  }
+  useEffect(() => {
+    return backendConnected(c => {
+      setConnected(c);
```
thisconnect, line 29: c -> connected

frontends/web/src/connected.tsx
```diff
@@ -14,46 +14,28 @@
...
-        <div className="app" style={{ padding: 40 }}>
-                    The WebSocket closed. Please restart the backend and reload this page.
-        </div>
-      );
-    }
+  if (!connected) {
     return (
-      <div>{children}</div>
+      <div className="app" style={{ padding: 40 }}>
+                The WebSocket closed. Please restart the backend and reload this page.
```
thisconnect, line 36: since you are here, please fix the indentation

frontends/web/src/connected.tsx
```diff
@@ -14,46 +14,28 @@
  * limitations under the License.
  */
```
thisconnect, line 15: please add `Copyright 2024 Shift Crypto AG`

thisconnect (changes requested):

frontends/web/src/connected.tsx
```diff
@@ -14,46 +15,28 @@
...
+  const [connected, setConnected] = useState(true);
 
-  public componentWillUnmount() {
-    this.unsubscribe();
-  }
+  useEffect(() => {
+    return backendConnected(connected => {
+      setConnected(connected);
+    });
+  }, []);
```
thisconnect, line 32: this is shorthard and returns the unsibscibe function directly. In most other places we store it in a variable first `const unsubscribe = ` and the `return unsubscribe;`.

Note this could even be written shorted as: `useEffect(() => backendConnected(setConnected));` but it is not very obvious that there is an unsubscribe.

I personally think it's ok like this. cc @Beerosagos any preference?

thisconnect (approved):
thanks LGTM, added question for beersagos

frontends/web/src/connected.tsx
NicolaLS: I think easy to read code is important but with react effect hooks you kind of need to know how they work even to understand the more verbose version, so I think the even shorter version you suggested is best (no strong opinion).

nit: should have `[]` dependency array, otherwise it will re-run every time the component is re-rendered which is not really necessary (but would also work)

thisconnect: Note: I don't want to propose shorter version, but rather get feedback if we should use longer version for readability:

```js
useEffect(() => {
  const unsubscribe = backendConnected(connected => {
    setConnected(connected);
  });
  return unsubscribe;
}, []);
```

Beerosagos: > I personally think it's ok like this. cc @Beerosagos any preference?

No strong opinion, either. Since there is the `return` statement I think it is readable enough :+1:

PR #2710, opened 2024-05-10
frontend: Maintain chart timeframe when rotating currencies

strmci:
Previously, the chart would reset to 'All' whenever the currency was changed. This commit ensures that the selected timeframe remains consistent when rotating currencies.


shonsirsha (changes requested):
This did not work for me. It always resets to the initial state `all`... 🤔  which makes sense to me, as after changing the chart display option (e.g `displayWeek`), we didnt save that state anywhere where it won't get reset. 

We could probably save this state in a context.

shonsirsha (approved):
LGTM. Thought it was for when changing default currencies in the settings, but rather when user rotates the currency in the account summary page / chart.tsx itself. Thanks!

PR #2711, opened 2024-05-13
backend/bitbox02bootloader: allow fixing broken install

benma:
Problem: a new device ships with the latest firmware monotonic version. User
sees 'Install' because the device is erased (no firmware on it), but during
first install of firmware, unplugs. Upon replug, the device will show 'Invalid
firmware' and boot into bootloader again. Since the monotonic version of the
firmware to be installed is the same as the version stored on the device, and
the device is not erased (half of the firmware is on it), the user sees no
button to upgrade/instll anymore.

This commit fixes this by still showing the upgrade button if the versions are
the same, but the device firmware hash does not match the expected firmware
hash.

Replaces https://github.com/digitalbitbox/bitbox-wallet-app/pull/2700 - @thisconnect suggested to use the firmware hash instead, which is a much nicher solution as it does not show the upgrade button to every user who enters the bootloader, just to the ones where the firmware install is actually broken - thanks @thisconnect :pray:

### DIFF
--- a/backend/devices/bitbox02bootloader/device.go
+++ b/backend/devices/bitbox02bootloader/device.go
@@ -17,6 +17,8 @@
 package bitbox02bootloader
 
 import (
+	"bytes"
+	"encoding/hex"
 	"fmt"
 	"sync"
 
@@ -240,6 +242,11 @@ func (device *Device) VersionInfo() (*VersionInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+	currentFirmwareHash, _, err := device.Device.GetHashes(false, false)
+	if err != nil {
+		return nil, err
+	}
+
 	latestFw, err := bundledFirmware(device.Device.Product())
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


benma:
@Beerosagos for efficiency I already merged the https://github.com/digitalbitbox/bitbox02-api-go/pull/93, exposing the two functions needed here. PTAL anyway and I can add fixes in a new PR should you find any issues there.

benma (commented):
The 'Take me back' button could/should also be hidden if the install is broken, as the device would not be able to boot anyway.

benma:
> The 'Take me back' button could/should also be hidden if the install is broken, as the device would not be able to boot anyway.

On 2nd thought I'll skip this, as it's a bunch of code overhead and only works in this particular edge case - if any regular update is interrupted the button would still be shown as before, and we cannot fix this easily. While not great, clicking the button would still show "Invalid firmware" on the device and the user can fix it by installing again.

Beerosagos:
> @Beerosagos for efficiency I already merged the [digitalbitbox/bitbox02-api-go#93](https://github.com/digitalbitbox/bitbox02-api-go/pull/93), exposing the two functions needed here. PTAL anyway and I can add fixes in a new PR should you find any issues there.

the PR LGTM! I only found a tiny nit in a comment (see https://github.com/digitalbitbox/bitbox02-api-go/pull/93/commits/c5e4f77d941fafbe2b133a756e915dae3b3379c8#r1601100822). Probably it doesn't even worth fixing it..

benma:
@Beerosagos did you mean this? https://github.com/digitalbitbox/bitbox02-api-go/pull/94

Beerosagos:
> @Beerosagos did you mean this? [digitalbitbox/bitbox02-api-go#94](https://github.com/digitalbitbox/bitbox02-api-go/pull/94)

Yup! Thanks

Beerosagos (approved):
tACK!

benma:
Rebased

PR #2720, opened 2024-05-20
frontend: persist chart timeframe

shonsirsha:
Persisting chart's timeframe using the AppContext so the timeframe doesn't reset when we navigate through the app.

### DIFF
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -23,8 +23,9 @@ import { Amount } from '../../../components/amount/amount';
 import { PercentageDiff } from './percentage-diff';
 import { Filters } from './filters';
 import { getDarkmode } from '../../../components/darkmode/darkmode';
-import { TChartDisplay, TChartFiltersProps } from './types';
+import { TChartFiltersProps } from './types';
 import { DefaultCurrencyRotator } from '../../../components/rates/rates';
+import { AppContext } from '../../../contexts/AppContext';
 import styles from './chart.module.css';
 
 export interface FormattedLineData extends LineData {
@@ -40,7 +41,6 @@ type ChartProps = {
 };
 
 type State = {
-  display: TChartDisplay;
   source: 'daily' | 'hourly';
   difference?: number;
   diffSince?: string;
@@ -58,6 +58,8 @@ type FormattedData = {
   [key: number]: string;
 }
 class Chart extends Component<Props, State> {
+  static contextType = AppContext;
+  context!: React.ContextType<typeof AppContext>;
   private ref = createRef<HTMLDivElement>();
   private refToolTip = createRef<HTMLSpanElement>();
   private chart?: IChartApi;
@@ -81,7 +83,6 @@ class Chart extends Component<Props, State> {
   };
 
   public readonly state: State = {
-    display: 'all',
     source: 'daily',
     toolTipVisible: false,
     toolTipValue: undefined,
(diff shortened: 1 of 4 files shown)


strmci (dismissed):
I tested it, and it doesn't work for me either while rotating currencies or navigating to accounts/settings and back to My portfolio. The filter shows the correct timeframe, but the chart resets to `all`. I managed to fix it by keeping the `switch` in the `createChart` and changing it to `switch (this.context.chartDisplay)`, but maybe it can be fixed in some other way, too. PTAL

shonsirsha:
@strmci , Oops! Totally correct and thanks for testing. I tried removing it to test the behaviour and never added it back. It's back now. PTAL 🙏

strmci:
> @strmci , Oops! Totally correct and thanks for testing. I tried removing it to test the behaviour and never added it back. It's back now. PTAL 🙏

thanks, tested, LGTM, requesting review also from @thisconnect

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -335,8 +337,9 @@ class Chart extends Component<Props, State> {
       this.chart.applyOptions({ timeScale: { timeVisible: false } });
     }
     this.setState(
-      { display: 'month', source: 'daily' },
+      { source: 'daily' },
       () => {
+        this.context.setChartDisplay('month');
         if (!this.chart) {
```
thisconnect, line 343: Before this setstate callback ensured that both states were updated.

Afaik context function `setChartDisplay` is not synchronously. So when this callback runs the context not updated.

Now it seems that the intention is to fist update the state, and wait for context update.

Would it work if you set context before state?

That seems closer to how it was before, and React may have the chance to schedule the context and state update together?

Also it doesn't look like it should wait until state is updated before changing context. wdyt?

thisconnect (changes requested):

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -23,9 +23,10 @@ import { Amount } from '../../../components/amount/amount';
 import { PercentageDiff } from './percentage-diff';
 import Filters from './filters';
 import { getDarkmode } from '../../../components/darkmode/darkmode';
-import { TChartDisplay, TChartFiltersProps } from './types';
+import { TChartFiltersProps } from './types';
```
thisconnect, line 26: since `TChartDisplay` isn't used in Chart component, maybe move the type to appprovider?

frontends/web/src/routes/account/summary/chart.tsx
shonsirsha: Yeah the intention here is to actually only set the `setChartDisplay` after the `source` state has been changed.

I think setting context before state would also work, as you said it's closer to the previous behaviour (async, these two states not dependant on each other).

thisconnect: Thinking about that we will at some point chagne this from class->functional I prefer if the context was set before the state.
Else it will look like it is intentional to wait for state change and then update context.

please change to set context first 🙏

shonsirsha: > please change to set context first

Yes, I'm changing it to setting the context state first. :)

shonsirsha:
Rebased. Changes made. Thanks @thisconnect , PTAL 🙏

thisconnect (approved):
tested LGTM thanks 👍

PR #2721, opened 2024-05-20
frontend: convert InlineMessage to functional and use TS

shonsirsha:


frontends/web/src/components/inlineMessage/inlineMessage.tsx
```diff
@@ -14,25 +15,30 @@
...
 
-export default class InlineMessage extends Component {
-  deactivate = () => {
-    this.props.onEnd();
+type TProps = {
+    type: string;
+    message: string;
+    align: string;
+    onEnd: () => void;
+}
```
thisconnect, line 25: I wanted to ask to also type type ('success') and align 'left' props, but this component is only used in tor proxy settings. It is probably not really worth it. I think when we update tor-proxy config we might even drop this component in favor or Message or Status component.

thisconnect (approved):
untested LGTM thanks

PR #2724, opened 2024-05-20
frontend/coin-control: add address reuse warning

strmci:
Add a warning to the coin control modal when a user selects one or more UTXOs from reused addresses. The warning: "One or more UTXOs have a reused address. Be aware  the receiver will see all UTXOs associated with those addresses."

The UTXOs now have red badge when their address was re-used.

### DIFF
--- a/frontends/web/src/routes/account/send/utxos.tsx
+++ b/frontends/web/src/routes/account/send/utxos.tsx
@@ -32,6 +32,8 @@ import { ExternalLink } from '../../../components/icon';
 import { Amount } from '../../../components/amount/amount';
 import { FiatConversion } from '../../../components/rates/rates';
 import { getScriptName } from '../utils';
+import { Message } from '../../../components/message/message';
+import { Badge } from '../../../components/badge/badge';
 import style from './utxos.module.css';
 
 export type TSelectedUTXOs = {
@@ -56,6 +58,7 @@ export const UTXOs = ({
   const { t } = useTranslation();
   const [utxos, setUtxos] = useState<TUTXO[]>([]);
   const [selectedUTXOs, setSelectedUTXOs] = useState<TSelectedUTXOs>({});
+  const [reusedAddressUTXOs, setReusedAddressUTXOs] = useState(0);
 
   useEffect(() => {
     getUTXOs(accountCode).then(setUtxos);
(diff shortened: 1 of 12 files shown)


Beerosagos (changes requested):
Nice! Just noticed a small bug.

I also feel like the duplicated address should be somehow highlighted. Otherwise it can be hard to find those out.

strmci:
@Beerosagos @thisconnect 

- fixed the repeated warning for each UTXO's script type
- the warning is shown only when the user selects the UTXO from the reused address
- I changed the warning from `Status` to `Message` so it is not all red
- added a badge to all UTXOs from re-used addresses
- CHANGELOG updated

image

PTAL

frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -56,6 +58,7 @@ export const UTXOs = ({
   const { t } = useTranslation();
   const [utxos, setUtxos] = useState<TUTXO[]>([]);
   const [selectedUTXOs, setSelectedUTXOs] = useState<TSelectedUTXOs>({});
+  const [showAddressReusedMessage, setShowAddressReusedMessage] = useState(0);
```
Beerosagos, line 61: nit: I find the name of this const slightly misleading. Maybe smth like `selectedReusedAddresses`? Since it is a number and not a bool.

Beerosagos (approved):
LGTM with a couple nits :+1: 

Nit: maybe we can vertically center better the badge?

image

thisconnect:
> Nit: maybe we can vertically center better the badge?

I'll take a look, ideally badge would be nice and inline with text. Not sure yet what causes this.

frontends/web/src/routes/account/send/utxos.tsx
strmci: Updated to `reusedAddressUTXOs`.

thisconnect:
> I'll take a look, ideally badge would be nice and inline with text. Not sure yet what causes this.

tl;dr it was a combination of inconsitent line-heights (parent element) and vertical centering. 

branch with fix
https://github.com/thisconnect/bitbox-wallet-app/tree/frontend-fix-badge-centering

diff
https://github.com/BitBoxSwiss/bitbox-wallet-app/compare/master...thisconnect:frontend-fix-badge-centering?expand=1

@strmci @Beerosagos let me know if you want to merge this PR first and if I should open a followup PR or feel free to chery-pick my commit into this change.

strmci:
@thisconnect
I cherry-picked your Badge improvement, thanks.

PR #2725, opened 2024-05-22
frontend: refactor passphrase.tsx

NicolaLS:
This is a followup on #2713 that refactors the component to be more concise, and easier to understand/maintain.

- Set passphraseEnabled to decide whether to display the disable or the enable wizard.
- When one wizard is rendered it knows the status of the passphraseEnabled. Also the disable wizard does not need to know about the steps. The enable wizard knows where to start end stop and can manage that state on its own.
- SetPassphrase is more explicit and less error prone than toggling it.
- Remove unnecessary API request after receiving a successful response after setting the passphrase on the device.
- Remove some code duplication when rendering steps

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/passphrase.tsx
+++ b/frontends/web/src/routes/device/bitbox02/passphrase.tsx
@@ -27,12 +27,9 @@ import { View, ViewButtons, ViewContent, ViewHeader } from '../../../components/
 import { PointToBitBox02 } from '../../../components/icon';
 import { Status } from '../../../components/status/status';
 
-// enabling has 6 dialogs with information
-const INFO_STEPS_ENABLE = 5;
-
-// disabling passphrase shows only 1 info dialog
-const INFO_STEPS_DISABLE = 0;
-
+// The enable wizard has five steps that can be navigated by clicking
+// 'back' or 'continue'. On the last step the passphrase will be enabled.
+const FINAL_INFO_STEP = 5;
 const CONTENT_MIN_HEIGHT = '38em';
 
 type TProps = {
@@ -44,9 +41,9 @@ type TStatus = 'info' | 'progress' | 'success';
 export const Passphrase = ({ deviceID }: TProps) => {
   const { t } = useTranslation();
   const navigate = useNavigate();
-  const [infoStep, setInfoStep] = useState(0);
+
   const [status, setStatus] = useState<TStatus>('info');
-  const [passphraseEnabled, setPassphraseEnabled] = useState<boolean>();
+  const [isEnabled, setIsEnabled] = useState<boolean>();
 
   useEffect(() => {
     getDeviceInfo(deviceID).then(result => {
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -184,147 +147,58 @@ export const Passphrase = ({ deviceID }: TProps) => {
...
+  type TStepData = { titleKey: string; messageKey: string; buttonTextKey: string; }[]
+
+  const stepData: TStepData = [
+    { titleKey: 'passphrase.intro.title', messageKey: 'passphrase.intro.message', buttonTextKey: 'passphrase.what.button' },
+    { titleKey: 'passphrase.what.title', messageKey: 'passphrase.what.message', buttonTextKey: 'passphrase.why.button' },
+    { titleKey: 'passphrase.why.title', messageKey: 'passphrase.why.message', buttonTextKey: 'passphrase.considerations.button' },
+    { titleKey: 'passphrase.considerations.title', messageKey: 'passphrase.considerations.message', buttonTextKey: 'passphrase.how.button' },
+    { titleKey: 'passphrase.how.title', messageKey: 'passphrase.how.message', buttonTextKey: 'passphrase.summary.button' },
+    { titleKey: 'passphrase.summary.title', messageKey: 'passphrase.summary.understandList', buttonTextKey: 'passphrase.enable' },
+  ];
```
thisconnect, line 186: In the rest of the codebase translated strings are consistently always in the `t()` function, this makes it easy to do global search example: `t('passphrase.summary.understandList')`.

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -184,147 +147,58 @@ export const Passphrase = ({ deviceID }: TProps) => {
...
+
+  const renderSteps = (idx: number) => {
+    return (<View
```
thisconnect, line 189: We don't do it everywhere consistently but for readability 
```suggestion
    return (
      <View
```

NicolaLS:
I used `t(..)` in the object directly, and reformatted the `renderSteps` return. PTAL @thisconnect :)

thisconnect (changes requested):

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -56,77 +50,46 @@ export const Passphrase = ({ deviceID }: TProps) => {
...
-      // each page has a continue button that jumps to the next or finally toggles passphrase
-      // infoStep counts down in decreasing order
       setPassphraseEnabled(mnemonicPassphraseEnabled);
-      setInfoStep(mnemonicPassphraseEnabled ? INFO_STEPS_DISABLE : INFO_STEPS_ENABLE);
     }).catch(console.error);
-  }, [deviceID, setPassphraseEnabled, setInfoStep, t]);
+  }, [deviceID, t]);
 
-  const togglePassphrase = async () => {
+  const setPassphrase = async (enabled: boolean) => {
```
thisconnect, line 57: just nit: it all makes sense but we have now `setPassphrase` and `setPassphraseEnabled` with very similar naming this is slightly confusing.

Not sure maybe change the state to `isEnabled`/`setIsEnabled` or something better?

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -27,12 +27,6 @@ import { View, ViewButtons, ViewContent, ViewHeader } from '../../../components/
 import { PointToBitBox02 } from '../../../components/icon';
 import { Status } from '../../../components/status/status';
 
-// enabling has 6 dialogs with information
-const INFO_STEPS_ENABLE = 5;
```
thisconnect, line 31: Nit: We had this number in one place with a comment what it is all about.

Now this is in 4 different places in the component without any comment (or I missed it somehow), do you prefer having it this way or would you mind adding something similar back with a useful comment?

thisconnect (changes requested):
lgtm, with 2 nits/questions

frontends/web/src/routes/device/bitbox02/passphrase.tsx
NicolaLS: imo. the comment(s) were necessary because the logic was so complicated/intertwined. now it is so straight forward that a developer working on this code will understand what is happening without comment imo.
- the  infoStep is only relevant for the enable wizard now
- the info step starts at 0 , if you press back it means you abort
- when the infoStep reaches 5 it will set the passphrase on continue because its the last step

this is all really obvious from the code and does not need comments imo. but let me know if you disagree, ofc. I can add some.

frontends/web/src/routes/device/bitbox02/passphrase.tsx
NicolaLS: (good point, pushed)

frontends/web/src/routes/device/bitbox02/passphrase.tsx
thisconnect: I agree, however I don't think it is ideal to have the number 5 hardcoded in 4 places. Think about if we need to add or remove one more pages.

but just a nit. It is ok for me

frontends/web/src/routes/device/bitbox02/passphrase.tsx
thisconnect: thanks

frontends/web/src/routes/device/bitbox02/passphrase.tsx
NicolaLS: totally agree, should I add a `const FINAL_INFO_STEP = 5` ?

thisconnect (approved):
tested 6a1526e73e0f064aa4c3b7abb3f33bba9b5b5bd6 LGTM thanks

PR #2726, opened 2024-05-23
backend/accounts_test: fix testAccountsByKeystore occasional error

Beerosagos:
testAccountsByKeystore test was occasionally failing the check on the number of loaded accounts for each keystore
(see https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2716#issuecomment-2114690746)

This fix excludes the `HiddenBecauseUnused` accounts from the check and should fix the issue.

### DIFF
--- a/backend/accounts_test.go
+++ b/backend/accounts_test.go
@@ -47,16 +47,20 @@ const (
 	regtestDisabled = false
 )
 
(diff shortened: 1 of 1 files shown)


benma (approved):
LGTM

PR #2727, opened 2024-05-27
frontend: create dmg installer in osx build

sutterseba:
Installing application with `.dmg` installers is a familiar and intuitive process for macOS users, ensuring applications are installed and updated correctly.

Using [create-dmg](https://github.com/create-dmg/create-dmg) we can create a nice looking `.dmg` installer with custom icons and background. 

The Makefile is adjusted with a new osx-create-dmg target, which can be run manually after the BitBox.app has been built and notarized.

Preview:
Screenshot 2024-05-27 at 09 11 10


benma (changes requested):
Very nice! I'll try it out.

Please update the docs/BUILD.md file to change the instructions. I think instead of 

    /usr/bin/ditto -c -k --keepParent BitBox.app BitBox.zip

it should be the DMG-creation, and then the DMG can be submitted for notarization instead of the zip file. I'll test that though.

scripts/osx-create-dmg.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+APP="frontends/qt/build/osx/BitBox.app"
+INSTALLER="frontends/qt/build/osx/BitBoxApp Installer.dmg"
+
+if ! type create-dmg > /dev/null; then
+    echo "create-dmg not available. Installing with Homebrew."
+    brew install create-dmg
+fi
```
benma, line 9: Imo it's not very nice to mess with someone's system without prompt/confirmation. I prefer to exit with an error and instruct the user to run the command to fix it.

scripts/osx-create-dmg.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+APP="frontends/qt/build/osx/BitBox.app"
+INSTALLER="frontends/qt/build/osx/BitBoxApp Installer.dmg"
```
benma, line 4: Please remove the space in the filename as spaces are a bit annoying to deal with on the terminal. Before release the file will be manually renamed containing the version anyway.

benma (approved):
tACK, thanks a lot! Everything seems to work fine.

PR #2728, opened 2024-05-27
frontend: convert chart.tsx component

NicolaLS:
Convert the Chart class component in chart.tsx to a functional component.

There some general improvements that could be made, but this PR aims to perform minimal changes to only transform the component from a class component to a functional one. I will open another PR with a general refactor when this one was merged.

note: I moved the `renderDate` function out of the component, but this improvement should've been in the next PR not this one, lmk if I should put it back.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
+import type { LineData } from 'lightweight-charts';
 import { apiGet, apiPost } from '@/utils/request';
-import type { ChartData } from '@/routes/account/summary/chart';
 import type { TDetailStatus } from './bitsurance';
 import type { SuccessResponse } from './response';
 import { Slip24 } from 'request-address';
@@ -185,6 +185,12 @@ export const getInfo = (code: AccountCode) => {
 export const init = (code: AccountCode): Promise<null> => {
   return apiPost(`account/${code}/init`);
 };
+
+export type FormattedLineData = LineData & {
+  formattedValue: string;
+};
+
+export type ChartData = FormattedLineData[];
 
 export type TSummaryResponse = {
     success: true;
--- a/frontends/web/src/routes/account/summary/chart.tsx
+++ b/frontends/web/src/routes/account/summary/chart.tsx
@@ -14,282 +14,186 @@
  * limitations under the License.
  */
 
+import { useCallback, useContext, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import { createChart, IChartApi, LineData, LineStyle, LogicalRange, ISeriesApi, UTCTimestamp, MouseEventParams, ColorType, Time } from 'lightweight-charts';
-import { Component, createRef, ReactChild } from 'react';
-import { TSummary } from '@/api/account';
-import { translate, TranslateProps } from '@/decorators/translate';
+import type { TSummary, ChartData } from '@/api/account';
+import { usePrevious } from '@/hooks/previous';
 import { Skeleton } from '@/components/skeleton/skeleton';
 import { Amount } from '@/components/amount/amount';
 import { PercentageDiff } from './percentage-diff';
 import { Filters } from './filters';
 import { getDarkmode } from '@/components/darkmode/darkmode';
-import { TChartFiltersProps } from './types';
 import { DefaultCurrencyRotator } from '@/components/rates/rates';
 import { AppContext } from '@/contexts/AppContext';
 import styles from './chart.module.css';
 
-export interface FormattedLineData extends LineData {
-  formattedValue: string;
-}
-
-export type ChartData = FormattedLineData[];
-
-type ChartProps = {
-  data: TSummary;
-  noDataPlaceholder?: ReactChild;
-  hideAmounts: boolean;
+type TProps = {
+  data?: TSummary;
+  noDataPlaceholder?: JSX.Element;
+  hideAmounts?: boolean;
 };
 
-type State = {
-  source: 'daily' | 'hourly';
-  difference?: number;
-  diffSince?: string;
-  toolTipVisible: boolean;
-  toolTipValue?: string;
-  toolTipTop: number;
-  toolTipLeft: number;
-  toolTipTime: number;
-  isMobile: boolean;
-}
-
-type Props = ChartProps & TranslateProps;
+const defaultData: Readonly<TSummary> = {
+  chartDataMissing: true,
+  chartDataDaily: [],
+  chartDataHourly: [],
+  chartFiat: 'USD',
+  chartTotal: null,
+  formattedChartTotal: null,
+  chartIsUpToDate: false,
+  lastTimestamp: 0,
+};
 
 type FormattedData = {
   [key: number]: string;
 }
-class Chart extends Component<Props, State> {
-  static contextType = AppContext;
-  context!: React.ContextType<typeof AppContext>;
-  private ref = createRef<HTMLDivElement>();
-  private refToolTip = createRef<HTMLSpanElement>();
-  private chart?: IChartApi;
-  private lineSeries?: ISeriesApi<'Area'>;
-  private height: number = 300;
-  private mobileHeight: number = 150;
-  private formattedData?: FormattedData;
 
-  static readonly defaultProps: ChartProps = {
-    data: {
-      chartDataMissing: true,
-      chartDataDaily: [],
-      chartDataHourly: [],
-      chartFiat: 'USD',
-      chartTotal: null,
-      formattedChartTotal: null,
-      chartIsUpToDate: false,
-      lastTimestamp: 0,
-    },
-    hideAmounts: false,
+const getUTCRange = () => {
+  const now = new Date();
+  const utcYear = now.getUTCFullYear();
+  const utcMonth = now.getUTCMonth();
+  const utcDate = now.getUTCDate();
+  const utcHours = now.getUTCHours();
+  const to = new Date(Date.UTC(utcYear, utcMonth, utcDate, utcHours, 0, 0, 0));
+  const from = new Date(Date.UTC(utcYear, utcMonth, utcDate, utcHours, 0, 0, 0));
+  return {
+    utcYear,
+    utcMonth,
+    utcDate,
+    to,
+    from,
   };
+};
 
-  public readonly state: State = {
-    source: 'daily',
+const renderDate = (
+  date: number,
+  lang: string,
+  src: string
+) => {
+  return new Date(date).toLocaleString(
+    lang,
+    {
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+      ...(src === 'hourly' ? {
+        hour: '2-digit',
+        minute: '2-digit',
+      } : null)
+    }
+  );
+};
+
+export const Chart = ({
+  data = defaultData,
+  noDataPlaceholder,
+  hideAmounts = false
+}: TProps) => {
+  const height: number = 300;
+  const mobileHeight: number = 150;
+  const hasData = data.chartDataDaily && data.chartDataDaily.length > 0;
+  const hasHourlyData = data.chartDataHourly && data.chartDataHourly.length > 0;
+
+  const { t, i18n } = useTranslation();
+  const { chartDisplay, setChartDisplay } = useContext(AppContext);
+
+  const ref = useRef<HTMLDivElement>(null);
+  const refToolTip = useRef<HTMLSpanElement>(null);
+  const chart = useRef<IChartApi>();
+  const chartInitialized = useRef(false);
+  const lineSeries = useRef<ISeriesApi<'Area'>>();
+  const formattedData = useRef<FormattedData>({});
+
+  const [source, setSource] = useState<'daily' | 'hourly'>(chartDisplay === 'week' ? 'hourly' : 'daily');
+  const [difference, setDifference] = useState<number>();
+  const [diffSince, setDiffSince] = useState<string>();
+  const [isMobile, setIsMobile] = useState(window.innerWidth <= 640);
+  const [tooltipData, setTooltipData] = useState<{
+    toolTipVisible: boolean;
+    toolTipValue?: string;
+    toolTipTop: number;
+    toolTipLeft: number;
+    toolTipTime: number;
+  }>({
     toolTipVisible: false,
-    toolTipValue: undefined,
     toolTipTop: 0,
     toolTipLeft: 0,
     toolTipTime: 0,
-    isMobile: false,
-  };
+  });
 
-  public componentDidMount() {
-    if (this.state.source !== 'hourly' && this.context.chartDisplay === 'week') {
-      this.setState({ source: 'hourly' });
-    }
-    this.createChart();
-  }
+  const prevChartDataDaily = usePrevious(data.chartDataDaily);
+  const prevChartDataHourly = usePrevious(data.chartDataHourly);
+  const prevChartFiat = usePrevious(data.chartFiat);
+  const prevHideAmounts = usePrevious(hideAmounts);
 
-  public componentWillUnmount() {
-    window.removeEventListener('resize', this.onResize);
-    if (this.chart) {
-      this.chart.timeScale().unsubscribeVisibleLogicalRangeChange(this.calculateChange);
-      this.chart.unsubscribeCrosshairMove(this.handleCrosshair);
-    }
-  }
+  const setFormattedData = (chartData: ChartData) => {
+    formattedData.current = {};
 
-  public componentDidUpdate(prev: Props) {
-    const { chartDataDaily, chartDataHourly } = this.props.data;
-    if (!this.chart) {
-      this.createChart();
-    }
-    if (
-      (this.lineSeries && prev.data.chartDataDaily && prev.data.chartDataHourly && chartDataDaily && chartDataHourly)
-        && (
-          prev.data.chartDataDaily.length !== chartDataDaily.length
-          || prev.data.chartDataHourly.length !== chartDataHourly.length
-        )
-    ) {
-      const data = this.state.source === 'hourly' ? chartDataHourly : chartDataDaily;
-      this.lineSeries.setData(data);
-      this.chart?.timeScale().fitContent();
-      this.setFormattedData(data);
-    }
-
-    if (prev.hideAmounts !== this.props.hideAmounts) {
-      this.chart?.applyOptions({
-        leftPriceScale: {
-          visible: this.props.hideAmounts ? false : !this.state.isMobile,
-        }
-      });
-    }
-
-    if (this.props.data.chartFiat !== prev.data.chartFiat) {
-      this.reinitializeChart();
-    }
-    if (!this.hasData() && this.props.data.chartIsUpToDate) {
-      this.removeChart();
-    }
-  }
-
-  private hasData = (): boolean => {
-    return this.props.data.chartDataDaily && this.props.data.chartDataDaily.length > 0;
-  };
-
-  private hasHourlyData = (): boolean => {
-    return this.props.data.chartDataHourly && this.props.data.chartDataHourly.length > 0;
-  };
-
-  private setFormattedData(data: ChartData) {
-    this.formattedData = {} as FormattedData;
-
-    data.forEach(entry => {
-      if (this.formattedData) {
-        this.formattedData[entry.time as number] = entry.formattedValue;
+    chartData.forEach(entry => {
+      if (formattedData.current) {
+        formattedData.current[entry.time as number] = entry.formattedValue;
       }
     });
-  }
+  };
 
-  private createChart = () => {
-    const { data: { chartDataMissing } } = this.props;
-    const darkmode = getDarkmode();
-    if (this.ref.current && this.hasData() && !chartDataMissing) {
-      if (!this.chart) {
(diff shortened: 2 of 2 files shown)


frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -49,120 +43,176 @@ type State = {
...
     toolTipValue: undefined,
     toolTipTop: 0,
     toolTipLeft: 0,
     toolTipTime: 0,
-    isMobile: false,
-  };
-
-  public componentDidMount() {
-    this.createChart();
+  });
```
thisconnect, line 101: split into source/setSource, difference/setDifference, diffSince/setDiffSince and the rest tooltip/setTooltip ?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -49,120 +43,176 @@ type State = {
...
+
+  const [prevChartDataDaily, setPrevChartDataDaily] = useState(data.chartDataDaily);
+  const [prevChartDataHourly, setPrevChartDataHourly] = useState(data.chartDataHourly);
+  const [prevChartFiat, setPrevChartFiat] = useState(data.chartFiat);
+  const [prevHideAmounts, setPrevHideAmounts] = useState(hideAmounts);
+
+  if (prevChartDataDaily !== data.chartDataDaily) {
```
thisconnect, line 108: could hooks/usePrevious be helpful here?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -314,292 +362,237 @@ class Chart extends Component<Props, State> {
       to,
       from,
     };
-  };
+  }
```
thisconnect, line 365: looks like this function could be moved out of the component next to renderDate

thisconnect (changes requested):

NicolaLS:
addressed everything, PTAL @thisconnect :)

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -28,141 +29,195 @@ import { DefaultCurrencyRotator } from '../../../components/rates/rates';
...
-        && (
-          prev.data.chartDataDaily.length !== chartDataDaily.length
-          || prev.data.chartDataHourly.length !== chartDataHourly.length
-        )
-    ) {
-      const data = this.state.source === 'hourly' ? chartDataHourly : chartDataDaily;
-      this.lineSeries.setData(data);
-      this.chart?.timeScale().fitContent();
-      this.setFormattedData(data);
+  });
```
thisconnect, line 162: I believe this effect is not needed, as the previous effect already and unconditionally calls initChart().

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -28,141 +29,195 @@ import { DefaultCurrencyRotator } from '../../../components/rates/rates';
...
+  useEffect(() => {
+    initChart();
+    return () => {
+      window.removeEventListener('resize', onResize);
+      if (chart.current) {
+        chart.current.timeScale().unsubscribeVisibleLogicalRangeChange(calculateChange);
+        chart.current.unsubscribeCrosshairMove(handleCrosshair);
+      }
+    };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
thisconnect, line 155: is this needed?

thisconnect (changes requested):

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -231,59 +286,57 @@ class Chart extends Component<Props, State> {
...
+      chart.current.remove();
+      chart.current = undefined;
+      window.removeEventListener('resize', onResize);
     }
-  };
+  }
 
-  private onResize = () => {
-    const isMobile = window.innerWidth <= 640;
-    this.setState({ isMobile });
```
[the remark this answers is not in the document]
NicolaLS: because it is calculated every time the component renders now. see [here](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2728/files#diff-c761d5505dba4edd4496dfe892f927231782977bae81a155d645a6a04dd7ab59R88).

edit: ah, the component does not re-render automatically when the window re-sizes..good catch. I will convert `isMobile` to state again, and initialize it just like right now, but also `setIsMobile` in `onResize` then.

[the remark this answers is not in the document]
thisconnect: can you past full links instead of "here", in some other comment you have like 3 links "here" "here" "this" this is hard for me to read.

also this should not update on ever render but on resize.

thisconnect:
> componentDidMount and componentWillUnmount are replaced by [this](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2728/files#diff-c761d5505dba4edd4496dfe892f927231782977bae81a155d645a6a04dd7ab59R140).

I believe "this" and a few other links broke. also it's hard to follow please consider using less links or paste full links. but if you rebase / squash etc. such links might break.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -28,141 +29,195 @@ import { DefaultCurrencyRotator } from '../../../components/rates/rates';
...
+  }, [source, chartDisplay]);
 
-  private hasHourlyData = (): boolean => {
-    return this.props.data.chartDataHourly && this.props.data.chartDataHourly.length > 0;
-  };
 
-  private setFormattedData(data: ChartData) {
-    this.formattedData = {} as FormattedData;
+  function setFormattedData(chartData: ChartData) {
+    formattedData.current = {} as FormattedData;
```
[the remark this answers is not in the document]
NicolaLS: removed `as FormattedData`, `as ChartData`, and ` as TChartFiltersProps`

frontends/web/src/routes/account/summary/chart.tsx
NicolaLS: yes, I also thought that, but I thought I should put it in there since there was [this](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2728/files#diff-c761d5505dba4edd4496dfe892f927231782977bae81a155d645a6a04dd7ab59L109).

So yes logically it does not need to be there but if we want to strictly say we only want to convert the component it should be there? But I'll remove it and test if everything still works.

frontends/web/src/routes/account/summary/chart.tsx
NicolaLS: we explicitly want to run the effect **once** which is possible by passing an empty dependency array according to the react docs. however eslint wants us to either add the used dependencies, or remove the dependency array (which would cause it to run every render). 

so we either ignore the warning with the comment or remove the dependency array but check if it already ran once e.g
```
const chartInitialized = useRef(false);

useEffect..
  if(!chartInitialized) ...chartInitialized = true, initChart() return cleanup
  else do nothing
 ```
 
 so the function in use effect would return `() => void | undefined` not sure if thats even possible ? can it have _no cleanup function_ when some condition is not met?

Comment (NicolaLS) [reply frontends/web/src/routes/account/summary/chart.tsx:155]: tldr; yes otherwise we get eslint errors

**Comment by NicolaLS** [comment]
> > componentDidMount and componentWillUnmount are replaced by [this](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2728/files#diff-c761d5505dba4edd4496dfe892f927231782977bae81a155d645a6a04dd7ab59R140).
> 
> I believe "this" and a few other links broke. also it's hard to follow please consider using less links or paste full links. but if you rebase / squash etc. such links might break.

ah I see..I think I will just remove the note now, thanks for letting me know. it was not super important just meant as a quick overview before starting the review but you already did that. (the links just pointed to the respective `useEffect`s..)

**Comment by NicolaLS** [comment]
> tested [b95ae02](https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/b95ae02c0b9347e753ed643957dfb8fbd06e39a7) webdev/servewallet and initChart is called, but somehow the chart isn't loading.
> Screenshot 2024-05-30 at 11 13 54

actually, I forgot testing after removing the `useEffect` that calls `initChart` if `!chart.current`..but I tested right now without changing anything and `webdev/servewallet` seem to work fine for me on firefox and chrome..I'll look into this though.

**Comment by NicolaLS** [comment]
(rebase master force push )

**Comment by NicolaLS** [comment]
> tested [b95ae02](https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/b95ae02c0b9347e753ed643957dfb8fbd06e39a7) webdev/servewallet and initChart is called, but somehow the chart isn't loading.
> Screenshot 2024-05-30 at 11 13 54

@thisconnect so it is definitely the `useEffect` I removed, I added it back in and the chart shows again. I would argue that for a transformation of the code it is also correct that it is in there, since the original code also contains:
```ts
public componentDidUpdate(prev: Props) {
    const { chartDataDaily, chartDataHourly } = this.props.data;
    if (!this.chart) {
      this.createChart();
    }
```

which translates to the `useEffect` that runs every render. I agree that it should not be like this, but would it make sense to refactor this in another PR? Of course I can do it in this PR too but I think it would be better in another PR.

Do you have any env variables for react? Maybe the chart shows for me because react runs `useEffect` twice in dev mode (so you're not in dev mode?)

frontends/web/src/api/account.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
+import { LineData } from 'lightweight-charts';
```
thisconnect, line 17:
```suggestion
import type { LineData } from 'lightweight-charts';
```

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -15,157 +15,215 @@
...
-      chartDataHourly: [],
-      chartFiat: 'USD',
-      chartTotal: null,
-      formattedChartTotal: null,
-      chartIsUpToDate: false,
-      lastTimestamp: 0,
-    },
-    hideAmounts: false,
+
+function getUTCRange() {
```
thisconnect, line 52: please use arrow functions if possible, I recently changed other functions see https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2731

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -15,157 +15,215 @@
...
+  return {
+    utcYear,
+    utcMonth,
+    utcDate,
+    to,
+    from,
   };
+}
+
+function renderDate(date: number, lang: string, src: string) {
```
thisconnect, line 69: please use newlines if there is more than 1-2 simple props.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -475,134 +478,109 @@ class Chart extends Component<Props, State> {
...
+
+  const {
+    toolTipVisible,
+    toolTipValue,
+    toolTipTop,
+    toolTipLeft,
+    toolTipTime,
+  } = tooltipData;
+
+
```
thisconnect, line 508: nit: for readability and consistency there is no need to have multiple newlines, here and before

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -15,157 +15,215 @@
  */
 
 import { createChart, IChartApi, LineData, LineStyle, LogicalRange, ISeriesApi, UTCTimestamp, MouseEventParams, ColorType, Time } from 'lightweight-charts';
-import { Component, createRef, ReactChild } from 'react';
-import { ISummary } from '../../../api/account';
-import { translate, TranslateProps } from '../../../decorators/translate';
+import { useContext, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { usePrevious } from '../../../hooks/previous';
+import { ISummary, ChartData } from '../../../api/account';
```
thisconnect, line 21:
```suggestion
import type { ISummary, ChartData } from '../../../api/account';
```

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -15,157 +15,215 @@
  */
 
 import { createChart, IChartApi, LineData, LineStyle, LogicalRange, ISeriesApi, UTCTimestamp, MouseEventParams, ColorType, Time } from 'lightweight-charts';
-import { Component, createRef, ReactChild } from 'react';
-import { ISummary } from '../../../api/account';
-import { translate, TranslateProps } from '../../../decorators/translate';
+import { useContext, useEffect, useRef, useState } from 'react';
```
thisconnect, line 18: nit: would be nice if react import is always first.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -15,157 +15,215 @@
...
-          visible: this.props.hideAmounts ? false : !this.state.isMobile,
-        }
-      });
+  useEffect(() => {
+    if (source !== 'hourly' && chartDisplay === 'week') {
+      setSource('hourly');
     }
+    initChart();
+    return () => {
+      window.removeEventListener('resize', onResize);
```
thisconnect, line 161: initChart conditionally adds the event listener, but here it always removes the event.

I think this should be changed somehow. I guess the event can always be added?

thisconnect (changes requested):
found a weird issue, something seems wrong:

- click week 
- rotate to BTC
- white screen & js error

Screenshot 2024-06-07 at 09 19 21

frontends/web/src/routes/account/summary/chart.tsx
NicolaLS: Good to know, so I'll always use them now. I was really confused on how to decide which ones to use..I kept the normal functions so the `git diff` is better because they don't need to be moved around (hoisting)

NicolaLS:
> found a weird issue, something seems wrong:
> 
>     * click week
> 
>     * rotate to BTC
> 
>     * white screen & js error
> 
> 
> Screenshot 2024-06-07 at 09 19 21

ah yes, this also happens when clicking on some account now and then going back to the account overview..its weird because I'm pretty sure that this was not an issue and some point, I don't know what change caused this (maybe it was an issue, I'm not hundred percent sure)

- the problem is we set the state of `chartDisplay` and `source` when the switch in `initChart` matches anything (this is why there is no issue if 'all' is selected. this causes the `chart.current` to stay uninitialized causing the infinite loop.
I will just fix the chart initialization now, thought it would be better in another PR but it seems like I can only get this one to work with a proper chart initialization logic.

- I changed everything in the first commit but it stays broken with the issue you described here
- I will add another commit fixing the initialization, but only for a easy review, the commits need to be squashed since the first one is broken

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -209,10 +210,13 @@ export const Chart = ({
   };
 
   const removeChart = () => {
-    if (chart.current) {
-      chart.current.remove();
-      chart.current = undefined;
+    if (chartInitialized.current) {
       window.removeEventListener('resize', onResize);
+      chart.current?.timeScale().unsubscribeVisibleLogicalRangeChange(calculateChange);
+      chart.current?.unsubscribeCrosshairMove(handleCrosshair);
```
NicolaLS, line 216: imo. `chart.remove()` should clean up everything but from a quick look at the function documentation I don't know if it does so I left the unsubscribing in here just for safety..wdyt @thisconnect should I remove these?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -101,10 +101,11 @@ export const Chart = ({
   const ref = useRef<HTMLDivElement>(null);
   const refToolTip = useRef<HTMLSpanElement>(null);
   const chart = useRef<IChartApi>();
+  const chartInitialized = useRef(false);
```
NicolaLS, line 104: I guess we could also use `chart.current` but I like it like that, it is more clear even if its a bit redundant. but no strong opinion, let me know if I should remove this and use `chart.current` (!== undefined) to check if the chart was initialized.

NicolaLS (commented):
@thisconnect I addressed all changes in the first commit and added a second one that refactors the chart initialization. When everything is fine I can squash them before merge. 
PTAL :)

frontends/web/src/routes/account/summary/chart.tsx
NicolaLS: actually typescript did not warn me when I changed this, had to move the functions around quite a bit so the order is correct and I did not get any runtime errors...

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -475,134 +293,280 @@ class Chart extends Component<Props, State> {
...
 
-export { HOC as Chart };
+  const chartHeight = `${!isMobile ? height : mobileHeight}px`;
+
+  return (
+    <section className={styles.chart}>
+      <header>
+        <div className={styles.summary}>
+          <div className={styles.totalValue}>
+            {formattedChartTotal !== null ? (
```
NicolaLS, line 516: note: can be `undefined` needs to be fixed but wait for master and rebase

thisconnect: did you rebase?

NicolaLS: Yes I did rebase, `formattedChartTotal` should now be either `null` or some value because #2850 now uses `success` and `error` response.

unrelated to this: I saw some other comments above, I thought I did address all change requests/review but will take another look and ping you if I think it is ready for review again, thanks :)

NicolaLS:
@thisconnect all change requests should be fixed/outdated now, I did not respond to all the comments individually to create less noise. PTAL or resolve the threads that are satisfied in case there is more to do (thanks :))

NicolaLS:
(rebased without changes)

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -433,45 +425,36 @@ export const Chart = ({
...
     return () => {
-      if (chart.current) {
-        window.removeEventListener('resize', onResize);
-        chart.current.timeScale().unsubscribeVisibleLogicalRangeChange(calculateChange);
-        chart.current.unsubscribeCrosshairMove(handleCrosshair);
-      }
+      removeChart();
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
```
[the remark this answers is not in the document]
NicolaLS: I can't really remember exactly why but I think there was an issue with infinite render-loops when I did this, even with `useCallback`. I just remembered that when starting to change it like you said, adding the two dependencies here means I have to add lots of dependencies to those functions with `useCallback`.

I remember that no matter what I tried I got infinite re-render loops like this, I think the issue was that we have to add dependencies on object fields (e.g `data.chartDataDaily`) but I'm going to look into this again.

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -433,45 +425,36 @@ export const Chart = ({
...
 
   useEffect(() => {
-    if (!chart.current) {
-      return;
+    if (!chartInitialized.current) {
+      initChart();
     }
+  });
```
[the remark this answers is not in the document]
NicolaLS: Yes I intentionally added this so it stays the same as it currently is:
https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/fb2888332c16bcc2b92f574b268db63b0d085c54/frontends/web/src/routes/account/summary/chart.tsx#L110-L114

I guess we can remove it, it should not be possible for a variable to become `undefined`/`null` if it is not explicitly assigned to it. So it should be enough to initialize it once with the other `useEffect`.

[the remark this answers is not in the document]
NicolaLS: *also should never be set to false if we don't explicitly assign false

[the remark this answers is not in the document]
NicolaLS: Adding the `initChart` and `removeChart` to the dependency array of the other `useEffect` and then removing this one works.

frontends/web/src/routes/account/summary/chart.tsx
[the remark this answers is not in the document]
NicolaLS: So I just tried it again and this time there are no problems at all, not sure what I did wrong last time.

NicolaLS:
@shonsirsha thanks, adding `initChart`/`removeChart` and using `useCallback` fixed the chart so that the other `useEffect` that ran on every render could be removed. I did this in another commit but let me know if I should squash all three into one, only the last into the second one before merge.

thisconnect:
@NicolaLS please squash and rebase

thisconnect (changes requested):
lgtm with nit, also please reabase

NicolaLS:
@thisconnect first push is rebase without changes but fixing merge conflicts, I basically overwrote these commits:
https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/c62a5c289d5b60aa9092e8c94bbccf157708dcbb#diff-c761d5505dba4edd4496dfe892f927231782977bae81a155d645a6a04dd7ab59L525
and
https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/97c80ba63f9e1bb71f252da65b5fc838d552997b#diff-c761d5505dba4edd4496dfe892f927231782977bae81a155d645a6a04dd7ab59R141

But it looks like the new version of `chart.tsx` already fixes/addresses what these commits did.

Second push changed the nit.

thisconnect (approved):
tested LGTM thanks 👍

PR #2730, opened 2024-05-28
frontend/feetargets: load fee targets only once

benma:
Due to the useEffect fetching the fee targets depending on config, it was executed many times when going to the send screen, while it should only be called once. This commit fixes this.

While not strictly necessary, I also moved the fee target fetching its own `useLoad()`, which should decrease the chance that we will accidentally increase the number of requests again.


thisconnect (approved):
LGTM 👍

PR #2731, opened 2024-05-28
frontend: prefer arrow functions

thisconnect:
Refactored to use arrow functions to provide a more concise syntax and ensure a consistent style across the codebase. Arrow functions are preferred in this context as they do not bind their own this, making them a better fit for functional components and callbacks.

Traditional function expressions were retained where necessary, such as for methods requiring their own this context, constructor functions, or where hoisting is necessary.

TypeScript's type system helps identify issues related to this binding and potential hoisting problems.


shonsirsha (approved):
LGTM thanks

thisconnect:
rebased

PR #2733, opened 2024-05-29
Frontend type CoinCode

thisconnect:


frontends/web/src/api/account.ts
```diff
@@ -31,14 +31,18 @@ export type CoinUnit = 'BTC' | 'sat' | 'LTC' | 'ETH' | 'TBTC' | 'tsat' | 'TLTC'
 
 export type ERC20TokenUnit = 'USDT' | 'USDC' | 'LINK' | 'BAT' | 'MKR' | 'ZRX' | 'WBTC' | 'PAXG' | 'DAI';
 
+export type ERC20TokenCode = 'erc20Test' | 'eth-erc20-usdt' | 'eth-erc20-usdc' | 'eth-erc20-link' | 'eth-erc20-bat' | 'eth-erc20-mkr' | 'eth-erc20-zrx' | 'eth-erc20-wbtc' | 'eth-erc20-paxg' | 'eth-erc20-dai0x6b17';
+
+export type CoinOrTokenCode = CoinCode | ERC20TokenCode;
```
Beerosagos, line 36: Imo it would be more readable if we defined `CoinCode = 'btc' | ... | 'sepeth' | ERC20TokenCode`

I don't think we'll often need to differentiate between Coins and Tokens, and this would still allow to do it, right?

thisconnect: I'm not 100% if I understand.

> I don't think we'll often need to differentiate between Coins and Tokens, and this would still allow to do it, right?

In add account we only allow adding Coins but not tokens.. I believe this change reflects better how the app works, and I think we sohuldn't mix tokens into CoinCodes.

frontends/web/src/api/account.ts
```diff
@@ -104,7 +108,7 @@ export const getAccountsTotalBalance = (): Promise<TAccountsTotalBalanceResponse
 };
 
 export type TCoinsTotalBalance = {
-  [key: string]: IAmount;
+  [key in CoinOrTokenCode]: IAmount;
```
thisconnect, line 111: Hm, I think this is not 100% correct, looking at the response it seems the the backend only returns used coins
probably should change to:

```suggestion
  [key in CoinOrTokenCode]?: IAmount;
```

I'll check the other maps in this file too...

frontends/web/src/api/account.ts
```diff
@@ -70,11 +74,11 @@ export const getAccounts = (): Promise<IAccount[]> => {
 };
 
 export type TAccountsBalanceByCoin = {
-    [key: string]: IAmount;
+    [key in CoinOrTokenCode]: IAmount;
```
thisconnect, line 77:
```suggestion
    [key in CoinOrTokenCode]?: IAmount;
```
otherwise it is typed that the backend alwas returns a map with all coins/tokens, but it only returns a map with used coins/tokens

Screenshot 2024-05-30 at 09 46 28

frontends/web/src/api/account.ts
thisconnect: I'll fix this up into the first commit, TS seems happy about it and doesn't complain about any other issues when making this key optional.

thisconnect: fixed

frontends/web/src/api/account.ts
Beerosagos: I see the point. I just don't like much having a type that is used a lot in the frontend that is called with a name like "ThisOrThat". I'd prefer to keep using the world Coin in its more general sense (btc/ltc/eth/erc20token), as I think it is more readable. What about finding a different name than 'Coin' for btc/ltc/eth and the other layer1 coins? But I can't think of a good name for that. wdyt? 

I mean smth like:
- Layer1Coin = 'btc' | 'eth' | ....
- ERC20Token = 'eth-erc20-usdt' | 'eth-erc20-usdc' | ....
- Coin = Layer1Coin | Erc20Token

But `Layer1Coin` sucks :grimacing:

Beerosagos: maybe `AccountCoin`?

thisconnect: I agree it's not a great name, but somewhat accurate.

Accounts coin code can be a coin or token... hm. CryptoCode ? 🙃  I'll ping others maybe somebody has a nicer name.

thisconnect:
rebased

frontends/web/src/api/account.ts
thisconnect: changed to `CoinCode = NativeCoinCode | ERC20CoinCode`

thisconnect: done also squashed the commits @Beerosagos PTAL

frontends/web/src/api/backend.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { AccountCode, CoinCode } from './account';
+import { AccountCode, CoinCode, ERC20CoinCode } from './account';
```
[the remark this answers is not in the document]
thisconnect: Yes 💯

thisconnect:
There is some overlap with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2736
I'll wait with this PR...

thisconnect:
rebased, waiting with merging as there is some overlap with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2736

Beerosagos:
> rebased, waiting with merging as there is some overlap with #2736

Just approved #2736 ! Pls, lmk when this is ready for review :)

thisconnect:
> Just approved https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2736 ! Pls, lmk when this is ready for review :)

ack, then I'll rebase on top of that branch

thisconnect:
rebased on top of strmci:total_coins_update f76468a60254b047c27df729b565cb367c56d227 so it has that commit in this branch as well.

@Beerosagos ready for review.

frontends/web/src/routes/account/summary/summarybalance.tsx
```diff
@@ -107,7 +95,7 @@ export const SummaryBalance = ({
           <tbody>
             { accounts.length > 0 ? (
               coins.map(coinCode => {
-                const balanceRows = accountsPerCoin[coinCode].map(account =>
+                const balanceRows = accountsPerCoin[coinCode]?.map(account =>
```
Beerosagos, line 98: nit: I see that `coins` is obtained from `accountsPerCoin` keys. Do we need the `?` anyway? I think that if we have the coinCode, we should also have at least one account. wdyt? It can't hurt, anyway.

Beerosagos (approved):
tACK with a small possible nit :) Very nice! Thanks

PR #2737, opened 2024-05-31
qt/windows: fix utf8 encoding of notes etc.

benma:
When entering chars like 'ü' as a tx note on Windows, the char was encoded as "\xef\xbf\xbd", which is displayed as '�'.

The frontend handled it properly, but when sending the `QString` to Go, the conversion used this function:

https://doc.qt.io/qt-5/qstring.html#toLocal8Bit

in

```
backendCall(queryID, query.toLocal8Bit().constData());
```

which uses whatever local codec is defined.

In Go strings are UTF-8, so using `toUtf8()` instead is appropriate and fixes the issue.

We change all instances of toLocal8bit to toUtf8 for the same reason, fixing possibly other existing bugs on Windows.


thisconnect (approved):
untested LGTM

PR #2738, opened 2024-05-31
ci: check that locize placeholders are the same in all languages

benma:
Translators sometimes accidentally translate `{{placeholder}}` variables, like "name"=>"nombre", breaking the translation.

This script checks that for each key in English (reference lang), all translations contain the same placeholders.


benma:
CI fails with the expected errors in the locize files - we should probably fix all issues so make CI pass here, then merge to prevent future issues.

benma:
@thisconnect done

thisconnect:
thank you. fixed
- SL, made same change as it is in buy.info.disclaimer.protection.descriptionGeneric
- FR, fixed placeholders
- FA and AR google translated
- IT fixed placeholder
- ja used gpt to translate again (it's only some warning about bitsurance and new tx alert message)
- es fixed placeholder
- he deepled
- zh deepled

benma:
Thanks. One remaining:

```

+ ./scripts/check-locize-placeholders.py
Issues found in fa:
  - Placeholder mismatch at 'notification.newTxs_one': '{{count}} معاملات جدید در: {{accountName}}' - found ['{{count}}', '{{accountName}}']; expected ['{{accountName}}']
 ```

thisconnect:
> Thanks. One remaining:

oups 🙈 sorry, should be fixed now.

benma:
Yeah it's fixed now, thanks.

thisconnect (approved):
LGTM thank you

baizon:
Why was the `scripts/check-locize-placeholders.py` written in python? Why not Go, IMO it's not good to introduce another programming language.

thisconnect:
> Why not Go

or Node.js 😇

baizon:
Indeed, because of `frontends/web/src/locales/...` i would see it within Node and "frontend".

benma:
> Why was the `scripts/check-locize-placeholders.py` written in python? Why not Go, IMO it's not good to introduce another programming language. More complexity, more to maintain, etc.

Good point about another language, but it's available anyway in the Docker image.

Python felt like the best tool for the job - it's is kind of perfect for simple scripts like this imho. No strong opinion though, Go would have been fine too, but I don't think I'll re-do it now.

baizon:
Yes, but like I said, then there is a Docker Python dependency. What if Python 4 is released and no more support for code fragments, etc. IMO the solution should be in Go, Node, or bash.

benma:
Fyi, this is not the first Python script. ./scripts/get_version has been there since the beginning and is literally untouched since 5 years :grin: as a measure of how much of a maintenance burden it is. 

> Yes, but like I said, then there is a Docker Python dependency. What if Python 4 is released and no more support for code fragments, etc. IMO the solution should be in Go, Node, or bash.

I see your point, but these risks are not realistic imho. Python is available (has always been, will likely always be), Python 4 is not coming for many years if ever, breaking changes are very rare (not sure I ever experiences a breaking change in Python except 2->3). I really don't expect much Python-related maintenance overhead for this.

PR #2740, opened 2024-06-03
frontend: make reusable pill button & pill button group

shonsirsha:
Right now, we're only using such components in `filters.tsx`. However, this will change in the future as other components may incorporate these pill buttons for filtering / tabbing-like functionality.

I didn't make a new variant in the button component because we wanna keep the styling on the parent container of the buttons (`PillButtonGroup`) as we want them to look & behave the same (size-wise, etc) on every components we want to have them in. So instead, I made a `PillButtonGroup` and `PillButton` components to be used together.

### DIFF
--- a/frontends/web/src/routes/account/summary/filters.tsx
+++ b/frontends/web/src/routes/account/summary/filters.tsx
@@ -16,6 +16,7 @@
 
 import { useTranslation } from 'react-i18next';
 import { TChartFiltersProps } from './types';
+import { PillButton, PillButtonGroup } from '../../../components/pillbuttongroup/pillbuttongroup';
 import styles from './chart.module.css';
 
 export const Filters = ({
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/account/summary/filters.tsx
```diff
@@ -1,3 +1,4 @@
+
```
thisconnect, line 1: remove?

thisconnect (changes requested):
untested LGTM

please copy the PR description to commit message so we have it in the git history.

nit: how about renaming the file from `pillbuttongroup` to just `pill` ?

shonsirsha:
Thanks, PTAL @thisconnect !

shonsirsha:
> nit: how about renaming the file from pillbuttongroup to just pill ?

Hm.. To me just `pill` isn't descriptive enough tbh. 😇 I think naming it pill**buttongroup** is clearer to describe that it's a:

1. component to be used with **pillbutton**, and 
2. on first glance, only having `<Pill>` somewhere in the codebase to me, also imply that it's just as a "pill". Except it's not - it's a container and not even a "pill" per-se.

thisconnect:
or just PillButton ?

thisconnect (approved):
tested LGTM

PR #2741, opened 2024-06-04
backend/lightning: don't expose mnemonic words for lightning

strmci:
This commit ensures the mnemonic words created
for lightning wallet are not exposed to frontend.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -1529,8 +1529,29 @@ func (handlers *Handlers) postExportLog(r *http.Request) interface{} {
 	return result{Success: true}
 }
 
+type lightningAccountConfigWithoutMnemonic struct {
+	RootFingerprint jsonp.HexBytes     `json:"rootFingerprint"`
+	Code            accountsTypes.Code `json:"code"`
+	Number          uint16             `json:"num"`
+}
+
+type lightningConfigWithoutMnemonic struct {
+	Accounts []*lightningAccountConfigWithoutMnemonic `json:"accounts"`
+}
+
 func (handlers *Handlers) getLightningConfigHandler(_ *http.Request) interface{} {
-	return handlers.backend.Config().LightningConfig()
+	lightningAccounts := handlers.backend.Config().LightningConfig().Accounts
+	accounts := make([]*lightningAccountConfigWithoutMnemonic, len(lightningAccounts))
+	for i, account := range lightningAccounts {
+		accounts[i] = &lightningAccountConfigWithoutMnemonic{
+			RootFingerprint: account.RootFingerprint,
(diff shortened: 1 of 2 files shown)


backend/config/lightning.go
```diff
@@ -23,7 +23,7 @@ import (
 // LightningAccountConfig is the configuration of a single Lightning account.
 type LightningAccountConfig struct {
 	// Mnemonic is the wallet node generated from the device entropy.
-	Mnemonic string `json:"mnemonic"`
+	Mnemonic string `json:"-"`
```
benma, line 26: With this you delete the mnemonic also from the config, and the user can't spend anymore :cold_sweat: You even changed the test above to reflect this :sweat_smile:

benma (changes requested):

backend/config/lightning.go
strmci: hmm, I'll look into it, thanks.

Beerosagos (changes requested):
Plase, also remove the mnemonic from the frontend object :)

strmci:
> Plase, also remove the mnemonic from the frontend object :)

Removed.

Beerosagos (approved):
utACK

backend/handlers/handlers.go
```diff
@@ -1530,7 +1530,11 @@ func (handlers *Handlers) postExportLog(r *http.Request) interface{} {
 }
 
 func (handlers *Handlers) getLightningConfigHandler(_ *http.Request) interface{} {
-	return handlers.backend.Config().LightningConfig()
+	lightningConfig := handlers.backend.Config().LightningConfig()
+	for _, account := range lightningConfig.Accounts {
+		account.Mnemonic = ""
```
thisconnect, line 1535: Why set to empty string? Is there a way to remove the mnemonic field?

strmci: It's now removed by adding `omitempty` to the json descriptor (thanks @Beerosagos)

benma: This still removes the mnemonic from the actual lightning account object in the config, so it will be gone afterwards, and also persisted on the next config save() I think.

To do this properly, you'd have to change the handler that returns the data to the frontend, probably recreating a structure with the relevant data for the frontend without the mnemonic.

Beerosagos: Dang, @benma is right. I didn't realize there was a pointer to the LightningAccountConfig, I thought it would have been a deep copy. Sorry for not noticing it :pray:

strmci:
@Beerosagos @benma 
I recreated the lightning config without a mnemonic in the handler, PTAL

benma (approved):
utACK

PR #2742, opened 2024-06-04
frontend: refactor password.jsx into typescript

NicolaLS:
Refactor the `password.jsx` file into typescript.

I had to cast to the concrete type a lot of times, I'm not sure if that makes sense since it can throw an error, but how _should_ an error be handled? e.g. `this.setState({ [stateKey]: value } as Pick<TState, keyof TState>, this.validate);`

### DIFF
--- a/frontends/web/src/components/password.tsx
+++ b/frontends/web/src/components/password.tsx
@@ -1,5 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
+ * Copyright 2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,13 +16,33 @@
  */
 
 import { Component, createRef } from 'react';
+import { TranslateProps, translate } from '@/decorators/translate';
 import { Input, Checkbox, Field } from './forms';
 import { alertUser } from './alert/Alert';
 import style from './password.module.css';
-import { withTranslation } from 'react-i18next';
 
-export const PasswordInput = (props) => {
-  const { seePlaintext, ...rest } = props;
+const excludeKeys = /^(Shift|Alt|Backspace|CapsLock|Tab)$/i;
+
+const hasCaps = (event: KeyboardEvent) => {
+  const key = event.key;
+  // will return null, when we cannot clearly detect if capsLock is active or not
+  if (key.length > 1 || key.toUpperCase() === key.toLowerCase() || excludeKeys.test(key)) {
+    return null;
+  }
+  // ideally we return event.getModifierState('CapsLock')) but this currently does always return false in Qt
+  return key.toUpperCase() === key && key.toLowerCase() !== key && !event.shiftKey;
+};
+
+type TPropsPasswordInput = {
+  seePlaintext?: boolean;
+  id?: string;
+  idPrefix?: string;
+  label: string;
+  placeholder?: string;
+  onInput?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  value: string;
+}
+export const PasswordInput = ({ seePlaintext, ...rest }: TPropsPasswordInput) => {
   return (
     <Input
       type={seePlaintext ? 'text' : 'password'}
@@ -30,24 +51,51 @@ export const PasswordInput = (props) => {
   );
 };
 
-class PasswordSingleInputClass extends Component {
+type TProps = {
+  idPrefix?: string;
+  pattern?: string;
+  autoFocus?: boolean;
+  disabled?: boolean;
+  label?: string;
+  placeholder?: string;
+  title?: string;
+  showLabel?: string;
+  onValidPassword: (password: string | null) => void;
+}
+
+type TPasswordSingleInputProps = TProps & TranslateProps;
+
+type TState = {
+  password: string;
+  seePlaintext: boolean;
+  capsLock: boolean;
+}
+
+class PasswordSingleInputClass extends Component<TPasswordSingleInputProps, TState> {
+  private regex?: RegExp;
+
   state = {
     password: '',
     seePlaintext: false,
     capsLock: false
   };
 
-  password = createRef();
+  password = createRef<HTMLInputElement>();
 
   idPrefix = () => {
     return this.props.idPrefix || '';
   };
 
-  UNSAFE_componentWillMount() {
-    window.addEventListener('keydown', this.handleCheckCaps);
-  }
+  handleCheckCaps = (event: KeyboardEvent) => {
+    const capsLock = hasCaps(event);
+
+    if (capsLock !== null) {
+      this.setState({ capsLock });
+    }
+  };
 
   componentDidMount() {
+    window.addEventListener('keydown', this.handleCheckCaps);
     if (this.props.pattern) {
       this.regex = new RegExp(this.props.pattern);
     }
@@ -60,8 +108,9 @@ class PasswordSingleInputClass extends Component {
     window.removeEventListener('keydown', this.handleCheckCaps);
   }
 
-  tryPaste = event => {
-    if (event.target.type === 'password') {
+  tryPaste = (event: React.ClipboardEvent<HTMLInputElement>) => {
+    const target = event.currentTarget;
+    if (target.type === 'password') {
       event.preventDefault();
       alertUser(this.props.t('password.warning.paste', {
         label: this.props.label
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/device/bitbox01/unlock.jsx
```diff
@@ -135,12 +135,10 @@ class Unlock extends Component {
                         <PasswordSingleInput
                           autoFocus
                           id="password"
-                          type="password"
```
thisconnect, line 138: 👍

frontends/web/src/routes/device/components/skipfortesting.tsx
```diff
@@ -58,11 +58,9 @@ export const SkipForTesting = ({
       <Dialog open={dialog} title={title} onClose={() => setDialog(false)}>
         <form onSubmit={registerTestingDevice}>
           <PasswordSingleInput
-            type="password"
             autoFocus
             label="Test Password"
-            onValidPassword={setTestPIN}
-            value={testPIN} />
+            onValidPassword={(pw) => pw ? setTestPIN(pw) : setTestPIN('')}/>
```
thisconnect, line 63: with typing this you found a small bug here. 👍

frontends/web/src/components/password.tsx
```diff
@@ -302,9 +340,10 @@ class PasswordRepeatInputClass extends Component {
   }
 }
 
-export const PasswordRepeatInput = withTranslation(null, { withRef: true })(PasswordRepeatInputClass);
+const HOCRepeat = translate(undefined, { withRef: true })(PasswordRepeatInputClass);
+export { HOCRepeat as PasswordRepeatInput };
 
-const MatchesPattern = ({ regex, value = '', text }) => {
+const MatchesPattern = ({ regex, value = '', text }: { regex: RegExp | undefined, value: string, text: string | undefined }) => {
```
thisconnect, line 346: please change to `type MatchesPatternProps = {`

and use MatchesPatternProps here.

frontends/web/src/components/password.tsx
```diff
@@ -316,7 +355,7 @@ const MatchesPattern = ({ regex, value = '', text }) => {
 
 const excludeKeys = /^(Shift|Alt|Backspace|CapsLock|Tab)$/i;
 
-const hasCaps = ({ key }) => {
+const hasCaps = ({ key }: { key: string }) => {
```
thisconnect, line 358: not sure if better to use `{ key: string }` or just `KeyboardEvent`

frontends/web/src/components/password.tsx
```diff
@@ -211,18 +249,18 @@ class PasswordRepeatInputClass extends Component {
...
     const stateKey = event.target.id.slice(this.idPrefix().length);
-    this.setState({ [stateKey]: value }, this.validate);
+    this.setState({ [stateKey]: value } as Pick<TState, keyof TState>, this.validate);
   };
 
-  handleCheckCaps = event => {
+  handleCheckCaps = (event: KeyboardEvent) => {
     const capsLock = hasCaps(event);
-        if (capsLock != null) { // eslint-disable-line
+    if (capsLock != null) { // eslint-disable-line
```
thisconnect, line 263: since you are here and this is now typed, it looks like ` // eslint-disable-line` isnt needed anymore
```suggestion
    if (capsLock !== null) {
```

frontends/web/src/components/password.tsx
thisconnect: actually there is another bug with `event.shiftKey` at the end of this function, please remove both `@ts-ignore` and `eslint-disable-next-line` and check for a solution 🙏

thisconnect (changes requested):

frontends/web/src/components/password.tsx
```diff
@@ -88,16 +120,16 @@ class PasswordSingleInputClass extends Component {
...
+  handleFormChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    let value: string | boolean = event.target.value;
     if (event.target.type === 'checkbox') {
       value = event.target.checked;
     }
-    const stateKey = event.target.id.slice(this.idPrefix().length);
-    this.setState({ [stateKey]: value }, this.validate);
+    const stateKey = event.target.id.slice(this.idPrefix().length) as keyof TState;
+    this.setState({ [stateKey]: value } as Pick<TState, keyof TState>, this.validate);
   };
```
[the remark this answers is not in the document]
NicolaLS: I'll take a note and either open a PR or issue.

frontends/web/src/components/password.tsx
NicolaLS: ah I think the whole function can just return `event.getModifierState('CapsLock')` then, the bug with that always returning false on qt was probably because `event` was not globally available and it was also not passed to the function.

Actually we can remove the function completely if `event.getModifierState` works, will test on qt now.

NicolaLS: Looks like adding the event listener does not even work when it is done in `UNSAFE_componentWilMount`! I'll remove the function and add it to `componentDidMount`

I noticed I can't really test the input dialog on qt, but I'll put the same logic in some other component just to test.

NicolaLS: Ok so `event.getModifierState('CapsLock')` still does not work on qt linux.

NicolaLS:
@thisconnect thanks, PTAL.

note: I removed the `UNSAFE_componentWillMount` and added the `keydown` event listener in the `componentDidMount` now because (at least on `webdev` firefox) the listener did not work (the method ran, but the listener was either not added correctly or broken).

frontends/web/src/components/password.tsx
```diff
@@ -302,9 +349,15 @@ class PasswordRepeatInputClass extends Component {
...
 
-export const PasswordRepeatInput = withTranslation(null, { withRef: true })(PasswordRepeatInputClass);
+const HOCRepeat = translate(undefined, { withRef: true })(PasswordRepeatInputClass);
+export { HOCRepeat as PasswordRepeatInput };
 
-const MatchesPattern = ({ regex, value = '', text }) => {
+type MatchesPatternProps = {
+  regex: RegExp | undefined,
+  value: string,
+  text: string | undefined,
```
thisconnect, line 358: here and in other type in this file `,` -> `;`

frontends/web/src/components/password.tsx
```diff
@@ -153,29 +195,40 @@ class PasswordSingleInputClass extends Component {
 
 }
 
-export const PasswordSingleInput = withTranslation(null, { withRef: true })(PasswordSingleInputClass);
+const HOC = translate(undefined, { withRef: true })(PasswordSingleInputClass);
+export { HOC as PasswordSingleInput };
+
 
+type TPasswordRepeatProps = TPasswordSingleInputProps & { repeatLabel?: string; repeatPlaceholder: string; };
```
thisconnect, line 202: please use newlines for readabilty

thisconnect (changes requested):

frontends/web/src/components/password.tsx
thisconnect: yeah I believe @shonsirsha mentioned something like that before..

frontends/web/src/components/password.tsx
thisconnect: ptal

thisconnect: PTAL 🙏

thisconnect (changes requested):

frontends/web/src/components/password.tsx
thisconnect: self-review the PR after you push then you see :)

thisconnect (changes requested):
sorry for the slow review. looks all good with a typescript nits in 2 places.

please update then I do a final test with BB01

NicolaLS:
@thisconnect thanks, first push was the actual change, last push is a rebase on master.

frontends/web/src/components/password.tsx
```diff
@@ -60,8 +108,9 @@ class PasswordSingleInputClass extends Component {
     window.removeEventListener('keydown', this.handleCheckCaps);
   }
 
-  tryPaste = event => {
-    if (event.target.type === 'password') {
+  tryPaste = (event: React.ClipboardEvent<HTMLInputElement>) => {
+    const target = event.target as HTMLInputElement;
```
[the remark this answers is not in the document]
NicolaLS: oh makes sense, I did not understand why _typescript was so stupid_ but I used the wrong property..

thisconnect (approved):
tested LGTM

PR #2743, opened 2024-06-05
Frontend fix css regression

thisconnect:


NicolaLS:
tested `AppImage` LGTM! 
- Sidebar scroll works
- Pocket and Moonpay show properly now

thisconnect:
added a commit that should fix:
- height not expanding to full height when the content is smaller than the window height
- dialogs always using full height

what to test:
- waiting view
- dialog i.e. enable tor proxy
- bootloader view (using centeredcontent comp)

NicolaLS:
tested `webdev` firefox (BB02B and software keystore), `AppImage` on linux and the android app. the problem with the dialog expanding is fixed, not sure what the issue with bootloader view and waiting view was but they looked normal to me too.

shonsirsha:
Tested on Chrome webdev macOS. Seems to fix the regressions. 👍 


blv

mb

bbtc

tp

password

shonsirsha (approved):
LGTM, thanks.

PR #2745, opened 2024-06-06
frontend: add red dot in sidebar and settings when fw can update

thisconnect:
Users miss that the firmware can update or are just not aware.

Added red dot in the sidebar (or on device tab) indicating that there is something important.

In the sidebar this only checks for the first connected device.

### DIFF
--- a/frontends/web/src/components/sidebar/sidebar.tsx
+++ b/frontends/web/src/components/sidebar/sidebar.tsx
@@ -15,13 +15,15 @@
  * limitations under the License.
  */
 
-import React, { useContext, useEffect } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { useLocation } from 'react-router';
 import { Link, NavLink } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import { useKeystores } from '../../hooks/backend';
-import { IAccount } from '../../api/account';
+import type { TDevices } from '../../api/devices';
+import type { IAccount } from '../../api/account';
 import { deregisterTest } from '../../api/keystores';
+import { getVersion } from '../../api/bitbox02';
 import coins from '../../assets/icons/coins.svg';
 import ejectIcon from '../../assets/icons/eject.svg';
 import shieldIcon from '../../assets/icons/shield_grey.svg';
@@ -31,7 +33,7 @@ import settingsGrey from '../../assets/icons/settings-alt_disabled.svg';
 import deviceSettings from '../../assets/icons/wallet-light.svg';
 import { debug } from '../../utils/env';
 import { AppLogoInverted, Logo } from '../icon/logo';
-import { CloseXWhite, USBSuccess } from '../icon';
+import { CloseXWhite, RedDot, USBSuccess } from '../icon';
 import { getAccountsByKeystore, isAmbiguiousName, isBitcoinOnly } from '../../routes/account/utils';
 import { SkipForTesting } from '../../routes/device/components/skipfortesting';
 import { Badge } from '../badge/badge';
@@ -41,6 +43,7 @@ import style from './sidebar.module.css';
 
 type SidebarProps = {
   deviceIDs: string[];
+  devices: TDevices;
   accounts: IAccount[];
 };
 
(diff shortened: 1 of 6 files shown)


frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -79,8 +80,17 @@ const Sidebar = ({
 }: SidebarProps) => {
   const { t } = useTranslation();
   const { pathname } = useLocation();
+  const [ canUpgrade, setCanUpgrade ] = useState(false);
   const { activeSidebar, sidebarStatus, toggleSidebar } = useContext(AppContext);
 
+  useEffect(() => {
+    if (deviceIDs.length) {
+      getVersion(deviceIDs[0]).then(versionInfo => setCanUpgrade(versionInfo.canUpgrade));
```
thisconnect, line 88: In the sidebar this only checks for the first deviceID, in settings tab it checks every deviceID.

If we want proper multi device support here, we could check every deviceID and add the dot if one of them canUpgrade.

Should update this? cc @benma

benma: Even though we support only one, it's easy to check all of them, right? So why not.

I think this code fails if a BitBox01 is connected, or the BitBox02 bootloader. You probably need to pass `devices` not `deviceIDs` to the sidebar, which contains the `TProductName`, which is `'bitbox' | 'bitbox02' | 'bitbox02-bootloader'`, so you should only call getVersion if the product name is `bitbox02`

thisconnect:
added changelog entry

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -75,12 +78,26 @@ const eject = (e: React.SyntheticEvent): void => {
...
 }: SidebarProps) => {
   const { t } = useTranslation();
   const { pathname } = useLocation();
+  const [ canUpgrade, setCanUpgrade ] = useState(false);
   const { activeSidebar, sidebarStatus, toggleSidebar } = useContext(AppContext);
 
+  useEffect(() => {
+    const deviceList = Object.entries(devices).filter(([_, productName]) => productName === 'bitbox02');
+    if (deviceList.length) {
+      Promise.all(
```
benma, line 92: It does not matter much with only one device, but if we had multiple, you'd want to not call getVersion on all of them but abort early if one returns true.

I asked ChatGPT to rewrite it with this in mind and it spit out this:

```
useEffect(() => {
  const checkUpgradableDevices = async () => {
    const bitboxDevices = Object.keys(devices).filter(deviceID => devices[deviceID] === 'bitbox02');

    for (const deviceID of bitboxDevices) {
      const { canUpgrade } = await getVersion(deviceID);
      if (canUpgrade) {
        setCanUpgrade(true);
        return; // Exit early as we found an upgradable device
      }
    }
  };

    checkUpgradableDevices();
}, [devices]);
```

imo this is also clearer to read. Promise.all etc. are a bit hard to read for me.

wdyt?

thisconnect: hm... Now we have the opposite problem, I find it hard to read. It was my intention to do the calls in parallel and set state as soon as the first result returns true.

I guess am not used to await in for loops and find it not natural (yet).  I should see how this transpiles to Promises to really understand. 

By reading your/gpt version I can't even tell if it is parallel or in sequence.

I'll try to understand your version if you read Promise.all doc :)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

thisconnect: @benma changed to async / sequential calls, tested with 1 device :)

Screenshot 2024-06-13 at 08 27 17

Screenshot 2024-06-13 at 08 27 30

benma: I actually didn't consider concurrent vs sequential execution at all :flushed: :thinking: 

The for-loop is sequential. Imo it's fine to keep your newest version even though it is not concurrent, as it's imho much clearer, and a bit less needless load on the backend (if we ever have multiple device support :joy:).

benma (approved):
utACK

PR #2752, opened 2024-06-08
feat: added dependabot

baizon:
Hi, 
I've added [dependabot](https://github.com/dependabot). I recommend to activate it, so automated (security) updates are performed. I started with updates for `github-actions` and `gomod`. You can add more components ([package-ecosystems](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)). You can activate dependabot in the [Code security and analysis section](https://github.com/BitBoxSwiss/bitbox-wallet-app/settings/security_analysis).
For more information see the [Dependabot quickstart guide](https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"


benma:
Thanks, but we already have dependabot alerts, possibly only visible to members of this repo. Is this something different?

baizon:
This is configuration, so dependabot will create PR for updates (for go and github-actions) automatically. That would be helpfull, because something like https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2754 wouldn't happen.

benma (approved):
Let's try it. I think a weekly PR to update deps might be a bit too much overhead, but let's see. Can always decrease the frequency later if needed.

benma (changes requested):
Actually you have an accidental merge commit in this PR, please push only one commit.

baizon:
You're correct, I've fixed it.

benma (approved):

PR #2754, opened 2024-06-08
chore(deps): update all go dependencies

baizon:
- Updated Go to 1.21
- Updated all Go module dependencies


benma:
Thanks for your contribution!

Could you please split the Go1.21 fixes into one commit, and the dep updates in one other commit? Could you also provide the exact commands you used to update the deps, so I can more easily check that I arrive at the same diff?

PR #2758, opened 2024-06-09
refactor: fixed go lint issues

baizon:
- Update `bitboxswiss/bitbox-wallet-app-ci` docker image to version `23`
- Update `.golangci.yml` to be compatible with version `1.59.0`
- Fixed all the go lint issues
- Added `golangci-lint config verify` to `ci.sh` to check the go lint configuration (`.golangci.yml`)

### DIFF
--- a/backend/rates/history.go
+++ b/backend/rates/history.go
@@ -178,11 +178,11 @@ func (updater *RateUpdater) updateHistory(ctx context.Context, coin, fiat string
 
 	updater.historyMu.Lock()
 	defer updater.historyMu.Unlock()
-	allRates := append(updater.history[bucketName], fetchedRates...)
-	sort.Slice(allRates, func(i, j int) bool {
-		return allRates[i].timestamp.Before(allRates[j].timestamp)
+
+	updater.history[bucketName] = append(updater.history[bucketName], fetchedRates...)
+	sort.Slice(updater.history[bucketName], func(i, j int) bool {
+		return updater.history[bucketName][i].timestamp.Before(updater.history[bucketName][j].timestamp)
 	})
-	updater.history[bucketName] = allRates
 
 	return len(fetchedRates), nil
 }
--- a/backend/rates/history_test.go
+++ b/backend/rates/history_test.go
@@ -123,7 +123,7 @@ func TestFetchGeckoMarketRangeInvalidCoinFiat(t *testing.T) {
 		var updater RateUpdater
 		g := fixedTimeRange(time.Now().Add(-time.Hour), time.Now())
 		_, err := updater.fetchGeckoMarketRange(ctx, test.coin, test.fiat, g)
-		assert.Error(t, err, "fetchGeckoMarketRange(%q, %q) returned nil error", test.coin, test.fiat)
+		require.Error(t, err, "fetchGeckoMarketRange(%q, %q) returned nil error", test.coin, test.fiat)
 		cancel()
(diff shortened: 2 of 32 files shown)


backend/rates/history.go
```diff
@@ -178,7 +178,8 @@ func (updater *RateUpdater) updateHistory(ctx context.Context, coin, fiat string
 
 	updater.historyMu.Lock()
 	defer updater.historyMu.Unlock()
-	allRates := append(updater.history[bucketName], fetchedRates...)
+	updater.history[bucketName] = append(updater.history[bucketName], fetchedRates...)
+	allRates := updater.history[bucketName]
 	sort.Slice(allRates, func(i, j int) bool {
 		return allRates[i].timestamp.Before(allRates[j].timestamp)
 	})
```
benma, line 185: the line below this one is not needed I think. For clarity maybe it's better to just do this?

```
	updater.history[bucketName] = append(updater.history[bucketName], fetchedRates...)
	sort.Slice(updater.history[bucketName], func(i, j int) bool {
		return updater.history[bucketName][i].timestamp.Before(updater.history[bucketName][j].timestamp)
	})
	return len(fetchedRates), nil
```

wdyt?

backend/rates/history_test.go
```diff
@@ -42,9 +42,7 @@ func TestPriceAt(t *testing.T) {
 		{7.25, time.Date(2020, 9, 3, 18, 0, 0, 0, time.UTC)},
 	}
 	for _, test := range tt {
-		assert.Equal(t,
-			test.wantValue,
-			updater.HistoricalPriceAt("btc", "USD", test.at), "at = %s", test.at)
+		assert.InDelta(t, test.wantValue, updater.HistoricalPriceAt("btc", "USD", test.at), 0, "at = %s", test.at)
```
benma, line 45: This is an instance of making code less readable to please the linter. Can the `InDelta` check only be disabled?

benma (changes requested):
Very nice, thanks a lot! I left two small nits.

backend/rates/history.go
baizon: Yes, this is a better solution imho. Also the reassignment `updater.history[bucketName] = allRates` is not useful and just more memory association. I've applied your change.

baizon (commented):
Build in review comments.

backend/rates/history_test.go
```diff
@@ -42,7 +42,10 @@ func TestPriceAt(t *testing.T) {
 		{7.25, time.Date(2020, 9, 3, 18, 0, 0, 0, time.UTC)},
 	}
 	for _, test := range tt {
-		assert.InDelta(t, test.wantValue, updater.HistoricalPriceAt("btc", "USD", test.at), 0, "at = %s", test.at)
+		//nolint:testifylint
```
benma, line 45: There are 4 other instances of InDelta that you could revert.

benma (changes requested):

backend/rates/history_test.go
baizon: A, ok, i thought you mean it just for that one case. I'll revert it and exclude the rule globally.

benma (commented):
Thanks a lot. Please squash all commits, then we can merge.

baizon:
Done.

benma (approved):
LGTM thanks!

PR #2759, opened 2024-06-10
feat: added trivy repository mode scan

baizon:
Hi,
I've added a Trivy action scanner (in repository mode), to scan for security vulnerabilities. The results would be sent to the security tab (can be changed). I'm using it for all my repositories and recommend it.
If you like the idea, You can ask me anything. A example run can be seen on my fork [here](https://github.com/baizon/bitbox-wallet-app/actions/runs/9444306412/job/26009589307).
For more information see: https://github.com/aquasecurity/trivy-action?tab=readme-ov-file#usage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,24 @@ env:
   GITHUB_BUILD_DIR: ${{github.workspace}}
 
 jobs:
+  trivy:
+    name: trivy
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


benma:
Thanks. The results so far look identical to dependabot, plus a few false positives. I personally think one such tool is enough, so I slightly lean against adding this tool in addition to dependabot. Wdyt @Beerosagos?

benma (approved):

baizon:
Another use case would be, that with every PR that you get you can see if a security vulnerability is introduced. I'm no sure, if dependabot is checking the PR.

benma:
Alright, let's give it a shot. Thanks.

PR #2761, opened 2024-06-11
frontend: transform aopp.tsx to funcitonal component

NicolaLS:
Tested by patching the `Start` function in `backend` like [this](https://github.com/benma/bitbox-wallet-app/commit/0529615c76c1cb6f2845ef6b9fe4a6aeee171269).

~~I actually forgot to remove the commit, but I guess it is good to leave it in there for reviewers to test the PR. I can remove it in the end. (note the delay is 15 seconds when testing).~~ removed it now.

I tested on `webdev` firefox with BB02B keystore and `servewallet-mainnet`. I tested the following `case` statements form the return: `error` `inactive` `user-approval` `awaiting-keystore`  `signing` 

I think `choosing-account` and `syncing` was tested too and looked fine but it was to fast to take a screenshot.

I think I did not test `success` because I don't know a callback to test aopp with.

Some pictures from testing:

aopp-test4 png
aopp-test3
aopp-test2
test-aopp1

### DIFF
--- a/frontends/web/src/components/aopp/aopp.tsx
+++ b/frontends/web/src/components/aopp/aopp.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021 Shift Crypto AG
+ * Copyright 2021-2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,21 +14,20 @@
  * limitations under the License.
  */
 
-import React, { Component, ReactNode } from 'react';
+import React, { ReactNode, useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import * as accountAPI from '@/api/account';
 import * as aoppAPI from '@/api/aopp';
-import { translate, TranslateProps } from '@/decorators/translate';
 import { equal } from '@/utils/equal';
 import { SimpleMarkup } from '@/utils/markup';
 import { View, ViewHeader, ViewContent, ViewButtons } from '@/components/view/view';
 import { Message } from '@/components/message/message';
 import { Button, Field, Label, Select } from '@/components/forms';
 import { CopyableInput } from '@/components/copy/Copy';
-import { Cancel, PointToBitBox02 } from '@/components/icon';
+import { PointToBitBox02 } from '@/components/icon';
 import { VerifyAddress } from './verifyaddress';
 import { Vasp } from './vasp';
 import styles from './aopp.module.css';
-import { TUnsubscribe } from '@/utils/transport-common';
 
 type TProps = {
   children: ReactNode;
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +38,127 @@ const Banner = ({ children }: TProps) => (
...
-      ) {
-        shouldUpdateAccountCodeDefault = true;
-      }
-      return { aopp };
-    }, () => { // callback when state did update
-      if (shouldUpdateAccountCodeDefault) {
-        this.setAccountCodeDefault();
+  useEffect(() => {
+    setAccountCodeDefault(aopp, setAccountCode);
+    unsubscribe.current = aoppAPI.subscribeAOPP(aopp => updateAOPP(aopp));
```
thisconnect, line 74: I think you don't need to store the unsubscribe callback in a ref.
```
const unsubscribe =  aoppAPI.subscribeAOPP...
....
return () => unsubscribe();
```

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +38,127 @@ const Banner = ({ children }: TProps) => (
   <div className={styles.banner}>{children}</div>
 );
 
-type State = {
-  accountCode: accountAPI.AccountCode;
-  aopp?: aoppAPI.Aopp;
-}
-
-type Props = TranslateProps;
+const setAccountCodeDefault = (aopp: aoppAPI.Aopp | undefined, setAccountCode: React.Dispatch<React.SetStateAction<string>>) => {
```
thisconnect, line 41: use newlines for arguments whenever there is something more complicated like this.

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +38,127 @@ const Banner = ({ children }: TProps) => (
...
-      return;
+  useEffect(() => {
+    if (shouldUpdateAccountCodeDefault) {
+      setAccountCodeDefault(aopp, setAccountCode);
     }
-    if (aopp.accounts.length) {
-      this.setState({ accountCode: aopp.accounts[0].code });
-    }
-  }
+  }, [shouldUpdateAccountCodeDefault, aopp]);
```
thisconnect, line 88: I don't see why shouldUpdateAccountCodeDefault is now a state, couldn't you remove the effect and use shouldUpdateAccountCodeDefault just internally in updateAOPP function as it was before?

thisconnect (changes requested):

frontends/web/src/components/aopp/aopp.tsx
NicolaLS: oh true, no idea why I did this..

frontends/web/src/components/aopp/aopp.tsx
NicolaLS: It's a bit subtle but I think it is correct to do it like this. Reason:

Before we used the callback function that can also be passed to `setState(newState, callback)` this function runs after the component rendered (see [this](https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/23e06b19a2e1d358d9363b0498ec14b20661c504/frontends/web/src/components/aopp/aopp.tsx#L81)).

We passed the boolean into this function, which means it will be _carried over_ to the new render, just like you can e.g. capture some variable when creating a go routine. 

The proper way to implement this in a functional component (run some function (the callback) after the render when certain state changed, is by using `useEffect` with a dependency on that state. Now we also want to check the boolean there, but if it was not state or a ref this value would get lost, unlike when we passed it to the callback directly. 

I guess it could be a `ref` though. Should I change it?

NicolaLS:
@thisconnect thanks :) PTAL, I added a comment regarding the `shouldUpdateAccountDefault` explaining the reasoning.

thisconnect (changes requested):

frontends/web/src/components/aopp/aopp.tsx
NicolaLS: Right, we do not need it. We get the `new_aopp` when the effect (now following your suggestion) re-runs because aopp changed and can `setAccountDefault(new_aopp, setAccountCode)` immediately.

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +37,124 @@ const Banner = ({ children }: TProps) => (
...
-      }
-    });
-  }
+  useEffect(() => {
+    setAccountCodeDefault(aopp, setAccountCode);
+    const unsubscribe = aoppAPI.subscribeAOPP(aopp => updateAOPP(aopp));
+    aoppAPI.getAOPP().then(aopp => setAopp(aopp));
+    return unsubscribe;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
NicolaLS: Ah yes makes sense...

NicolaLS:
@thisconnect followed your suggestions and removed the `shouldUpdateAccountDefault` stuff, and tested again.

NicolaLS:
not sure why CI fails, running `make ci` locally succeeds

thisconnect:
> not sure why CI fails, running make ci locally succeeds

not related to your PR, we have a new script that checks for missing placeholders in translations https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/3d34cb6146a36a96c76963b7caaa9bd2019297a5
and also Ci does localize check https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/2a2e2b84325adb8a58a3dc1fa74db158f2305c50/scripts/ci.sh#L35

master was failing for a while as some placeholders in translations were missing, but it should pass since https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2794

thisconnect:
> actually it was a different error, some transient backend/handler test failed for some reason. Re-run was successful.

oups I just assumed and didn't check 😇 sorry

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +37,114 @@ const Banner = ({ children }: TProps) => (
...
-  }
+  const [accountCode, setAccountCode] = useState<accountAPI.AccountCode>('');
+  const [aopp, setAopp] = useState<aoppAPI.Aopp>();
 
-  private updateAOPP(aopp: aoppAPI.Aopp) {
-    let shouldUpdateAccountCodeDefault = false;
-    this.setState(currentState => {
-      if (aopp?.state === 'choosing-account'
+  useEffect(() => {
+    setAccountCodeDefault(aopp, setAccountCode);
```
thisconnect, line 61: looking at the git diff setAccountCodeDefault is confusing me a bit and seems to work differently now.

thisconnect (changes requested):
cc @shonsirsha can you take this review?

frontends/web/src/components/aopp/aopp.tsx
NicolaLS: I defined it outside of the component so `aopp` and `setAccountCode` need to be passed to it but it is essentially the same: `aopp` comes from the state so same as [this](https://github.com/NicolaLS/bitbox-wallet-app/blob/3205b4cfde03f303cd76ff586d69fe9f5a1aa1e7/frontends/web/src/components/aopp/aopp.tsx#L89C7-L89C33). and `setAccountCode` sets the `accountCode` state so same as [this](https://github.com/NicolaLS/bitbox-wallet-app/blob/3205b4cfde03f303cd76ff586d69fe9f5a1aa1e7/frontends/web/src/components/aopp/aopp.tsx#L94).

NicolaLS: I can put it back into the component so it does not need any arguments, I just like the _functional style_ more for some reason but I think it does not really have any impact on anything here (e.g. performance)

shonsirsha:
> cc @shonsirsha can you take this review?

Sorry just saw this! Yes, I can. 👍 

@NicolaLS , is it ready to be re-reviewed? 😇

NicolaLS:
> > cc @shonsirsha can you take this review?
> 
> Sorry just saw this! Yes, I can. 👍
> 
> @NicolaLS , feel free to tag me as a reviwer once it's ready to be re-reviewed 😇 ty

@shonsirsha Thanks, yes it is ready (just rebased)

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +37,114 @@ const Banner = ({ children }: TProps) => (
...
+        verticallyCentered
+        width="580px">
+        <ViewHeader title={t('aopp.errorTitle')}>
+          <p>{domain(aopp.callback)}</p>
+        </ViewHeader>
+        <ViewContent>
+          <Message type="error">
+            <Cancel className={styles.smallIcon} />
+            {t(`error.${aopp.errorCode}`, { host: domain(aopp.callback) })}
+          </Message>
```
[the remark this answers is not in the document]
NicolaLS: I removed the `Cancel` and now it looks like this:

image

Did you mean it is totally unused in the frontend or totally unused in the component ? I'll check if it can be removed globally and do that in a separate commit then.

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -38,164 +37,114 @@ const Banner = ({ children }: TProps) => (
   <div className={styles.banner}>{children}</div>
 );
 
-type State = {
-  accountCode: accountAPI.AccountCode;
-  aopp?: aoppAPI.Aopp;
-}
-
-type Props = TranslateProps;
+const setAccountCodeDefault = (
```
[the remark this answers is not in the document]
NicolaLS: true, I generally like a more functional approach so the dependencies are clear, also technically this improves performance because the function is not defined/initialized on every render but performance does not really matter here xd.

I'm going to put it back in the component

NicolaLS:
(addressed change requests in interactive rebase, might add a commit to remove `Cancel` globally now if possible)

NicolaLS:
@shonsirsha thanks, I addressed everything PTAL.
>This looks strange now with our new Message component. I think we can remove the Cancel icon (and remove it from codebase too if it's totally unused).

I searched for `Cancel` and it is used in `KeystoreConnectPrompt` in the `switch(data.typ) ...case 'error': ..` so I don't think we can remove it. It is also used in `IconAndMessage` <- `MessageWaitDialog` <- `Send`

thisconnect (approved):
LGTM

PR #2763, opened 2024-06-11
chore(deps): update golangci to 1.59.0

baizon:
I just found that updating golangci there is a manual change needed in the Makefile to also bump the version.


benma (approved):
Thx!

PR #2765, opened 2024-06-12
frontend: sell UI components with mocked API calls.

shonsirsha:
Sell UI components with mocked API calls.

Mock API calls:
1. getExchangeSellSupported (only region `CH` is supported)
2. getSellExchangeSupportedAccounts (`ETH` not supported)

Also moved `/buy` dir  to `/exchange` dir, and  control firmware update dialog from AppContext through `firmwareUpdateDialogOpen` - so it's controllable from other pages such as the new sell page.

[Preview]:

Available:
avail

Unavailable due to region:
region

Unavailable due to selected account / coin (using ETH):
cointype

Unavailable due to out-of-date firmware:

https://github.com/BitBoxSwiss/bitbox-wallet-app/assets/28676406/b935cc16-d043-410f-8590-a6f08158b906


frontends/web/src/routes/exchange/components/sell.tsx
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    setExchangeNotAvailableReason(undefined);
+
+  }, [showMoonpay, showPocket, supportedAccounts, accountCode, versionInfo]);
+
+  const constructErrorMessage = () => {
+    switch (exchangeNotAvailableReason) {
+    case 'coinCode':
+      return t('No sell options available for this coin type.');
```
shonsirsha, line 148: These translations will come later

frontends/web/src/routes/exchange/components/sell.tsx
```diff
@@ -0,0 +1,217 @@
+import { useNavigate } from 'react-router-dom';
```
shonsirsha, line 1: This component has many duplicated code from `exchange.tsx` and needs to be refactored (both this and `exchange.tsx`). But not doing it now as it'd be a refactoring PR instead.

frontends/web/src/routes/exchange/exchange.tsx
```diff
@@ -202,39 +213,62 @@ export const Exchange = ({ code, accounts }: TProps) => {
...
+                            </Button>}
+                        <Button
+                          primary
+                          disabled={!selectedExchange}
+                          onClick={goToExchange} >
+                          {t('button.next')}
+                        </Button>
+                      </div>}
+                    </>
+                  }
```
shonsirsha, line 256: This living in `exchange.tsx` isn't ideal. IMO, we should also create a similar component like `sell.tsx` for this and call it `buy.tsx`. So here in `exchange.tsx` should have both `sell` and `buy` component separately and clearly defined.

We can then also elliminate the duplicated useEffects. I'll do all of the refactoring part once the BE is integrated.

frontends/web/src/api/exchanges.ts
```diff
@@ -86,3 +94,11 @@ export const getExchangeBuySupported = (code: AccountCode) => {
     return apiGet(`exchange/buy-supported/${code}`);
   };
 };
+
+export const getExchangeSellSupported = () => {
+  // TODO: Remove mock and use actual endpoint once it's ready
+  return (): Promise<SupportedExchanges> => {
+    return Promise.resolve({ exchanges: ['moonpay'] });
```
thisconnect, line 101: please hardcode to 'pocket' for now.

frontends/web/src/components/sidebar/sidebar.tsx
```diff
@@ -215,7 +215,7 @@ const Sidebar = ({
                   <img draggable={false} src={coins} alt={t('sidebar.exchanges')}/>
                 </div>
                 <span className={style.sidebarLabel}>
-                  {hasOnlyBTCAccounts ? t('accountInfo.buyCTA.buy', { unit: 'Bitcoin' }) : t('sidebar.buy')}
+                  {hasOnlyBTCAccounts ? t('accountInfo.exchangeCTA.exchange', { unit: 'Bitcoin' }) : t('sidebar.exchange')}
```
thisconnect, line 218: here unit is always 'Bitcoin' but in other places it can be 'crypto' but sidebar.exchange is always 'Exchange crypto'.

How about removing sidebar.exchange.

Also maybe consider moving the 'accountInfo.exchangeCTA.exchange' key to 'exchange.exchangeCTA' or similar, so that it is not under accountInfo?

frontends/web/src/routes/exchange/components/sell.tsx
```diff
@@ -0,0 +1,216 @@
...
+                  setFirmwareUpdateDialogOpen(true);
+                  navigate(`/settings/device-settings/${deviceIDs[0]}`);
+                }}
+                transparent>
+                {t('Update now')}
+              </Button>
+            }
+          </div>
+        )}
+        {!exchangeNotAvailableReason && allExchangeDeals && allExchangeDeals.exchanges.map(exchange => exchange.supported && (
```
thisconnect, line 183: for better readability maybe save in a `const hasExchange` or similar?

frontends/web/src/routes/exchange/components/sell.tsx
```diff
@@ -0,0 +1,216 @@
...
+      </div>
+      {!exchangeNotAvailableReason &&
+          <div className={style.buttonsContainer}>
+            {supportedAccounts.length > 1 &&
+                    <Button
+                      className={style.buttonBack}
+                      secondary
+                      onClick={() => navigate('/buy/info')}>
+                      {t('button.back')}
+                    </Button>}
```
thisconnect, line 205: nit: strange indentation, also for readability  `&& ( <JSXstuff> )` instead of `&& <JSXstuff>`

thisconnect (changes requested):
nice, tested a bit and commented

shonsirsha:
@thisconnect , thanks! PTAL

frontends/web/src/api/exchanges.ts
```diff
@@ -33,6 +33,14 @@ export const getExchangesByRegion = (code: string) => {
   };
 };
 
+
+export const getSellExchangesByRegion = () => {
+  return (): Promise<ExchangeRegionList> => {
+    return Promise.resolve({ regions: [{ code: 'CH', isMoonpayEnabled: true, isPocketEnabled: false }] });
+  };
+};
```
thisconnect, line 41: just fyi: We used the "curry" function style in other places if the api fn takes an argument, but then started just doing `useLoad(() => apiCall(withArg))` instead of `useLoad(apiCall(withArg))` (which requires the currying). 
I think we should just use `useLoad(() => apiCall(withArg))` when possible and not curry. wdyt?

frontends/web/src/locales/en/app.json
```diff
@@ -25,6 +25,9 @@
         "start": "Get started by depositing some coins to the wallet or buying directly in the BitBoxApp."
       }
     },
+    "exchangeCTA": {
+      "exchange": "Exchange {{unit}}"
+    },
```
thisconnect, line 30: duplicate of `exchange.exchangeCTA`?

thisconnect (changes requested):

frontends/web/src/locales/en/app.json
shonsirsha: Ups good point - i forgot the remove the old one. Removed :)

frontends/web/src/api/exchanges.ts
shonsirsha: Yeah I agree - basically immediately return the apiGet, instead of returning a promise, then inside of that promise we return the apiGet, right?

So, do this:

```ts
export const getExchangeBuySupported = (code: AccountCode): Promise<SupportedExchanges> => {
   return apiGet(`exchange/buy-supported/${code}`);
};
```

Instead of the current way:
```ts
export const getExchangeBuySupported = (code: AccountCode) => {
  return (): Promise<SupportedExchanges> => {
    return apiGet(`exchange/buy-supported/${code}`);
  };
};
```

is my understanding right?

thisconnect: Yes exactly, and I think we started doing it like this with `useLoad(() => getVersionInfo(deviceID))` in a few places.

shonsirsha: Yup - changes made. Looks cleaner without currying.

frontends/web/src/routes/settings/components/device-settings/firmware-setting.tsx
```diff
@@ -71,11 +72,11 @@ const FirmwareSetting = ({ deviceID, versionInfo, asButton = false }: TProps) =>
         />
       )}
       <UpgradeDialog
-        open={dialogOpen && canUpgrade}
+        open={firmwareUpdateDialogOpen && canUpgrade}
         versionInfo={versionInfo}
         confirming={confirming}
         onUpgradeFirmware={handleUpgradeFirmware}
-        onClose={() => setDialogOpen(false)}
+        onClose={() => setFirmwareUpdateDialogOpen(false)}
```
thisconnect, line 79: > Also control firmware update dialog from AppContext through
`firmwareUpdateDialogOpen` - so it's controllable from other
pages such as the new sell page.

would be very nice if this was a separate commit.

thisconnect (approved):
code LGTM deferring to @Beerosagos

frontends/web/src/routes/settings/components/device-settings/firmware-setting.tsx
shonsirsha: Hehe, okay done :)

frontends/web/src/api/exchanges.ts
```diff
@@ -33,6 +33,12 @@ export const getExchangesByRegion = (code: string) => {
   };
 };
 
+
+export const getSellExchangesByRegion = (): Promise<ExchangeRegionList> => {
```
Beerosagos, line 37: I think that we can avoid this endpoint, at least for now. Pocket sell feature will be available in every region where buy is available.

frontends/web/src/routes/exchange/components/sell.tsx
```diff
@@ -0,0 +1,211 @@
...
+  accountCode,
+  deviceIDs,
+  selectedRegion,
+  onSelectExchange,
+  selectedExchange
+}: TProps) => {
+  const { t } = useTranslation();
+  const { setFirmwareUpdateDialogOpen } = useContext(AppContext);
+
+  const exchangeDeals = useLoad(() => exchangesAPI.getExchangeDeals());
```
Beerosagos, line 37: exchange deals endpoint is now only about buy deals. How do we want to handle this? 

Option 1: We could split into two endpoints getExchangeBuyDeals and getExchangeSellDeals
Option 2: we could add a param field for the type of deal (buy/sell) we are looking for
Option 2: We could add a "type" field to the Deal struct (this looks more cumbersome to me).

I like option 2, wdyt?

frontends/web/src/api/exchanges.ts
shonsirsha: Ah, good to know. Okay!

frontends/web/src/routes/exchange/components/sell.tsx
shonsirsha: Yes, let's do the second option.

shonsirsha:
@Beerosagos do you want to take over here while implementing the BE? Or merge it as it is, and then once the BE is ready (with what we discussed above), we adjust this FE to work with the BE endpoints?

Beerosagos:
> @Beerosagos do you want to take over here while implementing the BE? Or merge it as it is, and then once the BE is ready (with what we discussed above), we adjust this FE to work with the BE endpoints?

Feel free to merge as is, thanks! This way we can go on in parallel on the backend integration and the buy refactor you mentioned in the previous comments

PR #2766, opened 2024-06-12
backend: sort unconfirmed transactions by created time

strmci:
This commit updates the transaction history sorting to handle multiple unconfirmed transactions. These transactions are now sorted by their created time.

### DIFF
--- a/backend/accounts/transaction.go
+++ b/backend/accounts/transaction.go
@@ -131,14 +131,17 @@ func (tx *TransactionData) isConfirmed() bool {
 
 // byHeight defines the methods needed to satisify sort.Interface to sort transactions by their
 // height. Special case for unconfirmed transactions (height <=0), which come last. If the height
-// is the same for two txs, they are sorted by the created (first seen) time instead.
+// is the same for two txs, or both txs are unconfirmed they are sorted by the created (first seen)
+// time instead.
 type byHeight []*TransactionData
 
 func (s byHeight) Len() int { return len(s) }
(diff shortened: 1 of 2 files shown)


backend/accounts/transaction.go
```diff
@@ -131,14 +131,17 @@ func (tx *TransactionData) isConfirmed() bool {
...
+// time instead.
 type byHeight []*TransactionData
 
 func (s byHeight) Len() int { return len(s) }
 func (s byHeight) Less(i, j int) bool {
 	// Secondary sort by the time we've first seen the tx in the app.
-	if s[i].Height == s[j].Height && s[i].CreatedTimestamp != nil && s[j].CreatedTimestamp != nil {
-		return s[i].CreatedTimestamp.Before(*s[j].CreatedTimestamp)
+	if s[i].CreatedTimestamp != nil && s[j].CreatedTimestamp != nil {
+		if (s[i].Height == s[j].Height) || (s[i].Height <= 0 && s[j].Height <= 0) {
```
benma, line 142: How does this change the behavior? Unconfirmed transactions have height=-1 I thought, so the `s[i].Height == s[j].Height` check that was there before  should have caught that, or not?

benma (changes requested):
Please adjust the unit test to cover sorting by CreatedTimeStamp with the various options.

strmci:
> Please adjust the unit test to cover sorting by CreatedTimeStamp with the various options.

@benma updated, PTAL

Beerosagos (approved):
tACK!

PR #2769, opened 2024-06-17
frontend: bump @testing-library/jest-dom version

NicolaLS:
Bump the version of the @testing-library/jest-dom to the latest version 6.4.6.

There should be no breaking changes (since only minor version was updated), see the CHANGELOG:
https://github.com/testing-library/jest-dom/releases

Tested by running `act` locally since this is for testing this should be sufficient.


thisconnect (approved):
LGTM

PR #2771, opened 2024-06-18
frontend: bump @testing-library/react version

NicolaLS:
Bump @testing-library/react to latest version 16.0.0. Also installs @testing-library/dom as a peer dependency, see:
https://github.com/testing-library/react-testing-library/releases/tag/v16.0.0

There are breaking changes from 15.0.0 to 16.0.0 (none from 14 to 15) but they do not cause any problems, see link above. Tested with CI and `webdev` (just opened a software keystore and clicked on some stuff / `webdev` not really relevant to test this)

### DIFF
--- a/frontends/web/package-lock.json
+++ b/frontends/web/package-lock.json
@@ -25,7 +25,7 @@
       "devDependencies": {
         "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
         "@testing-library/jest-dom": "^6.4.6",
-        "@testing-library/react": "^14.2.1",
+        "@testing-library/react": "^16.0.0",
         "@testing-library/user-event": "^14.5.2",
         "@types/node": "^20.12.7",
         "@types/react": "18.3.3",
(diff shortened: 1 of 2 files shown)


NicolaLS:
>@testing-library/dom was moved to a peer dependency and needs to be explicitly installed.

I'm a bit confused what this means, isn't it pulled in automatically when it is declared as a peer dependency? `npm ls @testing-library/dom` shows it is installed now but from that sentence it sounds like we are supposed to explicitly put it in the `package.json`? @thisconnect

thisconnect:
> @types/react-dom needs to be installed if you're typechecking files using @testing-library/react

yeah I think so. deferring to @shonsirsha

shonsirsha:
Sorry, @NicolaLS  - I was in the wrong dir. I tried `npm ls @testing-library/dom` and indeed it shows that it's installed. Even after removing both `node_modules` and `packge-lock`.

The answer to your question is most likely to be [this](https://github.com/npm/feedback/discussions/225). My npm version is `9.5.0` so I expected this behaviour. I'm assuming your npm ver is >= 7?

So yeah I guess it's fine to leave it without adding that pkg explicitly in the `package.json` - unless we want to control the version of the `@testing-library/dom`.  However, it seems that all tests are still running fine so no need for now I'd say.

NicolaLS:
> Sorry, @NicolaLS - I was in the wrong dir. I tried `npm ls @testing-library/dom` and indeed it shows that it's installed. Even after removing both `node_modules` and `packge-lock`.
> 
> > npm ls @testing-library/dom shows it is installed now but from that sentence ... ?
> 
> The answer to your question is most likely to be [this](https://github.com/npm/feedback/discussions/225). My npm version is `9.5.0` so I expected this behaviour. I'm assuming your npm ver is >= 7?
> 
> So yeah I guess it's fine to leave it without adding that pkg explicitly in the `package.json` - unless we want to control the version of the `@testing-library/dom`. However, it seems that all tests are still running fine so no need for now I'd say.

Yes my npm is `10.8.1`. Thanks for the article!!

shonsirsha (approved):
LGTM tested webdev & test suites ran smoothly! Thanks.

PR #2776, opened 2024-06-22
frontend: refactor transaction.tsx

NicolaLS:
Refactor the transaction.tsx component to be more concise and easier to understand, also fix some minor nits like using template literals for element class names.

The PR implements all suggestions from #2549 and does not do anything else. **The commits can be reviewed own their own.** But the [first](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2776/commits/3621dcde23cb5579d8e35e2c4feca70eff3349f9) and the [third](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2776/commits/388c932dc78feee5eee0bbc92843427f08472fd7) could be skipped because the changed code is moved in other commits.

The last commit moves, and changes the transaction dialog to its own component which is why everything editing something in `Dialog` prior to that commit can also be reviewed later in the last commit. 

Please note that the many additions are caused by the license and other overhead from extracting sub-components. I think the refactor is almost addition/deletion neutral without those.

### DIFF
--- a/frontends/web/src/components/transactions/components/address-or-txid.tsx
+++ b/frontends/web/src/components/transactions/components/address-or-txid.tsx
@@ -1,0 +1,73 @@
+/**
+ * Copyright 2024 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CopyableInput } from '@/components/copy/Copy';
+import transactionStyle from '@/components/transactions/transactions.module.css';
+import parentStyle from '@/components/transactions/transaction.module.css';
+
+type TPropsTxAddress = {
+  label: string;
+  addresses: string[];
+}
+
+export const TxAddress = ({
+  label,
+  addresses,
+}: TPropsTxAddress) => {
+  return (
+    <div className={transactionStyle.activity}>
+      <span className={parentStyle.label}>
+        {label}
+      </span>
+      <span className={parentStyle.address}>
+        {addresses[0]}
+        {addresses.length > 1 && (
+          <span className={parentStyle.badge}>
+                    (+{addresses.length - 1})
+          </span>
+        )}
+      </span>
+    </div>
+  );
+};
+
+type TPropsTxDetailCopyableValues = {
+  label: string;
+  values: string[];
+}
+
+export const TxDetailCopyableValues = ({
+  label,
+  values,
+}: TPropsTxDetailCopyableValues) => {
+  return (
+    <div className={`${parentStyle.detail} ${parentStyle.addresses}`}>
+      <label>{label}</label>
+      <div className={parentStyle.detailAddresses}>
+        {values.map((addrOrTxID) => (
+          <CopyableInput
+            key={addrOrTxID}
+            alignRight
+            borderLess
+            flexibleHeight
+            className={parentStyle.detailAddress}
+            value={addrOrTxID}
+          />
+        ))}
+      </div>
+    </div>
+  );
+};
--- a/frontends/web/src/components/transactions/components/date.tsx
+++ b/frontends/web/src/components/transactions/components/date.tsx
@@ -1,0 +1,49 @@
+/**
+ * Copyright 2024 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useTranslation } from 'react-i18next';
+import { parseTimeShort } from '@/utils/date';
+import { TxDetail } from './detail';
+import transactionsStyle from '@/components/transactions/transactions.module.css';
+import parentStyle from '@/components/transactions/transaction.module.css';
+
+type TProps = {
+  time: string | null;
+}
+
+export const TxDate = ({ time }: TProps) => {
+  const { i18n, t } = useTranslation();
+  const shortDate = time ? parseTimeShort(time, i18n.language) : '---';
+  return (
+    <div className={transactionsStyle.date}>
+      <span className={parentStyle.columnLabel}>
+        {t('transaction.details.date')}:
+      </span>
+      <span className={parentStyle.date}>{shortDate}</span>
+    </div>
+  );
+};
+
+export const TxDateDetail = ({ time }: TProps) => {
+  const { i18n, t } = useTranslation();
+  const shortDate = time ? parseTimeShort(time, i18n.language) : '---';
+  return (
+    <TxDetail
+      label={t('transaction.details.date')}>
+      {shortDate}
+    </TxDetail>
+  );
+};
--- a/frontends/web/src/components/transactions/components/details.tsx
+++ b/frontends/web/src/components/transactions/components/details.tsx
@@ -1,0 +1,198 @@
+/**
+ * Copyright 2024 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { ITransaction, IAmount, getTransaction } from '@/api/account';
+import { A } from '@/components/anchor/anchor';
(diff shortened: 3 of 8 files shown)


frontends/web/src/components/transactions/components/details.tsx
```diff
@@ -0,0 +1,190 @@
...
+            />
+          </TxDetail>
+          <TxDetail label={t('transaction.confirmation')}>{numConfirmations}</TxDetail>
+          <TxStatus
+            status={status}
+            numConfirmations={numConfirmations}
+            numConfirmationsComplete={numConfirmationsComplete}
+            detail
+          />
+          <TxDate time={time} detail />
```
thisconnect, line 95: Instead of 1 component with 2 layouts depending on detail prob. I think you should just export 2 different components from that module without using detail prop.

thisconnect (changes requested):
partly reviewed

frontends/web/src/components/transactions/components/address-or-txid.tsx
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CopyableInput } from '../../copy/Copy';
+import transactionStyle from '../transactions.module.css';
```
shonsirsha, line 18: With #2781 merged, we should start using absolute imports for imports that are `../` or higher ('`../../` etc..). Thanks 👍

frontends/web/src/components/transactions/components/details.tsx
NicolaLS: yes this makes sense, I also changed it for all other components (that you did not review yet)

NicolaLS:
@thisconnect rebased and applied all suggestions. the commit messages should change too for some commits now, but I'll do that after everything is reviewed.

frontends/web/src/components/transactions/components/date.tsx
```diff
@@ -0,0 +1,55 @@
...
+import parentStyle from '@/components/transactions/transaction.module.css';
+
+const parseTimeShort = (time: string, lang: string) => {
+  const options = {
+    year: 'numeric',
+    month: 'numeric',
+    day: 'numeric',
+  } as Intl.DateTimeFormatOptions;
+  return new Date(Date.parse(time)).toLocaleString(lang, options);
+};
```
shonsirsha, line 29: Can we put this in a `utils` file?

frontends/web/src/components/transactions/components/address-or-txid.tsx
```diff
@@ -0,0 +1,83 @@
...
+  addresses,
+  txid,
+}: TPropsAddressOrTxID) => {
+  const values = addresses ? addresses : [txid];
+
+  return (
+    <div className={`${parentStyle.detail} ${parentStyle.addresses}`}>
+      <label>{label}</label>
+      <div className={parentStyle.detailAddresses}>
+        {values.map((addr_or_txid) => (
```
shonsirsha, line 71: Again, for consistency, let's use camelCase here e.g `addrOrTxId` instead of `snake_case`.

frontends/web/src/components/transactions/components/details.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { Dialog } from '@/components/dialog/dialog';
+import { FiatConversion } from '@/components/rates/rates';
+import { Amount } from '@/components/amount/amount';
+import { Note } from '@/components/transactions/note';
+import { TxDetail } from './detail';
+import { Arrow } from './arrow';
+import { TxDateDetail } from './date';
+import { TxStatusDetail } from './status';
+import { TxAddressOrTxID } from './address-or-txid';
+import parentStyle from '../transaction.module.css';
```
shonsirsha, line 14: Let's use absolute import here :)

NicolaLS:
@shonsirsha thanks! rebased/applied chagnes, PTAL

frontends/web/src/components/transactions/components/details.tsx
```diff
@@ -0,0 +1,187 @@
...
+  sign,
+  typeClassName,
+  feeRatePerKb,
+  explorerURL,
+}: TProps) => {
+  const { t } = useTranslation();
+
+  const transactionInfo = useRef<ITransaction | null>();
+
+  useEffect(() => {
```
NicolaLS, line 55: I just noticed this causes **all** single transactions to be fetched at once instead only when user clicks on the details. I have to add `if(!transactionInfo.current && open)` to prevent this, will do this later.

NicolaLS:
@shonsirsha I added one commit to fix the problem that all transactions are fetched immediately instead of when the details dialog opens, this can be in the `create DetailsDialog` though so we can squash it but I thought it'd be better to do it in another commit for now.

frontends/web/src/components/transactions/components/details.tsx
```diff
@@ -0,0 +1,187 @@
...
+  time,
+  amount,
+  sign,
+  typeClassName,
+  feeRatePerKb,
+  explorerURL,
+}: TProps) => {
+  const { t } = useTranslation();
+
+  const [transactionInfo, setTransactionInfo] = useState<ITransaction | null>();
```
shonsirsha, line 53: `transactionInfo` here has the (implicit) type of `ITransaction | null | undefined`. 

nit: Should we just do the following:

```suggestion
  const [transactionInfo, setTransactionInfo] = useState<ITransaction | null>(null);
```

? 🤔 

Since `getTransaction` returns either `ITransaction | null` anyway (not `undefined`), at least from its type.

frontends/web/src/components/transactions/components/address-or-txid.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <label>{label}</label>
+      <div className={parentStyle.detailAddresses}>
+        {values.map((addrOrTxID) => (
+          <CopyableInput
+            key={addrOrTxID}
+            alignRight
+            borderLess
+            flexibleHeight
+            className={parentStyle.detailAddress}
+            value={addrOrTxID} />
```
shonsirsha, line 78: nit:
```suggestion
            value={addrOrTxID} 
       />
```

😄

shonsirsha (commented):
Code LGTM (some nits). I'll test in a bit

shonsirsha (approved):
LGTM tested. 👍 

I don't have that many transactions, feel free if you'd like to test as well @thisconnect

frontends/web/src/components/transactions/components/details.tsx
NicolaLS: yes definitely

NicolaLS:
(addressed nits)

frontends/web/src/components/transactions/components/details.tsx
thisconnect: looks like this was done in last commit, and fixes and issue that was introduced in 2nd last commit?

suqash both commits?

thisconnect (changes requested):
please also take a look at the commits, it looks to me that all "extract" commits should be squashed into 1 commit. 

and the 2 last commits into 1 commit I made a comment

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -51,97 +50,47 @@ export const Transaction = ({
...
   const sign = ((type === 'send') && '−') || ((type === 'receive') && '+') || '';
   const typeClassName = (status === 'failed' && style.failed) || (type === 'send' && style.send) || (type === 'receive' && style.receive) || '';
-  const sDate = time ? parseTimeShort(time) : '---';
-  const statusText = t(`transaction.status.${status}`);
-  const progress = numConfirmations < numConfirmationsComplete ? (numConfirmations / numConfirmationsComplete) * 100 : 100;
 
   return (
-    <div className={[style.container, index === 0 ? style.first : ''].join(' ')}>
-      <div className={[parentStyle.columns, style.row].join(' ')}>
+    <div className={`${style.container}${index === 0 ? ' ' + style.first : ''}`}>
```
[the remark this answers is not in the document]
NicolaLS: with this there would be a trailing space if `index !==0` but I guess that is not a problem.

frontends/web/src/components/transactions/components/show-details-button.tsx
```diff
@@ -0,0 +1,39 @@
...
+  hideOnMedium?: boolean;
+}
+
+export const ShowDetailsButton = ({
+  onClick,
+  expand,
+  hideOnMedium,
+}: TProps) => {
+  return (
+    <div className={`${transactionStyle.action} ${hideOnMedium ? transactionStyle.hideOnMedium : transactionStyle.showOnMedium}`}>
```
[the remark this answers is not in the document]
NicolaLS: we use this element twice, one time with `hideOnMedium` and one time [without](https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/8b04d02bc69541958d7d62ba42b91c6b7a762374/frontends/web/src/components/transactions/transaction.tsx#L159) it. I think it needs to be like this no?

frontends/web/src/components/transactions/components/address-or-txid.tsx
```diff
@@ -0,0 +1,84 @@
...
+type TPropsAddressOrTxID = (TOnlyTxID | TOnlyAdresses) & {
+  label: string;
+}
+
+export const TxAddressOrTxID = ({
+  label,
+  addresses,
+  txid,
+}: TPropsAddressOrTxID) => {
+  const values = addresses ? addresses : [txid];
```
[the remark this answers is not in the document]
NicolaLS: Oh yes, makes sense now after splitting up into `TxAddress` and `TxAddressOrTxID`

frontends/web/src/components/transactions/components/show-details-button.tsx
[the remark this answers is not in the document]
thisconnect: ah true, I missed this in the old version.

NicolaLS:
@thisconnect thanks :) addressed everything, PTAL

2
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const parseTimeShort = (time: string, lang: string) => {
+  const options: Intl.DateTimeFormatOptions = {
+    year: 'numeric',
+    month: 'numeric',
+    day: 'numeric',
+  } as Intl.DateTimeFormatOptions;
+  return new Date(Date.parse(time)).toLocaleString(lang, options);
+};
```
thisconnect, line 38: remove? :)

NicolaLS: I have no idea how this happened, sorry.

thisconnect: no problem, maybe quickly check your own PR after pushing, then you'd see such accidental commits.

thisconnect (approved):
LGTM thanks

PR #2778, opened 2024-06-24
backend/bitbox02: bundle v9.19.0 firmwares

benma:


Beerosagos (approved):
utACK! Maybe add a line in the changelog?

PR #2786, opened 2024-07-01
frontend: improve firmware update indicator on small screen

thisconnect:
The position of the red dot in settings on mobile looked visually a bit off.

Placed the dot in the settings item on the right so it has same look and style as when we use the red dot in other places in the settings item.


shonsirsha (approved):
LGTM

PR #2788, opened 2024-07-01
frontend: label change

NicolaLS:
Label utxos that are to change addresses from our account (subaccounts
respectively) as change to give the user more information about the utxo
when selecting it in coin control.
requestly-isChange-for-all-tx-small png
requestly-isChange-for-all-tx

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -717,15 +717,7 @@ func (account *Account) Transactions() (accounts.OrderedTransactions, error) {
 	if account.fatalError.Load() {
 		return nil, errp.New("can't call Transactions() after a fatal error")
 	}
-	return account.transactions.Transactions(
-		func(scriptHashHex blockchain.ScriptHashHex) bool {
-			for _, subacc := range account.subaccounts {
-				if subacc.changeAddresses.LookupByScriptHashHex(scriptHashHex) != nil {
-					return true
-				}
-			}
-			return false
-		})
+	return account.transactions.Transactions(account.IsChange)
 }
 
 // GetUnusedReceiveAddresses returns a number of unused addresses. Returns nil if the account is not initialized.
@@ -861,6 +853,7 @@ type SpendableOutput struct {
 	*transactions.SpendableOutput
 	OutPoint wire.OutPoint
 	Address  *addresses.AccountAddress
+	IsChange bool
 }
 
 // SpendableOutputs returns the utxo set, sorted by the value descending.
@@ -873,12 +866,14 @@ func (account *Account) SpendableOutputs() []*SpendableOutput {
 		panic(err)
 	}
 	for outPoint, txOut := range utxos {
+		scriptHashHex := blockchain.NewScriptHashHex(txOut.TxOut.PkScript)
 		result = append(
 			result,
 			&SpendableOutput{
 				OutPoint:        outPoint,
 				SpendableOutput: txOut,
-				Address:         account.getAddress(blockchain.NewScriptHashHex(txOut.TxOut.PkScript)),
+				Address:         account.getAddress(scriptHashHex),
+				IsChange:        account.IsChange(scriptHashHex),
 			})
 	}
 	return sortByAddresses(result)
@@ -905,6 +900,17 @@ func (account *Account) VerifyExtendedPublicKey(signingConfigIndex int) (bool, e
 		)
 	}
 	return false, nil
+}
+
+// IsChange returns true if there is an address corresponding to the provided scriptHashHex in our
+// accounts change address chain. It returns false if no address can be found.
+func (account *Account) IsChange(scriptHashHex blockchain.ScriptHashHex) bool {
+	for _, subacc := range account.subaccounts {
+		if subacc.changeAddresses.LookupByScriptHashHex(scriptHashHex) != nil {
+			return true
+		}
+	}
+	return false
 }
 
 // SignBTCAddress returns an unused address and makes the user sign a message to prove ownership.
--- a/backend/coins/btc/account_test.go
+++ b/backend/coins/btc/account_test.go
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package btc_test
+package btc
 
 import (
 	"crypto/sha256"
@@ -24,7 +24,6 @@ import (
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	accountsTypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
-	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/blockchain"
 	blockchainMock "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/blockchain/mocks"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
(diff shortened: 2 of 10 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -1530,6 +1530,7 @@
     "coincontrol": {
       "address": "Address",
       "addressReused": "Address re-used",
+      "change": "Change",
```
thisconnect, line 1533: hm not sure if this needs to be translated or if we just keep the label as "change", but no strong opinion.

Beerosagos: I think keeping a label makes sense, in case we want to translate. We can always keep the label untranslated across languages.

thisconnect: the other day we had an error something about WC namespace which we found out it makes more sense to not translate. But translators might miss that and translate anyway.

thisconnect: asked a bit around, people would like to see it translated ie. "Wechselgeld" in German.

lets keep as is.

Beerosagos (commented):
LGTM! Only one doubt about the first commit comment:

`Possible false negative in case the user used another
wallet and derived some change address behind the current change gap
limit.`

I think that in case the user received on a change address uncovered by our gap limit he will miss the whole utxo, not only the `change` badge. wdyt?

backend/coins/btc/account.go
```diff
@@ -873,12 +874,19 @@ func (account *Account) SpendableOutputs() []*SpendableOutput {
 		panic(err)
 	}
 	for outPoint, txOut := range utxos {
+		isChange := false
+		for _, subacc := range account.subaccounts {
```
benma, line 878: this is the same code as https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/81b3b257f9bad89ec94b1e6939461b2fc6960cc4/backend/coins/btc/account.go#L722-L726

would be nice to move that to a method on Account, with a unit test, so it can be reused

NicolaLS:
> LGTM! Only one doubt about the first commit comment:
> 
> `Possible false negative in case the user used another wallet and derived some change address behind the current change gap limit.`
> 
> I think that in case the user received on a change address uncovered by our gap limit he will miss the whole utxo, not only the `change` badge. wdyt?

Right, I don't know why I thought he would see the utxo. I'll amend the commit message, thanks.

backend/coins/btc/account.go
NicolaLS: Maybe it would be better to make the [AccountAddress](https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/160d1adebbabd220d83dfcffbfa0a3dda85b7a9b/backend/coins/btc/addresses/address.go#L34) _change aware_ and then use [getAddress](https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/160d1adebbabd220d83dfcffbfa0a3dda85b7a9b/backend/coins/btc/transaction.go#L218) instead of `IsChange`, here is how it could look like:
https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/3b7fa1246e4cc722a908e0cc9eefedc680000dff

I think it makes sense for AccounAdress to know whether it is change or not. Often times we'll want to get some address anyways, and even if we really only care for the `IsChange` value the overhead (looking at the receiveAdresses) does not matter at all since a hash map is used for the lookup (O(1)) I think this would make the code easier to understand. 

Please lmk what you think @benma

NicolaLS: (added three commits after the initial backend commit, can squash them in case we want to do it like this)

thisconnect (approved):
frontend LGTM

thisconnect (changes requested):

frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -141,6 +141,12 @@ export const UTXOs = ({
                           </Badge> :
                           null
                         }
```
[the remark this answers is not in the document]
NicolaLS: good idea
image

backend/coins/btc/account.go
thisconnect: please squash

thisconnect (approved):
frontend LGTM

Beerosagos:
@NicolaLS is this ready for BE review or does it need other changes?

NicolaLS:
> @NicolaLS is this ready for BE review or does it need other changes?

It's ready, I added the `IsChange` method to the account and added a test (also fixed test package name). I still think something like [this](https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/3b7fa1246e4cc722a908e0cc9eefedc680000dff#diff-f35c451e2c99215268316520d28911334d439c389500338e7df38710a3f8e5a7) would be better rather than adding a method to account, but this could be a future PR/refactor too.

backend/coins/btc/account_test.go
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package btc_test
```
Beerosagos, line 16: @benma I guess there was a reason why this package was named `btc_test`, but I'm not sure which was it. Is it ok to move inside `btc` package?

backend/coins/btc/account.go
```diff
@@ -907,6 +902,17 @@ func (account *Account) VerifyExtendedPublicKey(signingConfigIndex int) (bool, e
 	return false, nil
 }
 
+// IsChange returns true if there is an address corresponding to the provided scriptHashHex in our
+// accounts change address chain. It returns false if no addresss can be found.
```
Beerosagos, line 906: typo: addresss

backend/coins/btc/account_test.go
```diff
@@ -190,11 +189,26 @@ func TestSignAddress(t *testing.T) {
...
 	require.Error(t, err)
-	address, signature, err := btc.SignBTCAddress(account, "Hello there", signing.ScriptTypeP2WPKH)
+	address, signature, err := SignBTCAddress(account, "Hello there", signing.ScriptTypeP2WPKH)
 	require.NoError(t, err)
 	require.NotEmpty(t, address)
 	require.Equal(t, base64.StdEncoding.EncodeToString([]byte("signature")), signature)
 
 }
+
+func TestIsChange(t *testing.T) {
```
Beerosagos, line 201: Would be nice also to test for false positives: non-change addresses should not be mistakenly identified as change

Beerosagos (changes requested):

backend/coins/btc/account_test.go
NicolaLS: @benma could you take a look at this ? I can't imagine any reason why tests that test code from a certain package should be outside of that package (unit tests). but if there is/was a reason I can just undo the change in this PR.

I already addressed the other change requests from @Beerosagos

benma: What was the reason to change it in the first place?

Often one wants to test the API of a package, not its internals, which is why the tests are often in a separate package (`..._test`). 

It does not really matter much though, so feel free to change it. Afaik the Go stdlib tests are mostly not in external packages either, and that should be a good reference.

NicolaLS:
@Beerosagos adressed all change requests and @benma is okay with changing the tests package if I understood correctly:
 https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2788#discussion_r1770985912

So should be ready for final review/merge.

Beerosagos (approved):
tACK :pray:

thisconnect:
please add a changelog entry 🙏

NicolaLS:
@Beerosagos can we merge this?

Beerosagos:
> @Beerosagos can we merge this?

Yes, thanks

NicolaLS:
lint fail should be: 
https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/cc9fb570194ea55f427330dcb27c3923b75c0ad1

trivy also intermittent fail, @thisconnect could you run CI again and if it succeeds merge the PR?

thisconnect:
@NicolaLS please rebase

Beerosagos:
> @NicolaLS please rebase

Sorry, I missed your comment before merging :pray:

PR #2790, opened 2024-07-03
frontend: improve translations with crypto, bitcoin or coincode

thisconnect:
In some languages Bitcoin or crypto can have special ending depending
on the context, so translating crypto does not work in all cases.

Added Bitcoin and crypto context with option to show coinCode,
if needed.

The buy title will now use coin code instead of coinname, this is
so that we do not need a different translation key for each altcoin.
With that the buy workflow can show 'Buy Bitcoin' or 'Buy crypto',
and once an account is selected it will show the coin code of the
altcoin (unless it is Bitcoin), i.e. 'Buy LTC'.

For simplicity the guide is just using crypto or bitcoin.
Also buy.info.disclaimer keys need to be updated with coinCode.

Related https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/b617b15427ffac7e0c0e9e1b7c2a8d01b59cefd0

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -454,19 +454,18 @@
       },
       "noExchanges": "Sorry, there are no available exchanges in this region.",
       "region": "Region",
-      "selectRegion": "Not specified",
-      "title": "Buy {{name}}"
+      "selectRegion": "Not specified"
     },
     "info": {
       "continue": "Agree and continue",
       "crypto": "crypto",
       "disclaimer": {
         "intro": [
-          "We partner with MoonPay to offer you a seamless way to buy {{name}} directly within the BitBoxApp. It's just a few clicks.",
-          "MoonPay is a platform that makes it easy and quick to buy {{name}} in over 160+ countries."
+          "We partner with MoonPay to offer you a seamless way to buy {{coinCode}} directly within the BitBoxApp. It's just a few clicks.",
+          "MoonPay is a platform that makes it easy and quick to buy {{coinCode}} in over 160+ countries."
         ],
         "payment": {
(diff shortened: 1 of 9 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -454,19 +454,18 @@
...
         "intro": [
-          "We partner with MoonPay to offer you a seamless way to buy {{name}} directly within the BitBoxApp. It's just a few clicks.",
-          "MoonPay is a platform that makes it easy and quick to buy {{name}} in over 160+ countries."
+          "We partner with MoonPay to offer you a seamless way to buy {{coinCode}} directly within the BitBoxApp. It's just a few clicks.",
+          "MoonPay is a platform that makes it easy and quick to buy {{coinCode}} in over 160+ countries."
         ],
         "payment": {
-          "details": "You can buy {{name}} instantly via MoonPay with the following payment methods. Credit or debit card orders are instant and convenient, but more expensive due to increased chargeback risk. We recommend using the bank transfer option for larger amounts. The minimum fee is 4 USD/EUR or equivalent.",
+          "details": "You can buy {{coinCode}} instantly via MoonPay with the following payment methods. Credit or debit card orders are instant and convenient, but more expensive due to increased chargeback risk. We recommend using the bank transfer option for larger amounts. The minimum fee is 4 USD/EUR or equivalent.",
```
thisconnect, line 468: We usually do not change existing keys in commits, however I thought I keep so I can update locize after merging this. Here and in other places.

thisconnect:
CI isn't happy that I renamed the keys, this is expected and good that it reports this.

shonsirsha (approved):
tested LGTM

PR #2791, opened 2024-07-03
frontend: improve translations with crypto, bitcoin or coincode

thisconnect:
In some languages Bitcoin or crypto can have special ending depending
on the context, so translating crypto does not work in all cases.

Added Bitcoin and crypto context with option to show coinCode,
if needed.

For simplicity the guide is just using crypto or bitcoin.
Also buy.info.disclaimer keys need to be updated with coinCode.

Related b617b15427ffac7e0c0e9e1b7c2a8d01b59cefd0


shonsirsha (approved):
untested LGTM, tested #2790

PR #2795, opened 2024-07-04
backend/handlers: fix err handling in getExchangeBuySupported

Beerosagos:
If getExchangeBuySupported was called after an account was closed it was returning a generic error, causing the frontend to display the generic error popup when disconnecting the bitbox every now and then.

This makes the endpoint to return an empty list of supported exchanges if the account is not available.


Beerosagos:
There is a CI error, but looks unrelated to the PR

thisconnect:
> There is a CI error, but looks unrelated to the PR

yeah sorry, it's our new translation check and will be fixed with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2794

benma (approved):
utACK

PR #2801, opened 2024-07-09
frontend: uninstall deprecated &  unmaintained deps

shonsirsha:
Several dependencies are marked as "deprecated":

```
WARN deprecated @babel/plugin-proposal-private-methods@7.18.6:
WARN deprecated @babel/plugin-proposal-numeric-separator@7.18.6
WARN deprecated @babel/plugin-proposal-nullish-coalescing-operator@7.18.6
WARN deprecated @babel/plugin-proposal-class-properties@7.18.6
WARN deprecated @babel/plugin-proposal-private-property-in-object@7.21.11
WARN deprecated @babel/plugin-proposal-optional-chaining@7.21.0
```

These come from `babel-preset-react-app`, that's used by `eslint-config-react-app`. Another unmaintained dep is `@babel/plugin-proposal-private-property-in-object"`. These pkgs are needed for our eslint `"extends": ["react-app"]` that's most likely added by CRA (create-react-app) which we don't use.

This commit removes those unmaintained pkgs, as well as the usage of `react-app` in our eslint.

Doing all of the above gets rid of the mentioned deprecation warnings and usage of those deps from the project.

To keep linting rules in place, we added new deps for eslint plugins & rules to replace the outdated `react-app` eslint extension.

We need at least node version 18.18.x (recommended: 20.x) to be able to install the new deps.


shonsirsha:
There are still deprecation warnings:

```bash
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory...
npm warn deprecated @humanwhocodes/config-array@0.11.14: Use @eslint/config-array instead
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated @humanwhocodes/object-schema@2.0.3: Use @eslint/object-schema instead
```

These are due to `vite` using an outdated version of `eslint` which requires other dependencies that are also outdated. [See on StackOverflow for full explanation.](https://stackoverflow.com/a/78527383)

### Should we fix this?

> TypeScript-ESLint Maintainer here. Ignore the warning - it's useless here. It's a transitive dependency 7 layers deep and it's only a deprecation warning. 

[See on StackOverflow.](https://stackoverflow.com/a/78527383)
 
Testing on the [ViteJS react-ts template](https://vite.new/react-ts) from the [official ViteJS website](https://vitejs.dev/guide/#trying-vite-online), I can also see the exact same dep warnings:


POJ

Therefore, I conclude that these warnings are "harmless" and we don't need to necessarily need fix them.

thisconnect:
With this change `make webtest` is very noisy and has lots of warnings. 

maybe this is relevant
https://github.com/vitest-dev/vitest/issues/4055#issuecomment-1712792750

shonsirsha:
> With this change `make webtest` is very noisy and has lots of warnings.


Good catch, it was due to `eslint()` being ran on tests. Disabled `eslint()` vite plugin for `webtest`.

frontends/web/.eslintrc.json
```diff
@@ -21,7 +31,21 @@
...
+    "prefer-const": "off",
+    "react/prop-types": "off",
+    "@typescript-eslint/no-unused-vars": "off",
+    "no-extra-boolean-cast": "off",
+    "@typescript-eslint/ban-types": "off",
+    "no-undef": "off",
+    "no-empty": "off",
+    "react/display-name": "off",
+    "@typescript-eslint/ban-ts-comment": "off",
+    "no-async-promise-executor": "off"
```
shonsirsha, line 48: Extra explanation: we had to add this in order to not change anything in our codebase (to make linter happy basically). Our previous eslint extension (that we removed) most likely had these already, so we didn't need to declare it here.

frontends/web/vite.config.mjs
```diff
@@ -16,7 +16,7 @@ export default defineConfig(() => {
     },
     plugins: [
       react(),
-      eslint(),
+      env.mode !== "test" && eslint(),
```
thisconnect, line 19: nit please use single quotes 😇

frontends/web/package.json
```diff
@@ -3,16 +3,19 @@
...
     "@types/react": "18.3.3",
     "@types/react-dom": "18.3.0",
+    "@typescript-eslint/eslint-plugin": "^7.15.0",
+    "@typescript-eslint/parser": "^7.15.0",
     "@vitejs/plugin-react": "^4.0.4",
     "eslint": "^8.57.0",
-    "eslint-config-react-app": "^7.0.1",
+    "eslint-plugin-react": "^7.34.3",
+    "eslint-plugin-react-hooks": "^4.6.2",
+    "eslint-plugin-react-refresh": "^0.4.7",
```
thisconnect, line 18: the old plugin contained one more for testinglibrary, could make sense to add that wdyt?

- https://www.npmjs.com/package/eslint-config-react-app?activeTab=dependencies
- https://www.npmjs.com/package/eslint-plugin-testing-library

frontends/web/.eslintrc.json
```diff
@@ -21,7 +31,21 @@
     "arrow-spacing": "error",
     "space-infix-ops": "error",
     "react/no-unused-prop-types": "error",
-    "react/jsx-equals-spacing": ["error", "never"]
+    "react/jsx-equals-spacing": ["error", "never"],
+    "react/react-in-jsx-scope": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "no-case-declarations": "off",
+    "react/no-children-prop": "off",
+    "prefer-const": "off",
```
thisconnect, line 39: nice.
I randomly tested enabling this.. looks like we could enable some in the future once this is merged. (and maybe once we have less staging branches)

thisconnect (approved):

thisconnect:
LGTM with suggestion

frontends/web/.eslintrc.json
shonsirsha: Good find! This might be the reason why we saw the noise on.`make webtest` and needed this https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2801#issuecomment-2217855467.

frontends/web/package.json
shonsirsha: Good find! This might be the reason why we saw the noise on.`make webtest` and needed this https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2801#issuecomment-2217855467.

PR #2803, opened 2024-07-10
backend/btc: add support for payment request transactions

Beerosagos:
Payment requests have been supported by bitbox02 firmware since v9.19.0. This integrates the feature in the app, mocking a payment request in the Pocket widget section. Once the Pocket sell widget integration is completed the mock function will be removed.

Note: to test this it is necessary to flash a dev device with a FW having the test merchant enabled.

### DIFF
--- a/backend/accounts/account.go
+++ b/backend/accounts/account.go
@@ -33,15 +33,33 @@ type AddressList struct {
 	Addresses  []Address
 }
 
+// TextMemo represents a slip-0024 text memo.
+type TextMemo struct {
+	Note string
+}
+
+// PaymentRequest contains the data needed to fulfill a slip-0024 payment request.
+// Text memos are the only memo type supported, currently.
+type PaymentRequest struct {
+	RecipientName string
+	Memos         []TextMemo
+	Nonce         []byte
+	TotalAmount   uint64
+	Signature     []byte
+	// TxOut is a pointer to the TxOut which will satisfay the payment request.
+	TxOut *wire.TxOut
+}
+
 // TxProposalArgs are the arguments needed when creating a tx proposal.
 type TxProposalArgs struct {
 	RecipientAddress string
 	Amount           coin.SendAmount
 	FeeTargetCode    FeeTargetCode
 	// Only applies if FeeTargetCode == Custom. It is provided in sat/vB for BTC/LTC and Gwei for ETH.
-	CustomFee     string
-	SelectedUTXOs map[wire.OutPoint]struct{}
-	Note          string
+	CustomFee       string
+	SelectedUTXOs   map[wire.OutPoint]struct{}
+	Note            string
+	PaymentRequests []*PaymentRequest
 }
 
 // Interface is the API of a Account.
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -17,6 +17,7 @@ package handlers
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"math/big"
@@ -364,6 +365,51 @@ func (handlers *Handlers) getAccountBalance(*http.Request) (interface{}, error) 
 	}, nil
 }
 
+type slip24Request struct {
+	RecipientName string `json:"recipientName"`
+	Nonce         string `json:"nonce"`
+	Memos         []struct {
+		Type string `json:"type"`
+		Text string `json:"text"`
+	} `json:"memos"`
+	Outputs []struct {
+		Amount  uint64 `json:"amount"`
+		Address string `json:"address"`
+	} `json:"outputs"`
+	Signature string `json:"signature"`
+}
+
+func (slip24 slip24Request) toPaymentRequest() (*accounts.PaymentRequest, error) {
+	if len(slip24.Outputs) != 1 {
+		return nil, errp.New("Missing or multiple payment request output unsupported")
+	}
+
+	if len(slip24.Nonce) > 0 {
+		return nil, errp.New("Nonce value unsupported")
+	}
+
+	sigBytes, err := base64.StdEncoding.DecodeString(slip24.Signature)
+	if err != nil {
+		return nil, err
+	}
+
+	memos := []accounts.TextMemo{}
+	for _, memo := range slip24.Memos {
+		if memo.Type != "text" {
+			return nil, errp.New("Payment request non-text memo unsupported")
+		}
+		memos = append(memos, accounts.TextMemo{Note: memo.Text})
+	}
+
+	return &accounts.PaymentRequest{
+		RecipientName: slip24.RecipientName,
+		Nonce:         nil,
+		Signature:     sigBytes,
+		TotalAmount:   slip24.Outputs[0].Amount,
+		Memos:         memos,
+	}, nil
+}
+
 type sendTxInput struct {
 	accounts.TxProposalArgs
 }
@@ -374,11 +420,12 @@ func (input *sendTxInput) UnmarshalJSON(jsonBytes []byte) error {
 		SendAll   string `json:"sendAll"`
 		FeeTarget string `json:"feeTarget"`
 		// Provided in Sat/vByte for BTC/LTC and in Gwei for ETH.
-		CustomFee     string   `json:"customFee"`
-		Amount        string   `json:"amount"`
-		SelectedUTXOS []string `json:"selectedUTXOS"`
-		Note          string   `json:"note"`
-		Counter       int      `json:"counter"`
+		CustomFee       string         `json:"customFee"`
+		Amount          string         `json:"amount"`
+		SelectedUTXOS   []string       `json:"selectedUTXOS"`
+		Note            string         `json:"note"`
+		Counter         int            `json:"counter"`
+		PaymentRequests *slip24Request `json:"paymentRequest"`
 	}{}
 	if err := json.Unmarshal(jsonBytes, &jsonBody); err != nil {
 		return errp.WithStack(err)
@@ -406,6 +453,13 @@ func (input *sendTxInput) UnmarshalJSON(jsonBytes []byte) error {
 		input.SelectedUTXOs[*outPoint] = struct{}{}
 	}
 	input.Note = jsonBody.Note
+	if jsonBody.PaymentRequests != nil {
+		paymentRequest, err := jsonBody.PaymentRequests.toPaymentRequest()
+		if err != nil {
+			return err
+		}
+		input.PaymentRequests = append(input.PaymentRequests, paymentRequest)
+	}
 	return nil
 }
 
--- a/backend/devices/bitbox02/keystore.go
+++ b/backend/devices/bitbox02/keystore.go
@@ -20,6 +20,7 @@ import (
 	"math/big"
 	"sync"
 
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/blockchain"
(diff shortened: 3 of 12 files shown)


frontends/web/src/api/account.ts
```diff
@@ -296,13 +296,29 @@ export const getReceiveAddressList = (code: AccountCode) => {
...
+// PaymentRequest contains the data needed to fulfill a slip-0024 payment request.
+// Text memos are the only memo type supported, currently.
+export type TPaymentRequest = {
+  recipientName: string;
+  memos: TTextMemo[];
+  nonce: string | null;
+  totalAmount: number;
+  signature: string;
+};
+
```
benma, line 313: Adding this and converting from the [slip24 types](https://github.com/pocketbitcoin/request-address/blob/main/src/slip24.ts#L3) seems like overkill, as they are virtually the same. Why not send the slip24 object directly to the backend and use their format to deserialize?

backend/coins/btc/handlers/handlers.go
```diff
@@ -364,6 +364,52 @@ func (handlers *Handlers) getAccountBalance(*http.Request) (interface{}, error)
...
+		if memo.Type != "text" {
+			return nil, errp.New("Payment request non-text memo unsupported")
+		}
+		paymentRequest.Memos = append(paymentRequest.Memos, accounts.TextMemo{Note: memo.Text})
+	}
+
+	if len(slip24.Outputs) != 1 {
+		return nil, errp.New("Missing or multiple payment request output unsupported")
+	}
+	stringAmount := fmt.Sprintf("%.8f", slip24.Outputs[0].Amount)
```
benma, line 404: Isn't the request-address slip24 amount in sats? Why all of this conversion stuff?

backend/coins/btc/handlers/handlers.go
```diff
@@ -364,6 +364,52 @@ func (handlers *Handlers) getAccountBalance(*http.Request) (interface{}, error)
...
+	Nonce         string `json:"nonce"`
+	Memos         []struct {
+		Type         string `json:"type"`
+		Text         string `json:"text"`
+		Refund       string `json:"refund,omitempty"`
+		CoinPurchase struct {
+			CoinType int    `json:"coinType"`
+			Amount   string `json:"amount,omitempty"`
+			Address  string `json:"address,omitempty"`
+		} `json:"coinPurchase"`
```
benma, line 378: I would skip this as it's unused. By having it here it looks like it's needed.

backend/coins/btc/handlers/handlers.go
```diff
@@ -364,6 +364,52 @@ func (handlers *Handlers) getAccountBalance(*http.Request) (interface{}, error)
...
+	} `json:"memos,omitempty"`
+	Outputs []struct {
+		Amount  float64 `json:"amount"`
+		Address string  `json:"address"`
+	} `json:"outputs,omitempty"`
+	Signature string `json:"signature,omitempty"`
```
benma, line 384: why omitEmpty in all the fields? This skips the fields if empty when marshaling, but we don't serialize this to JSON, right?

backend/accounts/account.go
```diff
@@ -33,15 +33,33 @@ type AddressList struct {
...
 }
 
+// TextMemo represents a slip-0024 text memo.
+type TextMemo struct {
+	Note string `json:"text"`
+}
+
+// PaymentRequest contains the data needed to fulfill a slip-0024 payment request.
+// Text memos are the only memo type supported, currently.
+type PaymentRequest struct {
```
benma, line 43: why does this have json tags? It's not json serialized/deserialized is it?

backend/coins/btc/handlers/handlers.go
Beerosagos: Dang, you're right :man_facepalming: I don't know why I thought it was in BTC

backend/accounts/account.go
```diff
@@ -33,15 +33,33 @@ type AddressList struct {
...
+	RecipientName string     `json:"recipientName"`
+	Memos         []TextMemo `json:"memos"`
+	Nonce         string     `json:"nonce"`
```
benma, line 46: make the nonce `[]byte`

backend/coins/btc/handlers/handlers.go
```diff
@@ -364,6 +364,52 @@ func (handlers *Handlers) getAccountBalance(*http.Request) (interface{}, error)
...
+}
+
+func (slip24 slip24Request) toPaymentRequest() (*accounts.PaymentRequest, error) {
+	paymentRequest := &accounts.PaymentRequest{
+		RecipientName: slip24.RecipientName,
+		Nonce:         slip24.Nonce,
```
benma, line 390: I would put `Nonce: nil` here and return an error if slip24.Nonce is not empty, as we don't support it anyway.

backend/devices/bitbox02/keystore.go
```diff
@@ -452,14 +453,65 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
...
+		sigBytes, err := hex.DecodeString(txPaymentRequest.Signature)
+		if err != nil {
+			return nil, err
+		}
+
+		nonceBytes, err := hex.DecodeString(txPaymentRequest.Nonce)
+		if err != nil {
+			return nil, err
+		}
+
```
benma, line 480: It's not clear the request-address nonce is supposed to be hex-encoded. In my above comment I suggest to just error if it is not empty and use `Nonce []byte`, so you can skip the conversion here.

@davidknezic could you make the nonce type `Uint8Array` for clarity here? Or remove it and re-add it only once we needed.

https://github.com/pocketbitcoin/request-address/blob/main/src/slip24.ts#L5

frontends/web/src/api/account.ts
```diff
@@ -18,6 +18,7 @@ import { apiGet, apiPost } from '@/utils/request';
 import type { ChartData } from '@/routes/account/summary/chart';
 import type { TDetailStatus } from './bitsurance';
 import type { SuccessResponse } from './response';
+import { Slip24 } from 'request-address/dist/slip24';
```
benma, line 21: The /dist/ looks very weird :joy: Surprised this would even work. @davidknezic could you also export the Slip24 type in https://github.com/pocketbitcoin/request-address/blob/main/src/index.ts?

backend/coins/btc/handlers/handlers.go
Beerosagos: I already check for the memo type to be `text`. That should already cover it with a proper error message. I'll just remove the unused stuff.

backend/coins/btc/handlers/handlers.go
Beerosagos: Right, I'll remove it.

backend/accounts/account.go
Beerosagos: Leftover from the initial implementation :smile:

frontends/web/src/api/account.ts
Beerosagos: lol, didn't notice it :sweat_smile:

backend/devices/bitbox02/keystore.go
Beerosagos: Also, could it make sense that the payment request message contains a custom fee that Pocket may want to suggest for the transaction?

backend/accounts/account.go
Beerosagos: Done, and also for the Signature

backend/devices/bitbox02/keystore.go
benma: Not sure. It's not in SLIP24, and to me it feels like the sender should have control over the fee, not the recipient.

benma: @thisconnect the sender in this case is the BitBoxApp, not the user :P It chooses the highest fee rate for now.

backend/coins/btc/handlers/handlers.go
```diff
@@ -364,6 +365,51 @@ func (handlers *Handlers) getAccountBalance(*http.Request) (interface{}, error)
...
+func (slip24 slip24Request) toPaymentRequest() (*accounts.PaymentRequest, error) {
+	if len(slip24.Outputs) != 1 {
+		return nil, errp.New("Missing or multiple payment request output unsupported")
+	}
+
+	if len(slip24.Nonce) > 0 {
+		return nil, errp.New("Nonce value unsupported")
+	}
+
+	sigBytes, err := hex.DecodeString(slip24.Signature)
```
benma, line 391: Per README example, request-address's signature is a base64 encoded DER-encoded signature (~70 bytes).

(The BitBox02 takes a compact 64 byte encoding)

benma (changes requested):

frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -175,6 +176,67 @@ export const Pocket = ({ code }: TProps) => {
...
+      address: message.bitcoinAddress,
+      amount: parsedAmount.amount,
+      feeTarget: 'high',
+      customFee: '',
+      sendAll: 'no',
+      selectedUTXOs: [],
+      paymentRequest: message.slip24,
+    };
+
+    let result = await proposeTx(code, txInput);
```
benma, line 202: Not sure but you probably have to wipe the proposed tx note afterwards (maybe in a finally{} clause?), to avoid the bug that was fixed in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2739

frontends/web/src/api/account.ts
davidknezic: Just published `v0.0.9` which includes an export of the `Slip24` type.

frontends/web/src/routes/exchange/pocket.tsx
Beerosagos: Could not reproduce the bug, but done anyway. It can't harm

Beerosagos:
@benma PTAL

frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -175,6 +176,83 @@ export const Pocket = ({ code }: TProps) => {
...
+    if (result.success) {
+      let txNote = t('buy.pocket.paymentRequestNote') + ' ' + message.slip24.recipientName;
+      await proposeTxNote(code, txNote);
+      sendTx(code).then((sendResult: ISendTx) => {
```
benma, line 206: you are in a nice async function, so please don't mix `await` and `.then` :smile: 

You can do `try { const sendResult = await sendTx(code); ... } finally { ... }`

benma (changes requested):

frontends/web/src/routes/exchange/pocket.tsx
Beerosagos: fixed :+1:

frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -175,6 +176,85 @@ export const Pocket = ({ code }: TProps) => {
     });
   };
 
+  const handlePaymentRequest = useCallback(async (message: PaymentRequestV0Message) => {
```
benma, line 179: I think useCallback here doesn't do anything right? It's not used a dep to any other hook.

benma (approved):
utACK with minor nit

frontends/web/package.json
```diff
@@ -34,7 +34,7 @@
     "react-i18next": "14.1.2",
     "react-router-dom": "6.4.3",
     "react-select": "5.8.0",
-    "request-address": "0.0.6"
+    "request-address": "^0.0.9"
```
thisconnect, line 37: nit: for production deps other deps don't use semver ranges, so we can explicitly decide and test new versions.

frontends/web/src/routes/exchange/pocket.tsx
Beerosagos: Good catch! Thanks

frontends/web/package.json
Beerosagos: Thanks! Updated

PR #2807, opened 2024-07-11
Update Breez SDK to v0.5.0

dangeross:
This PR fixes the checksum issue for the `websocket` dependency, removes currently unused Breez SDK types from the frontend and upgrades the breez-sdk-go dependency to v0.5.0

**[SDK v0.5.0](https://github.com/breez/breez-sdk/releases/tag/0.5.0):** 
- Greenlight signer upgrade.
- Keep closed channels in pending state until sweep.

**[SDK v0.4.2](https://github.com/breez/breez-sdk/releases/tag/0.4.2):** 
- Fix wrong route hints when changing channel policy.
- Fix balance update after receive.


Beerosagos (commented):
Very nice, thanks! 
Just one nit: go.mod and go.sum are updated in two different commits. Is there a reason for that? Otherwise, could you fix it? :pray:

PR #2814, opened 2024-07-17
frontend/dialog: remove disableClose param

benma:
Only used in the factory reset dialog during isConfirming, but when isConfirming is true, the FactoryResetWaitDialog is shown and the original dialog is hidden anyway.


thisconnect (approved):
untested LGTM

PR #2815, opened 2024-07-17
frontend: error on react hook exhaustive deps

thisconnect:
This should error if some deps are missing when using useEffect.


shonsirsha:
Thanks. Can you add in `vite.config.mjs`:

`import eslint from 'vite-plugin-eslint'` 

`env.mode !== 'test' && eslint(),` as one of the plugins?

thisconnect:
thanks, added

> Edit: Otherwise vite won't pick up the errors / warnings "reported" by eslint.

shonsirsha (approved):
LGTM

PR #2818, opened 2024-07-17
frontend/dialog: remove disableEscape param, use onClose instead

benma:
If onClose is undefined, there is on close button, and ESC should then also not work, and vice versa.

This also fixes some dialogs that could be escaped via ESC that shouldn't have, like the BB02 backup check dialog.

### DIFF
--- a/frontends/web/src/components/dialog/dialog.tsx
+++ b/frontends/web/src/components/dialog/dialog.tsx
@@ -28,7 +28,6 @@ type TProps = {
     large?: boolean;
     slim?: boolean;
     centered?: boolean;
-    disableEscape?: boolean;
     onClose?: (e?: Event) => void;
     children: React.ReactNode;
     open: boolean;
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/device/bitbox01/restore.tsx
```diff
@@ -137,7 +137,6 @@ class Restore extends Component<Props, State> {
           activeDialog && (
             <DialogLegacy
               title={t('backup.restore.title')}
-              disableEscape={isConfirming || isLoading}
```
thisconnect, line 140: This is a different component DialogLegacy, where you might want to keep it or also apply the change of Dialog component

thisconnect (changes requested):

frontends/web/src/routes/device/bitbox01/restore.tsx
benma: good catch, reverted.

thisconnect (approved):
LGTM

PR #2819, opened 2024-07-18
android: make back button actually go back

benma:
**Behavior**

Before it would always prompt to quit the app. Now, the back button behavior is:

- if a frontend component is using `useBackButton(handler)`, then handler is executed. If the handler returns false, stop. This is useful for dialog components to inhibit the Android back button or make it close the dialog.

- If there is no active frontend back button handler, or a handler was executed and returns true, then the default behavior kicks in:
  - if we can go back in browser history, we go back
  - if we can't we prompt exiting the app

This commit contains this with a few frontend components modified to use the useBackButton hook:
- wait dialogs make the back button do nothing, so user can't exit blocking dialogs when e.g. input on the BitBox02 is required
- close normal dialogs if they can be closed

**Future work to fix history back behavior everywhere**

There are many places where we need to fix navigate()/history.back() behavior to make history.back() work as expected. This commit contains one such fix in the BB02 passphrase flow. In general we want to navigate(-1) instead of navigate forward on a back/abort click, and navigate(-1) afte a flow finished instead of navigating forward, so the back button does bring you back into the previous process.

Future PRs should fix it everywhere, e.g. "Back" buttons should mostly just do `navigate(-1)`.

**webdev**

In webdev, the history back button has the same behavior as on Android except for where useBackButton() is used, which is basically only in dialogs, so one can test the back button behavior almost completely in webdev. One can call `window.onBackButtonPressed()` in the console invoke an active useBackButton hook handler like Android does if one needs to test that, e.g. in a dialog.

**ALTERNATIVE dimsmissed**:

I spent an obscene amount of time trying to solve intercepting "history back" in the frontend only using a combination of history.pushState() and window.onpopstate(), and made it work, but with downsides: - navigate() or history.pushState() could not be called until the useBackButton hook became incative, which is difficult to get right, e.g. `navigate()` after a blocking BB02 operation would happen while the useBackButton hook was still active to inhibit the back button, leading to chaos.  - onpopstate() only fires after history back was executed and cannot be prevented, so the common workaround is to first do a pushstate(), then intercept it in onpopstate() and pushState() again immediately. Since history.pushState() and history.back() are asynchronous operations, and useCallBack() could be rapidly remounted, it would fall out of sync, so a queue processor was needed.

Generally the code complexity of the frontend-only solution was huge and added requirements for how to use it correctly made me abandon the approach.

**Code of alternative solution that was abandoned for the record**:

```backbutton.ts
import { useContext, useEffect, useRef } from 'react';
import { BackButtonContext, THandler } from '@/contexts/BackButtonContext';

export const useBackButton = (handler: THandler) => {
  const { pushHandler, popHandler } = useContext(BackButtonContext);

  // We don't want to re-trigger the handler effect below when the handler changes, no need to
  // repeat the push/pop pair unnecessarily.
  const handlerRef = useRef<(() => void)>(handler);
  useEffect(() => {
    handlerRef.current = handler;
  }, [handler]);

  useEffect(() => {
    pushHandler(handlerRef.current);
    return popHandler;
  }, [handlerRef, pushHandler, popHandler]);
};

// A convenience component that makes sure useBackButton is only used when the component is rendered.
// This avoids complicated useEffect() uses to make sure useBackButton is only active depending on
// rendering conditions.
// This also is useful if you want to use this hook in a component that is still class-based.
export const UseBackButton = ({ handler }: { handler: THandler }) => {
  useBackButton(handler);
  return null;
};
```

```BackButtonContext.ts
import { ReactNode, useContext, createContext, useEffect, useState, useCallback } from 'react';
import { usePrevious } from '@/hooks/previous';
import { runningOnMobile } from '@/utils/env';
import { AppContext } from './AppContext';

export type THandler = () => void;

type TProps = {
  pushHandler: (handler: THandler) => void;
  popHandler: () => void;
}

export const BackButtonContext = createContext<TProps>({
  pushHandler: () => {
    console.error('pushHandler called out of context');
    return true;
  },
  popHandler: () => {
    console.error('popHandler called out of context');
    return true;
  },
});

let queue: {
  action: 'pushState' | 'back';
  invoke?: () => void;
}[] = [];
let isProcessing = false;
let ignoreNext = 0;

function processQueue() {
  if (isProcessing || queue.length === 0) {
    return;
  }
  isProcessing = true;

  const item = queue.shift();
  if (!item) {
    return;
  }
  if (item.action === 'pushState') {
    console.log('QUEUE PUSH');
    window.history.pushState('BLOCKED', '', window.location.href);
  } else {
    console.log('QUEUE POP');
    window.history.back();
  }
  if (item.invoke) {
    item.invoke();
  }

  // Ensure the browser has time to process each navigation
  setTimeout(() => {
    isProcessing = false;
    processQueue();
  }, 100); // Adjust delay based on testing
}

type TProviderProps = {
  children: ReactNode;
}

export const BackButtonProvider = ({ children }: TProviderProps) => {
  const [handlers, sethandlers] = useState<THandler[]>([]);
  const { guideShown, setGuideShown } = useContext(AppContext);
  const previousGuideShown = usePrevious(guideShown);

  const callTopHandler = useCallback(() => {
    console.log('CALL TOP', handlers.length);

    if (handlers.length > 0) {
      const topHandler = handlers[handlers.length - 1];
      topHandler();
      return true;
    }
    return false;
  }, [handlers]);

  useEffect(() => {
    const handler = () => {
      if (ignoreNext > 0) {
        console.log('IGNORED', ignoreNext);
        ignoreNext--;
        return;
      }

      if (callTopHandler()) {
        queue.push({ action: 'pushState' });
        processQueue();
      }
    };
    window.addEventListener('popstate', handler);
    return () => {
      window.removeEventListener('popstate', handler);
    };
  }, [callTopHandler]);

  const pushHandler = useCallback((handler: THandler) => {
    console.log('pushHandler');
    sethandlers((prevStack) => [...prevStack, handler]);
    queue.push({ action: 'pushState' });
    processQueue();
  }, []);

  const popHandler = useCallback(() => {
    console.log('popHandler');
    sethandlers((prevStack) => prevStack.slice(0, -1));
    queue.push({ action: 'back' });
    ignoreNext++;
    processQueue();
  }, []);

  // On mobile, the guide covers the whole screen.
  // Make the back button remove the guide first.
  // On desktop the guide does not cover everything and one can keep navigating while it is visible.
  useEffect(() => {
    if (!runningOnMobile) {
      return;
    }
    if (guideShown && !previousGuideShown) {
      pushHandler(() => setGuideShown(false));
    }
    if (!guideShown && previousGuideShown) {
      popHandler();
    }
  }, [pushHandler, popHandler, guideShown, previousGuideShown, setGuideShown]);

  return (
    <BackButtonContext.Provider value={{ pushHandler, popHandler }}>
      {children}
    </BackButtonContext.Provider>
  );
};
```

### DIFF
--- a/frontends/web/src/hooks/backbutton.ts
+++ b/frontends/web/src/hooks/backbutton.ts
@@ -1,0 +1,65 @@
+/**
+ * Copyright 2024 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useContext, useEffect, useRef } from 'react';
+import { BackButtonContext, THandler } from '@/contexts/BackButtonContext';
+
+/**
+ * The Android back button will call this handler while this hook is active.
+ * The handler can perform an action and:
+ * - return false to stop any further action
+ * - return true for Android to perform the default back operation, which is going back in browser
+ * history if possible, or prompting to quit the app.
+*/
+export const useBackButton = (handler: THandler) => {
+  const { pushHandler, popHandler } = useContext(BackButtonContext);
+
+  // We don't want to re-trigger the handler effect below when the handler changes, no need to
+  // repeat the push/pop pair unnecessarily.
+  const handlerRef = useRef<THandler>(handler);
+  useEffect(() => {
+    handlerRef.current = handler;
+  }, [handler]);
+
+  useEffect(() => {
+    const handler = handlerRef.current;
+    pushHandler(handler);
+    return () => popHandler(handler);
+  }, [handlerRef, pushHandler, popHandler]);
+};
+
+/**
+ * A convenience component that makes sure useBackButton is only used when the component is rendered.
+ * This avoids complicated useEffect() uses to make sure useBackButton is only active depending on
+ * rendering conditions.
+ * This also is useful if you want to use this hook in a component that is still class-based.
+ * MUST be unmounted before any calls to `navigate()`.
+ */
+export const UseBackButton = ({ handler }: { handler: THandler }) => {
+  useBackButton(handler);
+  return null;
+};
+
+/**
+ * Same as UseBackButton, but with a default handler that does nothing and disables the Android back
+ * button completely.
+ */
+export const UseDisableBackButton = () => {
+  useBackButton(() => {
+    return false;
+  });
+  return null;
+};
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -582,12 +582,17 @@ public class MainActivity extends AppCompatActivity {
         }
     }
 
-    // The app cannot currently handle the back button action to allow users
-    // to move between screens back and forth. What happens is the app is "moved"
-    // to background as if "home" button were pressed.
-    // To avoid unexpected behaviour, we prompt users and force the app process
-    // to exit which helps with preserving phone's resources by shutting down
-    // all goroutines.
+    // Handle Android back button behavior:
+    //
+    // By default, if the webview can go back in browser history, we do that.
+    // If there is no more history, we prompt the user to quit the app. If
+    // confirmed, the app will be force quit.
+    //
+    // The default behavior can be modified by the frontend via the
+    // window.onBackButtonPressed() function. See the `useBackButton` React
+    // hook. It will be called first, and if it returns false, the default
+    // behavior is prevented, otherwise we proceed with the above default
+    // behavior.
     //
     // Without forced app process exit, some goroutines may remain active even after
     // the app resumption at which point new copies of goroutines are spun up.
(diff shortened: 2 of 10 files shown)


frontends/web/src/hooks/backbutton.ts
```diff
@@ -0,0 +1,60 @@
...
+  // We don't want to re-trigger the handler effect below when the handler changes, no need to
+  // repeat the push/pop pair unnecessarily.
+  const handlerRef = useRef<THandler>(handler);
+  useEffect(() => {
+    handlerRef.current = handler;
+  }, [handler]);
+
+  useEffect(() => {
+    pushHandler(handlerRef.current);
+    return popHandler;
```
thisconnect, line 39: popHandler currently removes the last handler in the internal array... but I wonder if for any reason there is more than one useBackButton handlers at the same time and some mount and unmount if just removing the last one is always correct.

Alternative which seems more save would be to pass the handler also to popHandler and then let the function search for this exact function and remove just that funciton from the array.

We have something similar here: https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/32bc3d6f4f91af7cb21abb5391b3263687f50e12/frontends/web/src/utils/event.ts#L81-L82

frontends/web/src/hooks/backbutton.ts
```diff
@@ -0,0 +1,60 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useContext, useEffect, useRef } from 'react';
+import { BackButtonContext, THandler } from '@/contexts/BackButtonContext';
+
+/*
```
thisconnect, line 20:
```suggestion
/**
```

frontends/web/src/hooks/backbutton.ts
```diff
@@ -0,0 +1,60 @@
...
+    return popHandler;
+  }, [handlerRef, pushHandler, popHandler]);
+};
+
+// A convenience component that makes sure useBackButton is only used when the component is rendered.
+// This avoids complicated useEffect() uses to make sure useBackButton is only active depending on
+// rendering conditions.
+// This also is useful if you want to use this hook in a component that is still class-based.
+// MUST be unmounted before any calls to `navigate()`.
+export const UseBackButton = ({ handler }: { handler: THandler }) => {
```
thisconnect, line 48: please change to docstring

frontends/web/src/hooks/backbutton.ts
```diff
@@ -0,0 +1,60 @@
...
+  useBackButton(handler);
+  return null;
+};
+
+// Same as UseBackButton, but with a default handler that does nothing and disables the Android back
+// button completely.
```
thisconnect, line 54: please change to docstring

thisconnect (approved):
very nice with question and some docstring nit other than that LGTM

thisconnect (approved):
thank you LGTM
- tested history-back after abort-passphrase
- history-back during wait-dialog on confirm-enable-passphrase

PR #2822, opened 2024-07-22
frontend/exchange: refactor exchange deals logic

Beerosagos:
Exchange logic was largely implemented in the frontend. With this, all the logic is moved to the backend, and the whole code is simplified.

Main changes:
- getExchangeDeals now has three params: coin, region and action (buy/sell) and returns already filtered exchange deals, with best deal flag already set
- mv sell.tsx buysell.tsx: this component now has an action param, that allows to display different deals on the base of the enabled buy/sell tab.
- all the /buy/ path has been moved to /exchange/
- a lot of frontend logic (e.g. to find the best deal) is not necessary anymore and has been removed.
- pocket sell mockup button removed, as it is not necessary anymore

Still todo: we should check if the bb02 firmware supports silent payments and can sell through Pocket using a backend endpoint instead of just checking the version in the frontend. This will be done in a future commit.


Beerosagos:
ping @benma for BE review :) :pray:

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -0,0 +1,155 @@
...
+  action,
+  setInfo,
+}: TProps) => {
+  const { t } = useTranslation();
+  const { setFirmwareUpdateDialogOpen } = useContext(AppContext);
+
+  const exchangeDeals = useLoad(() => exchangesAPI.getExchangeDeals(action, accountCode, selectedRegion), [action, selectedRegion]);
+  const [exchangeDealsError, setExchangeDealsError] = useState('');
+  const [updateFirmwareToSell, setUpdateFirmwareToSell] = useState(false);
+  // FIXME: we should make a BE `can-sell` endpoint
```
benma, line 42: don't let your dreams be dreams :stuck_out_tongue: 

Just add `supportsPocketPaymentRequest` or sth like that here: https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/39f90f84abb1326d208447a099a6bf27c607db2c/backend/devices/bitbox02/handlers/handlers.go#L304

benma (approved):
only light review, utACK with nits

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const versionInfo = useLoad(() => getVersion(deviceIDs[0]), [deviceIDs[0]]);
```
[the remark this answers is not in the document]
Beerosagos: yeah this is why I left the comment: I felt this PR was already big enough, and this improvement was out of scope. I was planning to make a second one that adds a `canSell` or `canSilentPayment` or similar to the keystore interface.

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const navigate = useNavigate();
+
+  useEffect(() => {
```
shonsirsha, line 46: Can we add short comments on what each of them does here?:)

Tbh I'm not the biggest fan of using many `useEffect`s in one file (as it can be kinda hard to read) but sometimes it's unavoidable.

We don't hv to do it here, but we can turn these into separate hooks e.g  `useCheckExchangeDealsError()`, `useCheckFirmwareUpdate()` and so on (with those comments) for better readability.

shonsirsha (approved):
FE LGTM, non-blocking small comments

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -0,0 +1,155 @@
...
+    if (exchangeDeals?.success === false) {
```
shonsirsha, line 47: I noticed this small render bug / hiccup: 

https://github.com/user-attachments/assets/f10f593e-89cf-4ad3-87f7-fb0d7ea2a21f

I believe this was caused by the `exhcangeDealsError` state that was set when we click on "Sell" (while it has no sell option). That state wasnt reset on changing the tab (clicking "Buy"). So this error message is tehcnically for "Sell" (? 🤔thus actually an incorrect message).

Suggestion: 
-reset exchangeDealsError state on tab change 
-& optionally add a loading state instead of showing the (inaccurate) error message. 

WDYT?

shonsirsha (changes requested):
Noticed a small render/state bug

frontends/web/src/locales/en/app.json
```diff
@@ -708,6 +708,14 @@
     "wrongKeystore": "Wrong wallet connected. Please make sure to insert the correct device matching this account.",
     "wrongKeystore2": " If you are using the optional passphrase, make sure you have entered the correct passphrase for the account."
   },
+  "exchange": {
+    "buySell": {
+      "coinNotSupported": "No {{action}} options available for this coin type.",
+      "regionNotSupported": "No {{action}} options available for this region.",
+      "updateFirmware": "Update firmware to sell",
+      "updateNow": "Update now"
```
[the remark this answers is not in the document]
Beerosagos: no strong opinion, but I think that specifying that the upgrade is needed to sell could be useful. Otherwise could be a bit weird to get a generic fw upgrade request there, imo.

frontends/web/src/routes/exchange/components/buysell.tsx
Beerosagos: see my other comment below :)

backend/exchanges/exchanges.go
```diff
@@ -119,3 +143,70 @@ func ListExchangesByRegion(account accounts.Interface, httpClient *http.Client)
...
+		return nil, errp.New("CoinNotSupported")
+	}
+
+	var userRegion *ExchangeRegion
+	if len(regionCode) > 0 {
+		exchangesByRegion := ListExchangesByRegion(account, httpClient)
+		for _, region := range exchangesByRegion.Regions {
+			if region.Code == regionCode {
+				// to avoid exporting loop refs
+				region := region
```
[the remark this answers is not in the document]
Beerosagos: yeah, that would be nice. We have the same workaround in other loops too, maybe this deserve its own PR?

frontends/web/src/routes/exchange/components/buysell.tsx
Beerosagos: I removed the whole `exchangeDealsError ` state and am now storing the whole `exchangeDealsResponse` as a state containing both legitimate results and errors. This allowed to remove one of the `useEffects`.

I didn't touch the firmware part as I expect this to be changed to a `useLoad()` that stores a bool for 'supportsSilentPayments' or 'canSell' or smth similar.

frontends/web/src/routes/exchange/components/buysell.tsx
Beerosagos: Nice catch, thanks! I added a loading `Skeleton` and updated the `useLoad` hook to set back the response to `undefined` while waiting for a new result. Hope that is ok.

Beerosagos:
@benma @shonsirsha updated the code and answered the comments, PTAL :pray:

frontends/web/src/locales/en/app.json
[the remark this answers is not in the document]
benma: we should have a general "A firmware update is required to make use of this feature" text, as we have some instances already and there will be more. Having a specific string for all of them seems overkill.

backend/exchanges/exchanges.go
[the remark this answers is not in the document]
benma: It will be changed in 1.22 anyway, and we have a PR for it underway already.

benma (approved):
utACK

shonsirsha (approved):
LGTM thanks!

backend/handlers/handlers.go
```diff
@@ -239,10 +239,10 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/electrum/check", handlers.postElectrumCheck).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/socksproxy/check", handlers.postSocksProxyCheck).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/exchange/by-region/{code}", handlers.getExchangesByRegion).Methods("GET")
-	getAPIRouterNoError(apiRouter)("/exchange/deals", handlers.getExchangeDeals).Methods("GET")
-	getAPIRouterNoError(apiRouter)("/exchange/buy-supported/{code}", handlers.getExchangeBuySupported).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/exchange/deals/{action}/{code}", handlers.getExchangeDeals).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/exchange/supported/{code}", handlers.getExchangeSupported).Methods("GET")
```
benma, line 243: if code is an account code, why are these functions not living in btc/handlers.go? :nerd_face:

Beerosagos: On second thought, there is a reason: both these functions need `backend.HTTPClient()`. In general, all the exchange-related functions that use the account code also need some backend element. The only exception is `getExchangeSupported()` which could be easily moved, but I'm not sure if it makes much sense to keep all the exchange-related handlers in one place, except this. wdyt?

PR #2823, opened 2024-07-22
Frontend use navigate

thisconnect:


thisconnect:
I noticed we sometimes import the useNavigate hook from 'react-router' and sometimes from 'react-router-dom'.
I'll push an additional commit to fix this.

Beerosagos (approved):
untested LGTM :+1:

PR #2826, opened 2024-07-23
frontend: use navigate -1 for back buttons

thisconnect:
Back buttons should not add new history entries and behave same
as native back buttons by going one step back in history.

also: added functional back button component for send view class component

### DIFF
--- a/frontends/web/src/components/backbutton/backbutton.tsx
+++ b/frontends/web/src/components/backbutton/backbutton.tsx
@@ -1,0 +1,55 @@
+/**
+ * Copyright 2024 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ReactNode, useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { useEsc } from '@/hooks/keyboard';
+import { Button } from '@/components/forms/button';
+
+type TBackButton = {
+  children: ReactNode;
+  className?: string;
+  disabled?: boolean;
+  enableEsc?: boolean;
+};
+
+export const BackButton = ({
+  children,
+  className,
+  disabled,
+  enableEsc,
+}: TBackButton) => {
+  const navigate = useNavigate();
+
+  const handleBack = useCallback(() => {
+    if (!disabled) {
+      navigate(-1);
+    }
+  }, [disabled, navigate]);
+
+  useEsc(useCallback(() => enableEsc && handleBack(), [enableEsc, handleBack]));
+
+  return (
+    <Button
+      disabled={disabled}
+      className={className}
+      onClick={handleBack}
+      secondary
+    >
+      {children}
+    </Button>
+  );
+};
--- a/frontends/web/src/components/forms/button.tsx
+++ b/frontends/web/src/components/forms/button.tsx
@@ -37,6 +37,8 @@ type TProps = TButtonStyleProp & {
   children: ReactNode;
 }
 
+type TButtonLink = LinkProps & TProps;
+
 export const ButtonLink = ({
   primary,
   secondary,
@@ -46,14 +48,14 @@ export const ButtonLink = ({
   children,
   disabled,
   ...props
-}: LinkProps & TProps) => {
+}: TButtonLink) => {
   const classNames = [
     style[
       (primary && 'primary')
-            || (secondary && 'secondary')
-            || (transparent && 'transparent')
-            || (danger && 'danger')
-            || 'button'
+      || (secondary && 'secondary')
+      || (transparent && 'transparent')
+      || (danger && 'danger')
+      || 'button'
     ], className
   ].join(' ');
 
@@ -75,6 +77,8 @@ export const ButtonLink = ({
   );
 };
 
+type TButton = TProps & ComponentPropsWithoutRef<'button'>;
+
 export const Button = ({
   type = 'button',
   primary,
@@ -84,13 +88,14 @@ export const Button = ({
   className = '',
   children,
   ...props
-}: TProps & ComponentPropsWithoutRef<'button'>) => {
+}: TButton) => {
   const classNames = [
-    style[(primary && 'primary')
-            || (secondary && 'secondary')
-            || (transparent && 'transparent')
-            || (danger && 'danger')
-            || 'button'
+    style[
+      (primary && 'primary')
+      || (secondary && 'secondary')
+      || (transparent && 'transparent')
+      || (danger && 'danger')
+      || 'button'
     ], className
   ].join(' ');
 
--- a/frontends/web/src/routes/account/receive/receive.tsx
+++ b/frontends/web/src/routes/account/receive/receive.tsx
@@ -1,6 +1,6 @@
 /**
  * Copyright 2018 Shift Devices AG
- * Copyright 2023 Shift Crypto AG
+ * Copyright 2023-2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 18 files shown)


frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -280,11 +280,11 @@ export const Receive = ({
                       primary>
                       {t('receive.verifyBitBox02')}
                     </Button>
-                    <ButtonLink
+                    <Button
```
benma, line 283: Why don't you use the Backbutton component here and in other places?

frontends/web/src/routes/account/send/components/back.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useNavigate } from 'react-router-dom';
+import { Button, TButton } from '@/components/forms/button';
+
+type TBackButton = Omit<TButton, 'onClick'>;
```
benma, line 20: you don't actually propagate all the other props to <Button>, so why this type? you only need `children`.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -637,11 +638,9 @@ class Send extends Component<Props, State> {
                         disabled={!this.getValidTxInputData() || !valid || isUpdatingProposal}>
                         {t('send.button')}
                       </Button>
-                      <ButtonLink
-                        secondary
-                        to={`/account/${code}`}>
+                      <BackButton secondary>
```
benma, line 641: I am confused, `secondary` is hardcoded in BackButton and this `secondary` here is unused I think. Why is nothing complaining about unused props?

benma (changes requested):

frontends/web/src/routes/account/receive/receive.tsx
thisconnect: BackButton is just a simple wrapper so we can use the hook in class based component. 

In one place it shares the onClick function with useEsc and I guess we do that more in the future. So should the new BackButton do this as well? what if there is another popup that listens to esc etc?

I prefer to keep it simple for now, but no strong opinion.

frontends/web/src/routes/account/send/components/back.tsx
thisconnect: good catch it should pass props.. but lets just find out if we want BackButton everywhere or if this is just a temporary component that can be removed once Send is a functional component.

frontends/web/src/routes/account/send/send.tsx
thisconnect: TButton accepts this prop. This is a mistake, I'll update once we decide if BackButton has a future or is just temp.

benma: Imo it has a future just so the `navigate(-1)` onclick thing does not have to be repeated everywhere. then we could also easily hide it completely on Android if we wanted to.

thisconnect: ack 👍

frontends/web/src/routes/account/send/services.ts
```diff
@@ -37,8 +37,7 @@ export const txProposalErrorHandling = (registerEvents: () => void, unregisterEv
     return { feeError: t(`send.error.${errorCode}`) };
   default:
     if (errorCode) {
-      unregisterEvents();
-      alertUser(errorCode, { callback: registerEvents });
+      alertUser(errorCode);
```
thisconnect, line 40: this has the side-effect that if an unknown error occurred and Alert is shown the user can hit the esc key and go back.

thisconnect:
@benma I pushed one more commit that updates BackButton and added canGoBack prop, so each component can pass a boolean if the back button should be active or disabled. Should also address the other comments about passong props.

I'm not sure if you had something like this in mind.

I keep it as separate commit and if ok I'll update all other views to use this BackButton.

Or let me know how you'd like the BackButton to work.

frontends/web/src/components/backbutton/backbutton.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ComponentPropsWithoutRef, useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { useEsc } from '@/hooks/keyboard';
+import { Button } from '@/components/forms/button';
+
+type TBackButton = Omit<ComponentPropsWithoutRef<'button'>, 'onClick'> & {
```
benma, line 22: shouldn't this be using button.tsx:TProps as you are passing props to the Button component?

frontends/web/src/components/backbutton/backbutton.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }, [canGoBack, navigate]);
+
+  useEsc(handleBack);
+
+  return (
+    <Button
+      secondary
+      onClick={handleBack}
+      disabled={!canGoBack}
+      {...props}>
```
benma, line 45: Looks like `disabled` can be passed in these props too or not? Could be confusing having both canGoBack and disabled. Imho I'd just remove `canGoBack` and use `disabled` instead where needed.

benma (commented):
I still think having ESC behave like a back button everywhere is not normal and not intuitive. ESC in the send screen is probably a bit weird too, but one could consider send/receive as kind of dialogs on the account (even if they are not). But going e.g setings->device-settings->manage accounts, ESC going back to device settings is not intuitive at all.

Maybe just add a `enableEsc` prop to enable it in the send screen to not change current behavior for now?

frontends/web/src/components/backbutton/backbutton.tsx
thisconnect: Yes it probably should, but I tried and could not get it to work, as TButton has a bit a special type for `primary` OR `secondary` OR `transparent`.

And BackButton should always be `secondary` anyway, and never have onClick. So I figured it is not worth bothering extending the type of TButton.

frontends/web/src/components/backbutton/backbutton.tsx
thisconnect: makes sense

frontends/web/src/components/backbutton/backbutton.tsx
benma: why even pass on props in the first place? maybe better just make an explicit list of props that are needed for clarity. Or maybe Pick<> instead of Omit<> for the ones that we actually use and want to pass on.

thisconnect: Pushed a commit to simplify BackButton, I think disabled, enableEsc, className and children are sufficient.

Children is probably not really necessary, but I like the flexibility so that the parent component can decide to use button.back or "cancel" abort etc. depending on the view/popup

frontends/web/src/components/forms/button.tsx
```diff
@@ -84,13 +88,14 @@ export const Button = ({
...
-            || (danger && 'danger')
-            || 'button'
+    style[
+      (primary && 'primary')
+      || (secondary && 'secondary')
+      || (transparent && 'transparent')
+      || (danger && 'danger')
+      || 'button'
     ], className
   ].join(' ');
```
thisconnect, line 101: I kept this changes, but should probably move to a separate commit.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -637,11 +625,11 @@ class Send extends Component<Props, State> {
                         disabled={!this.getValidTxInputData() || !valid || isUpdatingProposal}>
                         {t('send.button')}
                       </Button>
-                      <ButtonLink
-                        secondary
-                        to={`/account/${code}`}>
+                      <BackButton
+                        disabled={activeCoinControl || activeScanQR}
+                        enableEsc>
```
thisconnect, line 630: Setting the button to disabled it not really needed as the UI elements are covered by either coin-control or qr-scan popups.

Similar situation as in receive, but there it currently only conditionally sets enableEsc `<BackButton enableEsc={!addressTypeDialog && !verifying}>`

thisconnect:
I'll make nicer commits when we are happy with the props. @benma PTAL

frontends/web/src/components/backbutton/backbutton.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: TBackButton) => {
+  const navigate = useNavigate();
+
+  const handleBack = useCallback(() => {
+    if (!disabled) {
+      navigate(-1);
+    }
+  }, [disabled, navigate]);
+
+  useEsc(() => enableEsc && handleBack());
```
benma, line 43: i am never quite sure how it works, but maybe you need useCallback(() => ...)) here to avoid re-renders?

benma (approved):
utACK

frontends/web/src/components/backbutton/backbutton.tsx
thisconnect: I tested a bit and found a perf issue, but the other way around. 

If there are many re-renders this causes the internal useKeyboard hook to remove and re-attach the keyboad event listener on document.

On transaction overview this re-attached the event over 100 times (1 per tx). Added a commit to address this and not render Dialog when there is no data.

PR #2829, opened 2024-07-26
backend/bitbox02/keystore: use bitbox02-api-go util function

benma:
Updated dep in 6a75faeb3, we can use a new function added there: `NewBTCPrevTxFromBtcd()`.


Beerosagos (approved):
tACK

PR #2831, opened 2024-07-29
frontend: remove "Back" text from back button in mobile header

shonsirsha:
We removed it as it's not necessarry to have the text especially since it overlaps with other components in certain ocassions.

Preview:
sads


thisconnect (approved):
untested LGTM

PR #2834, opened 2024-07-29
frontend: add history entry navigating to add-account

thisconnect:
Navigating to add-account should add a history entry so that back goes back to manage-accounts.


shonsirsha (approved):
LGTM

PR #2836, opened 2024-07-29
backend/transaction: auto select highest fee for payment requests

Beerosagos:
There was a bug in the payment request implementation that didn't correctly support mempool.space fees and resulted in a fee error during the tx proposal on production environment.

This commit moves the fee target selection for payment requests in the BE and fixes the bug.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -480,7 +480,7 @@ func (account *Account) Notifier() accounts.Notifier {
 // For the other coins or in case mempool.space is not available it fallbacks on Bitcoin Core.
 // The minimum relay fee is used as a last resource fallback in case also Bitcoin Core is
 // unavailable.
-func (account *Account) feeTargets() []*FeeTarget {
+func (account *Account) feeTargets() FeeTargets {
 	// for mainnet BTC we fetch mempool.space fees, as they should be more reliable.
 	var mempoolFees *accounts.MempoolSpaceFees
 	if account.coin.Code() == coin.CodeBTC {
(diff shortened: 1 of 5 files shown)


backend/coins/btc/transaction.go
```diff
@@ -60,7 +60,10 @@ func (account *Account) getFeePerKb(args *accounts.TxProposalArgs) (btcutil.Amou
 	}
 	var feeTarget *FeeTarget
 	for _, target := range account.feeTargets() {
-		if target.code == args.FeeTargetCode {
+		if len(args.PaymentRequests) > 0 {
+			// for payment requests we just take the last element of the array, which is the highest fee target
+			feeTarget = target
```
benma, line 65: This is a bit fragile as it relies on implicit assumptions. It would be better to make `type FeeTargets []*FeeTarget`, then add a `highest()` method on it so it is explicit.

frontend still has `feeTarget: 'high',` which is confusing. Could it be removed, or if not, add a comment that the presence of a paymentRequest overrides it.

benma (changes requested):

backend/coins/btc/transaction.go
Beerosagos: Great suggestion! Done :+1: 

I also changed the feeTarget in the FE from `high` to `custom` and added a comment to clarify :pray:

backend/coins/btc/account.go
```diff
@@ -475,12 +475,36 @@ func (account *Account) Notifier() accounts.Notifier {
 	return account.notifier
 }
 
+// FeeTargets represents an array of FeeTarget pointers.
+type FeeTargets []*FeeTarget
+
+// Highest returns the feeTarget with the highest fee.
+func (feeTargets FeeTargets) Highest() *FeeTarget {
```
benma, line 482: unexport `highest()`, it's only used internally.

backend/coins/btc/account.go
```diff
@@ -475,12 +475,36 @@ func (account *Account) Notifier() accounts.Notifier {
...
+	if len(feeTargets) < 1 {
+		return nil
+	}
+
+	highestFeeIndex := 0
+	var highestFee int64
```
benma, line 488: i think the below loop does not work if all fee targets are nil.

you could fix it by having a `var highestFeeTarget *FeeTarget`, and in the loop do something simliar to this:

```
if feeTarget == nil || feeTarget.feeRatePerKb == nil {
			continue
}
if highestFeeTarget == nil || *feeTarget.feeRatePerKb > *highestFeeTarget.feeRatePerKb {
    highestFeeTarget = feeTarget
}
```

benma (changes requested):

backend/coins/btc/account.go
Beerosagos: nice catch, thanks!

Beerosagos:
@benma all done, PTAL :pray:

backend/coins/btc/account.go
```diff
@@ -475,12 +475,34 @@ func (account *Account) Notifier() accounts.Notifier {
...
 
+// FeeTargets represents an array of FeeTarget pointers.
+type FeeTargets []*FeeTarget
+
+// highest returns the feeTarget with the highest fee.
+func (feeTargets FeeTargets) highest() *FeeTarget {
+	if len(feeTargets) < 1 {
+		return nil
+	}
+
```
benma, line 486: technically not needed any longer

benma (approved):
utACK. unit tests would be nice as usual :innocent:

Beerosagos:
Moved all the `FeeTargets` code to `coins/btc/feetarget.go` and added unit tests. Thanks :pray:

PR #2838, opened 2024-07-31
frontend: back behavior

thisconnect:
- cancel keystore prompt on back
- block back button while verifying receive address


shonsirsha (approved):
lgtm

PR #2839, opened 2024-07-31
frontend: move bip85 to its own route

thisconnect:
Similiar to the passphrase workflow, the bip85 workflow should have its own route.

So the back button goes back to Manage-device where the user came from and navigate(-1) for abort/success to get back.


shonsirsha (approved):
LGTM

PR #2840, opened 2024-07-31
frontend: fix guide open on app restart

thisconnect:
After closing the guide by going back, the guide opens on next app
restart.

Reason was BackButtonProvider closed the guide by calling:
setGuideShown but this function doesn't update the config and
should probably be internal to AppProvider.


shonsirsha (approved):
Good find! Lgtm

PR #2841, opened 2024-08-01
frontend/android: fix startup crash

Beerosagos:
A recent upgrade to targetSdkVersion 34 (ffc030aa6) caused a crash when starting the app. This fixes the crash adding permission requests in the manifest and updating the `registerReceiver` call in `MainActivity` to v34 requirements.


benma:
Damn Android is so fragile :sweat_smile: thanks a lot!

Issue #1728, opened 2022-07-25
Feature Request: Option to split consolidated transactions 


chkpnt:
When a transaction contains multiple outputs which belong to the same (BitBox) wallet, this transaction is only shown once in the transaction overview, but with multiple addresses in the transaction details.

This might be the way a "bitcoin transaction" works technically, but from a user perspective, I expect that every receipt to an address is a seperate "transaction".

Therefore, there should be an option to "de-consolidate" those transactions in order to be able to use different notes for the different addresses. Currently, in the export, the note is the same for all "transactions" within that on-chain transaction.

benma:
Interesting. That however could lead to other complications, e.g. that potentially the tx fees would be listed twice (but only paid once).

Could your problem maybe be solved in other ways? For example, make the notes field multiline so you can enter more elaborate notes.

chkpnt:
As it's about incoming transactions, there is no fee.

If the notes belong to the output address instead of the transaction, that could be an option. But I can't imagine how to solve this in the UI. From the UI perspective, I still find it more intuitive to have one "transaction" per output address in the transaction overview, as it is in the export.

How did I noticed this issue? I withdrew from an exchange to several addresses with the intention to write different notes to them, as soon as the transactions are visible in the BitBox-App. But as the exchange used one "bitcoin transaction" for my several "transactions", there is only one "transaction" in the transaction overview.

PR #2843, opened 2024-08-05
frontend/backend: show outputs as individual row for receive transactions

NicolaLS:
Closes: #1728 

Display each output of a **receive** transaction as one transaction in the app. The outputs data is only added to the endpoint response if the transaction is of type receive. Non-utxo model transactions e.g. ETH will just contain a single item in the `outputs` field which is weird but fine imo.

In the current implementation the index used for the **note key** (`txid:idx`) is set/defined by the order of the `Adresses` in the `TransactionData` struct returned by `txInfo` ([here](https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/0c82d7b67b8f84c751dd19a2e4095aa0aa4fdde4/backend/coins/btc/transactions/transactions.go#L377))
- if the logic there changes all notes would be broken
- the way the frontend treats the `outputs` field right now kind of implies that we are dealing with the actual `outpoint`/output index of the transaction which is not true and could be confusing when working on the code after some time.

Probably it would be better to add `index` to `TransactionOutput` and actually set it to the transactions output index and then use that in the frontend. Let me know if you think the current implementation is fine or if I should change that.

### DIFF
--- a/backend/accounts/notes/notes.go
+++ b/backend/accounts/notes/notes.go
@@ -112,11 +112,12 @@ func (notes *Notes) SetTxNote(txID string, note string) (bool, error) {
 }
 
 // TxNote fetches a note for a transaction. Returns the empty string if no note was found.
-func (notes *Notes) TxNote(txID string) string {
+// NOTE: noteID is either txid or txid:n, n MIGHT NOT be the outpoint from the bitcoin transaction.
+func (notes *Notes) TxNote(noteID string) string {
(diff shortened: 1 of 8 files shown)


NicolaLS:
This can be tested by checking out #2844 
```bash
./scripts/regtest/run_regtest.sh
make servewallet-regtest
make webdev
cd ./scrips/regtest/
source examples.sh
mine_block 101
example_bitbox_receive_many
```

PR #2845, opened 2024-08-06
frontend: disable native back when showing iframes

thisconnect:
The app should disable native back when the user is in an iframe so that they do not accidentally leave its workflow.


shonsirsha (approved):
LGTM thanks.

PR #2854, opened 2024-08-16
Update Breez SDK to v0.5.2

dangeross:
This PR updates the Breez SDK to v0.5.2 which includes a Greenlight fix to prevent nodes getting stuck (desync). It also adds new API methods to support retrieval of diagnostic data, the service health check, and reporting of payment failures.

- **Diagnostic Data** - Collects data from the SDK state and Greenlight node into text which can be shared for debugging issues.
- **Service Health Check** - Retrieves the current service status from the Breez server.
- **Report Payment Failure** - Can be used to report to the Breez server a payment failure using the failed payment hash. This shares details for the payment and its failure reason with Breez for analysis.

Note that (currently) this PR does not include UI changes to use these API methods, only to expose them so they can be integrated into the payment flow.


Beerosagos (approved):
tACK, thanks @dangeross

PR #2858, opened 2024-08-21
frontend: hide taproot account info

sutterseba:
Using Taproot extended public keys in raw xpub format is not widely supported and can lead to dangerous misuse. This hides the Taproot extended public key and displays an info message instead.

Most wallets will default to P2PKH when scanning a raw P2TR xpub, resulting in a non-standard combination of script type and derivation path. Recovering funds sent to addresses like this is difficult and potentially unsafe for users, as the BitBox02 does not support spending from P2PKH outputs.

As long as the BitBoxApp does not support output descriptors, expert users should instead connect their device to other wallet software to view detailed account information.

Screenshot 2024-08-21 at 09 37 27

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -26,6 +26,7 @@
     "extendedPublicKey": "Extended public key",
     "label": "Account info",
     "scriptType": "Script type",
+    "taproot": "The Taproot extended public key is not shown to protect from accidental misuse, as raw formats for this script type are not widely supported. If you want to view this key, connect your device to wallet software that supports the output descriptor format.",
     "title": "Account information",
     "verify": "Verify on device",
     "xpubTypeChangeBtn": {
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -26,6 +26,7 @@
     "extendedPublicKey": "Extended public key",
     "label": "Account info",
     "scriptType": "Script type",
+    "taproot": "The Taproot extended public key is not shown to protect from accidental misuse, as raw formats for this script type are not widely supported. Expert users who want to view this key should connect their device to wallet software that supports the output descriptor format.",
```
benma, line 29: i recently read somewhere that it's not good to call users users :smile:  Maybe "If you want to view this key ..."?

frontends/web/src/routes/account/info/info.tsx
```diff
@@ -80,16 +81,27 @@ export const Info = ({
...
-                key={viewXPub}
-                account={account}
-                code={code}
-                info={config}
-                signingConfigIndex={viewXPub}>
-                <BackButton enableEsc>
-                  {t('button.back')}
-                </BackButton>
-              </SigningConfiguration>
+              { isBitcoinOnly(account.coinCode) && (viewXPub === 1) ? (
```
benma, line 84: checking the index is fragile. you can check info.signingConfigurations[viewXPub] directly:  `signingConfig.bitcoinSimple && signingConfig.bitcoinSimple.scriptType === 'p2tr'` or something like that

benma (changes requested):
Thanks!

benma (approved):
utACK

PR #2860, opened 2024-08-21
frontend: only pass account to Send component

NicolaLS:
The Send component interacts with one account per transaction which is why it makes sense to refactor the component to only take one account instead of a list of all accounts and the code of the relevant account.

This has several benefits:
- reduce code size
- remove non-null assertions that should not be used (`getAccount()!.coinCode` for example)
- makes intention clear: Send can only take a valid account, if there is no account there is no Send.
- makes Send cleaner/easier to understand by removing prop, helper function, null/undefined checks.

Small PR working towards **convert send.tsx to functional component**. I will open a few small PR's and try to reduce the size of `Send`/make it easier to understand, and then eventually open a bigger PR to convert it to its own functional component.

But it will not be possible to do all refactoring before the big PR, for example a custom hook might be useful but can only be used when the component is functional itself.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -34,7 +34,7 @@ import { getConfig } from '@/utils/config';
 import { FeeTargets } from './feetargets';
 import { signConfirm, signProgress, TSignProgress } from '@/api/devicessync';
 import { UnsubscribeList, unsubscribe } from '@/utils/subscriptions';
-import { isBitcoinBased, findAccount } from '@/routes/account/utils';
+import { isBitcoinBased } from '@/routes/account/utils';
 import { ConfirmingWaitDialog } from './components/dialogs/confirm-wait-dialog';
 import { SendGuide } from './send-guide';
 import { MessageWaitDialog } from './components/dialogs/message-wait-dialog';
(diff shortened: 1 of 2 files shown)


NicolaLS:
returning `null` if `findAccount` returns nothing feels lazy but is actually the graceful outcome of the `Send` component right now:
https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/87d6219ddb566a52aa410c54cd75d63669e5cec6/frontends/web/src/routes/account/send/send.tsx#L525

But right now it could also crash because of the non-null assertion used.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -161,26 +149,22 @@ class Send extends Component<Props, State> {
       signConfirm(() =>
         this.setState({ signConfirm: true })
       ),
-      syncdone((code) => {
-        if (this.props.code === code) {
```
thisconnect, line 165: I think you should keep this condition somehow, haven't tested but maybe `this.props.account.code === code`  or similar?

thisconnect (changes requested):

frontends/web/src/routes/account/send/send.tsx
NicolaLS: changed it in first push, also rebased in second push.

thisconnect (approved):
tested LGTM

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -520,10 +501,7 @@ class Send extends Component<Props, State> {
       signConfirm
     };
 
-    const account = this.getAccount();
-    if (!account) {
-      return null;
-    }
+    const account = this.props.account;
```
thisconnect, line 504: I'll change to desctruct at the beginning of the render function

PR #2861, opened 2024-08-21
frontend/backend: pass final tx. note to `sendtx` and remove `propose-tx-note` handler

NicolaLS:
Noticed this when planning the refactor of (frontend) `send.tsx`, this PR does not change anything and should be a pure refactor.
- the `propose-tx-note` handler did nothing but setting a field on `BaseAccount` it did not validate anything, thus removing it and just using the final tx note reduces the code size but changes nothing.

We initially separated the note from the txProposal because we did not want to perform any actions when (e.g calculate fee) when user types a note:
https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/4eb692c84351dbd23ae391e99ffd5d8452db08f7

It would be good to validate the note though to match the constrains here, but that would not be a refactor so maybe in another PR:
https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/87d6219ddb566a52aa410c54cd75d63669e5cec6/backend/accounts/notes/notes.go#L92

### DIFF
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -73,7 +73,6 @@ func NewHandlers(
 	handleFunc("/verify-extended-public-key", handlers.ensureAccountInitialized(handlers.postVerifyExtendedPublicKey)).Methods("POST")
 	handleFunc("/sign-address", handlers.ensureAccountInitialized(handlers.postSignBTCAddress)).Methods("POST")
 	handleFunc("/has-secure-output", handlers.ensureAccountInitialized(handlers.getHasSecureOutput)).Methods("GET")
-	handleFunc("/propose-tx-note", handlers.ensureAccountInitialized(handlers.postProposeTxNote)).Methods("POST")
 	handleFunc("/notes/tx", handlers.ensureAccountInitialized(handlers.postSetTxNote)).Methods("POST")
 	handleFunc("/connect-keystore", handlers.ensureAccountInitialized(handlers.postConnectKeystore)).Methods("POST")
 	handleFunc("/eth-sign-msg", handlers.ensureAccountInitialized(handlers.postEthSignMsg)).Methods("POST")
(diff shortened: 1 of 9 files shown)


NicolaLS:
(has a one line conflict with #2860)

backend/coins/btc/handlers/handlers.go
```diff
@@ -410,7 +409,11 @@ func (input *sendTxInput) UnmarshalJSON(jsonBytes []byte) error {
 }
 
 func (handlers *Handlers) postAccountSendTx(r *http.Request) (interface{}, error) {
-	err := handlers.account.SendTx()
+	var finalNote string
+	if err := json.NewDecoder(r.Body).Decode(&finalNote); err != nil {
+		handlers.log.WithError(err).Error("Failed to unmarshal transaction note")
```
NicolaLS, line 414: note: I decided to ignore the error on purpose here and only log it. it is better if users contact us with a _bug that their notes do not get applied_ rather then _I can't send any coins anymore_

this should not happen anyways, and if it does proceeding with the zero string value is like the user never entered anything into the note field.

NicolaLS: *what I mean is I don't return an error here and continue on purpose

benma: please add this as a comment, otherwise the next person to look at this will think it's a bug and return an error :smile:

benma (changes requested):
Very nice, this is a great simplification. I am trying to think why I made this very clumsy way of adding the note, and if there was maybe a good reason for it, but I think it was just a blunder :flushed:

benma:
Also it should be one commit, it's one logical change.

NicolaLS:
(rebased)

NicolaLS:
@benma thanks :) I rebased, squashed, and added a comment explaining why we don't return an error when unmarshaling the tx. note fails, PTAL.

benma (approved):
tACK

PR #2863, opened 2024-08-26
frontend: revamp send confirm dialog / view

shonsirsha:
Using `View fullscreen` instead of WaitDialog and UI improvement in general in order to improve UX (easier to read).

a_
b_


frontends/web/src/routes/account/send/components/confirm/confirm.module.css
```diff
@@ -0,0 +1,86 @@
...
+  margin: 0;
+  word-break: break-all;
+}
+
+.confirmItem ul {
+  list-style-type: square;
+  margin: 0;
+  margin-top: var(--space-quarter);
+  padding-left: var(--space-half);
+}
```
shonsirsha, line 29: Lmk if there's already a way we use square bullet points somewhere - or if we wanna style it a certain way (it's pretty "plain" for now).

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -613,8 +613,7 @@ class Send extends Component<Props, State> {
                   </Column>
                 </Grid>
               </ViewContent>
-              <ConfirmingWaitDialog
-                paired={paired}
```
shonsirsha, line 617: Let me know if we also wanna have the conditional render of `paired` somewhere in `ConfirmSend`. 🤔

thisconnect: actually that is for BB01 😬 

Ideally find a solution that we don't have to test BB01 too much.. should we keep confirm-wait-dialog.tsx and conditionally show it for BB01?

or maybe split send into 2 different components like we do for receive https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/bdf270c0bfc824f3dfda91176f76852eb4a524f7/frontends/web/src/routes/account/receive/index.tsx#L34-L36

wdyt?

shonsirsha:
Thanks! PTAL @thisconnect

shonsirsha:
0da
da

Screenshots for second commit (design change; font size & text colour).

thisconnect (approved):
LGTM with design nit

PR #2865, opened 2024-08-26
ci: install nodejs with github actions

thisconnect:
Simplify nodejs setup, use GH actions instead of homebrew and nvm

https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-nodejs


benma (approved):
LGTM

PR #2866, opened 2024-08-26
frontend: remove unused btcUnit from send.tsx

NicolaLS:
Remove unused state variable `btcUnit` from the Send component. Probably it was not noticed that it is not used anymore because the linter does not complain since we set it with `getConfig`.

Small PR in preparation for `send.tsx` refactor.


NicolaLS:
@thisconnect amended the commit message

thisconnect (approved):
untested LGTM

PR #2867, opened 2024-08-26
frontend: remove redundant fiatUnit state in Send

NicolaLS:
Remove the `fiatUnit` state in Send because it is redundant since we can access `this.props.activeCurrency` since this commit: 373b7e89a2cbf6d4cf8f51ac69743b9a7b4efd61

Small PR preparation for `send.tsx` refactor.


thisconnect (approved):
tested LGTM thanks 👍

PR #2868, opened 2024-08-26
frontend: don't convert to fiat twice in send

NicolaLS:
The fiat amount does not have to be calculated when receiving the response from submitting the tx. proposal because the `result.amount` is exactly the same as the `amount` in the components state. When this state is set we already convert to fiat.

### DIFF
--- a/frontends/web/src/routes/account/send/send.tsx
+++ b/frontends/web/src/routes/account/send/send.tsx
@@ -236,7 +236,7 @@ class Send extends Component<Props, State> {
     };
   };
 
-  private validateAndDisplayFee = (updateFiat: boolean = true) => {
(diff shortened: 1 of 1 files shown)


thisconnect:
Unfortunately it is currently needed with this change it introduces a bug:

when the user clicks "Send all" / "Send selected coin" the fiat field isn't updated.

NicolaLS:
> Unfortunately it is currently needed with this change it introduces a bug:
> 
> when the user clicks "Send all" / "Send selected coin" the fiat field isn't updated.

@thisconnect oh right...mb should have noticed this. Looks like there is no way around it actually because we would not want to calculate the "all" amount in the frontend, we just propse with "sendall: true" and then use the returned amount.

PR #2877, opened 2024-09-02
frontend: remove "Security Model" from exchange guide

shonsirsha:
The text is different depending on what service you use, and if one buys or sells. Each service already has the security model text in its disclaimer.

Also changed from "Buy Guide" to "Exchange guide" 

gfea


benma (approved):
utACK

PR #2879, opened 2024-09-02
backend: update to version 4.44.0

benma:
Added missing CHANGELOG entry.
Reworded the Pocket sell CHANGELOG entry, the technical detail is unimportant for the release note. Also the BTC are not sold *to* Pocket, but Pocket sells it for the customer as a broker.


thisconnect (approved):
LGTM

PR #2884, opened 2024-09-02
frontend: use Buy & sell instead of Exchange

shonsirsha:
Also removed unused `generic.exchange` translation key from codebase.
bs1
bs2
bs3


thisconnect (changes requested):

thisconnect (approved):
untested LGTM

PR #2888, opened 2024-09-03
ci: build iOS app

benma:
For now we ignore signing the app, publishing it or even running it. This first step is to simply compile the iOS app to check for build errors.

In the future, we may want to consider using github secrets to automatically sign and push to testflight.

The macos runner image is bumped to 14, which defaults to XCode
15. The older version had some build errors with the iOS app.

I also tried porting the macOS Qt job to run on macos-14 which runs on ARM, using scripts/osx-brew.sh to install the x84_64 version of Qt, which is needed so the app can be built for x84_64. It works but we postpone this until necessary, as it slows down CI a lot (emulates all of compilation in x86_64 on ARM).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,3 +115,37 @@ jobs:
         with:
           path: BitBoxApp-macos.zip
           name: BitBoxApp-macos-${{github.sha}}.zip
+  ios:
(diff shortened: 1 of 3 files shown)


NickeZ (approved):
lgtm

PR #2891, opened 2024-09-04
ios: centralize and set version data

benma:
See https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project.


NickeZ (approved):
lgtm!

PR #2893, opened 2024-09-04
frontend: put send request result in `Send` state

NicolaLS:
Put the result from the send request into the `Send` components state and use it directly instead of `isSent` and `isAborted`. This simplifies the component and will make the refactor to a functional component easier because we'll be able to use a `useEffect` with a dependncy on the `sendResult` state that can `switch` over it and also reset the form.


NicolaLS:
Actually the error handling can also go into the sub-component `MessageWaitDialog`...I'll do this in another commit but can squash it later.

thisconnect (approved):
tested LGTM

PR #2902, opened 2024-09-10
frontend: spacing fixes

shonsirsha:
1st commit to fix spacing for Send screen
Before:
before_

After:
after

2nd commit to fix spacing for Message component in the address type dialog
Before:
__before

After:
_after

### DIFF
--- a/frontends/web/src/routes/account/receive/receive.module.css
+++ b/frontends/web/src/routes/account/receive/receive.module.css
@@ -66,3 +66,7 @@
     text-decoration: underline;
     padding: var(--space-quarter);
 }
+
+.messageContainer {
+    margin-top: var(--spacing-default);
+}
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/send/send.tsx
```diff
@@ -527,14 +527,16 @@ class Send extends Component<Props, State> {
...
                   <label className="labelXLarge">{t('send.transactionDetails')}</label>
-                  { coinControl && (
-                    <Button
-                      className="m-bottom-quarter p-right-none"
-                      transparent
-                      onClick={this.toggleCoinControl}>
-                      {t('send.toggleCoinControl')}
-                    </Button>
-                  )}
+                  <div className={style.coinControlButtonContainer}>
```
thisconnect, line 530: I haven't tested but shouldn't that div element be inside the condition in case a user doesn't have coin-control enabled?

thisconnect (approved):
untested LGTM

PR #2904, opened 2024-09-10
backend: increase API response size limit for available countries

thisconnect:
The previous limit of 81,920 bytes caused issues due to increased response sizes, likely due to additional countries or fields in the JSON data. This change increases the limit to 100,000 bytes to accommodate current data and provide a buffer for future growth.

### DIFF
--- a/backend/exchanges/pocket.go
+++ b/backend/exchanges/pocket.go
@@ -98,7 +98,7 @@ func GetPocketSupportedRegions(httpClient *http.Client) (map[string]PocketRegion
 	var regionsList []PocketRegion
 	endpoint := fmt.Sprintf("%s/availabilities", pocketAPILiveURL)
 
-	_, err := util.APIGet(httpClient, endpoint, "", 81920, &regionsList)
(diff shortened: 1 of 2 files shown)


backend/exchanges/pocket.go
```diff
@@ -98,7 +98,7 @@ func GetPocketSupportedRegions(httpClient *http.Client) (map[string]PocketRegion
 	var regionsList []PocketRegion
 	endpoint := fmt.Sprintf("%s/availabilities", pocketAPILiveURL)
 
-	_, err := util.APIGet(httpClient, endpoint, "", 81920, &regionsList)
+	_, err := util.APIGet(httpClient, endpoint, "", 100000, &regionsList)
```
benma, line 101: This is too small of an increase I think. You can easily bump it 10x without any downsides and then we can be more confident there will not be a similar issue in the future.

benma (changes requested):

backend/exchanges/pocket.go
thisconnect: updated, yeah my increase was a bit more conservative.

benma (approved):
utACK

PR #2907, opened 2024-09-11
frontend: remove i18n placeholder in error messages

thisconnect:
These error messages should use context so that the whole message can be translated for each language.

Adding a placeholder does often not work in every langauge and the term 'buy' or 'sell' was not translatable in this case.

Removed the action placeholder for these messages, this needs to be updated in locize as well in all langauges.

### DIFF
--- a/frontends/web/src/routes/exchange/components/buysell.tsx
+++ b/frontends/web/src/routes/exchange/components/buysell.tsx
@@ -73,10 +73,10 @@ export const BuySell = ({
     }
   }, [exchangeDealsResponse, onSelectExchange]);
 
-  const constructErrorMessage = (action: exchangesAPI.TExchangeAction): string | undefined => {
+  const constructErrorMessage = (): string | undefined => {
     if (exchangeDealsResponse?.success === false) {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -73,10 +73,10 @@ export const BuySell = ({
...
 
-  const constructErrorMessage = (action: exchangesAPI.TExchangeAction): string | undefined => {
+  const constructErrorMessage = (): string | undefined => {
     if (exchangeDealsResponse?.success === false) {
       if (exchangeDealsResponse.errorCode) {
-        return t('exchange.buySell.' + exchangeDealsResponse.errorCode, { action });
+        return t('exchange.buySell.' + exchangeDealsResponse.errorCode);
       }
       return exchangeDealsResponse.errorMessage;
     } else if (paymentRequestError) {
```
benma, line 82: else after return :(

benma (approved):
utACK, thanks :heart:

PR #2908, opened 2024-09-12
frontend: fix button fontsize on mobile

thisconnect:
On small screens the Buy&sell button font size is larger than the other buttons, this is probably a regression that was introduced by some renaming.


thisconnect:
current 
Screenshot 2024-09-12 at 10 28 08


with this change:
Screenshot 2024-09-12 at 10 25 23

benma (approved):
utACK

PR #2910, opened 2024-09-13
frontend/backend: bb01 remove `EventSignConfirm`

NicolaLS:
Remove the `EventSignConfirm` and stop using it in the frontend because we decided it would be a nice cleanup.

- it looks like there's some more unused CSS in `dialog.module.css` and `dialog-legacy.module.css` but I decided to only remove the CSS rules that targeted classes that are removed by this PR. I might open another PR later to remove more CSS
- the `paired` prop was used to decide if it should show **both steps** or only _Continue on your BB02_ since we always **only** show the latter now I decided to remove the `paired` prop too and use the styling for `paired == false` of course we can change it to the styling of `paired == true` that would mean we add the `padding-left` and `margin-top` to the `confirmationLabel` again)

this is how it looks like now:

image


NicolaLS:
@thisconnect do I have to manually delete **all** keys in `locales/` or do the other ones get deleted automatically if the `en` keys are deleted ?

benma (commented):
Pls squash, it's one logical commit, similar to https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2861#issuecomment-2309465050. Think of it in terms of reverting commits: would it ever make sense to revert one commit but not the other? If not, it should be one.

benma:
> @thisconnect do I have to manually delete **all** keys in `locales/` or do the other ones get deleted automatically if the `en` keys are deleted ?

Just in `en`. Maybe you could add a note to https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/docs/i18n.md about this.

NicolaLS:
> > @thisconnect do I have to manually delete **all** keys in `locales/` or do the other ones get deleted automatically if the `en` keys are deleted ?
> 
> Just in `en`. Maybe you could add a note to https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/docs/i18n.md about this.

thanks, I took a note and will open a small doc PR later

thisconnect (approved):
tested LGTM

PR #2911, opened 2024-09-14
frontend: reduce spacing on mobile for settings subtitle text

shonsirsha:
Before:

b_

After:

a_


thisconnect (approved):
untested LGTM thanks 👍

PR #2913, opened 2024-09-16
docs/i18n: add more details on add/remove keys

NicolaLS:
Clarify that **only** `en` keys should be added, removed or changed.

ref: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2910#issuecomment-2349689642

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -37,7 +37,12 @@ Or use a build of the app with an inline editor (see above).
 
 ## For developers
 
-New keys can be added to `frontends/web/src/locales/en/app.json`.
+Keys can be added to or removed from `frontends/web/src/locales/en/app.json`. Keys for other
+languages **must not** be added or removed manually.
+
+Existing keys **should not** be modified directly, but sometimes it is still useful to modify `en` keys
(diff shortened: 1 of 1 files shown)


docs/i18n.md
```diff
@@ -37,7 +37,9 @@ Or use a build of the app with an inline editor (see above).
 
 ## For developers
 
-New keys can be added to `frontends/web/src/locales/en/app.json`.
+New keys can be added to `frontends/web/src/locales/en/app.json`. Keys for other languages **must not**
+be added manually. The same applies to the removal or modification of keys, only `en` keys should be
+removed/edited.
```
thisconnect, line 42: existing `en` keys have to be updated in locize.

For bigger PR's I find it still useful to change en keys and commit to git. That way we see which keys should be manually updated in locize (after the PR is merged).

NicolaLS:
side note: not sure what the desired line-length is, I added line breaks at 100 without styling (so if there is **a** =5 letters I count it as 1)

NicolaLS:
@thisconnect PTAL

thisconnect (approved):
LGTM

PR #2914, opened 2024-09-16
ios: add icon and Info.plist

benma:
Fixing errors found when uploading to TestFlight.

The icon was created by resizing frontends/android/icon_source.png to 1024x1024, the required size.

THe `<key>ITSAppUsesNonExemptEncryption</key>` Info.plist entry is added after App Store Connect prompted to declare this. See:

https://developer.apple.com/documentation/security/complying-with-encryption-export-regulations

As far as I can tell, we use no non-exempt encryption algorithm directly or in dependencies.


NickeZ (approved):
utACK!

PR #2919, opened 2024-09-18
frontend: use View for bb02bootloader component

shonsirsha:


thisconnect (approved):
untested LGTM

I assume mobile view looks ok?

PR #2922, opened 2024-09-18
Silent Payments send support

benma:


benma:
I tested:

- sending to a SP address with 'Send all'
- sending to a SP address with a specific amount
- sending using all supported UTXO types as inputs (p2wpkh, p2wpkh-p2sh, p2tr) 
- sending with a normal recipient address (to see the previous stuff didn't break)
- sending to a SP address with an outdated firmware (clear error msg appears saying an upgrade is required)
- Pocket sell flow with payment requests, to see that part did not break either

And an end-to-end test with sending funds from BitBoxApp to a CakeWallet silent payment account.

PR #2925, opened 2024-09-19
frontend: update isAmbiguousName function

strmci:
- rename function to correct typo
- update function parameter type from `TAccountsByKeystore[]` 
  to new type `TKeystoreName[]` to  ease type requirements and 
  prepare for integration with the lightning account


thisconnect (approved):
untested LGTM

PR #2926, opened 2024-09-19
Frontend transaction revamp

thisconnect:
Redesigned transaction history in account overview to improve the
list for mobile, better legibility and nicer look.

This revamp should also slightly improve performance, as it uses
less DOM elements, moved SVG from JSX to img assets and does not
use a Dialog for each transaction.


thisconnect:
rebased onto master without conflicts

thisconnect:
rebased

thisconnect:
sometimes when there is pending tx it cuts off the progress ring, I'll fix and squash into "frontend: revamp transaction history"

thisconnect:
closing in favor of https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2936

PR #2928, opened 2024-09-19
prevent sleep on macOS during long running queries

benma:
To prevent USB communication interruptions when e.g. recovering from 24 words.

See https://github.com/BitBoxSwiss/bitbox02-api-go/pull/111

```
go get github.com/BitBoxSwiss/bitbox02-api-go@88d5934
go mod tidy
go mod vendor
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Bitcoin: add support for sending to silent payment (BIP-352) addresses
+- Prevent macOS from going to sleep on long running interactions with the BitBox
 
 ## 4.44.0
 - Bundle BitBox02 firmware version v9.20.0
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/BitBoxSwiss/bitbox-wallet-app
 go 1.22
 
 require (
-	github.com/BitBoxSwiss/bitbox02-api-go v0.0.0-20240918121853-2221e9876d20
+	github.com/BitBoxSwiss/bitbox02-api-go v0.0.0-20240925080402-a2115fee878e
 	github.com/BitBoxSwiss/block-client-go v0.0.0-20240516081043-0d604acd6519
 	github.com/btcsuite/btcd v0.24.2
 	github.com/btcsuite/btcd/btcec/v2 v2.3.4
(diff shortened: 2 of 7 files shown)


vendor/github.com/BitBoxSwiss/bitbox02-api-go/util/sleep/sleep_darwin.go
```diff
@@ -0,0 +1,55 @@
...
+}
+*/
+import "C"
+
+// Prevent prevents macOS from going to sleep. Must be paired with `Allow()`.
+func Prevent() {
+	C.preventSleep()
+}
+
+// Allow allows macOS from going to sleep.
```
NickeZ, line 52: > Allow allows macOS from going to sleep.

Should probably be "Allow allows macOS to go to sleep."

thisconnect: Is it specific to macOS tho?

Testing on Ubuntu, with this it seems that it doesn't leave the workflow anymore when the screenlocks appears on Ubuntu.

Before: the workflow aborted silently when Ubuntu locked the screen.

thisconnect: actually nevermind, I tried to reproduce with 4.43 on Ubuntu and there is not the same issue with macOS loosing connection.

I think before I lost only connection on Ubuntu once it suspended.

thisconnect (approved):
LGTM

vendor/github.com/BitBoxSwiss/bitbox02-api-go/util/sleep/sleep_darwin.go
benma: thanks @NickeZ, fixed in https://github.com/BitBoxSwiss/bitbox02-api-go/pull/112 and updated the dep here again.

benma: @thisconnect as you have reviewed here, it is only macOS for now :grin: https://github.com/BitBoxSwiss/bitbox02-api-go/pull/111/files

PR #2931, opened 2024-09-25
Ios testnet

benma:


NickeZ:
To reduce confusion I think it makes sense to have a separate launcher icon (can be super uggly) and have some red overlay, always visible, that says "testnet".

benma:
> To reduce confusion I think it makes sense to have a separate launcher icon (can be super ugly) and have some red overlay, always visible, that says "testnet".

@thisconnect could you please make a separate PR for a sticky warning banner at the  top of the app in testnet?

Can use 

    import { getTesting } from '@/api/backend';

NickeZ (approved):
tACK

PR #2932, opened 2024-09-25
ios: allow app to access camera

benma:
Needed for QR code scanning.

### DIFF
--- a/frontends/ios/BitBoxApp/BitBoxApp/Info.plist
+++ b/frontends/ios/BitBoxApp/BitBoxApp/Info.plist
@@ -4,5 +4,7 @@
 <dict>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
+	<key>NSCameraUsageDescription</key>
+	<string>We need access to the camera to scan QR codes.</string>
 </dict>
 </plist>
--- a/frontends/ios/BitBoxApp/BitBoxApp/WebView.swift
+++ b/frontends/ios/BitBoxApp/BitBoxApp/WebView.swift
@@ -171,5 +171,17 @@ struct WebView: UIViewRepresentable {
             }
             return nil
         }
+
+
+        // Automatically grant camera permission when used in the webview.
+        // The camera permission was already granted at install time via
+        // the NSCameraUsageDescription Info.plist entry.
+        func webView(_ webView: WKWebView,
+                     requestMediaCapturePermissionFor origin: WKSecurityOrigin,
+                     initiatedByFrame frame: WKFrameInfo,
+                     type: WKMediaCaptureType,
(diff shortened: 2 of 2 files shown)


NickeZ:
Is it possible to test this in the simulator? I get "The operation Aborted", but I'm not sure if that is expected in the simulator

benma:
> Is it possible to test this in the simulator? I get "The operation Aborted", but I'm not sure if that is expected in the simulator

For me I get a prompt to allow permission, then I get a fake videostream in the simulator :thinking: 

I pushed this to TestFlight (as BitBoxApp Testnet) now so we can do a live test too.

benma:
@NickeZ it works for @jadzeidan but if he taps it it looks like a videoplayer that can be paused :sweat_smile:

NickeZ:
> @NickeZ it works for @jadzeidan but if he taps it it looks like a videoplayer that can be paused 

Scanning works, but the experience sucks.

NickeZ:
There is no difference between this build and the previous build. So I'm not sure if this change does anything.

NickeZ:
Camera works equally bad for me in both builds. I think we cannot do much about it since we use the built in web viewer? Or perhaps we can improve the experience by requesting an entitlement? https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_device_camera

benma:
@NickeZ both builds as in master vs this PR? Before this PR, @jadzeidan and I could not load the camera at all, and there was no prompt for allowing it. Are you sure you are comparing the right things?

benma:
> I think we cannot do much about it since we use the built in web viewer? 

We'll see, maybe there is a way to access a native scanner anyway. Out of scope for this PR though.

> Or perhaps we can improve the experience by requesting an entitlement? https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_device_camera

This is for macOS afaik. We use this here: https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/2ad55f250224a27a3678d016f235f2703d3024da/frontends/qt/resources/MacOS/entitlements.plist#L13

NickeZ:
> @NickeZ both builds as in master vs this PR? Before this PR, @jadzeidan and I could not load the camera at all, and there was no prompt for allowing it. Are you sure you are comparing the right things?

I'm going back and forth between versions using testflight. But in the app we don't have the commit hash or the build number, so I cannot verify the actual version that I'm running.

NickeZ (approved):
After some testing I think this solves part of the problem indeed.

benma:
@NickeZ updated this PR to fix all issues, please take a look again

NickeZ (approved):
tACk, works great!

PR #2934, opened 2024-09-26
ci: pump go version for appveyor

thisconnect:


thisconnect:
> Why? We are still at Go 1.22

I know but appveyor CI passes builds with 1.23.1 somehow it doesn't like 1.22 🤷 

was just a test.

benma:
Seems a bit like a temporary service issue with choco.

PR #2939, opened 2024-09-27
ios: fix exporting files (CSV exports, log export, notes export)

benma:
Similar to Android, we set `ExportsDir()` to `""` and let the backend environment `getSaveFilename()` construct the full path.

Mobile handling of the dirs/files is a big hack and works on implicit assumptions, e.g. that `getSaveFilename()` is always called with `filepath.Join(exportsDir, filename)`, which is only the filename on mobile as `ExportsDir()` returns `""`. This should be cleaned up, but for now iOS is made to behave like Android in this regard.

### DIFF
--- a/frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
+++ b/frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
@@ -38,10 +38,12 @@ protocol SetMessageHandlersProtocol {
     func setMessageHandlers(handlers: MessageHandlersProtocol)
 }
 
-class GoEnvironment: NSObject, MobileserverGoEnvironmentInterfaceProtocol {
-    func getSaveFilename(_ p0: String?) -> String {
-        // TODO
-        return ""
+class GoEnvironment: NSObject, MobileserverGoEnvironmentInterfaceProtocol, UIDocumentInteractionControllerDelegate {
+    func getSaveFilename(_ fileName: String?) -> String {
+        let tempDirectory = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
+        // fileName cannot be nil this is called by Go and Go strings cannot be nil/null.
+        let fileURL = tempDirectory.appendingPathComponent(fileName!)
+        return fileURL.path
     }
 
     func auth() {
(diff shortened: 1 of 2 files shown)


benma:
Ping @Beerosagos in case you missed it

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
```diff
@@ -38,10 +38,11 @@ protocol SetMessageHandlersProtocol {
...
 }
 
-class GoEnvironment: NSObject, MobileserverGoEnvironmentInterfaceProtocol {
-    func getSaveFilename(_ p0: String?) -> String {
-        // TODO
-        return ""
+class GoEnvironment: NSObject, MobileserverGoEnvironmentInterfaceProtocol, UIDocumentInteractionControllerDelegate {
+    func getSaveFilename(_ fileName: String?) -> String {
+        let tempDirectory = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
+        let fileURL = tempDirectory.appendingPathComponent(fileName!)
```
Beerosagos, line 44: I think we need a guard for `fileName`, like the one used in `systemOpen` for the urlString

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
```diff
@@ -75,10 +76,44 @@ class GoEnvironment: NSObject, MobileserverGoEnvironmentInterfaceProtocol {
...
+        guard let scene = UIApplication.shared.connectedScenes
+                .filter({ $0.activationState == .foregroundActive })
+                .first as? UIWindowScene else {
+            return nil
+        }
+
+        return scene.windows.first(where: { $0.isKeyWindow })?.rootViewController
+    }
+
+    func systemOpen(_ urlString: String?) throws {
```
Beerosagos, line 90: The function doesn't throw any exception, I think you can remove the `throws` keyword

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
benma: I am unsure how filename could actually be nil here, any clue? That's why I just used `fileName!`.

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
benma: I don't think we can remove it - this whole class adheres to `MobileserverGoEnvironmentInterfaceProtocol`, i.e. it follows the Environment interface in backend.go. I assume `gomobile` translates a Go error return to `throws`:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/c4867b84198ce4dee877b981ebfb0ad1a00a6bf7/backend/backend.go#L150

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
Beerosagos: I agree it should not happen, but from what I could understand, the app would crash if that happened. I just felt that in the future we may exploit this method somewhere else without realizing that it is not safe against null values. wdyt?

benma: I fixed it, but I still disagree :tongue:  this is in the GoEnv interface, so it will only ever be called by the Go backend, and there is no "nil" string in Go. I am unsure why gomobile even types this as optional, it seems almost like a bug.

Beerosagos: Ah now I see your point! I didn't consider that strings are not pointers in Go :innocent:  Then I agree with you. Feel free to put it back to how it was initially. Maybe it would be nice to have a comment explicitly stating why we don't need the guard, tho

benma: Please approve PR if ok

Beerosagos (approved):

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
Beerosagos: Sorry! I thought I already approved it!

PR #2941, opened 2024-09-30
frontend/settings: enable log export on Android

Beerosagos:
Previously, log export button was hidden on Android, but the functionality actually works. This commit enabled it.


benma (approved):
utACK

benma (changes requested):
Add CHANGELOG pls

Beerosagos:
@benma done, PTAL

benma (approved):

PR #2944, opened 2024-10-01
frontend: allow action buttons to grow on mobile

thisconnect:
With different translations the action buttons (send, receive, buy & sell and WC) did not break nicely on small screen.

Due to limited button height the text could be outside of the button and only visible in darkmode but not in normal mode (white on gray).

Changed to let the buttons break onto 2 lines of text if there is not enough horozontal space.

### DIFF
--- a/frontends/web/src/routes/account/account.module.css
+++ b/frontends/web/src/routes/account/account.module.css
@@ -22,6 +22,7 @@
 
 .actionsContainer {
     display: flex;
+    flex-wrap: nowrap;
     transform: translateY(-36%);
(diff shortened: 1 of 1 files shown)


thisconnect:
Current master:

Screenshot 2024-10-01 at 14 02 05
Screenshot 2024-10-01 at 14 02 21

thisconnect:
With this change:
Screenshot 2024-10-01 at 14 02 42
Screenshot 2024-10-01 at 14 02 51
Screenshot 2024-10-01 at 14 03 02

benma (approved):
utACK :smile:

PR #2945, opened 2024-10-01
buy & sell i18n fix to master

benma:
So we can push the keys to locize from master. This commit will then be applied also to a new branch for a 4.44 fix.


Beerosagos (approved):
LGTM

PR #2951, opened 2024-10-02
android: specify foreground service type for API level >=29

benma:
https://developer.android.com/reference/android/app/Service#startForeground(int,%20android.app.Notification)

> Note: Beginning with SDK Version
Build.VERSION_CODES.UPSIDE_DOWN_CAKE, apps targeting SDK Version Build.VERSION_CODES.UPSIDE_DOWN_CAKE or higher are not allowed to start foreground services without specifying a valid foreground service type in the manifest attribute
R.attr.foregroundServiceType. See Behavior changes: Apps targeting Android 14 for more details.

In the simulator for Android 34 the app seems to work without this change.


benma:
The AndroidManifest only specifies dataSync, so maybe on Android 34, the deviceConnected one is not actually in use? With this commit, on Android 34 both types would be specified in the code call, which takes precedence over the manifest.

PR #2954, opened 2024-10-03
CHANGELOG: add v4.44.1

benma:


benma:
This is the diff of current master to v4.44.1, apart from the v4.44.1 version bumps (that are ignored as master is already at v4.45.0).

Beerosagos (approved):
LGTM

PR #2955, opened 2024-10-03
frontend: utilize send-wrapper to lift some stuff from the `Send` component

NicolaLS:
Two small improvements towards the functional refactor.

Two things:

1. I moved this:
https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/9ed78882c0dd6a9fc49a8d519d4888a171a8eb57/frontends/web/src/routes/account/send/send.tsx#L134-L137

Into the send-wrapper and I'm not so sure if that's a good idea. But imo. there is no point in showing the error/returning only when the user hits send, because that is the ultimate goal of the component anyways.

2. I use `bb01Paired` instead of `device` (which was actually `product`) now to decide whether to show BB01 or BB02 stuff. This is a bit implicit but I think it is easy to understand/does not make the code harder to understand. Ofc. we can just add the device prop. back in if you think it is worth it.

### DIFF
--- a/frontends/web/src/routes/account/send/components/confirm/confirm.tsx
+++ b/frontends/web/src/routes/account/send/components/confirm/confirm.tsx
@@ -25,28 +25,23 @@ import { TConfirmSendProps } from '@/routes/account/send/components/confirm/type
 import { ConfirmingWaitDialog } from '@/routes/account/send/components/confirm/dialogs/confirm-wait-dialog';
 import style from './confirm.module.css';
 
+type TConfimrSend = { bb01Paired: boolean | undefined } & TConfirmSendProps;
+
 type TFiatValueProps = {
   baseCurrencyUnit: ConversionUnit;
   amount: string
 }
 
-export const ConfirmSend = (props: TConfirmSendProps) => {
-  switch (props.device) {
-  case 'bitbox':
-    return (
-      <ConfirmingWaitDialog
-        {...props}
-      />
-    );
-  case 'bitbox02':
-    return (
-      <BB02ConfirmSend
-        {...props}
-      />
-    );
-  default:
-    return null;
-  }
+export const ConfirmSend = (props: TConfimrSend) => {
+  return (props.bb01Paired !== undefined ? (
+    <ConfirmingWaitDialog
+      {...props}
+    />
+  ) : (
+    <BB02ConfirmSend
+      {...props}
+    />
+  ));
 };
 
 export const BB02ConfirmSend = ({
@@ -57,7 +52,7 @@ export const BB02ConfirmSend = ({
   selectedUTXOs,
   coinCode,
   transactionDetails
-}: Omit<TConfirmSendProps, 'device'>) => {
+}: TConfirmSendProps) => {
 
   const { t } = useTranslation();
   const {
--- a/frontends/web/src/routes/account/send/send-wrapper.tsx
+++ b/frontends/web/src/routes/account/send/send-wrapper.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 Shift Crypto AG
+ * Copyright 2024 Shift Crypto AG
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


thisconnect (changes requested):

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -516,20 +501,16 @@ class Send extends Component<Props, State> {
                   </Column>
                 </Grid>
               </ViewContent>
-
-              {device && (
```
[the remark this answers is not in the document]
NicolaLS: I could also change the meaning of `bb01Paired: boolean | undefined` to `undefined` == no device connected and then use `true/false` to check whether its a bb01 or bb02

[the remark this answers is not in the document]
NicolaLS: nvm, that does not work

frontends/web/src/routes/account/send/components/confirm/confirm.tsx
```diff
@@ -25,12 +25,14 @@ import { TConfirmSendProps } from '@/routes/account/send/components/confirm/type
 import { ConfirmingWaitDialog } from '@/routes/account/send/components/confirm/dialogs/confirm-wait-dialog';
 import style from './confirm.module.css';
 
+type TConfimrSend = { bb01Paired: boolean | undefined } & TConfirmSendProps;
```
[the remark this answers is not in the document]
NicolaLS: oh..yes...I probably `git commit --amend` when addressing your prior comment and forgot there was two commits -.- good catch.

NicolaLS:
@thisconnect PTAL

frontends/web/src/routes/account/send/send-wrapper.tsx
```diff
@@ -29,14 +32,40 @@ type TSendProps = {
...
+  const [noMobileChannelError, setNoMobileChannelError] = useState<boolean>();
+
   const account = findAccount(accounts, code);
+  const product = deviceIDs.length > 0 ? devices[deviceIDs[0]] : undefined;
+
+  useEffect(() => {
+    if (account && product && product === 'bitbox') {
+      const fetchData = async () => {
+        try {
+          const mobileChannel = await hasMobileChannel(product)();
```
thisconnect, line 48: I get a 404 :(

Screenshot 2024-11-25 at 16 40 13

NicolaLS: oh, it is wrong to pass `product` to `hasMobileChannel` not sure why I did that. I'll update it.

NicolaLS: same for `getDeviceInfo` actually

NicolaLS:
@thisconnect thanks for the review, it was just a stupid logical error hope it works now. forgot to rebase/push before the change so the last push is the rebase and the one before the edit.

frontends/web/src/routes/account/send/send-wrapper.tsx
```diff
@@ -29,14 +32,40 @@ type TSendProps = {
...
+          console.error(error);
+        }
+      };
+      fetchData();
+    }
+  }, [account, deviceIDs, devices]);
+
+  if (noMobileChannelError) {
+    alertUser(t('warning.sendPairing'));
+    return;
```
thisconnect, line 62: Why  stop the rendering with a `return` here?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -131,10 +116,6 @@ class Send extends Component<Props, State> {
   }
 
   private send = async () => {
-    if (this.state.noMobileChannelError) {
-      alertUser(this.props.t('warning.sendPairing'));
-      return;
```
thisconnect, line 136: The intention of this `return` was to stop the send function.

thisconnect (changes requested):

frontends/web/src/routes/account/send/send-wrapper.tsx
thisconnect: I assume this was copied over, see my other comment. I opened a PR to fix this https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3089

thisconnect:
merged with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2955

(without the `return;` see comment above)

PR #2959, opened 2024-10-07
frontend: change layout manage accounts buttons (small devices)

shonsirsha:
Previews:

a
b

### DIFF
--- a/frontends/web/src/routes/account/add/add.tsx
+++ b/frontends/web/src/routes/account/add/add.tsx
@@ -237,7 +237,7 @@ export const AddAccount = ({ accounts }: TAddAccountGuide) => {
                   </Step>
                 </Steps>
               </div>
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/add/add.tsx
```diff
@@ -237,7 +237,7 @@ export const AddAccount = ({ accounts }: TAddAccountGuide) => {
                   </Step>
                 </Steps>
               </div>
-              <div className="row flex flex-row flex-between m-bottom" style={{ flexDirection: 'row-reverse' }}>
+              <div className={styles.buttonsContainer} style={{ flexDirection: 'row-reverse' }}>
```
thisconnect, line 240: I'd slightly prefer to refactor this instead with View/ViewButtons components, but I guess the layout wouldn't look  100% same.

thisconnect (approved):

PR #2962, opened 2024-10-07
frontend: remove unused state in send component

NicolaLS:
Remove the unused `account` state from the `State` type declaration because it is never used.

Remove the `paired?` state from the `Send` `State` declaration and also remove its usage in the return statement. It is always `undefined` which means `hidden={paired !== false}` is always `true`. This means we can remove it completely.


NicolaLS:
the merge conflicts are going to be super annoying / risky to mess something up. I'll open one PR with many commits..

PR #2963, opened 2024-10-07
frontend: improve send error handling

NicolaLS:
The `txProposalErrorHandling` function should **only** return errors and not set any state. Although this isn't causing an issue right now, the function should not set state in principle. 

Return only `TProposalError` from the function but make all fields optional. Additionally, this will be necesary for an upcoming refactor of the `Send` component.

To keep the logic the same we set `proposedFee` to `undefined` according to the same logic (ie. `{}` or `{ amountError: ... }` is returned) but in the `Send` component not in the error handling function.


NicolaLS:
the merge conflicts are going to be super annoying / risky to mess something up. I'll open one PR with many commits..

PR #2964, opened 2024-10-08
frontend: increase mobile settings back button's height

shonsirsha:
to align better with the page's title and guide (and to increase touch space) the height of the back button is increased to 24px.

Preview:
fhuda


thisconnect (approved):
untested LGTM

PR #2966, opened 2024-10-08
Final improvments before refactor

NicolaLS:
After #2955 

Refactoring the `Send` component I noticed these as either improvements or required preparation for the refactor. The commits are quite small and I don't think they need to be tested individually, just test HEAD of the PR, I can also squash them all into one "Send component improvements" commit ?

Please review each commit individually.

Putting the result of the proposal request into the state changes the most:
be190c667e638cad6bf975d427be3c8fadcd641a

But it also fixes a bug that exists on master right now, to reproduce:

1. Type some value X into the amount field. It will show Y in the fiat field.
2. Deleted some part of the address, to make it invalid.
3. Type in the amount field again it will show Y' and X'.
4. Click on send all, now it shows Y but in fiat still X'.

After this commit there will never be any `proposedStuff` in case the last proposal (and thus partially our current state) is invalid. To prevent this the "send all" checkbox is disabled on address/fee errors. And we fallback on the last typed amount/fiat in case we end up in the state of `sendAll = true` but `proposalResult = undefined`.

In case you think we should reset the custom fee I can drop this:
9c6765387c165e122b8248889f82a53908f2cf43

But it will let us remove the handler function later and only pass the state setter when we refactor `Send` to functional in case we don't reset the custom fee.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -65,7 +65,7 @@ export interface IAccount {
   active: boolean;
   watch: boolean;
   coinCode: CoinCode;
-  coinUnit: string;
+  coinUnit: CoinUnit;
   coinName: string;
   code: AccountCode;
   name: string;
--- a/frontends/web/src/routes/account/send/components/confirm/confirm.tsx
+++ b/frontends/web/src/routes/account/send/components/confirm/confirm.tsx
@@ -31,22 +31,15 @@ type TFiatValueProps = {
 }
 
 export const ConfirmSend = (props: TConfirmSendProps) => {
-  switch (props.device) {
-  case 'bitbox':
-    return (
-      <ConfirmingWaitDialog
-        {...props}
-      />
-    );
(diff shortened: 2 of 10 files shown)


NicolaLS:
(rebased onto #2955)

NicolaLS:
> Is it ready to be reviewed? :) @NicolaLS

Yes, please review the commits individually and ignore the first two as they belong to the other PR. I know this is not great for review (having commits from another PR) but I thought since they are small changes it would be nice if this could be reviewed simultaneously.

NicolaLS:
rebased on #2955 2955

thisconnect:
Ideally, this PR should be a refactoring only and not change the behavior.

> But it also fixes a bug that exists on master right now, to reproduce:

The bug fix should be it's own PR.

thisconnect:
I cherry-picked some of the commits into PR https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3091

PR #2970, opened 2024-10-09
bump minimum platform support to Windows 10, macOS 10.14

benma:
We upgraded to Go 1.22 a while ago:

https://go.dev/wiki/MinimumRequirements

> For Go 1.21 and later: Windows 10 and higher or Windows Server 2016
> and higher.

> Go 1.21 and later will support macOS Catalina 10.15 or newer; see https://go.dev/doc/go1.20#darwin.

See also:

- https://github.com/golang/go/issues/57003
- https://github.com/golang/go/issues/57125
- https://github.com/golang/go/issues/64622#issuecomment-1847475161

Windows 7 has oficially been EOL'd in Jan 2020, almost 5 years ago. It
is reasonable to not officially support it.


Beerosagos (approved):
LGTM

PR #2971, opened 2024-10-09
frontend: fix spacing between receiver address and tx detail label

shonsirsha:
Before:
image

After (without coin control):

egrfde

After (with coin control):

fdas


thisconnect (approved):
untested LGTM

PR #2973, opened 2024-10-10
frontend: revert f564cfa94f0979fd3efff2b638ce1e933f4e63f0

shonsirsha:
f564cfa94f0979fd3efff2b638ce1e933f4e63f0 causes infinite loop / calls to the AOPP endpoint, thus we revert it for now.


NicolaLS:
Opened #2976 with a fix.

PR #2974, opened 2024-10-10
Update Breez SDK to v0.6.1

dangeross:
This PR updates the Breez SDK to v0.6.1 which includes:
- Trampoline payment support - the application will by default attempt to send via a trampoline payment (if the trampoline payment fails the SDK falls back to a normal send payment)
- Upgrading VLS to 0.12.0 - to fix some further desync edge cases


frontends/web/src/routes/lightning/send/send.tsx
```diff
@@ -252,6 +252,7 @@ export const Send = () => {
   const [customAmount, setCustomAmount] = useState<number>();
   const [rawInputError, setRawInputError] = useState<string>();
   const [sendError, setSendError] = useState<string>();
+  const [useTrampoline] = useState<boolean>(true);
```
Beerosagos, line 255: Why use a state here?

dangeross: I added it in case you wanted to switch between trampoline/non-trampoline in the app.

Beerosagos: I'd remove it for now, since we're not using it. We can always put it back :) Thanks :pray:

Beerosagos (approved):
LGTM

PR #2991, opened 2024-10-23
frontend: make qrscanner more resilient to re-render

thisconnect:
I.e. when resizing the window or rotating the screen it sometimes crashed.


Beerosagos (approved):
tACK on Qt linux and Android. I'm not a super fan fo the fixed sleep time but it seems to work pretty well and I couldn't find any good alternative

thisconnect:
rebased with CHANGELOG conflict

PR #2992, opened 2024-10-23
frontend: add BTC Direct OTC information

sutterseba:
* Adds an info message on the buy and sell screen about BTC Direct and inks to their OTC service. This info will be shown if Pocket Bitcoin is available as an exchange option, as they are both available in the same regions (i.e. the EEA).
* Whitelist BTC Direct URL
* Adds an optional icon parameter to the Message component, allowing to use a custom icon with any of the existing Message types.

**Preview:**

Screenshot 2024-10-23 at 14 22 34


frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -133,6 +154,22 @@ export const BuySell = ({
           </div>
         )}
       </div>
+      {exchangeDealsResponse?.success &&
+        exchangeDealsResponse.exchanges.some(exchange => exchange.exchangeName === 'pocket') &&
```
thisconnect, line 158: tiny nit: for readability I'd prefer to wrap JSX in parenthesis
```suggestion
        exchangeDealsResponse.exchanges.some(exchange => exchange.exchangeName === 'pocket') && (
```

frontends/web/src/routes/exchange/exchange.module.css
```diff
@@ -94,4 +94,17 @@
 
 .updateButton {
     height: var(--item-height-small);
+}
+
+.infoContainer {
+    margin-top: var(--space-default);
+    .link {
+      color: var(--color-default);
+    }
```
thisconnect, line 103: Vite uses PostCSS under the hood which supports nesting afaik, also CSS got native nesting in more modern engines (our webview uses currently smth like chrome 83, but this will prob change with Qt 6.2)
https://caniuse.com/css-nesting
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting

but we never used nesting in the codebase afaik and I don't think it should be necessary here so please try to make it not nested.

thisconnect:
The vertical centering is now strange. I believe you'll have to increase the min-height of the exchangeContainer somehow probably not too much. Please test with different countries and also sell.

I wish we would have used View component for the exchange selection. Not sure if we should try to fix the current version or just move to View component.. cc @shonsirsha  wdyt?

Also back button only shows if there are options, but not if there are none. But looks it was like this before.

See screenshots, on mobile there should be enought space, but the bottons are out of the visible viewport

Screenshot 2024-10-24 at 13 29 52


Screenshot 2024-10-24 at 13 22 52

thisconnect:
@sutterseba please also add a CHANGELOG entry.

thisconnect:
> Since we'll move to be using the View component soon here and it's such a small UI bug, I'm fine with it for now.

ack. We might run into similar issue there if we want to maintain vertical centered + fixed height so that it doesn't jump. As the content is now often longer, we should consider not having it vertically centered anymore. But you can also look at improving it when moving to View component.

thisconnect (approved):
LGTM

thisconnect:
please rebase

frontends/web/src/locales/en/app.json
```diff
@@ -402,6 +402,55 @@
...
+              "link": "BitBox02 security threat model",
+              "text": "BTC Direct's Private Trading Desk is an external service not covered by the BitBox02 Security Threat model. However, BTC Direct is well-acquainted with BitBox products and offers secure, knowledgeable guidance. BitBox products verify crypto addresses during transactions to ensure secure delivery.",
+              "title": "Security"
+            },
+            "title": "Disclaimer"
+          },
+          "info": "Get a personal advisor, reduced fees and place orders directly through your contact at BTC Direct.",
+          "infobox": {
+            "introbtc": "Get personal assistance for high-value, over-the-counter (OTC) Bitcoin trades.",
+            "intromulti": "Get personal assistance for high-value over-the-counter (OTC) trades for Bitcoin, Ethereum stablecoins, and more.",
```
thisconnect, line 442: nit: this should have been using i18n context https://www.i18next.com/translation-function/context

 i.e. `intro_btconly` and `intro_multi`

but I think it is ok to keep as is.

frontends/web/src/components/terms/btcdirect-terms.tsx
```diff
@@ -0,0 +1,107 @@
...
+              &nbsp;
+              {t('buy.exchange.infoContent.btcdirect.disclaimer.paymentMethods.sell2')}
+            </p>
+          </li>
+        </ul>
+        <p>{t('buy.exchange.infoContent.btcdirect.disclaimer.paymentMethods.fee')}</p>
+        <h2 className={style.title}>{t('buy.exchange.infoContent.btcdirect.disclaimer.security.title')}</h2>
+        <p>{t('buy.exchange.infoContent.btcdirect.disclaimer.security.text')}</p>
+        <p>
+          <A href="https://bitbox.swiss/bitbox02/threat-model/">
```
thisconnect, line 78: note for myself: threat-model is english only so it doesn't need to point to different langs.

frontends/web/src/routes/exchange/btcdirect.tsx
```diff
@@ -0,0 +1,47 @@
...
+import style from './iframe.module.css';
+import { Header } from '@/components/layout';
+import { useTranslation } from 'react-i18next';
+
+export const BTCDirect = () => {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+
+  const openBTCDirect = () => {
+    open(getBTCDirectLink());
```
thisconnect, line 29: unfortunately there is a DOM api called open which doesn't work on Android https://developer.mozilla.org/en-US/docs/Web/API/Window/open

please add `import { open } from '@/api/system';`.

thisconnect (changes requested):

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -130,6 +161,28 @@ export const BuySell = ({
                 onClickInfoButton={() => setInfo(buildInfo(exchange))}
               />
             ))}
+            {exchangeDealsResponse?.exchanges.some(exchange => exchange.exchangeName === 'pocket') && (
```
thisconnect, line 164: the problem with this is that Pocket doesnt sell altcoins and so therefore BTCDirect never shows up if an altcoin account was selected.

thisconnect (approved):
tested LGTM 👍 

with that one open issue that the pocket deals do not include altcoins. We'll fix this in a new PR.

PR #2993, opened 2024-10-23
frontend: lazy-load languages

### LINKED ISSUE #1547: Explore not loading all languages
Currently we bundle all languages as part of the main bundle, but we keep adding more and more.

After upgrading  react-i18next https://github.com/digitalbitbox/bitbox-wallet-app/issues/1543 we should analyze the perf impact of those languages, memory, load speed etc and consider loading the languages only when needed.

cstenglein:
fixes #1547 

## Before
before

## After
after

Shouldn't have that much impact though since the languages are loaded locally.

### DIFF
--- a/frontends/web/src/i18n/i18n.ts
+++ b/frontends/web/src/i18n/i18n.ts
@@ -17,36 +17,59 @@
 
 import i18n from 'i18next';
 import { getNativeLocale } from '@/api/nativelocale';
-import appTranslationsAR from '@/locales/ar/app.json';
-import appTranslationsCS from '@/locales/cs/app.json';
-import appTranslationsDE from '@/locales/de/app.json';
-import appTranslationsEN from '@/locales/en/app.json';
-import appTranslationsFR from '@/locales/fr/app.json';
-import appTranslationsJA from '@/locales/ja/app.json';
-import appTranslationsRU from '@/locales/ru/app.json';
-import appTranslationsMS from '@/locales/ms/app.json';
-import appTranslationsNL from '@/locales/nl/app.json';
-import appTranslationsPT from '@/locales/pt/app.json';
-import appTranslationsHI from '@/locales/hi/app.json';
-import appTranslationsBG from '@/locales/bg/app.json';
-import appTranslationsTR from '@/locales/tr/app.json';
-import appTranslationsZH from '@/locales/zh/app.json';
-import appTranslationsFA from '@/locales/fa/app.json';
-import appTranslationsES from '@/locales/es/app.json';
-import appTranslationsSL from '@/locales/sl/app.json';
-import appTranslationsHE from '@/locales/he/app.json';
-import appTranslationsIT from '@/locales/it/app.json';
 import { languageFromConfig } from './config';
 import { localeMainLanguage } from './utils';
 import { setConfig } from '@/utils/config';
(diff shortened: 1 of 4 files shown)


thisconnect:
Very nice. This cuts the size of the main production JavaScript bundle by half (tested with `make buildweb`).

Somehow it doesn't initialize the language from the config correctly, but in the settings shows the correct lang.


Screenshot 2024-10-24 at 11 34 50

thisconnect:
buildweb with master

master-buildweb

buildweb with this branch

lazy-load-buildweb

thisconnect:
webdev is not good to test, in this PR I added `vite preview` so we can test prodcution build. I tested your branch

looks like this change reduced JS Heap and scripting time 

current
profiling-master

with lazy-load branch
profiling-lazyload

cstenglein:
Fixed the loading of the language, but not that happy with the implementation with the changing boolean.

cstenglein:
Fixed the test and squashed my commits. Should work now.

thisconnect:
Thank you so much @cstenglein !

One thing please change the prefix in the commit message from `(web):` to `frontend:` so it is consistent with other commits https://github.com/BitBoxSwiss/bitbox-wallet-app/commits/master/ 🙏

frontends/web/src/i18n/i18n.ts
```diff
@@ -59,56 +82,49 @@ i18Init.init({
...
-  // the newly selected language lng to make the app use native-locale again.
-  // This also covers partial matches. For example, if native locale is pt_BR
-  // and the app has only pt translation, assume they match.
-  //
-  // Since userLanguage is stored in the backend config as a string,
-  // setting it to null here in JS turns it into an empty string "" in Go backend.
-  // This is ok since we're just checking for a truthy value in the language detector.
-  return getNativeLocale().then((nativeLocale) => {
+// load the default language first
+loadLanguage(defaultLang);
```
thisconnect, line 95: I found a small bug in the Qt app:

When the userLanguage in config.json is an empty string (or not defined) the Qt BBApp will always start with English instead of using the native language. 

We purposely store an empty string in case the user changes the language to be the same as the system.

How to test:
- change system to something else than English, i.e. German
- change userLanguage to empty string in `~/Library/Application\ Support/bitbox/config.json`
- open the Qt app, English loads, but expected is that the BBapp loads in German

I can't reproduce this well in webdev as native-locale gives me an empty string on webdev so it always goes to En for me in webdev.

thisconnect: I played a bit around and the 2nd commit  in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3053 fixes the issue for me. Could you test with this and if it works for you squash it into your commit?

thisconnect (changes requested):

frontends/web/src/i18n/i18n.ts
thisconnect: Your version actually partly tries to load to German (with de system and userlang empty string) but the UI is still in English:
Bildschirmfoto 2024-11-15 um 12 18 39

thisconnect:
merged with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3053

PR #3002, opened 2024-10-28
frontend: fix actionbuttons for eth

thisconnect:
ETH can have long decimals (up to 18) and usually shows 4 action buttons: buy, send, receive and walletconnect.

Fixed that the balance is not covered by the action buttons.

Tested with `make servewallet-mainnet-prodservers` to see all action buttons including buy&sell


thisconnect:
with this PR action buttons should break before amounts break. Amounts if there is not enough room to display coin and fiat on the same line.


Screenshot 2024-10-28 at 15 09 14

PR #3003, opened 2024-10-29
qt: create windows desktop shortcut during installation

jadzeidan:
Currently, Windows users need to manually create a desktop shortcut. 
This PR creates a desktop shortcut during the installation process.

### DIFF
--- a/frontends/qt/setup.nsi
+++ b/frontends/qt/setup.nsi
@@ -103,6 +103,9 @@ Section -Main SEC0000
     #File /r /x Makefile* @abs_top_srcdir@/doc\*.*
     SetOutPath $INSTDIR
     WriteRegStr HKCU "${REGKEY}\Components" Main 1
+
+    # Create a shortcut on the desktop
+    CreateShortCut "$DESKTOP\$(^Name).lnk" "$INSTDIR\${APP_EXE}" "" "" 0
 SectionEnd
 
 Section -post SEC0001
@@ -164,8 +167,7 @@ Section -un.post UNSEC0001
     DeleteRegKey HKCU "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$(^Name)"
     Delete /REBOOTOK "$SMPROGRAMS\$StartMenuGroup\Uninstall $(^Name).lnk"
     Delete /REBOOTOK "$SMPROGRAMS\$StartMenuGroup\$(^Name).lnk"
-    Delete /REBOOTOK "$SMPROGRAMS\$StartMenuGroup\BitBox.lnk"
-    Delete /REBOOTOK "$SMSTARTUP\Bitcoin.lnk"
+    Delete /REBOOTOK "$DESKTOP\$(^Name).lnk"
     Delete /REBOOTOK $INSTDIR\uninstall.exe
     Delete /REBOOTOK $INSTDIR\debug.log
     Delete /REBOOTOK $INSTDIR\db.log
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Improve the UI of buy & sell page for mobile devices
 - Fixed export to CSV for ERC20 tokens.
 - Added support for xpub_required in AOPP.
+- Create desktop shortcut by default on Windows during installation
 
 # 4.46.3
 - Fix camera access on linux


jadzeidan:
@benma please have a look.

I looked into whether it is possible to add a checkbox to the last screen to create a shortcut, however it doesn't seem possible to do without removing the "Run BitBoxApp" checkbox, which I don't think it we should remove.

Therefore, the shortcut created by default. I think this is totally acceptable behavior for Windows users. Many apps create a desktop shortcut and it is expected. For those that don't want a a shortcut, it is trivial to delete from the desktop.

frontends/qt/setup.nsi
```diff
@@ -103,6 +104,12 @@ Section -Main SEC0000
     #File /r /x Makefile* @abs_top_srcdir@/doc\*.*
     SetOutPath $INSTDIR
     WriteRegStr HKCU "${REGKEY}\Components" Main 1
+
+    # Explicitly copy the icon file to the installation directory
+    File "${ICONDIR}\icon.ico"
+
+    # Create a shortcut on the desktop
+    CreateShortCut "$DESKTOP\${APP_NAME}.lnk" "$INSTDIR\${APP_EXE}" "" "$INSTDIR\icon.ico" 0
```
benma, line 112: the app already embeds an icon doesn't it? can you try simply setting the icon string to `""` and not copying the icon again to the install folder?

benma: I meant to try like this:

    CreateShortCut "$DESKTOP\${APP_NAME}.lnk" "$INSTDIR\${APP_EXE}" "" "" 0

benma: ping @jadzeidan

jadzeidan: Tested and that works 👍 

Will update PR

frontends/qt/setup.nsi
```diff
@@ -103,6 +104,12 @@ Section -Main SEC0000
     #File /r /x Makefile* @abs_top_srcdir@/doc\*.*
     SetOutPath $INSTDIR
     WriteRegStr HKCU "${REGKEY}\Components" Main 1
+
+    # Explicitly copy the icon file to the installation directory
+    File "${ICONDIR}\icon.ico"
```
benma, line 109: Seems like this is not needed anymore, right?

benma (changes requested):

frontends/qt/setup.nsi
jadzeidan: Correct, forgot to remove it in last update. Fixed now.

jadzeidan:
@benma done, PTAL

frontends/qt/setup.nsi
```diff
@@ -103,6 +104,9 @@ Section -Main SEC0000
     #File /r /x Makefile* @abs_top_srcdir@/doc\*.*
     SetOutPath $INSTDIR
     WriteRegStr HKCU "${REGKEY}\Components" Main 1
+
+    # Create a shortcut on the desktop
+    CreateShortCut "$DESKTOP\${APP_NAME}.lnk" "$INSTDIR\${APP_EXE}" "" "" 0
```
benma, line 109: Other lines use `$(^Name)`, presumably to reference the Name "BitBoxApp" as defined in the first line, so I think you can avoid defining another APP_NAME variable and use that instead.

benma (changes requested):
@jadzeidan I asked o3-mini to review this and it said that the shortcut will not be removed with the uninstaller. I tested it and it is right.

When looking at the uninstaller section, there seems to be these weird lines that is probably there by mistake:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/555c3ba846761d1a9e1884ca66a5ce5898b24d3f/frontends/qt/setup.nsi#L166-L167

BitBox.lnk and Bitcoin.lnk are not created by the installer (and never were afaik) so they should not be removed.

Let me know if you prefer me to finish off the PR or if you wanna do the changes yoursef.

jadzeidan:
> @jadzeidan I asked o3-mini to review this and it said that the shortcut will not be removed with the uninstaller. I tested it and it is right.

That is a a good catch, but it is also true for how it currently works (shortcut is not removed on uninstall now). So should that be fixed here or is it technically an independent improvement that should have it's own PR?

benma:
> > @jadzeidan I asked o3-mini to review this and it said that the shortcut will not be removed with the uninstaller. I tested it and it is right.
> 
> That is a a good catch, but it is also true for how it currently works (shortcut is not removed on uninstall now). So should that be fixed here or is it technically an independent improvement that should have it's own PR?

What do you mean? Now there is no desktop shortcut - this PR introduces it but does not remove it on uninstall. Or do yo mean there is another shortcut that is not properly removed?

jadzeidan:
> > > @jadzeidan I asked o3-mini to review this and it said that the shortcut will not be removed with the uninstaller. I tested it and it is right.
> > 
> > 
> > That is a a good catch, but it is also true for how it currently works (shortcut is not removed on uninstall now). So should that be fixed here or is it technically an independent improvement that should have it's own PR?
> 
> What do you mean? Now there is no desktop shortcut - this PR introduces it but does not remove it on uninstall. Or do yo mean there is another shortcut that is not properly removed?

I mean if I create a shortcut manually, I would still expect it to remove the shortcut on uninstall.

benma (approved):
tACK

PR #3004, opened 2024-10-29
frontend: use skeletons for account page instead of spinner

shonsirsha:
Preview opening account pages:
_(**Network is throttled (slowed down)** to preview the skeletons, which also causes icons to load slower)_

https://github.com/user-attachments/assets/ee86f4ef-ab01-4923-a762-71e14d373583

Preview **offline error (before):**
offline_err_b4

Preview **offline error (after, with this PR):**

offline_err_after

Preview **status not synced (before):**

synced_false_b4

Preview **status not synced (after, with this PR):**

sycned_false_after


frontends/web/src/routes/account/actionButtons.tsx
```diff
@@ -61,17 +62,42 @@ export const ActionButtons = ({ canSend, code, coinCode, exchangeSupported, acco
...
+        <>
+          {accountDataLoaded ?
+            <Link key="exchange" to={`/exchange/info/${code}`} className={style.exchange}>
+              <span>{t('generic.buySell')}</span>
+            </Link> :
+            <span key="buySellDisabled" className={`${style.exchange} ${style.disabled}`}>
+              {t('generic.buySell')}
+            </span>
+          }
+        </>
```
thisconnect, line 86: fragment is not necessary here and next one for walletConnect

frontends/web/src/routes/account/actionButtons.tsx
```diff
@@ -61,17 +62,42 @@ export const ActionButtons = ({ canSend, code, coinCode, exchangeSupported, acco
...
-      {walletConnectEnabled && <Link key="wallet-connect" to={`/account/${code}/wallet-connect/dashboard`} className={style.walletConnect}>
-        <WalletConnectLight width={24}/> {!isLargeTablet && <span>Wallet Connect</span>}
-      </Link>}
+
+      {exchangeSupported && (
+        <>
+          {accountDataLoaded ?
+            <Link key="exchange" to={`/exchange/info/${code}`} className={style.exchange}>
+              <span>{t('generic.buySell')}</span>
+            </Link> :
```
thisconnect, line 81: nit: I find this hard to read
please use `accountDataLoaded ? ( \newline ) : ( \newline ) ` here and in other places.. 

would it help if I try to find an eslint rule to force this?

thisconnect (changes requested):

thisconnect:
> should I just squash it into 1 commit?

yes it's fine to have it in 1 commit

thisconnect:
tested a bit and it and it LGTM, one problem I encountered in offline mode is changing to an account loads with loading AND offline error message. I think it should only show offline error without loading message.

after navigating a bit between accounts the view becomes just empty. But this issue already exist on master and doesn't get worse with this PR. It would be nice if it would still load the account with skeletons and offline error.

here is a screenshot where it shows offlienError AND loading-data-from-the-blockchain, I think if we know there is an error we don't need to show the info about loading.

Screenshot 2024-10-31 at 10 54 54

PR #3006, opened 2024-10-30
frontend: reset percentage diff on chart

thisconnect:
In some situation the percentage difference did not update.

Test:
- have 1 empty remembered wallet with no history
- unlock a wallet with a history so the chart loads
- remove the wallet
- expected result: total should be 0 with no percentage difference


thisconnect:
> Account summary crashes when testing this PR

@shonsirsha thanks for catching this, resetting in removeChart war conceptually wrong. I moved it to the rendering function.

PR #3007, opened 2024-10-30
frontend: remove localised warning

thisconnect:
If the system native locale returns an empty string, this localize function is very noisy and returns a big warning.

In this case the app fallsback to default de-ch formatting.

Dropped the warning.


frontends/web/src/hooks/localized.ts
```diff
@@ -35,9 +35,7 @@ export const useLocalizedPunctuation = (
       );
       decimal = parts.find(part => part.type === 'decimal')?.value || decimal;
       group = parts.find(part => part.type === 'group')?.value || group;
-    } catch (error) {
-      console.warn(error);
-    }
+    } catch (error) {}
```
benma, line 38: Or just `catch {}`, as you don't use the var.

PR #3013, opened 2024-10-31
Moonpay

benma:
Somehow since Qt 6, https://buy.moonpay.com/serviceWorker.js is loaded
in the external browser when loading the moonpay widget, as if it was
a link clicked by the user.

I can't see a way to identify this as being different to a link click.

I tried:

- acceptNavigationRequest(), but that is only called once a page is
being loaded (the main app index, or the iframe)
- interceptRequest info.navigationtype() is "link clicked" even for
the JS file

Maybe we could make it work by having our createWindow()
implementation actually create a new hidden window and intercept
requests on it to open in a new browser, but that seems as if it might
be unreliable too and too much effort to implement and test.


NickeZ (approved):
lgtm

Another option I thought of was to allow all `js` and `css` earlier, around where we blanket allow `qrc://` and `blob://`. But maybe that is wrong for some other reason.

PR #3014, opened 2024-10-31
frontend: fix account action buttons alignment regression

thisconnect:
The action buttons are not aligned with the right edge of the content area, but instead just follow the total balance.

This happens also in Qt6.2.


benma:
> This happens also in Qt6.2.

Where else does it happen? :smile: 

You can merge whenever btw - Linux and Windows are already on Qt 6 on master.

benma (approved):
utACK

thisconnect:
> Where else does it happen? 😄

removed from commit message.

PR #3015, opened 2024-11-03
CI: Send messages to mattermost with artifact links

NickeZ:
Post messages in channels for build successes and failures of `push` kind of events.

I decided to use two channels because I think different people care about successful and unsuccessful builds. So you can be in the artifacts channel without getting spammed with fail builds for example.

I also added "if-no-files-found: error", since I think it is an error if the final apps are not created.

I didn't wait for the "dry-run" to complete in my repo before opening this PR because pulling the docker images from GHCR takes forever.

https://github.com/NickeZ/bitbox-wallet-app/actions/runs/11650613418

edit: dry run finished, there was a bug. The bug is fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,8 @@ jobs:
           OS_NAME: linux
   android:
     runs-on: ubuntu-22.04
+    outputs:
+      artifact-url: ${{ steps.upload.outputs.artifact-url }}
     steps:
       - name: Clone the repo
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -163,3 +181,42 @@ jobs:
...
+
+      - name: Message for failure
+        uses: mattermost/action-mattermost-notify@master
+        continue-on-error: true
+        if: job.status == 'failure'
+        with:
+          MATTERMOST_WEBHOOK_URL: ${{ secrets.MM_WEBHOOK_URL }}
+          PAYLOAD: |-
+            {
+              "channel": "eng-build-failures",
```
benma, line 218: does anyone need this? people working on their PRs will check their PRs.

NickeZ: This is in case the main branch fails to build. Which hopefully never happens.

benma: oh wait this is for pushes to branches in this repo too, nvm

benma (approved):
utACK

PR #3018, opened 2024-11-04
frontend: use skeletons for account page

shonsirsha:
Created and used BalanceSkeleton and TransactionHistorySkeleton in the appropriate places within the account page.

Also removed the blocking Spinner view, and replaced it with the Status component, informing the user the respective errors or message.

Same PR as #3004 , just using my own fork instead of pushing directly to origin.


thisconnect:
tested LGTM,

I notice first time navigating to an account in offline it shows the account with skeletons, but then when changing between different accounts it renders an empty view.

But it's the same on master, so I think this is not a new issue and still good to merge as is.

thisconnect (approved):
LGTM thanks

PR #3020, opened 2024-11-05
backend: remove "No priority" from fee options

strmci:
The "No priority" option has been removed because it can be confusing for users, and selecting it during high fee times can cause transactions to get stuck. Advanced users can still set custom fees if needed.


benma (changes requested):
changelog

Beerosagos:
I am going to print that ^ and attach it in front of my desk :joy:

strmci:
@benma updated :)

backend/coins/btc/account.go
```diff
@@ -496,14 +496,12 @@ func (account *Account) feeTargets() FeeTargets {
...
 	if mempoolFees != nil {
 		feeTargets = FeeTargets{
-			{blocks: 12, code: accounts.FeeTargetCodeMempoolEconomy},
 			{blocks: 3, code: accounts.FeeTargetCodeMempoolHour},
 			{blocks: 2, code: accounts.FeeTargetCodeMempoolHalfHour},
 			{blocks: 1, code: accounts.FeeTargetCodeMempoolFastest},
 		}
 	} else {
 		feeTargets = FeeTargets{
-			{blocks: 24, code: accounts.FeeTargetCodeEconomy},
```
benma, line 506: I think this should not be removed right? This is the Bitcoin Core estimator if mempool.space is down.

Beerosagos: I think it would make sense to remove the Bitcoin Core economy option, for the same reasons. But if we want to do that, than we should clean it up completely, as done for the mempool one.

benma (approved):
utACK - @Beerosagos good to merge?

Beerosagos (approved):
partially tested ACK

PR #3021, opened 2024-11-05
backend: update to version 4.46.0

benma:


Beerosagos (changes requested):
I think that qt/setup.nsi should be updated too

benma:
> I think that qt/setup.nsi should be updated too

:flushed: 

Great catch! Fixed

Beerosagos (approved):
LGTM

PR #3026, opened 2024-11-06
Frontend handle drop

thisconnect:
The app could show SHA256 checksums when dropping app files.


benma (commented):
It's a fun idea, but I am not a fan of the sha256 hidden feature. First because it's hidden (features should have proper UI), but:

From a security perspective, it is questionable to use the app that already might be compromised to check the next version, and you can't bootstrap it. An independent tool is safer.

thisconnect:
ack makes sense.. was just for fun

thisconnect:
replaced by https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3025

PR #3032, opened 2024-11-07
frontend: fix estimate symbol

thisconnect:
The estimate (≈) symbol should be placed before the amount and sign separated by a white-space.

Example:

    ≈ +100.-


frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -157,21 +157,34 @@ const Amounts = ({
   const conversion = amount?.conversions && amount?.conversions[defaultCurrency];
   const sign = getTxSign(type);
   const txTypeClass = `txAmount-${type}`;
+  const conversionPrefix = amount.estimated ? '\u2248' : false; // ≈
```
benma, line 160: a char or a bool is a bit weird for a type. null or undefined would be a better fit right?

benma (approved):
utACK

PR #3033, opened 2024-11-07
frontend: fix back button with UseBackButton

benma:
The Android back button stopped working after the back button was used to close a dialog. The reason is that the backbutton pop handler did not remove the handler, as `slice(index, 1)` does not remove the element at `index`, it in fact keeps it. When properly removing the handler, back button behavior works as expected again.


benma:
@thisconnect to emulate the Android back button in webdev, enter this in the console:

```
if(window.onBackButtonPressed()) { history.back()}
```

thisconnect (approved):
thank you 🙏 

tested LGTM

PR #3034, opened 2024-11-07
frontend/auth: fix auth screen glitch on startup

Beerosagos:
The auth screen was sometimes showed for a very short time at startup even when it was disabled. This fixes the glitch, hiding the auth screen details while the backend checks if the auth config is enabled.

### DIFF
--- a/frontends/web/src/components/auth/authrequired.tsx
+++ b/frontends/web/src/components/auth/authrequired.tsx
@@ -89,18 +89,21 @@ export const AuthRequired = () => {
         textCenter
         verticallyCentered
         withBottomBar>
-        <ViewHeader small title={t('auth.title')} />
-        <ViewContent children={undefined} minHeight="0" />
-        <ViewButtons>
-          <Button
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/auth/authrequired.tsx
```diff
@@ -89,18 +89,21 @@ export const AuthRequired = () => {
...
-          <Button
-            autoFocus
-            primary
-            hidden={authForced.current}
-            disabled={authenticating}
-            onClick={newAuthentication}>
-            {t('auth.authButton')}
-          </Button>
-        </ViewButtons>
+        { !authenticating && (
```
benma, line 92: a bit hard to undestand to me - `authenticating` is false by default, so how does this help in not showing the auth stuff at startup?

why this and not `authRequired && ( ... )`?

benma (approved):
utACK

Beerosagos:
rebased

PR #3035, opened 2024-11-07
frontend: change transaction info for send-to-self

thisconnect:
Until now a send-to-self transaction looked more like an outgoing transaction with the amount sent.

Changed to only show the fee paid as outgoing and the amount of coin as conversion amount.

### DIFF
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -219,10 +219,10 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
 		Time:         formattedTime,
 		Addresses:    addresses,
 		Note:         handlers.account.TxNote(txInfo.InternalID),
+		Fee:          feeString,
 	}
 
 	if detail {
-		txInfoJSON.Fee = feeString
 		switch handlers.account.Coin().(type) {
 		case *btc.Coin:
 			txInfoJSON.VSize = txInfo.VSize
--- a/frontends/web/src/components/transactions/transaction.tsx
+++ b/frontends/web/src/components/transactions/transaction.tsx
@@ -28,12 +28,13 @@ import { getTxSign } from './utils';
 import styles from './transaction.module.css';
 
 type TTransactionProps = ITransaction & {
-  onShowDetail: (internalID: ITransaction['internalID']) => void
+  onShowDetail: (internalID: ITransaction['internalID']) => void;
 }
 
 export const Transaction = ({
   addresses,
   amountAtTime,
+  fee,
   onShowDetail,
   internalID,
   note,
@@ -65,7 +66,11 @@ export const Transaction = ({
           time={time}
           type={type}
         />
-        <Amounts amount={amountAtTime} type={type} />
+        <Amounts
+          amount={amountAtTime}
+          fee={fee}
+          type={type}
+        />
         <button
           className={styles.txShowDetailBtn}
           onClick={() => !isMobile && onShowDetail(internalID)}
@@ -146,11 +151,13 @@ const Status = ({
 
 type TAmountsProps = {
   amount: IAmount;
+  fee: IAmount;
   type: TTransactionType;
 }
 
 const Amounts = ({
   amount,
+  fee,
   type,
 }: TAmountsProps) => {
   const { defaultCurrency } = useContext(RatesContext);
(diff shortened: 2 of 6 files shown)


backend/coins/btc/handlers/handlers.go
```diff
@@ -221,8 +221,8 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
 		Note:         handlers.account.TxNote(txInfo.InternalID),
 	}
 
+	txInfoJSON.Fee = feeString
 	if detail {
```
thisconnect, line 225: I slightly prefer if we could just pass all that data anyways.

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -146,25 +151,47 @@ const Status = ({
...
+  //   if (
+  //     type === 'send_to_self'
+  //     && !fee?.amount
+  //     && accountCode && internalID
+  //   ) {
+  //     getTransaction(accountCode, internalID).then(transaction => {
+  //       setFee(transaction?.fee);
+  //     });
+  //   }
+  // }, [accountCode, fee, internalID, type]);
```
thisconnect, line 181: If we don't have the fee in the /transactions endpoint we would need to fetch it somehow, with this PR we need it for send to self tx.

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -146,25 +151,47 @@ const Status = ({
...
+
+  const sendToSelf = type === 'send_to_self';
+  const coinAmount = sendToSelf ? fee.amount : amount.amount;
```
thisconnect, line 184: this should work for coins, but for tokens, we'd need fee.unit, as when a erc20-USDT is sent to self (for whatever reason), the fee is in ETH .. or gas?

thisconnect:
rebased

thisconnect:
Screenshot_2024-11-12_at_08 17 29

backend/coins/btc/handlers/handlers.go
thisconnect: interestingly TypeScript already expected that there is fee, also it thinks there is gas and nonce.  I am planing to open another PR to fix the types and make gas/nonce only be in /transaction?id= endpoint and optional.

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/12df2de2626dec6e53970f4ddbed45cb888450d0/frontends/web/src/api/account.ts#L247-L250

Tomasvrba:
@thisconnect nack from me, personally showing the fee as the bigger number (and therefore the main number always being negative) seems really counterintuitive, I would switch them around, the send amount with the "send to self" icon as the main number and the fee as the smaller number below (ideally with the label "fee")

thisconnect:
@Tomasvrba did you test with tokens as well?

Tomasvrba:
> @Tomasvrba did you test with tokens as well?

Yes, same issue from my perspective, since it's a token "account" I would expect the main tx number be the token transferred in the tx, not the eth-denominated fee

Screenshot From 2024-11-22 11-58-18

thisconnect:
rebased

thisconnect:
>  showing the fee as the bigger number (and therefore the main number always being negative) seems really counterintuitive, I would switch them around, the send amount with the "send to self" icon as the main number and the fee as the smaller number below (ideally with the label "fee")

As discussed internally: 

For send-to-self transactions we keep showing the fee as the bigger number so that the user see what they spent to do so.

In ERC20 tokens the fee is in ETH that is how Ethereum works, but

…there seems to be no good use-case to do send-to-self on account based coins/tokens. If you find out how contract interaction would look like in the app where the user initiates some kind of withdrawal which results in a deposit to the user's address please let me know.

thisconnect:
rebased, cc @Beerosagos PTAL at the backend change.

backend/coins/btc/handlers/handlers.go
```diff
@@ -221,8 +221,8 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
 		Note:         handlers.account.TxNote(txInfo.InternalID),
 	}
 
+	txInfoJSON.Fee = feeString
```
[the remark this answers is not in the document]
thisconnect: done

Beerosagos (approved):
backend LGTM

thisconnect:
rebased

PR #3042, opened 2024-11-11
frontend: better font size & spacing during firwmware install

shonsirsha:
Same as #3028  (closed due to wrong head branch).

fw_update

fw_install

fw_additional


thisconnect (approved):
untested LGTM

PR #3046, opened 2024-11-11
backend: add litecoin to btcdirect supported coins

sutterseba:
BTC Direct supports buying and selling LTC through their private trading desk, so the link and info should also be shown to Litecoin accounts on exchange selection.


sutterseba:
rebased to release-v4.46.0

PR #3048, opened 2024-11-12
frontend/qrcodescanner: improve qrscanner component

Beerosagos:
7c0b2f6 introduced a 300ms wait between the QrScanner creation and its start to fix a bug on re-renders of the component, but it seems not to be enough on certain devices, which still show the bug.

It would be nice to change the 300ms fixed wait to a better way to know when the scanner is ready to be started but we could not find it.

This commit improves the general structure of the component, to limit race conditions across re-renders and seems to improve the behavior and fix the bug with the tested devices.

### DIFF
--- a/frontends/web/src/hooks/qrcodescanner.ts
+++ b/frontends/web/src/hooks/qrcodescanner.ts
@@ -34,61 +34,82 @@ export const useQRScanner = (
   const { t } = useTranslation();
   const [initErrorMessage, setInitErrorMessage] = useState();
   const scanner = useRef<QrScanner | null>(null);
-
-  useEffect(() => {
-    if (videoRef.current && !scanner.current) {
-      scanner.current = new QrScanner(
-        videoRef.current,
-        result => {
-          scanner.current?.stop();
-          onResult(result);
-        }, {
-          onDecodeError: err => {
-            const errorString = err.toString();
-            if (err && !errorString.includes('No QR code found')) {
-              onError(err);
-            }
-          },
-          highlightScanRegion: true,
-          highlightCodeOutline: true,
-          calculateScanRegion: (v) => {
-            const videoWidth = v.videoWidth;
-            const videoHeight = v.videoHeight;
-            const factor = 0.5;
-            const size = Math.floor(Math.min(videoWidth, videoHeight) * factor);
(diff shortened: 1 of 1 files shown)


thisconnect:
I cannot reproduce the issue @strmci can you test this?

strmci:
> I cannot reproduce the issue @strmci can you test this?

I already tested it, @Beerosagos sent me the apk

thisconnect:
> I already tested it

and?

strmci:
> > I already tested it
> 
> and?

worked fine, on my Android

frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -34,61 +34,82 @@ export const useQRScanner = (
...
+        loading.current = false;
         const cameraNotFound = stringifiedError === 'Camera not found.';
         setInitErrorMessage(cameraNotFound ? t('send.scanQRNoCameraMessage') : stringifiedError);
+        onError(error);
       }
     })();
+
+    return () => {
+      (async() => {
+        while (loading.current) {
```
thisconnect, line 100: these while loops make me a little nervous 😅

Beerosagos: :joy: I know it's not exactly the react way to solve a problem like this. Feel free to suggest changes

Beerosagos:
> > > I already tested it
> > 
> > 
> > and?
> 
> worked fine, on my Android

@strmci could you please test again? This is not exactly the same apk you tested, even tho there should not be any relevant difference <3

strmci:
> > > > I already tested it
> > > 
> > > 
> > > and?
> > 
> > 
> > worked fine, on my Android
> 
> @strmci could you please test again? This is not exactly the same apk you tested, even tho there should not be any relevant difference <3

sure, can you please share apk with me?

strmci:
@thisconnect @Beerosagos 
tested again on my phone, qr scanner after rotation works fine, it takes approx 2 seconds to reload the camera after the rotation

thisconnect (approved):
could not reproduce on Android nor macOS, but looks good. Android 14 and macOS 14.7 load the camera

PR #3050, opened 2024-11-14
frontend: fix minor spacing issue on medium size screens

thisconnect:
Before the chart on account-summary goes into mobile view below 640px causing the total balance and filters to have no spacing to the edge of the window if the window was between 640px and 678px.

Fixed by increasing the mobile style to be applied until 678px.


thisconnect:
with this change the total balance and filters will go into mobile version is under 768px
Bildschirmfoto 2024-11-14 um 12 30 50

Beerosagos (approved):
utACK

PR #3053, opened 2024-11-15
Lazy load lang defaults

thisconnect:


frontends/web/src/i18n/i18n.ts
```diff
@@ -91,9 +91,6 @@ i18Init.init({
   }
 });
 
-// load the default language first
-loadLanguage(defaultLang);
```
thisconnect: This is the relevant change: in this commit I changed @cstenglein initial default loading with a initialized callback see below. cc @shonsirsha

thisconnect:
rebased

thisconnect:
rebased

thisconnect:
I noticed that for me on macOS (at least in `make servewallet`) the native locale endpoint returns an empty string.

```
 % locale
LANG=""
LC_COLLATE="C"
LC_CTYPE="UTF-8"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL=

% defaults read -g AppleLanguages
(
    "en-US",
    "de-CH"
)

% defaults read -g AppleLocale
en_US@rg=chzzzz

```

added a commit to "fix" the backend by trying `defaults read -g AppleLocale` or just falling back to en_US, @Beerosagos could you take a look at my last commit?

cmd/servewallet/main.go
```diff
@@ -103,6 +103,18 @@ func (webdevEnvironment) NativeLocale() string {
 	if v == "" {
 		v = os.Getenv("LANG")
 	}
+	// try macOS specific AppleLocale
+	if v == "" {
```
benma, line 107: ` && runtime.GOOS == "darwin"`

benma (changes requested):
Checked only the servewallet locale commit.

Wouldn't it be also good to have the frontend be able to handle invalid or missing locale info? Not sure what the issue was exactly, but I can imagine that the locale might not be detected properly on some platforms in some cases.

cmd/servewallet/main.go
thisconnect: thanks ❤️  updated with `&& runtime.GOOS == "darwin"`

thisconnect:
> Wouldn't it be also good to have the frontend be able to handle invalid or missing locale info?

I does already to a certain degree, but I need to test a bit more, what happens if there is really weird invalid data I believe worst case it just fallsback to EN.

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/9a74ab410863f1a01ae8effc82a8040eab340b66/frontends/web/src/i18n/utils.ts#L17-L25

> Not sure what the issue was exactly

If no userLang has been set the app tries to fallback to the navite lang from the OS, if that doesn't work it fallsback to 'en'.

I found and fixed another issue, that missing keys didn't fallback to en (fixed with last commit).

I'll test a bit more with invalid data later...

thisconnect:
rebased

thisconnect:
> I'll test a bit more with invalid data later.

pushed a commit that updated i18n tests with unknown, empty and invalid native locales

benma (approved):

cstenglein:
Hi,

sorry for not answering, have been busy these last months. :(
Thank you for finishing the PR on my behalf 🙏!

thisconnect:
@cstenglein thank you for the initial version 🙏

PR #3054, opened 2024-11-15
backend: re-enable copyloopvar linter and refactor code

strmci:
Since Go 1.22, the issue of the loop variable being reused in closures has been fixed, so the `copyloopvar` workaround is no longer needed. This commit re-enables the `copyloopvar` linter, disables exportloopref linter (deprecated and replaced by copyloopvar) and refactors the code to remove redundant assignments. See also: https://go.dev/doc/go1.22

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -86,7 +86,7 @@ linters:
     - wsl
     - execinquery
     - ireturn
-    - copyloopvar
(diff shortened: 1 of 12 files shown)


Beerosagos (approved):
utACK :+1:

PR #3056, opened 2024-11-18
android: handle device disconnect while the app is in the background

benma:
The ACTION_USB_DEVICE_DETACHED intent in MainActivity was not handled by the app while it was in the background.

Resulting bug:
- when you disconnect an unlocked device while the app is in the background, and bring the app to the foreground again, the previous unlocked keystore is visible, as the backend didn't register the disconnet.
- same when the user disconnects the device while the app is in the background, and reconnects it before bringing it to the foreground. Ths is a special case, simply calling `updateDevice()` in `onResume()` would not handle this correctly, as a device would be connected, and an intermediate disconnect could not be detected.

To fix this, we detect the DETACH intent in the service, and trigger the backend to handle it.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
@@ -5,12 +5,18 @@ import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.app.Service;
+import android.content.BroadcastReceiver;
+import android.content.Context;
 import android.content.Intent;
+import android.content.IntentFilter;
 import android.os.Binder;
 import android.os.Build;
 import android.os.IBinder;
+import android.hardware.usb.UsbManager;
 
 import androidx.core.app.NotificationCompat;
+import androidx.lifecycle.ViewModelProvider;
+import androidx.lifecycle.ViewModelStoreOwner;
 
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -30,6 +36,20 @@ public class GoService extends Service {
     private final String channelId = "21";
 
     private final int notificationId = 8;
+
+    private ViewModelStoreOwner viewModelStoreOwner;
+
+    private final BroadcastReceiver usbStateReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            String action = intent.getAction();
+            if (viewModelStoreOwner != null && UsbManager.ACTION_USB_DEVICE_DETACHED.equals(action)) {
+                GoViewModel viewModel = new ViewModelProvider(viewModelStoreOwner).get(GoViewModel.class);
+                viewModel.setDevice(null);
+                Mobileserver.usbUpdate();
+            }
+        }
+    };
 
     @Override
     public void onCreate() {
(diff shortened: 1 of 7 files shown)


thisconnect:
tested and works on my Android

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -31,6 +37,20 @@ public class GoService extends Service {
 
     private final int notificationId = 8;
 
+    private ViewModelStoreOwner viewModelStoreOwner;
+
+    private final BroadcastReceiver usbStateReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            String action = intent.getAction();
+            if (viewModelStoreOwner != null && UsbManager.ACTION_USB_DEVICE_DETACHED.equals(action)) {
```
Beerosagos, line 46: `viewModelStoreOwner` never seems to be initialized, is there some piece missing?

Beerosagos (changes requested):
It doesn't seem to work on my Android. Not sure how it could work for @thisconnect, I think there is something missing in the code

thisconnect:
> Not sure how it could work for @thisconnect

I'll try again

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
benma: Wow, true and very confusing, as I tested it and @thisconnect did too :thinking: I'll check again later to see what went wrong.

benma: Indeed the call to init viewModelStoreOwner was somehow missing. I added it back and now it seems to work again for me. Please check @Beerosagos @thisconnect

thisconnect:
Tested LGTM disconnecting the device while the app is in the background, going back to the app it now shows the waiting screen 

also connecting the device while the app is in the background asks shows a dialog "Open BitBoxApp DEBUG to handle BitBox02?" confirming this brings the app to the foreground showing he "Unlock" view.

I'm not sure why it worked last time when it shouldn't have worked 🤔

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
```diff
@@ -68,12 +88,26 @@ public void onCreate() {
         // focus. This is needed to avoid timeouts when the backend is polling the BitBox for e.g.
         // an address verification.
         startForeground(notificationId, notification);
+
+        // Register USB broadcast receiver to detect USB disconnects, even while the app is in the
+        // background.
+        IntentFilter filter = new IntentFilter();
+        filter.addAction(UsbManager.ACTION_USB_DEVICE_DETACHED);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
Beerosagos, line 97: The version code ('O' -> 26) seems wrong. The RECEIVER_EXPORTED should be available since version 33 (https://developer.android.com/reference/android/content/Context#RECEIVER_EXPORTED)

Beerosagos (changes requested):

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/GoService.java
benma: Good catch, I had only looked at registerReceiver itself:

https://developer.android.com/reference/android/content/Context#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter,%20int)

Fixed.

Beerosagos: I noticed it as it was a warning in my Android Studio

PR #3064, opened 2024-11-21
qt: force text/html as the mimetype for index.html

### LINKED ISSUE #3061: v4.46.0 AppImage no UI
Running the AppImage (Arch, KDE Wayland) using remote debugging gives the following errors in the console.

```
screencast.js:1 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'url')
    at b._requestNavigationHistory (screencast.js:1:18111)
screencast.js:1 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'url')
    at b._requestNavigationHistory (screencast.js:1:18111)
```

The error doesn't occur with v4.45.0

benma:
Fixes #3061.

Without this, on linux the local mimetype database (influenced by various packages installed on the system) can mess with the mimetype resolution of our .html file served locally, resulting in a blank page. Notably, FireFox on linux sets application/x-extension-html as the mimetype for .html files when made the default browser.

This problem is new in Qt 6, on Qt 5.15 it worked fine for some reason.

The general solution is to hardcode mime-types in a custom url scheme handler, but that requires using a different scheme to `qrc:`, which we can't do right now as Moonpay whitelisted only `qrc:`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 # 4.46.1
 - Fix Android app crash on old Android versions
+- Fix Linux blank screen issue related to the local mimetype database
 
 # 4.46.0
 - Android: enable export logs feature
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -15,6 +15,7 @@
 #include <singleapplication.h>
 #include <QApplication>
 #include <QCoreApplication>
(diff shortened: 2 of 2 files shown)


benma:
This breaks navigation behavior, e.g. going back from send/receive goes back to portfolio instead of the account :exploding_head:

PR #3065, opened 2024-11-22
Linux blank screen fix

benma:
The only "problem" left is that the console logs

    Please register the custom scheme 'bitboxapp' via QWebEngineUrlScheme::registerScheme() before installing the custom scheme handler.

But it seems to work anyway. I couldn't figure out how to get rid of this message.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -20,6 +20,11 @@
 #include <QWebEnginePage>
 #include <QWebChannel>
 #include <QWebEngineUrlRequestInterceptor>
+#include <QWebEngineUrlScheme>
+#include <QWebEngineUrlSchemeHandler>
+#include <QWebEngineUrlRequestJob>
+#include <QMimeDatabase>
+#include <QFile>
 #include <QContextMenuEvent>
 #include <QMenu>
 #include <QRegularExpression>
@@ -51,6 +56,11 @@
 #include "webclass.h"
 
 #define APPNAME "BitBoxApp"
+
+// Custom scheme so we can intercept and serve local files.
+// If you change this, MoonPay may break, as this scheme is whitelisted as a frame ancestor in
+// their CSP response headers.
+static const char* scheme = "bitboxapp";
 
 static QWebEngineView* view;
 static QWebEnginePage* mainPage;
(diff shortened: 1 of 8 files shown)


frontends/qt/main.cpp
```diff
@@ -111,12 +121,60 @@ class WebEnginePage : public QWebEnginePage {
...
+        QString hardcodedMimeType;
+        if (path.endsWith(".html")) {
+            hardcodedMimeType = "text/html";
+        } else if (path.endsWith(".js")) {
+            hardcodedMimeType = "application/javascript";
+        } else if (path.endsWith(".css")) {
+            hardcodedMimeType = "text/css";
+        } else if (path.endsWith(".svg")) {
+            hardcodedMimeType = "image/svg+xml";
+        } else {
```
thisconnect, line 155: The only ones missing as far as I see are .webm (for the password input video) and .png

benma: Pls provide the mimetypes.

The only practical isues we had were with .html and .js, the others I put just in case but are not fixing any known bugs, so it's not important to add all of them right away.

thisconnect: in chromium
- ".png" shows "image/png" 
- ".webm" uses "video/webm"

I didn't find any other endings in `frontends/web/build`

frontends/qt/main.cpp
```diff
@@ -111,12 +121,60 @@ class WebEnginePage : public QWebEnginePage {
...
+// Custom scheme handler so we can force mime type resolution manually.
+// Without this, on linux the local mimetype database can mess with the mimetype resolution.
+// Programs like monodevelop or Steam/Proton install weird mime types making for example our
+// .js or .html files be served with the wrogn mimetype, resulting in the webengine displaying a
+// blank page only.
+class SchemeHandler : public QWebEngineUrlSchemeHandler {
+public:
+    // Similar like the built-in qrc scheme handler, but with hardcoded mime-types.
+    // https://github.com/qt/qtwebengine/blob/v6.2.4/src/core/net/qrc_url_scheme_handler.cpp
+    void requestStarted(QWebEngineUrlRequestJob *request) override {
```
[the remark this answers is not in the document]
benma: Not sure, the original qrc handler does nothing of the sort either:

https://github.com/qt/qtwebengine/blob/v6.2.4/src/core/net/qrc_url_scheme_handler.cpp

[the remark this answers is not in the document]
Beerosagos: Chat says it's part of the `fileInfo` call, but I can't find any proof in the Qt docs.

PR #3074, opened 2024-11-26
qt: fix cam access by restoring window.isSecureContext

benma:
In d78380f45, we switched from qrc:/ to the bitboxapp:/ custom scheme so we can force mimetypes in the scheme handler. This somehow made window.isSecureContext be false, breaking camera (which requires a secure context).

The issue was that our scheme handler was never registered, because it was configued with HostAndPort but no default port was specified. We could specify one, or switch to Host only, but in our case Path seems to be most appropriate.

The frontend navigation is changed from browser router to hash based router, as somehow navigation breaks in the custom scheme, as the frontend tries to actually request the individual pages from QtWebEngine. Not sure why it was working before or what causes it, but a local hash based router seems to work.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -184,8 +184,8 @@ public:
         // We treat the exchange pages specially because we need to allow exchange
         // widgets to load in an iframe as well as let them open external links
         // in a browser.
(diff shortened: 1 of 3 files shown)


thisconnect (approved):
tested LGTM

Beerosagos (commented):
tACK on linux and Android 11

PR #3078, opened 2024-11-27
frontend/android: create new android id for lnalpha

Beerosagos:
Now it is possible to compile the app using separated Android app id, which allows the LN app to co-exhists with the debug one.

`make android-ln` to compile it,
`make deploy-ln` to install through adb.


NickeZ (changes requested):
* Some of the logotypes seem to be the old image
* I think we should build both the regular android debug build and the android lightning build. So duplicate the whole android build (we can use "matrix strategy" to avoid code duplication, but don't waste time on it if you don't want to.
* We should change name of the non-prod apps to "BBApp Debug" and "BBApp Lightning". Otherwise you just see "BitBoxApp ...".

Beerosagos:
Updated:
> * Some of the logotypes seem to be the old image

Yes, that is on purpose. We need both a round and a square(ish) icon. The one with the lightning in the corner is good for the last one.

>     * I think we should build both the regular android debug build and the android lightning build. So duplicate the whole android build (we can use "matrix strategy" to avoid code duplication, but don't waste time on it if you don't want to.

Not sure what you meant with the matrix strategy, but I think now it should be good


>     * We should change name to "BBApp Lightning". Otherwise you just see "BitBoxApp ...".

Done, but I left the "Alpha" too as I think it is an important warning. Ack?

PTAL :pray:

NickeZ (approved):
lgtm

PR #3080, opened 2024-11-28
frontend: full width for Dropdowns in General settings on mobile

shonsirsha:
Before:

After:
altpu


thisconnect (approved):
untested LGTM thanks

PR #3081, opened 2024-11-28
frontend/notes: filter text files on notes import

Beerosagos:
and add mime type handling in Android filepicker.

Before this, the file picker was allowing the user to select non-text files for notes import. Now this choice is restricted to *.txt files on Qt and text/plain files on Android and iOS (as it is not possible to filter single file extensions with the default file picker).


NickeZ:
tACK, other file types are greyed out now. 🎉

NickeZ (approved):

PR #3082, opened 2024-11-28
frontend: fix mobile remember wallet settings layout

shonsirsha:
Previously, the switch collapses if it has no enough space causing inconsistent layout between accounts. Now, that switch always stays on right edge top of the container. If there's no enough space, the name of the account and the switch label will collapse.

Before: 
Sasd9
Sasd

After:
Scx7

After, on super small screen:

Sasd4
(the fingerprint is hidden when this screenshot was taken).


thisconnect (approved):
tested in webdev LGTM thanks

PR #3083, opened 2024-11-28
ci: build windows app and report artifact

sutterseba:
This adds a GitHub Action job to build the Windows app and create the installer. Artifacts will be reported on Mattermost, like with other platforms.

Notable differences to our previous Appveyor CI:

* The GitHub windows-2019 image comes with Vistual Studio Enterprise instead of the Community edition. A directoy junction is created to point calls from the Community to the Enterprise path.
* NSIS, MinGW and Make are pre-installed already, not requiring any installations with choco anymore.
* Qt is not pre-installed anymore and now installed using an Action.

Because there are no changes to other files, the Appveyor build will continue to work and can be used in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,9 +181,53 @@ jobs:
         run: |
           make gomobileinit
           (cd $GOPATH/$GO_SRC_DIR; make ios)
+  windows:
+    runs-on: windows-2019
+    defaults:
+      run:
+        shell: bash
+    env:
+      MINGW_BIN: /C/mingw64/bin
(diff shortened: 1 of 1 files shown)


NickeZ (approved):
lgtm

benma:
We can delete the appveyor one then right? I don't see the need to maintain two Windows CI builds.

NickeZ:
> We can delete the appveyor one then right? I don't see the need to maintain two Windows CI builds.

I recommended that we wait a couple of weeks and test out this integration first. but I have nothing against killing appveyor.

PR #3086, opened 2024-12-03
frontend/widget: fix loosing iframe ref on screen rotation

Beerosagos:
Pocket and Bitsurance integrations both feature an address signing request in their widgets. With the previous implementation any re-rendering of the component (e.g. because of screen rotation on phone) caused the loss of the widget reference and the fail of the request. This commit fixes the bug.

### DIFF
--- a/frontends/web/src/routes/exchange/pocket.tsx
+++ b/frontends/web/src/routes/exchange/pocket.tsx
@@ -15,7 +15,7 @@
  */
 
 import { useTranslation } from 'react-i18next';
-import { useState, useEffect, createRef } from 'react';
+import { useState, useEffect, createRef, useRef } from 'react';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -53,7 +53,7 @@ export const Pocket = ({ code, action }: TProps) => {
   const accountInfo = useLoad(getInfo(code));
 
   const ref = createRef<HTMLDivElement>();
-  const iframeRef = createRef<HTMLIFrameElement>();
+  const iframeRef = useRef<HTMLIFrameElement | null>(null);
```
thisconnect, line 56: even thought Moonpay might not need it, I would do it there too for consistency

thisconnect (approved):
thanks 🙏 LGTM with comment

frontends/web/src/routes/exchange/pocket.tsx
thisconnect: oups 😅 .. thanks for checking 🙏

PR #3087, opened 2024-12-03
ci: add artifact url to windows outputs

sutterseba:
* Fixes missing artifact URL in Mattermost artifact reports
* Upload action will fail now if no files are found (like with the other job)


NickeZ (approved):
lgtm

PR #3092, opened 2024-12-05
frontend: keep fees dropdown always enabled

thisconnect:
The fees dropdown show useful information such as expected
time, or with config.frontend.expertFee current cost sat/vB.

This change enables opening the dropdown to see the current fees.

Added disabled to the 'custom' option so it is only selectable if
there is valid txinput data (such as address and amount).


thisconnect:
Screenshot 2024-12-05 at 08 14 12

PR #3093, opened 2024-12-05
frontend: hide estimated symbol for send-to-self transactions

thisconnect:
As the send-to-self amount is in coins there is no need to show an estimate symbol in front of the amount.


thisconnect:
Screenshot 2024-12-04 at 11 16 39

PR #3094, opened 2024-12-05
frontend: use blocking view for create ln wallet step

thisconnect:
As the user needs to confirm on the device, the app UI should block and disallow going back.

Screenshot 2024-12-05 at 13 56 04


Beerosagos (approved):
tACK! Very nice <3

thisconnect:
updated the message

Screenshot 2024-12-09 at 12 01 09

PR #3103, opened 2024-12-10
container: add container image version

Beerosagos:
This will allow to checkout older commits and run `make dockerdev` without worrying if the container image version changed.


scripts/dockerdev.sh
```diff
@@ -38,13 +35,15 @@ dockerdev () {
     fi
 
     # If already running, enter the container.
-    if $RUNTIME ps | grep -q $container_name; then
+    if $RUNTIME ps | grep -q $image_name:$CONTAINER_VERSION; then
```
benma, line 38: this seems wrong  `docker ps` shows the running containers, which are named with the container name, not the image name. If I have the same image running under a different name, I don't want `dockerdev` to use that.

scripts/dockerdev.sh
```diff
@@ -25,11 +26,7 @@ fi
 
 dockerdev () {
     local container_name=bitbox-wallet-dev
-
-    if ! $RUNTIME images | grep -q bitbox-wallet-app; then
-        echo "No bitbox-wallet docker image found! Maybe you need to run 'make dockerinit'?" >&2
-        exit 1
-    fi
+    local image_name=bitbox-wallet-app
```
benma, line 29: should be shiftcrypto/bitbox-wallet-app, shiftcrypto/ is part of the image name.

benma (changes requested):

scripts/dockerdev.sh
Beerosagos: Good point, added a second filter :+1:

Beerosagos:
@benma updated, PTAL :pray:

benma (approved):
utACK

PR #3109, opened 2024-12-31
chore: sort web assets before generating RCC file

tensor5:
The order of `find` results is non-deterministic, which breaks build reproducibility.


thisconnect:
LGTM

thank you @tensor5

benma:
@tensor5 are you trying to get reproducible builds? Does it work? Would be surprising if this was all that was needed!

Which platform did you build it on?

tensor5:
@benma, we use this patch in NixOS for the BitBox desktop app, and it works. Package [here](https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name/bi/bitbox).

thisconnect (approved):
LGTM

PR #3115, opened 2025-01-07
frontend: remove hide amounts settings

shonsirsha:
We'd like to show the hide amounts button (on the header) by default to improve the UX, since it's a common feature among similar apps.

This means, we don't need the hide amounts settings anymore.


thisconnect:
👍  this probably should have a changelog entry.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Use native scrollbars on macOS, iOS and Android
 - Fix address signing fail on screen rotation for Pocket and Bitsurance
 - Restrict selection to text files when importing notes
+- Remove hide amounts settings
```
thisconnect, line 9: Nit: if I read this it would be unclear to me if this feature is removed or just the setting.
How about something like permanently enable hide amount feature or similar?

thisconnect (approved):
tested webdev LGTM with nit

PR #3116, opened 2025-01-07
container: fix bug in make dockerdev

Beerosagos:
Recent commit ace02fe8aeb23c4ad2b49ec9c144397d32a6d383 introduced versioning for the app container. The commit introduced a bug that causes `make dockerdev` to fail when there is a stopped container that matches the current version. This fixes the bug.


benma (approved):
LGTM

PR #3122, opened 2025-01-13
frontend: enable clipboard read/write for android/qt

Beerosagos:
This enables the use of the `navigator.clipboard` JS api to read/write the clipboard on both Qt and Android and in particular:
- fixes the copy buttons in the Pocket order confirmation page
- updates the api of the CopyableInput component

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -224,6 +224,17 @@ public class MainActivity extends AppCompatActivity {
 
         vw.setWebViewClient(new WebViewClient() {
             @Override
+            public void onPageFinished(WebView view, String url) {
+                super.onPageFinished(view, url);
+                // override the default readText method, that doesn't work
+                // because of read permission denied.
+                view.evaluateJavascript(
+                        "navigator.clipboard.readText = () => {" +
+                        "    return androidClipboard.readFromClipboard();" +
+                        "};", null);
+            }
+
+            @Override
             public WebResourceResponse shouldInterceptRequest(final WebView view, WebResourceRequest request) {
                 if (request != null && request.getUrl() != null) {
                     String url = request.getUrl().toString();
@@ -349,7 +360,7 @@ public class MainActivity extends AppCompatActivity {
 
         final String javascriptVariableName = "android";
         vw.addJavascriptInterface(new JavascriptBridge(this), javascriptVariableName);
-
+        vw.addJavascriptInterface(new ClipboardHandler(this), "androidClipboard");
         vw.loadUrl(BASE_URL + "index.html");
 
         // We call updateDevice() here in case the app was started while the device was already connected.
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -23,6 +23,7 @@
 #include <QWebEngineUrlScheme>
 #include <QWebEngineUrlSchemeHandler>
 #include <QWebEngineUrlRequestJob>
+#include <QWebEngineSettings>
 #include <QMimeDatabase>
 #include <QFile>
 #include <QContextMenuEvent>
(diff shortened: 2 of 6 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -349,7 +360,7 @@ public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathC
 
         final String javascriptVariableName = "android";
         vw.addJavascriptInterface(new JavascriptBridge(this), javascriptVariableName);
-
+        vw.addJavascriptInterface(new ClipboardHandler(this), "AndroidClipboard");
```
benma, line 363: camelCase maybe? PascalCase is uncommon in JS

frontends/qt/main.cpp
```diff
@@ -453,6 +454,8 @@ int main(int argc, char *argv[])
     QWebChannel channel;
     channel.registerObject("backend", webClass);
     view->page()->setWebChannel(&channel);
+    view->settings()->setAttribute(QWebEngineSettings::JavascriptCanAccessClipboard, true);
+    view->settings()->setAttribute(QWebEngineSettings::JavascriptCanPaste, true);
```
benma, line 458: Why do we need this?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -223,6 +223,17 @@ public void run() {
         vw.getSettings().setMediaPlaybackRequiresUserGesture(false);
 
         vw.setWebViewClient(new WebViewClient() {
+            @Override
+            public void onPageFinished(WebView view, String url) {
+                super.onPageFinished(view, url);
+                // override the default readText method, that doesn't work
+                // because of read permission denied.
```
benma, line 230: Why should JS read from clipboard? Please update comment to say where/why we need this.

frontends/qt/main.cpp
Beerosagos: Without it, the `navigator.clipboard` API calls fail on Qt

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
Beerosagos: This is currently not used on master, but
- with this we could implement e.g. a button to paste an address in the send view
- we need this in the LN branch to paste invoices (we used a workaround in the past to overcome the issue there)

I see how it is a bit strange to enable it without directly using it, but It seemed a bit overkill to split the PR and moving this part to the LN branch, also bc I think that not allowing js paste is a bit of a limitation for the app interface. Btw, reading from clipboard is disabled by default in the iframes (even tho I guess that a malicious js dependency could do it in the app itself).

If you agree I can add a generic comment, but I can't mention a specific "where" we need it rn.

cc @thisconnect @shonsirsha

benma: :+1: 

What is the 'read permission' exactly that is denied, and why is it denied? Just wondering if there is a more straight forward way to make it work than this big workaround.

benma: Maybe the above info would be good in the commit msg at least

Beerosagos: I agree this is a big workaround. [Here](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#browser_compatibility) it says that `navigator.clipboard.readText()` should be supported since Android Webview 66, but `The user must grant the clipboard-read permission.`.

I couldn't find a way to provide that permission to the webview, and in my tests the `clipboard.writeText()` seems to work out of the box, while the `readText()` fails with a `Failed to execute 'readText' on 'Clipboard': Read permission denied.`.  I guess that since reading the clipboard is more sensible than writing it, from a security POV, there are different default permissions.

This is the only working solution I found (see also https://stackoverflow.com/questions/61243646/clipboard-api-call-throws-notallowederror-without-invoking-onpermissionrequest/61546346#61546346)

Beerosagos:
Updated the commit msg with more details and the PascalCase thing

Beerosagos:
rebased

benma (approved):
utACK

PR #3124, opened 2025-01-14
frontend: include fee in transaction summary

bznein:
When displaying the summary for a `send` transaction, instead of showing the received amount, show the "deducted" amount instead, i.e. received + fee.

Before 
image

After
image
Only change is in the highlighted field


backend/coins/btc/handlers/handlers.go
```diff
@@ -55,7 +55,8 @@ type Handlers struct {
 
 // NewHandlers creates a new Handlers instance.
 func NewHandlers(
-	handleFunc func(string, func(*http.Request) (interface{}, error)) *mux.Route, log *logrus.Entry) *Handlers {
```
bznein, line 58: This (and another change below) are done by `gofmt`, I might have a version different from the rest of the team, I'll check that soon

bznein:
Thanks for the review @Beerosagos, comments addressed this is ready for a re-review :)

backend/coins/btc/handlers/handlers.go
```diff
@@ -180,8 +182,10 @@ func (handlers *Handlers) ensureAccountInitialized(h func(*http.Request) (interf
 // If `detail` is false, Coin related details, fees and historical fiat amount won't be included.
 func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail bool) Transaction {
 	var feeString FormattedAmount
+	var deductedAmount FormattedAmount
 	if txInfo.Fee != nil {
 		feeString = handlers.formatAmountAsJSON(*txInfo.Fee, true)
+		deductedAmount = handlers.formatAmountAsJSON(coin.SumAmounts(txInfo.Amount, *txInfo.Fee), false)
```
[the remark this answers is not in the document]
thisconnect: why is it named deducted amount?

deducted - Taken away. Opposite of added.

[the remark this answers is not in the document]
bznein: > why is it named deducted amount?
> 
> deducted - Taken away. Opposite of added.

Isn't that correct? This is the total amount that was deducted from the account, i.e. sent + fee.
For reference, deducted is the term that was used in the issue and I used as reference :)

backend/coins/btc/handlers/handlers.go
```diff
@@ -189,14 +190,21 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
...
 		formattedTime = &t
 		amountAtTime = handlers.formatAmountAtTimeAsJSON(txInfo.Amount, txInfo.Timestamp)
+		if txInfo.Fee != nil {
+			deductedAmount = handlers.formatAmountAtTimeAsJSON(coin.SumAmounts(txInfo.Amount, *txInfo.Fee), txInfo.Timestamp)
+		}
 	} else if txInfo.CreatedTimestamp != nil {
 		t := txInfo.CreatedTimestamp.Format(time.RFC3339)
 		formattedTime = &t
 		amountAtTime = handlers.formatAmountAtTimeAsJSON(txInfo.Amount, txInfo.CreatedTimestamp)
+		if txInfo.Fee != nil {
```
Beerosagos, line 205: nit: instead of duplicating the code (and to improve also the existing duplicated one), you could define a timestamp var which gets a value depending on `txInfo.Timestamp`/`txInfo.CreatedTimestamp`, and use it to initialize the other depending vars

Beerosagos: Also another possible nit: I'm wondering if it makes sense to compute the deductedAmount only for `send` txs. In send-to-self the fee is not nil, so we compute the deducted amount, which is not used anywhere, also because it is technically wrong, as the only deducted value is the fee. Just thinking, no strong opinion here

bznein: > nit: instead of duplicating the code (and to improve also the existing duplicated one), you could define a timestamp var which gets a value depending on `txInfo.Timestamp`/`txInfo.CreatedTimestamp`, and use it to initialize the other depending vars

Yeah I was wondering the same but I wasn't sure about modifying unrelated code. definitely agree with you and going for this change.



> Also another possible nit: I'm wondering if it makes sense to compute the deductedAmount only for `send` txs. In send-to-self the fee is not nil, so we compute the deducted amount, which is not used anywhere, also because it is technically wrong, as the only deducted value is the fee. Just thinking, no strong opinion here

I think this one also makes sense!

backend/coins/btc/handlers/handlers.go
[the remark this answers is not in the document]
thisconnect: ah, makes total sense 😇  deducted from the account 👍

backend/coins/btc/handlers/handlers.go
```diff
@@ -189,14 +190,18 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
 		Unit:   amount.Unit,
 	}
 	var formattedTime *string
-	if txInfo.Timestamp != nil {
-		t := txInfo.Timestamp.Format(time.RFC3339)
+	var deductedAmount FormattedAmount
+	timestamp := txInfo.Timestamp
+	if timestamp != nil {
```
Beerosagos, line 195:
```suggestion
	if timestamp == nil {
```
:innocent:

Beerosagos (changes requested):

backend/coins/btc/handlers/handlers.go
```diff
@@ -189,14 +190,18 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
...
+		t := timestamp.Format(time.RFC3339)
 		formattedTime = &t
 		amountAtTime = handlers.formatAmountAtTimeAsJSON(txInfo.Amount, txInfo.Timestamp)
-	} else if txInfo.CreatedTimestamp != nil {
-		t := txInfo.CreatedTimestamp.Format(time.RFC3339)
-		formattedTime = &t
-		amountAtTime = handlers.formatAmountAtTimeAsJSON(txInfo.Amount, txInfo.CreatedTimestamp)
+		if txInfo.Fee != nil && txInfo.Type == accounts.TxTypeSend {
+			deductedAmount = handlers.formatAmountAtTimeAsJSON(coin.SumAmounts(txInfo.Amount, *txInfo.Fee), txInfo.Timestamp)
```
Beerosagos, line 203:
```suggestion
			deductedAmount = handlers.formatAmountAtTimeAsJSON(coin.SumAmounts(txInfo.Amount, *txInfo.Fee), timestamp)
```
Here and also for `amountAtTime`

Beerosagos (changes requested):

Beerosagos (approved):
tACK! :partying_face:

backend/coins/btc/handlers/handlers.go
```diff
@@ -189,14 +190,18 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
...
+	if timestamp == nil {
+		timestamp = txInfo.CreatedTimestamp
+	}
+	if timestamp != nil {
+		t := timestamp.Format(time.RFC3339)
 		formattedTime = &t
-		amountAtTime = handlers.formatAmountAtTimeAsJSON(txInfo.Amount, txInfo.CreatedTimestamp)
+		amountAtTime = handlers.formatAmountAtTimeAsJSON(txInfo.Amount, timestamp)
+		if txInfo.Fee != nil && txInfo.Type == accounts.TxTypeSend {
+			deductedAmount = handlers.formatAmountAtTimeAsJSON(coin.SumAmounts(txInfo.Amount, *txInfo.Fee), timestamp)
```
benma, line 203: @bznein @Beerosagos I don't think this works with ERC-20 tokens, where the fee has a different unit (ETH) than the ERC-20 token. See `txInfo.FeeIsDifferentUnit`.

Furthermore, the handler should ideally not contain logic, but only convert http inputs and outputs. Shouldn't this stuff be wherever TransactionData is being prepared?

Beerosagos: Shoot, you're right about the ERC20 think!

About the place in the code for this change, I guess you're right too :innocent: I didn't consider it a problem as I thought the deducted amount as a mere data representation used in the frontend and not useful in other parts of the backend, but I can see how it would be better to have it in TransactionData with some unit testing.

Beerosagos: Btw, while changing this, I think that it would be useful to add the coin type to the Tx, to make conversion estimation checks easier in the frontend cc @thisconnect

PR #3125, opened 2025-01-15
ci: replace deprecated set-output command

sutterseba:
GitHub Actions is deprecating the set-output command. This updates the workflow to use the new OUTPUT environment file instead.

Removes setting an environment variable since it is unused in the job.

See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


Beerosagos (approved):
LGTM :pray:

PR #3126, opened 2025-01-15
frontend: allow frontend to run not on port 8080

bznein:
This PR makes it possible to run the frontend on a port different from 8080.

There are 2 main changes:

1) Vite config is changed by adding `strictPort = true`
2) The port can be specified by an env var `VITE_PORT`. If not specified, it will default to 8080.

I feel 1 is the "mandatory" part of the PR. It makes no sense to allow the frontend to try different ports as it won't be able to connect to the backend anyway.

part 2 is optional, I think it's nice to allow to use another port. But if people object, I'd happily remove it and just leave 1.

### DIFF
--- a/frontends/web/vite.config.mjs
+++ b/frontends/web/vite.config.mjs
@@ -1,10 +1,12 @@
-import { defineConfig } from 'vite';
+import { defineConfig, loadEnv } from 'vite';
 import react from '@vitejs/plugin-react';
 import checker from 'vite-plugin-checker';
 import eslint from 'vite-plugin-eslint';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
 export default defineConfig((env) => {
+  const envVars = loadEnv(env.mode, process.cwd(), '')
+  const port = envVars.VITE_PORT
   return {
     // Relative base path so the js/css files are referenced with `./index-...js` instead of
     // `/index-...js`. This makes it easier to find these files in iOS.
@@ -29,5 +31,9 @@ export default defineConfig((env) => {
       pool: 'forks',
       setupFiles: './vite.setup-tests.mjs',
     },
+    server: {
+      port: typeof port !== 'undefined' ? port : 8080,
+      strictPort: true,
+    }
   };
 });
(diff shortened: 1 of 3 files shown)


frontends/web/vite.config.mjs
```diff
@@ -29,5 +31,9 @@ export default defineConfig((env) => {
       pool: 'forks',
       setupFiles: './vite.setup-tests.mjs',
     },
+    server: {
+      port: typeof port !== 'undefined' ? port : 8080,
```
benma, line 35: undefined is not a string is it? just `!== undefined`?

VITE_PORT is not an official env name right? Maybe `FRONTEND_PORT` or similar is better, in case we stop using vite? (we switched the frontend bundler in the past, maybe we will again).

benma (changes requested):
Nice!

Most the the PR description should be in the commit message body. `better logic for undefined` is not as descriptive as the PR description :smile:

bznein:
> Nice!
> 
> Most the the PR description should be in the commit message body. `better logic for undefined` is not as descriptive as the PR description 😄

You know, I have NO idea how that ended up there! Thanks for catching it :D

frontends/web/vite.config.mjs
bznein: > undefined is not a string is it? just `!== undefined`?

I was going by this: https://stackoverflow.com/questions/5113374/javascript-check-if-variable-exists-is-defined-initialized and if I use `undefined` not wrapped in quotes, it doesn't work (it doesn't detect that is undefined, so it tries to use `port`, which is undefined, and falls back to VITE default port (5173)


> VITE_PORT is not an official env name right? Maybe `FRONTEND_PORT` or similar is better, in case we stop using vite? (we switched the frontend bundler in the past, maybe we will again).

According to [this](https://vite.dev/guide/env-and-mode#:~:text=To%20prevent%20accidentally%20leaking%20env%20variables%20to%20the%20client%2C%20only%20variables%20prefixed%20with%20VITE_%20are%20exposed%20to%20your%20Vite%2Dprocessed%20code.%20e.g.%20for%20the%20following%20env%20variables%3A) it needs to be prefixed by `VITE`

PR #3129, opened 2025-01-16
backend: do not log error containing serialized tx

bznein:
The error message coming from electrum contains the serialized transaction and we do not want it in the logs, for privacy reasons, so we strip the entire error.

I opened a PR on their side to remove the raw tx from the error altogether: https://github.com/spesmilo/electrumx/pull/286/files

### DIFF
--- a/backend/coins/btc/electrum/client.go
+++ b/backend/coins/btc/electrum/client.go
@@ -18,6 +18,9 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/blockchain"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
@@ -120,7 +123,8 @@ func (c *client) TransactionBroadcast(transaction *wire.MsgTx) error {
 	rawTxHex := hex.EncodeToString(rawTx.Bytes())
 	txID, err := c.client.TransactionBroadcast(context.Background(), rawTxHex)
 	if err != nil {
-		return err
+		// Return a new error, stripping the rawTxHex from it, if it is there.
+		return errors.New(strings.ReplaceAll(err.Error(), fmt.Sprintf("[%s]", rawTxHex), ""))
 	}
 	if txID != transaction.TxHash().String() {
 		return errp.New("Response is unexpected (transaction hash mismatch)")


backend/coins/btc/handlers/handlers.go
```diff
@@ -493,7 +493,11 @@ func (handlers *Handlers) postAccountSendTx(r *http.Request) (interface{}, error
 		return map[string]interface{}{"success": false, "aborted": true}, nil
 	}
 	if err != nil {
-		handlers.log.WithError(err).Error("Failed to send transaction")
+		if strings.Contains(err.Error(), "min relay fee not met") {
```
bznein, line 496: I feel like this string could go in a constant somewhere, I'm just not sure where. Open to suggestions!

benma: Looking at error of the source, it seems min relay fee is not the only error where the raw tx will appear, it could be any error. 

It seems more robust to actually filter out the raw tx than match one specific error, like you suggested.

In this case you could move this logic to https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/backend/coins/btc/electrum/client.go#L117, closer to the source where the rawTxHex is already available.

benma (changes requested):

bznein:
@benma made the changes you requested.

The error now shows up like this (and analogous in the logs) 
image

The `[]` looks bad. I thought about replacing `[rawTxHex]` but that would be fragile in case the rawtx is not wrapped in square brackets.

We could also replace any `[]` remaining after stripping the hex.
Or we could replace rawTxHex with something like <redacted> but that would probably confusing for users. 

WDYT?

benma:
> @benma made the changes you requested.
> 
> The error now shows up like this (and analogous in the logs) image
> 
> The `[]` looks bad. I thought about replacing `[rawTxHex]` but that would be fragile in case the rawtx is not wrapped in square brackets.
> 
> We could also replace any `[]` remaining after stripping the hex. Or we could replace rawTxHex with something like `<redacted>` but that would probably be confusing for users.
> 
> WDYT?

Why not just remove `[<tx>]` including the brackets? According to the electrumx source the brackets are always there.

bznein:
> > @benma made the changes you requested.
> > The error now shows up like this (and analogous in the logs) image
> > The `[]` looks bad. I thought about replacing `[rawTxHex]` but that would be fragile in case the rawtx is not wrapped in square brackets.
> > We could also replace any `[]` remaining after stripping the hex. Or we could replace rawTxHex with something like `<redacted>` but that would probably be confusing for users.
> > WDYT?
> 
> Why not just remove `[<tx>]` including the brackets? According to the electrumx source the brackets are always there.

Makes sense, I thought maybe it's a bit more fragile in case they change the error message, but maybe it's unlikely enough!

benma (approved):
utACK

PR #3130, opened 2025-01-16
backend: show deducted amount only for non-erc20.

bznein:
Move the logic to compute the deducted amount out of the handlers exclude erc20 tokens from the computation.

### DIFF
--- a/frontends/web/src/components/transactions/transaction.tsx
+++ b/frontends/web/src/components/transactions/transaction.tsx
@@ -33,8 +33,7 @@ type TTransactionProps = ITransaction & {
 export const Transaction = ({
   addresses,
   amountAtTime,
-  fee,
-  deductedAmount,
+  deductedAmountAtTime,
   onShowDetail,
   internalID,
   note,
@@ -67,8 +66,8 @@ export const Transaction = ({
           type={type}
         />
         <Amounts
-          amount={type === 'send' ? deductedAmount : amountAtTime}
-          fee={fee}
+          amount={amountAtTime}
+          deductedAmount={deductedAmountAtTime}
           type={type}
         />
         <button
@@ -151,18 +150,18 @@ const Status = ({
 
 type TAmountsProps = {
   amount: IAmount;
-  fee: IAmount;
+  deductedAmount: IAmount,
   type: TTransactionType;
 }
 
 const Amounts = ({
   amount,
-  fee,
+  deductedAmount,
   type,
 }: TAmountsProps) => {
   const txTypeClass = `txAmount-${type}`;
-  const sendToSelf = type === 'send_to_self';
   const sign = getTxSign(type);
+  const recv = type === 'receive';
 
   return (
     <span className={`${styles.txAmountsColumn} ${styles[txTypeClass]}`}>
@@ -170,16 +169,16 @@ const Amounts = ({
       <span className={styles.txAmount}>
         {sign}
         <Amount
-          amount={sendToSelf ? fee.amount : amount.amount}
-          unit={amount.unit}
+          amount={recv ? amount.amount : deductedAmount.amount}
+          unit={recv ? amount.unit : deductedAmount.unit}
         />
         <span className={styles.txUnit}>
           {' '}
-          {sendToSelf ? fee.unit : amount.unit}
+          {deductedAmount.unit}
         </span>
       </span>
       {/* </data> */}
-      <ConversionAmount amount={amount} type={type} />
+      <ConversionAmount amount={amount} deductedAmount={deductedAmount} type={type} />
     </span>
   );
 };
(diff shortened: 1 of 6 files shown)


benma (commented):
https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/690e15561e6a6e088c59526d52cefee347cce97e/frontends/web/src/components/transactions/transaction.tsx#L178-L181

Shouldn't this now simply be

    amount={amount.amount}

cc @Beerosagos

And:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/690e15561e6a6e088c59526d52cefee347cce97e/frontends/web/src/components/transactions/transaction.tsx#L71

What about send-to-self here?

Generally `          amount={type === 'send' ? deductedAmount : amountAtTime}` is very confusing - shouldn't `deductedAmount` be named `deductedAmountAtTime` for consistency if it is the historical value?

Beerosagos:
> Generally           amount={type === 'send' ? deductedAmount : amountAtTime} is very confusing - shouldn't deductedAmount be named deductedAmountAtTime for consistency if it is the historical value?

@benma technically yes, the 'AtTime' postfix is used to indicates the amount with historical rates conversions (in order to distinguish it from the current-rates-conversions amount). But I don't think it makes much sense to rename 'deductedAmount' to 'deductedAmountAtTime`, as we don't need to distinguish it from a deductedAmount with current rates. No strong opinion, anyway.
 I'm starting to think that it would be better to move conversions outside the Amount struct. Not something for this PR, anyway.

bznein:
@benma this is again ready for review. Please note that I removed the computation of `deductedAmount` for the `sendToSelf` case. In that case we do not need a `deductedAmount` since we already treat `send_to_self` differently and we show `fee` separately from `amount`: https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/frontends/web/src/components/transactions/transaction.tsx#L184-L202

image

bznein:
As discussed offline, I changed the logic to pass everything to `Amounts` without any `txType` check.

The checks are done inside `Amounts` 


This is the situation now:

- A new field `deductedAmountAtTime` is presented by the backend. 
   - for a `receive` tx, this is set to 0
   - for a `send` tx, this is equal to the amount sent + the fee
   - for a `send-to-self` tx, this is equal to the fee
   
 Effectively, this represents how much money left the account when performing the tx.
 
 In the frontend, both `amountAtTime` and `deductedAmountAtTime` are passed to `Amounts`.
 
 `Amounts` then uses `deductedAmount` for `send` and `send-to-self` TXs, and `amount` for `receive` TXs.
 
 Additionally, `ConversionAmount` (new function introduced by https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3128) is called with an argument that is `deductedAmount` in case of a `send` (in this case we want to show the conversion for the entirety of the balance that left the account), but it's `amount` in the case of a `recv` (as `deductedAmount` is 0) and a `send-to-self` (because in that case we show something different, i.e. the value that was sent back to itself, without the fee)

frontends/web/src/components/transactions/transaction.tsx
```diff
@@ -151,35 +150,36 @@ const Status = ({
...
         />
         <span className={styles.txUnit}>
           {' '}
-          {sendToSelf ? fee.unit : amount.unit}
+          {deductedAmount.unit}
         </span>
       </span>
       {/* </data> */}
-      <ConversionAmount amount={amount} type={type} />
+      <ConversionAmount amount={recv || sendToSelf ? amount : deductedAmount} type={type} />
```
benma, line 182: Same as with Amounts, it would probably make sense to pass amount, deductedAmount and type here and handle the type logic inside. Otherwise  we have the same confusion as before by sending in different data depending on the type, and inside the component doing different things based on the type.

`ConversionAmount` is also the wrong name for a component that in some cases does not show any conversions at all, but we already have the `Amount` component too, so it's not clear to me how to name these thing :sweat_smile: We can postpone that.

benma (changes requested):

frontends/web/src/components/transactions/transaction.tsx
bznein: > Same as with Amounts, it would probably make sense to pass amount, deductedAmount and type here and handle the type logic inside. Otherwise we have the same confusion as before by sending in different data depending on the type, and inside the component doing different things based on the type.
> 

Done

> `ConversionAmount` is also the wrong name for a component that in some cases does not show any conversions at all, but we already have the `Amount` component too, so it's not clear to me how to name these thing 😅 We can postpone that.

I can see that :) I didn't name it originally and I think if we want to rename it can be done in a different PR :)

benma (approved):
tACK, thank a lot!

PR #3132, opened 2025-01-20
frontend: skeleton when txhistory conversion amount isnt available

shonsirsha:
Sometimes, when the connection is slow, the conversion amount on the tx history list will render "---".

This is actually down in the `amount` component. However, we'd like to use skeleton here on the txlist to better indicate loading for this particular view.

Showing the skeleton:

rup

### DIFF
--- a/frontends/web/src/components/amount/amount.tsx
+++ b/frontends/web/src/components/amount/amount.tsx
@@ -126,7 +126,7 @@ export const FormattedAmount = ({
   unit,
   removeBtcTrailingZeroes,
   alwaysShowAmounts = false,
-}: TProps) => {
+}: Omit<TProps, 'allowRotateCurrencyOnMobile'>) => {
   const { hideAmounts } = useContext(AppContext);
   const { decimal, group } = useContext(LocalizationContext);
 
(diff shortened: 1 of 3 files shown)


benma:
@Beerosagos is the amount not shown when the historical conversion rates are not avaliable, i.e. waiting for them to be downloaded? If so, this could take a while right? Not sure if a loading skeleton is suitable if it could take minutes to become available :thinking:

Beerosagos:
> @Beerosagos is the amount not shown when the historical conversion rates are not avaliable, i.e. waiting for them to be downloaded? If so, this could take a while right? Not sure if a loading skeleton is suitable if it could take minutes to become available 🤔

Yeah, I agree. If the tx is still unconfirmed, or recent enough (less than 2hours), the amount is estimated using the current rates. But otherwise, if the historical rates are not available, the '---' is shown and that could last for a long-ish time.

thisconnect:
code LGTM

about the visual change:

On the one hand it might confuse people to see loading skeleton when they scroll down.

On the other hand, not noticing anything when searching for a tx  and then detecting the subtle  "---" might be even more confusing :) 

so personally I find the "more in your face" skeleton ok, but no strong opinion.

thisconnect (changes requested):

PR #3133, opened 2025-01-20
frontend: increase confirm send width

shonsirsha:
Previous size was a bit too small causing selected coins (if coin control is enabled) to be broken into two lines. We increase the size so it stays in one line.


thisconnect (approved):
untested LGTM

PR #3136, opened 2025-01-21
backend: replace ratelimiter with x/time/rate.

bznein:
Removes the ratelimiter package and replaces its usages with x/time/rate.


backend/coins/eth/etherscan/etherscan.go
```diff
@@ -50,17 +52,22 @@ const ERC20GasErr = "insufficient funds for gas * price + value"
 type EtherScan struct {
 	url        string
 	httpClient *http.Client
+	limiter    rate.Limiter
```
benma, line 55: usually `*rate.Limiter` so you don't need to dereference `*rate.NewLimiter(...)` below, or is there a reason for making this a value instead of a pointer?

backend/rates/history.go
```diff
@@ -278,38 +278,36 @@ func (updater *RateUpdater) fetchGeckoMarketRange(ctx context.Context, coin, fia
 	}
 
 	// Make the call, abiding the upstream rate limits.
-	msg := fmt.Sprintf("fetch coingecko coin=%s fiat=%s start=%s", coin, fiat, timeRange.start)
+	//	msg := fmt.Sprintf("fetch coingecko coin=%s fiat=%s start=%s", coin, fiat, timeRange.start)
```
benma, line 281: undo?

bznein: Hey yeah, the issue is that this was passed to a fn call that doesn't exist anymore, I think for logging purposes and this is a bit harder now to achieve. I was planning to find a way around this and only then open for review :)

bznein: I don't think there is a way here to have the log message show up periodically as we did before. `Wait()` is a blocking call  and we have no way of loggin periodically as we did before.

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -35,11 +35,13 @@ import (
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"golang.org/x/time/rate"
 )
 
-// CallInterval is the duration between etherscan requests.
+// CallsPerSec is thenumber of etherscanr equests allowed
+// per second.
 // Etherscan rate limits to one request per 0.2 seconds.
-var CallInterval = 260 * time.Millisecond
+var CallsPerSec = 3.8
```
bznein, line 44: ~= 1/0.26 (previous duration between requests)

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -50,17 +52,22 @@ const ERC20GasErr = "insufficient funds for gas * price + value"
...
 func NewEtherScan(url string, httpClient *http.Client) *EtherScan {
 	return &EtherScan{
 		url:        url,
 		httpClient: httpClient,
+		limiter:    rate.NewLimiter(rate.Limit(CallsPerSec), 1),
 	}
 }
 
 func (etherScan *EtherScan) call(params url.Values, result interface{}) error {
+	if err := etherScan.limiter.Wait(context.TODO()); err != nil {
```
bznein, line 68: `Wait` needs a `context` that we do not have here. I'm open to suggestions. FWIW we didn't have it before either

bznein (commented):
@benma this is overall ready for review. 

the changes are mostly on `backend/rates` and `backend/coins/eth/etherscan`

I was able to test them in `rates` by adding a log line whenever we make the API call and decrease massively the number of calls allowed per second.

However, as discussed offline, I currently have no way to test them on `etherscan`, so I won't be anyway merging this until I can.

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -323,7 +330,7 @@ func (etherScan *EtherScan) Transactions(
 	result := struct {
 		Result []*Transaction
 	}{}
-	if err := etherScan.call(params, &result); err != nil {
+	if err := etherScan.call(context.TODO(), params, &result); err != nil {
```
bznein, line 333: the linter was forcing me to pass the context down the stack (which I think makes absolutely sense!) but there are two places (this function and another one) in which we don't have it, so I'm open to suggestions here

bznein:
Closing in favour of https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3155

PR #3138, opened 2025-01-21
backend: add BtcDirect to exchange deals.

bznein:
Remove the hardcoded BtcDirect widget from the frontend code, ad instead add it to the returned list of exchange deals in the backend.

### DIFF
--- a/backend/exchanges/btcdirect.go
+++ b/backend/exchanges/btcdirect.go
@@ -20,25 +20,59 @@ import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 )
 
+const (
+	// BTCDirectName is the name of the exchange, it is unique among all the supported exchanges.
+	BTCDirectName = "btcdirect"
+)
+
 var regions = []string{
 	"AD", "AT", "BE", "BG", "CH", "CZ", "DE", "DK", "EE", "ES", "FI", "FR", "GR",
 	"HR", "HU", "IS", "IT", "LI", "LT", "LU", "LV", "MD", "ME", "MK", "NL", "NO",
 	"PL", "PT", "RO", "RS", "SE", "SI", "SK", "SM"}
 
-// GetBtcDirectSupportedRegions reutrn a string slice of the regions where BTC direct services
+// GetBtcDirectSupportedRegions returns a string slice of the regions where BTC direct services
 // are available.
 func GetBtcDirectSupportedRegions() []string {
 	return regions
 }
 
-// IsBtcDirectSupported is true if coin.Code and region are supported by BtcDirect.
-func IsBtcDirectSupported(coinCode coin.Code, region string) bool {
+// isRegionSupportedBtcDirect returns whether a specific region (or an empty one
+// for "unspecified") is supported by BtcDirect.
+func isRegionSupportedBtcDirect(region string) bool {
+	return len(region) == 0 || slices.Contains(regions, region)
+}
+
+// IsBtcDirectSupported is true if coin.Code is supported by BtcDirect.
+func IsBtcDirectSupported(coinCode coin.Code) bool {
 	supportedCoins := []coin.Code{
 		coin.CodeBTC, coin.CodeTBTC, coin.CodeLTC, coin.CodeTLTC, coin.CodeETH, coin.CodeSEPETH,
 		"eth-erc20-usdt", "eth-erc20-usdc", "eth-erc20-link"}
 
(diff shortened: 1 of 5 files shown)


backend/exchanges/btcdirect.go
```diff
@@ -31,14 +36,31 @@ func GetBtcDirectSupportedRegions() []string {
 	return regions
 }
 
-// IsBtcDirectSupported is true if coin.Code and region are supported by BtcDirect.
-func IsBtcDirectSupported(coinCode coin.Code, region string) bool {
+// IsBtcDirectSupported is true if coin.Code is supported by BtcDirect.
+func IsBtcDirectSupported(coinCode coin.Code) bool {
```
bznein, line 40: I changed this logic to make it matching the structure we use for Pocket and Moonpay (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/4815ca377bc7f3e912830d7ae0195244286f9223/backend/exchanges/exchanges.go#L138-L139). To check that the region is supported we can use (as below) a simple `slice.contains()`


As an aside, I wonder if we would benefit by declaring these as interfaces instead, so every one will have methods with the exact same signature

backend/exchanges/btcdirect.go
```diff
@@ -31,14 +36,31 @@ func GetBtcDirectSupportedRegions() []string {
...
+		ExchangeName: BTCDirectName,
+		Deals: []*ExchangeDeal{
+			{
+				Fee:     4.9,
+				Payment: CardPayment,
+				IsFast:  true,
+			},
+			{
+				Fee:     1.9,
+				Payment: BankTransferPayment,
```
bznein, line 62: I took these values from the info box that compare after clicking "next", but I'm not sure if they are 100% correct

bznein: I know read it more carefully and those are placeholders values copied from the MoonPay one :D

thisconnect: yeah they are just a copy, afaik we use those value in the in app guide

Screenshot 2025-01-21 at 18 32 41

We don't need fee for the BTC Direct widget but I guess for the BTC Direct in-app guide.

thisconnect (approved):
frontend changes LGTM :)

backend/handlers/handlers.go
```diff
@@ -1370,7 +1370,7 @@ func (handlers *Handlers) getBtcDirectSupported(r *http.Request) interface{} {
 	regionCode := r.URL.Query().Get("region")
 	return Result{
 		Success:   true,
-		Supported: exchanges.IsBtcDirectSupported(acct.Coin().Code(), regionCode),
+		Supported: exchanges.IsBtcDirectSupported(acct.Coin().Code()) && exchanges.IsRegionSupportedBtcDirect(regionCode),
```
Beerosagos, line 1373: I think that we should refactor a bit this endpoint:
- rename to getExchangeBtcDirectOTCSupported (don't know why it wasn't already like that)
- maybe move the logic inside a dedicated method exchanges.IsBtcDirectOTCSupported ?

Beerosagos: Also, could you add `IsBtcDirectSupported` check inside `getExchangeSupported` (and maybe move that logic inside exchanges too)? :pray: 

Otherwise we won't consider btcDirect when loading the accounts list in the buy&sell page

backend/exchanges/btcdirect.go
Beerosagos: Good point, please do it! :D

It could also be used to uniform the endpoints that provide info for the exchange widgets loading, e.g., `getExchangeMoonpayBuyInfo` and `getExchangePocketURL`, but maybe it is something for a future PR.

backend/exchanges/btcdirect.go
Beerosagos: We need fees and payment methods to correctly load the badges in the `BuySell` compontent

Beerosagos (changes requested):

backend/exchanges/btcdirect.go
bznein: > Good point, please do it! :D

Do you mean refactoring as an interface? If that's ok for you, I'd create an issue and work on that on a separate PR :)

bznein: (I actually had a look and it takes a bit more work than I thought. Still worth doing imho)

Beerosagos (approved):
utACK :pray:

PR #3139, opened 2025-01-22
frontend: set app status to default on wallet restore/create abort

shonsirsha:
Not setting the app status to default after aborting, will cause
the app to be stuck on the same view.

The view:

Screenshot 2025-01-22 at 03 55 13


thisconnect (approved):
tested LGTM

just fyi: I ran into the re-insert bug that will be fixed next firmware release see https://github.com/BitBoxSwiss/bitbox02-firmware/commit/f4c67e8047ed57ebe571d8d4c6c34f048f87abd5

PR #3140, opened 2025-01-23
frontend: add dummy get-btc-direct-info api 

thisconnect:


frontends/web/src/api/exchanges.ts
```diff
@@ -99,6 +99,28 @@ export const getPocketURL = (action: TExchangeAction): Promise<TPocketUrlRespons
...
+};
+
+export const getBTCDirectInfo = (
+  action: TExchangeAction,
+): Promise<TBTCDirectInfoResponse> => {
+  console.log(action);
+  // TODO: change to return apiGet(`exchange/btc-direct/info/${action}`); or similar
+  return Promise.resolve({
+    success: true,
+    url: '/btcdirect/fiat-to-coin.html', // local static file for testing
```
thisconnect, line 118: I expect that in production mode we will point to a subdomain where we host the content on fiat-to-coin.html

PR #3141, opened 2025-01-23
frontend/android: upgrade gradle to 8.8.0

Beerosagos:
Done using Android Studio upgrade assistant.

### DIFF
--- a/frontends/android/BitBoxApp/gradle/wrapper/gradle-wrapper.properties
+++ b/frontends/android/BitBoxApp/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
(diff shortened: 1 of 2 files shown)


frontends/android/BitBoxApp/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10.2-bin.zip
```
benma, line 3: 8.10 or 8.8?

Beerosagos: This line refers to the version of the Gradle build tool, which is a separate entity from the Gradle plugin itself (which is updated to v8.8.0). See also https://developer.android.com/build/releases/gradle-plugin#updating-gradle

benma (approved):
utACK

PR #3142, opened 2025-01-27
   backend: provide getBtcDirectInfo API endpoint

bznein:
Provide an endpoint to be called from the frontend. This endpoints returns the APIKey, the URL address and, if the action is "buy", the address.

### DIFF
--- a/backend/exchanges/btcdirect.go
+++ b/backend/exchanges/btcdirect.go
@@ -17,13 +17,28 @@ package exchanges
 import (
 	"slices"
 
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 )
 
 const (
 	// BTCDirectName is the name of the exchange, it is unique among all the supported exchanges.
 	BTCDirectName = "btcdirect"
+
+	btcDirectTestApiKey = "6ed4d42bd02eeac1776a6bb54fa3126f779c04d5c228fe5128bb74e89ef61f83"
+
+	btcDirectProdAPiKey = "7d71f633626901d5c4d06d91f7d0db2c15cdf524ddd0ebcd36f4d9c4e04694cd"
+
+	btcDirectTestUrl = "/btcdirect/fiat-to-coin.html"
+
+	btcDirectProdUrl = "https://bitboxapp.shiftcrypto.io/widgets/btcdirect/v1/fiat-to-coin.html"
 )
+
+type btcDirectInfo struct {
+	Url     string
+	ApiKey  string
+	Address *string
+}
 
 var regions = []string{
 	"AT", "BE", "BG", "CH", "CY", "CZ", "DE", "DK", "EE", "ES", "FI", "FR", "GR",
@@ -75,4 +90,24 @@ func BtcDirectDeals() *ExchangeDealsList {
 			},
 		},
 	}
+}
+
+// BtcDirectInfo returns the information needed to interact with BtcDirect.
+// If `devServers` is true, it returns testing URL and ApiKey.
+func BtcDirectInfo(action ExchangeAction, acct accounts.Interface, devServers bool) btcDirectInfo {
+	res := btcDirectInfo{
+		Url:    btcDirectProdUrl,
+		ApiKey: btcDirectProdAPiKey,
+	}
+
+	if devServers {
+		res.Url = btcDirectTestUrl
+		res.ApiKey = btcDirectTestApiKey
+	}
+
+	if action == BuyAction {
+		addr := acct.GetUnusedReceiveAddresses()[0].Addresses[0].EncodeForHumans()
+		res.Address = &addr
+	}
+	return res
 }
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -243,6 +243,7 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/exchange/deals/{action}/{code}", handlers.getExchangeDeals).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/supported/{code}", handlers.getExchangeSupported).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/btcdirect-otc/supported/{code}", handlers.getExchangeBtcDirectOTCSupported).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/exchange/btcdirect/info/{action}/{code}", handlers.getExchangeBtcDirectInfo).Methods("GET")
 	getAPIRouter(apiRouter)("/exchange/moonpay/buy-info/{code}", handlers.getExchangeMoonpayBuyInfo).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/pocket/api-url/{action}", handlers.getExchangePocketURL).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/pocket/verify-address", handlers.postPocketWidgetVerifyAddress).Methods("POST")
@@ -1436,6 +1437,33 @@ func (handlers *Handlers) getExchangeMoonpayBuyInfo(r *http.Request) (interface{
 		Address: buy.Address,
 	}
 	return resp, nil
+}
+
+func (handlers *Handlers) getExchangeBtcDirectInfo(r *http.Request) interface{} {
+	type result struct {
+		Success      bool    `json:"success"`
+		ErrorMessage string  `json:"errorMessage"`
+		Url          string  `json:"url"`
+		ApiKey       string  `json:"apiKey"`
+		Address      *string `json:"address"`
+	}
+
+	code := accountsTypes.Code(mux.Vars(r)["code"])
+	acct, err := handlers.backend.GetAccountFromCode(code)
+	accountValid := acct != nil && acct.Offline() == nil && !acct.FatalError()
+	if err != nil || !accountValid {
+		return result{Success: false, ErrorMessage: "Account is not valid."}
+	}
+
+	action := exchanges.ExchangeAction(mux.Vars(r)["action"])
+	btcInfo := exchanges.BtcDirectInfo(action, acct, handlers.backend.DevServers())
+
+	return result{
+		Success: true,
+		Url:     btcInfo.Url,
+		ApiKey:  btcInfo.ApiKey,
+		Address: btcInfo.Address,
+	}
 }
 
 func (handlers *Handlers) getExchangePocketURL(r *http.Request) interface{} {
--- a/frontends/web/src/api/exchanges.ts
+++ b/frontends/web/src/api/exchanges.ts
@@ -103,6 +103,7 @@ export type TBTCDirectInfoResponse = {
   success: true;
   url: string;
   apiKey: string;
+  address?: string;
 } | {
   success: false;
   errorMessage: string;
@@ -110,15 +111,9 @@ export type TBTCDirectInfoResponse = {
 
 export const getBTCDirectInfo = (
   action: TExchangeAction,
+  code: string,
 ): Promise<TBTCDirectInfoResponse> => {
-  console.log(action);
-  // TODO: change to return apiGet(`exchange/btc-direct/info/${action}`); or similar
-  return Promise.resolve({
-    success: true,
-    url: '/btcdirect/fiat-to-coin.html', // local static file for testing
-    apiKey: '6ed4d42bd02eeac1776a6bb54fa3126f779c04d5c228fe5128bb74e89ef61f83', // debug
-    // apiKey: '7d71f633626901d5c4d06d91f7d0db2c15cdf524ddd0ebcd36f4d9c4e04694cd', // prod
-  });
+  return apiGet(`exchange/btcdirect/info/${action}/${code}`);
 };
 
 export type SupportedExchanges= {
--- a/frontends/web/src/routes/exchange/btcdirect.tsx
+++ b/frontends/web/src/routes/exchange/btcdirect.tsx
@@ -18,7 +18,7 @@ import { useState, useEffect, createRef, useContext, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { getBTCDirectInfo } from '@/api/exchanges';
 import { AppContext } from '@/contexts/AppContext';
-import { AccountCode, getReceiveAddressList, IAccount } from '@/api/account';
+import { AccountCode, IAccount } from '@/api/account';
 import { useLoad } from '@/hooks/api';
 import { useDarkmode } from '@/hooks/darkmode';
 import { UseDisableBackButton } from '@/hooks/backbutton';
@@ -51,13 +51,7 @@ export const BTCDirect = ({ accounts, code }: TProps) => {
   const { isDarkMode } = useDarkmode();
 
   const iframeRef = useRef<HTMLIFrameElement | null>(null);
-  const btcdirectInfo = useLoad(() => getBTCDirectInfo('buy'));
-
-  const receiveAddresses = useLoad(getReceiveAddressList(code));
-
-  // TODO: address should probably come from the backend, i.e. ETH address
-  const p2wpkhAddresses = receiveAddresses?.find(({ scriptType }) => scriptType === 'p2wpkh')?.addresses || [];
-  const address = p2wpkhAddresses[0]?.address || '';
+  const btcdirectInfo = useLoad(() => getBTCDirectInfo('buy', code));
 
   const [agreedTerms, setAgreedTerms] = useState(false);
   const [iframeLoaded, setIframeLoaded] = useState(false);
@@ -124,7 +118,7 @@ export const BTCDirect = ({ accounts, code }: TProps) => {
               <div style={{ height }}>
                 <UseDisableBackButton />
                 {!iframeLoaded && <Spinner text={t('loading')} />}
-                { address && btcdirectInfo?.success && (
+                { btcdirectInfo?.success && (
                   <iframe
                     onLoad={() => {
                       setIframeLoaded(true);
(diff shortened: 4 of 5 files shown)


backend/handlers/handlers.go
```diff
@@ -243,6 +243,7 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/exchange/deals/{action}/{code}", handlers.getExchangeDeals).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/supported/{code}", handlers.getExchangeSupported).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/btcdirect-otc/supported/{code}", handlers.getExchangeBtcDirectOTCSupported).Methods("GET")
+	getAPIRouter(apiRouter)("/exchange/btcdirect-otc/info/{action}/{code}", handlers.getBtcDirectInfo).Methods("GET")
```
thisconnect, line 246: just tiny nit: probably we should drop the `-otc` part for api endpoints that are for both the widget and the otc link.

In a past release we had only the OTC link so then it made sense.

backend/exchanges/btcdirect.go
```diff
@@ -17,14 +17,27 @@ package exchanges
...
 
 const (
 	// BTCDirectName is the name of the exchange, it is unique among all the supported exchanges.
 	BTCDirectName = "btcdirect"
+
+	btcDirectTestApiKey = "6ed4d42bd02eeac1776a6bb54fa3126f779c04d5c228fe5128bb74e89ef61f83"
+
+	btcDirectProdAPiKey = "7d71f633626901d5c4d06d91f7d0db2c15cdf524ddd0ebcd36f4d9c4e04694cd"
+
+	btcUrl = "/btcdirect/fiat-to-coin.html"
```
thisconnect, line 32: maybe change to btcDirectUrl so it is consistent with the other variables ?

backend/exchanges/btcdirect.go
```diff
@@ -76,3 +89,21 @@ func BtcDirectDeals() *ExchangeDealsList {
...
 }
+
+// BtcDirectInfo returns the information needed to interact with BtcDirect.
+func BtcDirectInfo(action ExchangeAction, acct accounts.Interface) btcDirectInfo {
+	res := btcDirectInfo{
+		Url:    btcUrl,
+		ApiKey: btcDirectProdAPiKey,
+	}
+
+	if _, isTestnet := coin.TestnetCoins[acct.Coin().Code()]; isTestnet {
```
Beerosagos, line 100: I think it would be more correct to check for `backend.DevServers()`, instead of checking for testnet accounts. It can be recovered from the handler and passed here as an argument. But I see that we do the same for Moonpay, so no strong opinion.

backend/exchanges/btcdirect.go
```diff
@@ -76,3 +89,21 @@ func BtcDirectDeals() *ExchangeDealsList {
...
+		res.ApiKey = btcDirectTestApiKey
+	}
+
+	if action == BuyAction {
+		addr := acct.GetUnusedReceiveAddresses()[0].Addresses[0].EncodeForHumans()
```
Beerosagos, line 105: I don't like it much to get the first script type in the addresses list without checking which one it is, but I see that we do the same for Moonpay, so I guess it is okay.. wdyt @benma ?

backend/handlers/handlers.go
```diff
@@ -1438,6 +1439,32 @@ func (handlers *Handlers) getExchangeMoonpayBuyInfo(r *http.Request) (interface{
 	return resp, nil
 }
 
+func (handlers *Handlers) getBtcDirectInfo(r *http.Request) (interface{}, error) {
+	code := accountsTypes.Code(mux.Vars(r)["code"])
+	acct, err := handlers.backend.GetAccountFromCode(code)
+	if err != nil {
+		return nil, err
+	}
+
```
Beerosagos, line 1448: Maybe check also for `acct != nil && acct.Offline() == nil && !acct.FatalError()` like we do in other places?

backend/handlers/handlers.go
```diff
@@ -1438,6 +1439,32 @@ func (handlers *Handlers) getExchangeMoonpayBuyInfo(r *http.Request) (interface{
 	return resp, nil
 }
 
+func (handlers *Handlers) getBtcDirectInfo(r *http.Request) (interface{}, error) {
```
Beerosagos, line 1442: Please, rename to `getExchangeBtcDirectInfo` to follow the existing convention :innocent: 
Also, returning `(interface{}, error)` is kind of deprecated, as it is not correctly handled in the frontend. Better to just return `interface{}` and use the `success` param and an error message

Beerosagos (changes requested):

backend/exchanges/btcdirect.go
bznein: yeah I just copied the code over from Moonpay :) I'm not familiar with this logic so I don't know which one would be better

backend/exchanges/btcdirect.go
bznein: Same here as before, just copied from Moonpay, open to suggestions!

backend/exchanges/btcdirect.go
thisconnect: Yes, please change to devserver. It turned out that in sandbox we still need to provide mainnet address to be able to test. We can't really test in testnet.

So locally we can test with `make servewallet-mainnet` and `make webdev`

backend/exchanges/btcdirect.go
```diff
@@ -17,14 +17,27 @@ package exchanges
...
 
 const (
 	// BTCDirectName is the name of the exchange, it is unique among all the supported exchanges.
 	BTCDirectName = "btcdirect"
+
+	btcDirectTestApiKey = "6ed4d42bd02eeac1776a6bb54fa3126f779c04d5c228fe5128bb74e89ef61f83"
+
+	btcDirectProdAPiKey = "7d71f633626901d5c4d06d91f7d0db2c15cdf524ddd0ebcd36f4d9c4e04694cd"
+
+	btcDirectUrl = "/btcdirect/fiat-to-coin.html"
```
thisconnect, line 32: please keep `/btcdirect/fiat-to-coin.html` in testing mode and for mainnet use `https://bitboxapp.shiftcrypto.io/widgets/btcdirect/v1/fiat-to-coin.html` (nothing there yet).

backend/exchanges/btcdirect.go
thisconnect: so the condition should check if the app is in mainnet, then use prod api keys and url.
if in testnet use test api key and local file.

backend/exchanges/btcdirect.go
benma: Yeah, if we change it we should change it for both in a different PR.

backend/exchanges/btcdirect.go
thisconnect: I've discussed with @benma in dm. For testing we should modify `/etc/hosts` and remove fiat-to-coin.html completely from this repo.

With that btcDirectUrlTesting is not needed anymore, sorry for the confusion.

backend/handlers/handlers.go
```diff
@@ -1438,6 +1439,33 @@ func (handlers *Handlers) getExchangeMoonpayBuyInfo(r *http.Request) (interface{
...
+		Url     string  `json:"url"`
+		ApiKey  string  `json:"apiKey"`
+		Address *string `json:"address"`
+	}{}
+
+	code := accountsTypes.Code(mux.Vars(r)["code"])
+	acct, err := handlers.backend.GetAccountFromCode(code)
+	accountValid := acct != nil && acct.Offline() == nil && !acct.FatalError()
+	if err != nil || !accountValid {
+		res.Success = false
```
benma, line 1454: Imho readability would be a bit better to define the struct only above, and instantiate locally, like `return &result{.. }`, like in this example: https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/eb7ec055d00179a9e13afdc148b0ac6aaf9463b4/backend/handlers/handlers.go#L1577

Just a minor nit though.

backend/exchanges/btcdirect.go
benma: Pocket uses `DevServers()`, which in this case makes sense I think - you want to run the local file only with `make servewallet` right? If a customer runs `BitBox -testnet` they should not be served the local file!

Pls then also update https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/eb7ec055d00179a9e13afdc148b0ac6aaf9463b4/backend/arguments/arguments.go#L51-L55, which has a mention about Pocket already.

thisconnect: > you want to run the local file only with make servewallet right?

Yes correct, `make servewallet` for the local file, and `make servewallet-mainnet` to `bitboxapp.shiftcrypto.io`.

thisconnect (approved):
untested LGTM

backend/handlers/handlers.go
```diff
@@ -1438,6 +1439,32 @@ func (handlers *Handlers) getExchangeMoonpayBuyInfo(r *http.Request) (interface{
...
+
+	code := accountsTypes.Code(mux.Vars(r)["code"])
+	acct, err := handlers.backend.GetAccountFromCode(code)
+	accountValid := acct != nil && acct.Offline() == nil && !acct.FatalError()
+	if err != nil || !accountValid {
+		return result{Success: false}
+	}
+
+	action := exchanges.ExchangeAction(mux.Vars(r)["action"])
+	btcInfo := exchanges.BtcDirectInfo(action, acct, handlers.backend.DevServers())
```
thisconnect, line 1458: I believe here it should pass Testing() to get the prod url with `make servewallet-mainnet` 

```suggestion
	btcInfo := exchanges.BtcDirectInfo(action, acct, handlers.backend.Testing())
```

thisconnect (changes requested):

backend/handlers/handlers.go
Beerosagos: @thisconnect I think this is correct. `DevServers ()` is true for both `make servewallet` and `make servewallet-mainnet`, while it is false for `servewallet-mainnet-prodservers`

backend/exchanges/btcdirect.go
```diff
@@ -17,14 +17,29 @@ package exchanges
...
 
 const (
 	// BTCDirectName is the name of the exchange, it is unique among all the supported exchanges.
 	BTCDirectName = "btcdirect"
+
+	btcDirectTestApiKey = "6ed4d42bd02eeac1776a6bb54fa3126f779c04d5c228fe5128bb74e89ef61f83"
+
+	btcDirectProdAPiKey = "7d71f633626901d5c4d06d91f7d0db2c15cdf524ddd0ebcd36f4d9c4e04694cd"
+
+	btcDirectUrlTesting = "/btcdirect/fiat-to-coin.html"
```
Beerosagos, line 32: nit: it would be nice to stick to the same naming for URL and apikey
```suggestion
	btcDirectTestUrl = "/btcdirect/fiat-to-coin.html"
```

backend/handlers/handlers.go
thisconnect: ack, I forgot about `make servewallet-mainnet-prodservers` makes sense.

thisconnect: in this case, lets get this merged? @Beerosagos ?

thisconnect (approved):

backend/handlers/handlers.go
```diff
@@ -242,7 +242,8 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/exchange/by-region/{code}", handlers.getExchangesByRegion).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/deals/{action}/{code}", handlers.getExchangeDeals).Methods("GET")
 	getAPIRouterNoError(apiRouter)("/exchange/supported/{code}", handlers.getExchangeSupported).Methods("GET")
-	getAPIRouterNoError(apiRouter)("/exchange/btcdirect-otc/supported/{code}", handlers.getExchangeBtcDirectOTCSupported).Methods("GET")
+	getAPIRouterNoError(apiRouter)("/exchange/btcdirect/supported/{code}", handlers.getExchangeBtcDirectOTCSupported).Methods("GET")
```
Beerosagos, line 245: Here it should still be `exchange/btcdirect-otc/supported`.
the OTC service of btc-direct is something a bit different from the other widget/deals things

backend/handlers/handlers.go
```diff
@@ -1438,6 +1439,32 @@ func (handlers *Handlers) getExchangeMoonpayBuyInfo(r *http.Request) (interface{
...
+		Url     string  `json:"url"`
+		ApiKey  string  `json:"apiKey"`
+		Address *string `json:"address"`
+	}
+
+	code := accountsTypes.Code(mux.Vars(r)["code"])
+	acct, err := handlers.backend.GetAccountFromCode(code)
+	accountValid := acct != nil && acct.Offline() == nil && !acct.FatalError()
+	if err != nil || !accountValid {
+		return result{Success: false}
```
Beerosagos, line 1454: The frontend result type also has an `errorMessage`. I'd declare it here too and use it to return "Account not valid" or smth similar.

backend/handlers/handlers.go
Beerosagos: Left a few comments

frontends/web/src/api/exchanges.ts
```diff
@@ -140,6 +135,6 @@ export type TBtcDirectResponse = {
 
 export const getBtcDirectOTCSupported = (code: AccountCode, region: ExchangeRegion['code']) => {
   return (): Promise<TBtcDirectResponse> => {
-    return apiGet(`exchange/btcdirect-otc/supported/${code}?region=${region}`);
+    return apiGet(`exchange/btcdirect/supported/${code}?region=${region}`);
```
Beerosagos, line 138: Same here: see the comment in handlers.go

Beerosagos (changes requested):

backend/exchanges/btcdirect.go
Beerosagos: Using devServers it would use the local file also for `make servewallet-mainnet`, see my other comment below :)

Do we plan to keep the local files in the future, or do we want to point to `bitboxapp.shiftcrypto.io` for the sandbox too, eventually?

thisconnect: For now the local file helps me a lot for local development, i.e. changing communication to postmessage was super simple to implement, https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3146
(not this is already on bitboxapp.shiftcrypto.io and the PR is only to keep both versions aligned)

We can remove the local file in the future if you prefer.

thisconnect: > do we want to point to bitboxapp.shiftcrypto.io for the sandbox too, eventually?

yes that could make sense but sandbox needs a mainnet address, I guess it should use `make servewallet-mainnet-prodservers` for production (as you recommended) and  `make servewallet-mainnet` for sandbox?

thisconnect:
I tested this commit with the frontend change see pr https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3146

The widget loads from shiftcrypto.io  tested on macos qt 🎉  build from #3146 .

In -testnet the widget loads their sandbox but shows an wrong api key error. 

Testnet addresses are not supported in their sandbox so we can't really test in their sandbox at the moment, but this is currently expected.

Beerosagos:
> In -testnet the widget loads their sandbox but shows an wrong api key error.

Do you mean that the API key actually depends on testnet vs mainnet?

thisconnect:
> Do you mean that the API key actually depends on testnet vs mainnet?

They provide 2 api keys one for prod and one for sandbox, unfortunately both currently only support mainnet addresses. So we can't really test in -testnet.

However with the local files the widget still loads in testnet, with the restriction that it errors at some point during the order process, as testnet addresses are not supported.

In my draft PR https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3146  it errors earlier with wrong api key, but that is something I would like to take a look at with BTC Direct once we have this and the frontend PR merged.

backend/exchanges/btcdirect.go
```diff
@@ -17,14 +17,29 @@ package exchanges
...
+
+	btcDirectUrlProd = "https://bitboxapp.shiftcrypto.io/widgets/btcdirect/v1/fiat-to-coin.html"
```
thisconnect, line 34: FYI, in this is just the buy widget, in a 2nd step we should also add the sell widget i.e. /btcdirect/coin-to-fiat.html

frontends/web/src/api/exchanges.ts
```diff
@@ -103,22 +103,17 @@ export type TBTCDirectInfoResponse = {
   success: true;
   url: string;
   apiKey: string;
+  address?: string;
```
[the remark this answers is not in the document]
bznein: Can you elaborate on this? My frontend knowledge is EXTREMELY limited :D

frontends/web/src/routes/exchange/btcdirect.tsx
```diff
@@ -51,13 +51,9 @@ export const BTCDirect = ({ accounts, code }: TProps) => {
...
   const iframeRef = useRef<HTMLIFrameElement | null>(null);
-  const btcdirectInfo = useLoad(() => getBTCDirectInfo('buy'));
+  const btcdirectInfo = useLoad(() => getBTCDirectInfo('buy', code));
 
-  const receiveAddresses = useLoad(getReceiveAddressList(code));
-
-  // TODO: address should probably come from the backend, i.e. ETH address
-  const p2wpkhAddresses = receiveAddresses?.find(({ scriptType }) => scriptType === 'p2wpkh')?.addresses || [];
-  const address = p2wpkhAddresses[0]?.address || '';
+  const address = btcdirectInfo?.success ? btcdirectInfo.address : '';
```
bznein, line 56: Not super super sure how to handle the case in which success is false here, though it is also handled below (line 123)

thisconnect: It will only load the iframe if address has a "truthy" value (see below `{ address && `), empty string is falsy.

Truthy/falsy is often a bit weird to read. You could change from empty string to undefined then it might be more clear.

Beerosagos: I don't think you need to define the `address` const at all. There is a check for `btcdirectInfo?.success` just before including the iframe, so you can just call `btcdirectInfo.address` there, as done for the apiKey and the url.

bznein: Good point, done!

Beerosagos (approved):
partially tested ACK

PR #3152, opened 2025-02-03
frontend: use postmessage to communicate with the iframe

thisconnect:
The previous approach depended on the iframe having access to the parent context, but this wont work when the iframe is hosted on a different domain.

Instead of relaxing webengine restrictions this change simply uses postMessage api to pass data from and to the iframe.

### DIFF
--- a/frontends/web/src/routes/exchange/btcdirect.tsx
+++ b/frontends/web/src/routes/exchange/btcdirect.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useState, useEffect, createRef, useContext, useRef } from 'react';
+import { useState, useEffect, createRef, useContext, useRef, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { getBTCDirectInfo } from '@/api/exchanges';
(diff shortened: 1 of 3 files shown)


thisconnect:
previous draft (that included the backend change) https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3146

frontends/web/src/routes/exchange/btcdirect.tsx
```diff
@@ -87,15 +87,47 @@ export const BTCDirect = ({ accounts, code }: TProps) => {
     }, 200);
   };
 
+  const locale = i18n.resolvedLanguage ? localeMapping[i18n.resolvedLanguage] : 'en-GB';
+
+  const onMessage = useCallback((event: MessageEvent) => {
+    if (
+      !account
+      || !btcdirectInfo?.success
+      || (!isDevServers && event.origin !== 'https://bitboxapp.shiftcrypto.io')
```
Beerosagos, line 96: Could we use btcdirect.url to derive the domain?

Beerosagos (approved):
utACK with a small comment :pray:

PR #3154, opened 2025-02-04
frontend: set width for dropdowns in settings page

shonsirsha:
Currently, the width of these dropdowns in the settings page
follow the content inside it, causing it to look inconsistent.

We unify their width to improve the UI.

Before:

pjxro

After:
razu4


thisconnect (approved):
LGTM thank you

PR #3155, opened 2025-02-04
backend: replace ratelimiter with x/time/rate.

bznein:
Removes the ratelimiter package and replaces its
usages with x/time/rate.


bznein:
@benma I made a mess on the other PR trying to merge master into it. I opened a new one with the same diff.

I managed to test these changes also on the `etherscan.go` path in the same way (adding a log line and artificially limiting even further the number of calls per second) and confirmed that they do in fact happen less often.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	golang.org/x/crypto v0.32.0
 	golang.org/x/mobile v0.0.0-20240716161057-1ad2df20a8b6
 	golang.org/x/net v0.34.0
+	golang.org/x/time v0.5.0
```
benma, line 25: why v0.5.0? v0.10.0 is available also (even v0.11.0 is, but that seems to require Go 1.23).

backend/rates/gecko.go
```diff
@@ -16,21 +16,21 @@ const (
 	maxGeckoRange = 364 * 24 * time.Hour
 )
 
-// apiRateLimit specifies the minimal interval between equally spaced API calls
+// apiRateLimit specifies the maximum number of API calls per second
 // to one of the supported exchange rates providers.
-func apiRateLimit(baseURL string) time.Duration {
+func apiRateLimit(baseURL string) float64 {
```
benma, line 21: make the return type `rate.Limit` for clarity? then the callsite does not need to convert either.

backend/rates/rates.go
```diff
@@ -119,15 +120,15 @@ type RateUpdater struct {
...
+	geckoLimiter *rate.Limiter
 }
 
 // NewRateUpdater returns a new rates updater.
 // The dbdir argument is the location of a historical rates database cache.
 // The returned updater can function without a valid database cache but may be
 // impacted by rate limits. The database cache is transparent to the updater users.
 // To stay within acceptable rate limits defined by CoinGeckoRateLimit, callers can
-// use util/ratelimit package.
+// use https://pkg.go.dev/golang.org/x/time/rate package.
```
benma, line 131: remove this part? users don't have to do anything, rate limiting is includd in the RatesUpdater

backend/rates/rates.go
```diff
@@ -275,36 +276,42 @@ func (updater *RateUpdater) updateLast(ctx context.Context) {
 	}
 
 	var geckoRates map[string]map[string]float64
-	callErr := updater.geckoLimiter.Call(ctx, "updateLast", func() error {
-		ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
benma, line 279: you missed this line below in your http request, which should time out.

backend/rates/rates.go
```diff
@@ -275,36 +276,42 @@ func (updater *RateUpdater) updateLast(ctx context.Context) {
...
-		if err := json.Unmarshal(responseBody, &geckoRates); err != nil {
-			return errp.WithMessage(err,
-				fmt.Sprintf("could not parse rates response: %s", string(responseBody)))
-		}
-		return nil
-	})
-	if callErr != nil {
-		updater.log.WithError(callErr).Errorf("updateLast")
+	if err := updater.geckoLimiter.Wait(ctx); err != nil {
+		updater.log.WithError(err).Error("could not wait for rate limiter")
```
benma, line 280: please prefix all msgs with `updateLast:` like it used to be before, so the error is more easily identified in the logs.

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -35,11 +35,13 @@ import (
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"golang.org/x/time/rate"
 )
 
-// CallInterval is the duration between etherscan requests.
+// CallsPerSec is thenumber of etherscanr equests allowed
+// per second.
 // Etherscan rate limits to one request per 0.2 seconds.
-var CallInterval = 260 * time.Millisecond
+var CallsPerSec = 3.8
```
benma, line 44: can unexport this, it's only used locally.

benma (changes requested):
tested and seems to work, but left a few comments

go.mod
bznein: Yeah I'm not sure, I think it was auto-added. Will change it

benma: Now that we are back on 1.23 it could also be 0.11.0 :nerd_face: not very important though I assume

benma (approved):

PR #3157, opened 2025-02-04
backend/frontend: add option to restart in testnet

bznein:
Adds a button in the advanced settings that enables the app to start in testnet automatically on the next start. The setting is only valid for the next start, and it gets reset after that.

image

(note: the "enabled" text near the option has been removed, this is a slightly older screenshot)

Note that this PR does not automatically restart the application. This would require a lot more investigation on how to do it properly on all platforms.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -1196,7 +1196,7 @@ outer:
 // a user-facing setting. Now we simply use it for migration to decide which coins to add by
 // default.
 func (backend *Backend) persistDefaultAccountConfigs(keystore keystore.Keystore, accountsConfig *config.AccountsConfig) error {
-	if backend.arguments.Testing() {
+	if backend.Testing() {
 		if backend.arguments.Regtest() {
 			if backend.config.AppConfig().Backend.DeprecatedCoinActive(coinpkg.CodeRBTC) {
 				if _, err := backend.createAndPersistAccountConfig(
@@ -1458,7 +1458,7 @@ func (backend *Backend) maybeAddHiddenUnusedAccounts() {
 	switch {
 	case backend.arguments.Regtest():
 		coinCodes = []coinpkg.Code{coinpkg.CodeRBTC}
-	case backend.arguments.Testing():
+	case backend.Testing():
 		coinCodes = []coinpkg.Code{coinpkg.CodeTBTC, coinpkg.CodeTLTC}
 	default:
 		coinCodes = []coinpkg.Code{coinpkg.CodeBTC, coinpkg.CodeLTC}
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -232,20 +232,23 @@ type Backend struct {
 	tstCheckAccountUsed func(accounts.Interface) bool
 	// For unit tests, called when `backend.maybeAddHiddenUnusedAccounts()` has run.
 	tstMaybeAddHiddenUnusedAccounts func()
+
+	// testing tells us whether the app is in testing mode
+	testing bool
 }
 
 // NewBackend creates a new backend with the given arguments.
 func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backend, error) {
 	log := logging.Get().WithGroup("backend")
-	config, err := config.NewConfig(arguments.AppConfigFilename(), arguments.AccountsConfigFilename())
+	backendConfig, err := config.NewConfig(arguments.AppConfigFilename(), arguments.AccountsConfigFilename())
 	if err != nil {
 		return nil, errp.WithStack(err)
 	}
-	log.Infof("backend config: %+v", config.AppConfig().Backend)
-	log.Infof("frontend config: %+v", config.AppConfig().Frontend)
+	log.Infof("backend config: %+v", backendConfig.AppConfig().Backend)
+	log.Infof("frontend config: %+v", backendConfig.AppConfig().Frontend)
 	backendProxy := socksproxy.NewSocksProxy(
-		config.AppConfig().Backend.Proxy.UseProxy,
-		config.AppConfig().Backend.Proxy.ProxyAddress,
+		backendConfig.AppConfig().Backend.Proxy.UseProxy,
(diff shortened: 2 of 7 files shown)


backend/accounts.go
```diff
@@ -1196,7 +1196,7 @@ outer:
 // a user-facing setting. Now we simply use it for migration to decide which coins to add by
 // default.
 func (backend *Backend) persistDefaultAccountConfigs(keystore keystore.Keystore, accountsConfig *config.AccountsConfig) error {
-	if backend.arguments.Testing() {
+	if backend.Testing() {
```
bznein, line 1199: I replaced all occurrences of `backend.arguments.Testing()` with `backend.Testing`. Prior to this PR the two calls are the same (`backend.Testing()` calls `backend.arguments.Testing()`) but with this new PR there is more logic in `backend.Testing()` - The alternative would have be to change the field `arguments.testing` but since everything in arguments is unexported I prefer not to

backend/backend.go
```diff
@@ -232,30 +232,34 @@ type Backend struct {
...
 	tstMaybeAddHiddenUnusedAccounts func()
+
+	restartingInTestnet bool
 }
 
 // NewBackend creates a new backend with the given arguments.
 func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backend, error) {
 	log := logging.Get().WithGroup("backend")
-	config, err := config.NewConfig(arguments.AppConfigFilename(), arguments.AccountsConfigFilename())
+	backendConfig, err := config.NewConfig(arguments.AppConfigFilename(), arguments.AccountsConfigFilename())
```
bznein, line 242: renaming because I later need to use `config` as an import name

frontends/web/src/locales/en/app.json
```diff
@@ -1318,6 +1318,9 @@
       "customFees": {
         "description": "Lets you enter your own fee when sending."
       },
+      "restartInTestnet": {
+        "description": "Use testnet instead of mainnet."
```
bznein, line 1322: I'm not good with words :P I'm open to suggestions!

frontends/web/src/routes/settings/components/advanced-settings/restart-in-testnet-setting.tsx
```diff
@@ -0,0 +1,69 @@
+
+/**
+ * Copyright 2023-2024 Shift Crypto AG
```
benma, line 3: 2025

backend/backend.go
```diff
@@ -232,30 +232,34 @@ type Backend struct {
 	tstCheckAccountUsed func(accounts.Interface) bool
 	// For unit tests, called when `backend.maybeAddHiddenUnusedAccounts()` has run.
 	tstMaybeAddHiddenUnusedAccounts func()
+
+	restartingInTestnet bool
```
benma, line 236: This is redundant with `backend.config.AppConfig().Backend.RestartInTestnet`, so I would remove this here to improve clarity. Otherwise I am inclined  to think that this value is somehow special or different to what's in the config.

backend/config/config.go
```diff
@@ -102,6 +102,9 @@ type Backend struct {
 
 	// BtcUnit is the unit used to represent Bitcoin amounts. See `coin.BtcUnit` for details.
 	BtcUnit coin.BtcUnit `json:"btcUnit"`
+
+	// RestartInTestnet represents whether the app should launch in testnet on the next start.
+	RestartInTestnet bool `json:"restartInTestnet"`
```
benma, line 107: Maybe a better name is `StartInTestnet` or just `Testnet`? The flag is used to turn into testnet when starting, not before. This way it sounds like it would trigger a restart or something like that.

backend/backend.go
```diff
@@ -272,6 +276,16 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
+
+	// Only override if set in settings. If not, let the existing
+	// value (that could come from the --testnet flag) untouched.
+	if testnet {
+		backend.restartingInTestnet = true
+		if err := backendConfig.ModifyAppConfig(func(c *config.AppConfig) error { c.Backend.RestartInTestnet = false; return nil }); err != nil {
+			return nil, err
+		}
+	}
+
```
benma, line 288: This might be better in the `Start()` method instead of the constructor?

benma (changes requested):

backend/backend.go
bznein: of course! I completely missed that I could just use that! Thanks

benma: Right, my bad. For clarity, would this maybe help?

- rename it to `testing bool`
- set it to `backend.config.AppConfig().Backend.Testing || backend.arguments.Testing()` followed by modifying the config to set it back to `false`
- `func (backend *Backend) Testing() bool { return backend.testing }`

then the logic would be concentrated and easier to follow I think.

backend/config/config.go
```diff
@@ -102,6 +102,9 @@ type Backend struct {
 
 	// BtcUnit is the unit used to represent Bitcoin amounts. See `coin.BtcUnit` for details.
 	BtcUnit coin.BtcUnit `json:"btcUnit"`
+
+	// RestartInTestnet represents whether the app should launch in testnet on the next start.
```
benma, line 106: Mention that it resets to `false` on app launch?

backend/backend.go
```diff
@@ -638,6 +644,12 @@ func (backend *Backend) Start() <-chan interface{} {
 	backend.configureHistoryExchangeRates()
 
 	backend.environment.OnAuthSettingChanged(backend.config.AppConfig().Backend.Authentication)
+
+	if backend.testing {
```
benma, line 648: tiny nit, but shouldn't this be `backendConfig.AppConfig().Backend.StartInTestnet`? The config does not need to be touched if testing is true due to the `-testnet` flag.

benma (approved):
LGTM

PR #3158, opened 2025-02-05
backend: add support for xpub_required in AOPP.

bznein:


bznein:
> The frontend should tell the user that "{{host}}" is requesting the xpub and that it will be shared. Not sure if this should need a checkbox/button confirmation. I think a clearly visible bold note in a suitable place in the workflow should be sufficient.

Would something like this work? 

image

benma:
> > The frontend should tell the user that "{{host}}" is requesting the xpub and that it will be shared. Not sure if this should need a checkbox/button confirmation. I think a clearly visible bold note in a suitable place in the workflow should be sufficient.
> 
> Would something like this work?
> 
> image

I'd make it an info instead of a warning if possible. 

@jadzeidan please suggest some text that makes it clear that the service is requesting the xpub and that it will be shared in the end.

frontends/web/src/components/aopp/aopp.tsx
```diff
@@ -110,10 +111,18 @@ export const Aopp = () => {
...
-                host: `<strong>${domain(aopp.callback)}</strong>`
+                host: `<strong>${host}</strong>`
               })} />
             )}
             withLogoText={t('aopp.addressRequestWithLogo')} />
+          {
+            aopp.xpubRequired ?
+              (
+                <div>
+                  <Message type="info"> {aopp.xpubRequired ? t('aopp.xpubRequested', { host: `${host}` }) : ''} </Message>
```
benma, line 122: aopp.xpubRequired here is redundant with the same check above

frontends/web/src/locales/en/app.json
```diff
@@ -84,7 +84,8 @@
       "title": "Address successfully sent"
     },
     "syncing": "Syncing the account, please wait.",
-    "title": "Address request"
+    "title": "Address request",
+    "xpubRequested": "Your extendend public key will be shared with {{host}}"
```
benma, line 88: @jadzeidan is this message good or do you want to suggest a change?

bznein:
This is the revised info box:

image

benma (changes requested):
changelog

jadzeidan:
> @bznein This is the revised info box:

Nit: Please left align the text in the info box 🙂

frontends/web/src/components/message/message.module.css
```diff
@@ -10,6 +10,7 @@
     margin-bottom: var(--spacing-default);
     overflow: hidden;
     padding: var(--spacing-default);
+    text-align: left;
```
benma, line 13: Does this now apply to all messages? Is this okay @thisconnect? This was added in response to https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3158/#issuecomment-2647999195

thisconnect: ack to set left align on all message components, until we have proper right-to-left langauge support (one day).

frontends/web/src/locales/en/app.json
```diff
@@ -72,7 +72,9 @@
   },
   "aopp": {
     "addressRequest": "{{host}} is requesting a receiving address.",
-    "addressRequestWithLogo": "is requesting a receiving address",
+    "addressRequestWithLogo": "is requesting a receiving address.",
+    "addressRequestWithLogoAndXPub": "is requesting a receiving address and public key (xPub).",
+    "addressRequestWithXPub": "{{host}} is requesting a receiving address and public key (xPub).",
```
thisconnect, line 77: Please lowercase "xpub" so it is consistent with existing messages / in-app guide etc.

PR #3159, opened 2025-02-06
frontend: fix missing base currency in btc-direct widget

thisconnect:
Instead of deriving the coin/token unit from coinCode somehow,
the app can just pass account.coinUnit.

Fixes missing baseCurrency error when using ERC20 tokens, with the
drawback that in testnet the app now passes TBTC. BTC Direct does
not support testnet addresses nor testnet coins and will fallback
to just BTC.


bznein (approved):
utACK

PR #3160, opened 2025-02-08
backend: allow opening checkout.btcdirect.eu

thisconnect:
Some payment providers such as iDEAL or SOFORT need to breakout the BitBoxApp and be continued in a native browser.

From https://checkout.btcdirect.eu/ the user fills in the bank they are using and will get redirected to their banking environment.

On mobile the bank may use intent links to automatically open and continue in the banking app.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -275,6 +275,7 @@ public class MainActivity extends AppCompatActivity {
                     patterns.add(Pattern.compile("^(.*\\.)?pocketbitcoin\\.com$"));
                     patterns.add(Pattern.compile("^(.*\\.)?moonpay\\.com$"));
                     patterns.add(Pattern.compile("^(.*\\.)?bitsurance\\.eu$"));
+                    patterns.add(Pattern.compile("^(.*\\.)?btcdirect\\.eu$"));
 
                     for (Pattern pattern : patterns) {
                         if (pattern.matcher(request.getUrl().getHost()).matches()) {


backend/backend.go
```diff
@@ -88,6 +88,7 @@ var fixedURLWhitelist = []string{
 	"https://pocketbitcoin.com/",
 	// BTCDirect
 	"https://btcdirect.eu/",
+	"https://checkout.btcdirect.eu/",
```
thisconnect, line 91: adding here is not strictly needed unless we add an in-app link to this url.

benma: Maybe remove it then?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -275,6 +275,7 @@ public boolean shouldOverrideUrlLoading(WebView view,  WebResourceRequest reques
                     patterns.add(Pattern.compile("^(.*\\.)?pocketbitcoin\\.com$"));
                     patterns.add(Pattern.compile("^(.*\\.)?moonpay\\.com$"));
                     patterns.add(Pattern.compile("^(.*\\.)?bitsurance\\.eu$"));
+                    patterns.add(Pattern.compile("^(.*\\.)?btcdirec\\.eu$"));
```
benma, line 278: `btcdirec` is missing a `t`?

benma (changes requested):

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
thisconnect: 🙈

backend/backend.go
thisconnect: The only reason I'd keep it here is so we know this is an url we allow.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
thisconnect: 🙏 thanks for catching, fixed.

PR #3166, opened 2025-02-10
ios: listen on our product characteristic

benma:
- Similar to the USB HID descriptor product string
- Empty `""` means the device has not set it yet
- "no connection" is defined by the bluetooth firmware as not connected to either firmware/bootloader yet
- The bluetooth firmware updates it dynamically, as the device can reboot into bootloader/firwmware without bluetooth disconnecting. The app listens on the change and connects/disconnects the right device accordingly.
- For now we map to "BitBox02BTC" to get a connection going, later we will modify bitbox02-api-go to extend it with the official new product strings so the backend can identify them


NickeZ (approved):
lgtm

PR #3167, opened 2025-02-10
frontend: fix export to CSV for ERC20 Tokens.

bznein:
ERC20 tokens are now exported with the correct value for "unit" Added a new column "FeeUnit" as the fee is displayed in ETH/wei for ERC20 tokens.

### DIFF
--- a/backend/accounts/baseaccount.go
+++ b/backend/accounts/baseaccount.go
@@ -245,6 +245,7 @@ func (account *BaseAccount) ExportCSV(w io.Writer, transactions []*TransactionDa
 		"Amount",
 		"Unit",
 		"Fee",
+		"Fee Unit",
 		"Address",
 		"Transaction ID",
 		"Note",
@@ -260,11 +261,17 @@ func (account *BaseAccount) ExportCSV(w io.Writer, transactions []*TransactionDa
 			TxTypeSendSelf: "sent_to_yourself",
 		}[transaction.Type]
 		feeString := ""
+		feeUnit := ""
 		fee := transaction.Fee
 		if fee != nil {
 			feeString = fee.BigInt().String()
+			feeUnit = account.Coin().SmallestUnit()
 		}
 		unit := account.Coin().SmallestUnit()
+		if transaction.IsErc20 {
+			unit = account.Coin().Unit(false)
+		}
+
 		timeString := ""
 		if transaction.Timestamp != nil {
 			timeString = transaction.Timestamp.Format(time.RFC3339)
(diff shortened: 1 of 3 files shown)


backend/accounts/baseaccount.go
```diff
@@ -245,6 +246,7 @@ func (account *BaseAccount) ExportCSV(w io.Writer, transactions []*TransactionDa
 		"Amount",
 		"Unit",
 		"Fee",
+		"Fee Unit",
```
benma, line 249: Might adding a column break Cointracking? @Beerosagos

backend/accounts/baseaccount.go
```diff
@@ -273,12 +281,22 @@ func (account *BaseAccount) ExportCSV(w io.Writer, transactions []*TransactionDa
...
 				transactionType = "sent_to_yourself"
 			}
+
+			amount := addressAndAmount.Amount.BigInt()
+			amountExport := amount.String()
+
+			// When dealing with ERC20 tokens, we need to divide the amount by the token's decimals.
+			if isERC20 {
+				erc20amount := new(big.Float).Quo(new(big.Float).SetInt(amount), new(big.Float).SetInt(coin.DecimalsExp(account.Coin())))
+				amountExport = erc20amount.String()
```
benma, line 291: Couldn't you simply use `Coin().FormatAmount()` here?

backend/accounts/baseaccount.go
```diff
@@ -253,6 +255,7 @@ func (account *BaseAccount) ExportCSV(w io.Writer, transactions []*TransactionDa
 		return errp.WithStack(err)
 	}
 
+	isERC20 := account.Coin().IsERC20()
```
benma, line 258: I think instead if adding a interface function for all coins, the info is already in `transaction.IsERC20`.

Alternatively you could match the specific coin: `ethCoin, ok := account.Coin().(*eth.Coin); isERC20 := ok && ethCoin.ERC20Token() != nil`

bznein: Yeah this is just a draft pr and it's actually broken right now (doesn't build due to import cycle), so it's going to change

bznein: > Alternatively you could match the specific coin: `ethCoin, ok := account.Coin().(*eth.Coin); isERC20 := ok && ethCoin.ERC20Token() != nil`

Fun fact, before I noticed that we have `IsERC20` in the tx data, I was doing this, but it's not possible as it creates an import cycle (we can't import `eth` here as `eth` itself imports `accounts`) that's the reason for that workaround of the interface :)

backend/accounts/baseaccount.go
Beerosagos: I asked them

backend/accounts/baseaccount.go
```diff
@@ -289,6 +304,7 @@ func (account *BaseAccount) ExportCSV(w io.Writer, transactions []*TransactionDa
 			// a multitx is output in one row per receive address. Show the tx fee only in the
 			// first row.
 			feeString = ""
+			feeUnit = ""
```
benma, line 307: In receive transactions in Bitcoin, the feeString is `""`, but the unit "satoshi" is shown. But here, if the fee is `""`, the fee unit is not shown. Would be good to be consistent, i.e. show the feeUnit always, or never show it if the feeString is `""`.

benma (changes requested):
tested LGTM, left a minor nit.

backend/accounts/baseaccount.go
bznein: I went with the second: never showing it if the `feeString` is empty. LMK if this looks ok to you

bznein: And I added some unit tests to make sure :P

backend/accounts/baseaccount.go
Beerosagos: They replied it's fine :+1:

PR #3168, opened 2025-02-11
frontend: fix account selector title on buy / sell page

shonsirsha:
Rename 'Buy & Sell' title to 'Select account'
for clarity in the account selector of Buy & Sell page.

The same behaviour can be seen for the Receive account selector.


thisconnect (approved):
LGTM

PR #3171, opened 2025-02-11
Update btcdirect payment methods with fees

thisconnect:


frontends/web/src/routes/exchange/components/infocontent.tsx
```diff
@@ -139,14 +176,71 @@ export const BTCDirectInfo = ({ accounts }: TBTCDirectInfo) => {
...
+  cardFee,
+  bankTransferFee,
+  sofortFee,
+  bancontactFee,
+}: TBTCDirectInfoProps) => {
+  const { t } = useTranslation();
+  const formattedCardFee = cardFee && cardFee * 100;
+  const formattedBankTransferFee = bankTransferFee && bankTransferFee * 100;
+  const formattedSofortFee = sofortFee && sofortFee * 100;
+  const formattedBancontactFee = bancontactFee && bancontactFee * 100;
```
thisconnect, line 204: this is not ideal, with 0.036 fee I get 3.5999999999999996%

probably better we do such math stuff in the backend 😇

Beerosagos: I think that we could store the fees in the backend deals as 3.6 instead of 0.036 so that we don't have to convert in the frontend. Would that be good enough?

thisconnect: yes that makes sense. Looks to me that we do not use fee other than isBest calculation here https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/bbdc3b87016b33dc2b6ac46eb4a5cc0d65d2c1eb/backend/exchanges/exchanges.go#L214-L223

 change in last commit

Beerosagos: Awesome, ty

thisconnect: changed to percentages in last commit.

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -45,21 +44,6 @@ type TProps = {
...
-  case 'nl':
-    return 'https://btcdirect.eu/nl-nl/private-trading-contact?BitBox';
-  case 'es':
-    return 'https://btcdirect.eu/es-es/private-trading-contactanos?BitBox';
-  case 'fr':
-    return 'https://btcdirect.eu/fr-fr/private-trading-contact?BitBox';
-  default:
-    return 'https://btcdirect.eu/en-eu/private-trading-contact?BitBox';
-  }
-};
```
thisconnect, line 61: moved to infocontent component so that we have less circular imports.

thisconnect:
@Beerosagos PTAL

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -45,21 +44,6 @@ type TProps = {
   setInfo: (into: TInfoContentProps) => void;
```
Beerosagos, line 44: typo: into

frontends/web/src/routes/exchange/components/exchange-providers.tsx
```diff
@@ -47,6 +47,10 @@ const PaymentMethod = ({ methodName }: TPaymentMethodProps) => {
         {t('buy.exchange.creditCard')}
       </span>
     );
+  case 'sofort':
+  case 'bancontact':
+    // hide these payment methods in the overview
+    return '';
```
Beerosagos, line 53: This is not really robust, imo. If the only available deals in a specific situation were the hidden ones, this would show a component without available options. It would be better to have a "hidden" field in the deal struct that can be set in the backend and properly checked in the frontend.

backend/exchanges/btcdirect.go
```diff
@@ -80,14 +80,22 @@ func BtcDirectDeals() *ExchangeDealsList {
 		ExchangeName: BTCDirectName,
 		Deals: []*ExchangeDeal{
 			{
-				Fee:     3,
+				Fee:     3, // 3%
 				Payment: CardPayment,
-				IsFast:  true,
+				IsLocal: true,
```
Beerosagos, line 85: I think that we should keep the `isFast` too

Beerosagos (commented):
Nice! I'm not sure about hiding deals in the frontend, tho. See my comment below :innocent:

backend/exchanges/btcdirect.go
thisconnect: If we do that with the current system it would show up, but I think we don't want to have multiple isfast badges.

Screenshot 2025-02-12 at 13 30 01

frontends/web/src/routes/exchange/components/buysell.tsx
thisconnect: done, fixedup into this commit

frontends/web/src/routes/exchange/components/exchange-providers.tsx
thisconnect: I don't know sofort or bancontact (not available in CH), aren't these some kind  of fiat banktransfers anyway?

Hidden field sounds more robust, but I am not 100% confident about how to do the backend changes... I'll dm you later.

thisconnect: as discussed I'll leave in ths PR, and we can add isHidden in a new PR

backend/exchanges/exchanges.go
```diff
@@ -101,17 +101,22 @@ const (
...
-// - Fee indicates form the percentage that goes to the exchange in a float representation (e.g. 0.01 -> 1%).
+// - Fee that goes to the exchange in percentage.
 // - Payment is the payment method offered in the deal (usually different payment methods bring different fees).
 // - IsFast is usually associated with card payments. It is used by the frontend to display the `fast` tag in deals list.
 type ExchangeDeal struct {
 	Fee     float32       `json:"fee"`
 	Payment PaymentMethod `json:"payment"`
 	IsFast  bool          `json:"isFast"`
 	IsBest  bool          `json:"isBest"`
+	IsLocal bool          `json:"isLocal"`
```
thisconnect, line 119: as discussed in dm, we keep the badges we have right now, but skip "Local payments" for now and rethink what badges it should show.

thisconnect: done, @Beerosagos PTAL

backend/exchanges/btcdirect.go
```diff
@@ -80,14 +80,21 @@ func BtcDirectDeals() *ExchangeDealsList {
 		ExchangeName: BTCDirectName,
 		Deals: []*ExchangeDeal{
 			{
-				Fee:     3,
+				Fee:     3, // 3%
 				Payment: CardPayment,
-				IsFast:  true,
```
Beerosagos, line 85: I think we should keep the isFast here like for the other card payments (e.g. moonpay)

Beerosagos (commented):
Left a last comment :innocent:

frontends/web/src/routes/exchange/components/buysell.tsx
```diff
@@ -109,9 +93,17 @@ export const BuySell = ({
   };
 
   const buildInfo = (exchange: exchangesAPI.ExchangeDeals): TInfoContentProps => {
-    const cardFee = exchange.deals && exchange.deals.find(feeDetail => feeDetail.payment === 'card')?.fee;
-    const bankTransferFee = exchange.deals && exchange.deals.find(feeDetail => feeDetail.payment === 'bank-transfer')?.fee;
-    return { info: exchange.exchangeName, cardFee, bankTransferFee };
+    const cardFee = exchange.deals.find(feeDetail => feeDetail.payment === 'card')?.fee;
+    const bankTransferFee = exchange.deals.find(feeDetail => feeDetail.payment === 'bank-transfer')?.fee;
+    const bancontactFee = exchange.deals.find(feeDetail => feeDetail.payment === 'bancontact')?.fee;
+    const sofortFee = exchange.deals.find(feeDetail => feeDetail.payment === 'sofort')?.fee;
```
Beerosagos, line 99: Another one, sorry: I think that this could scale better by refactoring the payment methods fees as a map.

Beerosagos: Not a blocker, but I think it could simplify the code a bit

thisconnect: yeah this function just got a bit more ugly :)

backend/exchanges/btcdirect.go
thisconnect: ack. fixedup and force pushed.

thisconnect: fixed, @Beerosagos PTAL should be all good now to merge this

Beerosagos (approved):
tACK :+1:

PR #3172, opened 2025-02-11
frontend: show correct amount for ERC20 tokens.

bznein:
Default to show the amount of the tx for all transactions, and then add the fee for non-erc20 ones.


benma (approved):
tACK thanks!

PR #3176, opened 2025-02-12

benma:
```
go get github.com/BitBoxSwiss/bitbox02-api-go@2b90fad
go mod tidy
go mod vendor
```

### DIFF
--- a/backend/devices/bitbox02/handlers/handlers.go
+++ b/backend/devices/bitbox02/handlers/handlers.go
@@ -289,10 +289,15 @@ func (handlers *Handlers) postSetMnemonicPassphraseEnabled(r *http.Request) inte
 func (handlers *Handlers) getVersionHandler(_ *http.Request) interface{} {
 	currentVersion := handlers.device.Version()
 	newVersion := bitbox02bootloader.BundledFirmwareVersion(handlers.device.Product())
-
+	var newVersionStr string
+	var canUpgrade bool
+	if newVersion != nil {
+		newVersionStr = newVersion.String()
+		canUpgrade = newVersion.AtLeast(currentVersion) && currentVersion.String() != newVersion.String()
+	}
 	return struct {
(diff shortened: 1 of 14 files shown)


bznein (approved):
LGTM

benma:
@bznein pushed one more commit, pls check again :innocent:

backend/devices/bitbox02/handlers/handlers.go
```diff
@@ -289,10 +289,17 @@ func (handlers *Handlers) postSetMnemonicPassphraseEnabled(r *http.Request) inte
...
 	newVersion := bitbox02bootloader.BundledFirmwareVersion(handlers.device.Product())
-
+	var newVersionStr string
+	if newVersion != nil {
+		newVersionStr = newVersion.String()
+	}
+	canUpgrade := false
+	if newVersion != nil {
+		canUpgrade = newVersion.AtLeast(currentVersion) && currentVersion.String() != newVersion.String()
+	}
```
bznein, line 299: Can we just move `canUpgrade := false` (nit: I think for consistency `var canUpgrade bool` would be nicer) above and have everything in a single if?

```go
var newVersionStr string
var canUpgrade bool

if newVersion != nil {
                neeVersionStr = newVersion.String()
		canUpgrade = newVersion.AtLeast(currentVersion) && currentVersion.String() != newVersionStr
	}
```

(also: wdyt about changing the implementation of `String()` in `bitbox02-api-go` to add 
```go
if version == nil { return "" }
```
This way we wouldn't need a `newVersionStr`. But this might be out of scope for this PR :)

benma: Done the first part. I don't think changing `String()` is a good idea, as I think normally methods, even on pointer receivers, assume the instance is not nil, so returning `""` there is quite implicit. Explicit > impllicit. Looking at the call site one in that case would look like it would panic.

PR #3177, opened 2025-02-13
Frontend btcdirect disclaimertext

thisconnect:


frontends/web/src/locales/en/app.json
```diff
@@ -462,7 +482,6 @@
               "title": "Fees for Buying"
             },
             "intro": "European buying of cryptocurrencies with EUR.",
-            "learnmore": "Learn more about BTC Direct",
```
thisconnect, line 465: just fyi: this is not yet in locize.

moved "learnmore" key one level up so as it is the same text for "learn more" regardless where the link is.

but later I realized that there are a few small inconsitencies how all this  guide text is structured.

frontends/web/src/components/terms/btcdirect-otc-terms.tsx
```diff
@@ -26,26 +26,27 @@ type TProps = {
...
+  case 'nl':
+    return 'https://btcdirect.eu/nl-nl/privacy-policy?BitBox';
+  case 'es':
+    return 'https://btcdirect.eu/es-es/privacy-policy?BitBox';
+  case 'fr':
+    return 'https://btcdirect.eu/fr-fr/privacy-policy?BitBox';
+  default:
+    return 'https://btcdirect.eu/en-eu/privacy-policy?BitBox';
+  }
+};
```
thisconnect, line 42: The other links come from routes/exchange/components/infocontent.tsx

but it seems completely wrong to import from routes here.

I considered moving terms into routes/exchanges, but one of the terms is also used for routes/bitsurance, so I just left it as is now.

Beerosagos (approved):
tACK :+1:

PR #3178, opened 2025-02-13
backend: add isHidden property in exchange deals

Beerosagos:
This allows to hide specific deals in the frontend deals list. e.g. when the exchange offers several payment methods with different fees. The deals are still used to show all the payments and fees in the info box.

### DIFF
--- a/backend/exchanges/exchanges.go
+++ b/backend/exchanges/exchanges.go
@@ -108,14 +108,17 @@ const (
 )
 
 // ExchangeDeal represents a specific purchase option of an exchange.
-// - Fee that goes to the exchange in percentage.
-// - Payment is the payment method offered in the deal (usually different payment methods bring different fees).
-// - IsFast is usually associated with card payments. It is used by the frontend to display the `fast` tag in deals list.
 type ExchangeDeal struct {
-	Fee     float32       `json:"fee"`
+	// Fee that goes to the exchange in percentage.
+	Fee float32 `json:"fee"`
+	// Payment is the payment method offered in the deal (usually different payment methods bring different fees).
 	Payment PaymentMethod `json:"payment"`
-	IsFast  bool          `json:"isFast"`
-	IsBest  bool          `json:"isBest"`
+	// IsFast is usually associated with card payments. It is used by the frontend to display the `fast` badge in deals list.
+	IsFast bool `json:"isFast"`
+	// IsBest is assigned to the deal with the lowest fee, it is used to show the `best deal` badge in the frontend.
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/exchange/components/exchange-providers.tsx
```diff
@@ -32,28 +32,27 @@ type TPaymentMethodProps = { methodName: ExchangeDeal['payment'] };
...
-  case 'bancontact':
-    // hide these payment methods in the overview
-    return '';
-  default:
-    return <>{methodName}</>;
+    icon = isDarkMode ? <CreditCard /> : <CreditCardDark />;
+    label = 'buy.exchange.creditCard';
+    break;
+  // case 'sofort':
+  // case 'bancontact':
```
Beerosagos, line 47: Not sure if we want to add icons for sofort and bancontact :thinking:

thisconnect (changes requested):

frontends/web/src/routes/exchange/components/exchange-providers.tsx
```diff
@@ -32,28 +32,27 @@ type TPaymentMethodProps = { methodName: ExchangeDeal['payment'] };
...
+  let icon = <>{methodName}</>;
+  let label = '';
   switch (methodName) {
   case 'bank-transfer':
-    return (
-      <span className={style.paymentMethodName}>
-        {isDarkMode ? <Bank /> : <BankDark />}
-        {t('buy.exchange.bankTransfer')}
-      </span>
-    );
```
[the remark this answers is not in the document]
Beerosagos: Ack, I just thought that with other payment methods, it would duplicate too much code, but since we are not adding any other method that's absolutely fine.

Beerosagos:
@thisconnect updated, PTAL :pray:

backend/exchanges/exchanges.go
```diff
@@ -236,7 +239,7 @@ func GetExchangeDeals(account accounts.Interface, regionCode string, action Exch
 		bestDealIndex := 0
 		for i, deal := range deals {
 			oldBestDeal := deals[bestDealIndex]
-			if deal.Fee < oldBestDeal.Fee {
+			if !deal.IsHidden && deal.Fee < oldBestDeal.Fee {
```
bznein, line 242: [out of scope for this PR] I still think that if we keep the "best" badge we need to decide how to deal with multiple deals offering the same fee

bznein (approved):
LGTM (mostly looked at the backend, but the frontend looks good to me as well)

PR #3180, opened 2025-02-13
Exchanges cleanup

Beerosagos:
Some refactoring and cleanups in the code, see commit messages

### DIFF
--- a/frontends/web/src/api/exchanges.ts
+++ b/frontends/web/src/api/exchanges.ts
@@ -17,20 +17,8 @@
 import { AccountCode } from './account';
 import { apiGet, apiPost } from '@/utils/request';
 
-export type ExchangeRegionList = {
-  regions: ExchangeRegion[];
-}
-
-export type ExchangeRegion = {
-  code: string;
-  isMoonpayEnabled: boolean;
-  isPocketEnabled: boolean;
-}
-
-export const getExchangesByRegion = (code: string) => {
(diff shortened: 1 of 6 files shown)


frontends/web/src/api/exchanges.ts
```diff
@@ -17,19 +17,9 @@
...
-  code: string;
-  isMoonpayEnabled: boolean;
-  isPocketEnabled: boolean;
-}
-
-export const getExchangesByRegion = (code: string) => {
-  return (): Promise<ExchangeRegionList> => {
-    return apiGet(`exchange/by-region/${code}`);
+export const getExchangeRegionCodes = () => {
+  return (): Promise<string[]> => {
```
thisconnect, line 21: Nice that makes everything simpler.

Now that the `code` argument is gone, it doesn't need to return another function.

You could just return `return apiGet('exchange/region-codes');` here and pass it later to useLoad without calling the function:

`const regionCodes = useLoad(exchangesAPI.getExchangeRegionCodes);`

thisconnect (commented):
frontend LGTM with small suggestion

bznein (approved):
LGTM on the backend. The frontend goes a bit aboe my expertise :)

thisconnect (changes requested):

thisconnect (approved):
tested LGTM with optional suggestion

PR #3182, opened 2025-02-16
frontend: make locize-pull

benma:


frontends/web/src/locales/en/app.json
```diff
@@ -68,13 +68,11 @@
       "step": "Finished",
       "title": "Account added"
     },
-    "title": "Add account"
+    "title": "+ Add account"
```
benma, line 71: this seems wrong, is this correct @thisconnect? does not seem like a `+` should be part of i18n/translations.

thisconnect: Yeah this is new and was added on purpose, it needs to be part of the translation as  right-to-left langs should put it on the other side.

frontends/web/src/locales/de/app.json
```diff
@@ -760,7 +760,7 @@
         "kyc": {
           "link": "Lese die Pocket FAQs",
           "p1": "Für den Handel bis zu 950 EUR (1000 CHF) pro Tag sind keine weiteren Dokumente erforderlich.",
```
thisconnect, line 762: I believe this is now wrong.

frontends/web/src/locales/en/app.json
```diff
@@ -68,13 +68,11 @@
...
   },
   "aopp": {
     "addressRequest": "{{host}} is requesting a receiving address.",
-    "addressRequestWithLogo": "is requesting a receiving address.",
-    "addressRequestWithLogoAndXPub": "is requesting a receiving address and public key (xpub).",
-    "addressRequestWithXPub": "{{host}} is requesting a receiving address and public key (xpub).",
```
thisconnect, line 77: These keys are new and should not be removed. Did you push?

https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/5220df6948aff065fb7bc1f75f4c4c57e45c37d5#diff-6631e5e7b36d6349336cdf14e4f91eeeb8f284d449b8281af375936bbe94ae79

frontends/web/src/locales/en/app.json
```diff
@@ -86,8 +84,7 @@
       "title": "Address successfully sent"
     },
     "syncing": "Syncing the account, please wait.",
-    "title": "Address request",
-    "xpubRequested": "Sharing your xpub lets external services see your account addresses. Your coins are still safe and remain in your control."
```
thisconnect, line 90: We should have this as well.

thisconnect (changes requested):
@benma did you push and pull?

frontends/web/src/locales/de/app.json
thisconnect: We should update the English version and update the daily kyc-light faq.

frontends/web/src/locales/en/app.json
benma: I did, but after pull :see_no_evil: fixed now

frontends/web/src/locales/de/app.json
thisconnect: updated in locize

thisconnect (approved):
untested LGTM

note: that `exchange.pocket.terms.kyc.p1 ` in English has changed in locize, in case you like to include this as well.

frontends/web/src/locales/de/app.json
benma: pulled

frontends/web/src/locales/en/app.json
jadzeidan: @thisconnect It's meant to be an SVG, will text "+" look good? :)

thisconnect: @jadzeidan I wasn't aware of that. It looks as good as in your mockups.

PR #3188, opened 2025-02-17
ios: list bluetooth devices, let user tap to connect

benma:
Instead of connecting to the first discovered device, we list all discovered BitBox02s and let the user connect to one.

This commit does not yet properly handle:
- listing the devices ordered by discovery time - newest should come on bottom so the list stays stable when new devices are added, so the user does not accidentally tap the wrong BitBox (wants to tap one, but a newly discovered one is put there right before they tap)


NickeZ (commented):
tACK

NickeZ (approved):

PR #3189, opened 2025-02-18
backend: fix restart in testnet

benma:
Typo in config var, and the flag was never reset.


bznein (approved):
Tested ACK thank you so much!

PR #3190, opened 2025-02-19
frontend: fix unsupported css property

shonsirsha:
Bug exists across platforms due to an unsupported CSS property in the browser engine: `justify-content: end;`.

Our QT is running version 90.x, which doesn’t support this property. Support for justify-content: end; begins in Chrome 93.x.

<hr />

- [ ] updating the Changelog
- [ ] writing unit tests


thisconnect (approved):
thanks LGTM

PR #3191, opened 2025-02-19
bb01: don't load keystore/account

benma:
BitBox01 has been end-of-life'd a long time ago. Since we want to refactor transaction signing and simplify accounts management, BitBox01 is once again a maintenance burden. For this reason we remove accounts/keystore support for BitBox01. In follow-up commits, we can remove its keystore, remove split accounts, etc.

We keep device management in the app for now so users can check and create backups, etc.

Users can continue using BitBoxApp v4.47.0 to send out coins if needed.


frontends/web/src/routes/device/waiting.tsx
```diff
@@ -31,9 +34,18 @@ import style from './bitbox01/bitbox01.module.css';
 export const Waiting = () => {
   const { t } = useTranslation();
   const { isDarkMode } = useDarkmode();
-
+  const navigate = useNavigate();
   const keystores = useKeystores();
-  const devices = useSync(getDeviceList, syncDeviceList);
+  const devices = useDefault(useSync(getDeviceList, syncDeviceList), {});
+
+  // BitBox01 does not have any accounts anymore, so we route directly to the device settings.
```
thisconnect, line 41: small note: If a user has "remembered wallets" the app wont load waiting route. Unlocking a BB01 in this case will still go to account-summary. But I don't think it is worth catching this.

thisconnect (approved):
tested LGTM

PR #3193, opened 2025-02-19
backend: allow firmware upgrade when app is offline.

bznein:
Run ensureAddresses in a separate goroutine.
This way the call is not blocking and the initializeLock can be released, which allows the device to be restarted in bootloader mode and the firmware upgrade process to start.

This way, `ensureAddresses` will keep running in the background and if connection is restored later, it will terminate correctly.

This has been tested on MacOS with Qt: note that if running this in webdev mode it won't work, as when the app is offline there is no way to reach the "manage device" option as it will hang forever.

The following in a screen recording that shows the correct behaviour.

https://github.com/user-attachments/assets/3818a8b1-2730-4e6e-a14f-b00632f720da

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/btc/account_test.go
+++ b/backend/coins/btc/account_test.go
@@ -21,6 +21,7 @@ import (
 	"math/big"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	accountsTypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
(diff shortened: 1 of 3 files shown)


backend/coins/btc/account_test.go
```diff
@@ -110,7 +111,7 @@ func TestAccount(t *testing.T) {
 	account := mockAccount(t, nil)
 	require.False(t, account.Synced())
 	require.NoError(t, account.Initialize())
-	require.True(t, account.Synced())
+	require.Eventually(t, func() bool { return account.Synced() }, time.Minute, time.Second)
```
benma, line 114: TIL Eventually :exploding_head: 

Do you know how long it takes? Maybe a shorter interval could make sense.

benma (approved):
LGTM

backend/coins/btc/account_test.go
bznein: I think it barely needs to wait, I can definitely try to lower the value and see what happens :)

PR #3201, opened 2025-02-27
test: ensure that account is eventually synced.

bznein:
With ensureAddresses running in a separate goroutine, it might happen that after a call to account.Initialize() we perform checks that fail because the ensureAddresses goroutine hasn't completed yet.

By adding a call to require.Eventually we ensure that the account gets eventually synced (within a second) before proceeding with the rest of the test.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/btc/account_test.go
+++ b/backend/coins/btc/account_test.go
@@ -111,7 +111,7 @@ func TestAccount(t *testing.T) {
 	account := mockAccount(t, nil)
 	require.False(t, account.Synced())
 	require.NoError(t, account.Initialize())
-	require.Eventually(t, func() bool { return account.Synced() }, time.Second, time.Millisecond*200)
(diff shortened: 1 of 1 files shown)


backend/coins/btc/account_test.go
```diff
@@ -189,6 +191,7 @@ func TestInsuredAccountAddresses(t *testing.T) {
 func TestSignAddress(t *testing.T) {
 	account := mockAccount(t, nil)
 	require.NoError(t, account.Initialize())
+	require.Eventually(t, func() bool { return account.Synced() }, time.Second, time.Millisecond*200)
```
benma, line 194: nit: isn't `func() bool { return account.Synced() }` the same as `account.Synced`?

benma (approved):
utACK

PR #3206, opened 2025-03-03
frontend: move contact us position in guide

shonsirsha:
fsd


thisconnect:
found just a tiny edgecase nit: could you add some padding top to the appendix. In case when there is many guide entries or small screen height:

main version:

Screenshot 2025-03-03 at 09 28 25

with this change:

Screenshot 2025-03-03 at 09 28 40

thisconnect (approved):
tested LGTM thank you 👍

PR #3212, opened 2025-03-04
qt: fix uri scheme handling on macOS

benma:


bznein:
Tested and it does work in the app is already running, but if the app is not running, it will open it but not show the AOPP screen :(

benma:
> Tested and it does work in the app is already running, but if the app is not running, it will open it but not show the AOPP screen :(

Are you sure that in this case, not another copy of your BitBoxApp in your system (with the bug present) is being opened?

bznein:
> > Tested and it does work in the app is already running, but if the app is not running, it will open it but not show the AOPP screen :(
> 
> Are you sure that in this case, not another copy of your BitBoxApp in your system (with the bug present) is being opened?

I'll triple check :)

bznein (approved):
As usual, you were absolutely right :)

Tested ACK! Thanks!

PR #3213, opened 2025-03-04
frontend fix transactions

thisconnect:
- fixes breaking unit (U S D) onto multiple newlines on small screen 
- unit should be gray and smaller regression
- hide plus minus sign if amount is not yet known
- fix long amounts on small screens

### DIFF
--- a/frontends/web/src/components/amount/amount.module.css
+++ b/frontends/web/src/components/amount/amount.module.css
@@ -1,3 +1,9 @@
+.amount {
+    flex-shrink: 1;
+    overflow: hidden;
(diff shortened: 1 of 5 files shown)


thisconnect:
Before (current release-v4.47.0)

Screenshot 2025-03-04 at 11 04 29

with this PR 

Screenshot 2025-03-04 at 11 03 59

thisconnect:
@shonsirsha PTAL please also try to test and break it.

thisconnect:
macos qt (from this build) looks ok to me.

PR #3215, opened 2025-03-04
frontend: fix shown amount for unconfirmed ETH txs

bznein:
Fixes the shown amount by allowing `formatJSONAmountAtTime` to work with a nil timestamp

image

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/components/amount/conversion-amount.tsx
+++ b/frontends/web/src/components/amount/conversion-amount.tsx
@@ -54,7 +54,7 @@ export const ConversionAmount = ({
 
   return (
     <span className={styles.txConversionAmount}>
-      {conversion && amountToShow ? (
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/amount/conversion-amount.tsx
```diff
@@ -54,7 +54,7 @@ export const ConversionAmount = ({
 
   return (
     <span className={styles.txConversionAmount}>
-      {conversion && amountToShow ? (
+      {(conversion || sendToSelf) && amountToShow ? (
```
bznein, line 57: If this is a `sendToSelf`, we do not need `conversion` as we show the coin amount, and not its conversion into a fiat value.

benma (approved):
tACK, thanks!

PR #3219, opened 2025-03-05
frontend/balance: add BTC/sat rotating unit in account balance

Beerosagos:
Previously it was possible to switch between BTC and sat unit using a switch component in the general settings. This change allows the user to switch unit more easily without leaving the account view.

This also heavily refactors the FiatConversion component, which is now renamed AmountWithUnit.

frontend/web/src/components/rates/rates.* have been moved and renamed to frontend/web/src/components/amount/amount-with-unit.*

Note: while our currency rotational unit is based on the fact that all the fiat conversions are fetched inside the amounts, this doesn't happen for btc/sat switching. This means that I had to "observe" the unit change in the account page to update the balance and the transaction list, i.e. it is an async change. It seems good enough to me (and I really don't want to over complicate things adding BTC/sat conversion in the amount), but let me know what you think about this.


frontends/web/src/components/amount/amount-with-unit.tsx
```diff
@@ -0,0 +1,103 @@
+/**
+ * Copyright 2018 Shift Devices AG
+ * Copyright 2020 Shift Crypto AG
```
thisconnect, line 3: 2020-2025

frontends/web/src/components/amount/amount.tsx
```diff
@@ -92,21 +83,27 @@ const formatBtc = (
...
+  onMobileClick?: () => Promise<void>;
+};
+
 export const Amount = ({
   amount,
   unit,
   removeBtcTrailingZeroes,
   alwaysShowAmounts = false,
-  allowRotateCurrencyOnMobile = false,
+  onMobileClick,
```
thisconnect, line 99: 👍

frontends/web/src/components/balance/balance.tsx
```diff
@@ -73,7 +70,7 @@ export const Balance = ({
               {' '}{balance.incoming.unit} /
               <span className={style.incomingConversion}>
                 {' '}
-                <FiatConversion amount={balance.incoming} noBtcZeroes/>
+                <AmountWithUnit amount={balance.incoming} removeBtcTrailingZeroes convertToFiat/>
```
thisconnect, line 73: Small regression when sending, I believe due to the `unstyled` prop that was there before.

before this commit:
Screenshot 2025-03-12 at 12 14 23

with this commit the BTC unit is very large:
Screenshot 2025-03-12 at 12 14 00

frontends/web/src/components/amount/amount.tsx
```diff
@@ -92,21 +83,27 @@ const formatBtc = (
   );
 };
 
+type TProps = {
+  amount: string;
+  unit: CoinUnit | ConversionUnit;
+  removeBtcTrailingZeroes?: boolean;
+  alwaysShowAmounts?: boolean
```
thisconnect, line 90:
```suggestion
  alwaysShowAmounts?: boolean;
```

thisconnect (changes requested):
very nice thank you @Beerosagos !

I found a small bug with incoming amount unit see comment.

Beerosagos:
Rebased ^

frontends/web/src/components/balance/balance.tsx
Beerosagos: Nice catch! Thanks :pray:

Beerosagos:
@thisconnect thanks for the review! I updated the PR, PTAL :pray:

thisconnect (approved):
partly tested LGTM with tiny nit

Beerosagos:
rebased ^

PR #3231, opened 2025-03-10
We should avoid setting GO env variables

NickeZ:
It should be enough to temporarily set GOPATH during docker image build to install tools in a shared directory and then add that directory to $PATH.

The defaults on unix are:
GOPATH = ~/go
GOROOT = /usr/local/go
GOMODCACHE = GOPATH[0]/pkg/mod

I will have to test this myself first when I have better internet and can use docker.


Makefile
```diff
@@ -14,8 +14,6 @@
 
 SHELL    := /bin/bash
 WEBROOT  := frontends/web
-GOPATH   ?= $(HOME)/go
```
NickeZ, line 17: GOPATH can be any directory on your system. In Unix examples, we will set it to $HOME/go (the default since Go 1.8).

NickeZ:
Ready for review, need help publishing container tag `:29` to ghcr.io.

scripts/github-ci.sh
```diff
@@ -22,24 +22,18 @@ if [ "$OS_NAME" == "linux" ]; then
...
-    # GitHub checkout action (git clone) seem to require current work dir
-    # to be the root of the repo during its clean up phase. So, we push it
-    # here and pop in the end.
-    pushd ../ && cp -a bitbox-wallet-app $GOPATH/$(dirname $GO_SRC_DIR)
-    cd $GOPATH/$GO_SRC_DIR
-    make "$WHAT"
-    popd
+    export PATH="~/go/bin:$PATH"
+    mkdir -p $(go env GOPATH)/$(dirname $GO_SRC_DIR)
+    cp -a ../bitbox-wallet-app $(go env GOPATH)/$(dirname $GO_SRC_DIR)
```
benma, line 37: this seems wrong, there should be no need to copy the repo somewhere else.

NickeZ: Is that something from how go used to work? that the go code had to be copied to a certain directory structure?

NickeZ: There is this comment earlier:

> # Because we need to compile some Go code without modules,
> # the source must be placed in a specific directory as expected by Go.
> # The path is relative to GOPATH.
> GO_SRC_DIR=src/github.com/BitBoxSwiss/bitbox-wallet-app

benma (approved):
utACK

PR #3234, opened 2025-03-11
Shutdown

benma:
First commit is from https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3230

### DIFF
--- a/backend/bridgecommon/bridgecommon.go
+++ b/backend/bridgecommon/bridgecommon.go
@@ -363,8 +363,9 @@ func Shutdown() {
 
 	log := logging.Get().WithGroup("server")
 	if globalShutdown != nil {
+		log.Info("Shutdown about to be called")
 		globalShutdown()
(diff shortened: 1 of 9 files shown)


backend/bridgecommon/bridgecommon.go
```diff
@@ -363,8 +363,9 @@ func Shutdown() {
 
 	log := logging.Get().WithGroup("server")
 	if globalShutdown != nil {
+		log.Info("Shutdown about to be called")
```
Beerosagos, line 366: nit: is this still needed, or is it a debug leftover?

Beerosagos (approved):
tACK :+1:

backend/bridgecommon/bridgecommon.go
benma: I thought it's a helpful line anyway in case Shutdown gets stuck for whatever reason

PR #3235, opened 2025-03-12
backend: show estimated conversions for unconfirmed ETH txs and remove skeleton when conversions are missing.

bznein:
ETH transactions do not have a "Created" timestamp we can't provide an estimated conversion based on that. We use real-time conversion instead. 

We also remove the skeleton for the cases in which a conversion is still not available: 
image

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -121,15 +121,27 @@ func (handlers *Handlers) formatAmountAsJSON(amount coin.Amount, isFee bool) For
 
 func (handlers *Handlers) formatAmountAtTimeAsJSON(amount coin.Amount, timeStamp *time.Time) FormattedAmount {
 	accountCoin := handlers.account.Coin()
-	conversions := map[string]string{}
-	estimated := false
-	if timeStamp != nil {
+	rateUpdater := handlers.account.Config().RateUpdater
+	formatBtcAsSat := util.FormatBtcAsSat(handlers.account.Config().BtcCurrencyUnit)
+	var conversions map[string]string
+	var estimated bool
+
+	if timeStamp == nil {
+		conversions = coin.Conversions(
+			amount,
+			accountCoin,
+			false,
+			rateUpdater,
+			formatBtcAsSat,
+		)
+		estimated = true
+	} else {
 		conversions, estimated = coin.ConversionsAtTime(
 			amount,
-			handlers.account.Coin(),
+			accountCoin,
 			false,
-			handlers.account.Config().RateUpdater,
-			util.FormatBtcAsSat(handlers.account.Config().BtcCurrencyUnit),
+			rateUpdater,
+			formatBtcAsSat,
 			timeStamp,
 		)
 	}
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/amount/conversion-amount.tsx
```diff
@@ -70,12 +69,8 @@ export const ConversionAmount = ({
             unit={conversionUnit}
           />
         </>
-      ) : (
-        <div className={styles.skeletonContainer}>
-          <Skeleton />
-        </div>
-      )}
-      <span className={styles.txUnit}>{' '}{conversionUnit}</span>
+      ) : null }
```
Beerosagos, line 72: cc @thisconnect @shonsirsha ack on this? 

I think that we should use skeletons only while waiting for backend responses, while here the data is just missing.

backend/coins/btc/handlers/handlers.go
```diff
@@ -121,15 +121,27 @@ func (handlers *Handlers) formatAmountAsJSON(amount coin.Amount, isFee bool) For
...
-	conversions := map[string]string{}
-	estimated := false
-	if timeStamp != nil {
+	rateUpdater := handlers.account.Config().RateUpdater
+	formatBtcAsSat := util.FormatBtcAsSat(handlers.account.Config().BtcCurrencyUnit)
+	var conversions map[string]string
+	var estimated bool
+
+	if timeStamp == nil {
+		conversions = coin.Conversions(
```
Beerosagos, line 130: Even if I suggested this approach, I'm still not sure if it wouldn't be better to check for the nil timestamp and possibly fallback on `coin.Conversions` directly inside `coin.ConversionsAtTime`.
No strong opinion, in any case. If you like this let's keep it.

Beerosagos (changes requested):
Please, add some info in the commit messages about why the changes are required :innocent:

frontends/web/src/components/amount/conversion-amount.tsx
```diff
@@ -70,12 +69,8 @@ export const ConversionAmount = ({
...
+      { (conversion || sendToSelf) ? <span className={styles.txUnit}>{' '}{conversionUnit}</span> : null }
```
thisconnect, line 73: nit: I know we don't have prettier (yet), but it would help me reading it a lot by separating jsx and js with some newlines, it also makes a nicer git history in the future 
```suggestion
      { (conversion || sendToSelf) ? (
        <span className={styles.txUnit}>
          {' '}
          {conversionUnit}
        </span>
      ) : null }
```

backend/coins/btc/handlers/handlers.go
bznein: I think the current way makes it more explicit so I'm fine with it, but also I don't feel strongly about it :)

frontends/web/src/components/amount/conversion-amount.tsx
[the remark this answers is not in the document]
bznein: Not sure I follow, the logic for the `if` is slightly different here than from anywhere else. can you clarify what you mean please? :)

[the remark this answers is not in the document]
Beerosagos: Yes that is different, but if we decide that instead of showing the skeleton we want to hide the whole line (doesn't make sense to me to hide only the fiat conversion and show the unit), I think that the logic used to decide when showing the unit should be the same of the amount, and hence we don't need separate if statements.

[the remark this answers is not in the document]
bznein: Ok I *think* I understood, LMK if it seems ok now :)

Beerosagos (approved):
tACK! :pray:

PR #3236, opened 2025-03-12
btcdirect: update private trading desk links

sutterseba:
BTC Direct updated their private trading desk links and asked us to change them.


thisconnect (approved):
LGTM thanks

PR #3240, opened 2025-03-17
qt/linux: don't bundle libwayland-client.so.0

benma:
This fixes this crash:

```
BitBox: symbol lookup error: /lib/x86_64-linux-gnu/libEGL_mesa.so.0:
undefined symbol: wl_display_create_queue_with_name
```

which is likely caused by the target system having a newer version of Mesa that is incompatible with the verison of libwayland-client.so.0 that we bundled.

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos (approved):
LGTM

PR #3244, opened 2025-03-19
ios: update device name in frontend

benma:
The name can be updated either the discovery method being called again for the same device, or via the peripheralDidUpdateName method.


NickeZ (approved):
lgtm

PR #3245, opened 2025-03-19
backend/bitbox02: give simulator test more time to connect

benma:
CI often fails in this step, this is an attempt to fix that.


Beerosagos (approved):
LGTM!

PR #3248, opened 2025-03-21
frontend: remove unused translations

thisconnect:
Removed some leftover translations:

- bitbox02Wizard.initialize.tip is still used but removed all other unused keys in bitbox02Wizard.initialize
- removed bitbox02Wizard.insertSDCard
- removed bitbox02Wizard.success.text


bznein (approved):
LGTM sorry forgot about this one!

PR #3249, opened 2025-03-24
frontend: add received date to coin control UTXO list

strmci:
Add the received date (block header timestamp) to the UTXO detail in the coin control view. This makes it easier for users to view the date, which can be helpful for various reasons, without having to open each UTXO individually in the block explorer.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -840,7 +840,7 @@ func sortByAddresses(result []*SpendableOutput) []*SpendableOutput {
 	for _, s := range sums {
 		outputs := grouped[s.address]
 		sort.Slice(outputs, func(i, j int) bool {
-			return outputs[i].Value > outputs[j].Value
+			return outputs[i].TxOut.Value > outputs[j].TxOut.Value
 		})
 		newResult = append(newResult, outputs...)
 	}
--- a/backend/coins/btc/handlers/handlers.go
+++ b/backend/coins/btc/handlers/handlers.go
@@ -311,18 +311,24 @@ func (handlers *Handlers) getUTXOs(*http.Request) (interface{}, error) {
 	for _, output := range spendableOutputs {
 		address := output.Address.EncodeForHumans()
 		addressReused := addressCounts[address] > 1
-
+		var formattedTime *string
+		timestamp := output.HeaderTimestamp
+		if timestamp != nil {
+			t := timestamp.Format(time.RFC3339)
+			formattedTime = &t
+		}
 		result = append(result,
 			map[string]interface{}{
-				"outPoint":      output.OutPoint.String(),
-				"txId":          output.OutPoint.Hash.String(),
-				"txOutput":      output.OutPoint.Index,
-				"amount":        coin.ConvertBTCAmount(handlers.account.Coin(), btcutil.Amount(output.TxOut.Value), false, accountConfig.RateUpdater),
-				"address":       address,
-				"scriptType":    output.Address.AccountConfiguration.ScriptType(),
-				"note":          handlers.account.TxNote(output.OutPoint.Hash.String()),
-				"addressReused": addressReused,
-				"isChange":      output.IsChange,
+				"outPoint":        output.OutPoint.String(),
+				"txId":            output.OutPoint.Hash.String(),
+				"txOutput":        output.OutPoint.Index,
+				"amount":          coin.ConvertBTCAmount(handlers.account.Coin(), btcutil.Amount(output.TxOut.Value), false, accountConfig.RateUpdater),
+				"address":         address,
+				"scriptType":      output.Address.AccountConfiguration.ScriptType(),
+				"note":            handlers.account.TxNote(output.OutPoint.Hash.String()),
+				"addressReused":   addressReused,
+				"isChange":        output.IsChange,
+				"headerTimestamp": formattedTime,
 			})
 	}
 
--- a/backend/coins/btc/transactions/transactions.go
+++ b/backend/coins/btc/transactions/transactions.go
@@ -15,6 +15,8 @@
 package transactions
 
 import (
+	"time"
+
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/blockchain"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/headers"
@@ -34,7 +36,8 @@ import (
 
 // SpendableOutput is an unspent coin.
 type SpendableOutput struct {
-	*wire.TxOut
+	TxOut           *wire.TxOut
+	HeaderTimestamp *time.Time
 }
 
 // ScriptHashHex returns the hash of the PkScript of the output, in hex format.
@@ -235,7 +238,8 @@ func (transactions *Transactions) SpendableOutputs() (map[wire.OutPoint]*Spendab
 
 				if confirmed || transactions.allInputsOurs(dbTx, txInfo.Tx) {
 					result[outPoint] = &SpendableOutput{
-						TxOut: txOut,
+						TxOut:           txOut,
+						HeaderTimestamp: txInfo.HeaderTimestamp,
 					}
 				}
 			}
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -433,6 +433,7 @@ export type TUTXO = {
   scriptType: ScriptType;
   addressReused: boolean;
   isChange: boolean;
+  headerTimestamp: string | null;
 };
 
 export const getUTXOs = (code: AccountCode): Promise<TUTXO[]> => {
--- a/frontends/web/src/routes/account/send/utxos.tsx
+++ b/frontends/web/src/routes/account/send/utxos.tsx
@@ -33,8 +33,9 @@ import { Amount } from '@/components/amount/amount';
 import { getScriptName } from '@/routes/account/utils';
 import { Message } from '@/components/message/message';
 import { Badge } from '@/components/badge/badge';
-import style from './utxos.module.css';
+import { parseTimeShort } from '@/utils/date';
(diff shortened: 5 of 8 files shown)


backend/coins/btc/transactions/transactions.go
```diff
@@ -221,6 +223,18 @@ func (transactions *Transactions) SpendableOutputs() (map[wire.OutPoint]*Spendab
 	})
 }
 
+// GetHeaderTimestamp retrieves the header timestamp for a given transaction hash.
+func (transactions *Transactions) GetHeaderTimestamp(txHash chainhash.Hash) (*time.Time, error) {
+	transactions.synchronizer.WaitSynchronized()
+	return DBView(transactions.db, func(dbTx DBTxInterface) (*time.Time, error) {
```
benma, line 229: This adds one dbTx per UTXO. In transactions.SpendableOutputs() we aleady call `TxInfo()` on each UTXO, so it would be more natural to simply fetch the timestamp there.

backend/coins/btc/transactions/transactions.go
```diff
@@ -221,6 +223,18 @@ func (transactions *Transactions) SpendableOutputs() (map[wire.OutPoint]*Spendab
...
+		txInfo, err := dbTx.TxInfo(txHash)
+		if err != nil {
+			return nil, err
+		}
+		return txInfo.HeaderTimestamp, nil
```
benma, line 234: HeaderTimestamp is `nil` when the utxo is unconfirmed. If nil you should take the `CreatedTimestamp`, like here:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/backend/coins/btc/handlers/handlers.go#L205-L208

It would probably be good to add a method `Timestamp()` to the `TransactionData` struct and add this logic there.

backend/coins/btc/handlers/handlers.go
```diff
@@ -365,15 +365,16 @@ func (handlers *Handlers) getUTXOs(*http.Request) (interface{}, error) {
...
+				"outPoint":        output.OutPoint.String(),
+				"txId":            output.OutPoint.Hash.String(),
+				"txOutput":        output.OutPoint.Index,
+				"amount":          handlers.formatBTCAmountAsJSON(btcutil.Amount(output.TxOut.Value), false),
+				"address":         address,
+				"scriptType":      output.Address.Configuration.ScriptType(),
+				"note":            handlers.account.TxNote(output.OutPoint.Hash.String()),
+				"addressReused":   addressReused,
+				"isChange":        output.IsChange,
+				"headerTimestamp": output.HeaderTimestamp,
```
benma, line 377: doesn't the timestamp need to be formatted like here?

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/backend/coins/btc/handlers/handlers.go#L215

frontends/web/src/api/account.ts
```diff
@@ -408,6 +408,7 @@ export type TUTXO = {
   scriptType: ScriptType;
   addressReused: boolean;
   isChange: boolean;
+  headerTimestamp: string;
```
benma, line 411: I think in theory both the HeaderTimestamp and the CreatedTimestamp could be `nil`, so that means headerTimestamp should be `string | null`.

benma (changes requested):

frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -127,6 +128,14 @@ export const UTXOs = ({
                       </span>
                       <FiatConversion alwaysShowAmounts amount={utxo.amount} unstyled noAction/>
                     </div>
+                    <div className={style.date}>
+                      <span className={style.label}>
+                        {t('send.coincontrol.receivedDate')}:
+                      </span>
+                      <span className={style.shrink}>
+                        <Date time={utxo.headerTimestamp} />
```
Beerosagos, line 136: I know this date format is used in the tx list of the account, but there the transactions are sorted by time. Here I find it a bit confusing to see dates like `Saturday, March 22`  or `Wednesday, October 30, 2024`. I think it would be easier to have a `dd/mm/YYYY` format or similar.

thisconnect: Please dont' hardcode `dd/mm/YYYY` but use `{parseTimeShort(utxo.headerTimestamp, i18n.language)}`, so the date is localized and short.

```js
new Date().toLocaleString('en-US', {
      month: 'short',
      day: 'numeric',
      year: 'numeric'
    })
"Apr 4, 2025" 

new Date().toLocaleString('de-CH', {
      month: 'short',
      day: 'numeric',
      year: 'numeric'
    })
"4. Apr. 2025" 
```

frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -128,6 +129,16 @@ export const UTXOs = ({
...
                       <AmountWithUnit alwaysShowAmounts amount={utxo.amount} convertToFiat/>
                     </div>
+                    <div className={style.date}>
+                      <span className={style.label}>
+                        {t('send.coincontrol.receivedDate')}:
+                      </span>
+                      <span className={style.shrink}>
+                        {utxo.headerTimestamp
+                          ? parseTimeShort(utxo.headerTimestamp, i18n.language)
+                          : '-'}
```
thisconnect: Why does the dash now show for unconfirmed utxo's?

currently unconfirmed utxo's show "Received date". But if the tx gets confirmed days later, the "received date" is a bit misleading.

The code has a fallback to dash `–`. It seems useful to see that the UTXO is not yet confirmed with the dash.


In the future  we could add an "unconfirmed" badge.

Screenshot 2025-06-06 at 18 37 23

strmci: It falls back now to `CreatedTimestamp` if there is no `HeaderTimestamp`, but I agree.

Shouldn't we display the Received date only as a HeaderTimestamp and for other cases (unconfirmed tx, blockchain not synced yet), show only `Received date: -`? cc @Beerosagos @benma

backend/coins/btc/account.go
```diff
@@ -851,9 +852,10 @@ func sortByAddresses(result []*SpendableOutput) []*SpendableOutput {
 // SpendableOutput is an unspent coin.
 type SpendableOutput struct {
 	*transactions.SpendableOutput
-	OutPoint wire.OutPoint
-	Address  *addresses.AccountAddress
-	IsChange bool
+	OutPoint        wire.OutPoint
+	Address         *addresses.AccountAddress
+	IsChange        bool
+	HeaderTimestamp *time.Time
```
Beerosagos: Isn't the headerTimestamp already included in the `SpendableOutput`? I think you can avoid adding it here too, right?

frontends/web/src/routes/account/send/utxos.tsx
Beerosagos: Yes, I agree. I think that the UTXO should only have a confirmation date, not a creation date (in the backend, we fallback on that for Tx details to keep ETH compatibility, but this is not needed here).

I'm not a fan for the `-`, tho. Maybe a `Pending` label?

backend/accounts/transaction.go
```diff
@@ -128,6 +128,15 @@ type TransactionData struct {
 	IsErc20 bool
 }
 
+// Timestamp returns Header Timestamp, if the tx hasn't been confirmed yet, it returns Created Timestamp.
+func (tx *TransactionData) Timestamp() *time.Time {
```
Beerosagos: If we don't use this for the UTXO date I think we can remove it

Beerosagos:
> I'm not sure if the formatting with RFC3339 is necessary, isn't that done automatically in time package ?
https://pkg.go.dev/time#Time.MarshalJSON

Imo it would be good to format it explicitly for consistency.

Now sure if there is a nice way to add a test for `SpendableOutput.HeaderTimestamp`, maybe in `transactions_test.go:TestSpendableOutputs`? cc @benma 

It would be nice to move the logic outside of `handlers.go`, but since this PR has been opened for a while, I'd do it in a separate one.

Beerosagos (changes requested):

backend/coins/btc/account.go
strmci: Right, removed.

Beerosagos:
Backend LGTM! For the frontend, I meant just a text label, for `Pending`, not a badge. But I'll leave the last word to @thisconnect on this :innocent:

thisconnect:
> Backend LGTM! For the frontend, I meant just a text label, for Pending, not a badge. But I'll leave the last word to @thisconnect on this 😇

we can always add a badge later if it makes sense.

frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -128,6 +129,16 @@ export const UTXOs = ({
...
                       <AmountWithUnit alwaysShowAmounts amount={utxo.amount} convertToFiat/>
                     </div>
+                    <div className={style.date}>
+                      <span className={style.label}>
+                        {t('send.coincontrol.receivedDate')}:
+                      </span>
+                      <span className={style.shrink}>
+                        {utxo.headerTimestamp
+                          ? parseTimeShort(utxo.headerTimestamp, i18n.language)
+                          : t('send.coincontrol.pending')}
```
benma: How about reusing `transaction.status.pending`?

frontends/web/src/routes/account/send/utxos.tsx
```diff
@@ -128,6 +129,16 @@ export const UTXOs = ({
                       </span>
                       <AmountWithUnit alwaysShowAmounts amount={utxo.amount} convertToFiat/>
                     </div>
+                    <div className={style.date}>
+                      <span className={style.label}>
+                        {t('send.coincontrol.receivedDate')}:
```
benma: Reuse `transaction.details.date` here? Avoid adding more things to translate that we already have.

benma (changes requested):
Much better, thanks.

strmci:
@benma I've updated to reuse existing strings, please check.

benma (approved):
tACK, nice!

PR #3254, opened 2025-03-25
frontend/account: remove flicker

benma:
Before the status is loaded, we returned null, causing a flicker when changing accounts, even for static elements.

We can simply proceed with rendering and handle all elements that depend on `status` conditionally to avoid the flicker.


frontends/web/src/routes/account/account.tsx
```diff
@@ -227,24 +225,19 @@ export const Account = ({
...
-    // happens briefly before `setStatus(undefined)` stops rendering again below.
-    return null;
-  }
-  if (!account || status === undefined) {
+  if (!account) {
     return null;
   }
 
-  if (status.fatalError) {
+  if (status !== undefined && status.fatalError) {
```
thisconnect, line 232:
```suggestion
  if (status?.fatalError) {
```

thisconnect (approved):
nice, tested LGTM with small suggestion

PR #3257, opened 2025-03-26
Ios screenlock

benma:
@bznein fyi:

Cancelling an authentication normally should show the 'Authenticate' button to try again, but cancelling the auth when enabling the screen lock in the advanced settings should do nothing (just cancel the screen lock setup). Before this PR, cancelling the setup was buggy.


bznein (approved):
LGTM!

Tested with an emulated iPhone through XCode. Both the "cancel" bug and FaceID (https://developer.apple.com/documentation/xcode/testing-complex-hardware-device-scenarios-in-simulator#Test-Face-ID-or-Touch-ID-authentication) work! The only thing I wasn't able to reproduce is that two failed FaceID match fall back to the passcode auth (as mentioned in the commit message), but maybe that is a limitation of the emulator (or mine :D)

PR #3259, opened 2025-03-27
backend: add argument to force highest fee rate in tx proposal

sutterseba:
Previously, the highest fee rate was automatically applied to all transaction proposals with a SLIP-24 payment request. For selling and spending integrations that do not use SLIP-24 payment requests, there should also be an option to trigger this same behaviour.

This commit replaces the previous conditioning on payment requests with a general boolean field that is set independently and can be used for other integrations. This also allows setting a custom fee rate for transaction proposals with payment requests, if needed.

### DIFF
--- a/backend/coins/btc/transaction.go
+++ b/backend/coins/btc/transaction.go
@@ -39,8 +39,7 @@ const unitSatoshi = 1e8
 // fee target (priority) if one is given, or the provided args.FeePerKb if the fee taret is
 // `FeeTargetCodeCustom`.
 func (account *Account) getFeePerKb(args *accounts.TxProposalArgs) (btcutil.Amount, error) {
-	isPaymentRequest := args.PaymentRequest != nil
-	if args.FeeTargetCode == accounts.FeeTargetCodeCustom && !isPaymentRequest {
+	if args.FeeTargetCode == accounts.FeeTargetCodeCustom && !args.UseHighestFee {
 		float, err := strconv.ParseFloat(args.CustomFee, 64)
 		if err != nil {
 			return 0, err
@@ -60,7 +59,7 @@ func (account *Account) getFeePerKb(args *accounts.TxProposalArgs) (btcutil.Amou
 	}
 
 	var feeTarget *FeeTarget
-	if isPaymentRequest {
+	if args.UseHighestFee {
 		feeTarget = account.feeTargets().highest()
 	} else {
 		for _, target := range account.feeTargets() {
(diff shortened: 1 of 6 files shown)


frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -212,6 +212,7 @@ export const Pocket = ({ code, action }: TProps) => {
       sendAll: 'no',
       selectedUTXOs: [],
       paymentRequest: message.slip24,
+      useHighestFee: true
```
thisconnect, line 215: Shouldn't setting useHighestFee be enough or  it still required/makes-sense  to pass: `feeTarget: 'custom', customFee: '',`?

backend/coins/btc/transaction.go
```diff
@@ -39,8 +39,7 @@ const unitSatoshi = 1e8
 // fee target (priority) if one is given, or the provided args.FeePerKb if the fee taret is
 // `FeeTargetCodeCustom`.
 func (account *Account) getFeePerKb(args *accounts.TxProposalArgs) (btcutil.Amount, error) {
```
benma, line 41: This whole file is severly lacking unit tests. Some day :sob:

frontends/web/src/routes/exchange/pocket.tsx
sutterseba: It's still a required field but ignored if `useHighestFee` is true, just like previously when there was a payment request

benma (changes requested):

backend/coins/btc/transaction.go
```diff
@@ -39,8 +39,7 @@ const unitSatoshi = 1e8
...
-	isPaymentRequest := args.PaymentRequest != nil
-	if args.FeeTargetCode == accounts.FeeTargetCodeCustom && !isPaymentRequest {
+	if args.FeeTargetCode == accounts.FeeTargetCodeCustom && !args.UseHighestFee {
```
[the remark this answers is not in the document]
sutterseba: but that part is specifically to evaluate the custom fee rate, no? `if args.UseHighestFee` is further down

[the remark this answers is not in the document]
benma: oh yes, I missed that :+1:

frontends/web/src/api/account.ts
```diff
@@ -320,12 +320,20 @@ export const getReceiveAddressList = (code: AccountCode) => {
...
-};
+} & (
+  {
+    useHighestFee: false;
+    customFee: string;
+    feeTarget: FeeTargetCode;
+  } | {
+    useHighestFee: true;
+    customFee?: never;
+    feeTarget?: never;
```
benma, line 334: I think you can just remove those.

PR #3280, opened 2025-04-02
backend/handlers: remove unused /rates

benma:
The frontend call that used it was removed in 4d30d09222eb35b248503ad4965b3ae70c40b151


Beerosagos (approved):
utACK!

PR #3281, opened 2025-04-03
tests: add tests for TxProposal.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests


bznein:
> * [nitpick] Consider renaming the interface 'Interface' to a more descriptive name such as 'TransactionsInterface' to improve code clarity.

This is actually the exact opposite as what the linter complained about :D since `TransactionsInterface` would cause stutter as the package is already called `Transactions`

backend/coins/btc/transaction_test.go
```diff
@@ -44,6 +46,65 @@ func mustKeypath(t *testing.T, keypath string) signing.AbsoluteKeypath {
...
+					return btcutil.Amount(10e4), nil
+				default:
+					return btcutil.Amount(10e6), nil
+				}
+			},
+		}
+	})
+	account.coin.blockchain = account.coin.makeBlockchain()
+	require.NoError(t, account.Initialize())
+	account.ensureAddresses()
```
benma, line 98: this is called by Initialize() already, so could be skipped? Seems too low level to have to deal with in a unit test.

backend/coins/btc/transaction_test.go
```diff
@@ -44,6 +46,65 @@ func mustKeypath(t *testing.T, keypath string) signing.AbsoluteKeypath {
 	return kp
 }
 
+type MockTransaction struct {
```
benma, line 49: Consider using `moq` to auto-generate a mock struct, like

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/b9cc719ea68ffffa3ed961d56504b2111dc4dd05/backend/accounts/account.go#L67

which generates backend/accounts/account/mocks/account.go.

Then you can can use it locally in the test below and set the functions directly, so it's closer to the test, like 

```
acccount.transactions = mocks.InterfaceMock{
    SpendableOutputsFunc: func() (...) { ... },
    ....
}
```

backend/coins/btc/transaction_test.go
bznein: Good catch, done

backend/coins/btc/transaction_test.go
bznein: Yep good idea, done

backend/coins/btc/transaction_test.go
```diff
@@ -269,10 +287,56 @@ func TestPickChangeAddressFails(t *testing.T) {
...
-	account.ensureAddresses()
 	require.Eventually(t, account.Synced, time.Second, time.Millisecond*200)
 
 	account.subaccounts = subaccounts{}
 	_, err := account.pickChangeAddress(nil)
 	require.ErrorContains(t, err, "Account has no subaccounts")
 }
+
+func TestTxProposal(t *testing.T) {
+	testCases := []struct {
```
benma, line 298: would be nice to also test:

- SpendAll (one of the two major modes of spending)
- SelectedUTXOs for coin control, with and without coin control
- that amount parsing works as expected if sats mode is enabled (user enters the amount in satohis instead of BTC)

bznein: Done!

bznein: > SelectedUTXOs for coin control, with and without coin control

Though I'm not sure what you mean "with and without coin control". Isn't specifying SelectedUTXOs in the args efffectively "coin control"?

benma: Yeah, it's the same.

By with and without coin control I meant: there should be a test using SelectedUTXOs with SendAll, and one using SelectedUTXOs with a regular amount (not send all). The first is now there, the second is still missing I think.

benma (changes requested):

backend/coins/btc/transaction_test.go
bznein: I see what you mean now, done!

benma (approved):
LGTM

PR #3282, opened 2025-04-03
ci: fix windows workflow

bznein:
Maybe there's a better way, and I can't really figure out the root cause, but this seems to fix the problem: 

https://github.com/bznein/bitbox-wallet-app/actions/runs/14241588861/job/39912541043

- [ ] updating the Changelog
- [ ] writing unit tests


.github/workflows/ci.yml
```diff
@@ -224,6 +224,7 @@ jobs:
         with:
           # Take Go version to install from go.mod
           go-version-file: 'go.mod'
+      - run: pip install wheel
```
benma, line 227: Always consider someone maintaining this in a few years wondering what this is for :grin: Please add a comment here (or in the commit msg) that explains what exactly this does.

It seems it's needed for `aqt` for some reason, the tool used by the Qt install action below to install Qt. Consider opening an issue at https://github.com/jurplel/install-qt-action/issues with the CI error and context and linking the issue here (so we know we can remove this again once fixed).

bznein: Thanks, done and opened https://github.com/jurplel/install-qt-action/issues/281

.github/workflows/ci.yml
```diff
@@ -224,6 +224,10 @@ jobs:
         with:
           # Take Go version to install from go.mod
           go-version-file: 'go.mod'
+      # We need to install wheel otherwise the below will fai
```
benma, line 227: s/fai/fail

PR #3285, opened 2025-04-07
frontend: fullscreen selector on mobile

shonsirsha:
to improve the mobile UX, we've introduced a new component called MobileFullscreenSelector.

This component is rendered automatically when the Dropdown component is used with the prop mobileFullScreen={true}. It is integrated in this commit across the following components:

- ActiveCurrenciesDropdownSetting
- DefaultCurrencyDropdownSetting
- LanguageDropdownSetting

Screenshot 2025-04-15 at 07 41 32
Screenshot 2025-04-15 at 07 41 14

Screenshot 2025-04-15 at 07 41 43

ab_3
ab_4
ab_5

### DIFF
--- a/frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
+++ b/frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
@@ -112,14 +112,21 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
         return false;
       }}
       />
-      <button
-        className={styles.mobileDropdownTrigger}
-        onClick={handleOpen}
-        type="button"
-      >
-        <span className={styles.mobileDropdownValue}>{displayValue}</span>
-        <div className={styles.dropdown} />
-      </button>
+      {onOpenChange ? (
+        <div
+          className={styles.mobileSelectorTrigger}
+        >
+          <span className={styles.mobileSelectorValue}>{displayValue}</span>
+        </div>
+      ) : (
+        <button
+          className={styles.mobileSelectorTrigger}
+          onClick={handleOpen}
+        >
+          <span className={styles.mobileSelectorValue}>{displayValue}</span>
+        </button>
+      )
+      }
 
       {isOpen && (
         <div className={styles.fullscreenOverlay}>
@@ -156,7 +163,6 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
                     key={String(option.value)}
                     className={`${styles.optionItem} ${isSelected(option) ? styles.selectedOption : ''}`}
                     onClick={(e) => handleSelect(option, e)}
-                    type="button"
                   >
                     <div className={styles.optionContent}>{renderOptions(option)}</div>
                   </button>
--- a/frontends/web/src/routes/settings/components/appearance/activeCurrenciesDropdownSetting.tsx
+++ b/frontends/web/src/routes/settings/components/appearance/activeCurrenciesDropdownSetting.tsx
@@ -26,7 +26,7 @@ const ActiveCurrenciesDropdownSetting = () => {
   const { t, i18n } = useTranslation();
   const { activeCurrencies, defaultCurrency } = useContext(RatesContext);
   const { formattedCurrencies } = useLocalizedFormattedCurrencies(i18n.language);
-  const [isMobileDropdownOpen, setIsMobileDropdownOpen] = useState(false);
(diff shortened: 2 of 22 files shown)


frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
```diff
@@ -0,0 +1,171 @@
...
+      <button
+        className={styles.mobileDropdownTrigger}
+        onClick={handleOpen}
+        type="button"
+      >
+        <span className={styles.mobileDropdownValue}>{displayValue}</span>
+        <div className={styles.dropdown} />
+      </button>
+
+      {isOpen && (
```
thisconnect, line 125: Currently if you use native back on Android it goes to mobilesettings, but it would be nice if it goes to general settings.

Could you intercept and prevent the native back behavior and close the fullscreen selector instead?

untested but something like this I believe:

```jsx
<UseBackButton handler={() => {
  handleBackButtonClick();
  return false;
}} />
```

frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const isOpen = isControlled ? controlledIsOpen : internalIsOpen;
+
+  useEffect(() => {
+    if (!isOpen) {
+      setSearchText('');
+    }
+  }, [isOpen]);
+
+  const handleOpen = () => {
+    if (isControlled && onOpenChange) {
```
thisconnect, line 59: `isControlled` already contains `onOpenChange`, so it looks like it is not needed to check for `onOpenChange` again.

small nit: I find it a bit hard to read:  `isOpen` arg, `isOpen` const, `controlledIsOpen`, `internalIsOpen`, `isControlled` maybe this could be slightly simplified? But if too much hassle leave as is.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Reduced support for BitBox01
 - Fix a bug that would prevent the app to perform firmware upgrade when offline.
 - Replace sidebar with bottom navigation bar for mobile devices
+- Introduce full screen selector for mobile
```
thisconnect, line 8: nit: as a user I might not understand what this means, consider mentioning "instead of dropdown"  or similar ?

thisconnect (changes requested):

thisconnect:
On settings page please vertically center the caret-right icons

signal-2025-04-09-100039

jadzeidan:
@shonsirsha Thanks for screen shots.

Regarding the caret, I also think it makes sense to center it as it looks a bit wrong to me otherwise.

I took a look at iOS settings, and they also center it when there are multi-lines in the same setting button:

image

frontends/web/src/routes/settings/components/appearance/activeCurrenciesDropdownSetting.tsx
```diff
@@ -25,13 +26,20 @@ const ActiveCurrenciesDropdownSetting = () => {
...
+  const [isMobileDropdownOpen, setIsMobileDropdownOpen] = useState(false);
+  const isMobile = useMediaQuery('(max-width: 768px)');
+
   return (
     <SettingsItem
+      disabled={!isMobile}
       collapseOnSmall
       settingName={t('newSettings.appearance.activeCurrencies.title')}
       secondaryText={t('newSettings.appearance.activeCurrencies.description')}
+      onClick={() => setIsMobileDropdownOpen(true)}
```
thisconnect, line 38: What do you think about conditionally adding this callback in case isMobielDropdownOpen is false.

Then SettingsItem should not render as button in case the dropdown is open.....

frontends/web/src/routes/settings/components/appearance/activeCurrenciesDropdownSetting.tsx
```diff
@@ -25,13 +26,20 @@ const ActiveCurrenciesDropdownSetting = () => {
...
       extraComponent={
         <ActiveCurrenciesDropdown
```
thisconnect, line 40: Since this is not always a Dropdown now, what do you think about renaming to something else like Selector or Picker?

but it seems you'd need to rename a bit more than just the component i.e. isMobileDropdownOpen -> isMobileSelectorOpen etc.

thisconnect (approved):
LGTM The 2 open points are just suggestions.

PR #3286, opened 2025-04-07
frontend: fix inconsistent font

shonsirsha:
Button and Textarea had different fonts compared to the rest of the app. To improve the UI, we'd like to unify the font family.


thisconnect (approved):
LGTM 👍

PR #3292, opened 2025-04-14
frontend: add and use 3rd party guide links and text

shonsirsha:
For third party services, we'd like to use their own links to be put in the Guide.

### DIFF
--- a/frontends/web/src/components/guide/entry.tsx
+++ b/frontends/web/src/components/guide/entry.tsx
@@ -23,7 +23,7 @@ export type TEntryProp = {
     title: string;
     text: string;
     link?: {
-        url: string;
+        url?: string;
         text: string;
     };
 }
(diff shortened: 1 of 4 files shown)


frontends/web/src/components/guide/guide.tsx
```diff
@@ -54,11 +59,35 @@ const Guide = ({ children, title = t('guide.title') }: TProps) => {
...
-              {t('guide.appendix.link')}
-            </A>
+            {service ? (
+              <>
+                <span className={style.question}>{t('guide.appendix.questionService', { serviceName: service.name })}</span>
+                <br />
+                <br />
+                {t('guide.appendix.textService', { serviceName: service.name })}
+                {' '}
+                {service.linkDisplay.includes('@') ? (
```
thisconnect, line 69: nit: in theory uri's could contain `@` symbol as well.

With that in mind an alternative may be to make link optional  `link?: string;` (as you don't use service.link in this case).

frontends/web/src/components/guide/guide.tsx
```diff
@@ -54,11 +59,35 @@ const Guide = ({ children, title = t('guide.title') }: TProps) => {
...
+                  <span className={style.link}>
+                    {t('guide.appendix.linkService', { link: service.linkDisplay })}
+                  </span>
```
thisconnect, line 72: I just realized that this is not ideal for RTL langauges. It was already wrong before, but with this PR it adds more trnaslations, it would be good to check if we can make it work with RTL before translators translate it.

thisconnect (approved):
tested LGTM, thank you.

PR #3296, opened 2025-04-17
backend: show send-to-self message for other accounts.

bznein:
Enable showing when a transaction's output is an address into a different account on the same keystore.

- [ ] updating the Changelog
- [ ] writing unit tests


backend/devices/bitbox02/keystore_simulator_test.go
```diff
@@ -412,12 +419,48 @@ func makeTx(t *testing.T, device *Device, recipient *maketx.OutputInfo) *btc.Pro
...
+		require.NoError(t, device.Keystore().SignTransaction(proposedTransaction))
+		require.NoError(t, proposedTransaction.Finalize())
+		require.NoError(
+			t,
+			btc.TxValidityCheck(
+				proposedTransaction.TXProposal.Transaction,
+				proposedTransaction.TXProposal.PreviousOutputs,
+				proposedTransaction.TXProposal.SigHashes()))
+
+		// Before simulator v9.20, address confirmation data was not written to stdout.
```
benma, line 457: v9.20 in comment, but v9.22 in the if check below. What about v9.20 and v9.21?

backend/devices/bitbox02/keystore_simulator_test.go
```diff
@@ -465,7 +508,11 @@ func TestSimulatorSignBTCTransactionSendSelfSameAccount(t *testing.T) {
 
 		// Before simulator v9.20, address confirmation data was not written to stdout.
 		if device.Version().AtLeast(semver.NewSemVer(9, 20, 0)) {
-			require.Contains(t, stdOut.String(), "ADDRESS: This BitBox02: bc1pg848p0rvmj0r3j064prlpw0gecyzkwlpt7ndzdlmh2mvkyu299psetgxhf")
+			if device.Version().AtLeast(semver.NewSemVer(9, 22, 0)) {
```
benma, line 511: No need for nesting here, >=v9.22 implies >=v9.20. 

Better imho: `if >= v9.22 { ...} else if >= v9.20 {... }`

backend/coins/btc/account.go
```diff
@@ -137,13 +140,13 @@ func NewAccount(
...
-		dbSubfolder:    "", // set in Initialize()
-		forceGapLimits: forceGapLimits,
-
-		log:        log,
-		httpClient: httpClient,
+		BaseAccount:                     accounts.NewBaseAccount(config, coin, log),
+		coin:                            coin,
+		dbSubfolder:                     "", // set in Initialize()
+		forceGapLimits:                  forceGapLimits,
+		getAddressFromDifferentAccounts: getAddress,
```
benma, line 147: why different names for the struct field and param name? By looking at the actual callback, it returns addresses not only from different accounts, but also from the same account, so the name is a bit confusing.

backend/coins/btc/account.go
```diff
@@ -129,6 +131,7 @@ func NewAccount(
 	config *accounts.AccountConfig,
 	coin *Coin,
 	forceGapLimits *types.GapLimits,
+	getAddress func(blockchain.ScriptHashHex) *addresses.AccountAddress,
```
benma, line 134: please document what exactly this function should do, it's unclear if it returns addresses from any account, or from any account from the same keystore, or from *other* account of the same keystore, etc.

benma (changes requested):

backend/devices/bitbox02/keystore_simulator_test.go
```diff
@@ -412,12 +419,48 @@ func makeTx(t *testing.T, device *Device, recipient *maketx.OutputInfo) *btc.Pro
...
+			}
+			return nil
+		},
 		Signatures: make([]*types.Signature, len(txProposal.Transaction.TxIn)),
 		FormatUnit: coinpkg.BtcUnitDefault,
 	}
 
 }
 
+func TestSimulatorSignBTCTransactionSendSelfDifferentAccount(t *testing.T) {
```
benma, line 436: Maybe also add a unit test that covers sending to an address of a different keystore (which I suspect does not currently work as expected).

benma (changes requested):

backend/devices/bitbox02/keystore_simulator_test.go
bznein: This is actually a leftover comment from a copy-paste :)

backend/devices/bitbox02/keystore_simulator_test.go
bznein: Thanks, done

backend/coins/btc/account.go
bznein: I totally agree, I found it hard to name things in this PR :) I went for "GetAddressFromSameKeystore" (based on another change you suggested) but open to different naming

backend/coins/btc/account.go
bznein: I tried to give it a better name and add a comment, LMK how it looks like now

backend/devices/bitbox02/keystore.go
```diff
@@ -422,15 +439,33 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
...
+				msgScriptType, ok := btcMsgScriptTypeMap[accountConfiguration.ScriptType()]
+				if !ok {
+					return errp.Newf("Unsupported script type %s", accountConfiguration.ScriptType())
+				}
+				outputScriptConfigIndex = addOutputScriptConfig(&messages.BTCScriptConfigWithKeypath{
+					ScriptConfig: firmware.NewBTCScriptConfigSimple(msgScriptType),
+					Keypath:      accountConfiguration.AbsoluteKeypath().ToUInt32(),
+				})
+				outputScriptIndex = &outputScriptConfigIndex
+				isOurs = true
```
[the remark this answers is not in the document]
bznein: Yeah by changing the callback function this will not be the case anymore (I still need to add a test for this though)

backend/accounts.go
```diff
@@ -874,12 +876,30 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
 		BtcCurrencyUnit:  backend.config.AppConfig().Backend.BtcUnit,
 	}
 
+	// This function is passed as a callback to the BTC account constructor. It is called when the
+	// keystore needs to determine whether an address belongs to a different account on the same
+	// keystore.
+	getAddressCallback := func(scriptHashHex blockchain.ScriptHashHex) *addresses.AccountAddress {
+		for _, account := range backend.accounts {
+			// This only makes sense for BTC accounts.
+			btcAccount, ok := account.(*btc.Account)
```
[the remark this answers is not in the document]
bznein: Thanks, I changed to check if it's the same keystore (not sure if this is the best way to achieve it), and tested manually that it works as intended; I'm still missing the unit test though

backend/devices/bitbox02/keystore_simulator_test.go
bznein: @benma I'm not sure that would give us any meaningful coverage; the logic to exclude different keystores is in the `GetAddress` function, which we mock here, so we would basically only change the mock function.

Another idea was to extract the `getAddressCallback` function into a separate fn and test it. However, the reason this also not works is that the functions needs to perform a type cast to `btc.Account` (as it needs to call `GetAddress`, which is not parts of the `account` interface) so this won't work with the mock type generated by `moq`. What do you think? I can't see a valid way to add a unit test for this but open to suggestions!

backend/devices/bitbox02/keystore_simulator_test.go
```diff
@@ -412,12 +419,77 @@ func makeTx(t *testing.T, device *Device, recipient *maketx.OutputInfo) *btc.Pro
 		GetPrevTx: func(chainhash.Hash) (*wire.MsgTx, error) {
 			return prevTx, nil
 		},
+		GetAddress: func(scriptHashHex blockchain.ScriptHashHex) (*addresses.AccountAddress, error) {
+			for _, address := range addrsInDifferentAccount {
```
benma, line 423: This is not a faithful mock, as this callback in prod also returns addresses of the same account, right? So you should include all addresses here :thinking:

backend/devices/bitbox02/keystore_simulator_test.go
```diff
@@ -412,12 +419,77 @@ func makeTx(t *testing.T, device *Device, recipient *maketx.OutputInfo) *btc.Pro
...
+			t,
+			btc.TxValidityCheck(
+				proposedTransaction.TXProposal.Transaction,
+				proposedTransaction.TXProposal.PreviousOutputs,
+				proposedTransaction.TXProposal.SigHashes()))
+
+		// Send-to-self message from different accounts is only available on v9.22 and later.
+		if device.Version().AtLeast(semver.NewSemVer(9, 22, 0)) {
+			require.Contains(t, stdOut.String(), "ADDRESS: This BitBox (account #2): bc1qeunlpt8umlyyv0mpzpe3uauqlezsh39ctwg4e5")
+		}
```
benma, line 460: I'd still add an `else { ... }` and include the expected output for v9.20-v9.21, for clarity.

benma (changes requested):

backend/coins/btc/sign.go
```diff
@@ -68,7 +69,8 @@ func (account *Account) signTransaction(
 	proposedTransaction := &ProposedTransaction{
 		TXProposal:                   txProposal,
 		AccountSigningConfigurations: signingConfigs,
-		GetAccountAddress:            account.getAddress,
+		GetAccountAddress:            account.GetAddress,
```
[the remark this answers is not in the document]
bznein: Done

backend/coins/btc/sign.go
```diff
@@ -37,6 +37,7 @@ type ProposedTransaction struct {
 	// Signatures collects the signatures, one per transaction input.
 	Signatures []*types.Signature
 	FormatUnit coin.BtcUnit
+	GetAddress func(blockchain.ScriptHashHex) (*addresses.AccountAddress, error)
```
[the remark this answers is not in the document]
bznein: Moved both to a single function, LMK how it looks to you!

backend/devices/bitbox02/keystore_simulator_test.go
bznein: Given that I consolidated two functions in a single one, I changed the mock, lmk what you think of  this one

backend/devices/bitbox02/keystore_simulator_test.go
bznein: Done

backend/accounts.go
```diff
@@ -874,12 +876,43 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
...
+		}
+		for _, account := range accountsByKeystore[hex.EncodeToString(rootFingerprint)] {
+			// This only makes sense for BTC accounts.
+			btcAccount, ok := account.(*btc.Account)
+			if !ok {
+				continue
+			}
+			// Skip accounts that are not BTC or TBTC.
+			coinCode := btcAccount.Coin().Code()
+			if coinCode != coinpkg.CodeBTC && coinCode != coinpkg.CodeTBTC {
```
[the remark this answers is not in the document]
bznein: Thanks, done. I'm not sure the way of checking the account is the best one (I really don't like it) but I couldn't find a better way, open to suggestions!

backend/accounts.go
```diff
@@ -872,12 +874,42 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
...
+		if err != nil {
+			return nil, false, err
+		}
+		for _, account := range accountsByKeystore[hex.EncodeToString(rootFingerprint)] {
+			// This only makes sense for BTC accounts.
+			btcAccount, ok := account.(*btc.Account)
+			if !ok {
+				continue
+			}
+			// Skip accounts that are not BTC or TBTC.
```
benma, line 894: comment out of date

backend/accounts.go
```diff
@@ -872,12 +874,42 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
...
+			coinCode := btcAccount.Coin().Code()
+			if coinCode != coinpkg.CodeBTC && coinCode != coinpkg.CodeTBTC && coinCode != coinpkg.CodeLTC {
```
benma, line 896: I think the check you want is `if coinCode != askingAccount.Coin().Code() { continue }`. This is what I meant by

> You should probably also check that the addresses returned here are from the same account code as the account asking for the address.

After all, it doesn't make sense to e.g. return a Litecoin address here if the account asking is a Bitocoin account.

(also you needlessly excluded Litecoin Testnet :P)

backend/devices/bitbox02/keystore.go
```diff
@@ -399,38 +414,55 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
...
-		isOurs := outputAccountAddress != nil
+		isOurs := outputAddress != nil
 		if !isChange && !keystore.device.Version().AtLeast(semver.NewSemVer(9, 15, 0)) {
 			// For firmware older than 9.15.0, non-change outputs cannot be marked internal.
 			isOurs = false
 		}
 
 		var keypath []uint32
 		var scriptConfigIndex int
+		var outputScriptIndex *uint32
```
benma, line 432: rename to `outputScriptConfigIndex`

backend/devices/bitbox02/keystore.go
```diff
@@ -399,38 +414,55 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
...
 		if isOurs {
-			keypath = outputAccountAddress.Configuration.AbsoluteKeypath().ToUInt32()
-			accountConfiguration := outputAccountAddress.AccountConfiguration
+			keypath = outputAddress.Configuration.AbsoluteKeypath().ToUInt32()
```
benma, line 434: Best not to supply this to the BitBox if `isOurs` is false, for clarity. I am a little surprised even that the BitBox doesn't error if this is supplied even if `ours` is false :yum:

benma (changes requested):

backend/coins/btc/sign.go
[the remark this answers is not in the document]
benma: Imho much better. Thanks!

backend/accounts.go
bznein: Removed

backend/devices/bitbox02/keystore.go
bznein: done

backend/devices/bitbox02/keystore.go
bznein: Yeah good idea (but no, it doesn't error :P) done

benma (approved):
tACK, thanks a lot!

PR #3298, opened 2025-04-18
chore: remove duplicate package imports

riskrose:
- [ ] updating the Changelog
- [ ] writing unit tests

remove duplicate package imports


benma (approved):
Thanks! I guess there might be more of these around. I wonder if there is a lint for this in golangci-lint that could be enabled.

PR #3301, opened 2025-04-22
update notes import and export file type to jsonl

sutterseba:
Changes the file type of notes exports to .jsonl (JSON Lines), which is a more common format used by compatible wallets as specified in BIP-329.

For example, exporting labels with Sparrow Wallet creates a .jsonl, which would previously have to be renamed manually in order to allow importing it in the BitBoxApp.

The supported import file types are updated accordingly, still allowing .txt imports for exports created prior to this change.

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -51,6 +51,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -352,13 +353,8 @@ public class MainActivity extends AppCompatActivity {
                 MainActivity.this.filePathCallback = filePathCallback;
                 String[] mimeTypes = fileChooserParams.getAcceptTypes();
                 String fileType = "*/*";
-                if (mimeTypes.length == 1) {
-                    // import notes form uses .txt file type, but is not supported here.
-                    if (".txt".equals(mimeTypes[0])) {
-                        fileType = "text/plain";
-                    } else if (MimeTypeMap.getSingleton().hasMimeType(mimeTypes[0])) {
-                        fileType = mimeTypes[0];
-                    }
+                if (mimeTypes.length == 1 && MimeTypeMap.getSingleton().hasMimeType(mimeTypes[0])) {
+                    fileType = mimeTypes[0];
                 }
                 mGetContent.launch(fileType);
                 return true;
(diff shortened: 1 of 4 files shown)


benma (changes requested):
I assume this also needs to be extended:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/1e5ec7466a96fc7655b8c6df0c728185aed1156e/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java#L356-L358

sutterseba:
@benma PTAL, change assumes that getAcceptTypes() returns two elements `.jsonl` and `.txt` (not `".jsonl,.txt"` as a single one). The latter should not happen afaiu but I can't really test on Android.

(Kept the existing check to map to a single valid type)

thisconnect:
tested a6d1b3e6fa1d32e600431fcfdefd21f70f8d30c4 on Android.
 
Importing `.txt` worked. somehow I can only see `.txt` and not `.jsonl`. If I browse to Downloads, I can see the `.jsonl` file but grayed out. Only the txt file is selectable.

sutterseba:
The filer picker shows `.jsonl` files as binary (i.e. it doesn't recognize it), so I opted to remove the file type restriction now. In notes import, any file should be able to be selected again. Not ideal, but restricting the picker to only allow .txt and .jsonl would probably require a more involved change.

sutterseba:
rebased, sry forgot to re-request review

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -352,13 +353,8 @@ public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathC
                 MainActivity.this.filePathCallback = filePathCallback;
                 String[] mimeTypes = fileChooserParams.getAcceptTypes();
                 String fileType = "*/*";
-                if (mimeTypes.length == 1) {
-                    // import notes form uses .txt file type, but is not supported here.
-                    if (".txt".equals(mimeTypes[0])) {
-                        fileType = "text/plain";
-                    } else if (MimeTypeMap.getSingleton().hasMimeType(mimeTypes[0])) {
-                        fileType = mimeTypes[0];
-                    }
```
benma, line 361: Checking this again and seeing that the previous commit to restrict to the right file types (https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/045aa86aa767f034c38b522e0dc1308dbc1e3540) is undone.

You write

>  The (android) filer picker shows .jsonl files as binary (i.e. it doesn't recognize it),

I guess it's not as easy as adding a ".jsonl" check same as the ".txt" check above? :thinking:

sutterseba: Yes, checking for both `jsonl` and `txt` exclusively like before does not work, hence the full fallback to `*/*`.  It would probably require a more involved custom change, which I'm not sure how to do.

benma:
Thanks, and sorry for the late approve! I rebased and resolved the CHANGELOG conflict so you don't have to bother :smile:

PR #3302, opened 2025-04-23
frontend: fix confirmation time for ERC20 tokens.

bznein:
The current GetCoinCode logic omits erc20 tokens, so we just use the "standard" text for confirmation time. 

With this change, we return "eth" for erc20 tokens so that we can show the correct confirmation time.

image

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/account/utils.ts
+++ b/frontends/web/src/routes/account/utils.ts
@@ -64,6 +64,9 @@ export const getCoinCode = (coinCode: CoinCode): CoinCode | undefined => {
   case 'sepeth':
     return 'eth';
   }
+  if (coinCode.startsWith('eth-erc20-')) {
+    return 'eth';
+  }
 };
 
 export const getScriptName = (scriptType: ScriptType): string => {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/utils.ts
```diff
@@ -64,6 +64,9 @@ export const getCoinCode = (coinCode: CoinCode): CoinCode | undefined => {
   case 'sepeth':
     return 'eth';
   }
+  if (coinCode.startsWith('eth-erc20-')) {
+    return 'eth';
+  }
 };
```
thisconnect, line 70: Looks like getCoinCode function is only used in feetargets component.

thisconnect (approved):
untested LGTM

frontends/web/src/routes/account/utils.ts
bznein: Yep I checked as I didn't want to break other stuff :) Happy to move it somewhere else if needed. I'll merge this PR now but I can open a new one if needed!

PR #3304, opened 2025-04-24
backend/bitbox02bootloader: add placeholder firmwares for BB02+

benma:
The bootloader assumes there are entries in this map for each product, otherwise it returns errors in various places. We add placeholder firmwares (same as BitBox02 firmwares but with adjusted magic bytes) so we can test the bootloader and firmware upgrades on BB02+ for now.


NickeZ (approved):
utack

PR #3306, opened 2025-04-24
chore: remove duplicate imports.

bznein:
Inspired by https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3298, I had ChatGPT come up with an ugly but working bash script that finds all duplicate imports. It looks like golangci doesn't have a linter for this specific.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -24,11 +24,9 @@ import (
 	"time"
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
-	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
 	accountsTypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/bitsurance"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
-	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/config"
@@ -253,7 +251,7 @@ func (backend *Backend) accountFiatBalance(account accounts.Interface, fiat stri
 		return nil, err
 	}
 
-	coinDecimals := coin.DecimalsExp(account.Coin())
+	coinDecimals := coinpkg.DecimalsExp(account.Coin())
 	price, err := backend.RatesUpdater().LatestPriceForPair(account.Coin().Unit(false), fiat)
 	if err != nil {
 		return nil, err
@@ -357,9 +355,9 @@ func (backend *Backend) LookupInsuredAccounts(accountCode accountsTypes.Code) ([
 					if !ok {
 						frontendConfig = make(map[string]interface{})
 					}
-					canceledAccounts, ok := frontendConfig["bitsuranceNotifyCancellation"].([]types.Code)
+					canceledAccounts, ok := frontendConfig["bitsuranceNotifyCancellation"].([]accountsTypes.Code)
 					if !ok {
-						frontendConfig["bitsuranceNotifyCancellation"] = []types.Code{bitsuranceAccount.AccountCode}
+						frontendConfig["bitsuranceNotifyCancellation"] = []accountsTypes.Code{bitsuranceAccount.AccountCode}
 					} else {
 						canceledAccounts = append(canceledAccounts, bitsuranceAccount.AccountCode)
 						frontendConfig["bitsuranceNotifyCancellation"] = canceledAccounts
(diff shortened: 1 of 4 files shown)


backend/accounts.go
```diff
@@ -25,11 +25,9 @@ import (
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
-	accountsTypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/bitsurance"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
-	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
```
benma, line 32: I think in case of dups, you should keep the named one and remove the unnamed one. The reason for naming these imports is to avoid clashes or generic names. `func foo(coin coin.Coin)` is no good :sweat_smile: 

Not sure what the idiomatic way is, as I assume renaming imports is not very idiomatic. I am afraid it might be one letter var name like `func foo(c coin.Coin)`, but I hate short names with a passion.

bznein:
> So where is the script? 😁 Could it be added to CI?

Well there's the script:

```
#!/bin/bash

find . -name "*.go" | while read -r file; do
  in_import_block=0
  imports=()

  while IFS= read -r line; do
    # Start of import block
    if [[ $line =~ ^[[:space:]]*import[[:space:]]*\( ]]; then
      in_import_block=1
      continue
    fi

    # End of import block
    if [[ $in_import_block -eq 1 && $line =~ ^[[:space:]]*\) ]]; then
      in_import_block=0
      continue
    fi

    # Single import line
    if [[ $line =~ ^[[:space:]]*import[[:space:]]+ ]]; then
      imp=$(echo "$line" | sed -E 's/.*"(.*)".*/\1/')
      imports+=("$imp")
      continue
    fi

    # Inside import block
    if [[ $in_import_block -eq 1 && $line =~ \" ]]; then
      imp=$(echo "$line" | sed -E 's/.*"(.*)".*/\1/')
      imports+=("$imp")
    fi

  done < "$file"

  # Now check for duplicates
  if [ ${#imports[@]} -gt 0 ]; then
    duplicates=$(printf "%s\n" "${imports[@]}" | sort | uniq -d)
    if [ -n "$duplicates" ]; then
      echo "In file $file:"
      echo "$duplicates" | while read -r dup; do
        echo "  Duplicate import: $dup"
      done
    fi
  fi
done
```

It can probably be added to the CI but I would rather write something non in bash :) easier to test and maintain

backend/accounts.go
bznein: > I think in case of dups, you should keep the named one and remove the unnamed one. The reason for naming these imports is to avoid clashes or generic names. `func foo(coin coin.Coin)` is no good 😅
> Not sure what the idiomatic way is, as I assume renaming imports is not very idiomatic.

Exactly, according to this: https://google.github.io/styleguide/go/decisions#import-renaming imports shouldn't be renamed unless there are good reasons

> The reason for naming these imports is to avoid clashes or generic names. `func foo(coin coin.Coin)` is no good 😅

I don't think this is solved by renaming the package import:
`func foo (coin coinpkg.Coin)` doesn't look much butter imho. In general having a type `Coin` in a package `coin` will inevitably leads us to situations like these.

benma: > `func foo (coin coinpkg.Coin)` doesn't look much better imho

I should have been more specific. It's not the optics, it's about the name shadowing. Inside the `foo` function you can't use the `coin` package anymore b/c it's shadowed by the `coin` argument. It's a bit silly that vars and packages live in the same namespace in Go, but since it is, we should not shadow.

bznein: I'd personally argue that it is better to rename the argument rather than the package (smaller scope, and it's easier to find a more meaningful name for an arg rather than a pkg) but I will not fight over it :D I'll make the changes!

bznein (commented):
@benma this is ready for review again. I changed it so that in case of duplicate imports, the unnamed one is removed and  the aliased one is kept.

Let me know if you want me to write a proper linter/script that detects duplicate imports; I have some free cycles to work on it but I'd prefer to have it in a proper programming language :)

backend/accounts.go
```diff
@@ -969,7 +967,7 @@ type scriptTypeWithKeypath struct {
 	keypath    signing.AbsoluteKeypath
 }
 
-// adds a combined BTC account with the given script types.
+// adds a combined BTC account with the given script accountsTypes.
```
benma, line 970: wrong search&replace

benma (changes requested):

backend/accounts.go
benma: It's not a fight :smile: I think your point makes sense and I'm okay with it too.

backend/accounts.go
bznein: whoops!

bznein: Fixed together with a few others, thanks for catching it!

benma (approved):

PR #3307, opened 2025-04-24
ble: Add support for larger ble packets

NickeZ:
This lowers the time to flash to 1m45s

~~Needs a new ble firmware that handles up to 320 bytes per ble write and also a new `bitbox02-api-go` that attempts to write more packets simultaneously.~~

edit: the required firmware is in the staging/ble branch. This PR has been rebased with new `bitbox02-api-go`.

### DIFF
--- a/frontends/ios/BitBoxApp/BitBoxApp/Bluetooth.swift
+++ b/frontends/ios/BitBoxApp/BitBoxApp/Bluetooth.swift
@@ -91,7 +91,7 @@ class BluetoothManager: NSObject, ObservableObject, CBCentralManagerDelegate, CB
     }
 
     func isConnected() -> Bool {
-        return isPaired && connectedPeripheral != nil && pReader != nil && pWriter != nil;
+        return isPaired && connectedPeripheral != nil && pReader != nil && pWriter != nil
     }
 
     func connect(to peripheralID: UUID) {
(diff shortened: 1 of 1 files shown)


frontends/ios/BitBoxApp/BitBoxApp/Bluetooth.swift
```diff
@@ -448,10 +461,63 @@ class BluetoothDeviceInfo: NSObject, MobileserverGoDeviceInfoInterfaceProtocol {
...
     init(bluetoothManager: BluetoothManager) {
         self.bluetoothManager = bluetoothManager
     }
+    func start() {
+        writer_thread = Thread.init(target: self, selector: #selector(writer), object: nil)
+        writer_thread!.start()
+
+    }
+
+    @objc func writer(_ _: Any) {
```
benma, line 478: Does this ever exit? What if you disconnect/reconnect? A new thread will start but will the old one close?

benma, line 478: why a thread instead of writing it sequentially in `func write()`? isn't the write function already called from some goroutine, so it shouldn't be blocking the UI?

benma (commented):
tACK

benma (approved):

NickeZ:
rebased and manually tested that things still work

PR #3308, opened 2025-04-25
frontend: remove bitbox01 mobile pairing and 2fa

thisconnect:
- removes BitBox01 mobile pairing and 2fa
- fixes in-app guide in the BitBox01 unlock view


frontends/web/src/api/bitbox01.ts
```diff
@@ -22,7 +22,6 @@ export type DeviceInfo = {
   lock: boolean;
   name: string;
   new_hidden_wallet: boolean;
-  pairing: boolean;
```
thisconnect, line 25: With this PR we could remove pairing from the backend as well.

PR #3311, opened 2025-04-28
frontend: don't show the BB02 logo when switching testnet mode

benma:
There is no action to be performed on the BB02.


thisconnect (approved):
tested LGTM

thanks

PR #3312, opened 2025-04-28
frontend: show global banners in the wait screen too

benma:
It's important to see the testnet warning early, otherwise it's very hard to see if one is in testnet mode.

It's also good to see the mobile data warning *before* one connects a device and all the data starts being used.


thisconnect (approved):
untested LGTM

PR #3316, opened 2025-04-28
backend: return transaction's ID from SendTx API.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/accounts/account.go
+++ b/backend/accounts/account.go
@@ -86,9 +86,9 @@ type Interface interface {
 	Transactions() (OrderedTransactions, error)
 	// Must enforce that initial sync is done before returning.
 	Balance() (*Balance, error)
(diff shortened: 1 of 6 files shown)


backend/accounts/account.go
```diff
@@ -88,7 +88,7 @@ type Interface interface {
 	Balance() (*Balance, error)
 	// SendTx signs and sends the active tx proposal, set by TxProposal. Errors if none
 	// available.
-	SendTx(txNote string) error
+	SendTx(txNote string) (string, error)
```
Beerosagos, line 91: Please, update the comment to explain the string return :)

Beerosagos (approved):
tACK with a small comment :)
Please also add a commit message explaining why the change is needed :innocent:

PR #3320, opened 2025-04-30
frontend: fix postmessage error

thisconnect:
This fixes the following error:

Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('bitboxapp:') does not match the recipient window's origin ('https://bitboxapp.shiftcrypto.io')

Posting 'configuration' in handleConfiguration already works and was missing in the two new postMessage's that are used to 'confirm-transaction-id' and 'cancel-order'.


Beerosagos (approved):
utACK!

PR #3321, opened 2025-04-30
eth: migrate etherscan API to v2

Tomasvrba:


backend/backend.go
```diff
@@ -517,19 +517,19 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 		coin = btc.NewCoin(coinpkg.CodeLTC, "Litecoin", "LTC", coinpkg.BtcUnitDefault, &ltc.MainNetParams, dbFolder, servers,
 			"https://blockchair.com/litecoin/transaction/", backend.socksProxy)
 	case code == coinpkg.CodeETH:
-		etherScan := etherscan.NewEtherScan("https://api.etherscan.io/api", backend.etherScanHTTPClient)
+		etherScan := etherscan.NewEtherScan("https://api.etherscan.io/v2/api?chainid=1", backend.etherScanHTTPClient)
```
benma, line 520: I think it's more proper to encode the chainID as a param in the `call()` function and not pass it as part of the url in here. In fact, you can drop the base url param as it's now the same for all. Rather, pass the chainID as an argument.

backend/coins/eth/etherscan/etherscan.go
```diff
@@ -69,7 +69,7 @@ func (etherScan *EtherScan) call(ctx context.Context, params url.Values, result
 		return errp.WithStack(err)
 	}
 	params.Set("apikey", apiKey)
-	response, err := etherScan.httpClient.Get(etherScan.url + "?" + params.Encode())
+	response, err := etherScan.httpClient.Get(etherScan.url + "&" + params.Encode())
```
benma, line 72: ... then this change is not necessary anymore.

benma (changes requested):

backend/backend.go
Tomasvrba: ahahah, I knew you'd want to do it that way :D

benma: :joy: It's not the first time you say this either :joy:

Tomasvrba: I changed it a few days ago btw

benma (approved):
tACK, thank you! I updated the commit to fix the CHANGELOG merge conflict and to put the line into a separate v4.47.3 release section.

PR #3322, opened 2025-05-02
refactor:replacing append([]T(nil), s...) by slices.Clone(s) added in go1.21

findnature:
Optimize code using a more modern writing style.   Official support from Go, for more details visit https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize.


benma (approved):
Thank you!

PR #3325, opened 2025-05-06
backend/btc/address: be explicit about chainIndex/addressIndex

benma:
The relative keypath always has two elements, the first element indicating change and the second indicating the address index.

A relative keypath of arbitrary length is harder to handle. E.g. it is not compatible with output descriptors with multipath indices (`xpub/<0;1>/*`), where also two elements are required to derive an address (the multipath index selecting between `<X;Y>` and the address index).

The unit tests change because the GetAddress test function used an empty relative path instead of a chainIndex and addressIndex, which now was forced to change with the stricter types.

I considered keeping `ChainIndex uint32` instead of `Change bool`,
because in multipaths, one could have more than two,
e.g. `xpub/<X;Y;Z>/*`, but I prefer the stricter type and it's not
likely we will ever need more than two (receive & change).

### DIFF
--- a/backend/coins/btc/addresses/address.go
+++ b/backend/coins/btc/addresses/address.go
@@ -49,20 +49,25 @@ type AccountAddress struct {
 // NewAccountAddress creates a new account address.
 func NewAccountAddress(
 	accountConfiguration *signing.Configuration,
-	keyPath signing.RelativeKeypath,
+	derivation types.Derivation,
 	net *chaincfg.Params,
 	log *logrus.Entry,
 ) *AccountAddress {
 
 	var address btcutil.Address
 	var redeemScript []byte
-	configuration, err := accountConfiguration.Derive(keyPath)
+	configuration, err := accountConfiguration.Derive(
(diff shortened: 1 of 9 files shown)


backend/coins/btc/addresses/address.go
```diff
@@ -47,22 +47,28 @@ type AccountAddress struct {
 }
 
 // NewAccountAddress creates a new account address.
+// The chainIndex is 0 for receive and 1 for change.
```
bznein, line 50: It's not clear what "chainIndex" is here; it's not one of the argument (though I assume you refer to the one accessed via `derivation.SimpleChainIndex()`? ) maybe we can make it clearer

bznein (commented):
Overall LGTM with one minor comment on docs

backend/coins/btc/addresses/address.go
benma: Ah it's a leftover comment from before I introduced the Derivation struct. Removed.

bznein (approved):
LGTM

PR #3329, opened 2025-05-07
backend: add eth account to software keystore.

bznein:
With this addition, we can use ETH accounts in a software keystore. As shown in the screenshot, accounts can be added, and they can correctly receive/send/send-to-self:

image

- [ ] updating the Changelog
- [ ] writing unit tests


benma (approved):
tACK

PR #3330, opened 2025-05-07
backend/account: fix bug in account discovery

Beerosagos:
Recent commit 726289c7eedbd5157cf5f514cca4e4ce2cc4e9b3 made the account Initialization asynchronous, moving the `account.ensureAddresses()` call on a separate thread. This caused a regression inside `backend.checkAccountUsed()`, where the method wasn't waiting for the complete account synchronization before marking it as used.

This commit fixes the issue waiting for account sync before checking if it used.

Note: The tests didn't catch the regression: I couldn't find an easy way to add some latency in the mocked addresses verification, but it would be nice to introduce it in a test to make it more reliable. Any suggestions?

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -750,12 +750,16 @@ func (backend *Backend) addAccount(account accounts.Interface) {
 		})
 		if event.Subject == string(accountsTypes.EventSyncDone) {
 			backend.notifyNewTxs(account)
+			go backend.checkAccountUsed(account)
 		}
 	})
+	if err := account.Initialize(); err != nil {
+		backend.log.WithError(err).Error("error initializing account")
+		return
+	}
 	if backend.onAccountInit != nil {
 		backend.onAccountInit(account)
 	}
-	go backend.checkAccountUsed(account)
 }
 
 // The accountsAndKeystoreLock must be held when calling this function.
@@ -1529,10 +1533,6 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 		}
 	}
 	log := backend.log.WithField("accountCode", account.Config().Config.Code)
-	if err := account.Initialize(); err != nil {
-		log.WithError(err).Error("error initializing account")
-		return
-	}
 	txs, err := account.Transactions()
 	if err != nil {
 		log.WithError(err).Error("discoverAccount")
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


backend/accounts.go
```diff
@@ -1533,6 +1533,9 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 		log.WithError(err).Error("error initializing account")
 		return
 	}
+	for !account.Synced() {
+		time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
Beerosagos: I agree is not very elegant, but why do you think it is not robust?

Beerosagos:
@benma updated, PTAL.

While working on this I also noticed a weird bug that was sometimes showing a wrong account list in the sidebar: after some debugging I found the cause in `backend.Accounts`.

backend/accounts.go
```diff
@@ -1533,45 +1533,53 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
...
 		}
-		if accountsConfig.IsKeystoreWatchonly(rootFingerprint) {
-			t := true
-			acct.Watch = &t
+		backend.emitAccountsStatusChanged()
+		backend.maybeAddHiddenUnusedAccounts()
+	}
+	account.Observe(func(event observable.Event) {
+		if event.Subject == string(accountsTypes.EventSyncDone) {
+			f()
```
benma, line 1580: There is already an `account.Observe()`  in `addAccount()`, imho it's better to just move the `go backend.checkAccountUsed(account)` there instead of having this big local `f := func() ... `  here. Wdyt?

Technically it's not guaranteed that the SyncDone event will be fired after we start to observe it :see_no_evil: But  that's probably a super-edge case that we don't need to worry about :smile:  

Now it also occurs to me that this PR probably also fixes the bug if you receive coins on the last account, you can't add another one until you re-open the app and it realizes the previous account was used. Now it will re-check once coins arrive :+1:

Beerosagos: > There is already an account.Observe() in addAccount(), imho it's better to just move the go backend.checkAccountUsed(account) there instead of having this big local f := func() ... here. Wdyt?

I tested that, but it waits until the sync is done before scanning the next account: the whole discovery is much slower.

> Now it also occurs to me that this PR probably also fixes the bug if you receive coins on the last account, you can't add another one until you re-open the app and it realizes the previous account was used. Now it will re-check once coins arrive 👍

Exactly :D

> Technically it's not guaranteed that the SyncDone event will be fired after we start to observe it 🙈 But that's probably a super-edge case that we don't need to worry about 😄

I could check if the account is already synced before setting the observer, wdyt?

backend/backend.go
```diff
@@ -590,7 +590,9 @@ func (backend *Backend) Testing() bool {
 // Accounts returns the current accounts of the backend.
 func (backend *Backend) Accounts() AccountsList {
 	defer backend.accountsAndKeystoreLock.RLock()()
-	return backend.accounts
+	accounts := AccountsList{}
+	accounts = append(accounts, backend.accounts...)
+	return accounts
```
benma, line 595: Use `slices.Clone()` instead, which I recently learned about here https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3322 :P

benma (commented):
utACK, but see suggestion to use `slices.Clone()`

benma (approved):

PR #3335, opened 2025-05-08
Continue with BTC address derivation using account config + derivation info

benma:


backend/keystore/keystore.go
```diff
@@ -92,9 +93,16 @@ type Keystore interface {
...
 	CanVerifyAddress(coin.Coin) (secureOutput bool, optional bool, err error)
 
-	// VerifyAddress outputs the public key at the given configuration for the given coin.
+	// VerifyAddressBTC displays a Bitcoin address for verification.
 	// Please note that this is only supported if the keystore has a secure output channel.
-	VerifyAddress(*signing.Configuration, coin.Coin) error
+	VerifyAddressBTC(
+		accountConfiguration *signing.Configuration,
+		derivation btctypes.Derivation,
+		coin coin.Coin) error
```
bznein, line 101: is there a specific reason we don't directly use `btc.Coin` and `eth.Coin` (below) so we don't have to cast it inside the method?

bznein (commented):
Overall LGTM just a couple of questions

backend/coins/btc/addresses/addresschain_test.go
```diff
@@ -142,8 +142,9 @@ func (s *addressChainTestSuite) TestEnsureAddresses() {
 	}
 	s.Require().Len(newAddresses, s.gapLimit)
 	for index, address := range newAddresses {
-		s.Require().Equal(uint32(index), address.Configuration.AbsoluteKeypath().ToUInt32()[1])
-		s.Require().Equal(getPubKey(index), address.Configuration.PublicKey())
+		s.Require().Equal(uint32(index), address.AbsoluteKeypath().ToUInt32()[1])
+		_ = getPubKey
```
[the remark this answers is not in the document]
benma: Thanks, yeah. Removed

backend/keystore/keystore.go
benma: Because of a circular import between the packages :sweat_smile: Maybe we can think another time of if and how to break the cycle.

PR #3337, opened 2025-05-09
frontend: show all sell options

thisconnect:


backend/coins/btc/handlers/handlers.go
```diff
@@ -875,7 +875,7 @@ func (handlers *Handlers) getHasPaymentRequest(r *http.Request) (interface{}, er
 	if !ok {
 		return response{
 			Success:      false,
-			ErrorMessage: "An account must be BTC based to support payment requests.",
+			ErrorMessage: "",
```
thisconnect, line 878: Temporary removed this error, but instead the app should probably just return BTC Direct as only sell option (now that it supports coins other than Bitcoin).

Beerosagos: Yeah, I agree :+1:

thisconnect: > Unless I'm mistaken there's not much we can do in the backend here, as the backend already returns the correct deals, i

oh, I was somehow under the assumption that the backend doesn't retrurn the correct deals. I'll check again…

PR #3340, opened 2025-05-09
qt: persist local storage

benma:
In Qt 5 this was on by default, but since our upgrade to Qt 6 this broke, as in Qt 6 one has to explicitly set the persistent storage path.


thisconnect (approved):
LGTM thank you

PR #3341, opened 2025-05-12
ios: handle version with and without `v` prefix

benma:
Before https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1447/, the product characteristic version string didn't include a 'v' prefix, but after it does.

The app backend requires it to parse the version.

We handle both cases for backwards compatibility.


NickeZ (approved):
lgtm

PR #3342, opened 2025-05-12
ios: set minimum-scale, and maximum-scale to disable zoom

shonsirsha:
setting intial-scale=1 is not enough to disable zooming on iOS app.


thisconnect:
Tested but does not affect behavior. BitBoapp on Android doesn't zoom in/out also in earlier versions.

PR #3343, opened 2025-05-12
Add initial Bitrefill widget to staging 

sutterseba:
Adds an initial proof of concept integration of the Bitrefill widget.

Still missing:

* Disclaimers (currently set to BTC direct instead)
* Some smaller UI elements

The long loading screen inside the widget on checkout is known and related to CSP settings of the captcha that the app tries to load.


frontends/web/src/locales/en/app.json
```diff
@@ -397,6 +397,12 @@
     "upgrade": "Upgrade"
   },
   "buy": {
+    "bitrefill": {
+      "error": {
+        "insufficientFunds": "Insufficient funds to buy gift card."
+      },
+      "transactionNote": "Sent to Bitrefill"
+    },
```
thisconnect, line 405: This file grew organically, since we started with buy and added sell later, we now have keys like: `buy.exchange.sell` 😅 

Maybe it would make sense to make a new top level "market" or "marketplace"  key? but buy works as well no strong opinion.

frontends/web/src/routes/exchange/bitrefill.tsx
```diff
@@ -0,0 +1,251 @@
...
+  code: AccountCode;
+};
+
+const getURLOrigin = (uri: string): string | null => {
+  try {
+    return new URL(uri).origin;
+  } catch (e) {
+    return null;
+  }
+};
```
thisconnect, line 66: nit: we have same (similar?) function in btcdirect, this could be shared in a new  `utils/url.ts or similar.

frontends/web/src/routes/exchange/bitrefill.tsx
```diff
@@ -0,0 +1,251 @@
...
+    case 'request-configuration': {
+      event.source?.postMessage({
+        event: 'configuration',
+        ref: bitrefillInfo.ref,
+        utm_source: 'BITBOX',
+        theme: isDarkMode ? 'dark' : 'light',
+        hl: i18n.resolvedLanguage ? localeMapping[i18n.resolvedLanguage] : 'en',
+        paymentMethods: account.coinCode ? coinMapping[account.coinCode] : 'bitcoin',
+        refundAddress: bitrefillInfo.address,
+        // Option to keep pending payment information longer in session, defaults to 'false'
```
thisconnect, line 134: could you add a link to their documentation somewhere (in this file or in commit message)?

frontends/web/src/routes/exchange/bitrefill.tsx
```diff
@@ -0,0 +1,251 @@
...
+        useHighestFee: true,
+        sendAll: 'no',
+        selectedUTXOs: [],
+        paymentRequest: null
+      };
+
+      let result = await proposeTx(code, txInput);
+      if (result.success) {
+        let txNote = `${t('buy.bitrefill.transactionNote')} (${String(invoiceId)})`;
+        const sendResult = await sendTx(code, txNote);
```
thisconnect, line 175: nice to pass invoiceId to tx note, seems useful for users that buy many vouchers.

nit: please use const if txNote variable is expected to not be re-assigned, otherwise `let`.

frontends/web/src/routes/exchange/exchange.tsx
```diff
@@ -134,7 +134,13 @@ export const Exchange = ({ code, accounts, deviceIDs }: TProps) => {
...
         </Dialog>
         <div className="innerContainer scrollableContainer">
-          <Header title={<h2>{title}</h2>} />
+          { activeTab === 'spend' ? (
+            <Header title={<h2>{t('generic.spend', {
+              context: hasOnlyBTCAccounts ? 'bitcoin' : 'crypto'
+            })}</h2>}/>
+          ) : (
+            <Header title={<h2>{title}</h2>}/>
+          )}
```
thisconnect, line 143: nit: an alternative could be to use 1 `<Header ` element and move the condition into title attribute.

thisconnect (approved):
Very nice LGTM, added some small comments

 also @Beerosagos please review if you have a moment.

frontends/web/src/api/exchanges.ts
```diff
@@ -21,7 +21,7 @@ export const getExchangeRegionCodes = (): Promise<string[]> => {
   return apiGet('exchange/region-codes');
 };
 
-export type TPaymentMethod = 'card' | 'bank-transfer' | 'bancontact' | 'sofort';
+export type TPaymentMethod = 'card' | 'bank-transfer' | 'bancontact' | 'sofort' | 'spend';
```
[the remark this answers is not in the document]
sutterseba: Yeah it's a bit ugly. I'm basically abusing the payment method field to pass a custom message (which depends on the provider). Strictly speaking "spend" is the payment method here, i.e. the wallet itself pays.

frontends/web/src/locales/en/app.json
sutterseba: Definitely, I haven't spent much time on lang keys yet, partly for this reason. Probably a larger refactor then with more structured buy/sell/spend categorization

thisconnect: it's very annoying to move keys onces they are translated. So if you want to change it would be better to move the new keys before translators start translating the new keys.

sutterseba:
@thisconnect addressed comments and also changed to the same tx note format now (as in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3351)

backend/exchanges/bitrefill.go
```diff
@@ -0,0 +1,101 @@
...
+// for "unspecified") is supported by Bitrefill.
+func IsRegionSupportedBitrefill(region string) bool {
+	return len(region) == 0 || slices.Contains(bitrefillRegions, region)
+}
+
+// IsBitrefillSupported is true if coin.Code is supported by Bitrefill.
+func IsBitrefillSupported(coinCode coin.Code) bool {
+	supportedCoins := []coin.Code{
+		coin.CodeBTC, coin.CodeTBTC, coin.CodeLTC, coin.CodeTLTC, coin.CodeETH, coin.CodeSEPETH,
+		"eth-erc20-usdc", "eth-erc20-usdt"}
```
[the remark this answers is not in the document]
thisconnect: I see, but maybe better if we test with `make servewallet-mainnet`?

thisconnect (approved):
very nice, initial Bitrefill pr LGTM

Lets get this merged to staging-bitrefill soon, you've already over-delivered and it looks like almost done and more than just a POC :)

ping @Beerosagos ptal

frontends/web/src/routes/exchange/bitrefill.tsx
```diff
@@ -0,0 +1,235 @@
...
+        sendAll: 'no',
+        selectedUTXOs: [],
+        paymentRequest: null
+      };
+
+      let result = await proposeTx(code, txInput);
+      if (result.success) {
+        const txNote = t('generic.paymentRequestNote', {
+          name: 'Bitrefill',
+          invoiceId,
```
thisconnect, line 157: oups in the other PR the placeholder is called orderId

backend/exchanges/bitrefill.go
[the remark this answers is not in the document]
sutterseba: Already removed it

frontends/web/src/routes/exchange/bitrefill.tsx
sutterseba: Updated

Beerosagos:
> ping @Beerosagos ptal

I already saw the first mention :joy: I'll have a look asap

thisconnect:
> I already saw the first mention 😂 I'll have a look asap

I didn't want to rush you, but wanted to hand it over to you. 😇

frontends/web/src/routes/exchange/bitrefill.tsx
```diff
@@ -0,0 +1,235 @@
...
+        address: paymentAddress,
+        amount: parsedAmount.amount,
+        // Always use highest fee rate for Bitrefill spend
+        useHighestFee: true,
+        sendAll: 'no',
+        selectedUTXOs: [],
+        paymentRequest: null
+      };
+
+      let result = await proposeTx(code, txInput);
```
thisconnect, line 153: tiny nit: by using `const` it would tell that this variable will not be re-assigned later.

frontends/web/src/routes/exchange/bitrefill.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const handleMessage = useCallback(async (event: MessageEvent) => {
+    if (
+      !account
+      || !bitrefillInfo?.success
+      || ![getURLOrigin(bitrefillInfo.url), 'https://embed.bitrefill.com'].includes(event.origin)
+    ) {
+      return;
+    }
+
+    const data = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
```
thisconnect, line 103: JSON.parse could throw an error if `event.data` is not valid JSON, this probably never happens, however it is coming from outside the BitBoxApp so it would not be wrong to wrap in try/catch

thisconnect (approved):
frontend LGTM, added a tiny nit

Beerosagos (approved):
I tested some changes in the code and I realized that the `Deal` model is deeply exploited in the FE implementation and that makes quite hard to move Bitrefill outside of it in a nice way. It probably doesn't make much sense from the effort pov.

I think that we could merge this and refactor later some part of the model, instead, to make it more flexible (e.g. not only valid for exchanges) so that it can fit Bitrefill and other future integrations. 

LGTM, great work @sutterseba :fire:

PR #3345, opened 2025-05-13

benma:
Compiled from
https://github.com/BitBoxSwiss/bitbox02-firmware/commit/2544f0889b9c7c5aba3b6bc2307efe443ce119dd.

BLE fw binary also taken from there.

We only update the multi firmware as all our testing devies are multis. We upgrade ignoring errors as it is unsigned (we prepend the v9.22 sigs just to have the right format).


NickeZ (approved):
Checked sha256 of the ble binary.

LGTM

PR #3348, opened 2025-05-14
frontend: move firmware required logic into pocket view

thisconnect:
Pocket Bitcoin was the only sell option before and requires a newer firmware. The sell tab in buy/sell just displayed a require-fw-update message instead of showing the sell option (Pocket).

Moved the required-firmware-update into pocket component so the app can shows all options.


bznein (approved):
utACK

PR #3350, opened 2025-05-15

benma:
Compiled from https://github.com/BitBoxSwiss/bitbox02-firmware/commit/3b387d8dce1cfa4c811e67b96b79599483a887b1

BLE fw binary also taken from there.

We only update the multi firmware as all our testing devies are multis. We upgrade ignoring errors as it is unsigned (we prepend the v9.22 sigs just to have the right format).


NickeZ (approved):
utack!

PR #3356, opened 2025-05-20
android: make cmake version a minimum

benma:
I could not figure out how to install 3.31.6 in my Android Studio, but later versions also work fine.


bznein (approved):
utACK

PR #3359, opened 2025-05-21
frontend: block ui during address verification

thisconnect:
Block the UI when there is a signing request so that the user cannot navigate when there is a signing request on the device.


Beerosagos (approved):
utACK

PR #3361, opened 2025-05-21
frontend: block ui in case of sign request

thisconnect:
The bitbox02 only can handle one request at a time, this means that an ongoing message signing request blocks the device until it is finished.

Currently, it is possible to request a message signing inside the pocket workflow and leave the workflow while the signing request is ongoing. If the user tried to make other operations, e.g. verifying an address or opening the manage device screen, each bb02 request remains pending.

By blocking everything but the iframe the user is forced to stay in the workflow and has to either complete or abort the signing.

When blocking the view should always surpress native back, by using

    <UseDisableBackButton />

### DIFF
--- a/frontends/web/src/routes/exchange/iframe.module.css
+++ b/frontends/web/src/routes/exchange/iframe.module.css
@@ -23,3 +23,18 @@
     position: relative;
     z-index: 2200;
 }
+
+.blocking {
+    background-color: var(--bg-transparent-dark);
+    bottom: 0;
+    left: 0;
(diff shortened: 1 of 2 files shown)


thisconnect:
Tested that the app navigation is darkened and not clickable with Qt on macOS and on Android.

thisconnect:
fyi, opened a draft PR to block the app ui in staging-btcdirect-sell https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3359

PR #3362, opened 2025-05-22

benma:
Compiled from https://github.com/BitBoxSwiss/bitbox02-firmware/commit/e3149e2566c796fb42c93c9b76f736f3d09551f0


NickeZ (approved):
utack

PR #3363, opened 2025-05-23
frontend: close sidebar on mobile when rotating to portrait

thisconnect:
Fixed an issue where the sidebar remains open after toggling via
the hamburger menu and rotating the screen.

Now, the sidebar is always closed on mobile devices when switching
to portrait orientation.


thisconnect:
tested on Android

PR #3367, opened 2025-05-26
ios: fix tight margins on fullscreen View

shonsirsha:
We previously added space but only the top.
Added bottom spacing now as well.

Regarding the "X" button was because it's an absolute posiitoned
item, so it didn't respect the top padding.  Thus added extra
space for its `top` property here.


benma (changes requested):
Unfortunately this did not fix it for me, the `X` is still in the same location next to the clock at the top and not tappable.

PR #3371, opened 2025-05-26
backend: don't require pairing code over BLE

benma:
Connecting to a BitBox02 over BLE already requires pairing, so the noise pairing is redundant. This does not prevent confirming the pairing code at any time after, it just means the confirmation is optional.

```
go get github.com/BitBoxSwiss/bitbox02-api-go@4c21ece5b08182e2946362303347bc4b50906e46
go mod tidy
go mod vendor
```


NickeZ (approved):
tACK (before PR was opened, I assume nothing changed)

PR #3373, opened 2025-05-27
stop scanning peripherals when the device picker is not mounted

benma:


NickeZ:
Should the frontend really stop the scan? shouldn't it stop by itself after a fixed amount of time?

PR #3376, opened 2025-05-28
backend/coins: refactor formatted amounts and add unit tests

Beerosagos:
This refactors the code to format amounts with fiat conversions, removing it from handlers into the coin package, and adds some tests.

### DIFF
--- a/backend/coins/btc/util/util.go
+++ b/backend/coins/btc/util/util.go
@@ -18,7 +18,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/ltc"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
 	"github.com/btcsuite/btcd/btcutil"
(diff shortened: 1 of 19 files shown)


backend/coins/btc/util/util.go
```diff
@@ -69,8 +68,3 @@ func AddressFromPkScript(pkScript []byte, net *chaincfg.Params) (btcutil.Address
 	}
 	return addresses[0], nil
 }
-
-// FormatBtcAsSat returns true if the btcUnit param is `[t]sat`.
```
Beerosagos, line 73: Removed, not needed anymore

backend/coins/coin/amount.go
```diff
@@ -76,6 +79,64 @@ func (amount Amount) BigInt() *big.Int {
 	return new(big.Int).Set(amount.n)
 }
 
+// FormattedAmountWithConversions and json tags.
+type FormattedAmountWithConversions struct {
```
benma, line 83: Maybe it would make sense to rename the frontend type from IAmount to IFormattedAmountWithConversions for consistency.

benma (approved):
utACK

backend/coins/coin/amount.go
Beerosagos: Removed the "Formatted" bit to make it shorter, since it is not really relevant in the frontend.

PR #3378, opened 2025-06-02
ios: disable landscape orientations

shonsirsha:
In the meantime, we'd like to disable landscape mode on iPhone and iPad devices due to UI inconsistencies in that
orientation.


benma (approved):
tACK

PR #3383, opened 2025-06-04
ios: show connection issues tips after 30s bluetooth scanning

shonsirsha:
After 30s of bluetooth scanning (scanning: true), we'll show a Link under the spinner in Bluetooth.tsx.

When clicked, this  will open a dialog.  This dialog contains connection issues tips for the user.

### DIFF
--- a/frontends/web/src/components/bluetooth/bluetooth.tsx
+++ b/frontends/web/src/components/bluetooth/bluetooth.tsx
@@ -15,6 +15,7 @@
  */
 
 import { useTranslation } from 'react-i18next';
+import { useEffect, useRef, useState } from 'react';
 import { useSync } from '@/hooks/api';
 import { connect, getState, syncState, TPeripheral } from '@/api/bluetooth';
 import { runningInIOS } from '@/utils/env';
@@ -22,6 +23,8 @@ import { Status } from '@/components/status/status';
 import { ActionableItem } from '@/components/actionable-item/actionable-item';
 import { Badge } from '@/components/badge/badge';
 import { HorizontallyCenteredSpinner, SpinnerRingAnimated } from '@/components/spinner/SpinnerAnimation';
+import { Button } from '@/components/forms';
+import { ConnectionIssuesDialog } from '@/components/bluetooth/connection-issues-dialog';
 import styles from './bluetooth.module.css';
 
 const isConnectedOrConnecting = (peripheral: TPeripheral) => {
@@ -31,6 +34,33 @@ const isConnectedOrConnecting = (peripheral: TPeripheral) => {
 const _Bluetooth = () => {
   const { t } = useTranslation();
   const state = useSync(getState, syncState);
+  const [showConnectionIssues, setShowConnectionIssues] = useState(false);
+  const [dialogOpen, setDialogOpen] = useState(false);
+  const scanningTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const TIMEOUT_MS = 30000;
+
+  useEffect(() => {
+    if (state?.scanning) {
+      scanningTimeoutRef.current = setTimeout(() => {
+        if (state.scanning && state.peripherals.length === 0) {
+          setShowConnectionIssues(true);
+        }
+      }, TIMEOUT_MS);
+    } else {
+      if (scanningTimeoutRef.current) {
+        clearTimeout(scanningTimeoutRef.current);
+        scanningTimeoutRef.current = null;
+      }
+      setShowConnectionIssues(false);
+    }
+
+    return () => {
+      if (scanningTimeoutRef.current) {
+        clearTimeout(scanningTimeoutRef.current);
+      }
+    };
+  }, [state?.scanning, state?.peripherals.length]);
+
   if (!state) {
     return null;
   }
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/device/waiting.tsx
```diff
@@ -89,7 +89,7 @@ export const Waiting = () => {
         <Entry entry={{
           link: {
             text: t('guide.waiting.getDevice.link.text'),
-            url: 'https://bitbox.shop/',
+            url: 'https://shop.bitbox.swiss/',
```
shonsirsha: Not exactly related but might as well change it here since it's small enough.

frontends/web/src/components/bluetooth/bluetooth.tsx
```diff
@@ -79,8 +107,25 @@ const _Bluetooth = () => {
...
+            <div className={styles.connectionIssuesLink}>
+              <ButtonLink
+                to="#"
+                transparent
+                onClick={(e) => {
+                  e.preventDefault();
+                  setDialogOpen(true);
+                }}
+              >
+                {t('novaConnectionIssues.havingConnectionIssues')}
```
benma: Prefix with `bluetooth.connectionIssues` would be a better namespace right? As it's in bluetooth.tsx, and it's not related to nova per se, but to bluetooth.

frontends/web/src/components/bluetooth/bluetooth.tsx
```diff
@@ -31,6 +34,31 @@ const isConnectedOrConnecting = (peripheral: TPeripheral) => {
 const _Bluetooth = () => {
   const { t } = useTranslation();
   const state = useSync(getState, syncState);
+  const [showConnectionIssues, setShowConnectionIssues] = useState(false);
+  const [dialogOpen, setDialogOpen] = useState(false);
+  const scanningTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const TIMEOUT_MS = 30000;
+
+  useEffect(() => {
+    if (state?.scanning) {
```
benma, line 43: If it's scanning and there are peripherals, but the user has not clicked on any yet, there are no connection issues, user just hasn't tapped on any yet.

frontends/web/src/components/bluetooth/bluetooth.tsx
shonsirsha: Fair point ty!

frontends/web/src/components/bluetooth/bluetooth.tsx
```diff
@@ -31,6 +34,33 @@ const isConnectedOrConnecting = (peripheral: TPeripheral) => {
...
+      scanningTimeoutRef.current = setTimeout(() => {
```
benma, line 44: If there is already a timeout going on, then this would create a new one and overwrite the ref without clearing the old one. I guess just check if the timeout ref is not null before creating a new timeout.

benma: On second look, the clearTimeout in the return value probably takes care of this.

frontends/web/src/locales/en/app.json
```diff
@@ -1363,6 +1364,20 @@
     "newTxs_one": "New transaction in: {{accountName}}",
     "newTxs_other": "{{count}} new transactions in: {{accountName}}"
   },
+  "novaConnectionIssues": {
+    "connectionIssueTips": "Connection issue tips",
+    "description1": "Only the BitBox02 Nova is compatible with iOS. For the BitBox02, use Windows, macOS, Android or Linux. You can purchase a BitBox02 Nova in our online shop:",
+    "description2": "The screen should light up. If the screen is not on, then the device is not receiving power.",
+    "description3": "Keep the BitBox02 Nova close to your iPhone/iPad. We recommend you plug the BitBox02 Nova directly into your device.",
+    "description4": "The BitBox02 Nova connects with iOS via bluetooth. Bluetooth can be disabled/enabled from the BitBoxApp settings on other platforms.",
```
benma: upper case Bluetooth

frontends/web/src/locales/en/app.json
```diff
@@ -1363,6 +1364,20 @@
...
+    "link1": "Order a BitBox02 Nova",
+    "link2": "Contact support",
+    "stillHavingIssues": "Still having issues?",
+    "title1": "Only use the BitBox02 Nova with iOS",
+    "title2": "Make sure the BitBox02 Nova is powered",
+    "title3": "Keep BitBox02 Nova in range",
+    "title4": "Make sure bluetooth is not deactivated on the BitBox02 Nova"
```
benma: uppercase Bluetooth

benma (changes requested):

frontends/web/src/locales/en/app.json
benma: Fixed and merged to unblock translators. Thanks!

frontends/web/src/components/bluetooth/bluetooth.tsx
benma: This seems to have been missed :( It's still named `novaConnectionIssues` not `bluetooth.connectionIssues`

shonsirsha: Oh .. 🤔 looking at current master, this has been (correctly) changed to bluetooth.connectionIssues though?

The novaConnectionIssues still exists, but those translation keys only concern BB02 Nova (the content always mentions BB02 Nova). Thus I think the naming is fine in this case.

PR #3384, opened 2025-06-04
backend/accounts: refactor accounts summary endpoint and functions

Beerosagos:
The code needed to load the accounts summary table was complicated and fragmented. This refactors some parts of the code, unifying some endpoints and adding unit tests.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -268,19 +268,95 @@ func (backend *Backend) accountFiatBalance(account accounts.Interface, fiat stri
 	return fiatValue, nil
 }
 
-// AccountsTotalBalanceByKeystore returns a map of accounts' total balances across coins, grouped by keystore.
-func (backend *Backend) AccountsTotalBalanceByKeystore() (map[string]KeystoreTotalAmount, error) {
-	totalAmounts := make(map[string]KeystoreTotalAmount)
-	fiat := backend.Config().AppConfig().Backend.MainFiat
+type coinFormattedAmount struct {
+	CoinCode        coinpkg.Code                           `json:"coinCode"`
+	CoinName        string                                 `json:"coinName"`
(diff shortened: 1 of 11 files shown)


backend/accounts.go
```diff
@@ -291,18 +367,74 @@ func (backend *Backend) AccountsTotalBalanceByKeystore() (map[string]KeystoreTot
...
+	keystoresBalance, err := backend.keystoresBalance()
+	if err != nil {
+		return nil, err
+	}
+	coinsTotalBalance, err := backend.coinsTotalBalance()
+	if err != nil {
+		return nil, err
+	}
+
+	return &AccountsBalanceSummary{
```
Beerosagos, line 434: Not sure if it would be better to return  an empty object instead of nil in case of errors

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -28,7 +28,7 @@ import { Status } from '@/components/status/status';
 import { GuideWrapper, GuidedContent, Header, Main } from '@/components/layout';
 import { View } from '@/components/view/view';
 import { Chart } from './chart';
-import { SummaryBalance } from './summarybalance';
+import { SummaryBalance as KeystoreBalance } from './summarybalance';
```
benma: please rename summarybalance to keystorebalance and SummaryBalance to KeystoreBalance for clarity.

benma (changes requested):
Very nice refactor :pray:

frontends/web/src/routes/account/summary/accountssummary.tsx
Beerosagos: Uh nice catch, thanks! Fixed, PTAL :pray:

PR #3396, opened 2025-06-11
backend/bitbox02: bundle v9.22.0 firmwares

benma:


NickeZ (approved):
ACK, confirmed the hashes by building on my machine

PR #3398, opened 2025-06-12
frontend: navigate to waiting screen also in /accounts/*

benma:
When there are no accounts (e.g. after unplugging the device), you should navigate back to the waiting screen if you are in the accounts page. The same was true for Manage accounts, but was missed when adding the new accounts page.


frontends/web/src/app.tsx
```diff
@@ -82,6 +82,10 @@ export const App = () => {
       currentURL.startsWith('/account-summary')
       || currentURL.startsWith('/add-account')
       || currentURL.startsWith('/settings/manage-accounts')
+      || currentURL.startsWith('/accounts/')
+      // Workaround on mobile where the bottom menu is not shown when there are no devices/accounts.
+      // If one is on "More" and the bottom menu disappears, one is stuck.
+      || currentURL === 'settings/more'
```
sutterseba:
```suggestion
      || currentURL === '/settings/more'
```

benma: :see_no_evil: fixed

sutterseba (approved):
tACK

PR #3400, opened 2025-06-12
[WIP - Do not review yet] - frontend: actively check for connection and show banner when offline.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests


backend/backend.go
```diff
@@ -299,9 +307,65 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
...
+		Timeout: time.Second,
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout: 3 * time.Second,
+			}).DialContext,
+		},
+	}
+	// Perform connectivity check.
+	resp, err := client.Get("https://clients3.google.com/generate_204")
```
bznein, line 328: This is not ideal as it would not work if Google is down, or blocked by a firewall, or something else.
We should probably have our own endpoint to ping?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -128,6 +128,7 @@ public void onReceive(Context context, Intent intent) {
         @Override
         public void onReceive(Context context, Intent intent) {
             Mobileserver.usingMobileDataChanged();
+            Mobileserver.forceConnectivityCheck();
```
bznein, line 131: This is not necessarily needed... the banner would update anyway after the backend detects the change in online status. However this makes things a bit faster because it triggers almost immediately after a change of status. Open to remove it if it's not worth it.

benma (commented):
Instead of making our own checks, have you considered using these APIs?

- https://doc.qt.io/archives/qt-5.15/qnetworkconfigurationmanager.html#isOnline
- https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback
- https://developer.android.com/reference/android/net/ConnectivityManager#getNetworkCapabilities(android.net.Network)
- https://m-mois.medium.com/how-to-check-internet-availability-in-swift-cae45f52d673

If feasible, they could be wrapped in our own little API in the `backend.Environment` interface.

bznein:
> Instead of making our own checks, have you considered using these APIs?
> 
> * https://doc.qt.io/archives/qt-5.15/qnetworkconfigurationmanager.html#isOnline
> * https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback
> * https://developer.android.com/reference/android/net/ConnectivityManager#getNetworkCapabilities(android.net.Network)
> * https://m-mois.medium.com/how-to-check-internet-availability-in-swift-cae45f52d673
> 
> If feasible, they could be wrapped in our own little API in the `backend.Environment` interface (or if notification based, add a `backend.OnlineStatusChanged(bool)` called from the native apps).

Thanks for the pointer, I misunderstood part of the issue, will work based on this :)

PR #3403, opened 2025-06-12
backend/bitbox02: don't re-use same device instance after reset

benma:
The `device.init()` after reset is causing issues on iOS, where the app sends the attestation request (first action during init) to a device that rebooted and connected, before the orientation screen passed, causing unexpected msgs and the app getting out of sync.


NickeZ (approved):
tACK 👌

PR #3417, opened 2025-06-18
backend/btcdirect: move btc direct sell addresses to the backend

Beerosagos:
so that they are fetched from there instead of being in the frontend logic.

### DIFF
--- a/backend/exchanges/btcdirect.go
+++ b/backend/exchanges/btcdirect.go
@@ -15,6 +15,7 @@
 package exchanges
 
 import (
+	"errors"
 	"slices"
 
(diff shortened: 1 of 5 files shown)


backend/exchanges/btcdirect.go
```diff
@@ -38,10 +39,20 @@ const (
...
+	"LINK": "0xE5b1F760BA4334bc311695e125861Eb5870018aD",
+}
+
 type btcDirectInfo struct {
-	Url     string
-	ApiKey  string
-	Address *string
+	Url      string
+	ApiKey   string
+	Address  string
```
benma: Would be helpful to document these fields.

PR #3418, opened 2025-06-24
Frontend format btcamounts

thisconnect:
In the transaction overview, the amounts are displayed with a space:
0.00 000 000, in the account balance without 0.00000000.

With a space, the amount is much easier to read consistent with other
amounts in the app.

Screenshot 2025-06-24 at 08 29 24
Screenshot 2025-06-24 at 08 29 13
Screenshot 2025-06-24 at 08 29 05

### DIFF
--- a/frontends/web/src/components/amount/amount-with-unit.tsx
+++ b/frontends/web/src/components/amount/amount-with-unit.tsx
@@ -14,21 +14,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import { useContext } from 'react';
-import { CoinUnit, ConversionUnit, TAmountWithConversions } from '@/api/account';
+import type { CoinUnit, ConversionUnit, TAmountWithConversions } from '@/api/account';
 import { RatesContext } from '@/contexts/RatesContext';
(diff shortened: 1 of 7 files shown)


frontends/web/src/components/amount/amount-with-unit.tsx
```diff
@@ -14,29 +14,28 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import { useContext } from 'react';
-import { CoinUnit, ConversionUnit, TAmountWithConversions } from '@/api/account';
+import type { CoinUnit, ConversionUnit, TAmountWithConversions } from '@/api/account';
```
Beerosagos, line 19: Why `type`?

thisconnect: to indicate to TypeScript that this is a Type-Only Import

> import type only imports declarations to be used for type annotations and declarations. It always gets fully erased, so there’s no remnant of it at runtime. Similarly, export type only provides an export that can be used for type contexts, and is also erased from TypeScript’s output.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

PR #3420, opened 2025-06-24
Fix sats

benma:


frontends/web/src/routes/accounts/all-accounts.tsx
```diff
@@ -48,13 +48,14 @@ const AccountItem = ({ account, hideAmounts }: { account: accountApi.IAccount, h
...
         if (!balance.success) {
+          setBalance(undefined);
           return;
         }
         const balanceData = balance.balance;
         if (balanceData.hasAvailable) {
-          setBalance(balanceData.available.amount);
+          setBalance(balanceData.available);
         } else {
-          setBalance('0');
```
benma, line 57: @thisconnect @shonsirsha please note to never show a `0` if a balance is not loaded yet or unavailable. It is bad practice and can induce heart attacks :sweat_smile:

benma: Actually in this instance I think I was wrong, `available` just means there is a non-zero balance. But we don't need to check for this here, `amount` will be `0` in this case already.

benma: Fixed up, PTAL again @thisconnect

thisconnect (approved):
tested LGTM thank you

PR #3421, opened 2025-06-24
frontend/qt: react to change in connectivity status.

bznein:
Instead of relying on a polling mechanism, we use QT's signals to monitor the  connectivity status and react when this changes by showing a banner. When the connectivity is resumed, we trigger a reconnect to our nodes.

- [ ] updating the Changelog
- [ ] writing unit tests


backend/backend.go
```diff
@@ -943,6 +949,33 @@ func (backend *Backend) HandleURI(uri string) {
 	}
 }
 
+// SetOnline sets the backend's online status and notifies the frontend.
+func (backend *Backend) SetOnline(online bool) {
+	backend.log.Infof("Setting online status to %v", online)
+	// If coming back online, trigger a reconnection.
+	if online && !backend.isOnline {
```
benma: I wonder if it would make sense to make backend.isOnline be of type `atomic.Bool`, as I guess this could be called from multiple threads/goroutines?

backend/backend.go
```diff
@@ -943,6 +949,33 @@ func (backend *Backend) HandleURI(uri string) {
...
+		backend.ManualReconnect()
+		for _, account := range backend.Accounts() {
+			ethAccount, ok := account.(*eth.Account)
+			if !ok {
+				continue
+			}
+			ethAccount.EnqueueUpdate()
+		}
```
benma: For maintainability, I think it may be better to move the ETH updates into `ManualReconnect()`, and only process if called with a `force bool` argument.

I imagine if we have more cases where we want to reconnect, or add more coin types in the future, checking `SetOnline()` could easily be missed.

backend/bridgecommon/bridgecommon.go
```diff
@@ -123,6 +123,16 @@ func HandleURI(uri string) {
 	globalBackend.HandleURI(uri)
 }
 
+// OnlineStatusChanged should be called when the network connection changed.
+func OnlineStatusChanged(isOnline bool) {
```
benma: I would just call this `SetOnline` too for consistency, all other globalBackend wrappers use the same name as the wrapped functions too.

backend/coins/eth/account.go
```diff
@@ -1024,3 +1024,8 @@ func (account *Account) MatchesAddress(address string) (bool, error) {
 	}
 	return false, nil
 }
+
+// EnqueueUpdate enqueues an update for the account.
+func (account *Account) EnqueueUpdate() {
+	account.enqueueUpdateCh <- struct{}{}
```
benma: Proably makes sense to wrap this in a select w/ default

```
select {
case account.enqueueUpdatech <- struct{}{}:
default:
}
```

This way if it is called multiple times, they don't all re-trigger updating, but get batched.

`SendTx()` at the bottom could then call this function too.

frontends/qt/network.h
```diff
@@ -0,0 +1,8 @@
+#ifndef NETWORK_H
+#define NETWORK_H
+#include <stdbool.h>
+
+void setupReachabilityNotifier();
+bool isReachable();
```
benma: Some quick docstrings would be nice.

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
+#include <QNetworkInformation>
+#include "libserver.h"
```
benma, line 2: `#include "network.h"` is missing

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void setupReachabilityNotifier() {
+    auto loaded = QNetworkInformation::loadBackendByFeatures(QNetworkInformation::Feature::Reachability);
+    if (!loaded) {
+        return;
```
benma, line 11: Could be useful to `goLog(...)` this condition.

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
...
+    }
+    QNetworkInformation* info = QNetworkInformation::instance();
+    if (info) {
+      QObject::connect(info, &QNetworkInformation::reachabilityChanged, [](QNetworkInformation::Reachability reachability){
+        bool isReachable = (reachability == QNetworkInformation::Reachability::Online);
+        onlineStatusChanged(isReachable);
+      });
```
benma: Indentation off, pls indent by 2 more spaces.

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void setupReachabilityNotifier() {
+    auto loaded = QNetworkInformation::loadBackendByFeatures(QNetworkInformation::Feature::Reachability);
+    if (!loaded) {
+        return;
+    }
+    QNetworkInformation* info = QNetworkInformation::instance();
+    if (info) {
+      QObject::connect(info, &QNetworkInformation::reachabilityChanged, [](QNetworkInformation::Reachability reachability){
+        bool isReachable = (reachability == QNetworkInformation::Reachability::Online);
```
benma: https://doc.qt.io/qt-6.8/qnetworkinformation.html#Reachability-enum

> QNetworkInformation::Reachability::Unknown	0	If this value is returned then we may be connected but the OS has still not confirmed full connectivity, or this feature is not supported.

If this happened, the user might see an offline banner when they are actually online.

Maybe include this with a comment to say that we'd rather not show a offline banner if the user is online than the other way around.

frontends/web/src/contexts/AppContext.tsx
```diff
@@ -25,6 +25,7 @@ type AppContextProps = {
     hideAmounts: boolean;
     isTesting: boolean;
     isDevServers: boolean;
+    isOnline: boolean | undefined;
```
benma: Fyi: this is the same as `isOnline?: boolean`

frontends/web/src/locales/en/app.json
```diff
@@ -1889,6 +1889,7 @@
   },
   "warning": {
     "coincontrol": "One or more UTXOs have a reused address. Be aware the receiver will see all UTXOs associated with those addresses.",
+    "offline": "You are currently offline. You can still view your accounts, but you will not be able to send transactions or see the latest exchange rates.",
```
benma, line 1892: Did you check if this is correct? I thought if offline, the accounts show a full page "connecting..." screen, but it may have changed at some point.

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+bool isReachable() {
+    auto loaded = QNetworkInformation::loadBackendByFeatures(QNetworkInformation::Feature::Reachability);
+    if (!loaded) {
+        return false;
+    }
+    QNetworkInformation* info = QNetworkInformation::instance();
+    if (info) {
+        return info->reachability() == QNetworkInformation::Reachability::Online;
+    } else {
```
benma: else not needed after `return` :nerd_face:

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
...
+        return false;
```
benma: return true? Better to not show a banner if offline than the other way.

frontends/qt/network.cpp
```diff
@@ -0,0 +1,29 @@
...
+        bool isReachable = (reachability == QNetworkInformation::Reachability::Online);
+        onlineStatusChanged(isReachable);
+      });
+    }
+}
+
+bool isReachable() {
+    auto loaded = QNetworkInformation::loadBackendByFeatures(QNetworkInformation::Feature::Reachability);
+    if (!loaded) {
+        return false;
```
benma: Same, return true?

frontends/web/src/locales/en/app.json
bznein: You are right, although if you go offline after loading the accounts, in that case you can still look at them. Maybe we can reword it better, or just omitting this part and saying something else?

backend/backend.go
bznein: Done, however I harcoded a call to `ManualReconnect(false)` in bridgecommon.go. Do you think we should instead change its signature so that we pass the flag directly from the callers?

backend/backend.go
```diff
@@ -549,7 +555,7 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 // ManualReconnect triggers reconnecting to Electrum servers if their connection is down.
 // Only coin connections that were previously established are reconnected.
 // Calling this is a no-op for coins that are already connected.
```
benma, line 557: Please update the docstring, which is now out of date

benma (approved):
tACK with a small comment

frontends/web/src/contexts/AppContext.tsx
thisconnect: almost the same

`isOnline?: boolean` means isOnline can be missing, but if there it is a boolean.

`isOnline: boolean | undefined` means isOnline key is always present but can have boolean or undefined value

PR #3422, opened 2025-06-24
Bluetooth settings section with firmware version

benma:
After this we'll add an item to enable/disable Bluetooth.


frontends/web/src/routes/settings/components/device-settings/bluetooth-firmware-setting.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <>
+      <SettingsItem
+        settingName={t('deviceSettings.firmware.title')}
+        secondaryText={secondaryText}
+        displayedValue={firmwareVersion}
+        extraComponent={extraComponent}
+      />
+    </>
```
thisconnect, line 42: untested LGTM

my only tiny nit was the document fragment is not needed here and in bb02-settings.tsx

but we have that probably in a few places in the codebase, lets just keep as is

benma: I'll change it in the other PR that builds on it :+1:

PR #3423, opened 2025-06-25
Settings toggle bluetooth

benma:
Builds on https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3422

### DIFF
--- a/frontends/web/src/api/bitbox02.ts
+++ b/frontends/web/src/api/bitbox02.ts
@@ -226,3 +226,7 @@ export const attestationCheckDone = (
 ): TUnsubscribe => {
   return subscribeEndpoint(`devices/bitbox02/${deviceID}/attestationCheckDone`, cb);
 };
+
+export const bluetoothToggleEnabled = (deviceID: string): Promise<SuccessResponse | FailResponse> => {
+  return apiPost(`devices/bitbox02/${deviceID}/bluetooth/toggle-enabled`);
+};
--- a/frontends/web/src/routes/settings/bb02-settings.tsx
+++ b/frontends/web/src/routes/settings/bb02-settings.tsx
@@ -31,6 +31,7 @@ import { Skeleton } from '@/components/skeleton/skeleton';
 import { AttestationCheckSetting } from './components/device-settings/attestation-check-setting';
 import { FirmwareSetting } from './components/device-settings/firmware-setting';
 import { BluetoothFirmwareSetting } from './components/device-settings/bluetooth-firmware-setting';
+import { BluetoothToggleEnabledSetting } from './components/device-settings/bluetooth-toggle-enabled-setting';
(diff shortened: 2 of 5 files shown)


frontends/web/src/routes/settings/bb02-settings.tsx
```diff
@@ -154,6 +156,20 @@ const Content = ({ deviceID }: TProps) => {
         }
       </div>
 
+      {/*"Bluetooth" section*/}
+      { deviceInfo && deviceInfo.bluetooth ? (
+        <>
```
thisconnect: nit: fragment is not necessary here as you only have 1 child you can just pass the div element

frontends/web/src/api/bitbox02.ts
```diff
@@ -222,3 +226,7 @@ export const attestationCheckDone = (
 ): TUnsubscribe => {
   return subscribeEndpoint(`devices/bitbox02/${deviceID}/attestationCheckDone`, cb);
 };
+
+export const bluetoothToggleEnabled = (deviceID: string) => {
```
thisconnect: Add return type ? 
```suggestion
export const bluetoothToggleEnabled = (deviceID: string): Promise<SuccessResponse | FailResponse> => {
```

frontends/web/src/routes/settings/components/device-settings/bluetooth-toggle-enabled-setting.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return bluetooth.enabled;
+  }, [deviceID]);
+
+  useEffect(() => {
+    updateEnabled();
+  }, [updateEnabled]);
+
+  const handleBluetoothToggleEnabled = async () => {
+    setShow(true);
+    await bluetoothToggleEnabled(deviceID).catch(console.error);
```
thisconnect: Shouldn't it check the bluetoothToggleEnabled response if success is true?

thisconnect (changes requested):
untested LGTM with question about api reponse type and check for succes.

frontends/web/src/api/bitbox02.ts
benma: Good catch. Added.

frontends/web/src/routes/settings/bb02-settings.tsx
benma: Removed :+1:

frontends/web/src/routes/settings/components/device-settings/bluetooth-toggle-enabled-setting.tsx
benma: Handled now (also in getDeviceInfo()) with an alert, not sure if it's the right way.

thisconnect (approved):
👍 untested LGTM

PR #3424, opened 2025-06-25
frontend: improve bottom navigation for 'accounts'

shonsirsha:
1. directly go to that account when theres only 1 active account
2. change wording from 'accounts' to 'account'

Screenshot 2025-06-25 at 10 29 07

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -210,7 +210,9 @@ export const App = () => {
             />
             <RouterWatcher />
           </div>
-          {showBottomNavigation && <BottomNavigation />}
+          {showBottomNavigation && (
+            <BottomNavigation activeAccounts={activeAccounts} />
+          )}
           <Alert />
           <Confirm />
         </div>
--- a/frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
+++ b/frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
@@ -17,11 +17,18 @@
 import { useTranslation } from 'react-i18next';
 import { Link, useLocation } from 'react-router-dom';
 import { AccountIconSVG, ExchangeIconSVG, MoreIconSVG, PortfolioIconSVG } from '@/components/bottom-navigation/menu-icons';
+import type { Account } from '@/api/aopp';
 import styles from './bottom-navigation.module.css';
 
-export const BottomNavigation = () => {
+type Props = {
+  activeAccounts: Account[];
+}
+
+export const BottomNavigation = ({ activeAccounts }: Props) => {
   const { t } = useTranslation();
   const { pathname } = useLocation();
+  const onlyHasOneAccount = activeAccounts.length === 1;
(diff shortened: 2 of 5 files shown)


frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
```diff
@@ -34,10 +39,14 @@ export const BottomNavigation = () => {
...
-        to="/accounts/all"
+        to={
+          onlyHasOneAccount && accountCode ?
+            `/account/${accountCode}` :
+            '/accounts/all'
+        }
       >
         <AccountIconSVG />
-        {t('settings.accounts')}
+        {onlyHasOneAccount ? t('walletConnect.signingRequest.account') : t('settings.accounts')}
```
thisconnect: Hm I am slightly not a fan of reuseing walletConnect translation here in bottom nav. Also we do reuse if from settings.accounts :) 

Could we add generic string somewhere in account. for account.account and account.accounts and then also reuse in settings and WC?

frontends/web/src/app.tsx
```diff
@@ -210,7 +210,12 @@ export const App = () => {
             />
             <RouterWatcher />
           </div>
-          {showBottomNavigation && <BottomNavigation />}
+          {showBottomNavigation && (
+            <BottomNavigation
+              onlyHasOneAccount={activeAccounts.length === 1}
+              accountCode={activeAccounts[0]?.code}
```
thisconnect: an alternative could be to just pass activeAccounts, then do the check inside BottomNavigation component

thisconnect (changes requested):
untested LGTM with recommendation

frontends/web/src/app.tsx
shonsirsha: @thisconnect , ah yeah thats a lot better. Thanks.

frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
shonsirsha: @thisconnect , yes sure. Added the `account.accounts`, `account.account` and used them.  Then removed `settings.account` and the wallet connect one

frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
```diff
@@ -18,10 +18,17 @@ import { useTranslation } from 'react-i18next';
 import { Link, useLocation } from 'react-router-dom';
 import { AccountIconSVG, ExchangeIconSVG, MoreIconSVG, PortfolioIconSVG } from '@/components/bottom-navigation/menu-icons';
 import styles from './bottom-navigation.module.css';
+import { Account } from '@/api/aopp';
```
thisconnect: I believe it should be

```suggestion
import type { IAccount } from '@/api/account';
```

thisconnect (approved):
untested LGTM but with one more change request (see type import from api/account comment)

PR #3426, opened 2025-06-25
frontend: use sat spacing in all-accounts

thisconnect:
Branched off of https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3420 (so this needs to be merged first)

- use sat spacing in all-accounts menu
- simplify and remove some dry code

Please note that this needed some CSS changes, so if we want sat spacing this needs a bit more visual testing

### DIFF
--- a/frontends/web/src/components/amount/amount-with-unit.module.css
+++ b/frontends/web/src/components/amount/amount-with-unit.module.css
@@ -2,6 +2,7 @@
   cursor: default;
   line-height: 1;
   margin: 0;
+  max-width: 100%;
   white-space: nowrap;
 }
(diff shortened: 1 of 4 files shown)


thisconnect:
Screenshot 2025-06-25 at 11 57 34

thisconnect:
There is one more issue with the connected keystore name, which doesn't exist on the account-summary, but I'll open a separate PR for this. Looks like this could go into a shared component

see difference between account-summary and all-account

Screenshot 2025-08-19 at 15 47 24
Screenshot 2025-08-19 at 15 47 12

PR #3427, opened 2025-06-25
android: react to change in network capabilities.

bznein:
NOTE: this is based on https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3421 so that one needs to be reviewed before this one. 

Use ConnectivityManager to react to onLost (connection entirely lost, e.g. going into airplane mode), onAvailable (the opposite), and onCapabilitiesChanged (other scenarios, for example the network loses access to the internet, but it still works locally)

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
@@ -16,6 +16,9 @@ import android.content.res.Configuration;
 import android.hardware.usb.UsbDevice;
 import android.hardware.usb.UsbManager;
 import android.net.ConnectivityManager;
+import android.net.Network;
+import android.net.NetworkRequest;
+import android.net.NetworkCapabilities;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
@@ -84,6 +87,51 @@ public class MainActivity extends AppCompatActivity {
     // This is for the file picker dialog invoked by file upload forms in the WebView.
     // Used by e.g. MoonPay's KYC forms.
     private ValueCallback<Uri[]> filePathCallback;
+
+    private ConnectivityManager connectivityManager;
+    private ConnectivityManager.NetworkCallback networkCallback;
+
+    private boolean hasInternetConnectivity(NetworkCapabilities capabilities) {
+        // To avoid false positives, if we can't obtain connectivity info,
+        // we return true.
+        // Note: this should never happen per Android documentation, as:
+        // - these can not be null it come from the onCapabilitiesChanged callback.
+        // - when obtained with getNetworkCapabilities(network), they can only be null if the
+        //   network is null or unknown, but we guard against both in the caller.
+        if (capabilities == null) {
+            Util.log("Got null capabilities when we shouldn't have. Assuming we are online.");
+            return true;
+        }
+
+
+        boolean hasInternet = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
+
+        // Use VALIDATED where available (API 23+)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            // We need to check for both internet and validated, since validated reports that the system
+            // found connectivity the last time it checked. But if this callback triggers when going offline
+            // (e.g. airplane mode), this bit would still be true when we execute this method.
+            boolean isValidated = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED);
+            return hasInternet && isValidated;
+        }
+
+        // Fallback for older devices
+        return hasInternet;
+    }
+
+    private void checkConnectivity() {
+        Network activeNetwork = connectivityManager.getActiveNetwork();
+
+        // If there is no active network (e.g. airplane mode), there is no check to perform.
+        if (activeNetwork == null) {
+            Mobileserver.setOnline(false);
+            return;
+        }
+
+        NetworkCapabilities capabilities = connectivityManager.getNetworkCapabilities(activeNetwork);
+
+        Mobileserver.setOnline(hasInternetConnectivity(capabilities));
+    }
 
     // Connection to bind with GoService
     private ServiceConnection connection = new ServiceConnection() {
@@ -369,11 +417,30 @@ public class MainActivity extends AppCompatActivity {
         // We call updateDevice() here in case the app was started while the device was already connected.
         // In that case, handleIntent() is not called with ACTION_USB_DEVICE_ATTACHED.
         this.updateDevice();
+
+        connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        networkCallback = new ConnectivityManager.NetworkCallback() {
+            @Override
+            public void onCapabilitiesChanged(android.net.Network network, android.net.NetworkCapabilities capabilities) {
+                super.onCapabilitiesChanged(network, capabilities);
+                Mobileserver.setOnline(hasInternetConnectivity(capabilities));
+            }
+            // When we lose the network, onCapabilitiesChanged does not trigger, so we need to override onLost.
+            @Override
+            public void onLost(Network network) {
+                super.onLost(network);
+                Mobileserver.setOnline(false);
+            }
+        };
+
     }
 
     private void startServer() {
         final GoViewModel gVM = ViewModelProviders.of(this).get(GoViewModel.class);
         goService.startServer(getApplicationContext().getFilesDir().getAbsolutePath(), gVM.getGoEnvironment(), gVM.getGoAPI());
+
+        // Trigger connectivity check (as the network may already be unavailable when the app starts).
+        checkConnectivity();
     }
 
     private static String readRawText(InputStream inputStream) throws IOException {
@@ -467,6 +534,14 @@ public class MainActivity extends AppCompatActivity {
                 });
             }
         });
+
+        NetworkRequest request = new NetworkRequest.Builder()
+            .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+            .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+            .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
+            .build();
+        // Register the network callback to listen for changes in network capabilities.
+        connectivityManager.registerNetworkCallback(request, networkCallback);
     }
 
     @Override
@@ -494,6 +569,9 @@ public class MainActivity extends AppCompatActivity {
 
         // Listen on changes in the network connection. We are interested in if the user is connected to a mobile data connection.
         registerReceiver(this.networkStateReceiver, new IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION));
+
+        // Trigger connectivity check (as the network may already be unavailable when the app starts).
+        checkConnectivity();
 
         Intent intent = getIntent();
         handleIntent(intent);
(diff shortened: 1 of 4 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -467,6 +493,30 @@ public void run() {
...
+            @Override
+            public void onLost(android.net.Network network) {
+                super.onLost(network);
+                Mobileserver.setOnline(false);
+            }
+            // The same is true for when the network is available again.
+            @Override
+            public void onAvailable(android.net.Network network) {
+                super.onAvailable(network);
+                Mobileserver.setOnline(true);
```
benma: onAvailable means there is a link (cable, WiFi), but not necessarily that you are online. E.g. WiFi catptive portals would return true here while not being connected. (see comment above in hasInternetConnectivity()) 

use 

```
NetworkCapabilities capabilities = connectivityManager.getNetworkCapabilities(network);
Mobileserver.setOnline(hasInternetConnectivity(capabilities));
```

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -85,6 +88,26 @@ public class MainActivity extends AppCompatActivity {
...
+    private ConnectivityManager.NetworkCallback networkCallback;
+
+    private boolean hasInternetConnectivity(NetworkCapabilities capabilities) {
+        // To avoid false positives, if we can't obtain connectivity info,
+        // we return true.
+        if (capabilities == null) {
+            return true;
+        }
+
+        return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
```
benma: This is not enough

According to

https://developer.android.com/reference/android/net/NetworkCapabilities

> The NET_CAPABILITY_INTERNET capability means that the network is configured to reach the general Internet. It may or may not actually provide connectivity ; the NET_CAPABILITY_VALIDATED bit indicates that the system found actual connectivity to the general Internet the last time it checked. Apps interested in actual connectivity should usually look at both these capabilities. 

DeepSeek suggests:

```
    // Use VALIDATED where available (API 23+)
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
        return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED);
    }
    // Fallback for older devices
    return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
```

benma (changes requested):
DeepSeek suggets these which seem reasonable:

---

3. **Improvements Needed**
   - Add `unregisterNetworkCallback()` in `onDestroy()` to prevent leaks
   - Handle null `activeNetwork` in connectivity checks
   - Use capability-aware `NetworkRequest`:
     ```java
     new NetworkRequest.Builder()
         .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
         .build()
     ```

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -467,6 +499,30 @@ public void run() {
...
+            @Override
+            public void onLost(android.net.Network network) {
+                super.onLost(network);
+                Mobileserver.setOnline(false);
+            }
+            // The same is true for when the network is available again.
+            @Override
+            public void onAvailable(android.net.Network network) {
+                super.onAvailable(network);
+                checkConnectivity();
```
benma: checkConnectivity uses `connectivityManager.getActiveNetwork();`, but there is a `network` argument here already. Could these be different? Anyway your version is better I think, in case there is a non-active network that is down it does not affect us, but maybe add a comment to make that explicit.

bznein: I think it's hard for them to be different, but I agree the current one seems better, will add a comment

bznein: Done, but also added a call to checkConnectivity to "onlost" for the same reason

bznein:
> DeepSeek suggets these which seem reasonable:
> 
> 3. **Improvements Needed**
>    
>    * Use capability-aware `NetworkRequest`:
>      ```java
>      new NetworkRequest.Builder()
>          .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
>          .build()
>      ```

I'm not sure about this one, the docs read:

>  Add the given capability requirement to this builder. These represent the requested network's required capabilities. Note that when searching for a network to satisfy a request, all capabilities requested must be satisfied.

My understanding is that if we add `NET_CAPABILITY_INTERNET`, the callback would not trigger when we are offline since this capability will not be satisfied by the current network. WDYT?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -85,6 +88,38 @@ public class MainActivity extends AppCompatActivity {
...
+        return capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
+    }
+
+    private void checkConnectivity() {
+        Network activeNetwork = connectivityManager.getActiveNetwork();
+
+        // Avoid false positives by assuming we are offline if activeNetwork is null
+        if (activeNetwork == null) {
+            Mobileserver.setOnline(true);
+        }
```
benma, line 129: return here?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -467,6 +505,36 @@ public void run() {
...
+                super.onAvailable(network);
+                // We call checkConnectivity and ignore the current network,
+                // so in that function we are sure we get the currently active one
+                // and not a different one
+                checkConnectivity();
+            }
+        };
+
+        // Register the network callback to listen for changes in network capabilities.
+        connectivityManager.registerNetworkCallback(new NetworkRequest.Builder().build(), networkCallback);
```
benma: Should this maybe be `registerDefaultNetworkCallback()`? This only listens on the active network, which is what we do anyway in `checkConnectivity()`.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -595,6 +666,7 @@ protected void onDestroy() {
             unbindService(connection);
         }
         super.onDestroy();
+        connectivityManager.unregisterNetworkCallback(networkCallback);
```
benma: `if (connectivityManager != null && networkCallback != null) {` for robustness.

I also just noticed that the register/deregister pair does not match. onCreate pairs with onDestroy, and onStart pairs with onStop. but currently the stuff is registered in onStart. 

Would it make sense to just move the setup to onCreate instead?

benma (changes requested):

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -595,6 +667,9 @@ protected void onDestroy() {
             unbindService(connection);
         }
         super.onDestroy();
+        if (connectivityManager != null && networkCallback != null) {
+            connectivityManager.unregisterNetworkCallback(networkCallback);
```
benma: Register is in onStart, so unregister should be in onStop.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -467,6 +506,36 @@ public void run() {
                 });
             }
         });
+
+        connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        networkCallback = new ConnectivityManager.NetworkCallback() {
```
benma: Maybe move these two to `onCreate` and leave `register...` in `onStart()`?

bznein:
@benma ready for review again, I tested it on Android and I believe it now covers all cases.

A few notes since many things changed since last version:

Removed callbacks to `onAvailable`, `onUnavailable`: these just return whether a network is available or not, no info about connectivity, and we should NOT call `getNetworkCapabilities` here, based on [the docs](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onAvailable(android.net.Network):~:text=Do%20NOT%20call,to%20other%20callbacks.)

Similarly, changed `onLost` to NOT call the network capabilities changed callback. Based again on [the docs](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onLost(android.net.Network):~:text=If%20the%20callback%20was%20registered%20with%20requestNetwork()%20or%20registerDefaultNetworkCallback()%2C%20it%20will%20only%20be%20invoked%20against%20the%20last%20network%20returned%20by%20onAvailable()%20when%20that%20network%20is%20lost%20and%20no%20other%20network%20satisfies%20the%20criteria%20of%20the%20request.) this will be called when the last network is lost, and no other network satisfies the criterias for the request

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -85,6 +88,51 @@ public class MainActivity extends AppCompatActivity {
...
+        // - these can not be null it come from the onCapabilitiesChanged callback.
+        // - when obtained with getNetworkCapabilities(network), they can only be null if the
+        //   network is null or unknown, but we guard against both in the caller.
+        if (capabilities == null) {
+            Util.log("Got null capabilities when we shouldn't have. Assuming we are online.");
+            return true;
+        }
+
+
+        boolean isValidated = capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED);
```
benma: I think this may crash on <23, and should be moved inside the `if` clause.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/MainActivity.java
```diff
@@ -467,6 +534,14 @@ public void run() {
...
             }
         });
+
+        NetworkRequest request = new NetworkRequest.Builder()
+            .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+            .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+            .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
+            .build();
+        // Register the network callback to listen for changes in network capabilities.
+        connectivityManager.registerNetworkCallback(request,networkCallback);
```
benma: nit formatting: space after comma

benma (changes requested):

benma:
@bznein I fixed up my two nits myself as I am about to make a RC1 now, which will include this PR.

benma (approved):
tACK. There is a brief flashing of the banner when going from WiFi to mobile connection, but imho not a big deal.

PR #3428, opened 2025-06-25
ios: react to change in connectivity.

bznein:
NOTE: this is based on https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3421 so that one needs to be reviewed before this one.

Use NwPathMonitor to monitor change in connectivity, and trigger the offline banner when going offline, and a manual reconnection when going back online.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/ios/BitBoxApp/BitBoxApp.xcodeproj/project.pbxproj
+++ b/frontends/ios/BitBoxApp/BitBoxApp.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		8E1846EB2E0AFF6000802D8B /* NetworkMonitor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E1846EA2E0AFF5B00802D8B /* NetworkMonitor.swift */; };
+		8E1846EC2E0AFF6000802D8B /* NetworkMonitor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E1846EA2E0AFF5B00802D8B /* NetworkMonitor.swift */; };
 		D700B5F62C884C34000496D4 /* Mobileserver.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = D76518BB2B1F8F7400DC03A9 /* Mobileserver.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		D700B5FD2CA2FFAF000496D4 /* WebView.swift in Sources */ = {isa = PBXBuildFile; fileRef = D76516322B1F3D1A00DC03A9 /* WebView.swift */; };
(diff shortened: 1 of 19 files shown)


bznein:
@benma This would be ready for review; however, while testing on the simulator, the banner seems to work fine when going offline, but not when going back online. I still don't have an iPhone to test on, so if you could test and report if it behaves fine on a real device it would be great!

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
```diff
@@ -184,6 +185,10 @@ struct BitBoxAppApp: App {
                     .edgesIgnoringSafeArea(.all)
                     .onAppear {
                         setupGoAPI(goAPI: goAPI)
+                        _ = NetworkMonitor.shared
```
benma: Why not use that? `isOnline = NetworkMonitor.shared.monitor.currentPath.status == .satisfied`

bznein:
> @benma This would be ready for review; however, while testing on the simulator, the banner seems to work fine when going offline, but not when going back online. I still don't have an iPhone to test on, so if you could test and report if it behaves fine on a real device it would be great!

@benma I received my iPhone and tested on it. Ready for review for real :)

frontends/ios/BitBoxApp/BitBoxApp.xcodeproj/project.pbxproj
```diff
@@ -464,7 +470,7 @@
 				INFOPLIST_KEY_UIApplicationSupportsIndirectInputEvents = YES;
 				INFOPLIST_KEY_UILaunchScreen_Generation = YES;
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad = "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
-				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone = "UIInterfaceOrientationPortrait";
```
bznein: These changes have been made by Xcode automatically, not sure the reason or whether they should be reverted or not

benma: git blame will tell you the reason for this line, it's deliberate :) should definitely revert

bznein: well git blame won't tell me why XCode changed it, I meant :D reverted!

frontends/ios/BitBoxApp/BitBoxApp/BitBoxAppApp.swift
bznein: `monitor` is private within NetworkMonitor so if we want to do this we need to make it non-private. I don't have a strong preference but I went for an alternative approach of adding an `isOnline` method and calling that instead

bznein:
ping :) @benma

benma (approved):
tACK

PR #3429, opened 2025-06-25
fronted/qt: fix web engine profile for external links

Beerosagos:
In 00d7607 we switched from the default profile to a dedicated one, to store profile data on disk. This change didn't involve the WebEnginePage used to display external resources (e.g. links in 3rd party widgets) and caused external links not to be displayed because of incompatible profiles.

This fixes the issue sharing the same profile also for external pages.

### DIFF
--- a/frontends/qt/main.cpp
+++ b/frontends/qt/main.cpp
@@ -353,8 +353,8 @@ int main(int argc, char *argv[])
         view->adjustSize();
     }
 
-    externalPage = new QWebEnginePage(view);
     profile = new QWebEngineProfile("BitBoxApp");
+    externalPage = new QWebEnginePage(profile, view);
     mainPage = new WebEnginePage(profile);
     view->setPage(mainPage);
 


frontends/qt/main.cpp
```diff
@@ -353,8 +353,8 @@ int main(int argc, char *argv[])
         view->adjustSize();
     }
 
-    externalPage = new QWebEnginePage(view);
     profile = new QWebEngineProfile("BitBoxApp");
+    externalPage = new QWebEnginePage(profile);
```
benma: shouldn't it be `new QWebEnginePage(profile, view), i.e. keep the parent the same?

benma (changes requested):
Wow, nasty.

frontends/qt/main.cpp
Beerosagos: Oh yes, I guess you're right! Fixed :pray:

benma (approved):
utACK

PR #3430, opened 2025-06-25
chore: switch from GHCR to Docker for CI.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,38 +13,23 @@ jobs:
   setup-env:
     runs-on: ubuntu-22.04
     outputs:
-      ci_image: ${{ steps.set-ci-image.outputs.ci_image }}
+      ci_image_version: ${{ steps.set-ci-image-version.outputs.ci_image_version }}
     steps:
       - name: Clone the repo
         uses: actions/checkout@v4
 
       - name: Read CI Image Version
-        id: set-ci-image
+        id: set-ci-image-version
         run: |
-          # Use alternative image when running on GitHub workflows CI to avoid potential
-          # rate limiting when executing jobs in parallel: they can't cache docker images
-          # and always pull.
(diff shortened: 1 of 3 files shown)


benma:
Re-assigning to @NickeZ as he is the ultimate CI guru :smile:

bznein:
> lgtm, did you test it in your own fork?

I did! You can see here: https://github.com/bznein/bitbox-wallet-app/actions a bunch of failed actions before I finally got it to work :)

.github/workflows/ci.yml
```diff
@@ -77,17 +63,18 @@ jobs:
         run: ./scripts/github-ci.sh android
         env:
           OS_NAME: linux
-          CI_IMAGE: ${{ needs.setup-env.outputs.ci_image }}
       - name: Upload APK
         id: upload
         uses: actions/upload-artifact@v4
         with:
-          path: frontends/android/BitBoxApp/app/build/outputs/apk/debug/app-debug.apk
+          path: /github/home/go/src/github.com/BitBoxSwiss/bitbox-wallet-app/frontends/android/BitBoxApp/app/build/outputs/apk/debug/app-debug.apk
```
benma, line 70: unusual to see absolute paths like this - why does relative not work?

bznein: It fails with with a "not found" error: https://github.com/bznein/bitbox-wallet-app/actions/runs/15879899679/job/44777165124

I wasn't able to figure out at which path is the step running... I tried adding a debug print on the step before and it looks like it should be able to find it: https://github.com/bznein/bitbox-wallet-app/actions/runs/15921416849/job/44908858374

But it doesn't.. I suppose there is some difference when running on a specifid container.

If it's a big deal, I can try figuring out what's happening anyway!

PR #3432, opened 2025-06-27
backend/accounts: skip unsupported coins when adding hidden accounts

benma:
With a BB02 BTC-only version, the log showed attempts to add LTC accounts, which were then skipped as unsupported. We can abort earlier in this case, which prevents the log from showing 'automatically created hidden account' for LTC accounts when LTC is not supported.


Beerosagos (approved):
utACK

PR #3434, opened 2025-07-01
frontend: toggle Bluetooth only on non-IOS devices

benma:


frontends/web/src/routes/settings/bb02-settings.tsx
```diff
@@ -108,7 +109,6 @@ const Content = ({ deviceID }: TProps) => {
           return;
         }
         setDeviceInfo(result.deviceInfo);
-        console.log('LOL', result);
```
thisconnect, line 111: 😅 👍

PR #3438, opened 2025-07-02
ci: add ad-hoc app signature for macos target

Beerosagos:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,15 @@ jobs:
           ./scripts/github-ci.sh qt-osx;
         env:
           OS_NAME: osx
+      - name: Ad-hoc sign BitBox.app
+        run: |
+          APP_PATH="$(go env GOPATH)/src/github.com/BitBoxSwiss/bitbox-wallet-app/frontends/qt/build/osx/BitBox.app"
+          echo "Signing app at $APP_PATH"
+          codesign --deep --force --verbose --sign - "$APP_PATH"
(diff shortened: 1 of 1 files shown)


Beerosagos:
@benma this seems to fix the issue for me and @thisconnect. Here it is quite easy to test, since the previous commit artifact is blocked by mac os, but maybe it would be better to bring it directly on master. wdyt?

.github/workflows/ci.yml
```diff
@@ -159,6 +159,15 @@ jobs:
           ./scripts/github-ci.sh qt-osx;
         env:
           OS_NAME: osx
+      - name: Ad-hoc sign BitBox.app
+        run: |
+          APP_PATH="$HOME/go/src/github.com/BitBoxSwiss/bitbox-wallet-app/frontends/qt/build/osx/BitBox.app"
```
benma: I am confused why this is the right path. Isn't the repo and build in `{{github.workspace}}`, which is the working dir? 

Doesn't a relative path work like in other jobs then? I.e. simply `frontends/qt/build/osx/BitBox.app`. 

If not, it would be good to use `go env GOPATH` instead of `$HOME/go`.

Beerosagos: It doesn't work with the relative path, but also the archive app part uses a global one.. anyway, I changed it to use the GOPATH as you suggested :pray:

benma (approved):
utACK

PR #3440, opened 2025-07-07
backend: add blockbook/etherscan wrapper.

bznein:
Note that this is targeting a staging branch; there are a few TODOs left that will be fixed before opening a PR against master.

This first implementation adds a wrapper and the call to `Balance` in blockbook. Tested by comparing the returned balance on `master` and here (and added a log line to make sure that the balance is indeed returned by Blockbook.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -226,11 +226,9 @@ type Backend struct {
 
 	socksProxy socksproxy.SocksProxy
 	// can be a regular or, if Tor is enabled in the config, a SOCKS5 proxy client.
-	httpClient          *http.Client
-	etherScanHTTPClient *http.Client
-	blockbookHTTPClient *http.Client
-	ratesUpdater        *rates.RateUpdater
-	banners             *banners.Banners
+	httpClient   *http.Client
+	ratesUpdater *rates.RateUpdater
+	banners      *banners.Banners
 
 	// For unit tests, called when `backend.checkAccountUsed()` is called.
 	tstCheckAccountUsed func(accounts.Interface) bool
@@ -293,8 +291,6 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.notifier = notifier
 	backend.socksProxy = backendProxy
 	backend.httpClient = hclient
-	backend.etherScanHTTPClient = hclient
-	backend.blockbookHTTPClient = hclient
 
 	ratesCache := filepath.Join(arguments.CacheDirectoryPath(), "exchangerates")
 	if err := os.MkdirAll(ratesCache, 0700); err != nil {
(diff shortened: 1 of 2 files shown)


backend/backend.go
```diff
@@ -291,6 +294,7 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.socksProxy = backendProxy
 	backend.httpClient = hclient
 	backend.etherScanHTTPClient = hclient
+	backend.blockbookHTTPClient = hclient
```
benma: Maybe in the past these were different, but now that they are all the same, it would make sense to just use `backend.httpClient` and remove the other two, right?

backend/coins/eth/blockbook/blockbook.go
```diff
@@ -0,0 +1,163 @@
...
+// Blockbook is a rate-limited blockbook api client.
+type Blockbook struct {
+	url        string
+	httpClient *http.Client
+	limiter    *rate.Limiter
+}
+
+// NewBlockbook creates a new instance of EtherScan.
+func NewBlockbook(chainId string, httpClient *http.Client) *Blockbook {
+	// TODO chainID is not used here, do we have blockbook running for SEPETH as well?
```
benma, line 52: Best to panic if chainID is not `1` until you find out, so it can never be used on the wrong chain.

backend/coins/eth/blockbook/blockbook.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return nil
+}
+
+func (blockbook *Blockbook) address(ctx context.Context, account common.Address, result interface{}) error {
+	params := url.Values{}
+	address := account.Hex()
+
+	if err := blockbook.call(ctx, path.Join("address", address), params, result); err != nil {
```
benma: This is a URL not a filepath, so on Windows this would use a backslash?

backend/coins/eth/rpcclient/wrapper/wrapper.go
```diff
@@ -0,0 +1,146 @@
...
+
+type client struct {
+	defaultRpcClient  rpcClientWithTransactionsSource
+	fallbackRpcClient rpcClientWithTransactionsSource
+
+	// TODO do we maybe want to add some fields to keep track
+	// of which one we are using? if we switch to the fallback one
+	// it might make sense to use that for a while instead of wasting
+	// time each time trying the default first and falling back to the
+	// other one later.
```
benma: Definitely! Maybe this package could help, but it may also be overkill :shrug: 

https://github.com/BitBoxSwiss/block-client-go/blob/master/failover/failover.go

backend/backend.go
```diff
@@ -539,9 +545,11 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 			nil)
 	case erc20Token != nil:
 		etherScan := etherscan.NewEtherScan("1", backend.etherScanHTTPClient)
-		coin = eth.NewCoin(etherScan, erc20Token.code, erc20Token.name, erc20Token.unit, "ETH", params.MainnetChainConfig,
+		blockBook := blockbook.NewBlockbook("1", backend.blockbookHTTPClient)
```
benma: Blockbook is able to deliver the results for the address for all tokens too in one shot, along with the main ETH address. I think it would be worth it to reduce the load on the backend. `tokenBalances` could be used as the `detail` param only if there are active ERC20 tokens too.  

Do you have a plan in mind for how to achieve that? If you use a separate blockBook instance here this will not be possible. but maybe you have something in mind.

backend/backend.go
bznein: Yep, done!

backend/backend.go
bznein: I haven't thought about yet in details, but my idea would be to change the interface (or have the wrapper not implement `rpcclient.interface`) so that the wrapper exposes one `Balance` that takes a parameter to mean if we either want only ETH, only ERC20, or both (assuming all three cases make sense, if only the latter does, there is no need for a parameter) and tries to call the Blockbook instance first (with a single call in any case), and if it fails back to Etherscan, it just makes separate calls as needed

backend/coins/eth/blockbook/blockbook.go
bznein: Good idea, done

backend/coins/eth/blockbook/blockbook.go
bznein: Yeah good catch, I moved away from path for other reasons and forgot about this one!

backend/coins/eth/rpcclient/wrapper/wrapper.go
bznein: Good shout, I'll take a look at it before merging to master!

backend/coins/eth/rpcclient/wrapper/wrapper.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// NewClient creates a new client that uses the default RPC client unless it fails,
+// in which case it falls back to the fallback RPC client.
+func NewClient(defaultRPCClient, fallbackRPCClient rpcClientWithTransactionsSource) *client {
```
benma: It's unusual to see an exported method return an instance an unexported struct. I'm somewhat surprised it even compiles :D

PR #3441, opened 2025-07-07
backend: migrate methods BlocKNumber and ERC20Balance to blockbook.

bznein:
To be reviewed after #3440 

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/eth/blockbook/blockbook.go
+++ b/backend/coins/eth/blockbook/blockbook.go
@@ -29,9 +29,11 @@ import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth/rpcclient"
 	ethtypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth/types"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/util/logging"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/sirupsen/logrus"
 	"golang.org/x/time/rate"
 )
 
@@ -45,6 +47,8 @@ type Blockbook struct {
 	url        string
 	httpClient *http.Client
 	limiter    *rate.Limiter
+	// TODO remove before merging into master?
+	log *logrus.Entry
 }
 
 // NewBlockbook creates a new instance of EtherScan.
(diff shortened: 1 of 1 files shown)


backend/coins/eth/blockbook/blockbook.go
```diff
@@ -134,7 +150,29 @@ func (blockbook *Blockbook) SendTransaction(ctx context.Context, tx *types.Trans
...
 func (blockbook *Blockbook) ERC20Balance(account common.Address, erc20Token *erc20.Token) (*big.Int, error) {
-	return nil, fmt.Errorf("Not yet implemented")
+	result := struct {
+		Tokens []struct {
+			Balance  string `json:"balance"`
+			Contract string `json:"contract"`
+		} `json:"tokens"`
+	}{}
+
+	// TODO why is there no context in the signature of this interface method?
```
benma, line 160: Probably just a miss, you should add it if it makes sense.

backend/coins/eth/blockbook/blockbook.go
```diff
@@ -134,7 +150,29 @@ func (blockbook *Blockbook) SendTransaction(ctx context.Context, tx *types.Trans
...
+	if err := blockbook.address(context.Background(), account, &result); err != nil {
```
benma, line 161: See also https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3440#discussion_r2196877309

It would be good to call `address()` only once for the main ETH account, the result contains the balance for the main ETH account and all the ERC20 token balances too if the detail param is set to "tokekenBalances" or higher.

Same for transactions, a separate call would not be needed there, it's included in the results with the right params.

bznein: as discussed on MM and per comment on the other PR, I definitely want to do it, just after migrating all necessary method first!

backend/coins/eth/blockbook/blockbook.go
bznein: I'd like to refactor our context usage in general. Other methods have context in the signature, but we mostly pass `context.TODO()`  😅

backend/coins/eth/blockbook/blockbook.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, errp.WithStack(err)
+	}
+
+	for _, token := range result.Tokens {
+		if token.Contract == erc20Token.ContractAddress().Hex() {
+			balance, ok := new(big.Int).SetString(token.Balance, 10)
+			if !ok {
+				return nil, errp.Newf("could not parse balance %q", token.Balance)
+			}
+			blockbook.log.Infof("ERC20 balance for %s: %s", erc20Token.ContractAddress().Hex(), balance.String())
```
benma: Logs should not contain this sensitive info, please remove.

benma (changes requested):

benma (approved):

PR #3443, opened 2025-07-09
backend: add Transactions implementation to blockbook.

bznein:
Note: based on #3440 

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/eth/blockbook/blockbook.go
+++ b/backend/coins/eth/blockbook/blockbook.go
@@ -22,15 +22,19 @@ import (
 	"math/big"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth/erc20"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth/rpcclient"
 	ethtypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth/types"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/util/logging"
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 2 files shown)


backend/coins/eth/blockbook/blockbook.go
```diff
@@ -0,0 +1,264 @@
...
+	isERC20 bool,
+	blockTipHeight *big.Int,
+	isInternal bool,
+	transactions []*Tx, address common.Address) ([]*accounts.TransactionData, error) {
+
+	// TODO figure out if needed. Etherscan.go uses this to compute the num of confirmations.
+	// But numConfirmations is already returned by the API call.
+	_ = blockTipHeight
+
+	_ = isInternal // TODO figure out how to deal with internal txs.
```
benma, line 159: Please do, my ERC20 accounts don't show any transactions with this PR :sweat_smile: I should probably review once it works for ETH and ERC20, or do you want to separate it into two PRs?

backend/coins/eth/blockbook/blockbook.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		from := tx.Vin[0].Addresses[0]
+		var to string
+		if len(tx.TokenTransfers) > 0 {
+			to = tx.TokenTransfers[0].To
+		} else {
+			to = tx.Vout[0].Addresses[0]
+		}
+		if ours != from && ours != to {
+			return nil, errp.New("transaction does not belong to our account")
```
benma, line 188: This errors on my ETH account I use for testing, because some spam tx touched it:

https://etherscan.io/tx/0x1a926b5296dc6803a8a07daeebbd6b1b6b476790af5719ec628d5eeca08027f2

This tx has a ton of ERC-1155 token transfers (NFTs?), one of which touches my address.

I am not sure so you should check, but I think the Etherscan-based account includes such transactions in the list if it originated from the account, in other words, if the ETH accounts paid the fees for it (which impacts the account balance and portfolio chart calculation). I guess it does not make sense to list "incoming" token transfers in the main ETH account, these are shown separately in the respective ERC20 accounts.

backend/coins/eth/blockbook/blockbook.go
bznein: I think since we are on a staging branch, is fine to do it in a separate PR. Definitely need to do some investigation for it :)

PR #3446, opened 2025-07-10
frontend: add backup reminder banner.

bznein:
For each wallet, the first time it crosses 1000 USD/CHF/EUR, display a banner suggesting users to create a paper backup of their seed. If the default currency is different from these three, we use USD.

If the keystore is connected, the banner shows a link to the device setting where the user can see their recovery words.
If the keystore is not connected, the banner asks the user to connect it.

If amounts are hidden, we do not display the banner at all.

Video (where I manually changed the threshold from 1000 to 1, you'll need to do the same to test it if you don't have at least 1000 in a wallet) - note: the video is slightly outdated but still shows the overall flow

https://github.com/user-attachments/assets/83619e6f-d659-4bd0-bd41-41d4981370ba

A few comments are left inline.

Tested on Android and Linux
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -343,6 +343,45 @@ type KeystoreBalance = struct {
 	CoinsBalance AmountsByCoin `json:"coinsBalance"`
 }
 
+// AccountsFiatAndCoinBalance returns the total fiat balance and the balance for each coin, of a list of accounts.
+func (backend *Backend) AccountsFiatAndCoinBalance(accounts AccountsList, fiatUnit string) (*big.Rat, map[coinpkg.Code]*big.Int, error) {
+	keystoreBalance := new(big.Rat)
+	keystoreCoinsBalance := make(map[coinpkg.Code]*big.Int)
+
+	for _, account := range accounts {
+		if account.Config().Config.Inactive || account.Config().Config.HiddenBecauseUnused {
+			continue
+		}
+		if account.FatalError() {
+			continue
+		}
+		err := account.Initialize()
+		if err != nil {
+			return nil, nil, err
+		}
+
+		accountFiatBalance, err := backend.accountFiatBalance(account, fiatUnit)
+		if err != nil {
+			return nil, nil, err
+		}
+		keystoreBalance.Add(keystoreBalance, accountFiatBalance)
+
+		coinCode := account.Coin().Code()
+		balance, err := account.Balance()
+		if err != nil {
+			return nil, nil, err
+		}
+		accountBalance := balance.Available().BigInt()
+		if _, ok := keystoreCoinsBalance[coinCode]; !ok {
+			keystoreCoinsBalance[coinCode] = accountBalance
+		} else {
+			keystoreCoinsBalance[coinCode] = new(big.Int).Add(keystoreCoinsBalance[coinCode], accountBalance)
+		}
+	}
+
+	return keystoreBalance, keystoreCoinsBalance, nil
+}
+
 // keystoresBalance returns a map of accounts' total balances across coins, grouped by keystore.
 func (backend *Backend) keystoresBalance() (map[string]KeystoreBalance, error) {
 	keystoreBalanceMap := make(map[string]KeystoreBalance)
@@ -353,38 +392,9 @@ func (backend *Backend) keystoresBalance() (map[string]KeystoreBalance, error) {
 		return nil, err
 	}
 	for rootFingerprint, accountList := range accountsByKeystore {
-		keystoreCoinsBalance := make(map[coinpkg.Code]*big.Int)
-		keystoreTotalBalance := new(big.Rat)
-		for _, account := range accountList {
-			if account.Config().Config.Inactive || account.Config().Config.HiddenBecauseUnused {
-				continue
-			}
-			if account.FatalError() {
-				continue
-			}
-			err := account.Initialize()
-			if err != nil {
-				return nil, err
-			}
-
-			accountFiatBalance, err := backend.accountFiatBalance(account, fiatUnit)
-			if err != nil {
-				return nil, err
-			}
-			keystoreTotalBalance.Add(keystoreTotalBalance, accountFiatBalance)
-
-			coinCode := account.Coin().Code()
-			balance, err := account.Balance()
-			if err != nil {
-				return nil, err
-			}
-			accountBalance := balance.Available().BigInt()
-			if _, ok := keystoreCoinsBalance[coinCode]; !ok {
-				keystoreCoinsBalance[coinCode] = accountBalance
-
-			} else {
-				keystoreCoinsBalance[coinCode] = new(big.Int).Add(keystoreCoinsBalance[coinCode], accountBalance)
-			}
+		keystoreTotalBalance, keystoreCoinsBalance, err := backend.AccountsFiatAndCoinBalance(accountList, fiatUnit)
+		if err != nil {
+			return nil, err
 		}
 
 		keystoreCoinsAmount := AmountsByCoin{}
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"os"
 	"runtime/debug"
+	"slices"
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
@@ -34,6 +35,7 @@ import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/bitsurance"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	accountHandlers "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/handlers"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/config"
@@ -72,6 +74,7 @@ type Backend interface {
 	Testing() bool
 	Accounts() backend.AccountsList
 	AccountsByKeystore() (backend.KeystoresAccountsListMap, error)
+	AccountsFiatAndCoinBalance(backend.AccountsList, string) (*big.Rat, map[coinpkg.Code]*big.Int, error)
 	Keystore() keystore.Keystore
 	AccountsBalanceSummary() (*backend.AccountsBalanceSummary, error)
 	OnAccountInit(f func(accounts.Interface))
@@ -251,8 +254,8 @@ func NewHandlers(
 	getAPIRouterNoError(apiRouter)("/aopp/cancel", handlers.postAOPPCancel).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/aopp/approve", handlers.postAOPPApprove).Methods("POST")
 	getAPIRouter(apiRouter)("/aopp/choose-account", handlers.postAOPPChooseAccount).Methods("POST")
-	getAPIRouterNoError(apiRouter)("/cancel-connect-keystore", handlers.postCancelConnectKeystore).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/connect-keystore", handlers.postConnectKeystore).Methods("POST")
+	getAPIRouterNoError(apiRouter)("/cancel-connect-keystore", handlers.postCancelConnectKeystore).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/set-watchonly", handlers.postSetWatchonly).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/on-auth-setting-changed", handlers.postOnAuthSettingChanged).Methods("POST")
 	getAPIRouterNoError(apiRouter)("/export-log", handlers.postExportLog).Methods("POST")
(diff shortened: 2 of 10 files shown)


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -115,6 +117,7 @@ export const AccountsSummary = ({
       ...prevBalances,
       [code]: balance.balance
     }));
+    setBalancesTrigger(Date.now());
```
bznein: This is the part I don't like... I need the banner to be able to trigger again when there is a call to "update" (otherwise it would not show up when opening the app, since the accounts are not loaded yet, or when a tx is received that bumps us over the thresold), so I came up with this solution with some AI help. I don't love it and I am open to more suggestions. CC @thisconnect :)

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -179,6 +182,14 @@ export const AccountsSummary = ({
             <Status hidden={!hasCard} type="warning">
               {t('warning.sdcard')}
             </Status>
+            {accountsByKeystore.map(({ keystore }) => (
+              <BackupReminder
+                key={keystore.rootFingerprint}
+                keystore={keystore}
+                deviceID={Object.keys(devices)[0]}
```
bznein: I worked on the assumption that there is only one value here when the keystore is connected, as we can't have multiple BB02 connected at the same time. It works, but are there possibly cases I missed here?

backend/handlers/handlers.go
```diff
@@ -1494,3 +1497,57 @@ func (handlers *Handlers) postBluetoothConnect(r *http.Request) interface{} {
...
+	if balance == nil {
+		handlers.log.Error("No balance found for the given root fingerprint")
+		return response{
+			Success: false,
+		}
+	}
+
+	return response{
+		Success: true,
+		Show:    balance.Cmp(big.NewRat(1000, 1)) > 0,
```
bznein: Change this "1000" to a lower value for testing

frontends/web/src/components/banners/backup.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(() => {
+    getShowBackupBanner(keystore.rootFingerprint).then(setBannerResponse);
+  }, [keystore.rootFingerprint, trigger]);
+
+  if (!bannerResponse || !bannerResponse.success) {
+    return null;
+  }
+
+  const deviceSettingsURL = `/settings/device-settings/${deviceID}`;
```
bznein: Is there a way to directly link to "show recovery words"?  

One solution I'd love would be a link that just "highlights" the "show recovery words" after link, which would be a much better UX than automatically "clicking" it.

Not sure if either is possible TBH

bznein: I tried something that highlights it, but it's very very ugly :)

frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -179,6 +187,14 @@ export const AccountsSummary = ({
             <Status hidden={!hasCard} type="warning">
               {t('warning.sdcard')}
             </Status>
+            {accountsByKeystore.map(({ keystore }) => (
+              <BackupReminder
+                key={keystore.rootFingerprint}
+                keystore={keystore}
```
benma, line 190: `accountsBalanceSummary?.keystoresBalance[keystore.rootFingerprint].total` contains the fiat total already, but I guess the issue is that you need to revert to USD of it's not one of the three currencies?

frontends/web/src/routes/account/summary/accountssummary.tsx
benma: A device might not be connected (watch-only), so `[0]` would throw an exception?

backend/accounts.go
```diff
@@ -343,6 +343,32 @@ type KeystoreBalance = struct {
 	CoinsBalance AmountsByCoin `json:"coinsBalance"`
 }
 
+// AccountsFiatBalance returns the total fiat balance of a list of accounts.
+func (backend *Backend) AccountsFiatBalance(accounts AccountsList, fiatUnit string) (*big.Rat, error) {
```
benma: We have a lot of balance computation functions, so imho it's important to reuse/refactor as we go instead of adding more of them.

`backend.keystoreBalance()` contains the same code as this function. How about calling this function there? So instead of

```
keystoreTotalBalance := new(big.Rat)
for _, account := range accountList { ... }
```

it could be

```
keystoreTotalBalance, err := backend.AccountsFiatBalance(accounts, fiatUnit)
...
```

frontends/web/src/components/banners/backup.tsx
```diff
@@ -0,0 +1,80 @@
...
+          })}
+      />
+      {keystore.connected ? (
+        <Link to={deviceSettingsURL}>{t('account.backupReminderLink')}</Link>
+      ) : (
+        <>
+          {t('connectKeystore.promptWithNameForBackup',
+            {
+              name: keystore.name
+            })}
```
benma: We have a connectKeystore() function with an associated UX that prompts to insert/unlock a keystore. It is currently tied to an account, but it could probably be refactored without much effort to work based on the root fingerprint only. 

See also:
- https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/63286f4a0618371f92e57f543b025c12215ddde5/backend/accounts.go#L908
- https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/63286f4a0618371f92e57f543b025c12215ddde5/frontends/web/src/api/account.ts#L455

Would it be better UX to have a link, which when clicked, prompts to enter the keystore, and then redirects to the relevant workflow?

frontends/web/src/components/banners/backup.tsx
benma: We only have the bb02 now, so there could be a direct link, but it would break modularity, as there could be different types of devices with different settings/endpoints/etc.

Your highlight attempt is not in this PR is it? It can probably be done with a GET or URL param, maybe @thisconnect can could worry about the UX. I'd leave the redirect to the settings as you did and create an issue to improve UX here if needed for @thisconnect or @shonsirsha to handle after this is merged.

frontends/web/src/components/banners/backup.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type BackupReminderProps = {
+  keystore: TKeystore;
+  deviceID: string;
+  trigger: number;
+}
+
+export const BackupReminder = ({ keystore, deviceID, trigger }: BackupReminderProps) => {
+  const { t } = useTranslation();
+  const [bannerResponse, setBannerResponse] = useState<TShowBackupBannerResponse|null>(null);
```
benma: Why does the linter not complain about missing spaces around `|` @thisconnect ? :smile:

frontends/web/src/routes/account/summary/accountssummary.tsx
benma: Why don't you use `accountsBalanceSummary` as the trigger instead? That var even contains the fiat total per keystore (in the current main fiat). I agree `Date.now()` is ugly :joy:

frontends/web/src/components/banners/backup.tsx
thisconnect: The linter currently ignores TypeScript.

frontends/web/src/routes/account/summary/accountssummary.tsx
bznein: I don't actually need the total in the banner, because the total is only used by the backend to determine whether the banner needs to be displayed.  but yes, I need to revert to USD as well.

(also, AFAIK `accountsBalanceSummary?.keystoresBalance[keystore.rootFingerprint].total` has the total as a string, and it would be ugly to have logic to decide whether to display the banner in the frontend AND operating on strings rather than numbers, imho)

frontends/web/src/components/banners/backup.tsx
bznein: > Your highlight attempt is not in this PR is it? 

Yeah that wasn't clear, but it was ugly enough to NOT include it in the PR :)

frontends/web/src/routes/account/summary/accountssummary.tsx
bznein: I have no idea why it doesn't (looking at the console, I see no errors with a watch-only wallet but not connected BB02), but I will change it anyway to be more robust

frontends/web/src/routes/account/summary/accountssummary.tsx
bznein: Done. I don't actually need the fiat total as discussed in another comment, but looks definitely much better

backend/accounts.go
bznein: Done! I had to slightly refactor the AccountsFiatBalance to als oreturn the coin balance otherwise the entire loop would still be needed, I think it looks fine now, lmk

frontends/web/src/components/banners/backup.tsx
bznein: It looks much better, thanks!

Right now, what I did is passing `accounts[0].code` to the backup banner since `connectKeystore` still only uses that to extract the keystore fingerprint and then uses that internally. Let me know if you want me to refactor the `connectKeystore` function in this/another PR!

benma: accounts[0] might not be attached to the same keystore, i.e. it could prompt for the wrong keystore.

If you are going to use the prompt, you should probably make a commit/PR to refactor it beforehand, and rebase this PR on top of that.

bznein: > accounts[0] might not be attached to the same keystore, i.e. it could prompt for the wrong keystore.

```  
{accountsByKeystore.map(({ keystore, accounts }) => (
              <BackupReminder
                key={keystore.rootFingerprint}
                keystore={keystore}
                accountsBalanceSummary={accountsBalanceSummary}
                accountCode={accounts[0].code}
              />
```

Since I'm looping through `accountsByKestore` shouldn't `accounts[0]` be attached to the right keystore? Still happy to create a different PR! Just wanted to clarify!

bznein: Refactor PR at https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3453

bznein: Rebased on that PR!

thisconnect:
fyi: I am planning to open a PR to change the show-recovery-words popup from using a popup to its own route & view (like enable passphrase). With that the banner can deeplink into show-recovery-words workflow and it looks nicer.

backend/handlers/handlers.go
```diff
@@ -1502,6 +1505,53 @@ func (handlers *Handlers) getOnline(r *http.Request) interface{} {
...
+	if err != nil {
+		handlers.log.WithError(err).Error("Could not retrieve fiat balance for account")
+		return response{
+			Success: false,
+		}
+	}
+
+	return response{
+		Success: true,
+		Show:    balance.Cmp(big.NewRat(10, 1)) > 0,
```
benma: That does not sound like it's 1000 :P

frontends/web/src/locales/en/app.json
```diff
@@ -2,6 +2,8 @@
   "account": {
     "account": "Account",
     "accounts": "Accounts",
+    "backupReminder": "Your wallet <strong>{{name}}</strong> just passed {{fiat}} 1000!\n\nWe recommend creating a paper backup for extra protection. It's quick and simple.",
```
benma: I'd remove "just", as it will be shown to any user who has had more than 1000 even for a long time.

Also, please make the variable contain "USD 1000", including the amount - we don't need the amount to be translated, and when we want to change it re-translate it.

Also, would be good to apply `coin.FormatAsCurrency()` to the actual threshold in the backend and `formatLocalizedAmount()` in the frontend, as we do with other fiat amounts.

frontends/web/src/components/banners/backup.tsx
```diff
@@ -0,0 +1,82 @@
...
+  if (!bannerResponse || !bannerResponse.success) {
+    return null;
+  }
+
+  const maybeNavigateToSettings = async (e: MouseEvent<HTMLAnchorElement>) => {
+    e.preventDefault();
+    const connectResult = await connectKeystore(keystore.rootFingerprint);
+    if (connectResult.success) {
+      const devices = await getDeviceList();
+      const deviceSettingsURL = `/settings/device-settings/${Object.keys(devices)[0]}`;
```
benma, line 66: Check that Object.keys(devices) is not empty, or you will redirect to `undefined`

benma (changes requested):
changelog

backend/handlers/handlers.go
bznein: Ouch! Thanks for spotting it!

frontends/web/src/components/banners/backup.tsx
bznein: Done

frontends/web/src/locales/en/app.json
bznein: I'm not entirely sure I follow what's the proper way to handle this. I made some changes but I'm not sure, especially on why we need to format in the backend as we don't really use the string there. However, please take a look and let me know if what I did now works for you, thanks!

frontends/web/src/locales/en/app.json
```diff
@@ -2,6 +2,8 @@
   "account": {
     "account": "Account",
     "accounts": "Accounts",
+    "backupReminder": "Your wallet <strong>{{name}}</strong> passed {{fiat}} {{threshold}}!\n\nWe recommend creating a paper backup for extra protection. It's quick and simple.",
```
benma, line 5: Thanks, imho this is much better.

benma (approved):
utACK

PR #3454, opened 2025-07-18
backend/bitbox02: bundle v9.23.1 firmwares

benma:


NickeZ (approved):
reproduced ACK

PR #3455, opened 2025-07-21
frontend: scroll up after accepting (widget) terms

shonsirsha:
Currently, after accepting a widget's terms the scroll position will stay in the bottom. This is not ideal because to use the widget we'd need to manually scroll up. This happens because the route did not change when the screen changes from the terms 'screen' to the widget 'screen'.

This commit intends to ensure that after every click on the accept btn of any widget terms, the scrollable container scrolls up, mimicking a scroll reset as if the user were to change route.

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -203,7 +203,7 @@ export const App = () => {
                 return null;
               })
             }
-            <div className={styles.contentContainer}>
+            <div id="content-container" className={styles.contentContainer}>
               <GlobalBanners />
               <AppRouter
                 accounts={accounts}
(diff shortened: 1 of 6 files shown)


frontends/web/src/app.tsx
```diff
@@ -203,7 +203,7 @@ export const App = () => {
                 return null;
               })
             }
-            <div className={styles.contentContainer}>
+            <div id="content-container" className={styles.contentContainer}>
```
thisconnect, line 206: Nit: In the future we should look into a different solution that doesn't need  an id and can be reused, maybe in a component.

thisconnect (approved):
untested LGTM

PR #3456, opened 2025-07-21
frontend: fix scrolling long dialog content on ios

thisconnect:
Using height: 100% does not work as expected on iOS and is the
reason that transaction details cannot be scrolled properly.

Also adjusted dialog paddings on ios to maintain usability


sutterseba (approved):
tested on iPhone, LGTM

PR #3459, opened 2025-07-22
ios: fix dialog X button cant be tapped on ipad landscape

shonsirsha:
The "X" button falls in the safe area making it impossible to tap on ipad landscape orientation. Also enabled tap-to-close on the overlay for all screen sizes (not just mobile).


sutterseba (approved):
tested (only on iPad) LGTM

PR #3467, opened 2025-07-24
qt: fix offline banner false positive

benma:
Beerosagos got the offline banner even though he was online. The reachability value for him was `Site`, which seems wrong. We only check `Disconnected` now to avoid false positives.


bznein (approved):
utACK  thanks for the quick fix!

PR #3471, opened 2025-07-28
[WIP - do not review] backend: fetch multiple ETH balances at the same time.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests


backend/accounts_test.go
```diff
@@ -179,11 +179,13 @@ func TestSortAccounts(t *testing.T) {
 		{Code: "acct-tbtc", CoinCode: coinpkg.CodeTBTC},
```
bznein, line 179: Changes in this file are done because `createAndAddAccount`, as the docs says, needs to be called while having ownership of the `accountsAndKeystoreLock`. This was not true before but not noticed since there was no concurrent access by the backend, as we do now

PR #3473, opened 2025-07-30
backend/handlers: fix header sync tx timestamp

Beerosagos:
The use of creation timestamp when the tx timestamp was not available caused the app to show the current date for all the btc/ltc transactions during the header syncing. This removes the use of the tx creation timestamp, as it doesn't seem to be really useful anyway.

Now during header sync the tx date is not showed at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix wrong btc/ltc transaction timestamp during header sync
 
 ## v4.48.1
 - Bundle BitBox02 firmware version v9.23.1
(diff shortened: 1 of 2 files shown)


benma (commented):
I think CreatedTimestamp was introduced to give unconfirmed transactions a timestamp:

https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/952/commits/029a54f04af98f7fa23f0fce9f22fb492a9c18d7

I haven't tested, but wouldn't this commit now show any date/time on unconfirmed transactions? 

And maybe mess up the fiat display in unconfirmed tranasctions then too (as `amountAtTime` depends on it)?

benma (approved):
utACK

PR #3478, opened 2025-08-01
backend: close Electrum/failover connections too on shutdown

benma:
Without this, if no backends/servers are configured, headers download would be stuck on some call that keeps failing over or retrying forever, and headers.Close() would be blocked from finishing.

This is a quick fix, there is still a delay to shut down until some Electrum call in headers.go fails (probably up to 30s due to the retry timeout?).

More things to do:

- Add a ctx var for every Electrum call that should get cancelled on shutdown to not delay
- Do not block closing headers on the same lock which is held during a blockchain call (use a different lock, or by the above ctx solution)

### DIFF
--- a/backend/coins/btc/coin.go
+++ b/backend/coins/btc/coin.go
@@ -312,12 +312,14 @@ func (coin *Coin) ValidateSilentPaymentAddress(address string) error {
 // Close implements coinpkg.Coin.
 func (coin *Coin) Close() error {
 	coin.log.Info("closing coin")
+	if coin.blockchain != nil {
+		coin.blockchain.Close()
+	}
 	if coin.headers != nil {
 		coin.log.Info("closing headers")
 		if err := coin.headers.Close(); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


Beerosagos (approved):
tACK!

> - Do not block closing headers on the same lock which is held during a
blockchain call (use a different lock, or by the above ctx solution)

Can you explain this? I don't understand which lock you are referring to

benma:
> tACK!
> 
> > * Do not block closing headers on the same lock which is held during a
> >   blockchain call (use a different lock, or by the above ctx solution)
> 
> Can you explain this? I don't understand which lock you are referring to

Sure, this one:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/backend/coins/btc/headers/headers.go#L488

`Close()` is blocked until the lock is released, which is also acquried when a header batch is being downloaded. If there is no Electrum backend, and because we currently don't time out Electrum calls that fail (we just keep "failing over" to another backend, but there are none in this case), downloading the headers batch never finishes and never releases the lock.

PR #3479, opened 2025-08-04
frontend: fix connection issues link style (bluetooth)

shonsirsha:
previously, this link was absolutely positioned causing inconsistent placement.


sutterseba (approved):
tested on iPhone

PR #3480, opened 2025-08-04
frontend: show red dot firmware upgrade for mobile

shonsirsha:
The red dot will start to show from the bottom menu (More), then Settings, and finally also on Manage Device on mobile.
Screenshot 2025-08-04 at 13 24 42

Screenshot 2025-08-04 at 07 21 11

### DIFF
--- a/frontends/web/src/hooks/api.ts
+++ b/frontends/web/src/hooks/api.ts
@@ -69,10 +69,10 @@ export const useLoad = <T>(
   const [response, setResponse] = useState<T>();
   const mounted = useMountedRef();
   const load = () => {
+    setResponse(undefined);
     if (apiCall === null) {
       return;
     }
-    setResponse(undefined);
     apiCall().then((data) => {
       if (mounted.current) {
         setResponse(data);
--- a/frontends/web/src/hooks/api.test.ts
+++ b/frontends/web/src/hooks/api.test.ts
@@ -60,6 +60,30 @@ describe('hooks for api calls', () => {
 
       await waitFor(() => expect(mockApiCall).toHaveBeenCalledTimes(4));
     });
+
+    it('clears response when apiCall becomes null', async () => {
+      const mockApiCall = vi.fn().mockImplementation(() => Promise.resolve({ canUpgrade: true }));
+
+      const { result } = renderHook(() => {
+        const [deviceID, setDeviceID] = useState('test-device');
+        const [isBitBox02, setIsBitBox02] = useState(true);
+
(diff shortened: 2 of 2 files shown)


thisconnect (changes requested):
found a small bug, in the desktop version the red dot disappears once the device is disconnected, but in the "More" and "Settings" the red dot stays.

frontends/web/src/routes/settings/components/tabs.tsx
```diff
@@ -71,17 +71,23 @@ export const Tab = ({
...
 }: TTab) => {
   const navigate = useNavigate();
+  const { pathname } = useLocation();
   const upgradeDot = canUpgrade ? (
     <RedDot className={styles.canUpgradeDot} width={8} height={8} />
   ) : null;
 
+  const isManageDeviceItem = url.includes('device-settings');
+  const mobileSettings = pathname === '/settings';
+  const showRedDotOnMobile = isManageDeviceItem && mobileSettings && canUpgrade;
```
thisconnect: Only showing the red-dot if `url.includes('device-settings');` makes sense so only links to device settings have reddot, but why does it need to check if pathname is '/settings' ?

shonsirsha: Hm.. good catch @thisconnect  -- for some reason I mistakenly thought we reuse `tabs.tsx` inside the `/device-settings` view itself. Since it's only used in `/settings` then this check is redundant.

frontends/web/src/hooks/api.ts
```diff
@@ -69,10 +69,10 @@ export const useLoad = <T>(
   const [response, setResponse] = useState<T>();
   const mounted = useMountedRef();
   const load = () => {
+    setResponse(undefined);
     if (apiCall === null) {
       return;
     }
```
shonsirsha, line 75: @thisconnect currently, the `response` state is only reset to `undefined` when `apiCall !== null`, due to the early return. Is that intentional? tbh seems like a bug to me. If the hook remains mounted and `apiCall` becomes `null`, it will still return the old response value from when `apiCall` wasn't `null` which causes the persisting red-dot bug.

The change I made here ensures that `response` is always reset, regardless of the `apiCall` value. WDYT?

--- 

Regarding the red-dot bug: 
1. this bug was present on the bottom navigation and "More" screen. 
2. However, as you mentioned it's not an issue in the `/settings` page with `TabWithVersionCheck`. From my observation, this is bc the component is remounted when the device is unplugged (likely due to conditional rendering).

PR #3486, opened 2025-08-07
Dockerfile: update Android NDK/SDK/targetSdkVersion/etc.

Beerosagos:
Android SDK 35 corresponds to Android 15.

- Update Docker base Android image to 9.3, the latest version on Ubuntu 22.04, which we use
- targetSdkVersion is updated to 35 as Google will enforce this starting August 31st 2025.
- minSdkVersion is updated to 24 (Android 7.0), to avoid build issues with the updated NDK (see the rest of the commit message).
- compileSdk is updated because Gradle 8 only works with that (SDK 35 was automatically downloaded by gradlew on each build in Docker as a result). The Docker image is updated with this so it is not downloaded every time.

NDK is updated to v28, the currently latest version, according to this comment:
https://github.com/golang/go/issues/68754#issuecomment-2523071214. The reason for this is given in
https://developer.android.com/about/versions/15/behavior-changes-all#16-kb:

>Support for 16 KB page sizes
>
>Historically, Android has only supported 4 KB memory page sizes, which has optimized system memory performance for the average amount of total memory that Android devices have typically had. Beginning with Android 15, AOSP supports devices that are configured to use a page size of 16 KB (16 KB devices). If your app uses any NDK libraries, either directly or indirectly through an SDK, then you will need to rebuild your app for it to work on these 16 KB devices.
>
>As device manufacturers continue to build devices with larger amounts of physical memory (RAM), many of these devices will adopt 16 KB (and eventually greater) page sizes to optimize the device's performance. Adding support for 16 KB page size devices enables your app to run on these devices and helps your app benefit from the associated performance improvements. Without recompiling, apps won't work on 16 KB devices in future Android releases.
>
>To help you add support for your app, we've provided guidance on how
>to check if your app is impacted, how to rebuild your app (if
>applicable), and how to test your app in a 16 KB environment using
>emulators (including Android 15 system images for the Android
>Emulator).

NOTE, relevant behavior changes in Android 15:
- Our dataSync foreground service will be timed out and lead to an exception after a total of 6h of background runtime in a 24h window: https://developer.android.com/about/versions/15/behavior-changes-15#datasync-timeout, unless we implement the onTimeout method

References:
- https://developer.android.com/about/versions/15/migration
- https://developer.android.com/about/versions/15/setup-sdk
- https://developer.android.com/about/versions/15/behavior-changes-all
- https://developer.android.com/about/versions/15/behavior-changes-15
- https://developer.android.com/about/versions/15/features

- [ ] updating the Changelog
- [ ] writing unit tests


backend/mobileserver/Makefile
```diff
@@ -2,7 +2,7 @@ include ../../android-env.mk.inc
 
 # Set -glflags to fix the vendor issue with gomobile, see: https://github.com/golang/go/issues/67927#issuecomment-2241523694
 build-android:
-	ANDROID_HOME=${ANDROID_SDK_ROOT} gomobile bind -x -a -glflags="-mod=readonly" -ldflags="-s -w" -target android .
+	ANDROID_HOME=${ANDROID_SDK_ROOT} gomobile bind -x -a -glflags="-mod=readonly" -ldflags="-s -w" -target android -androidapi 24 .
```
benma, line 6: Add a comment that this should correspond to `minSdkVersion` in build.gradle

frontends/android/BitBoxApp/app/build.gradle
```diff
@@ -1,12 +1,12 @@
...
 android {
-    buildToolsVersion '34.0.0'
-    compileSdk 34
+    buildToolsVersion '35.0.0'
+    compileSdk 35
     defaultConfig {
         applicationId "ch.shiftcrypto.bitboxapp"
-        minSdkVersion 21
-        targetSdkVersion 34
+        minSdkVersion 24
```
benma, line 10: Add comment that this should match `-androidapi` in `gomobile bind`

PR #3489, opened 2025-08-12
backend/usb: fix potential macOS communication issue with BitBox02

benma:


NickeZ (approved):
utACK

PR #3493, opened 2025-08-13
backend: fetch multiple ETH balances at once.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -308,6 +308,8 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.bluetooth = bluetooth.New(log)
 	backend.bluetooth.Observe(backend.Notify)
 
+	go backend.pollETHBalances()
+
 	return backend, nil
 }
 
@@ -555,6 +557,38 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 	return coin, nil
 }
 
+func (backend *Backend) pollETHBalances() {
+	timer := time.After(0)
+
+	for {
+		<-timer
+		if err := backend.updateETHAccounts(); err != nil {
+			backend.log.WithError(err).Error("could not update ETH accounts")
+		}
+		timer = time.After(eth.PollInterval)
+	}
+}
+
+func (backend *Backend) updateETHAccounts() error {
+	defer backend.accountsAndKeystoreLock.RLock()()
+	backend.log.Debug("Updating ETH accounts balances")
+
+	ethAccounts := make([]*eth.Account, 0, len(backend.accounts))
+	for _, account := range backend.accounts {
+		ethAccount, ok := account.(*eth.Account)
+		if ok {
+			ethAccounts = append(ethAccounts, ethAccount)
+		}
+
+	}
+
+	etherScanClient := etherscan.NewEtherScan("1", backend.etherScanHTTPClient)
+	if backend.Testing() {
+		etherScanClient = etherscan.NewEtherScan("11155111", backend.etherScanHTTPClient)
+	}
+	return eth.UpdateBalances(ethAccounts, etherScanClient)
+}
+
 // ManualReconnect triggers reconnecting to Electrum servers if their connection is down.
 // Only coin connections that were previously established are reconnected.
 // Calling this is a no-op for coins that are already connected.
@@ -592,12 +626,8 @@ func (backend *Backend) ManualReconnect(reconnectETH bool) {
 	}
 	if reconnectETH {
 		backend.log.Info("Reconnecting ETH accounts")
-		for _, account := range backend.accounts {
-			ethAccount, ok := account.(*eth.Account)
-			if !ok {
-				continue
-			}
-			ethAccount.EnqueueUpdate()
+		if err := backend.updateETHAccounts(); err != nil {
+			backend.log.WithError(err).Error("could not update ETH accounts")
 		}
 	}
 }
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -49,7 +49,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var pollInterval = 5 * time.Minute
+// PollInterval is the interval at which the account is polled for updates.
+var PollInterval = 5 * time.Minute
 
 func isMixedCase(s string) bool {
 	return strings.ToLower(s) != s && strings.ToUpper(s) != s
@@ -215,7 +216,25 @@ func (account *Account) Initialize() error {
 }
 
 func (account *Account) poll(initDone func()) {
-	timer := time.After(0)
+	updateAccount := func() {
+		balance, err := account.coin.client.Balance(context.TODO(), account.address.Address)
+		if err != nil {
+			account.log.WithError(err).Error("could not get balance")
+			return
+		}
+		if err := account.update(balance); err != nil {
+			account.log.WithError(err).Error("error updating account")
+			account.SetOffline(err)
+		} else {
+			account.SetOffline(nil)
+		}
+		if initDone != nil {
+			initDone()
+			initDone = nil
+		}
+	}
+
+	updateAccount()
 	for {
 		select {
 		case <-account.quitChan:
@@ -224,21 +243,10 @@ func (account *Account) poll(initDone func()) {
 			select {
 			case <-account.quitChan:
 				return
-			case <-timer:
 			case <-account.enqueueUpdateCh:
 				account.log.Info("extraordinary account update invoked")
 			}
-			if err := account.update(); err != nil {
-				account.log.WithError(err).Error("error updating account")
-				account.SetOffline(err)
-			} else {
-				account.SetOffline(nil)
-			}
-			if initDone != nil {
-				initDone()
-				initDone = nil
-			}
-			timer = time.After(pollInterval)
+			updateAccount()
 		}
 	}
 }
@@ -334,7 +342,7 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
 	return transactions, nil
 }
 
-func (account *Account) update() error {
+func (account *Account) update(balance *big.Int) error {
 	defer account.updateLock.Lock()()
 	defer account.Synchronizer.IncRequestsCounter()()
 
@@ -399,20 +407,13 @@ func (account *Account) update() error {
 		}
 	}
 
-	var balance *big.Int
+	pendingAmount := pendingTxsAmount(outgoingTransactionsData, account.coin.erc20Token != nil)
 	if account.coin.erc20Token != nil {
 		balance, err = account.coin.client.ERC20Balance(account.address.Address, account.coin.erc20Token)
 		if err != nil {
 			return errp.WithStack(err)
 		}
-	} else {
-		balance, err = account.coin.client.Balance(context.TODO(), account.address.Address)
-		if err != nil {
-			return errp.WithStack(err)
-		}
 	}
-
-	pendingAmount := pendingTxsAmount(outgoingTransactionsData, account.coin.erc20Token != nil)
 	account.balance = coin.NewAmount(balance.Sub(balance, pendingAmount))
 
 	return nil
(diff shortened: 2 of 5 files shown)


backend/coins/eth/etherscan/etherscan.go
```diff
@@ -457,6 +458,48 @@ func (etherScan *EtherScan) Balance(ctx context.Context, account common.Address)
...
+	params := url.Values{}
+	params.Set("module", "account")
+	params.Set("action", "balancemulti")
+
+	addresses := make([]string, len(accounts))
+	for i, account := range accounts {
+		addresses[i] = account.Hex()
+	}
+
+	params.Set("address", strings.Join(addresses, ","))
```
benma: https://docs.etherscan.io/api-endpoints/accounts

> up to 20 addresses per call

You should batch in chunks of 20 addresses then.

backend/coins/eth/account.go
```diff
@@ -399,20 +407,13 @@ func (account *Account) update() error {
 		}
 	}
 
-	var balance *big.Int
+	pendingAmount := pendingTxsAmount(outgoingTransactionsData, account.coin.erc20Token != nil)
 	if account.coin.erc20Token != nil {
 		balance, err = account.coin.client.ERC20Balance(account.address.Address, account.coin.erc20Token)
```
benma: Quite confusing to take a `balance` arg and then overwrite it. I'd centralize this and move the erc20 balance preparation also to the backend loop, so it's easier to follow and understand.

backend/coins/eth/account.go
```diff
@@ -334,7 +342,7 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
 	return transactions, nil
 }
 
-func (account *Account) update() error {
+func (account *Account) update(balance *big.Int) error {
```
benma: add docstring to explain the new argument

backend/backend.go
```diff
@@ -555,6 +557,38 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 	return coin, nil
 }
 
+func (backend *Backend) pollETHBalances() {
```
benma: This loop runs forever - should exit it upon backend.Close(). You would probably need to introduce quit channel, similar to https://github.com/BitBoxSwiss/block-client-go/pull/8.

backend/backend.go
```diff
@@ -555,6 +557,38 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
...
+	ethAccounts := make([]*eth.Account, 0, len(backend.accounts))
+	for _, account := range backend.accounts {
+		ethAccount, ok := account.(*eth.Account)
+		if ok {
+			ethAccounts = append(ethAccounts, ethAccount)
+		}
+
+	}
+
+	etherScanClient := etherscan.NewEtherScan("1", backend.etherScanHTTPClient)
```
benma: Was wondering about the rate limiting, and now see that https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/e509f1838ab5a81f571fe843919eb3e1ee1a755f#diff-ceb7e04e1612df83dc6c0f93e8dd3646a846c5949195088be13031f700264b67 introduced a bug:

Before, the rate limiting was enforced across all instances of the etherscan client, and now it's per-instance. This means that us making a separate instance per ERC-20 token and making new instances here goes over our intended rate limit.

Should probably instantiate the `rate.Limiter` instance once and pass it to all instances.

backend/backend.go
```diff
@@ -555,6 +557,38 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
...
+	if backend.Testing() {
```
benma: As mentioned before, this is not a robust check - we used to have multiple testnets with different chainIDs in the past (and might again in the future). 

eth.coin.ChainID() contains the chainID, could just use that?

backend/backend.go
```diff
@@ -592,12 +626,8 @@ func (backend *Backend) ManualReconnect(reconnectETH bool) {
 	}
 	if reconnectETH {
 		backend.log.Info("Reconnecting ETH accounts")
-		for _, account := range backend.accounts {
-			ethAccount, ok := account.(*eth.Account)
-			if !ok {
-				continue
-			}
-			ethAccount.EnqueueUpdate()
+		if err := backend.updateETHAccounts(); err != nil {
```
benma: This makes a call in parallel with the existing loop and does not obey the poll interval, right? 

Since the backend has a poll loop, why is there even a poll loop left in account.go? Could move all the polling logic of account.go (including enqueing an update, quitting etc) there instead of keeping two separate poll loops.

backend/coins/eth/account.go
```diff
@@ -215,7 +216,25 @@ func (account *Account) Initialize() error {
 }
 
 func (account *Account) poll(initDone func()) {
-	timer := time.After(0)
+	updateAccount := func() {
+		balance, err := account.coin.client.Balance(context.TODO(), account.address.Address)
```
benma: Why is there a poll loop here and in backend.go? Isn't the account now being updated twice every 5mins (pollInterval)?

benma (changes requested):

backend/coins/eth/account.go
bznein: This is not a real loop anymore, as it doesn't use the timer; it just waits for the enqueueUpdateChan to perform the update of a single account.I can probably see if it can be removed and merged into the backend loop though

backend/backend.go
bznein: > This makes a call in parallel with the existing loop and does not obey the poll interval, right?

Correct, I will try and merge everything in the backend loop (see other comment)

backend/backend.go
bznein: > eth.coin.ChainID() contains the chainID, could just use that?

Sure, thanks for pointing out; just ot make sure I understand: it is theoretically possible that in the loop above, different eth account will have different chainIDs, right? In that case I need to batch the calls to updateBalances and provide a different client each time

benma: Yes indeed, that is possible. It's not happening right now afaik, as we only have on mainnet and one testnet chain, but as I said, we used to have multiple testet chains, we might add more EVM mainnet chains, etc.

backend/coins/eth/account.go
benma: Ah I missed that it has no timer anymore. But as per the other comment, since the backend poll loop will also need a way to quit and enqueue, I guess we can do away with this function altogether.

Be mindful of `initDone()`, which takes care of the ETH account showing a loading screen until the first results are available. I guess this is a very clumsy way of achieving that (inherited from the BTC account where there are a lot more calls to sync), maybe one day we can do away with the sync counter.

backend/backend.go
```diff
@@ -308,6 +315,8 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.bluetooth = bluetooth.New(log)
 	backend.bluetooth.Observe(backend.Notify)
 
+	go backend.pollETHBalances()
```
benma: Should be moved to `Start()`

backend/coins/eth/account.go
```diff
@@ -1032,3 +1029,59 @@ func (account *Account) EnqueueUpdate() {
 	default:
 	}
 }
+
+// UpdateBalances updates the balances of the accounts in the provided slice.
+func UpdateBalances(accounts []*Account, etherScanClient *etherscan.EtherScan) error {
+	ethNonErc20Addresses := make([]ethcommon.Address, 0, len(accounts))
+	ethErc20Addresses := make([]ethcommon.Address, 0, len(accounts))
```
benma: This var is unused.

backend/backend.go
```diff
@@ -555,6 +564,42 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 	return coin, nil
 }
 
+func (backend *Backend) pollETHBalances() {
+	timer := time.After(0)
+
+	for {
+		select {
+		case <-backend.quit:
```
benma, line 585: Should do a nested check for this like in https://github.com/BitBoxSwiss/block-client-go/pull/8/files#diff-8d7b38d7d1da47b2e7c4b04453a94761e959c337444e1aa135dcaa88ac0f4392R174

The reason is that if timer and quit are available at the same time, you want to prioritize quit.

backend/backend.go
```diff
@@ -933,6 +974,8 @@ func (backend *Backend) Close() error {
 	if len(errors) > 0 {
 		return errp.New(strings.Join(errors, "; "))
 	}
+
+	backend.quit <- struct{}{}
```
benma: That should be `close(backend.quit)`.

bznein:
@benma I should have addressed all your comments, this is ready for another review :)

(also, should we add a changelog entry for this?)

backend/backend.go
```diff
@@ -555,6 +568,68 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
...
+				if err != nil {
+					backend.log.WithError(err).Errorf("could not query balance for account %s", account.Config().Config.Name)
+					continue
+				}
+				if err := account.Update(balance); err != nil {
+					backend.log.WithError(err).Errorf("could not update account %s", account.Config().Config.Name)
+					account.SetOffline(err)
+				} else {
+					account.SetOffline(nil)
+				}
```
benma: why not call `eth.UpdateBalances(account, ....)` instead? De-duplicate.

backend/backend.go
```diff
@@ -555,6 +568,68 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
 	return coin, nil
 }
 
+func (backend *Backend) pollETHAccounts() {
```
benma, line 574: When the app is started, time.After(0) will be handled immediately, likely before any accounts are loaded. When the accounts are then added (or when an ETH account is added manually later, etc), they will not be initialized for up to 5 minutes...

benma (changes requested):

backend/backend.go
```diff
@@ -555,6 +568,68 @@ func (backend *Backend) Coin(code coinpkg.Code) (coinpkg.Coin, error) {
...
+
+	for {
+		select {
+		case <-backend.quit:
+			return
+		default:
+			select {
+			case <-backend.quit:
+				return
+			case account := <-backend.enqueueUpdateForAccount:
```
benma, line 591: Here and in the cae below, run the case bodies in a goroutine, otherwise one account calling `EnqueueUpdate()` will be locked waiting until the rest has processed.

backend/backend.go
benma: This is still open - you want to kick the pollETHAccounts() loop into action (with a channel) so the next call is scheduled for 5min after this one

backend/coins/eth/account.go
```diff
@@ -1028,7 +992,75 @@ func (account *Account) MatchesAddress(address string) (bool, error) {
 // EnqueueUpdate enqueues an update for the account.
 func (account *Account) EnqueueUpdate() {
 	select {
-	case account.enqueueUpdateCh <- struct{}{}:
+	case account.enqueueUpdateCh <- account:
 	default:
```
benma: Now that the same channel handles multiple accounts, you should remove the default statement. Otherwise, if one account is enqueued to be updated, but not processed yet, further ones will be ignored. I suspect that if you load many ETH accounts like this, not all of them will actually be initialized.

backend/coins/eth/account.go
```diff
@@ -208,41 +208,12 @@ func (account *Account) Initialize() error {
 	)
 
 	account.coin.Initialize()
-	done := account.Synchronizer.IncRequestsCounter()
-	go account.poll(done)
+	account.initDone = account.Synchronizer.IncRequestsCounter()
+	account.EnqueueUpdate()
```
benma: `go account.EnqeueuUpdate()` to not block here in case the channel is full (considering my other comment where EnqueueUpdate() should drop the default case)

backend/coins/eth/account.go
```diff
@@ -1028,7 +992,75 @@ func (account *Account) MatchesAddress(address string) (bool, error) {
...
 func (account *Account) EnqueueUpdate() {
 	select {
-	case account.enqueueUpdateCh <- struct{}{}:
+	case account.enqueueUpdateCh <- account:
 	default:
 	}
 }
+
+// QueryBalance queries the coin clientfor the balance of the account.
+func (account *Account) QueryBalance() (*big.Int, error) {
```
benma: Not needed anymore

backend/coins/eth/account.go
```diff
@@ -1028,7 +992,75 @@ func (account *Account) MatchesAddress(address string) (bool, error) {
...
+	balances, err := etherScanClient.Balances(context.TODO(), ethNonErc20Addresses)
+	if err != nil {
+		return errp.WithStack(err)
+	}
+
+	for _, account := range accounts {
+		if account.isClosed() {
+			continue
+		}
+		address, err := account.Address()
```
benma, line 1049: If an error happens anywhere here, we probably want to set the account state into critical error or set it to offline, and continue with the other accounts. Otherwise one account's error will prevent all subsequent ones from being updated.

backend/backend.go
```diff
@@ -240,6 +242,15 @@ type Backend struct {
...
 	isOnline atomic.Bool
+
+	// quit is used to indicate to running goroutines that they should stop as the backend is being closed
+	quit chan struct{}
+
+	// enqueueUpdateForAccount is used to enqueue an update for an account.
+	enqueueUpdateForAccount chan *eth.Account
+
+	// updateETHAccountsCh is used to trigger an update of all ETH accounts.
+	updateETHAccountsCh chan struct{}
```
benma, line 253: I mentioned somewhere that we could wrap this into a dedicated struct and move all related methods there to pollute the backend less with ETH specifics. `ethUpdater *ETHUpdater` for example. Then unit tests could be added too more easily. Let's do this after this is merged?

bznein: Yeah can do of course in a next PR to keep the scope smaller. I

backend/coins/eth/account.go
```diff
@@ -1027,8 +991,67 @@ func (account *Account) MatchesAddress(address string) (bool, error) {
...
+				account.SetOffline(err)
+				continue
+			}
+		} else {
+			var ok bool
+			balance, ok = balances[address.Address]
+			if !ok {
+				errMsg := fmt.Sprintf("Could not find balance for address %s", address.Address.Hex())
+				account.log.Error(errMsg)
+				account.SetOffline(errp.Newf(errMsg))
```
benma: `errp.New` not `Newf`

bznein: Thanks, I wonder why the linter didn't get this one!

PR #3494, opened 2025-08-13
frontend: show selected coin's address on confirmation screen

strmci:
Display the selected coin’s address on the confirmation screen to make it clearer which address is being used. UTXOs are now grouped under their address.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/account/send/components/confirm/confirm.tsx
+++ b/frontends/web/src/routes/account/send/components/confirm/confirm.tsx
@@ -23,6 +23,7 @@ import { View, ViewContent, ViewHeader } from '@/components/view/view';
 import { Message } from '@/components/message/message';
 import { PointToBitBox02 } from '@/components/icon';
 import { FiatValue } from '../fiat-value';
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/account/send/components/confirm/confirm.tsx
```diff
@@ -23,6 +23,7 @@ import { View, ViewContent, ViewHeader } from '@/components/view/view';
 import { Message } from '@/components/message/message';
 import { PointToBitBox02 } from '@/components/icon';
 import { FiatValue } from '../fiat-value';
+import { TSelectedUTXOs } from '../../utxos';
```
thisconnect: tiny nit: you could tell typescript that this is only a type.

```suggestion
import type { TSelectedUTXOs } from '../../utxos';
```

strmci: updated, thanks

PR #3496, opened 2025-08-14
backend: only compute next nonce when needed.

bznein:
Instead of computing the next nonce in every single ETH account, we postpone its update to right before we need it.

The new method that updates the `nextNonce` field is called before any access to the `account.nextNonce` field.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -95,7 +95,6 @@ type Account struct {
 	updateLock   locker.Locker
 	balance      coin.Amount
 	blockNumber  *big.Int
-	nextNonce    uint64
 	transactions []*accounts.TransactionData
 
 	// if not nil, SendTx() will sign and send this transaction. Set by TxProposal().
@@ -304,6 +303,21 @@ func (account *Account) updateOutgoingTransactions(tipHeight uint64) {
 	}
 }
 
+func (account *Account) confirmedTransactions() ([]*accounts.TransactionData, error) {
+	var confirmedTransactions []*accounts.TransactionData
+	transactionsSource := account.coin.TransactionsSource()
+	if transactionsSource != nil {
+		var err error
+		confirmedTransactions, err = transactionsSource.Transactions(
+			account.blockNumber,
+			account.address.Address, account.blockNumber, account.coin.erc20Token)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return confirmedTransactions, nil
+}
+
 // outgoingTransactions gets all locally stored outgoing transactions. It filters out the ones also
 // present from the transactions source.
 func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData) (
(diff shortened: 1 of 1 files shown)


bznein:
ping @benma :)

backend/coins/eth/account.go
```diff
@@ -334,6 +349,35 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
 	return transactions, nil
 }
 
+func (account *Account) updateNextNonce() error {
```
benma: There is no need to store the nonce in the account struct anymore, you can just return it from this function (and rename the function accordingly).

backend/coins/eth/account.go
```diff
@@ -334,6 +349,35 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
...
+	account.nextNonce = nodeNonce
+
+	confirmedTransactions, err := account.confirmedTransactions()
+	if err != nil {
+		return errp.WithStack(err)
+	}
+	outgoingTransactions, err := account.outgoingTransactions(confirmedTransactions)
+	if err != nil {
+		return errp.WithStack(err)
+	}
```
benma: This is new behavior - the transactions are now fetched from etherscan while filling out the send form. I believe the `newTx()` function is called every time the user updates the send form :see_no_evil: So it's one more call everytime the user updates the send form (which is when `newTx()` is called). With the rate limit of ~4calls/sec it could also cause a noticeable delay when filling out the form (I didn't test).

Instead, you could use `account.transactions`.

benma (changes requested):

backend/coins/eth/account.go
```diff
@@ -334,6 +348,28 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
...
+	nodeNonce, err := account.coin.client.PendingNonceAt(context.TODO(), account.address.Address)
+	if err != nil {
+		return 0, err
+	}
+	nextNonce = nodeNonce
+
+	// In case the nodeNonce is not up to date, we fall back to our stored last nonce to compute the
+	// next nonce.
+	if len(account.transactions) > 0 {
+		localNonce := *account.transactions[len(account.transactions)-1].Nonce + 1
```
benma: There are two potential bugs here, one might have been there before:

First

Before, we used outgoingTransactions to check the highest nonce, and that list was ordered by nonce

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/e6f0142eb31ca2ee6bbe8ab0a62ce0065090e959/backend/coins/eth/db/interface.go#L30-L32

We want the highest nonce, and since the list is sorted descending, should we not have used `outgoingTransactions[0]` instead of the last element?

But now we use `account.transactions`, which is sorted by number of confirmations, not by nonce, so that does not work anyway. Maybe best to use `slices.MaxFunc(...)` and just get the max nonce that way?

Or maybe even better: use only the locally stored outgoing txs:

```
outgoingTranscations := account.outgoingTransactions(nil)
if len(outgoingTransactions) > 0 {
    localNonce := outgoingTranscations[0].Nonce + 1
}
````

Second, **important**

If using `account.transactions`, you can only consider nonces of transactions that are outgoing (type=send or type=sendToSelf), as the nonce is for transactions made from the account. Currently you also use incoming transacctions, which have the nonce of the sender. In my test account, this code lead to the next nonce being `4` (b/c the incoming tx had `3` as the nonce), even though my account only used one nonce before, so it should have been `1`). If you use only the outgoingTransactions again like above, that problem should be solved too.

File: backend/coins/eth/account.go [hunk 4]
```diff
@@ -334,6 +348,28 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
 	return transactions, nil
 }
 
+func (account *Account) updateNextNonce() (uint64, error) {
```
benma: Rename? we are not updating a struct field anymore.

benma (changes requested):

backend/coins/eth/account.go
bznein: great catch, thank you!

backend/coins/eth/account.go
```diff
@@ -334,6 +348,33 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
...
+
+	// In case the nodeNonce is not up to date, we fall back to our stored last nonce to compute the
+	// next nonce.
+	outgoingTransactions, err := account.outgoingTransactions(nil)
+	if err != nil {
+		return 0, errp.WithStack(err)
+	}
+
+	if len(outgoingTransactions) > 0 {
+		localNonce := outgoingTransactions[0].Transaction.Nonce()
```
benma: ` + 1`

PR #3498, opened 2025-08-18
ios: fix app flickers on start

shonsirsha:
Fixing it by adding a check in WebView.swift for  a message
sent from the react app signaling react app is ready.

On ready, it reveals the webview smoothly with transition.


frontends/ios/BitBoxApp/BitBoxApp/Assets.xcassets/dbb_logo_light.imageset/dbb_logo_dark.svg
benma, line 1: The image files appear twice in the PR. Is that required, could it be only one set?

frontends/ios/BitBoxApp/BitBoxApp/WebView.swift
```diff
@@ -156,6 +158,16 @@ struct WebView: UIViewRepresentable {
...
     
     class Coordinator: NSObject, WKNavigationDelegate, WKUIDelegate {
+        func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+            // change alpha back to 1.0 after a slight delay
+            // this is to prevent the white flash (flicker) on initial load
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.4) {
+                UIView.animate(withDuration: 0.4) {
+                    webView.alpha = 1.0
+                }
+            }
```
benma: This fixes the loading time of the app to 0.8s. This could be too long for some devices and too short for others, right?

Maybe we can have a smarter loading screen. What is the exact cause of the white flicker?

shonsirsha: > Maybe we can have a smarter loading screen. 

Makes sense

>  What is the exact cause of the white flicker?

I can't say for sure, but I think it's the webview loading nothing as react isn't ready yet (takes like a couple ms ), the webview's bgcolour at this time is white which most likely caused that flicker.

frontends/ios/BitBoxApp/BitBoxApp/Assets.xcassets/dbb_logo_light.imageset/dbb_logo_dark.svg
shonsirsha: Yes! Sorry - that was me playing around, deleted the unused.

frontends/ios/BitBoxApp/BitBoxApp/WebView.swift
```diff
@@ -156,6 +158,39 @@ struct WebView: UIViewRepresentable {
     }
     
     class Coordinator: NSObject, WKNavigationDelegate, WKUIDelegate {
+        func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+            // wait for React app to signal it's ready
+            // done to prevent the white flicker on app open
+            checkReactAppReady(webView: webView)
+        }
+        
+        private func checkReactAppReady(webView: WKWebView, retryCount: Int = 0) {
```
shonsirsha: Instead of using an arbitrary-valued delay, here we're checking for reactAppReady property instead. Check runs recursively until that value is set to true (with max retry).

This property is set by the useAppReady hook from the react app. cc @benma

benma: Nice!

Instead of polling, please call directly into Swift. See this as an example where TS is calling goCall(), and the call is handled in Swift:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/9ef6c607d1051fedf660d92801e0120cea8560d4/frontends/ios/BitBoxApp/BitBoxApp/WebView.swift#L109

This would be better as it requires fewer calls and has an even smaller delay until ready. When the react app is ready, tell Swift to transition.

frontends/web/src/hooks/appready.ts
```diff
@@ -0,0 +1,35 @@
...
+import { TDevices } from '@/api/devices';
+import { useEffect } from 'react';
+
+/**
+ * Hook to signal to native iOS WebView that React app is ready to be shown.
+ * This prevents the white flicker on app launch by keeping the WebView hidden
+ * until the React app has finished loading its initial data.
+ */
+export const useAppReady = (accounts?: IAccount[], devices?: TDevices) => {
+  useEffect(() => {
```
benma, line 26: Would it make sense to add a fallback?

```
useEffect(() => {
     const timer = setTimeout(() => {
        /* send appReady */ 
     }, 3000); 
     return () => clearTimeout(timer);
```

Just in case there is some weird delay in getting `accounts` and `devices` results.

Otherwise code looks good, I'll test it on a device tomorrow.

shonsirsha: Okay - added the fallback to call regardless of the `devices` || `accounts` results.

frontends/web/src/hooks/appready.ts
```diff
@@ -0,0 +1,50 @@
...
+    let appReadySent = false;
+
+    const sendAppReady = () => {
+      if (!appReadySent && window.runningOnIOS && window.webkit?.messageHandlers.appReady) {
```
benma: Please use `runningInIOS()` from env.ts.

frontends/web/src/hooks/appready.ts
```diff
@@ -0,0 +1,50 @@
...
+      sendAppReady();
+      return;
+    }
+
+    // fallback: send appReady after 3 seconds
+    // even if accounts/devices aren't loaded
+    const timer = setTimeout(() => {
+      sendAppReady();
+    }, 3000);
+
```
benma: This should be in a separate `useEffect(..., [])`, it should not reset when accounts/devices change.

frontends/web/src/hooks/appready.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (accounts !== undefined && devices !== undefined) {
+      sendAppReady();
+      return;
+    }
+
+    // fallback: send appReady after 3 seconds
+    // even if accounts/devices aren't loaded
+    const timer = setTimeout(() => {
+      sendAppReady();
```
benma: If the first call failed, this one will do anything due to `appReadySent` being true. Maybe just skip the appReadySent check?

frontends/web/src/hooks/appready.ts
```diff
@@ -0,0 +1,50 @@
...
+import { IAccount } from '@/api/account';
+import { TDevices } from '@/api/devices';
+import { useEffect } from 'react';
+
+/**
+ * Hook to signal to native iOS WebView that React app is ready to be shown.
+ * This prevents the white flicker on app launch by keeping the WebView hidden
+ * until the React app has finished loading its initial data.
+ */
+export const useAppReady = (accounts?: IAccount[], devices?: TDevices) => {
```
benma: can accounts and devices even be undefined at all? In app.tsx they use `useDefault(...)` so it seems they are never undefined. Then this hook could skip these params and just use `useEffect(..., [])`?

frontends/web/src/hooks/appready.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright 2024 Shift Crypto AG
```
benma, line 2: It's 2025 :)

benma (approved):
tACK

PR #3503, opened 2025-08-20
backend: move eth update logic into its own package.

bznein:
Note: this is based on https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3493 and it needs to be reviewed after it.

The only eth-balance specific code that remains in the backend is `updateETHAccounts` (which is passed to the updater as a callback) which I think needs to stay there as it needs access to both the accounts and keystore lock, and the slice of accounts.

- [ ] updating the Changelog
- [ ] writing unit tests


backend/backend.go
```diff
@@ -310,6 +302,7 @@ func NewBackend(arguments *arguments.Arguments, environment Environment) (*Backe
 	backend.socksProxy = backendProxy
 	backend.httpClient = hclient
 	backend.etherScanHTTPClient = hclient
+	backend.ethupdater = eth.NewUpdater(accountUpdate, backend.etherScanHTTPClient, backend.etherScanRateLimiter, backend.updateETHAccounts)
```
benma: I could swear we removed `etherScanHTTPClient` in a recent PR :thinking: I think I commented somewhere that we can remove it as it's the same as `backend.httpClient`, now that the rate limiter is separate from the http client.

backend/coins/eth/updater.go
```diff
@@ -0,0 +1,179 @@
...
+		updateAccounts:          updateETHAccounts,
+		log:                     logging.Get().WithGroup("ethupdater"),
+	}
+}
+
+// Close closes the updater and its channels.
+func (u *Updater) Close() {
+	close(u.quit)
+	close(u.enqueueUpdateForAccount)
+	close(u.updateETHAccountsCh)
```
benma: No need to close these explicitly afaik. They will be garbage collected, and no risk in making the poll loop handle these events (even though it should not happen with the double nesting on the quit channel).

`updateETHAccountsCh` should especially not be closed here as it was not created here either.

backend/coins/eth/updater.go
```diff
@@ -0,0 +1,179 @@
...
+		address, err := account.Address()
+		if err != nil {
+			u.log.WithError(err).Errorf("Could not get address for account %s", account.Config().Config.Code)
+			account.SetOffline(err)
+		}
+		var balance *big.Int
+		switch {
+		case IsERC20(account):
+			var err error
+			balance, err = account.ERC20Balance()
```
benma: Don't think it's worth introducing a new function for it, as it's only used once, it's short and can just be inlined here as before.

backend/coins/eth/updater.go
```diff
@@ -0,0 +1,179 @@
...
+	Balances(ctx context.Context, addresses []common.Address) (map[common.Address]*big.Int, error)
+}
+
+// Updater is a struct that takes care of updating ETH accounts.
+type Updater struct {
+	// quit is used to indicate to running goroutines that they should stop as the backend is being closed
+	quit chan struct{}
+
+	// enqueueUpdateForAccount is used to enqueue an update for a specific ETH account.
+	enqueueUpdateForAccount chan *Account
```
benma: Change the type to `<- chan *Account` to indicate it can only receive, not send/close.

backend/coins/eth/updater.go
```diff
@@ -0,0 +1,179 @@
...
+
+	etherscanClient      *http.Client
+	etherscanRateLimiter *rate.Limiter
+
+	// updateAccounts is a function that updates all ETH accounts.
+	updateAccounts func() error
+}
+
+// NewUpdater creates a new Updater instance.
+func NewUpdater(accountUpdate chan *Account, etherscanClient *http.Client, etherscanRateLimiter *rate.Limiter, updateETHAccounts func() error) *Updater {
```
benma: Please split into multiple lines (one per argument). Best to try keep lines <100 chars.

backend/coins/eth/updater.go
```diff
@@ -0,0 +1,179 @@
...
+				return
+			case account := <-u.enqueueUpdateForAccount:
+				go func() {
+					// A single ETH accounts needs an update.
+					etherScanClient := etherscan.NewEtherScan(account.ETHCoin().ChainIDstr(), u.etherscanClient, u.etherscanRateLimiter)
+					u.UpdateBalances([]*Account{account}, etherScanClient)
+				}()
+			case <-u.updateETHAccountsCh:
+				go updateAll()
+				timer = time.After(PollInterval)
```
benma: move the Pollnterval variable to this file and unexport it - it's not used outside this file anymore.

backend/backend.go
bznein: > I could swear we removed etherScanHTTPClient in a recent PR 🤔

Now that you mention, I think I remember it too :D maybe it was in a staging branch for blockbook? Anyway, will do :)

backend/coins/eth/updater.go
bznein: > Please split into multiple lines (one per argument). Best to try keep lines <100 chars.

Will do, but just for reference I didn't initially as I usually try to follow what I remember from Google's style guide (https://google.github.io/styleguide/go/guide.html#line-length)

bznein:
Ready for another review, thanks @benma !

backend/coins/eth/updater.go
benma: "if it feels too long", very clear :joy: 

In my experience 100 chars is pretty good and fits on split screens in vim/emacs on most displays.

benma (approved):
looks nice, thanks!

utACK

PR #3510, opened 2025-08-26
Frontend amounts should show 2 decimals

thisconnect:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/components/amount/amount.tsx
+++ b/frontends/web/src/components/amount/amount.tsx
@@ -49,6 +49,12 @@ const formatSats = (amount: string): JSX.Element => {
  * into a localized number format, given a custom group separator
  * and decimal separator.
  *
+ * Additional formatting:
+ * - If the amount has exactly one decimal digit (e.g., `"0.1"`),
+ *   it is padded to two decimals (`"0.10"`), since humans generally
+ *   expect at least two decimal places in currencies.
+ * - If the amount has no decimals, it is left untouched (`"10"` → `"10"`).
+ *
  * This function assumes the input amount always follows the Swiss format:
  * - Apostrophe `'` as the group separator
  * - Dot `.` as the decimal separator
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/amount/amount.tsx
```diff
@@ -44,13 +44,45 @@ const formatSats = (amount: string): JSX.Element => {
...
 
+/**
+ * Converts a Swiss-formatted amount string (e.g., `"10'000'000.00"`)
+ * into a localized number format, given a custom group separator
+ * and decimal separator.
+ *
+ * Additional formatting:
+ * - If the amount has exactly one decimal digit (e.g., `"0.1"`),
+ *   it is padded to two decimals (`"0.10"`), since humans generally
+ *   expect at least two decimal places in currencies.
```
benma, line 55: Why apply this to all when only USDT has a display problem? 

Also your commit messages don't contain the real reason for the change, which originated due to `USDT 0.1` looking weird next to `USD 0.10`, but there is no such problem for other coins or tokens. Just "humans expect it" does not really help future `git blame` users understand this change :smile:

thisconnect: If stores would for whatever reason sell products in smilie tokens they would write the price of a product with 2 decimals: `99.90 SmileyCoins` and not `99.9 SmileyCoins`. 

This is the commit message, but of course this is just a subjective claim:

> Humans expect a minimum of two decimals.

thisconnect: Do you think `99.9 Link` is better for users?

But no strong opinion, we can also try to just do it for stablecoins, then I need a good reason  why not all other tokens in the commit message, which I don't know why.

Also we can wontfix it, .. no strong opinion.

thisconnect: If we want it only for stable coins I'd prefer if the backend sends "0.10" as the function to format the amount currently does not know what currency and it would make everything a bit uglier.

thisconnect:
Closing, instead of formatting all non-bitcoin-style amounts, better would be if the backend returns stablecoin amounts already formatted with 2 decimals.

PR #3517, opened 2025-08-29
chore: enable recvcheck linter and fix issues

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -799,7 +799,7 @@ func (backend *Backend) registerKeystore(keystore keystore.Keystore) {
 		if err != nil {
 			return errp.WithMessage(err, "could not retrieve keystore name")
 		}
-		keystoreCfg := accountsConfig.GetOrAddKeystore(fingerprint)
+		keystoreCfg := config.GetOrAddKeystore(accountsConfig, fingerprint)
 		keystoreCfg.Name = keystoreName
 		keystoreCfg.LastConnected = time.Now()
 		return nil
--- a/backend/config/accounts.go
+++ b/backend/config/accounts.go
@@ -168,7 +168,7 @@ func (cfg AccountsConfig) IsAccountWatchonly(account *Account) (bool, error) {
 
 // GetOrAddKeystore looks up the keystore by root fingerprint. If it does not exist, one is added to
 // the list of keystores and the newly created one is returned.
-func (cfg *AccountsConfig) GetOrAddKeystore(rootFingerprint []byte) *Keystore {
+func GetOrAddKeystore(cfg *AccountsConfig, rootFingerprint []byte) *Keystore {
 	ks, err := cfg.LookupKeystore(rootFingerprint)
 	if err == nil {
 		return ks
--- a/backend/config/accounts_test.go
+++ b/backend/config/accounts_test.go
@@ -98,7 +98,7 @@ func TestGetOrAddKeystore(t *testing.T) {
 	fp1 := []byte("aaaa")
 	fp2 := []byte("bbbb")
 
-	ks := cfg.GetOrAddKeystore(fp1)
+	ks := GetOrAddKeystore(cfg, fp1)
 	ks.Name = "ks1"
 
 	require.Len(t, cfg.Keystores, 1)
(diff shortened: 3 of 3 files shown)


backend/signing/configuration.go
```diff
@@ -35,7 +35,7 @@ type KeyInfo struct {
 	ExtendedPublicKey *hdkeychain.ExtendedKey
 }
 
-func (ki KeyInfo) String() string {
+func (ki *KeyInfo) String() string {
```
Beerosagos, line 38: Nut sure, but I think it could make sense to add a check for ki != nil here, to avoid panics. wdyt?

backend/signing/configuration.go
```diff
@@ -46,7 +46,7 @@ type keyInfoEncoding struct {
 }
 
 // MarshalJSON implements json.Marshaler.
-func (ki KeyInfo) MarshalJSON() ([]byte, error) {
+func (ki *KeyInfo) MarshalJSON() ([]byte, error) {
```
Beerosagos, line 49: Same here, and also `ExtendedPublicKey` could be nil as well, so I'd add a check for that too

backend/signing/keypath.go
```diff
@@ -193,40 +194,43 @@ func NewAbsoluteKeypathFromUint32(elements ...uint32) AbsoluteKeypath {
...
-func (absoluteKeypath AbsoluteKeypath) ToUInt32() []uint32 {
-	return keypath(absoluteKeypath).toUInt32()
+func (absoluteKeypath *AbsoluteKeypath) ToUInt32() []uint32 {
+	return keypath(*absoluteKeypath).toUInt32()
 }
 
 // MarshalJSON implements json.Marshaler.
-func (absoluteKeypath AbsoluteKeypath) MarshalJSON() ([]byte, error) {
+func (absoluteKeypath *AbsoluteKeypath) MarshalJSON() ([]byte, error) {
 	return json.Marshal(absoluteKeypath.Encode())
```
Beerosagos, line 234: I think it makes sense to check for nil here too, to avoid unexpected panics on marshalling

backend/signing/keypath.go
```diff
@@ -236,7 +240,7 @@ func (absoluteKeypath *AbsoluteKeypath) UnmarshalJSON(bytes []byte) error {
 	if err := json.Unmarshal(bytes, &input); err != nil {
 		return errp.Wrap(err, "Could not unmarshal an absolute keypath.")
 	}
-	var err error
-	*absoluteKeypath, err = NewAbsoluteKeypath(input)
+	absKeypath, err := NewAbsoluteKeypath(input)
+	*absoluteKeypath = *absKeypath
```
Beerosagos, line 244: Better to check for err before doing this, I think. Otherwise, it could lead to panics.

backend/signing/configuration.go
```diff
@@ -142,11 +142,11 @@ func (configuration *Configuration) ScriptType() ScriptType {
 }
 
 // AbsoluteKeypath returns the configuration's keypath.
-func (configuration *Configuration) AbsoluteKeypath() AbsoluteKeypath {
+func (configuration *Configuration) AbsoluteKeypath() *AbsoluteKeypath {
```
Beerosagos, line 145: I think it would be better to return a copy of the object instead of a pointer, to avoid unwanted changes to the configuration

backend/signing/configuration.go
```diff
@@ -65,7 +65,7 @@ func (ki *KeyInfo) UnmarshalJSON(bytes []byte) error {
 		return errp.WithStack(err)
 	}
 	ki.RootFingerprint = rootFingerprint
-	ki.AbsoluteKeypath = encoding.Keypath
+	ki.AbsoluteKeypath = *encoding.Keypath
```
Beerosagos, line 68: Need to check for nil to avoid panics

Beerosagos (changes requested):

backend/signing/configuration.go
```diff
@@ -35,21 +35,21 @@ type KeyInfo struct {
...
+func (ki *KeyInfo) String() string {
 	return fmt.Sprintf("keypath=%s", ki.AbsoluteKeypath.Encode())
 }
 
 type keyInfoEncoding struct {
-	RootFingerprint string          `json:"rootFingerprint"`
-	Keypath         AbsoluteKeypath `json:"keypath"`
-	Xpub            string          `json:"xpub"`
+	RootFingerprint string           `json:"rootFingerprint"`
+	Keypath         *AbsoluteKeypath `json:"keypath"`
```
benma, line 44: this changes the json encoding, allowing or the json to not contain the keypath entry. 

Also AbsoluteKeypath is a type alias for a slice, and taking a pointer to a slice is almost never necessary or useful. I'd undo this for all cases in this PR that take a pointer to AbsoluteKeypath.

backend/backend.go
```diff
@@ -717,7 +717,7 @@ func (backend *Backend) Start() <-chan interface{} {
 	backend.usbManager = usb.NewManager(
 		backend.arguments.MainDirectoryPath(),
 		backend.arguments.BitBox02DirectoryPath(),
-		backend.socksProxy,
+		*backend.socksProxy,
```
benma, line 720: It doesn't make much sense to make backend.socksProxy a pointer but then dereference it, or leave it as a value in the Manager struct :see_no_evil:

benma (commented):
Personally not a big fan of this linter, don't think it adds much value, but has high potential for false positives (JSON encodings?) and needless churn :man_shrugging:

PR #3518, opened 2025-08-29
backend: batch calls to get the block number by chainID.

bznein:
We don't need to call the API endpoint for each ETH account, as every account on the same chain will have the same results.

Instead, we batch calls and compute the block number only once for each chainID.

- [ ] updating the Changelog
- [ ] writing unit tests


backend/coins/eth/updater.go
```diff
@@ -113,8 +115,19 @@ func (u *Updater) PollBalances() {
...
 }
 
-// UpdateBalances updates the balances of the accounts in the provided slice.
-func (u *Updater) UpdateBalances(accounts []*Account, etherScanClient BalanceFetcher) {
+// UpdateBalancesAndBlockNumber updates the balances of the accounts in the provided slice.
+func (u *Updater) UpdateBalancesAndBlockNumber(accounts []*Account, etherScanClient BalanceFetcher) {
+	if len(accounts) == 0 {
+		return
+	}
+	chainId := accounts[0].ETHCoin().ChainIDstr()
```
benma: Could also just use `ChainID()`

backend/coins/eth/updater.go
```diff
@@ -23,6 +24,7 @@ var pollInterval = 5 * time.Minute
 //go:generate moq -pkg mocks -out mocks/balancefetcher.go . Interface
 type BalanceFetcher interface {
```
benma: Maybe this should be renamed now, as it's not about balance fetching only now.

backend/coins/eth/updater.go
```diff
@@ -23,6 +24,7 @@ var pollInterval = 5 * time.Minute
...
 	Balances(ctx context.Context, addresses []common.Address) (map[common.Address]*big.Int, error)
+	rpcclient.Interface
```
benma: Maybe it makes sense to add the block number call only to not blow up this interface? Not sure which is better though.

benma (approved):
utACK, but see comments

backend/coins/eth/updater.go
bznein: I thought about that, but I was worried that the two things might get out of sync. Maybe I can simply add the block number call and a comment to keep them in sync? Not foolproof but a good compromise maybe

PR #3520, opened 2025-09-01
backend: remove QT workaround for dark mode.

bznein:
With Qt 5.15.2, dark mode was not working on Linux/Windows/MacOS We have since then updated to Qt 6 and the workaround is not needed anymore.

I have tested only on MacOS and dark mode works. Testing on other environments is welcome :) 

- [ ] updating the Changelog
- [ ] writing unit tests


benma (approved):
very nice! utACK

PR #3521, opened 2025-09-01
backend: re-enable the sandbox for Qt.

bznein:
The sandbox was disabled for Linux on Qt 5.15.2 as it would cause the app to show a blank screen. This is not needed anymore as we upgraded to Qt 6.

Needs to be tested by someone with Linux as I currently only have a Macbook to work on.

- [ ] updating the Changelog
- [ ] writing unit tests


benma (approved):
tACK on linux

PR #3523, opened 2025-09-01
backend: fix data race.

bznein:
This fixes a test failure (data race) that might have escaped before: I'm not able to repro it if running `make ci` inside the `dockerdev`, nor it happens in github CI. However, it happens 100% of the times if ran locally on my machine:

```
WARNING: DATA RACE
Write at 0x00c0003cf2e0 by goroutine 50:
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts.(*BaseAccount).SetOffline()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/accounts/baseaccount.go:143 +0x1ac7
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth.(*Updater).UpdateBalances()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/coins/eth/updater.go:182 +0x1a9d
  github.com/BitBoxSwiss/bitbox-wallet-app/backend.(*Backend).updateETHAccounts()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/backend.go:581 +0x59c
  github.com/BitBoxSwiss/bitbox-wallet-app/backend.(*Backend).updateETHAccounts-fm()
      <autogenerated>:1 +0x33
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth.(*Updater).PollBalances.func1()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/coins/eth/updater.go:85 +0x52

Previous read at 0x00c0003cf2e0 by goroutine 49:
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts.(*BaseAccount).Offline()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/accounts/baseaccount.go:136 +0x1718
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth.(*Updater).UpdateBalances()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/coins/eth/updater.go:175 +0x16ee
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth.(*Updater).PollBalances.func2()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/coins/eth/updater.go:102 +0x29a

Goroutine 50 (running) created at:
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth.(*Updater).PollBalances()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/coins/eth/updater.go:108 +0x364
  github.com/BitBoxSwiss/bitbox-wallet-app/backend.(*Backend).Start.gowrap2()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/backend.go:687 +0x33

Goroutine 49 (running) created at:
  github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth.(*Updater).PollBalances()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/coins/eth/updater.go:99 +0x305
  github.com/BitBoxSwiss/bitbox-wallet-app/backend.(*Backend).Start.gowrap2()
      /Users/bznein/Documents/bznein/bitbox-wallet-app/backend/backend.go:687 +0x33
==================
FAIL	github.com/BitBoxSwiss/bitbox-wallet-app/backend/bridgecommon	4.052s
```

SetOffline can be called by different goroutines.
Prior to the introduction of the eth updater, it was called only from within account.Update, where a lock is already held. However, the updater needs to call it when updating balances/block numbers, so we need to guard it.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/accounts/baseaccount.go
+++ b/backend/accounts/baseaccount.go
@@ -70,7 +70,7 @@ type BaseAccount struct {
 	// synced indicates whether the account has loaded and finished the initial sync of the
 	// addresses.
 	synced  atomic.Bool
-	offline error
+	offline atomic.Pointer[error]
 
 	// notes handles transaction notes.
 	notes *notes.Notes
@@ -128,13 +128,17 @@ func (account *BaseAccount) ResetSynced() {
 
 // Offline implements Interface.
 func (account *BaseAccount) Offline() error {
-	return account.offline
+	errorPtr := account.offline.Load()
+	if errorPtr == nil {
+		return nil
+	}
+	return *errorPtr
 }
 
 // SetOffline sets the account offline status and emits the EventStatusChanged() if the status
 // changed.
 func (account *BaseAccount) SetOffline(offline error) {
-	account.offline = offline
+	account.offline.Store(&offline)
 	account.Notify(observable.Event{
 		Subject: string(types.EventStatusChanged),
 		Action:  action.Reload,


bznein:
Now that I think about it, the issue is that the updater might receive a request to update a single account, while is performing update of all accounts.

I'm not sure the lock on the setoffline is the right solution, as there might be more nuanced cases in which a data race might happen (though right now the updater calls `account.Update()` for everything except the `SetOffline` method, so we are covered by the `updateLock`;

 I think we could either:

- Drop the request to update an account if the account is already being updated (might cause stale data?)
- Have a lock at the updater level, though I'm not sure a single lock might be the right move. maybe we need a lock for each account, so that if the UpdateBalances function is not updating the account for which we receive the request, we can update that one immediately.


WDYT @benma ?

bznein:
@benma as discussed offline, switched from a lock to an atomic.Value (with a small workaround, see comment in the code)

backend/accounts/baseaccount.go
```diff
@@ -70,14 +70,23 @@ type BaseAccount struct {
 	// synced indicates whether the account has loaded and finished the initial sync of the
 	// addresses.
 	synced  atomic.Bool
-	offline error
+	offline atomic.Value
```
benma: Maybe better to use `atomic.Pointer[error]`? Generic and type-safe, no need for type-casting.

bznein: Good idea, done

benma (approved):

PR #3524, opened 2025-09-01
backend/bitbox02: add integration test for payment requests

benma:
So refactorings related to BitBox02 keystore transactions are sure not to break payment requests.

This requires a simulator that is not yet released. Build it from latest bitbox02-firmware master and rename it to `simulator-v9.24.0` (version in filename is required so the tests can infer the version)

Then run using

```
SIMULATOR=/path/to/bitbox02-firmware/build-build-noasan/bin/simulator-v9.24.0 go test -mod=vendor -tags=bitbox02_simulator ./... -count=1 -v -run 'TestSimulatorSignBTCPaymentRequest'

### DIFF
--- a/backend/devices/bitbox02/keystore_simulator_test.go
+++ b/backend/devices/bitbox02/keystore_simulator_test.go
@@ -21,9 +21,11 @@ import (
 	"bufio"
 	"bytes"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"hash"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: backend/devices/bitbox02/keystore_simulator_test.go [hunk 1]
```diff
@@ -696,3 +702,120 @@ func TestSimulatorVerifyAddressETH(t *testing.T) {
...
+Test Merchant
+CONFIRM SCREEN END
+CONFIRM SCREEN START
+TITLE: Memo 1/2
+BODY: TextMemo line1
+CONFIRM SCREEN END
+CONFIRM SCREEN START
+TITLE: Memo 2/2
+BODY: TextMemo line2
+CONFIRM SCREEN END`) &&
```
bznein: I know this is all `gofmt` doing but maybe we can extract it (and the one above) in a var so it becomes a bit easier to read the code?

bznein (commented):
utACK (I can't run the test on MacOS but I trust that you ran it :D)

One minor comment, feel free to ignore or if you address it feel free to merge right after :)

bznein (approved):
Why do I keep forgetting to actually approve :D

PR #3525, opened 2025-09-01
chore: enable trimpath flag for go build.

bznein:
As per official docs, `-trimpath `removes all file system paths from the resulting executable.

This way, stack traces in the logs won't show hardcoded paths that can confuse users.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/mobileserver/Makefile
+++ b/backend/mobileserver/Makefile
@@ -3,8 +3,8 @@ include ../../android-env.mk.inc
 # Set -glflags to fix the vendor issue with gomobile, see: https://github.com/golang/go/issues/67927#issuecomment-2241523694
 build-android:
 	# androidapi version should match minSdkVersion in frontends/android/BitBoxApp/app/build.gradle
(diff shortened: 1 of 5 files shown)


bznein:
According to https://pkg.go.dev/golang.org/x/mobile/cmd/gomobile#:~:text=on%20the%20PATH.-,The%20build%20flags%20%2Da%2C%20%2Di%2C%20%2Dn%2C%20%2Dx%2C%20%2Dgcflags%2C%20%2Dldflags%2C%20%2Dtags%2C%20%2Dtrimpath%2C%20and%20%2Dwork%20are%20shared%20with%20the%20build%20command.%20For%20documentation%2C%20see%20%27go%20help%20build%27.,-Print%20version%20%C2%B6 we can use the flag also with gomobile. I'll push the change and test it first thing tomorrow :)

bznein:
@benma I checked by downlaoding the apk from latest master and from this PR, and scanning the files inside from absolute path, and I can confirm that absolute path are not there anymore also for android (I wasn't able to perform the same test for iOS, but since the effect is obtained by changing the build targets on mobileserver for both, I am confident it would be the same)

benma (approved):
nice, utACK

PR #3526, opened 2025-09-01
backend/bitbox02: simulator test that change is not confirmed

benma:
Easy to miss, if the change output info is missing and sent to the BitBox as an external address by accident, it will be explicitly confirmed. This helps catch regressions when refactoring signing logic.


bznein (commented):
utACK

bznein (approved):

PR #3527, opened 2025-09-01
frontend: remove currency selector and fix suggested account name for BTC-only.

bznein:
When there is only one coin supported, we want the back button on the choose name step to close the dialog. And we want the account name to be suggested.

- [x] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/account/add/add.tsx
+++ b/frontends/web/src/routes/account/add/add.tsx
@@ -68,7 +68,7 @@ export const AddAccount = ({ accounts }: TAddAccountGuide) => {
       setStep(onlyOneCoinIsSupported ? 'choose-name' : 'select-coin');
       setSupportedCoins(coins);
       if (onlyOneCoinIsSupported) {
-        setAccountCode(coins[0].suggestedAccountName);
+        setAccountName(coins[0].suggestedAccountName);
       }
       inputRef.current?.focus();
     } catch (err) {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/account/add/add.tsx
```diff
@@ -155,7 +159,10 @@ export const AddAccount = ({ accounts }: TAddAccountGuide) => {
           ref={inputRef}
           id="accountName"
           onInput={e => setAccountName(e.target.value)}
-          value={accountName} />
+          value={
+            accountName ||
+              (onlyOneSupportedCoin() ? supportedCoins[0].suggestedAccountName : '')
```
thisconnect: Could you move this into back function and use setAccountName ?

bznein: @thisconnect as discussed privately, I don't think `back` is the right place, but I moved the logic to `startProcess` and I think the result is much more cleaner now, wdyt?

frontends/web/src/routes/account/add/add.tsx
```diff
@@ -69,6 +69,7 @@ export const AddAccount = ({ accounts }: TAddAccountGuide) => {
       setSupportedCoins(coins);
       if (onlyOneCoinIsSupported) {
         setAccountCode(coins[0].suggestedAccountName);
```
thisconnect: It looks to me that you found a bug and  this line was intended to set `setAccountName` and not account code.

bznein: @thisconnect  I agree, I removed this line and replaced it with just `setAccountName`. Ready for review again :)

thisconnect (approved):
tested 082d721d9 LGTM thank you

PR #3529, opened 2025-09-02
ios/android: fix screen lock bugs

Beerosagos:
On some Android versions (e.g. Android 9) the authentication flow was causing an endless loop of auth requests. This was fixed by moving the setup of `BiometricAuthHelper` from `onStart` to `onCreate`.

In both Android and iOS the auth flow freezed if there was no authentication method setup on the device. This was fixed by refactoring the authentication flow code and adding a new possible authentication response `authres-missing`, that allows to handle this case.

### DIFF
--- a/backend/bridgecommon/bridgecommon.go
+++ b/backend/bridgecommon/bridgecommon.go
@@ -143,25 +143,16 @@ func TriggerAuth() {
 	globalBackend.TriggerAuth()
 }
 
-// CancelAuth triggers an authentication canceled notification.
-func CancelAuth() {
-	mu.Lock()
-	defer mu.Unlock()
-	if globalBackend == nil {
-		return
-	}
-	globalBackend.CancelAuth()
-}
-
 // AuthResult triggers an authentication result notification
 // on the base of the input value.
-func AuthResult(ok bool) {
+func AuthResult(result backend.AuthResultType) {
 	mu.Lock()
 	defer mu.Unlock()
 	if globalBackend == nil {
 		return
 	}
-	globalBackend.AuthResult(ok)
+
+	globalBackend.AuthResult(result)
 }
 
 // UsingMobileDataChanged should be called when the network connnection changed.
--- a/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/BiometricAuthHelper.java
+++ b/frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/BiometricAuthHelper.java
@@ -17,9 +17,23 @@ public class BiometricAuthHelper {
         void onSuccess();
         void onFailure();
         void onCancel();
+        void noAuthConfigured();
     }
 
     public static void showAuthenticationPrompt(FragmentActivity activity, AuthCallback callback) {
+
+        BiometricManager biometricManager = BiometricManager.from(activity);
+        int canAuthenticate = biometricManager.canAuthenticate(
+                BiometricManager.Authenticators.DEVICE_CREDENTIAL |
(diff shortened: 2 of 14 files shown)


backend/bridgecommon/bridgecommon.go
```diff
@@ -143,25 +143,16 @@ func TriggerAuth() {
...
-	if globalBackend == nil {
-		return
-	}
-	globalBackend.CancelAuth()
-}
-
 // AuthResult triggers an authentication result notification
 // on the base of the input value.
-func AuthResult(ok bool) {
+func AuthResult(result string) {
```
benma: I think it's better if this is of type AuthResult, with the conversion happening in mobileserver.go's `AuthResult` wrapper, because the string constants are defined there and not here.

frontends/web/src/components/auth/authrequired.tsx
```diff
@@ -50,23 +63,25 @@ export const AuthRequired = () => {
...
+            // it means that the user aborted the authentication flow needed to enable
+            // the screen lock.
+            dismissAuth();
+          }
+          break;
+        case 'authres-ok':
+          dismissAuth();
+          break;
+        case 'authres-missing':
+          setMissingAuth(true);
```
benma, line 83: missing `break;` statement

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/BiometricAuthHelper.java
```diff
@@ -17,9 +17,23 @@ public interface AuthCallback {
...
+        void noAuthConfigured();
     }
 
     public static void showAuthenticationPrompt(FragmentActivity activity, AuthCallback callback) {
+
+        BiometricManager biometricManager = BiometricManager.from(activity);
+        int canAuthenticate = biometricManager.canAuthenticate(
+                BiometricManager.Authenticators.DEVICE_CREDENTIAL |
+                        BiometricManager.Authenticators.BIOMETRIC_WEAK
+        );
```
benma, line 29: https://developer.android.com/reference/androidx/biometric/BiometricManager#canAuthenticate(int)

> Developers that wish to check for the presence of a PIN, pattern, or password on these versions should instead use [isDeviceSecure](https://developer.android.com/reference/android/app/KeyguardManager.html#isDeviceSecure()).

So maybe isDeviceSecure is the right function to call?

Beerosagos: Our current check is `BiometricManager.Authenticators.DEVICE_CREDENTIAL | BiometricManager.Authenticators.BIOMETRIC_WEAK`, which is correctly supported in our target APIs. No need to change that, imo.

frontends/web/src/locales/en/app.json
```diff
@@ -99,6 +99,8 @@
   },
   "auth": {
     "authButton": "Authenticate",
+    "dismissButton": "Cancel",
+    "missing": "Please, setup an authentication method on your device to continue",
```
benma: no comma after `Please` please :smile: 

Maybe better: "Please setup a screen lock in your device settings".

benma:
(Minor?) bug: if you setup screenlock on android, then disable the screen lock in the device settings, you will be stuck in an auth loop when you launch the app.

Beerosagos:
> (Minor?) bug: if you setup screenlock on android, then disable the screen lock in the device settings, you will be stuck in an auth loop when you launch the app.

Yes, but the auth screen message will warn you that you need to setup a screen lock to proceed. That sounds reasonable to me

benma (approved):
tACK

PR #3530, opened 2025-09-02
backend/devices: remove feature to get randomness from bb01

benma:
Reduce maintenance burden and locize costs.


frontends/web/src/locales/en/app.json
```diff
@@ -1922,4 +1918,4 @@
     "message": "Please connect your BitBox and tap the side to continue.",
     "title": "Welcome"
   }
-}
\ No newline at end of file
+}
```
thisconnect: weblint fails. maybe due to this newline?

benma: Yup, but locize format is broken, had to manually fix it :see_no_evil: 

https://github.com/locize/locize-cli/issues/107#issuecomment-3247986177

PR #3531, opened 2025-09-03
backend/frontend: add option to configure bitcoin gap limit

strmci:
Add a new expert setting that allows changing the gap limit used for transaction discovery on bitcoin receive and change addresses. Until now, this was only possible via command-line.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -29,6 +29,7 @@ import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/addresses"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/blockchain"
+	btctypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/types"
 	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/config"
@@ -1016,6 +1017,24 @@ outerLoop:
 	return ks, err
 }
 
+// gapLimits returns the gap limits to use, with arguments having priority over config settings.
+func (backend *Backend) gapLimits() *btctypes.GapLimits {
+	gapLimits := backend.arguments.GapLimits()
+
+	if gapLimits == nil {
+		configReceive := uint16(backend.config.AppConfig().Backend.GapLimitReceive)
+		configChange := uint16(backend.config.AppConfig().Backend.GapLimitChange)
+		if configReceive > 0 && configChange > 0 {
+			gapLimits = &btctypes.GapLimits{
+				Receive: configReceive,
+				Change:  configChange,
+			}
+		}
+	}
+
+	return gapLimits
+}
+
 // The accountsAndKeystoreLock must be held when calling this function.
 func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *config.Account) {
 	if backend.accounts.lookup(persistedConfig.Code) != nil {
@@ -1075,7 +1094,7 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
 		account = backend.makeBtcAccount(
 			accountConfig,
 			specificCoin,
-			backend.arguments.GapLimits(),
+			backend.gapLimits(),
 			getAddressCallback,
 			backend.log,
 		)
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -824,6 +824,14 @@
   "footer": {
     "appVersion": "App version:"
   },
+  "gapLimit": {
+    "change": "Gap limit for change addresses",
+    "description": "Configure custom gap limits for bitcoin receive and change addresses.",
+    "maxValue": "The gap limit must not exceed {{max}}.",
+    "minValue": "The gap limit must be at least {{min}}.",
+    "receive": "Gap limit for receive addresses",
+    "title": "Custom gap limit settings"
+  },
   "generic": {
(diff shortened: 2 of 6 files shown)


bznein (commented):
I tested without looking at the code on webdev and I have a couple of comments:

- The input field for gap limits has no maximum value; if I input a number that goes below the maximum int value, I get 

Image

and in the logs: 

```
ERRO[0144] endpoint failed                               error="json: cannot unmarshal number 2.011212121211232e+35 into Go struct field Backend.backend.gapLimitReceive of type int > github.com/BitBoxSwiss/bitbox-wallet-app/backend/handlers.(*Handlers).postAppConfig > /home/bznein/Shift/BitBoxSwiss/bitbox-wallet-app/backend/handlers/handlers.go:465 > github.com/BitBoxSwiss/bitbox-wallet-app/backend/handlers.NewHandlers.NewHandlers.NewHandlers.func2.func20.(*Handlers).apiMiddleware.func146 > /home/bznein/Shift/BitBoxSwiss/bitbox-wallet-app/backend/handlers/handlers.go:1077 > net/http.HandlerFunc.ServeHTTP > /usr/local/go/src/net/http/server.go:2294 > github.com/BitBoxSwiss/bitbox-wallet-app/backend/handlers.NewHandlers.NewHandlers.NewHandlers.func2.func20.ensureAPITokenValid.func147 > /home/bznein/Shift/BitBoxSwiss/bitbox-wallet-app/backend/handlers/handlers.go:1052 > net/http.HandlerFunc.ServeHTTP > /usr/local/go/src/net/http/server.go:2294 > github.com/gorilla/mux.(*Router).ServeHTTP > /home/bznein/Shift/BitBoxSwiss/bitbox-wallet-app/vendor/github.com/gorilla/mux/mux.go:212 > net/http.serverHandler.ServeHTTP > /usr/local/go/src/net/http/server.go:3301 > net/http.(*conn).serve > /usr/local/go/src/net/http/server.go:2102 > runtime.goexit > /usr/local/go/src/runtime/asm_amd64.s:1700" group=handlers
```

Not critical (unless someone tries to set an insame number).

- Also I see in the logs that `tbtc configuration="bitcoinSimple;scriptType=p2wpkh;keypath=m/84'/1'/5'" gap-limit=2000 group=addresses name="Bitcoin Testnet 6" net=testnet3` so it seems like 2000 as a maximum is enforced by the backend, so we should probably put that as max value?

- A UX thing I noticed: if I select the number and move the cursor outside the dialog (which I think is fairly common on pc) the dialog disappears, see 

https://github.com/user-attachments/assets/bb6502d8-3efd-48eb-9f62-8d93c5ac3fd9

Not sure if this happens only on webdev, I'll check later on other platforms

-  the current input field doesn't allow to empty the field, so if I want to go from 6 to, let's say, 10, I can't delete the 6 and type 10. I either need to increase with the arrows on the right, or write 10 after the 6, and then delete the 6

frontends/web/src/routes/settings/components/advanced-settings/custom-gap-limit-setting.tsx
```diff
@@ -0,0 +1,119 @@
...
+        onClick={() => setShowDialog(true)}
+      />
+      <Dialog
+        open={showDialog}
+        onClose={() => setShowDialog(false)}
+        title={t('gapLimit.title')}
+        small>
+        <div className="columnsContainer half">
+          <div className="columns half">
+            <div className="column">
```
thisconnect: I wanted to ask to use `<Grid>` and `<Column>` component here, but I don't think these are even needed. There is no columns here. Did a quick test and it renders fine without those 3 div's, please remove

 
Screenshot 2025-09-10 at 15 22 55
Screenshot 2025-09-10 at 15 22 21

frontends/web/src/routes/settings/components/advanced-settings/custom-gap-limit-setting.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }, [backendConfig]);
+
+  const handleSave = async () => {
+    const config = await setConfig({
+      backend: {
+        ...backendConfig,
+        gapLimitReceive,
+        gapLimitChange,
+      },
+    }) as TConfig;
```
thisconnect: I'd always prefer without typecasting if possible so that TypeScript can complain. Here it seems not needed and TS seems happy without.

```suggestion
    });
```

frontends/web/src/locales/en/app.json
```diff
@@ -824,6 +824,15 @@
   "footer": {
     "appVersion": "App version:"
   },
+  "gapLimit": {
+    "change": "Gap limit for change addresses",
+    "description": "Configure custom gap limits for bitcoin receive and change addresses.",
+    "maxValue": "The gap limit must not exceed {{max}}.",
+    "minValue": "The gap limit must be at least {{min}}.",
+    "receive": "Gap limit for receive addresses",
+    "resetToDefault": "Reset to default",
```
thisconnect: Small nit: as it is annoying to change keys later, I checked this one and noticed we already have same text in `settings.electrum.reset`. Therefore please consider moving this key to `generic.reset` or `generic.resetToDefault`. This allows us to use the same translation in the future. We can change the electrum key later.

strmci: moved to `generic`

backend/accounts.go
```diff
@@ -1072,10 +1073,23 @@ func (backend *Backend) createAndAddAccount(coin coinpkg.Coin, persistedConfig *
...
+		if gapLimits == nil {
+			configReceive := uint16(backend.config.AppConfig().Backend.GapLimitReceive)
+			configChange := uint16(backend.config.AppConfig().Backend.GapLimitChange)
+			if configReceive > 0 && configChange > 0 {
+				gapLimits = &btctypes.GapLimits{
+					Receive: configReceive,
+					Change:  configChange,
+				}
+			}
+		}
```
bznein: This function is already big enough so maybe it would be good to not add more code. I think we could create a `func (b *Backend) gapLimits()` that does all the same logic, WDYT? Then we can call it down on line 1078 similar to what we were doing before

strmci: yes, I agree, updated with a new function

PR #3533, opened 2025-09-03
frontend: refactor dialog to use state instead of refs for class toggling

thisconnect:
Previously the dialog relied on `useRef` to directly manipulate DOM classes (`classList.add/remove`). This made the component harder to reason about and coupled logic tightly to the DOM.

Now the dialog uses React state (`isActive`, `isOpen`, `isClosing`) to conditionally render CSS classes. This improves readability, keeps the component declarative, and avoids unnecessary ref usage except where focus management is required.

In preparation for https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3534

### DIFF
--- a/frontends/web/src/components/dialog/dialog.tsx
+++ b/frontends/web/src/components/dialog/dialog.tsx
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import React, { useCallback, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


thisconnect:
I tested `<Dialog>` in Tor proxy and language switch on Android and Qt/macOS seem to work as before with opening and closing animation.

PR #3534, opened 2025-09-04
frontend: enable nouncheckedindexedaccess typescript rule

thisconnect:


thisconnect:
Changed CSS modules in template literals from type casting `as string` (as it could theoretically become string "undefined") to empty string fallback `|| ''`. Fixed into commit: "fix nouncheckedindexedaccess in template literals"

PR #3538, opened 2025-09-08
Psbt

benma:


backend/coins/btc/maketx/txsize_internal_test.go
```diff
@@ -57,6 +59,51 @@ func makeSig() types.Signature {
...
+		// We assume SIGHASH_DEFAULT, which defaults to SIGHASH_ALL without needing to explicitly
+		// append it to the signature. See:
+		// https://github.com/bitcoin/bips/blob/97e02b2223b21753acefa813a4e59dbb6e849e77/bip-0341.mediawiki#taproot-key-path-spending-signature-validation
+		txWitness := wire.TxWitness{
+			signature.SerializeCompact(),
+		}
+		return []byte{}, txWitness
+	default:
+		panic("Unrecognized address type.")
+	}
```
bznein, line 104: I know we have different opinions on this so feel free to ignore this comment :D 

I personally like when the switch sets up the return values and then the return itself is outside the switch.

So something like

```golang
var signatureScript []byte
var txWitness wire.TxWitness
swtich {
// Everything the same without returns
}

return signatureScript, txWitness
```

(Or, if you want to save lines of code, you can even use named returns in the function signature :D)

.golangci.yml
```diff
@@ -120,6 +120,7 @@ linters:
     - predeclared
     - tenv
     - recvcheck
+    - godot
```
bznein, line 123: [nit] not super clear why this is included? Maybe worth a mention in the commit message (or a separate commit)?

backend/coins/btc/sign.go
```diff
@@ -44,66 +48,161 @@ type ProposedTransaction struct {
...
+	// Add key infos to outputs.
+	for index, txOut := range txProposal.Psbt.UnsignedTx.TxOut {
+		// outputAddress is the recipient address if it belongs to the same keystore. It is nil if
+		// the address is external.
+		outputAddress, sameAccount, err := p.GetKeystoreAddress(p.SendingAccount, blockchain.NewScriptHashHex(txOut.PkScript))
+		if err != nil {
+			return errp.Newf("failed to get address: %v", err)
+		}
+
+		_ = sameAccount
```
bznein: Is this leftover code?

backend/coins/btc/sign.go
```diff
@@ -44,66 +48,161 @@ type ProposedTransaction struct {
...
+	// Sanity check: see if the created transaction is valid.
+	if err := txValidityCheck(signedTx, p.TXProposal.PreviousOutputs,
+		p.TXProposal.SigHashes()); err != nil {
+		return nil, err
+	}
+	return signedTx, nil
+}
+
 // signTransaction signs all inputs. It assumes all outputs spent belong to this
 // wallet. previousOutputs must contain all outputs which are spent by the transaction.
```
bznein, line 168: maybe add what the new returned value is?

backend/coins/btc/maketx/txsize_internal_test.go
benma: maybe try to change my mind by explaining why this is better :joy:  in my view, returning in the cases is nicer, as you need to keep less context in your head. in this case, the context is where the variables are defined, and if they will be changed later again.

by returning early these things fall away.

.golangci.yml
benma: it was a false positive, and it's not worth wasting time over not finishing a docstring with a dot imho. added to the commit msg

backend/coins/btc/sign.go
benma: Done

backend/coins/btc/maketx/txsize_internal_test.go
benma: Btw, in this case the function was just moved from address.go, it's not a new function.

bznein: > maybe try to change my mind by explaining why this is better 😂 in my view, returning in the cases is nicer, as you need to keep less context in your head. in this case, the context is where the variables are defined, and if they will be changed later again.

I guess it really just boils down to personal preference :D Imho it makes it clearer as there is only one return and one place where the variables are defined. 

Anyway, by reading online, it seems like your version has a stronger supporter base, so I'll concede :D

PR #3545, opened 2025-09-10
backend: remove account names from log to preserve privacy

strmci:
Account names don’t provide useful information and could reveal sensitive data.


benma (approved):
utACK

PR #3549, opened 2025-09-10
frontend: use dropdown component in feetargets

thisconnect:
The native select dropdown sometimes renders with buggy visual effect on windows.

Fixed by changing to dropdown component which is a wrapper around react-select lib and doesn't ues native select element.

- [x] updating the Changelog


thisconnect:
rebased with small CHANGELOG conflict

PR #3550, opened 2025-09-11
Frontend new payment messages

thisconnect:


frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -67,6 +67,8 @@ export const Pocket = ({
 
   const ref = createRef<HTMLDivElement>();
   const iframeRef = useRef<HTMLIFrameElement | null>(null);
+  const sessionId = useRef(crypto.randomUUID());
```
TKone7: Instead of setting your own session id, my idea was for you to reply with the same id that was sent in the request analogue to how it's done in [JSON RPC](https://www.jsonrpc.org/specification#request_object).

So in a `Payment` message you would use the ID of the `PaymentRequest` sent by us. Similarly in `Address` and `RequestAddress`. A `Cancel` would refer to the ID of whatever request was preceding.

thisconnect: changed in last commit PTAL

frontends/web/package.json
```diff
@@ -38,7 +38,7 @@
     "react-i18next": "14.1.2",
     "react-router-dom": "6.4.3",
     "react-select": "5.8.0",
-    "request-address": "0.0.10"
+    "request-address": "pocketbitcoin/request-address#payment"
```
thisconnect: @TKone7 please let me know when 0.0.11 is published on npm https://www.npmjs.com/package/request-address

TKone7: It's published.

frontends/web/src/routes/exchange/pocket.tsx
TKone7: Cool, will test it now with our implementation.

TKone7:
Do you have any deadlines for the next release of the wallet app that we can aim for with this change?

thisconnect:
@TKone7 ack to merge?

frontends/web/src/routes/exchange/pocket.tsx
```diff
@@ -221,10 +256,14 @@ export const Pocket = ({
       alertUser(t('unknownError', { errorMessage: 'Missing payment request data' }));
       return;
     }
+    if (message.correlationId) {
+      setCorrelationId(message.correlationId);
```
TKone7: While testing I noticed I do not get the correct correlationId. Reason is that when you `setCorrelationId` here this is not reflected in `correlationId` until the next rerender. So further down in `sendPaymentTxId` or `sendCanceledMessage` the `correlationId` is either `undefined` (during the first time) or it's set to the previous value (on subsequent calls).

I don't see why you would need to store the correlation id in a state. I would simply read it here (line 259) into a local variable and then pass it on as an argument to `sendPaymentTxId` or `sendCanceledMessage`.

TKone7: I can propose a commit if you want me to.

thisconnect: yes please 🙏

TKone7: Does this make sense 19e9fc6820ad4a75f12a063663b2741abd0545ee?

thisconnect: yes 100% thank you so much. fixed up into this commit

thisconnect: @TKone7 I guess with the last change this PR is ok as is from your side? PTAL

TKone7 (commented):
Tested locally and can confirm that this works for us 👍🏽

frontends/web/src/routes/exchange/pocket.tsx
TKone7: Yes, perfect. 🟢

PR #3552, opened 2025-09-11
frontend: show bitrefill address in a blocking dialog

thisconnect:
So the user can confirm the payment on the device.

### DIFF
--- a/frontends/web/src/routes/market/bitrefill.tsx
+++ b/frontends/web/src/routes/market/bitrefill.tsx
@@ -18,11 +18,10 @@ import { useState, useEffect, useRef, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Header } from '@/components/layout';
 import { Spinner } from '@/components/spinner/Spinner';
-import { AccountCode, IAccount, proposeTx, sendTx, TTxInput } from '@/api/account';
+import { AccountCode, IAccount, proposeTx, sendTx, TTxInput, TTxProposalResult } from '@/api/account';
 import { findAccount, isBitcoinOnly } from '@/routes/account/utils';
 import { useDarkmode } from '@/hooks/darkmode';
 import { getConfig } from '@/utils/config';
-import style from './iframe.module.css';
 import { i18n } from '@/i18n/i18n';
 import { alertUser } from '@/components/alert/Alert';
 import { parseExternalBtcAmount } from '@/api/coins';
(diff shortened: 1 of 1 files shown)


frontends/web/src/routes/market/bitrefill.tsx
```diff
@@ -225,6 +235,23 @@ export const Bitrefill = ({ accounts, code }: TProps) => {
                     }}
                   />
                 )}
+
+                { verifyPaymentRequest && verifyPaymentRequest.success && (
+                  <WaitDialog title={t('receive.verifyBitBox02')}>
+                    <p>
+                      Address<br />
```
thisconnect: I forgot to translate those and will update the PR..

frontends/web/src/routes/market/bitrefill.tsx
```diff
@@ -137,11 +142,13 @@ export const Bitrefill = ({ accounts, code }: TProps) => {
       const parsedAmount = await parseExternalBtcAmount(paymentAmount.toString());
       if (!parsedAmount.success) {
         alertUser(t('unknownError', { errorMessage: 'Invalid amount' }));
+        setPendingPayment(false);
         return;
       }
       // Ensure expected payment method matches account
       if (coinMapping[account.coinCode] !== paymentMethod) {
         alertUser(t('unknownError', { errorMessage: 'Payment method mismatch' }));
+        setPendingPayment(false);
```
Beerosagos: I think we need a return too here

frontends/web/src/routes/market/bitrefill.tsx
```diff
@@ -61,6 +63,9 @@ export const Bitrefill = ({ accounts, code }: TProps) => {
   const config = useLoad(getConfig);
   const [agreedTerms, setAgreedTerms] = useState(false);
 
+  const [pendingPayment, setPendingPayment] = useState<boolean>(false);
+  const [verifyPaymentRequest, setVerifyPaymentRequest] = useState<TTxProposalResult & { address: string } | false>();
```
Beerosagos: Initialize to false?

Beerosagos (approved):
tACK with a couple of nits 🙏

frontends/web/src/routes/market/bitrefill.tsx
thisconnect: good catch 🙏

PR #3556, opened 2025-09-15
backend: fetch multiple xpubs for unified btc accounts at once

benma:
This is to reduce the number of Optiga secure chip operations needed in the BitBox02. Before it would be two operations per xpub, with 3 xpubs per account, with up to 6 accounts by default during discovery. This reduces the number of operations by a factor of 3 (one call for all three xpubs of an account).

For keystores that don't support it natively (e.g. older BitBox firmware), it falls back to making one API call per xpub like before.

(Depends on https://github.com/BitBoxSwiss/bitbox02-api-go/pull/138 being merged and included here first)


bznein (approved):
utACK nice one!

PR #3558, opened 2025-09-16
frontend: send to self dropdown

shonsirsha:
Created a new reusable component `InputWithDropdown` and used it for the Send screen (Receiver address input field). 

When there's no eligible account for send-to-self (same like prod):
Screenshot 2025-09-16 at 14 38 59

When there are eligible accounts:
Screenshot 2025-09-16 at 14 37 14

Screenshot 2025-09-16 at 14 43 00

Screenshot 2025-09-16 at 14 37 18

Screenshot 2025-09-16 at 14 37 20

Screenshot 2025-09-16 at 14 37 22

### DIFF
--- a/frontends/web/src/components/dropdown/dropdown.tsx
+++ b/frontends/web/src/components/dropdown/dropdown.tsx
@@ -37,7 +37,7 @@ type SelectProps<T = any, IsMulti extends boolean = false> = Omit<
   ReactSelectProps<TOption<T>>,
   'onChange'
 > & {
-  renderOptions: (selectedItem: TOption<T>) => ReactNode; // Function to render options and selected value for single dropdown
+  renderOptions?: (selectedItem: TOption<T>, isSelectedValue: boolean) => ReactNode; // Function to render options and selected value for single dropdown
   onChange: (
     newValue: IsMulti extends true ? TOption<T>[] : TOption<T>,
     actionMeta: ActionMeta<TOption<T>>
@@ -46,6 +46,7 @@ type SelectProps<T = any, IsMulti extends boolean = false> = Omit<
   isOpen?: boolean;
   onOpenChange?: (isOpen: boolean) => void;
   title?: string;
+  mobileTriggerComponent?: ReactNode | ((props: { onClick: () => void }) => ReactNode);
 };
 
 const DropdownIndicator = (props: DropdownIndicatorProps<TOption>) => (
@@ -59,7 +60,7 @@ const SelectSingleValue = (props: SingleValueProps<TOption> & { selectProps: any
 
   return (
     <components.SingleValue {...props}>
-      {renderOptions(props.data)}
+      {renderOptions(props.data, true)}
     </components.SingleValue>
   );
 };
@@ -68,7 +69,7 @@ const Option = (props: OptionProps<TOption> & { selectProps: any }) => {
   const { renderOptions } = props.selectProps;
   return (
     <components.Option {...props}>
-      {renderOptions(props.data)}
+      {renderOptions(props.data, false)}
     </components.Option>
   );
 };
@@ -109,6 +110,7 @@ export const Dropdown = <T, IsMulti extends boolean = false>({
   mobileFullScreen = false,
   isOpen,
   onOpenChange,
+  mobileTriggerComponent,
   ...props
 }: SelectProps<T, IsMulti>) => {
   const isMobile = useMediaQuery('(max-width: 768px)');
@@ -128,12 +130,13 @@ export const Dropdown = <T, IsMulti extends boolean = false>({
       <MobileFullscreenSelector
         title={title}
         options={options}
-        renderOptions={renderOptions}
+        renderOptions={renderOptions || (() => null)}
         value={props.value as any}
         onSelect={onChange}
         isMulti={props.isMulti}
         isOpen={isOpen}
         onOpenChange={onOpenChange}
+        triggerComponent={mobileTriggerComponent}
       />
     );
   }
--- a/frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
+++ b/frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
@@ -25,12 +25,13 @@ import styles from './mobile-fullscreen-selector.module.css';
 type Props<T, IsMulti extends boolean = false> = {
   title: string;
   options: TOption<T>[];
-  renderOptions: (option: TOption<T>) => ReactNode;
+  renderOptions: (option: TOption<T>, isSelectedValue: boolean) => ReactNode;
   value: IsMulti extends true ? TOption<T>[] : TOption<T>;
   onSelect: (newValue: IsMulti extends true ? TOption<T>[] : TOption<T>, actionMeta: ActionMeta<TOption<T>>) => void;
   isMulti?: boolean;
   isOpen?: boolean;
   onOpenChange?: (isOpen: boolean) => void;
+  triggerComponent?: ReactNode | ((props: { onClick: () => void }) => ReactNode);
 }
 
 export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
@@ -42,6 +43,7 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
   isMulti,
   isOpen: controlledIsOpen,
   onOpenChange,
+  triggerComponent,
 }: Props<T, IsMulti>) => {
   const [localIsOpen, setLocalIsOpen] = useState(false);
   const [searchText, setSearchText] = useState('');
@@ -112,10 +114,12 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
         return false;
       }}
       />
-      {onOpenChange ? (
-        <div
-          className={styles.mobileSelectorTrigger}
-        >
+      {triggerComponent ? (
+        typeof triggerComponent === 'function' ?
+          triggerComponent({ onClick: handleOpen }) :
+          triggerComponent
+      ) : onOpenChange ? (
+        <div className={styles.mobileSelectorTrigger}>
           <span className={styles.mobileSelectorValue}>{displayValue}</span>
         </div>
       ) : (
@@ -125,8 +129,7 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
         >
           <span className={styles.mobileSelectorValue}>{displayValue}</span>
         </button>
-      )
-      }
+      )}
 
       {isOpen && (
         <div className={styles.fullscreenOverlay}>
@@ -148,6 +151,7 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
 
             <div className={styles.searchContainer}>
               <input
+                id="search"
                 type="text"
                 className={styles.searchInput}
                 placeholder={`${t('generic.search')}`}
@@ -160,14 +164,16 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
               {filteredOptions.length > 0 ? (
                 filteredOptions.map((option) => (
                   <button
-                    key={String(option.value)}
+                    key={JSON.stringify(option.value)}
                     className={`
-                      ${styles.optionItem || ''}
-                      ${isSelected(option) && styles.selectedOption || ''}
-                    `}
+                    ${styles.optionItem || ''} 
+                    ${isSelected(option) ?
+                    styles.selectedOption || ''
+                    : ''}`
+                    }
                     onClick={(e) => handleSelect(option, e)}
                   >
-                    <div className={styles.optionContent}>{renderOptions(option)}</div>
+                    <div className={styles.optionContent}>{renderOptions(option, false)}</div>
                   </button>
                 ))
               ) : (
--- a/frontends/web/src/components/forms/input-with-dropdown.tsx
+++ b/frontends/web/src/components/forms/input-with-dropdown.tsx
@@ -1,0 +1,113 @@
+/**
+ * Copyright 2025 Shift Crypto AG
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { forwardRef } from 'react';
+import { TBaseInputProps } from './types';
+import { Dropdown, TOption } from '@/components/dropdown/dropdown';
+import { ChevronLeftDark } from '@/components/icon';
+import baseStyles from './input.module.css';
+import styles from './input-with-dropdown.module.css';
+
+export type TInputWithDropdownProps<T> = TBaseInputProps & {
+  dropdownOptions?: TOption<T>[];
+  dropdownValue?: TOption<T> | null;
+  onDropdownChange?: (selected: TOption<T>) => void;
+  dropdownPlaceholder?: string;
+  dropdownTitle?: string;
+  isOptionDisabled?: (option: TOption<T>) => boolean;
+  renderOptions?: (option: TOption<T>, isSelectedValue: boolean) => React.ReactNode;
+}
+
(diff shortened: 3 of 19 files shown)


frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -70,35 +83,108 @@ export const ReceiverAddressInput = ({
...
+  }));
+
+  const handleSendToAccount = useCallback(async (selectedOption: TOption<accountApi.IAccount | null>) => {
+    if (selectedOption.value === null) {
+      return;
+    }
+    const selectedAccountValue = selectedOption.value;
+    setSelectedAccount(selectedOption);
+    try {
+      const receiveAddresses = await getReceiveAddressList(selectedAccountValue.code)();
```
benma: If the account is not yet synced, this will never respond, and hang here forever I think. We need to adjust the backend to return a proper error here if unsynced, and then think about how to handle this problem :thinking:

frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
```diff
@@ -113,18 +115,30 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
...
-        >
-          <span className={styles.mobileSelectorValue}>{displayValue}</span>
-        </div>
+        triggerComponent ? (
+          typeof triggerComponent === 'function' ?
+            triggerComponent({ onClick: handleOpen }) :
+            triggerComponent
+        ) : (
+          <div
+            className={`${styles.mobileSelectorTrigger}`}
```
thisconnect: nit: please use
```suggestion
            className={styles.mobileSelectorTrigger}
```

As styles.mobileSelectorTrigger might be undefined and TS with complain in the future  with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3534

frontends/web/src/routes/account/send/components/inputs/receiver-address-wrapper.tsx
```diff
@@ -0,0 +1,197 @@
...
+      acc.active &&
+      acc.code !== currentAccount?.code
+    ) || [], [accounts, currentAccount]);
+
+  const shouldShowSendToAccountDropdown = sameCoinAccounts.length > 0;
+
+  const accountOptions: TAccountOption[] = sameCoinAccounts.map(account => ({
+    label: account.name,
+    value: account,
+    disabled: !accountSyncStatus[account.code]?.synced
```
shonsirsha: When not `synced`

Screenshot 2025-09-23 at 08 20 25

shonsirsha: cc @benma

frontends/web/src/components/forms/input-with-dropdown.tsx
```diff
@@ -0,0 +1,133 @@
...
+  children,
+  ...props
+}: TInputWithDropdownProps<any>, ref) => {
+  return (
+    <div className={[
+      styles.input,
+      styles[`align-${align}`],
+      className,
+      transparent ? styles.isTransparent : '',
+    ].join(' ')}>
```
thisconnect: nit: In other components you use multiline backticks, I find those much nicer and would love to use them everywhere instead of joining an array. WDYT?

frontends/web/src/components/dropdown/dropdown.tsx
```diff
@@ -46,9 +46,10 @@ type SelectProps<T = any, IsMulti extends boolean = false> = Omit<
   isOpen?: boolean;
   onOpenChange?: (isOpen: boolean) => void;
   title?: string;
+  mobileTriggerComponent?: ReactNode | ((props: { onClick: () => void }) => ReactNode);
 };
 
-const DropdownIndicator = (props: DropdownIndicatorProps<TOption>) => (
+export const DropdownIndicator = (props: DropdownIndicatorProps<TOption>) => (
```
thisconnect: Does this need to be exported?

frontends/web/src/components/forms/input-with-dropdown.tsx
```diff
@@ -0,0 +1,133 @@
...
+  dropdownOptions?: TOption<T>[];
+  dropdownValue?: TOption<T> | null;
+  onDropdownChange?: (selected: TOption<T>) => void;
+  dropdownPlaceholder?: string;
+  dropdownTitle?: string;
+  showDropdown?: boolean;
+  children?: React.ReactNode;
+} & Omit<HTMLProps<HTMLInputElement>, 'onInput'>
+
+export const InputWithDropdown = forwardRef<HTMLInputElement, TInputWithDropdownProps<any>>(({
```
thisconnect, line 34: I can't find where `<InputWithDropdown />` is used 🤔

frontends/web/src/routes/account/send/components/inputs/receiver-address-wrapper.tsx
```diff
@@ -0,0 +1,197 @@
...
+      subscriptions.push(syncdone(account.code, () => checkAccountStatus(account.code)));
+    });
+    return () => unsubscribe(subscriptions);
+  }, [sameCoinAccounts, checkAccountStatus]);
+
+  return (
+    <div className={[
+      styles.input,
+      styles['align-left'],
+    ].join(' ')}>
```
thisconnect: also here, I think backticks would be nicer.

thisconnect (changes requested):
super cool I tested a bit (not fully tested yet).

frontends/web/src/components/forms/input-with-dropdown.tsx
shonsirsha: Aah, this shouldve been used in the address-wrapper component. I created the component but stopped there :| thanks👍

frontends/web/src/components/dropdown/dropdown.tsx
shonsirsha: Nope! I think I was trying to use this, but not needed as we're already using it. Good catch,  thx.

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -1,3 +1,4 @@
+
```
thisconnect: newline

frontends/web/src/components/dropdown/dropdown.tsx
```diff
@@ -161,7 +166,6 @@ export const Dropdown = <T, IsMulti extends boolean = false>({
           : (onChange as (value: TOption<T>, actionMeta: ActionMeta<TOption<T>>) => void);
         handleChange(selected as any, actionMeta);
       }}
-      {...props}
```
thisconnect, line 164: this looks a bit like a risky change. :)

frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
```diff
@@ -113,18 +115,30 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
...
+            triggerComponent({ onClick: handleOpen }) :
+            triggerComponent
+        ) : (
+          <button
+            className={styles.mobileSelectorTrigger}
+            onClick={handleOpen}
+          >
+            <span className={styles.mobileSelectorValue}>{displayValue}</span>
+          </button>
+        )
```
thisconnect: I can't spot the difference, looks like code duplication ?

frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -1,3 +1,4 @@
+
```
thisconnect: newline :)

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -83,6 +86,7 @@ export const Send = ({
   const proposeTimeout = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   const [recipientAddress, setRecipientAddress] = useState<string>('');
+  const [actualRecipientAddress, setActualRecipientAddress] = useState<string>('');
```
thisconnect: when looking at this diff it is hard to understand what the intention of actualRecipientAddress is.

frontends/web/src/routes/account/send/components/confirm/confirm.tsx
```diff
@@ -123,10 +125,16 @@ export const ConfirmSend = ({
...
             <p className={`${style.valueOriginal}`}>
-              {recipientAddress || 'N/A'}
+              {accountName ? accountName : recipientAddress}
             </p>
+            {accountName && (
+              <span className={style.address}>
+                {recipientAddress}
+              </span>
+            )}
+            <p></p>
```
thisconnect: needed?

frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
shonsirsha: Yeah there are some duplication. Will simplify. But in essence it allows for 3 things to be used for the "trigger" for the mobile dropdown:
- use your own controlled (stateful) trigger 
- use the component's trigger but make it controlled (use your own state / handle)
- use the component's state and trigger

frontends/web/src/routes/account/send/components/confirm/confirm.tsx
shonsirsha: Oops!

frontends/web/src/components/dropdown/dropdown.tsx
```diff
@@ -161,7 +165,6 @@ export const Dropdown = <T, IsMulti extends boolean = false>({
...
```
thisconnect, line 164: Before props could overwrite existing props. With this change the existing props here will overwrite everything that comes via props. Is this intended and are there no side-effect since Dropdown is also used in other places?

frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
```diff
@@ -160,11 +163,8 @@ export const MobileFullscreenSelector = <T, IsMulti extends boolean = false>({
               {filteredOptions.length > 0 ? (
                 filteredOptions.map((option) => (
                   <button
-                    key={String(option.value)}
-                    className={`
-                      ${styles.optionItem || ''}
-                      ${isSelected(option) && styles.selectedOption || ''}
-                    `}
+                    key={JSON.stringify(option.value)}
+                    className={`${styles.optionItem || ''} ${isSelected(option) ? styles.selectedOption || '' : ''}`}
```
thisconnect: nit: for me it's slightly more readable with newlines.

frontends/web/src/components/forms/input-with-dropdown.tsx
```diff
@@ -0,0 +1,119 @@
...
+import styles from './input-with-dropdown.module.css';
+
+export type TInputWithDropdownProps<T> = {
+  align?: 'left' | 'right';
+  className?: string;
+  error?: string | object;
+  onInput?: (e: ChangeEvent<HTMLInputElement>) => void;
+  transparent?: boolean;
+  labelSection?: JSX.Element | undefined;
+  label?: string;
```
thisconnect: nit: instead of redefining, could we import from type TInputProps from input.tsx and extend?

thisconnect: … scrolling down and checking the rest of the component I found it does not use input.tsx...

it looks almost the same as in input.tsx and Dropdown is just after children. I didn't check closely if there is minor differences. Did you check if we could use Input component or is there some other structure needed ?

frontends/web/src/locales/en/app.json
```diff
@@ -1707,6 +1708,11 @@
     "priority": "Priority",
     "scanQR": "Scan QR code",
     "scanQRNoCameraMessage": "Camera not found. Please ensure that your device supports a camera and permissions are correctly set.",
+    "sendToAccount": {
+      "placeholder": "Select account",
+      "reset": "Reset selection",
```
thisconnect: I can't find where sendToAccount.reset was used, remove?

frontends/web/src/routes/account/utils.ts
```diff
@@ -148,3 +148,19 @@ export const getAccountsPerCoin = (accounts: IAccount[]): TAccountCoinMap => {
...
 };
+
+export const getAccountNumber = (targetAccount: IAccount, allAccounts: IAccount[]): number | null => {
+  try {
+    const sameKeystoreAndCoin = allAccounts.filter(account =>
+      account.keystore.rootFingerprint === targetAccount.keystore.rootFingerprint &&
+      account.coinCode === targetAccount.coinCode
+    );
+
+    sameKeystoreAndCoin.sort((a, b) => a.code.localeCompare(b.code));
```
thisconnect:
```suggestion
    sameKeystoreAndCoin.sort((a, b) => a.code.localeCompare(b.code, i18n.language));
```

frontends/web/src/routes/router.tsx
```diff
@@ -114,7 +114,8 @@ export const AppRouter = ({ devices, deviceIDs, devicesKey, accounts, activeAcco
   const AccSend = (<InjectParams>
     <SendWrapper
       code={'' /* dummy to satisfy TS */}
-      accounts={activeAccounts} />
+      accounts={activeAccounts}
```
thisconnect: When reading allAccounts, I wonder if that is accounts or just activeAccounts.

Here it passes activeAccounts -> accounts then in the wrapper accounts -> allAccounts,  as we already use activeAccounts, accounts and account, I'd prefer if the prop name was just activeAccounts also in send and sendwrapper..

but I see we use allAccounts also in other places, I think we should rename it there too (in a different PR)

frontends/web/src/routes/account/send/components/confirm/confirm.tsx
```diff
@@ -67,10 +70,20 @@ export const ConfirmSend = ({
...
     recipientAddress,
     activeCurrency: fiatUnit
   } = transactionDetails;
 
+  const selectedAccountWithNumber = selectedReceiverAccount && allAccounts
+    ? (() => {
+      const accountNumber = getAccountNumber(selectedReceiverAccount, allAccounts);
+      return accountNumber
+        ? { name: selectedReceiverAccount.name, number: accountNumber }
+        : { name: selectedReceiverAccount.name, number: undefined };
```
thisconnect: nit: getAccountNumber returns number | null, but if it would just `return;` here you could just pass accountNumber without the condition.

frontends/web/src/components/dropdown/dropdown.tsx
shonsirsha: I think I missed addressing this comment. Not intended. Thanks!

frontends/web/src/components/forms/input-with-dropdown.tsx
shonsirsha: Structurally it's indeed different, but at the same time there are duplicated props and classnames (but also very different ones to account for the dropdown). **I'll refactor and share the code which can be used for both components.** 

It's possible to make the input component more "flexible" but this increases its complexity (add prop for wrapper, another prop the dropdown component itself...) I prefer we completely separate them.

frontends/web/src/components/dropdown/mobile-fullscreen-selector.tsx
shonsirsha: Sure

frontends/web/src/routes/router.tsx
shonsirsha: > I'd prefer if the prop name was just activeAccounts also in send and sendwrapper.

Let's do that

frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -14,47 +14,67 @@
...
 }: TReceiverAddressInputProps) => {
   const { t } = useTranslation();
   const isMobile = useMediaQuery('(max-width: 768px)');
   const [activeScanQR, setActiveScanQR] = useState(false);
+  const accountCode = account?.code;
+
+  const accountsForReceiverDropdown = useMemo(() =>
+    activeAccounts?.filter(acc =>
+      acc.coinCode !== 'sepeth' &&
+      acc.coinCode !== 'eth' &&
```
thisconnect: tested servewallet-mainnet + webdev

ETH is excluded on purpose correct?

but erc20 tokens are still showing up in the dropdown, I assume those should be excluded as well?

Please mention in the commit message what the intention is or whether this is intended behavior.

thisconnect: if only btc and ltc should have the dropdown you could use isBitcoinBased from utils

thisconnect (approved):
LGTM

PR #3559, opened 2025-09-16
Frontend update dialog

thisconnect:
- [ ] updating the Changelog
- [ ] writing unit tests


strmci:
tACK

frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -44,126 +44,58 @@ export const Dialog = ({
...
-      }
-    };
-
-    // simulate transition duration fallback
-    const timer = setTimeout(finish, 400);
-    return () => clearTimeout(timer);
-  }, [onClose]);
+  }, [open, isVisible, onClose]);
+
+  // Close on ESC
```
shonsirsha: I think we can delete the comments (also in the new hook). Wdyt?

frontends/web/src/hooks/keyboard.ts
```diff
@@ -53,3 +53,68 @@ export const useEsc = (
...
+  const previouslyFocused = useRef<HTMLElement | null>(null);
+
+  // Focus trap handler
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (!active || !ref.current || e.key !== 'Tab') {
+      return;
+    }
+    const node = ref.current;
+    const focusables = node.querySelectorAll<HTMLElement>(
+      'a, button, input, textarea, select, [tabindex]:not([tabindex="-1"])'
```
shonsirsha: Kind of nit/question: there's a slight difference between this set of selector and the one below it. Does it make sense to unify them as a constant? 

Also, maybe makes sense to add `:not` disabled.. Afaik it's not possible to focus on disabled elements. I'm not sure how this would behave if one/more elements are disabled (would the focus be lost?)

thisconnect: when opening the dialog we want to focus the element that has an autofocus attribute.

The other selector list is just for the trap, when a user cycles through all elements with "Tab" key. Here we don't care about autofocus.

Will update to make this more explicit and so we can share the selector string.

PR #3560, opened 2025-09-16
maintenance: make BBApp able to talk to simulator.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests


backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -0,0 +1,106 @@
+package simulator
```
benma, line 15: License header missing

backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	device := bitbox02.NewDevice(SimulatorID, version, common.ProductBitBox02Multi,
+		&mocks.Config{}, communication,
+	)
+	return device, nil
+}
+
+// IsRunning checks if a BitBox02 simulator with the given version is running.
+func IsRunning(version string) bool {
+	cmd := exec.Command("pgrep", "-f", fmt.Sprintf("%s-simulator", version))
```
benma: Does this assume a certain filename format? If so, maybe not so robust :thinking: I wonder if there is a command to check if a port is being listened to / used instead.

backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := cmd.Run(); err != nil {
+		return false // no matching process
+	}
+	return true
+}
+
+// DeviceInfo implements usb.DeviceInfo for the simulator.
+// All methods return hardcoded values.
+type DeviceInfo struct {
```
benma: call it `simulatorDeviceInfo`?

backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Interface implements usb.DeviceInfo.
+func (d DeviceInfo) Interface() int {
+	return 0
+}
+
+// Serial implements usb.DeviceInfo.
+func (d DeviceInfo) Serial() string {
+	return "simulator"
```
benma: Since you decided to pass the simulator off as a USB device with a DeviceInfo (I assume so the USB manager is also tested as a result), then why not go all the way?

- Return the fw version here, then usb manager will parse it out automatically  
- Connect inside d.Open() and return the net-based ReadWriteCloser there
- Disconnect in d.Close()?

`manager.makeBitBox02()` will then automatically handle creating the device etc, and your custom isSimulator() case in manager.go would not be needed. In fact, all the changes to backend.go and manager.go would then probably not be needed.

backend/devices/bitbox02/simulator/simulator.go
bznein: Yeah I agree it's not the most robust solution. 

> I wonder if there is a command to check if a port is being listened to / used instead.

I tried that, but couldn't get it working. I don't know much about the details of networking but whenever I tried to check if the port was being listened to, the active connection with the simulator would drop.

backend/devices/bitbox02/simulator/simulator.go
bznein: Since it's already in the `simulator` package I'd rather not, to avoid the stuttering in `simulator.simulatorDeviceInfo`, WDYT?

backend/devices/bitbox02/simulator/simulator.go
bznein: Thanks, this looks so much cleaner! I'll push a commit soon!!

backend/devices/bitbox02/simulator/simulator.go
bznein: Another way would be to add a third argument that is the filename format. I don't like any of these three to be honest :D This one is the only one that I managed to get working

backend/devices/bitbox02/simulator/simulator.go
benma: makes sense! didn't realize this

backend/devices/bitbox02/simulator/simulator.go
benma:
```
func isRunning(port int) bool {
	cmd := exec.Command("lsof", "-i", fmt.Sprintf(":%d", port), "-sTCP:LISTEN")
	if err := cmd.Run(); err != nil {
		return false // No process listening on the port
	}
	return true
}
```

this seems to work for me. have you tried this?

cmd/servewallet/main.go
```diff
@@ -138,6 +144,8 @@ func main() {
 	devservers := flag.Bool("devservers", true, "switch to dev servers")
 	gapLimitsReceive := flag.Uint("gapLimitReceive", 0, "gap limit for receive addresses")
 	gapLimitsChange := flag.Uint("gapLimitChange", 0, "gap limit for change addresses")
+	simulatorPort := flag.Int("simulatorPort", -1, "port for the BitBox02 simulator")
```
benma: I'd default this to 15423 for convenience, and add another bool flag `simulator` to switch on the simulator.  It's always a bit annoying to look up the port. It should also be less confusing. 

Additionally, add a Makefile target `make servewallet-simulator` ?

cmd/servewallet/main.go
```diff
@@ -138,6 +144,8 @@ func main() {
...
+	simulatorVersion := flag.String("simulatorVersion", "", "version of the BitBox02 simulator")
```
benma: This and the product string are also an annoyance. If `NewDevice()` receives `nil` for version and product, it will infer it using the getInfo API endpoint:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/c9b4882325b847da24a5a120bc4a2f6b5dd30579/vendor/github.com/BitBoxSwiss/bitbox02-api-go/api/firmware/device.go#L138-L147

The problem right now is that manager.go always supplies it from the device info, but maybe it can be refactored a little to pass `nil` in the case of the simulator. 

The product also in this PR is hardcoded to BB02 Multi, but since v9.24.0, the simulator is a BB02 Nova Multi. But both would be inferred if not provided, if I am not wrong.

backend/devices/bitbox02/simulator/simulator.go
bznein: It does work, thanks!

cmd/servewallet/main.go
bznein: > I'd default this to 15423 for convenience, and add another bool flag simulator to switch on the simulator. It's always a bit annoying to look up the port. It should also be less confusing.

> Additionally, add a Makefile target make servewallet-simulator ?

Sure, I'll need it for Playwright later, is it ok in a different PR?

cmd/servewallet/main.go
bznein: I *think* I got what you meant :D Can you please take a look at the latest changes and let me know if this is what you meant? It seems to work fine (the code does need a few changes maybe, I quickly put together something just to test), as going to "manage device" shows the correct values (e.g. the version, which is not provided anymore as a flag)

backend/devices/bitbox02/simulator/simulator.go
benma: AI saves the day again :grin:

backend/devices/bitbox02/device.go
```diff
@@ -53,28 +53,31 @@ type Device struct {
...
 	communication firmware.Communication,
 	opts ...firmware.DeviceOption,
 ) *Device {
 	productName := BitBox02ProductName
 	// BitBox02Plus is the internal code name for the BitBox Nova.
-	if product == bitbox02common.ProductBitBox02PlusBTCOnly ||
-		product == bitbox02common.ProductBitBox02PlusMulti {
+	if product != nil && (*product == bitbox02common.ProductBitBox02PlusBTCOnly ||
+		*product == bitbox02common.ProductBitBox02PlusMulti) {
 		productName = BitBox02NovaProductName
```
benma, line 65: this also has to happen if the product name is inferred in instead of passed directly

backend/devices/bitbox02/device.go
```diff
@@ -53,28 +53,31 @@ type Device struct {
...
 	}
 	log := logging.Get().
 		WithGroup("device").
 		WithField("deviceID", deviceID).
-		WithField("productName", productName).
-		WithField("product", product)
+		WithField("productName", productName)
+
+	if product != nil {
+		log = log.WithField("product", *product)
```
benma: same, would be nice to log it also if it was inferred vs passed

backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Serial implements usb.DeviceInfo.
+func (d deviceInfo) Serial() string {
+	return ""
+}
+
+// Product implements usb.DeviceInfo.
+func (d deviceInfo) Product() string {
+	return bitbox02common.FirmwareDeviceProductStringBitBox02Multi
```
benma: would not need that anymore

benma (changes requested):
Actually forgot there were some comments left to address :innocent:

backend/devices/bitbox02/simulator/simulator.go
bznein: I actually asked AI and gave me a couple different options, but not that one :D

bznein:
@benma ready again :)

backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Interface implements usb.DeviceInfo.
+func (d deviceInfo) Interface() int {
+	return 0
+}
+
+// Serial implements usb.DeviceInfo.
+func (d deviceInfo) Serial() string {
+	return fmt.Sprintf("v%s", d.version.String())
```
benma: Fyi, `.String()` is technically not needed, as it is called when using `%s`

bznein: Apparently this causes the tests to fail: https://github.com/BitBoxSwiss/bitbox-wallet-app/actions/runs/17909896855/job/50918793278?pr=3560

PR #3568, opened 2025-09-22
frontend: add search to transaction history view

shonsirsha:
Enable searching transactions by note or transaction ID via a new component in the transaction history view. This improves UX by making it easier to find
specific records.

This new component replaces the position of the "Export" (export transactions) button, which has been moved to the account info view.

Screenshot 2025-09-22 at 06 18 47
Screenshot 2025-09-22 at 06 18 50
Screenshot 2025-09-22 at 06 19 00


frontends/web/src/locales/en/app.json
```diff
@@ -5,7 +5,7 @@
     "backupReminder": "Your wallet <strong>{{name}}</strong> passed {{fiat}} {{threshold}}!\n\nWe recommend creating a paper backup for extra protection. It's quick and simple.",
     "backupReminderLink": "Create paper backup",
     "disconnect": "Connection lost. Retrying…",
-    "export": "Export",
+    "export": "Export transactions",
```
shonsirsha, line 8: Need to change in locize.

frontends/web/src/hooks/debounce.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright 2018 Shift Devices AG
```
thisconnect:
```suggestion
 * Copyright 2025 Shift Crypto AG
```

frontends/web/src/hooks/debounce.ts
```diff
@@ -0,0 +1,40 @@
...
+      setDebouncedValue(value);
+    }, delay);
+
+    return () => {
+      clearTimeout(handler);
+    };
+  }, [value, delay]);
+
+  return debouncedValue;
+};
```
thisconnect, line 43: A few other hooks use useMounted internally would it make sense here too?

frontends/web/src/routes/account/account.tsx
```diff
@@ -311,37 +356,62 @@ const RemountAccount = ({
...
+                      </Button>
+                    </div>
+
+                    <div className={`${style.searchContainer} ${!showSearchBar ? style.searchHidden : ''}`}>
+                      <Input
+                        type="text"
+                        className={style.searchInput}
+                        placeholder="Search transactions..."
+                        value={searchTerm}
+                        onChange={(e) => setSearchTerm(e.currentTarget.value)}
```
thisconnect, line 377: Could you try to debounce here, so that the state doesn't change while typing ?

shonsirsha: Not 100% sure what you meant here. Could you elaborate?


We are already debouncing the filtering (which could be expensive with many transactions) - but the input field is updating `searchTerm` directly and will show the typed characters immediately (which is good). 

If we debounce this, we need to make the input component to be uncontrolled (otherwise it'll appear as a "lag" for the user).

frontends/web/src/routes/account/account.tsx
```diff
@@ -103,6 +107,37 @@ const RemountAccount = ({
...
+
+      return noteMatch || addressMatch || txIdMatch;
+    });
+  }, [transactions, debouncedSearchTerm]);
+
+  const hasFilteredTransactions = filteredTransactions.length > 0;
+
+  const handleShowDetail = useCallback((internalID: accountApi.ITransaction['internalID']) => {
+    setDetailID(internalID);
+  }, []);
```
thisconnect: nit: is this useCallback important or could we just use?
```ts
const handleShowDetail = (internalID: accountApi.ITransaction['internalID']) => setDetailID(internalID);
```
if it is important to use useCallback maybe add a comment so people in the future see it is needed for performance or similar.

frontends/web/src/routes/account/components/transaction-list.tsx
```diff
@@ -0,0 +1,73 @@
...
+        ))}
+      </>
+    ) : (
+      <p className={style.emptyTransactions}>
+        {t('transaction.no-results', { searchTerm: debouncedSearchTerm })}
+      </p>
+    );
+  }
+
+  return transactions?.success ? (
```
thisconnect: I am not sure if TransactionList should accept both: transactions and filteredTransactions. 

If there is no searchterm or searchterm is empty string, then filteredTransactions is the same as  transactions.list.

in account.tsx there is already some condition around TransactionList `{hasTransactions ? ( ) : transactions?.success && (` but half of the conditions still is in account.tsx but some conditions are here.

To me this component is not really useful or too tightly coupled with account.tsx, and not really standalone/reusable component. It might as well all be just in account.tsx which may make sense as well.

or what do you think?

frontends/web/src/routes/account/account.tsx
shonsirsha: Technically yes it improves performance if we also memoise Transaction component, but let's just get rid of this (in fact the whole handler). It's a very inexpensive function anyway.

frontends/web/src/routes/account/components/transaction-list.tsx
shonsirsha: I see your point in that it is confusing. I re-structured it. 
 
> To me this component is not really useful or too tightly coupled with account.tsx, and not really standalone/reusable component. 

Agree again on that this isn't a standalone/reusable but breaking component into parts isn't just for reusability but also readability / maintainability.  This pattern is something we already use. It's a lot easier to read the code for me personally, as well as to maintain it (e.g adding memoisation) if we extract it into a different component. 

Different file because `accounts.tsx` is already quite long, personally not a fan of creating another component there.

shonsirsha: > It's a lot easier to read the code for me personally

Maybe you'll agree to this after seeing the updated code 😄 This change you commented on indeed is kinda confusing.

frontends/web/src/routes/account/info/info.module.css
```diff
@@ -47,6 +47,12 @@
     margin-bottom: var(--space-half);
 }
 
+.exportButton {
+    margin-bottom: var(--space-eight);
+    padding-left: 0;
+
```
thisconnect: nit: remove newline 🙏

frontends/web/src/routes/account/account.tsx
```diff
@@ -316,37 +329,58 @@ const RemountAccount = ({
...
+                    </div>
+
+                    <div className={`${style.searchContainer} ${!showSearchBar ? style.searchHidden : ''}`}>
+                      <Input
+                        type="text"
+                        className={style.searchInput}
+                        placeholder="Search transactions..."
+                        value={searchTerm}
+                        onChange={(e) => setSearchTerm(e.currentTarget.value)}
+                        autoFocus={showSearchBar}
```
thisconnect: the autofocus doesn't work for me, I guess at this point the animation is sill happening and the button has focus.

But something we can improve later

shonsirsha: Yeah.. will fix with ref

PR #3570, opened 2025-09-22
frontend/qt: fix loading issue in new market routes

thisconnect:
Fixes a regresssion that iframes under the new market route do not load anymore and show the following error:

    The web page at ... might be temporarily down or it may have
    moved permanently to a new web address.

Introduced in:
- 160afee8f5b9498d4038205c4bdc282387c17307

- [ ] updating the Changelog
- [ ] writing unit tests


thisconnect:
tested with the Qt build from the CI, now all widgets load again on macOS

PR #3574, opened 2025-09-29
chore: increase servewallet timeout.

bznein:
To avoid spurious failures when the servewallet takes more times to be ready, we increase the timeout but also we log how long a successful connection took, in order to help keep track of it.

Looking at one execution, we were very close to the timeout limit so it does make sense that we had intermittent failures.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/tests/helpers/servewallet.ts
+++ b/frontends/web/tests/helpers/servewallet.ts
@@ -42,7 +42,7 @@ export async function waitForServewallet(
   servewalletPort: number,
   frontendPort: number,
   host: string,
-  timeout = 60000,
+  timeout = 90000,
 ) {
   const start = Date.now();
 
@@ -50,9 +50,12 @@ export async function waitForServewallet(
     try {
       await connectOnce(host, servewalletPort);
       await page.goto(`http://${host}:${frontendPort}`);
+      const elapsed = Date.now() - start;
+      console.log(`Connected to servewallet on ${host}:${servewalletPort} after ${elapsed} ms`);
       return;
     } catch {
       await new Promise(r => setTimeout(r, 200));
+
     }
   }
   throw new Error(`Timeout exceeded waiting for servewallet on ${host}:${servewalletPort}`);


frontends/web/tests/helpers/servewallet.ts
```diff
@@ -42,17 +42,22 @@ export async function waitForServewallet(
...
 ) {
   const start = Date.now();
 
   while (Date.now() - start < timeout) {
     try {
       await connectOnce(host, servewalletPort);
       await page.goto(`http://${host}:${frontendPort}`);
+      const elapsed = Date.now() - start;
+      console.log(`Connected to servewallet on ${host}:${servewalletPort} after ${elapsed} ms`);
+      ;
```
thisconnect: nit: not needed

frontends/web/tests/helpers/servewallet.ts
```diff
@@ -42,17 +42,22 @@ export async function waitForServewallet(
...
+
       return;
     } catch {
       await new Promise(r => setTimeout(r, 200));
+
     }
   }
   throw new Error(`Timeout exceeded waiting for servewallet on ${host}:${servewalletPort}`);
```
thisconnect, line 61: The timeout in the CI errored on this line, I wonder if we shouldn't put the timeout elapsed message before this line instead of in the while loop?

frontends/web/tests/helpers/servewallet.ts
bznein: I actually added it on purpose :) I think keeping track of it can help us make sure that the value we use is "reasonable". WDYT?

frontends/web/tests/helpers/servewallet.ts
bznein: I'm not sure what you mean, in the while we look a successful connection, here we return an error. What do you suggest we change here?

frontends/web/tests/helpers/servewallet.ts
bznein: (unless you meant the extra `;` that I mistakenly added and removed now :D)

frontends/web/tests/helpers/servewallet.ts
thisconnect: oups I didn't read it properly sorry, I looked at the new message that contains elapsed time, but expected it to be a fail error message. 

all good from my side 😇

frontends/web/tests/helpers/servewallet.ts
thisconnect: haha yes I meant only the extra `;`

PR #3577, opened 2025-09-29
backend/bitbox02: bundle v9.23.2 firmwares

benma:
We also rename the asset filenames to match the official release filenames for clarity/consistency.

The tests are split into subtests so the test output clearly shows all asset hashes were checked.

See fw release PR: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1610


Beerosagos (approved):
utACK

PR #3578, opened 2025-09-29
4.48.5: ios: fix timezone and CHANGELOG

benma:
The same issue we fixed on Android also exists on iOS. See comments in code for details.


Beerosagos (approved):
utACK

PR #3586, opened 2025-09-30
frontend: revamp upgrade bootloader layout

thisconnect:
Installing / upgrading looked a bit outdated.

- removed white box
- centered title with BitBox logo
- aligned text with progress bar
- positioned additional upgrade info before and after progress bar for better readability

### DIFF
--- a/frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.module.css
+++ b/frontends/web/src/components/devices/bitbox02bootloader/bitbox02bootloader.module.css
@@ -1,10 +1,16 @@
-.content {
-    font-size: 16px;
-    margin: 0;
+.upgradingTitle {
+    text-align: center;
 }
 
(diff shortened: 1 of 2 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -383,8 +383,8 @@
   "bootloader": {
     "button": "Upgrade firmware now",
     "button_install": "Install firmware now",
-    "progress": "Upgrading: {{progress}}%",
-    "progress_install": "Installing: {{progress}}%",
```
thisconnect, line 387: this is the reason the lint fails in CI, but keeping for now for testing.

Once approved I'll revert and change in locize. Then we need locize pull

thisconnect: updated in locize, will revert here so CI passes.

PR #3587, opened 2025-10-01
backend: fix native-locale on macos in webdev

thisconnect:


cmd/servewallet/main.go
```diff
@@ -113,6 +114,11 @@ func (webdevEnvironment) NativeLocale() string {
 	if v == "" {
 		v = os.Getenv("LANG")
 	}
+	if v == "" && runtime.GOOS == darwin {
```
bznein: just for my understanding: the first part of this `if` is only true if the env var above was not set right?

I think it would be slightly easier to read like this:

```golang
if v == "" {
  lang, ok := os.LookupEnv("LANG");
  if ok {
     v = lang
  } else {
     if out, err := exec.Command("defaults", "read", "-g", "AppleLocale").Output(); err == nil {
	v = strings.Split(strings.TrimSpace(string(out)), "@")[0]
     }
  }
}
```

cmd/servewallet/main.go
```diff
@@ -113,6 +114,11 @@ func (webdevEnvironment) NativeLocale() string {
...
+		if out, err := exec.Command("defaults", "read", "-g", "AppleLocale").Output(); err == nil {
```
bznein: question: what happens here if `err` is not `nil`? We just fallback to the current state (no locale set)? 

Maybe we can add a log in case we had an error?

cmd/servewallet/main.go
thisconnect: Yes exactly, this adds another fallback in case both LC_ALL and LANG return an empty string, which happens for me on macOS on the cli when running `make servewallet`. Qt app is fine, but in theory this fallback may also run in Qt if both LC_ALL/LANG return empty string.

thisconnect: In the previous version it also had a check for `runtime.GOOS == darwin` which is missing in your version. I guess ` exec.Command("defaults", "read", "-g", "AppleLocale")` only runs on macOS? 

I wondered about iOS and according to GPT

> On iOS, runtime.GOOS is also "darwin", but Go programs don’t typically run there in the same way, so we can safely assume macOS only. 

But I guess cmd/servewallet is only used on cli and not run in iOS anyways? but I am not sure. 

Would this be ok? (with a bit of help of gpt)

```go

// NativeLocale naively implements backend.Environment.
// This version is unlikely to work on Windows.
func (webdevEnvironment) NativeLocale() string {
	log := logging.Get().WithGroup("servewallet")
	v := os.Getenv("LC_ALL")
	if v == "" {
		if lang, ok := os.LookupEnv("LANG"); ok {
			v = lang
		} else if runtime.GOOS == "darwin" {
			out, err := exec.Command("defaults", "read", "-g", "AppleLocale").Output()
			if err != nil {
				log.Warnf("failed to read AppleLocale via defaults: %v", err)
			} else {
				v = strings.Split(strings.TrimSpace(string(out)), "@")[0]
			}
		}
	}

	// Strip charset from the LANG. It is unsupported by JS Date formatting
	// used in the frontend and breaks UI in unexpected ways.
	// We are always UTF-8 anyway.
	return strings.Split(v, ".")[0]
}
```

I added `log := logging.Get().WithGroup("servewallet")` as we do in other functions.

One more comment from GPT:

> Exec command safety: exec.Command("defaults", ...) will look in $PATH. On macOS this is fine, but if security is a concern, consider absolute path (/usr/bin/defaults).

bznein: Yeah I think that code would be fine (apologies, when writing mine I forgot the `runtime.GOOS check`). Not sure about the implication of the last message from ChatGPT though. Since we only run this for dev, I think it's fine

thisconnect:
just fyi: the issue that the locale isn't detected on macOS with `make servewallet` only surfaces if there is region in config.json so to reproduce one would need to `rm appfolder.dev/config.json` before `make servewallet`

thisconnect:
thank you.. unfortunately I ran into 1 more issue, fixed in PR https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3589 (based  top of this branch)

PR #3588, opened 2025-10-01
tests: enable automatic tests on Android.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,57 @@ jobs:
           path: /github/home/go/src/github.com/BitBoxSwiss/bitbox-wallet-app/frontends/android/BitBoxApp/app/build/outputs/apk/debug/app-debug.apk
           name: BitBoxApp-android-${{github.sha}}.apk
           if-no-files-found: error
+
+  e2e-android:
+    name: Android E2E Tests
+    needs: android
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        api-level: [30, 31, 32, 33, 34, 35]
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+
+      - name: Download APK
+        uses: actions/download-artifact@v4
+        with:
+          name: BitBoxApp-android-${{ github.sha }}.apk
+          path: frontends/mobiletests/apk
+
+      - name: Install dependencies
+        working-directory: frontends/mobiletests
+        run: npm ci
+
+      - name: Enable KVM group perms
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Install Appium & drivers
+        run: |
+          npm install -g appium@3.1.0
+          appium driver install uiautomator2
+
+      - name: Start Android emulator and run tests
+        uses: reactivecircus/android-emulator-runner@v2
+        env:
+          PLATFORM: Android
+          ANDROID_EMULATOR_WAIT_TIME_BEFORE_KILL: 5
+        with:
+          api-level: ${{ matrix.api-level }}
+          arch: x86_64
+          force-avd-creation: true
+          emulator-boot-timeout: 600
+          working-directory: frontends/mobiletests
+          emulator-options: "-no-window -no-audio -no-boot-anim -no-snapshot-load -no-snapshot-save"
+          script: ./run.sh
   qt-linux:
     runs-on: ubuntu-22.04
     needs: setup-env
--- a/README.md
+++ b/README.md
@@ -126,6 +126,11 @@ Tests are located under [`frontends/web/tests`](/frontends/web/tests) and can be
 `make webe2etest`
 
 More info can be found [here](/frontends/web/tests/README.md)
+Appium is used to perform automatic test on Android emulators.
+The test runs automatically on PRs and push to master in the Github CI. 
+
+`make mobilee2etest` can be used to launch the tests; however, when ran locally, this requires an Android emulator
+to be already running. In CI, the emulator is automatically spawned from a Github Action.
 
 #### Run the HTTP API
 
--- a/frontends/mobiletests/README.md
+++ b/frontends/mobiletests/README.md
@@ -1,0 +1,13 @@
+# Mobile e2e tests
+
+This folder contains the necessary files to setup and run tests against Android (and in the future iOS) devices.
+
+The tests are contained in the [e2e](./e2e) directory and use [Appium](https://github.com/appium/appium) to automate execution of actions on the app.
+
+To run the test manually, one can either run `make mobilee2etest` from the [root repo](../..) or execute [run.sh](./run.sh) directly.
+The script will start an Appium server, wait for it to be ready, and then call `npm run tests` which uses the testing framework [Mocha](https://mochajs.org/).
+
+Note that, in order for the test to start, an Android emulator must be running. The script doesn't start one as steps to do so tend to be somewhat platform-dependant. When running in CI, an emulator is automatically started through a Github Action.
+
+## Development
+To create a new test, simply add a new file called `<test-name>.test.js` to the `e2e` folder.
--- a/frontends/mobiletests/package.json
+++ b/frontends/mobiletests/package.json
@@ -1,0 +1,15 @@
+{
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "mocha e2e/*.test.js"
+  },
+  "description": "",
+  "devDependencies": {
+    "chai": "^6.0.1",
+    "mocha": "^11.7.2",
+    "webdriverio": "^9.19.2",
+    "appium": "^3.0.2",
+    "appium-uiautomator2-driver": "^5.0.3"
+  }
+}
--- a/frontends/mobiletests/run.sh
+++ b/frontends/mobiletests/run.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env sh
+set -e
+
+# --- Run Appium in background ---
+appium --allow-insecure=*:chromedriver_autodownload --port 4723 &
+APPIUM_PID=$!
+trap 'kill $APPIUM_PID || true' EXIT
+
+timeout=20
+elapsed=0
+interval=2
+
+until curl -s http://localhost:4723/status | grep '"ready":true'; do
+  if [ $elapsed -ge $timeout ]; then
+    echo "Appium did not become ready within $timeout seconds."
+    exit 1
+  fi
+  echo "Waiting for Appium..."
+  sleep $interval
+  elapsed=$((elapsed + interval))
+done
+
+# --- Temporarily disable exit on error to capture test status ---
+set +e
+# --- Run tests ---
+npm run test
+TEST_STATUS=$?
+set -e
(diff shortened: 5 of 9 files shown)


frontends/tests/e2e/base.test.js
```diff
@@ -0,0 +1,42 @@
+import { remote } from "webdriverio";
```
thisconnect: nit: other test files have the apache license header
tiny nit: `'` -> `"`

frontends/mobiletests/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "@wdio/local-runner": "^9.19.2",
+    "@wdio/mocha-framework": "^9.19.2",
+    "appium": "^3.0.2",
+    "appium-uiautomator2-driver": "^5.0.3",
+    "appium-xcuitest-driver": "^10.1.2",
+    "chai": "^6.0.1",
+    "mocha": "^11.7.2",
+    "wd": "^1.14.0",
+    "webdriverio": "^9.19.2"
+  }
```
thisconnect, line 14: With @wdio/*, webdriverio, Webdriver/Selenium 2 client, appium, mocha/mocha it's a bit tricky to get what the purpose of each is and how this all works together. I suppose they are all needed.
I hope we never have to migrate these to new major versions 😅

README.md
```diff
@@ -103,6 +103,9 @@ Playwright is used to perform automatic test on some use cases on the webdev ver
 Tests are located under [`frontends/web/tests`](/frontends/web/tests) and can be run with
 `make webe2etest`
 
+Appium is used to perform automatic test on Android emulators.
+The test runs automatically on PRs and push to master in the Github CI.
```
thisconnect, line 130: Maybe mention `make mobilee2etest` here or is it CI only?

frontends/tests/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "tests",
+  "version": "1.0.0",
+  "type": "module",
+  "main": "index.js",
+  "scripts": {
+    "test": "npx mocha e2e/base.test.js"
```
thisconnect: here npx should not be needed.
```suggestion
    "test": "mocha e2e/base.test.js"
```

frontends/tests/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "tests",
+  "version": "1.0.0",
```
thisconnect: name, version is not really needed here.

please add `"private": true,` so we don't accidentally npm-push it and it makes the intention clear.

frontends/tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "type": "module",
+  "main": "index.js",
```
thisconnect: "main" field can be removed.

It would serve as entry file so other project could `import tests from 'tests'`.

frontends/tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "scripts": {
+    "test": "npx mocha e2e/base.test.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
thisconnect: these 3 lines can be removed or or changed to apache

frontends/tests/e2e/base.test.js
```diff
@@ -0,0 +1,42 @@
...
+    const contexts = await driver.getContexts();
+    const webview = contexts.find((c) => c.startsWith("WEBVIEW_"));
+    if (webview) await driver.switchContext(webview);
+  });
+
+  after(async () => {
+    if (driver) await driver.deleteSession();
+  });
+
+  it("App main page loads", async () => {
```
thisconnect: 🎉  https://github.com/BitBoxSwiss/bitbox-wallet-app/actions/runs/18187607581/job/51775701847#step:8:702

.github/workflows/ci.yml
```diff
@@ -70,6 +70,55 @@ jobs:
...
+          npm install -g appium
+          appium driver install uiautomator2
+
+      - name: Start Android emulator and run tests
+        uses: reactivecircus/android-emulator-runner@v2
+        env:
+          PLATFORM: Android
+          ANDROID_EMULATOR_WAIT_TIME_BEFORE_KILL: 5
+        with:
+          api-level: 34
```
benma: Why 34? Should it be a matrix with all supported API levels?

frontends/tests/e2e/base.test.js
benma: Can or should the Android tests reuse Playwright tests? The code running on both platforms is nearly identical after all. Or is the plan to write most tests in Playwright and only stuff that is different on Android here?

What is the overall plan? Maybe worth writitng it down somewhere so people who will write tests know how to proceed.

thisconnect: Playwright and webdriver.io have different api to interact with the browser. So I don't think we can run the same tests.

https://playwright.dev/docs/writing-tests#basic-actions
https://webdriver.io/docs/api/mobile (or do we use https://github.com/admc/wd to interact with the browser?)

frontends/mobiletests/package.json
bznein: Thanks, there were definitely a bunch of leftovers here from experiments, I was able to clean this up quite a bit!

frontends/tests/e2e/base.test.js
bznein: @thisconnect is right. As far as I know there isn't really a way to reuse the tests. I will investigating whether we can at least translate one into the other easily or if we should keep it separated. I think the plan is still a bit in-flight and not set in stone yet. I'll make sure to track it in an issue.

.github/workflows/ci.yml
bznein: Absolutely right! This was more of a "let's get a single setup working first" but it does make sense to add all levels now. I'll do it before asking for  the next review

bznein: It turns out that API <30 don't work out of the box and require a bit more work/investigation. I will open an issue to track this but I think it would make this PR a bit too messy and big so if it's ok for you I'd keep it for another one

frontends/mobiletests/package.json
thisconnect: thank you 🙏

frontends/mobiletests/run.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env sh
+set -e
+
+# --- Run Appium in background ---
+npx appium --allow-insecure=*:chromedriver_autodownload --port 4723 &
+
+# Wait for Appium to be ready
+until curl -s http://localhost:4723/status | grep '"ready":true'; do
```
benma, line 13: should exit after like 5 iterations so it doesn't hang forever if something is wrong

.github/workflows/ci.yml
```diff
@@ -70,6 +70,57 @@ jobs:
...
           name: BitBoxApp-android-${{github.sha}}.apk
           if-no-files-found: error
+
+  e2e-android:
+    name: Android E2E Tests
+    needs: android
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        api-level: [30, 31, 33, 34, 35]
```
benma: 32 missing?

.github/workflows/ci.yml
```diff
@@ -70,6 +70,57 @@ jobs:
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
```
benma: the other entries have `20.x`, don't know what the difference is

frontends/mobiletests/README.md
benma, line 1: Maybe the folder could be named `mobiletests`, to reduce confusion with frontends/web/tests?

frontends/tests/e2e/base.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+    const opts =  {
+      path: '/',
+      port: 4723,
+      capabilities: {
+        platformName: 'Android',
+        'appium:deviceName': 'Android Emulator',
+        'appium:automationName': 'UiAutomator2',
+        'appium:app': './apk/app-debug.apk',
+        'appium:noReset': true,
```
benma: What does this mean? Isn't it better to have a fresh state for each test? Maybe add some comments about what assumptions the tests should make in the README

frontends/mobiletests/run.sh
```diff
@@ -0,0 +1,20 @@
...
+  echo "Waiting for Appium..."
+  sleep 2
+done
+
+# --- Run tests ---
+npm run test
+TEST_STATUS=$?
+echo "Tests finished with status: $TEST_STATUS"
+
+kill -9 $(pgrep -f appium)
```
benma: Why `-9`, that does not give the process a chance to cleanup? 

You can get the PID with `$!` after launching it, see the other comment.

frontends/mobiletests/run.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Wait for Appium to be ready
+until curl -s http://localhost:4723/status | grep '"ready":true'; do
+  echo "Waiting for Appium..."
+  sleep 2
+done
+
+# --- Run tests ---
+npm run test
+TEST_STATUS=$?
```
benma, line 27: If `npm run test` fails, doesn't the script abort due to `set -e` before we even get here?

If you did this so appium is killed on failure too, you could use trap instead:

```sh
set -e
npx appium ... &
APPIUM_PID=$!
trap 'kill -9 $APPIUM_PID' EXIT
...
```

.github/workflows/ci.yml
```diff
@@ -70,6 +70,57 @@ jobs:
...
+
+      - name: Enable KVM group perms
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Install Appium & drivers
+        run: |
+          npm install -g appium
```
benma: In run.sh you use `npx`, so it's a different appium installation (locally managed by npx) that ends up being used. Unless I misunderstand, either this global installation is not needed, or the use of `npx` is not needed.

Also, should/can the version of appium be fixed, or is it preferable if it just always uses the latest?

frontends/mobiletests/run.sh
bznein: (GH Actions have a timeout anyway, but yeah, I agree)

.github/workflows/ci.yml
bznein: I was thinking of only targeting API that were introduced with new Android versions, so I skipped 32 (which is part of Android 12 like level 31) but I changed it now

.github/workflows/ci.yml
bznein: me neither, but I will stick to 20.x for consistency!

frontends/tests/e2e/base.test.js
bznein: YEah you are probably right, I made many trial and error and I could have some leftover code that can be cleaned. I'll try without this flag and update the README if needed

.github/workflows/ci.yml
bznein: YEah I think both points make sense, I changed from `npx` to just `appium` and pinned to the current latest version

frontends/mobiletests/run.sh
benma: Is this true?

> If npm run test fails, doesn't the script abort due to set -e before we even get here?

bznein: Sorry that's weird, I could swear I pushed a change to this line that addresses it, and you are right, no clue what I have done instead... I will push a change asap

PR #3589, opened 2025-10-02
frontend: fix country-selector default

thisconnect:
react-select requires object identity matching, not just value equality. (according to GPT)

to test this in webdev it also need https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3587 
(and remove appfolder.dev/config.json)


thisconnect:
rebased

PR #3600, opened 2025-10-07
Frontend fix factory reset

thisconnect:
Fixed an issue with FactoryResetSetting where continuing to reset
the device should show a WaitDialog. But the simplified version in
a recent Dialog refactoring always called onClose as if the user
would aboard the workflow.

Regression introduced in:
- https://github.com/thisconnect/bitbox-wallet-app/commit/d5bb07888deac147a74b527c0251529f6ef0174f


thisconnect:
updated commit message, fixed setTimeout type, and cleanup DialogButtonsProps type

PR #3602, opened 2025-10-07
backend: change block explorer to mempool.space

strmci:
mempool.space is considered one of the best Bitcoin explorers and the preferred choice of many. The app already uses mempool.space for fee estimations.


Beerosagos (approved):
tACK 🙏

PR #3606, opened 2025-10-09
tests: add watch-only tests 1 and 2 to playwright.

bznein:
These are the watch-only tests that we usually run manually. However the ones that require the use of the passphrase are not implemented yet as they need changes to the simulator. 

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/tests/base.test.ts
+++ b/frontends/web/tests/base.test.ts
@@ -22,19 +22,19 @@ let servewallet: ServeWallet;
 
 test('App main page loads', async ({ page, host, frontendPort, servewalletPort }) => {
 
-    await test.step('Start servewallet', async () => {
-        servewallet = new ServeWallet(page, servewalletPort, frontendPort, host);
-        await servewallet.start();
-    });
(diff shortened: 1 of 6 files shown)


frontends/web/tests/base.test.ts
```diff
@@ -32,3 +34,7 @@ test('App main page loads', async ({ page, host, frontendPort, servewalletPort }
         await expect(body).toContainText('Please connect your BitBox and tap the side to continue.');
     });
 });
+
+test.afterAll(() => {
+    servewallet.stop();
+});
```
thisconnect, line 40: noticed we usually use 2 spaces, could you enable the linter to run on tests/ files?

I'm not 100% sure why it doesn't run, maybe due to https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/91c86119b2913a1b7b2e122e55e50800ef582551/frontends/web/vite.config.mjs#L36 ?

bznein: Done, thanks! I added a commit that addresses this

PR #3607, opened 2025-10-09
frontend: prefer 'type' over 'interface' for safer type definitions

thisconnect:
Type can represent unions, intersections, mapped types, tuples, and other
advanced types that interface cannot.

Interfaces with the same name are automatically merged in TypeScript,
which can lead to subtle bugs if two interfaces collide unknowingly.
Type aliases don’t merge, so reusing the same name will throw an error,
which is safer.

Retain 'interface' only when extending global types (e.g., Window) via
`declare global`.

Changed `interface IAccount` to `type TAccount` with T prefix for
consistency and to prevent name clashes with existing modules.

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -368,16 +368,6 @@ export const sendTx = (
 
 export type FeeTargetCode = 'custom' | 'low' | 'economy' | 'normal' | 'high' | 'mHour' | 'mHalfHour' | 'mFastest';
 
-export type TProposeTxData = {
-  address?: string;
-  amount?: number;
-  // data?: string;
-  feePerByte: string;
-  feeTarget: FeeTargetCode;
-  selectedUTXOs: string[];
-  sendAll: 'yes' | 'no';
-}
-
 export type TFeeTarget = {
   code: FeeTargetCode;
   feeRateInfo: string;


frontends/web/src/api/account.ts
```diff
@@ -368,33 +368,27 @@ export const sendTx = (
...
-    feePerByte: string;
-    feeTarget: FeeTargetCode;
-    selectedUTXOs: string[];
-    sendAll: 'yes' | 'no';
-}
-
-export interface IProposeTx {
-    aborted?: boolean;
-    success?: boolean;
-    errorMessage?: string;
```
thisconnect, line 384: IProposeTx was not used

frontends/web/src/globals.d.ts
```diff
@@ -14,28 +14,28 @@
...
 
-import { TPayload } from './utils/transport-common';
+import type { TPayload } from './utils/transport-common';
 
 export declare global {
-    interface Window {
-        qt?: { webChannelTransport: unknown; };
-        android?: {
-            call: (queryID: number, query: string) => void;
+  interface Window {
```
thisconnect, line 20: here we want to keep interface and extend global Window object

frontends/web/src/api/account.ts
```diff
@@ -368,33 +368,27 @@ export const sendTx = (
...
+export type TProposeTxData = {
+  address?: string;
+  amount?: number;
+  // data?: string;
```
bznein: Can we get rid of this as well?

frontends/web/src/api/backend.ts
```diff
@@ -19,36 +19,37 @@ import type { FailResponse, SuccessResponse } from './response';
...
+  name: string;
+  canAddAccount: boolean;
+  suggestedAccountName: string;
 }
 
-export interface ISuccess {
-    success: boolean;
-    errorMessage?: string;
-    errorCode?: string;
+// subtle difference FailResponse return message?: string; code?: number
```
bznein: This comment isn't really clear to me... can we make it more self-explanatory?

frontends/web/src/api/coins.ts
```diff
@@ -14,17 +14,17 @@
...
 import { apiPost, apiGet } from '@/utils/request';
 
 export type BtcUnit = 'default' | 'sat';
 
 export type TStatus = {
-    targetHeight: number;
-    tip: number;
-    tipAtInitTime: number;
-    tipHashHex: string;
+  targetHeight: number;
```
bznein, line 24: For my understanding? What triggered all of this whitespace changes? I'm fine with them I just want to make sure we all have the same auto-formatting active so that we don't get a bunch of unwanted changes each time :)

frontends/web/src/api/account.ts
thisconnect: actually we can completely remove this type, it seems to be a leftover and not used anymore. tx-propsal endpoint only uses TTxInput. 

I'll remove in an additional commit.

frontends/web/src/api/coins.ts
thisconnect: I formatted the ones I saw under api manually. 

For your other question, it seems we might be able to add automatic formatgin or at least linting with `@stylistic/eslint-plugin-ts`. We already use `@stylistic/eslint-plugin`.

I'll try later and open a new PR if I succeed with stylisitc

frontends/web/src/api/backend.ts
thisconnect: will fixup into "frontend: change interface to type"

frontends/web/src/api/account.ts
thisconnect: added one more commit to remove

frontends/web/src/api/backend.ts
thisconnect: squashed and force pushed

bznein: Much better now, thanks!

PR #3618, opened 2025-10-15
maintenance: only check if an account is used if not already set.

bznein:
When an account is set as used, there is no more need to check it again. By exiting early if an account is already used, we avoid a call to Transactions().

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -1722,20 +1722,23 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 			return
 		}
 	}
-	log := backend.log.WithField("accountCode", account.Config().Config.Code)
-	txs, err := account.Transactions()
-	if err != nil {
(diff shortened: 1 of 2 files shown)


backend/accounts.go
```diff
@@ -1722,6 +1722,12 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 			return
 		}
 	}
+
+	// If the account is already marked as used, nothing to do.
+	if account.Config().Config.Used {
+		return
```
benma: feels safer to continue with the code below, i.e. set HiddenBecauseUnused = true, and make the `maybeAddHiddenUnusedAccounts()` and emit event calls.

haven't checked all implications exactly, but by returning here the behavior might change because it is not doing the same thing anymore.

PR #3621, opened 2025-10-16
feat: allow Qt app to talk to the simulator.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ In order to do so:
 
 If the simulator is listening on a custom port, `--simulatorPort=<port>` must also be provided.
 
+Note: the simulator is currently only supported in the servewallet and in the Qt app and only when the app runs in testnet mode.
+
 #### Watch and build the UI
 
 Run `make webdev` to develop the UI inside a web browser (for quick development, automatic rebuilds
--- a/backend/bridgecommon/bridgecommon.go
+++ b/backend/bridgecommon/bridgecommon.go
@@ -286,6 +286,7 @@ func (env *BackendEnvironment) BluetoothConnect(identifier string) {
 // Serve serves the BitBox API for use in a native client.
 func Serve(
 	testnet bool,
+	simulator bool,
 	gapLimits *btctypes.GapLimits,
 	communication NativeCommunication,
 	backendEnvironment backend.Environment) {
@@ -303,6 +304,11 @@ func Serve(
 		WithField("goarch", runtime.GOARCH).
 		WithField("version", backend.Version).
 		Info("environment")
+
+	if simulator {
+		log.Info("Simulator mode enabled, ensuring testnet mode is enabled")
+		testnet = true
+	}
 
 	var err error
 	globalBackend, err = backend.NewBackend(
--- a/frontends/qt/server/server.go
+++ b/frontends/qt/server/server.go
@@ -68,6 +68,7 @@ import (
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/bridgecommon"
 	btctypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/types"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/devices/bitbox02/simulator"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/devices/usb"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/mobileserver"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/logging"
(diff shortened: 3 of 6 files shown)


frontends/qt/server/server.go
```diff
@@ -146,6 +149,15 @@ func serve(
 		}
 	}
 
+	deviceInfos := usb.DeviceInfos
+	if *useSimulator {
+		if !*testnet {
```
benma: This is not the only way to enable testnet, we now have the testnet app setting to restart in testnet via the config, so this is the wrong place to do this check.

bznein: Thanks good point. I moved the check down into the `bridgecommon.Serve` method, by adding a `simulator` arg and checking `backend.Testing()` there (as in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3622#discussion_r2436101365)

README.md
```diff
@@ -103,6 +103,8 @@ In order to do so:
 
 If the simulator is listening on a custom port, `--simulatorPort=<port>` must also be provided.
 
+Note: the simulator is currently only supported in the servewallet and in the Qt app.
```
benma: "... in testnet"

backend/bridgecommon/bridgecommon.go
```diff
@@ -318,6 +319,10 @@ func Serve(
 		log.WithError(err).Fatal("Failed to create backend")
 	}
 
+	if simulator && !globalBackend.Testing() {
+		log.WithError(errp.New("Simulator can only be used in testnet mode")).Fatal()
```
benma: No need to change, but fyi in case you didn't know about Panic: also could do `log.Panic("Simulator can only be used in testnet mode")`

backend/devices/bitbox02/simulator/simulator.go
```diff
@@ -67,6 +68,16 @@ func isRunning(port int) bool {
 	return true
 }
 
+// DeviceInfos returns the list of connected USB devices. If the simulator is running, it returns only
+// the simulator device info.
+// Otherwise, it returns the actual connected USB devices.
+func DeviceInfos() []usb.DeviceInfo {
```
benma: I would undo this and put it back into servewallet/main.go. It's only used there, and to me it makes more sense that the caller makes the distinction, and not the bitbox02/simulator package, similar like you do it in qt's server.go

backend/bridgecommon/bridgecommon.go
bznein: Thanks, I changed it as I think looks nicer

backend/devices/bitbox02/simulator/simulator.go
bznein: Done. The "ugly" part of this, is that there is duplication of code, as now servewallet/main's `DeviceInfos` is basically duplicated in `server.go` but I think we can live with it.

frontends/qt/server/server.go
```diff
@@ -109,6 +110,14 @@ func setOnline(isReachable bool) {
 	bridgecommon.SetOnline(isReachable)
 }
 
+func deviceInfos() []usb.DeviceInfo {
+	testDeviceInfo := simulator.TestDeviceInfo()
+	if testDeviceInfo != nil {
+		return []usb.DeviceInfo{*testDeviceInfo}
```
benma: why dereference? can just keep `testDeviceInfo` by ref right? Same in servewallet/main.go

backend/bridgecommon/bridgecommon.go
```diff
@@ -318,6 +319,10 @@ func Serve(
 		log.WithError(err).Fatal("Failed to create backend")
 	}
 
+	if simulator && !globalBackend.Testing() {
+		log.Panic("Simulator can only be used in testnet mode")
```
benma: This is not visible on the command line. Maybe also write to stderr so users can see what went wrong.

Thinking some more about this, maybe it's most ergonomical if `-simulator` just implied `-testnet`? Then you could remove the simulator flag here and this check here and just pass testnet=true if simulator is set.

Wdyt?

frontends/qt/server/server.go
```diff
@@ -121,6 +130,8 @@ func serve(
 	log := logging.Get().WithGroup("server")
 	log.WithField("args", os.Args).Info("Started Qt application")
 	testnet := flag.Bool("testnet", false, "activate testnets")
+	simulatorPort := flag.Int("simulatorPort", 15423, "port for the BitBox02 simulator")
+	useSimulator := flag.Bool("simulator", false, "use the BitBox02 simulator")
```
benma: mention that it implies testnet (if you make this change), or mention that testnet is required if you prefer to keep the testnet check

PR #3622, opened 2025-10-16
chore: disallow simulator together with mainnet in servewallet.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/cmd/servewallet/main.go
+++ b/cmd/servewallet/main.go
@@ -205,6 +205,10 @@ func main() {
 	fmt.Printf("Listening on: http://localhost:%d\n", port)
 
 	if *useSimulator {
+		// The simulator is only allowed when running in testnet mode.
+		if !backend.Testing() {
+			log.Panic("The BitBox02 simulator can only be used in testnet mode.")
+		}
 		simulator.Init(*simulatorPort)
 	}
 


cmd/servewallet/main.go
```diff
@@ -199,6 +199,10 @@ func main() {
 	fmt.Printf("Listening on: http://localhost:%d\n", port)
 
 	if *useSimulator {
+		// The simulator is only allowed when running in testnet mode.
+		if *mainnet {
```
benma: Probably the same problem as https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3621#discussion_r2435972289?

bznein: Thanks, I think in this case we can simply use `backend.Testing()` instead, since the `backend` is already constructed and `Testing()` takes care of the tesnet config option

cmd/servewallet/main.go
```diff
@@ -199,6 +199,10 @@ func main() {
 	fmt.Printf("Listening on: http://localhost:%d\n", port)
 
 	if *useSimulator {
+		// The simulator is only allowed when running in testnet mode.
+		if !backend.Testing() {
+			log.WithError(fmt.Errorf("The BitBox02 simulator can only be used in testnet mode.")).Fatal()
```
benma: log.Panic(...)

PR #3623, opened 2025-10-16
Frontend responsive fontsize android

thisconnect:
Updated the frontend to be responsive on Android. Ideally this PR would have no visual difference and just scales up if settings changed. But some CSS was not possible to make reponsive and stay 100% the same, that is for example the Guide where we used "em" sizes. The balance header that is used in different view such as account-overview and send now looks the same on both view before padding/spacing was different. Also we used many different font-sizes, some not intentional. This PR refactored font-size CSS variables and slightly reduced the number of different font-sizes, causing minor visual layout changes.

- the font-size and UI scales up if Android system settings "Font size" is increased
- Amounts that are too large to display should shorten i.e 0.123... BTC (also without increased font-size)
- shorten amounts also improve the all-accounts mobile view, especially with large amounts like for eth
- Balance component should now visually look the same on different views
- in scaled up mode some layout that before was left/right now can break onto 2 different lines if there is not enough space

The scaled up UI on Android looks currently so broken that any change is an improvement. But each change could also introduce a visual CSS regression. So it would help if the focus of the testing is that nothing broken in "normal" mode.

- [x] updating the Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Sort backups from newest to oldest in manage backups
 - Android: fix display of external links from Bitrefill
 - fix language sometimes not persistent across app restarts
+- Android: make the UI work with responsive font sizes and adhere to OS font size settings 
 
 ## v4.49.0
 - Bundle BitBox02 Nova firmware version v9.24.0


thisconnect:
Will eventually replace https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2699

thisconnect:
added some fixes (not net squashed) and rebased on top of current master

sutterseba (approved):
tested on iOS

PR #3625, opened 2025-10-16
backend/frontend: gate send-to-self dropdown on firmware support

benma:


backend/handlers/handlers.go
```diff
@@ -597,6 +598,55 @@ func (handlers *Handlers) getKeystores(*http.Request) interface{} {
...
+			Success:      false,
+			ErrorMessage: err.Error(),
+		}
+	}
+
+	if !bytes.Equal(rootFingerprint, connectedRootFingerprint) {
+		handlers.log.WithField("requested", rootFingerprintHex).Warn("features requested for non-connected keystore")
+		return response{
+			Success:      false,
+			ErrorMessage: "keystore not connected",
```
bznein: [nit] maybe something clearer that shows the difference between this case and the one where `connectedKeystore` is `nil`?

frontends/web/src/routes/account/send/send.tsx
```diff
@@ -135,18 +135,28 @@ export const Send = ({
...
+      const featuresResult = await getKeystoreFeatures(rootFingerprint);
+      if (!featuresResult.success) {
+        alertUser(featuresResult.errorMessage || t('genericError'));
+        return;
+      }
+      if (!featuresResult.features?.supportsSendToSelf) {
+        alertUser(t('device.firmwareUpgradeRequired'));
+        return;
+      }
+    }
```
thisconnect: in Pocket we use this

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/91c86119b2913a1b7b2e122e55e50800ef582551/frontends/web/src/routes/market/pocket.tsx#L385-L405

bznein (approved):
backend utACK with a minor comment (and I think you probably need to regenerate the mock file)

frontends/web/src/routes/account/send/send.tsx
benma: Nice. Will ask my agent to extract this into a resuable component later :grin:

benma: @thisconnect done, PTAL.

backend/handlers/handlers.go
benma: Done

frontends/web/src/routes/account/send/send.tsx
thisconnect: Tested with 9.18.0 and tried to find why it didn't show. But reading the code the FW upgrade dialog should only shows when handleSend is called. Could you change to it shows earlier?

I suggest either when the user clicks the dropdown or after choosing an option from the dropdown.

shonsirsha: Agree - I think it'd be nice to show it earlier, before the user presses the send btn

benma: Done now. I accidentally amended it into the previous commit, and it was too difficult to undo it, sorry!

~~This PR is now also building on top of https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3653~~ (rebased)

PTAL

frontends/web/src/app.tsx
```diff
@@ -216,7 +216,6 @@ export const App = () => {
             <AppRouter
               accounts={accounts}
               activeAccounts={activeAccounts}
-              deviceIDs={deviceIDs}
```
thisconnect, line 219: 😅

PR #3627, opened 2025-10-20
centralize version in VERSION file

benma:
Bumping version so far meant updating many files. This centralizes the version bumping to a single file.

In this first commit, we use the file to replace the hardcoding of the version in update.go and for the Linux packages.

The VERSION file follows the INI format, similar to TOML.

In later commits, the VERSION file will also get keys for the Android version code and the iOS build number. get_version.py will then probably be extended to be able to print the values of these keys as well.

Altneratives considered:
- use TOML: tommlib in Python is only available since v3.11, which is not in our Docker image yet, and the default Python version in the GitHub Windows runner now is still at 3.9.
- use JSON: uglier and harder to edit, more strict in formatting, can't add comments
- use `go:embed` instead of ldflags to embed the file, but go:embed can't embed files in parent directories. Solutions to this didn't look nice (symlink, or make root folder a Go package, or move the VERSION file to a Go package).


NickeZ:
IMO, please name the file `VERSION.toml`.

backend/versioninfo/versioninfo.go
```diff
@@ -0,0 +1,34 @@
...
+
+package versioninfo
+
+import (
+	"github.com/BitBoxSwiss/bitbox02-api-go/util/semver"
+)
+
+var (
+	// versionString is overridden at build time via go ldflags.
+	versionString = "1.0.0"
```
NickeZ: Why is 1.0.0 a good default? SHouldn't something that is obivously a non-release like "dev" or "0.0.1" be the default?

benma: 0.0.1 sounds good.

Didn't want a different string as it can't be parsed into the SemVer thing below.

The default will basically almost never be seen, all dev tools and prod/test builds get the right version. And if one sees it it's no big deal anyway I guess.

benma:
> > > IMO, please name the file `VERSION.toml`.
> > 
> > 
> > But it's not a TOML file, it's an INI file. `VERSION.ini`?
> > Edit: a concrete difference is that TOML requiures `"` string delimiters, while INI is the opposite (any string delimiters will be included in the final value as part of the string).
> 
> INI isn't well defined, but yeah configparser will give you the string raw. Can't you just parse out the string?

I could but is it worth it? Seems like bit like a hack. Why not just keep it as it is?

benma:
> and just put the version there unstructured and then just read the file. No need to parse anything.

I mentioned the reason in the commit msg: in future commits I'll add the iOS build number and Android version code, and I prefer to have all version related stuff in one file.

> OK, if you think INI is better

I don't think it's a better format, but TOML parsing is not readily available and pretending INI is TOML does not seem right :smile:

NickeZ:
You also can't reuse a build number, so it isn't great to put that in a file. With continuous delivery every merge to master gets pushed to the stores. So then you have to bump the file in every commit AFAIK.

benma:
@NickeZ in that case I go back to a plain text file, not ini/toml. See added commit (will squash in the end).

PR #3628, opened 2025-10-20
qt/windows: don't link libssp

benma:
https://www.msys2.org/news/?utm_source=chatgpt.com#2022-10-10-libssp-is-no-longer-required

> Building with _FORTIFY_SOURCE no longer requires explicitly linking
with libssp (-lssp) and enabling stack protection no longer pulls in libssp. This brings things in line with other platforms. Thanks to Martin Storsjö for implementing this in mingw-w64. Once all our affected packages are rebuilt we will remove the libssp package from our repo.

It seems maybe a recent GitHub Windows-2022 runner image update bumped the toolchain to one that removed libssp, as copying libssp failed with `cp: cannot stat 'C:/mingw64/bin/libssp-0.dll': No such file or directory`.

### DIFF
--- a/frontends/qt/BitBox.pro
+++ b/frontends/qt/BitBox.pro
@@ -32,7 +32,6 @@ include(external/singleapplication/singleapplication.pri)
 DEFINES += QAPPLICATION_CLASS=QApplication
 
 win32 {
-    # -llibssp would be nice to have on Windows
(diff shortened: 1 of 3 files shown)


benma:
Tested on my Windows, seems to work.

PR #3631, opened 2025-10-21
macOS: use version from APP_VERSION file

benma:
Centralize version bumping.


NickeZ:
I'm curious why you went with xcconfig in iOS but plist in macos? Both should support both ways right?

PR #3635, opened 2025-10-21
docker: increase version and fix image build bug

Beerosagos:
Recent commmit 72b248a4c9fd35cc297c77c6d56cbd14171c643e introduced a novel version.mk.inc file included by the Makefile. This broke the docker image building, because the file wasn't mounted before calling make.

This commit fixes the bug and increases the docker image version.


Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN --mount=target=/mnt,source=scripts DEBIAN_FRONTEND=noninteractive /mnt/docke
 ARG SYS_GOPATH=/opt/go
 ENV PATH=${SYS_GOPATH}/bin:/usr/local/go/bin:$PATH
 
-RUN --mount=target=/mnt/Makefile,source=Makefile GOPATH=${SYS_GOPATH} make -C /mnt envinit
+RUN --mount=target=/mnt/Makefile,source=Makefile --mount=target=/mnt/version.mk.inc,source=version.mk.inc GOPATH=${SYS_GOPATH} make -C /mnt envinit
```
benma, line 27: isn't APP_VERSION also needed then, otherwise `APP_VERSION := $(strip $(shell cat "$(VERSION_MK_ROOT)APP_VERSION"))` fails?

Thanks for making the image, I asked you for it so we can fix CI in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3626, but it won't work with this image as it is not based on the release-v4.49.0 branch, where version.mk.inc does not exist yet.

We need the update anyway so future locize pulls don't break CI, but I guess we will just do without the locize CI check in the release branch, it's not critical.

Beerosagos: > isn't APP_VERSION also needed then, otherwise APP_VERSION := $(strip $(shell cat "$(VERSION_MK_ROOT)APP_VERSION")) fails?

Looking at the code, I'd say you're right. But the build doesn't fail and I tried building a test android version: the app versions is displayed as expected. Not sure why :thinking:

Beerosagos:
Added a commit to fix the locize error

PR #3636, opened 2025-10-21
fix preposition in bluetooth's alertDisabled key

strmci:


frontends/web/src/locales/en/app.json
```diff
@@ -199,7 +199,7 @@
   },
   "bitbox02Settings": {
     "bluetoothToggleEnabled": {
-      "alertDisabled": "Bluetooth is now disabled on your BitBox. If you want to connect with iPhones on the future, you will need to re-enable it.",
+      "alertDisabled": "Bluetooth is now disabled on your BitBox. If you want to connect with iPhones in the future, you will need to re-enable it.",
```
benma, line 202: this has to be done on locize , which will then be pulled in with the next locize pull / PR. See https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/docs/i18n.md

strmci: updated in locize, closing this PR

PR #3637, opened 2025-10-21
frontend: fix supress esc key when confirming an address

thisconnect:
Pressing the ESC key during address verification reloads the page with the QR code dialog.

ESC or clicking on the dialog background, should be disabled (as in previous versions) when confirming the address.

### DIFF
--- a/frontends/web/src/components/dialog/dialog.tsx
+++ b/frontends/web/src/components/dialog/dialog.tsx
@@ -93,7 +93,7 @@ export const Dialog = ({
 
   // ESC closes dialog (fires onClose)
   useEsc(() => {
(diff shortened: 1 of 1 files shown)


frontends/web/src/components/dialog/dialog.tsx
```diff
@@ -117,8 +119,10 @@ export const Dialog = ({
 
   // Close button handler
   const handleCloseClick = useCallback(() => {
-    deactivate(true);
-  }, [deactivate]);
+    if (open && onClose) {
```
thisconnect, line 122: this check is not really necessary but I added it so the intention is clear.

PR #3638, opened 2025-10-22
frontend: fix appearence of upgrade dot (red dot) on ios & android

shonsirsha:
This commit fixes 2 issues:
1. First commit, UI issue where red dot appears not at the correct place in the bottom nav:

Before:
Screenshot 2025-10-22 at 07 10 00

After:
Screenshot 2025-10-22 at 07 15 30

2. 2nd commit: on iOS, the red dot did not appear on the bottom nav, as well as the more screen. This is fixed by putting `isBitBox02` in the dep. list.

### DIFF
--- a/frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
+++ b/frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
@@ -34,7 +34,7 @@ export const BottomNavigation = ({ activeAccounts, devices }: Props) => {
   const { pathname } = useLocation();
   const deviceID = Object.keys(devices)[0];
   const isBitBox02 = deviceID && devices[deviceID] === 'bitbox02';
-  const versionInfo = useLoad(isBitBox02 ? () => getVersion(deviceID) : null, [deviceID]);
+  const versionInfo = useLoad(isBitBox02 ? () => getVersion(deviceID) : null, [deviceID, isBitBox02]);
   const canUpgrade = versionInfo ? versionInfo.canUpgrade : false;
 
   const onlyHasOneAccount = activeAccounts.length === 1;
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/bottom-navigation/bottom-navigation.module.css
```diff
@@ -1,4 +1,5 @@
 .container {
+  align-items: center;
```
thisconnect, line 2: I was only aware of the icon alginment bug. But the commit message mentions another issue.

frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
```diff
@@ -34,7 +34,7 @@ export const BottomNavigation = ({ activeAccounts, devices }: Props) => {
   const { pathname } = useLocation();
   const deviceID = Object.keys(devices)[0];
   const isBitBox02 = deviceID && devices[deviceID] === 'bitbox02';
-  const versionInfo = useLoad(isBitBox02 ? () => getVersion(deviceID) : null, [deviceID]);
+  const versionInfo = useLoad(isBitBox02 ? () => getVersion(deviceID) : null, [deviceID, isBitBox02]);
```
thisconnect, line 37: Did the red dot not appear? Is this technically needed?

`isBitBox02` as well as `deviceID` are derived from devices, so  `deviceID`  would change anyway when a new devices is registered or unplugged. Or what am I missing?

Seems this should be a 2nd commit in addition to the alignment fix.

thisconnect:
maybe @sutterseba want to test this?

sutterseba (approved):
tested on iPhone

PR #3640, opened 2025-10-22
frontend: fix focus in dialog with multiple input fields

thisconnect:
When entering a number in the "Gap limit for change addresses" field within the dialog, the cursor jumps to the "Gap limit for receive addresses" field after typing a digit.

The cursor should remain in the Gap limit for change address field while typing.

### DIFF
--- a/frontends/web/src/hooks/keyboard.ts
+++ b/frontends/web/src/hooks/keyboard.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useRef } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 
 /**
  * gets fired on each keydown and executes the provided callback.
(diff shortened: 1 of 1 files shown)


frontends/web/src/hooks/keyboard.ts
```diff
@@ -71,59 +71,132 @@ export const useFocusTrap = (
   ref: React.RefObject<HTMLElement>,
   active: boolean,
 ) => {
+  const autofocusDelay = 50;
```
thisconnect, line 74: As the dialog changes from opening->open within 20ms, this 50ms wait time should be plenty enough.

frontends/web/src/hooks/keyboard.ts
```diff
@@ -71,59 +71,132 @@ export const useFocusTrap = (
...
+    (e: KeyboardEvent) => {
+      if (!trapEnabled.current || !ref.current || e.key !== 'Tab') {
+        return;
+      }
 
-    const first = focusables[0];
-    const last = focusables[focusables.length - 1];
+      const focusables = Array.from(
+        ref.current.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR)
+      ).filter((el) => el.offsetParent !== null); // skip hidden
```
thisconnect, line 89: filter was added by gpt and removes all elements that are hidden somehow

frontends/web/src/hooks/keyboard.ts
```diff
@@ -71,59 +71,132 @@ export const useFocusTrap = (
...
+      if (focusables.length === 0) {
+        return;
+      }
 
-  useKeydown(handleKeyDown);
+      const first = focusables[0];
+      const last = focusables[focusables.length - 1];
+      const current = document.activeElement as HTMLElement;
+
+      if (e.shiftKey && current && current === first) {
```
thisconnect, line 99: I added `&& current` as here both `first` and `current` could be `undefined`. And we don't want to continue in case of `undefined === undefined`.

PR #3641, opened 2025-10-23
tests: add GapLimits e2e test.

bznein:
Tested by running the test on `master` and then by cherry-picking the commit at https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3640/commits. The first one fails, the other one succeeds

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/tests/gapSettings.test.ts
+++ b/frontends/web/tests/gapSettings.test.ts
@@ -1,0 +1,89 @@
+/**
+*  Copyright 2025 Shift Crypto AG
+*
+*  Licensed under the Apache License, Version 2.0 (the "License");
+*  you may not use this file except in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*       http://www.apache.org/licenses/LICENSE-2.0
+*
+*  Unless required by applicable law or agreed to in writing, software
+*  distributed under the License is distributed on an "AS IS" BASIS,
+*  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+*  See the License for the specific language governing permissions and
+*  limitations under the License.
+*/
+
+
+
+import { test } from './helpers/fixtures';
+import { ServeWallet } from './helpers/servewallet';
+import { expect } from '@playwright/test';
+import { clickButtonWithText } from './helpers/dom';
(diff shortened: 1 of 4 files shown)


frontends/web/tests/gapSettings.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    await expect(body).toContainText('Please connect your BitBox and tap the side to continue.');
+  });
+
+  await test.step('Type into gap limit inputs', async () => {
+    await page.getByRole('link', { name: 'Settings' }).click();
+    await page.getByRole('link', { name: 'Advanced settings' }).click();
+    await page.getByRole('button', { name: 'Custom gap limit settings' }).click();
+
+    const recvInput = page.locator('#gapLimitReceive');
+    await recvInput.clear();
```
thisconnect, line 47: The Dialog component will focus the first focusable element inside the dialog. Would it make sense to check focus here and also later after the change gap was changed?


https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-focused

frontends/web/tests/gapSettings.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    await clickButtonWithText(page, 'Confirm');
+  });
+
+  await test.step('Verify gap limit inputs are correctly saved', async () => {
+    await page.getByRole('button', { name: 'Custom gap limit settings' }).click();
+    const recvInput = page.locator('#gapLimitReceive');
+    await expect(recvInput).toHaveValue('50');
+    const changeInput = page.locator('#gapLimitChange');
+    await expect(changeInput).toHaveValue('60');
```
thisconnect: A bug we had in past Dialog refactoring was that after editing the 2nd field, the focus jumped to the first input element. Could you test that the 2nd input field here still has focus?

https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-focused

bznein: Sure! I kinda did this indirectly by having playwright "type" as in a keyboard and then check the content after, but I think asserting for the focus is a good idea too.

PR #3643, opened 2025-10-23
Release v4.49.0 -> master

benma:
So we can locize push new keys.


thisconnect:
@benma there was one more commit that added a new translation in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/3643/commits/f7df9c0b53081dd68597f3e2018f09c94b36874c

PR #3916, opened 2026-03-04
frontend: move web config access into config context/provider

smokyisthatyou:
Add ConfigContext + ConfigProvider in frontends/web and refactor React components/hooks to consume getConfig/setConfig through useConfig() instead of importing utils/config directly. Keep non-React config consumers on the shared utils/config implementation path, and update related hook dependency arrays.

- [ ] updating the Changelog
- [ ] writing unit tests

cc @thisconnect

### DIFF
--- a/frontends/web/src/api/config.ts
+++ b/frontends/web/src/api/config.ts
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { apiGet, apiPost } from '@/utils/request';
+import type { Fiat } from '@/api/account';
+import type { BtcUnit } from '@/api/coins';
+
+type TElectrumServerInfo = Readonly<{
+  server: string;
+  tls: boolean;
+  pemCert: string;
+}>;
+
+type TBtcCoinConfig = Readonly<{
+  electrumServers: TElectrumServerInfo[];
+}>;
+
+/** BTC-based coin keys in backend config (see backend/config/config.go). */
+export type TConfigBackendBtcCoinKey = 'btc' | 'tbtc' | 'ltc' | 'tltc';
+
+// Mirrors backend/config/config.go ethCoinConfig: DeprecatedActiveERC20Tokens (JSON key
+// "activeERC20Tokens"). Deprecated — ERC20 activation is per-account in accounts config; kept
+// for migration / compatibility with persisted app config.
+type TEthCoinConfig = Readonly<{
+  activeERC20Tokens: string[];
+}>;
+
+export type TConfigBackendProxy = Readonly<{
+  useProxy: boolean;
+  proxyAddress: string;
+}>;
+
+/** Keys used by NewBadge to mark UI elements as seen. */
+export type TConfigFrontendBadgeKey =
+  | 'hasSeenMarketplaceNudge'
+  | 'hasSeenSwapMarketTab'
+  | 'hasSeenOtcMarketTab';
+
+/** Dynamic frontend keys written when dismissing Status banners. */
+type TConfigFrontendDismissibleDynamicKey =
+  | `update-${string}`
+  | `banner-backup-${string}`
+  | `banner-${string}-${string}`;
+
+/** Known static frontend keys written when dismissing Status banners. */
+type TConfigFrontendDismissibleKnownKey =
+  | 'walletConnectDisclaimerDismissed'
+  | 'skipTestingWarning'
+  | 'mobile-data-warning';
+
+export type TConfigFrontendDismissibleKey =
+  | TConfigFrontendDismissibleKnownKey
+  | TConfigFrontendDismissibleDynamicKey;
+
+export type TConfigFrontend = Readonly<{
+  guideShown?: boolean;
+  hideAmounts?: boolean;
+  darkmode?: boolean;
+  expertFee?: boolean;
+  coinControl?: boolean;
+  selectedExchangeRegion?: string;
+  hideEnableRememberWalletDialog?: boolean;
+  hasUsedWalletConnect?: boolean;
+  bitsuranceNotifyCancellation?: string[];
+
+  hasSeenMarketplaceNudge?: boolean;
+  hasSeenSwapMarketTab?: boolean;
+  hasSeenOtcMarketTab?: boolean;
+
+  skipBitrefillWidgetDisclaimer?: boolean;
+  skipBTCDirectWidgetDisclaimer?: boolean;
+  skipBTCDirectOTCDisclaimer?: boolean;
+  skipMoonpayDisclaimer?: boolean;
+  skipPocketDisclaimer?: boolean;
+  skipPocketOTCDisclaimer?: boolean;
+  skipBitsuranceDisclaimer?: boolean;
+  skipSwapkitDisclaimer?: boolean;
+
+  walletConnectDisclaimerDismissed?: boolean;
+  skipTestingWarning?: boolean;
+  'mobile-data-warning'?: boolean;
+}> & Readonly<{
+  [key in TConfigFrontendDismissibleDynamicKey]?: boolean;
+}>;
+
+export type TConfigBackend = Readonly<{
+  proxy: TConfigBackendProxy;
+  /**
+   * Deprecated global coin activation flags (backend/config/config.go: DeprecatedBitcoinActive,
+   * DeprecatedLitecoinActive, DeprecatedEthereumActive). Coins are configured per account now;
+   * kept for migration / compatibility with persisted app config.
+   */
+  bitcoinActive: boolean;
+  litecoinActive: boolean;
+  ethereumActive: boolean;
+  authentication: boolean;
+  btc: TBtcCoinConfig;
+  tbtc: TBtcCoinConfig;
+  rbtc: TBtcCoinConfig;
+  ltc: TBtcCoinConfig;
+  tltc: TBtcCoinConfig;
+  eth: TEthCoinConfig;
+  teth: Record<string, never>;
+  reth: Record<string, never>;
+  fiatList: Fiat[];
+  mainFiat: Fiat;
+  userLanguage: string;
+  btcUnit: BtcUnit;
+  startInTestnet: boolean;
+  gapLimitReceive: number;
+  gapLimitChange: number;
+}>;
+
+export type TConfig = {
+  readonly backend: TConfigBackend;
+  readonly frontend: TConfigFrontend;
+};
+
+/**
+ * Fetch config from the backend (see handlers.getAppConfig).
+ * Use setConfig from @/utils/config for partial merge-on-write updates.
+ */
+export const getConfig = (): Promise<TConfig> => apiGet('config');
+
+/**
+ * Post a config object to the backend.
+ */
+export const setConfig = (config: TConfig): Promise<void> => {
+  return apiPost('config', config);
+};
--- a/frontends/web/src/components/banners/offline-error.tsx
+++ b/frontends/web/src/components/banners/offline-error.tsx
@@ -1,9 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Message } from '../message/message';
-import { getConfig } from '@/utils/config';
+import { useConfig } from '@/contexts/ConfigProvider';
 import style from './offline-errors.module.css';
 
 type Props = {
@@ -13,13 +12,9 @@ type Props = {
 export const OfflineError = ({
   error,
 }: Props) => {
-
   const { t } = useTranslation();
-  const [usesProxy, setUsesProxy] = useState<boolean>();
-
-  useEffect(() => {
-    getConfig().then(({ backend }) => setUsesProxy(backend.proxy.useProxy));
-  }, []);
+  const { config } = useConfig();
+  const usesProxy = config?.backend.proxy?.useProxy ?? false;
 
   // Status: offline error
   const offlineErrorTextLines: string[] = [];
--- a/frontends/web/src/components/new-badge/new-badge.tsx
+++ b/frontends/web/src/components/new-badge/new-badge.tsx
@@ -2,15 +2,13 @@
 
 import { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import type { TConfigFrontendBadgeKey } from '@/api/config';
 import { Badge } from '@/components/badge/badge';
-import { useLoad } from '@/hooks/api';
-import { getConfig, setConfig } from '@/utils/config';
-
-type TConfigKey = 'hasSeenMarketplaceNudge' | 'hasSeenSwapMarketTab' | 'hasSeenOtcMarketTab';
+import { useConfig } from '@/contexts/ConfigProvider';
 
 type TProps = {
   className?: string;
-  configKey: TConfigKey;
+  configKey: TConfigFrontendBadgeKey;
   hideOnPathPrefix?: string;
   markAsSeen?: boolean;
   pathname?: string;
@@ -28,7 +26,7 @@ export const NewBadge = ({
   type = 'new',
 }: TProps) => {
   const { t } = useTranslation();
-  const config = useLoad(getConfig);
+  const { config, setConfig } = useConfig();
   const [showBadge, setShowBadge] = useState<boolean | undefined>(undefined);
 
   const persistAsSeen = useCallback(() => {
@@ -37,14 +35,13 @@ export const NewBadge = ({
     }
     setShowBadge(false);
     setConfig({ frontend: { [configKey]: true } });
-  }, [configKey, showBadge]);
+  }, [configKey, setConfig, showBadge]);
 
   useEffect(() => {
     if (!config) {
       return;
     }
-    const frontendConfig = config.frontend as Record<string, unknown> | undefined;
-    const hasSeenBadge = Boolean(frontendConfig?.[configKey]);
+    const hasSeenBadge = config.frontend[configKey];
     setShowBadge(currentShowBadge => (
       currentShowBadge === false ? false : !hasSeenBadge
     ));
--- a/frontends/web/src/components/status/status.tsx
+++ b/frontends/web/src/components/status/status.tsx
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { ReactNode, useCallback, useEffect, useState } from 'react';
-import { getConfig, setConfig } from '@/utils/config';
+import { ReactNode } from 'react';
+import type { TConfigFrontendDismissibleKey } from '@/api/config';
+import { useConfig } from '@/contexts/ConfigProvider';
 import { CloseXDark, CloseXWhite } from '@/components/icon';
 import { useDarkmode } from '@/hooks/darkmode';
 import { Message } from '@/components/message/message';
@@ -15,7 +16,7 @@ type TProps = {
   // used as keyName in the config if dismissing the status should be persisted, so it is not
   // shown again. Use an empty string if it should be dismissible without storing it in the
   // config, so the status will be shown again the next time.
-  dismissibleKey: string;
+  dismissibleKey: TConfigFrontendDismissibleKey | '';
   className?: string;
   children: ReactNode;
 };
@@ -28,21 +29,15 @@ export const Status = ({
   className = '',
   children,
 }: TProps) => {
-  // note: dismissible can be falsy i.e. empty string ''
-  const [show, setShow] = useState(dismissibleKey ? false : true);
-
+  const { config, setConfig } = useConfig();
   const { isDarkMode } = useDarkmode();
 
-  const checkConfig = useCallback(async () => {
-    if (dismissibleKey) {
-      const config = await getConfig();
-      setShow(!config ? true : !config.frontend[dismissibleKey]);
-    }
-  }, [dismissibleKey]);
-
-  useEffect(() => {
-    checkConfig();
-  }, [checkConfig]);
+  // note: dismissibleKey can be falsy i.e. empty string ''
+  const show = hidden
+    ? false
+    : dismissibleKey
+      ? (config ? !config.frontend[dismissibleKey] : true)
+      : true;
 
   const dismiss = async () => {
     if (!dismissibleKey) {
@@ -53,10 +48,9 @@ export const Status = ({
         [dismissibleKey]: true,
       }
     });
-    setShow(false);
   };
 
-  if (hidden || !show) {
+  if (!show) {
     return null;
   }
 
--- a/frontends/web/src/contexts/AppProvider.tsx
+++ b/frontends/web/src/contexts/AppProvider.tsx
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { ReactNode, useEffect, useState } from 'react';
-import { getConfig, setConfig } from '@/utils/config';
+import { useConfig } from './ConfigProvider';
 import { AppContext } from './AppContext';
 import { useLoad } from '@/hooks/api';
 import { useDefault } from '@/hooks/default';
@@ -19,6 +19,7 @@ type TProps = {
 };
 
 export const AppProvider = ({ children }: TProps) => {
+  const { config, setConfig } = useConfig();
   const nativeLocale = i18nextFormat(useDefault(useLoad(getNativeLocale), 'de-CH'));
   const isTesting = useDefault(useLoad(getTesting), false);
   const isOnline = useSync(getOnline, subscribeOnline);
@@ -62,19 +63,17 @@ export const AppProvider = ({ children }: TProps) => {
   }, [activeSidebar, isMobile, orientation]);
 
   useEffect(() => {
-    getConfig().then(({ frontend }) => {
-      if (frontend) {
-        if (frontend.guideShown !== undefined) {
-          setGuideShown(frontend.guideShown);
-        }
-        if (frontend.hideAmounts !== undefined) {
-          setHideAmounts(frontend.hideAmounts);
-        }
-      } else {
-        setGuideShown(true);
-      }
-    });
-  }, []);
+    if (!config) {
+      return;
+    }
+    const { frontend } = config;
+    if (frontend.guideShown !== undefined) {
+      setGuideShown(frontend.guideShown);
+    }
+    if (frontend.hideAmounts !== undefined) {
+      setHideAmounts(frontend.hideAmounts);
+    }
+  }, [config]);
 
   return (
     <AppContext.Provider
--- a/frontends/web/src/contexts/ConfigContext.tsx
+++ b/frontends/web/src/contexts/ConfigContext.tsx
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { createContext } from 'react';
+import type { TConfig } from '@/api/config';
+import type { TConfigUpdate } from '@/utils/config';
+
+export type TConfigContext = {
+  config: TConfig | undefined;
+  setConfig: (object: TConfigUpdate) => Promise<TConfig>;
+};
+
+export const ConfigContext = createContext<TConfigContext | undefined>(undefined);
--- a/frontends/web/src/contexts/ConfigProvider.tsx
+++ b/frontends/web/src/contexts/ConfigProvider.tsx
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { ReactNode, useCallback, useContext, useEffect, useState } from 'react';
+import { getConfig } from '@/api/config';
+import { setConfig as setConfigAPI } from '@/utils/config';
+import type { TConfig } from '@/api/config';
+import type { TConfigUpdate } from '@/utils/config';
+import { ConfigContext, TConfigContext } from './ConfigContext';
+
+type TProps = {
+  children: ReactNode;
+};
+
+export const ConfigProvider = ({ children }: TProps) => {
+  const [config, setConfigState] = useState<TConfig | undefined>(undefined);
+
+  useEffect(() => {
+    getConfig().then(setConfigState).catch(console.error);
+  }, []);
+
+  const setConfig = useCallback((object: TConfigUpdate) => {
+    return setConfigAPI(object).then(nextConfig => {
+      setConfigState(nextConfig);
+      return nextConfig;
+    });
+  }, []);
+
+  const value: TConfigContext = {
+    config,
+    setConfig
+  };
+
+  return (
+    <ConfigContext.Provider value={value}>
+      {children}
+    </ConfigContext.Provider>
+  );
+};
+
+export const useConfig = (): TConfigContext => {
+  const context = useContext(ConfigContext);
+  if (!context) {
+    throw new Error('useConfig must be used within ConfigProvider');
+  }
+  return context;
+};
--- a/frontends/web/src/contexts/DarkmodeProvider.tsx
+++ b/frontends/web/src/contexts/DarkmodeProvider.tsx
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { useState, useEffect, ReactNode, useCallback } from 'react';
-import { getConfig, setConfig } from '@/utils/config';
+import { useConfig } from './ConfigProvider';
 import { setDarkTheme, detectDarkTheme } from '@/api/darktheme';
 import { runningInAndroid } from '@/utils/env';
 import { useMediaQuery } from '@/hooks/mediaquery';
@@ -12,6 +12,7 @@ type TProps = {
 };
 
 export const DarkModeProvider = ({ children }: TProps) => {
+  const { config, setConfig } = useConfig();
   const [isDarkMode, setIsDarkMode] = useState(false);
   const androidPrefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)');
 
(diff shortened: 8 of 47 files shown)


frontends/web/src/hooks/bitsurance.ts
```diff
@@ -75,15 +75,17 @@ export const useBitsurance = (
       return;
     }
 
-    // we fetch the config after the lookup as it could have changed.
```
thisconnect, line 77: As this comment explicitly mentions to fetch config in case backend changed. Please keep getConfig and the comment.

Please squash the change into this commit.

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Fetch current config from the backend.
+ */
+export const getConfig = (): Promise<TConfig> => {
+  return apiGet('config');
+};
+
+/**
+ * Merge partial config with current, POST to backend, return new config.
+ * Source of truth remains the backend.
```
thisconnect: "Source of truth remains the backend" is not 100% correct as the function returns locally merged config object but doesn't get it from the backend. 

It could do a getConfig to request the up to date config object, but lets consider doing this later.

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,37 @@
...
+        frontend: Object.assign({}, currentConfig.frontend, pendingConfig.frontend, object.frontend)
+      });
+      pendingConfig = nextConfig;
+      return apiPost('config', nextConfig)
+        .then(() => {
+          pendingConfig = {};
+          return nextConfig;
+        });
+    });
+};
```
thisconnect, line 129: Ideally functions in `api/*` only wrap apiGet/apiPost and type function arguments and return types.

This function has a lot of logic, it would be nice if you could move the logic and pendingConfig global variable into the context. 

I am not sure if that is straight forward because of the code in i18n, maybe that would need to be refactored first.

frontends/web/src/components/banners/offline-error.tsx
```diff
@@ -13,14 +12,9 @@ type Props = {
...
   error,
 }: Props) => {
-
   const { t } = useTranslation();
-  const { getConfig } = useConfig();
-  const [usesProxy, setUsesProxy] = useState<boolean>();
-
-  useEffect(() => {
-    getConfig().then(({ backend }) => setUsesProxy(backend.proxy.useProxy));
-  }, [getConfig]);
```
thisconnect: nice, it looks like we could remove some more effects like this i.e in coincontrol.

frontends/web/src/api/config.ts
```diff
@@ -3,8 +3,8 @@
 import { apiGet, apiPost } from '@/utils/request';
 
 export type TConfig = {
-  backend?: unknown;
-  frontend?: unknown;
+  backend?: Record<string, unknown>;
+  frontend?: Record<string, unknown>;
```
thisconnect: When a component gets the new config object form the context, we should instruct typescript that this is a readonly object so components do not accidentally mutate the object.

`config.frontend.foo = 'bar';` should throw a typescript error.

I played around a bit and it seems is what it takes, please use this and make the config object readonly.

```diff
diff --git a/frontends/web/src/api/config.ts b/frontends/web/src/api/config.ts
index e8d1b1998..821ed8f60 100644
--- a/frontends/web/src/api/config.ts
+++ b/frontends/web/src/api/config.ts
@@ -3,11 +3,11 @@
 import { apiGet, apiPost } from '@/utils/request';
 
 export type TConfig = {
-  backend?: Record<string, unknown>;
-  frontend?: Record<string, unknown>;
+  readonly backend: Readonly<Record<string, unknown>>;
+  readonly frontend: Readonly<Record<string, unknown>>;
 };
 
-let pendingConfig: TConfig = {};
+let pendingConfig: Partial<TConfig> = {};
 
 /**
  * Fetch current config from the backend.
@@ -20,9 +20,9 @@ export const getConfig = (): Promise<TConfig> => {
  * Merge partial config with current, POST to backend, return new config.
  * Source of truth remains the backend.
  */
-export const setConfig = (object: TConfig): Promise<TConfig> => {
+export const setConfig = (object: Partial<TConfig>): Promise<TConfig> => {
   return getConfig()
-    .then((currentConfig = {}) => {
+    .then((currentConfig: Partial<TConfig> = {}) => {
       const nextConfig = Object.assign(currentConfig, {
         backend: Object.assign({}, currentConfig.backend, pendingConfig.backend, object.backend),
         frontend: Object.assign({}, currentConfig.frontend, pendingConfig.frontend, object.frontend)
diff --git a/frontends/web/src/contexts/ConfigContext.tsx b/frontends/web/src/contexts/ConfigContext.tsx
index 78ec8cdaa..abb975402 100644
--- a/frontends/web/src/contexts/ConfigContext.tsx
+++ b/frontends/web/src/contexts/ConfigContext.tsx
@@ -7,7 +7,7 @@ export type { TConfig };
 
 export type TConfigContext = {
   config: TConfig | undefined;
-  setConfig: (object: TConfig) => Promise<TConfig>;
+  setConfig: (object: Partial<TConfig>) => Promise<TConfig>;
 };
 
 export const ConfigContext = createContext<TConfigContext | undefined>(undefined);
diff --git a/frontends/web/src/contexts/ConfigProvider.tsx b/frontends/web/src/contexts/ConfigProvider.tsx
index 70338a6f2..55df49a8a 100644
--- a/frontends/web/src/contexts/ConfigProvider.tsx
+++ b/frontends/web/src/contexts/ConfigProvider.tsx
@@ -15,7 +15,7 @@ export const ConfigProvider = ({ children }: TProps) => {
     getConfig().then(setConfigState).catch(console.error);
   }, []);
 
-  const setConfig = useCallback((object: TConfig) => {
+  const setConfig = useCallback((object: Partial<TConfig>) => {
     return setConfigAPI(object).then(nextConfig => {
       setConfigState(nextConfig);
       return nextConfig;

```

frontends/web/src/contexts/RatesProvider.tsx
```diff
@@ -18,17 +18,18 @@ export const RatesProvider = ({ children }: TProps) => {
   const [activeCurrencies, setActiveCurrencies] = useState<Fiat[]>(['USD', 'EUR', 'CHF']);
   const [btcUnit, setBtcUnit] = useState<BtcUnit>('default');
 
-  const updateRatesConfig = useCallback(() => {
+  const updateRatesConfig = useCallback((): Promise<void> => {
     if (config === undefined) {
-      return;
+      return Promise.resolve();
```
thisconnect: You found a function that is only used here and doesn't need to be exported.

Keep as is for now, but in the future we should make this internal only and remove from RatesContext.tsx

frontends/web/src/contexts/RatesProvider.tsx
```diff
@@ -18,17 +18,18 @@ export const RatesProvider = ({ children }: TProps) => {
...
     }
     if (config.backend?.mainFiat) {
-      setDefaultCurrency(config.backend.mainFiat);
+      setDefaultCurrency(config.backend.mainFiat as Fiat);
     }
     if (config.backend?.fiatList && config.backend?.btcUnit) {
-      setActiveCurrencies(config.backend.fiatList);
-      setBtcUnit(config.backend.btcUnit);
+      setActiveCurrencies(config.backend.fiatList as Fiat[]);
+      setBtcUnit(config.backend.btcUnit as BtcUnit);
```
thisconnect: In a future improvement we should type all configs in backend and frontend, then we can remove this typecasting.

frontends/web/src/components/status/status.tsx
```diff
@@ -28,20 +28,16 @@ export const Status = ({
...
-    }
-  }, [dismissible]);
-
   useEffect(() => {
-    checkConfig();
-  }, [checkConfig]);
+    if (dismissible && config !== undefined) {
+      setShow(!config.frontend?.[dismissible]);
+    }
+  }, [dismissible, config]);
```
thisconnect: GPT mentions 

> Since coinControlEnabled is derived state, you might not need useEffect at all

here and probably in many other places, but lets keep it for now or do it in a later commit

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -27,56 +28,54 @@ export const DarkModeProvider = ({ children }: TProps) => {
...
-    getConfig()
-      .then(async config => {
-        let preferredDarkMode;
-        if (runningInAndroid()) {
-          preferredDarkMode = androidPrefersDarkMode;
-        } else {
-          preferredDarkMode = await detectDarkTheme();
-        }
-        if (preferredDarkMode === darkmode) {
-          // remove darkmode from config, so it use the same mode as the OS
```
thisconnect, line 62: please keep the comment, in general please always keep comment in the code and never just drop those they were added on purpose. Maybe something to add to Agents.md

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -27,56 +28,54 @@ export const DarkModeProvider = ({ children }: TProps) => {
...
+    (async () => {
+      let preferredDarkMode;
+      if (runningInAndroid()) {
+        preferredDarkMode = androidPrefersDarkMode;
+      } else {
+        preferredDarkMode = await detectDarkTheme();
+      }
+      if (preferredDarkMode === darkmode) {
+        const frontend = config.frontend && typeof config.frontend === 'object' ? config.frontend : {};
+        const { darkmode: _, ...rest } = frontend;
```
thisconnect: The version before was much simpler, in a previous comment I recommended making config readonly with that we also ensure that frontend is always present. Please revert to the version before if possible.

frontends/web/src/hooks/bitsurance.ts
```diff
@@ -74,15 +75,17 @@ export const useBitsurance = (
...
+    if (config === undefined) {
+      setInsured(false);
+      return;
+    }
+
+    let cancelledAccounts: string[] = (config.frontend?.bitsuranceNotifyCancellation as string[] | undefined) ?? [];
 
     if (cancelledAccounts?.includes(code)) {
       alertUser(t('account.insuranceExpired'));
-      // remove the pending notification from the frontend settings.
```
thisconnect, line 83: please keep the comment

thisconnect (changes requested):

frontends/web/src/api/config.ts
thisconnect: @smokyisthatyou the old comment was more correct.

frontends/web/src/api/config.ts
thisconnect: @smokyisthatyou please move the logic of getConfig and setConfig to utils/config,
but keep the api call + types here. 

So they work like all the other functions in `api/*` and contain no logic.

frontends/web/src/components/status/status.tsx
thisconnect: please check if this useEffect is straight forward to refactor and simplify.

thisconnect (changes requested):

frontends/web/src/utils/config.ts
```diff
@@ -1,40 +1,44 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { apiGet, apiPost } from '@/utils/request';
+import { getConfig as apiGetConfig, setConfig as apiSetConfig, type TConfig } from '@/api/config';
 
-type TConfig = {
-  backend?: unknown;
-  frontend?: unknown;
-};
+export type { TConfig };
```
thisconnect: There should be no need to export TConfig from the utils file. 

Types are currently defined in the `api/*` files and can be imported from there if needed.

frontends/web/src/utils/config.ts
```diff
@@ -1,40 +1,44 @@
...
+
+let pendingConfig: Partial<TConfig> = {};
```
thisconnect: Thinking about what it takes to move this into the react context.

The only place getConfig is used outside of react (component, hook, context) is in `i18n/config.ts` which is called once on app start. Here we could just use getConfig from api.

Similarly setConfig is used by i18n/i18n.ts, which is called once on startup (maybe not necessary) and when the user actively changes the language. Here I believe we cannot use setConfig from context, but there might be some workarounds (not 100% sure).

Other than for that we could move  pendingConfig + the logic in setConfig into a context in the future.

thisconnect: I think we should test/merge as is and in the future PR's:
1) try to move this pending variable and logic into context.
2) type config fully

both changes can be done independently

frontends/web/src/components/banners/offline-error.tsx
```diff
@@ -13,13 +12,9 @@ type Props = {
...
 }: Props) => {
-
   const { t } = useTranslation();
-  const [usesProxy, setUsesProxy] = useState<boolean>();
-
-  useEffect(() => {
-    getConfig().then(({ backend }) => setUsesProxy(backend.proxy.useProxy));
-  }, []);
+  const { config } = useConfig();
+  const usesProxy = (config?.backend?.proxy as { useProxy?: boolean } | undefined)?.useProxy;
```
thisconnect: Once we have config object fully typed in a future PR, this type casting can be simplified a lot or even removed completely.

frontends/web/src/components/status/status.tsx
```diff
@@ -52,7 +47,6 @@ export const Status = ({
         [dismissible]: true,
       }
     });
-    setShow(false);
   };
 
   if (hidden || !show) {
```
thisconnect: as `show` now takes the `hidden` prop into account, it looks like it would be enough to just check for `if (!show) {`

frontends/web/src/contexts/AppProvider.tsx
```diff
@@ -54,19 +55,18 @@ export const AppProvider = ({ children }: TProps) => {
...
-        }
-        if (frontend.hideAmounts !== undefined) {
-          setHideAmounts(frontend.hideAmounts);
-        }
-      } else {
-        setGuideShown(true);
+    const frontend = config?.frontend;
+    if (frontend && typeof frontend === 'object') {
+      if (frontend.guideShown !== undefined) {
+        setGuideShown(Boolean(frontend.guideShown));
```
thisconnect, line 61: Once we fully type TConfig in the future these safeguards can mostly be removed.

frontends/web/src/contexts/AppProvider.tsx
```diff
@@ -54,19 +55,18 @@ export const AppProvider = ({ children }: TProps) => {
...
       }
-    });
-  }, []);
+      if (frontend.hideAmounts !== undefined) {
+        setHideAmounts(Boolean(frontend.hideAmounts));
+      }
+    } else if (config !== undefined) {
```
thisconnect: Please keep as it was `} else {` the intention is that if the other conditions are false we want to always show the guide, regardless if config is undefined or not.

frontends/web/src/contexts/ConfigContext.tsx
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { createContext } from 'react';
+import type { TConfig } from '@/api/config';
+
+export type { TConfig };
```
thisconnect: no need to export TConfig here, we want to only import it from api 

(for now, until we one day maybe move all types into a single ./types directory, which could be generated from the go backend)

thisconnect (changes requested):
thank you I think we should merge this first step soon, as it reduces api calls to the config endpoint and is the base for further updates: like fully typing TConfig

smokyisthatyou:
@jadzeidan while working on this pr i think i introduced an old behaviour of the app that was opening the guide on first run. i implemented the code thinking it was the current expected flow, but @thisconnect pointed out this was actually deprecated in v4.40.

personally i’m not a huge fan of the auto-opening guide for aesthetic reasons, though I can see the utility. 
do you think we should reintroduce this or keep it as is?

jadzeidan:
thanks @smokyisthatyou for finding this 😄 

I agree that guide shouldn't open automatically, so let's keep the "new" behaviour.

smokyisthatyou:
@thisconnect rebased

smokyisthatyou:
@coderabbit review

frontends/web/src/components/banners/offline-error.tsx
```diff
@@ -13,13 +12,9 @@ type Props = {
...
 }: Props) => {
-
   const { t } = useTranslation();
-  const [usesProxy, setUsesProxy] = useState<boolean>();
-
-  useEffect(() => {
-    getConfig().then(({ backend }) => setUsesProxy(backend.proxy.useProxy));
-  }, []);
+  const { config } = useConfig();
+  const usesProxy = config?.backend.proxy.useProxy;
```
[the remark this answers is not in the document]
thisconnect: hm. I think it would be great if we could ensure that config and config.backend always exist.

So we could just use `config.backend.proxy?.useProxy`

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,136 @@
...
+export const getConfig = (): Promise<Partial<TConfig>> => {
+  return apiGet('config');
+};
+
+/**
+ * Post a config object to the backend.
+ */
+export const setConfig = (config: TConfig): Promise<void> => {
+  return apiPost('config', config);
+};
```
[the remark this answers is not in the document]
smokyisthatyou: the backend is currently expecting full config objects on POST, and not partial ones. it replaces the stored config rather than merging partial updates.

frontends/web/src/test/mock-config.ts
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import type { TConfig } from '@/api/config';
+
+const emptyBackendConfig = (): TConfig['backend'] => ({} as TConfig['backend']);
```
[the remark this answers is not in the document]
smokyisthatyou: i don't like the solution of adding defaults to TBackendConfig, as it will mean that we will need manual sync with the backend default, and that adds a maintenance burden. 
what if we normalise by merging `GET /config` over `GET /config/default`?

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { apiGet, apiPost } from '@/utils/request';
+import type { Fiat } from '@/api/account';
+import type { BtcUnit } from '@/api/coins';
+
+export type TElectrumServerInfo = Readonly<{
```
thisconnect: please only export the ones that are needed in other files

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export type TBtcCoinConfig = Readonly<{
+  electrumServers: TElectrumServerInfo[];
+}>;
+
+/** BTC-based coin keys in backend config (see backend/config/config.go). */
+export type TBtcCoinConfigKey = 'btc' | 'tbtc' | 'ltc' | 'tltc';
+
+export type TEthCoinConfig = Readonly<{
+  activeERC20Tokens: string[];
```
thisconnect, line 24: Since you are here. In the backend there is a comment that this is deprecated, please add comment here too.

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,136 @@
...
+export type TConfigUpdate = {
+  backend?: TBackendConfigUpdate;
+  frontend?: TFrontendConfigUpdate;
+};
+
+/**
+ * Fetch raw config from the backend. Keys may be missing; use getConfig from
+ * @/utils/config for a normalized TConfig.
+ */
+export const getConfig = (): Promise<Partial<TConfig>> => {
```
thisconnect: Partial is not deep-partial so it only affects frontend and backend key.

I think we can savely assume the backend always returns an object with frontend and backend.

with that we wont need to saveguard frontend or backend in components anymore.

frontends/web/src/contexts/ConfigProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { getConfig, setConfig as setConfigAPI } from '@/utils/config';
+import type { TConfig, TConfigUpdate } from '@/api/config';
+import { ConfigContext, TConfigContext } from './ConfigContext';
+
+type TProps = {
+  children: ReactNode;
+};
+
+export const ConfigProvider = ({ children }: TProps) => {
+  const [config, setConfigState] = useState<TConfig | undefined>(undefined);
```
thisconnect, line 15: If this was initialzed with `{backend: {}, frontend:{}}` instead of undefined we could drop the configcontext type `config: TConfig | undefined;` and  use `config.backend.foo` instead of `config?.backend...`

BUT then all the required keys in backend should be optional..
so I changed my mind and think it is probably better to keep as is.

and use `config?.backend.proxy?.proxystuff`

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,136 @@
...
+  Omit<Partial<TBackendConfig>, 'userLanguage'> & {
+    userLanguage?: string | null;
+  };
+
+export type TFrontendConfigUpdate = Partial<TFrontendConfig>;
+
+export type TConfigUpdate = {
+  backend?: TBackendConfigUpdate;
+  frontend?: TFrontendConfigUpdate;
+};
```
thisconnect, line 116: TConfigUpdate TBackendConfigUpdate is not used in api/config 

please move to the same file where it merges the config. I believ utils/config

frontends/web/src/components/banners/offline-error.tsx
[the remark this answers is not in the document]
thisconnect: We could initialize the context state with an object with frontend + backend keys. but then the types are not correct as all the backend is mandatory.

I guess we should use `config?.backend.proxy?.useProxy` (backend doesnt need it)

frontends/web/src/components/new-badge/new-badge.tsx
```diff
@@ -37,14 +35,13 @@ export const NewBadge = ({
...
-  }, [configKey, showBadge]);
+  }, [configKey, setConfig, showBadge]);
 
   useEffect(() => {
     if (!config) {
       return;
     }
-    const frontendConfig = config.frontend as Record<string, unknown> | undefined;
-    const hasSeenBadge = Boolean(frontendConfig?.[configKey]);
+    const hasSeenBadge = Boolean(config.frontend?.[configKey]);
```
thisconnect: Please double check but I believe  as you type TFrontendBadgeConfigKey to be always boolean, there shuold be no need to convert to boolean here and in other similar places


```suggestion
    const hasSeenBadge = config.frontend?.[configKey];
```

frontends/web/src/contexts/AppProvider.tsx
thisconnect: please take a look

thisconnect (commented):
partially reviewed

frontends/web/src/api/config.ts
thisconnect: ping ^ @smokyisthatyou please move the unused types

frontends/web/src/api/config.ts
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Fetch raw config from the backend. Response always includes `backend` and
+ * `frontend` objects (see handlers.getAppConfig). Nested backend fields may be
+ * incomplete; use getConfig from @/utils/config for a normalized TConfig.
+ */
+export const getConfig = () =>
+  apiGet('config') as Promise<
+    Readonly<{ backend: Partial<TBackendConfig>; frontend: TFrontendConfig }>
```
thisconnect: I think Partial<TBackendConfig> is wrong. The backend always returns TConfig which includes TBackendConfig. TBackendConfig is already typed readonly.

And instead of type casting "as Promise<...>" I would expect that you can just normaly type this.

Please double check but something like this should be sufficient here:

```ts
export const getConfig = (): Promise<TConfig> => apiGet('config');
```

frontends/web/src/components/new-badge/new-badge.tsx
```diff
@@ -2,15 +2,13 @@
...
-import { useLoad } from '@/hooks/api';
-import { getConfig, setConfig } from '@/utils/config';
-
-type TConfigKey = 'hasSeenMarketplaceNudge' | 'hasSeenSwapMarketTab' | 'hasSeenOtcMarketTab';
+import { useConfig } from '@/contexts/ConfigProvider';
 
 type TProps = {
   className?: string;
-  configKey: TConfigKey;
+  configKey: TFrontendBadgeConfigKey;
```
thisconnect: nit: maybe no so important (or for later), but should we consider renaming these and always prefix with TConfig?

TFrontendBadgeConfigKey -> TConfigFrontendBadgeKey ?

frontends/web/src/utils/config.ts
```diff
@@ -1,37 +1,52 @@
...
-import { apiGet, apiPost } from '@/utils/request';
+import { getConfig as apiGetConfig, setConfig as apiSetConfig, type TConfig, type TConfigUpdate } from '@/api/config';
 
-type TConfig = {
-  backend?: unknown;
-  frontend?: unknown;
-};
+let pendingConfig: TConfigUpdate = {};
+
+export const emptyBackendConfig = (): TConfig['backend'] => ({} as TConfig['backend']);
```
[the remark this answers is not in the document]
thisconnect: We trust the backend returns what is defined in api files.

frontends/web/src/utils/config.ts
```diff
@@ -1,37 +1,48 @@
...
- * get current configs
- * i.e. await getConfig()
- * returns a promise with backend and frontend configs
+ * Fetch current config from the backend.
  */
-export const getConfig = (): Promise<any> => {
-  return apiGet('config');
+export const getConfig = (): Promise<TConfig> => {
+  return apiGetConfig().then(normalizeConfig);
 };
```
thisconnect, line 16: As mention in previous comment we trust the backend returns what is defined in api files. For this reason the normalizing doesn't really make sense when we get the config. The normalizing makes sense when we post and update the config where it expects a full TConfig object.

Please remove getConfig here in utils/config and import getConfig from api/config directly.

frontends/web/src/utils/config.ts
```diff
@@ -1,37 +1,48 @@
...
-        frontend: Object.assign({}, currentConfig.frontend, pendingConfig.frontend, object.frontend)
-      });
+    .then((currentConfig) => {
+      const nextConfig: TConfig = {
+        backend: Object.assign(
+          {},
+          currentConfig.backend,
+          pendingConfig.backend,
+          object.backend,
+        ) as TConfig['backend'],
```
thisconnect, line 33: remove type casting

frontends/web/src/test/mock-config.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const emptyFrontendConfig = (): TConfig['frontend'] => ({});
+
+export const mockConfig = (overrides: Partial<TConfig> = {}): TConfig => ({
+  backend: { ...emptyBackendConfig(), ...overrides.backend },
+  frontend: { ...emptyFrontendConfig(), ...overrides.frontend },
+});
```
thisconnect: emptyBackendConfig and emptyFrontendConfig are just empty objects pretending to be real TBackendConfig objects and lie to Typescript :) 

I dislike maintaining an object  with fake properties too that needs to be in sync with backend too.

This mock helper function here seems somewhat useful for tests but often they just need 1 mocked key from the config for the test to work. Ideally the tests are contain everything need and mock the required config key for example feetarget.test.tsx.

So either we need to maintain a hardcoded TBackendConfig object or lie to typescript somewhere. Typecasting is usually a bit a red flag for me but in test and mock files it helps 😅 . 

the emptyBackendConfig and emptyFrontendConfig are not really needed, you can just merge the object and type case as TConfig

```tsx
export const mockConfig = (
  overrides: Partial<TConfig> = {}
): TConfig => ({
  backend: { ...overrides.backend },
  frontend: { ...overrides.frontend },
} as TConfig);
```

I made a comment in utils/config about  normalise by merging GET /config over GET /config/default

frontends/web/src/routes/settings/components/advanced-settings/custom-gap-limit-setting.tsx
```diff
@@ -5,18 +5,18 @@ import { useTranslation } from 'react-i18next';
...
 
 type TProps = {
   backendConfig?: TBackendConfig;
-  onChangeConfig: (config: TConfig) => void;
 };
 
-export const CustomGapLimitSettings = ({ backendConfig, onChangeConfig }: TProps) => {
+export const CustomGapLimitSettings = ({ backendConfig }: TProps) => {
   const { t } = useTranslation();
+  const { setConfig } = useConfig();
```
thisconnect: nit: it would be nice to take config from useConfig here then we don't need to bass backendConfig here.

frontends/web/src/routes/settings/components/advanced-settings/enable-auth-setting.tsx
```diff
@@ -1,21 +1,21 @@
...
 import { onAuthSettingChanged, TAuthEventObject, subscribeAuth, forceAuth } from '@/api/backend';
 import { runningInAndroid, runningInIOS } from '@/utils/env';
 
 type TProps = {
   backendConfig?: TBackendConfig;
-  onChangeConfig: Dispatch<TConfig>;
 };
 
-export const EnableAuthSetting = ({ backendConfig, onChangeConfig }: TProps) => {
+export const EnableAuthSetting = ({ backendConfig }: TProps) => {
```
thisconnect: nit: use config?.backend from useConfig(); a few lines below and drop passing config with backendConfig prop.

frontends/web/src/test/mock-config.ts
thisconnect: an more self contained alternative would be to just type cast in the test directly and drop this file

```ts
vi.mocked(config.getConfig).mockResolvedValue({
      frontend: {},
      backend: {},
    } as TConfig);
```

thisconnect (changes requested):
very nice almost there

 swap.tsx was recently merged, please also useConfig there instead of useLoad(getConfig)

frontends/mobiletests/e2e/export-logs-save.test.js
```diff
@@ -21,6 +21,20 @@ const getSaveDialogElement = async (driver) => {
   return null;
 };
 
+/** Dismiss the in-app guide overlay if it is open (blocks clicks on settings). */
+const dismissGuideIfPresent = async (driver) => {
+  await driver.execute(() => {
+    const overlays = document.querySelectorAll('div');
+    for (const el of overlays) {
```
thisconnect: Please use `data-testid` on the dialog component's overlay so you dont have to traverse the DOM here.

The react testing library style would more be to find web accessible elements by role atc. https://testing-library.com/docs/queries/about/#priority
but refactoring the diaog/overlay seems a bit out of scope of this PR so best to just add  `data-testid`  in this dialog.

frontends/web/src/contexts/DarkmodeProvider.tsx
```diff
@@ -27,56 +28,53 @@ export const DarkModeProvider = ({ children }: TProps) => {
...
-          // remove darkmode from config, so it use the same mode as the OS
-          const { darkmode, ...frontend } = config.frontend;
-          setConfig({
-            frontend: {
-              ...frontend,
-              darkmode: undefined,
-            },
-          });
-        } else {
-          // darkmode is different from OS, save to config
```
thisconnect, line 71: please keep the comment

frontends/web/src/routes/market/market.tsx
```diff
@@ -99,7 +99,7 @@ export const Market = ({
 
     if (config.frontend.selectedExchangeRegion) {
       // pre-select config region
-      setSelectedRegion(config.frontend.selectedExchangeRegion);
+      setSelectedRegion(String(config.frontend.selectedExchangeRegion));
```
thisconnect:
```suggestion
      setSelectedRegion(config.frontend.selectedExchangeRegion);
```

frontends/web/src/contexts/RatesProvider.tsx
```diff
@@ -13,26 +13,24 @@ type TProps = {
...
+    return Promise.resolve();
+  }, [config]);
 
-    if (appConf.backend?.fiatList && appConf.backend?.btcUnit) {
-      setActiveCurrencies(appConf.backend.fiatList);
-      setBtcUnit(appConf.backend.btcUnit);
-    }
-  };
+  useEffect(() => {
+    updateRatesConfig();
```
thisconnect: updateRatesConfig function is only used here, also the function returns a promise but that is not needed here. I think it is fine to just move the 5 lines logic of the function into the effect here.

frontends/web/src/routes/settings/advanced-settings.tsx
```diff
@@ -83,12 +49,12 @@ export const AdvancedSettings = ({ devices, hasAccounts }: TPagePropsWithSetting
...
-                <EnableAuthSetting backendConfig={backendConfig} onChangeConfig={setConfig} />
-                <EnableTorProxySetting proxyConfig={proxyConfig} onChangeConfig={setConfig} />
-                <RestartInTestnetSetting onChangeConfig={setConfig} />
-                <CustomGapLimitSettings backendConfig={backendConfig} onChangeConfig={setConfig} />
+                <EnableCustomFeesToggleSetting frontendConfig={config?.frontend} />
+                <EnableCoinControlSetting frontendConfig={config?.frontend} />
+                <EnableAuthSetting />
+                <EnableTorProxySetting proxyConfig={config?.backend.proxy} />
+                <RestartInTestnetSetting />
+                <CustomGapLimitSettings />
```
thisconnect, line 55: It would be nice to get rid of the remaining props where config is passed (frontendConfig and proxyConfig)

frontends/web/src/utils/config.ts
```diff
@@ -1,37 +1,72 @@
...
+ * Merge partial config with current, POST full TConfig to backend, return merged config.
+ * Does not refetch from the backend after POST.
  */
-export const setConfig = (object: TConfig) => {
-  return getConfig()
-    .then((currentConfig = {}) => {
-      const nextConfig = Object.assign(currentConfig, {
-        backend: Object.assign({}, currentConfig.backend, pendingConfig.backend, object.backend),
-        frontend: Object.assign({}, currentConfig.frontend, pendingConfig.frontend, object.frontend)
-      });
```
thisconnect, line 32: I dont understand why there is different behavior for merging frontend and backend. Or if you found a bug and this fixes something.

As this PR is already vvery big I'd prefer to keep the merging with Object.assign and you drop the new commit that add mergeBackend/mergeFrontend.

thisconnect (changes requested):

frontends/mobiletests/e2e/export-logs-save.test.js
```diff
@@ -0,0 +1,78 @@
...
+    await exportLogsButton.click();
+
+    // Allow native save picker to open after the export-log API call.
+    await driver.pause(3000);
+
+    await driver.switchContext('NATIVE_APP');
+    const saveDialog = await getSaveDialogElement(driver);
+    expect(saveDialog, 'expected Android save dialog to be visible').to.not.equal(null);
+  });
+});
```
thisconnect: In the meantime frontends/mobiletests/e2e were removed, I guess you added it back because of a merge conflict.

please remove, see https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4169

thisconnect (approved):
very nice LGTM

please remove the e2e file, other than that it's good to merge from my side.

PR #3961, opened 2026-03-30
backend: clean swakpit frontend/backend interaction

bznein:
Instead of returnign the whole response to the frontend, we only return what is needed to the frontend as QuoteRouteSummary.

We also change the account code type in the backend api, to make it more aligned to what other handlers do.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -1680,10 +1680,10 @@ func (handlers *Handlers) postConnectKeystore(r *http.Request) interface{} {
 
 func (handlers *Handlers) postSwapkitQuote(r *http.Request) interface{} {
 	type result struct {
-		Success      bool                   `json:"success"`
-		ErrorCode    errp.ErrorCode         `json:"errorCode,omitempty"`
-		ErrorMessage string                 `json:"errorMessage,omitempty"`
-		Quote        *swapkit.QuoteResponse `json:"quote,omitempty"`
+		Success      bool                        `json:"success"`
+		ErrorCode    errp.ErrorCode              `json:"errorCode,omitempty"`
+		ErrorMessage string                      `json:"errorMessage,omitempty"`
+		Routes       []swapkit.QuoteRouteSummary `json:"routes,omitempty"`
 	}
 	errorResult := func(code errp.ErrorCode, message string) result {
 		return result{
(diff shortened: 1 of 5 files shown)


frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -109,13 +109,13 @@ export const Swap = ({
...
-    const buyCoinCode = toAccount?.coinCode;
     const amount = Number(sellAmount);
 
     if (
-      !sellCoinCode
-      || !buyCoinCode
+      !sellAccountCode
+      || !buyAccountCode
+      || !fromAccount
+      || !toAccount
```
thisconnect: before sellCoinCode/buyCoinCode was derived from fromAccount/toAccount (TAccount object)

But now fromAccount/toAccount is only used in this condition.

I think it's save to remove fromAccount/toAccount here and from the effect's dependencyList.

It looks like toAccount is still used in a different place, but maybe we can remove that in the future too.

thisconnect (approved):
frontend LGTM with one small change

backend/handlers/handlers.go
```diff
@@ -1694,25 +1694,54 @@ func (handlers *Handlers) postSwapkitQuote(r *http.Request) interface{} {
...
 		request.SellAmount,
 	)
 	if quoteError != nil {
 		return errorResult(quoteError.ErrorCode, quoteError.Message)
 	}
+
 	return result{
 		Success: true,
-		Quote:   quoteResponse,
+		Routes:  swapkit.QuoteRouteSummariesFromResponse(quoteResponse),
```
Beerosagos, line 1745: nit: afaik we don't fetch quotes in any other place, so we could make `swapkit.NewQuoteFromCoinCode` return the object needed by the handler directly, and avoid the second helper

backend/market/swapkit/helpers.go
```diff
@@ -32,24 +34,43 @@ const apiKey = "0722e09f-9d3f-4817-a870-069848d03ee9"
 const ErrInvalidRequest errp.ErrorCode = "invalidRequest"
 
 // assetFromCoinCode translates an internal coin code into a SwapKit asset string.
-func assetFromCoinCode(coinCode string) (string, bool) {
-	asset, ok := swapkitAssetByCoinCode[strings.ToLower(strings.TrimSpace(coinCode))]
+func assetFromCoinCode(coinCode coinpkg.Code) (string, bool) {
+	asset, ok := swapkitAssetByCoinCode[strings.ToLower(strings.TrimSpace(string(coinCode)))]
```
Beerosagos, line 38: can we remove `ToLower` and `TrimSpace`?

backend/handlers/handlers.go
```diff
@@ -1694,25 +1694,54 @@ func (handlers *Handlers) postSwapkitQuote(r *http.Request) interface{} {
...
 	if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
 		return errorResult(swapkit.ErrInvalidRequest, "Request body is required and must be a valid JSON object.")
 	}
+	if request.SellAccountCode == "" {
+		return errorResult(swapkit.ErrInvalidRequest, "Missing sellAccountCode.")
+	}
+	if request.BuyAccountCode == "" {
+		return errorResult(swapkit.ErrInvalidRequest, "Missing buyAccountCode.")
+	}
+	if request.SellAccountCode == request.BuyAccountCode {
```
Beerosagos, line 1711: Maybe it would be better to move this check inside the swap logic? Comparing coins instead of accounts would be even better, I guess.

PR #3990, opened 2026-04-08
frontend/tx-detail-dialog: fix To label

benma:
The field always shows recipient address(es), not the sending address.


thisconnect (approved):
thank you 🙏 utACK

PR #3995, opened 2026-04-09
tests: move Playwright CI to stock image

bznein:
Run the Playwright job in the upstream Playwright container and adapt the regtest/AOPP  test setup to that environment.

Running on the Playwright container saves us the step of installing browsers, which can take a few minutes; sometimes, depending on network conditions, even up to 15.

Install the missing CI dependencies, make the regtest services reachable from the containerized job, add an RBTC electrum host override for CI, and harden the AOPP helper for raw signature recovery and missing RIPEMD160 support.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,13 @@ jobs:
           OS_NAME: linux
 
   playwright:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     needs: test-lint
+    container:
+      # Keep this in sync with frontends/web/package-lock.json.
+      image: mcr.microsoft.com/playwright:v1.57.0-jammy
+      # This below is needed to be able to run docker in the CI.
+      options: --volume /var/run/docker.sock:/var/run/docker.sock --add-host=host.docker.internal:host-gateway
     strategy:
       fail-fast: false
       matrix:
@@ -53,7 +58,18 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Install system dependencies
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends docker.io jq lsof python3-pip
+
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
       - name: Download simulator v9.24.0 from simulators.json
+        shell: bash
         run: |
           set -euo pipefail
           simulator_url=$(jq -r '.[] | select(.url | contains("firmware%2Fv9.24.0/")) | .url' backend/devices/bitbox02/testdata/simulators.json)
@@ -73,21 +89,12 @@ jobs:
       - name: Install dependencies
         run: |
           cd frontends/web
-          npm install
+          npm ci
 
       - name: Build frontend
         run: |
           cd frontends/web
           npm run build
-
-      - name: Install Playwright browsers
-        run: |
-          cd frontends/web
-          if [ "${{ matrix.project }}" = "Chromium" ]; then
-            npx playwright install --with-deps chromium
-          else
-            npx playwright install --with-deps webkit
-          fi
 
       - name: Setup Python 3
         uses: actions/setup-python@v6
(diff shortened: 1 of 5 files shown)


bznein:
@coderabbit review

bznein:
I left this onse on hold for a while as there were more pressing things, but I think now it's a good time to get this merged, as it can drastically improve run times for playwright

.github/workflows/ci.yml
```diff
@@ -43,8 +43,13 @@ jobs:
           OS_NAME: linux
 
   playwright:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     needs: test-lint
+    container:
+      # Keep this in sync with frontends/web/package-lock.json.
```
thisconnect: In case I update the lock file like here: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4144 is there anything else I should change to "keep in sync  frontends/web/package-lock.json." ?

.github/workflows/ci.yml
```diff
@@ -73,34 +89,26 @@ jobs:
       - name: Install dependencies
         run: |
           cd frontends/web
-          npm install
+          npm ci
```
thisconnect: 👍

.github/workflows/ci.yml
```diff
@@ -73,34 +89,26 @@ jobs:
...
           npm run build
 
-      - name: Install Playwright browsers
-        run: |
-          cd frontends/web
-          if [ "${{ matrix.project }}" = "Chromium" ]; then
-            npx playwright install --with-deps chromium
-          else
-            npx playwright install --with-deps webkit
-          fi
```
thisconnect, line 90: nice 👍  this can take 10+ minutes, see https://github.com/BitBoxSwiss/bitbox-wallet-app/actions/runs/25811176958/job/75828736293?pr=4144

bznein: Yeah that's the main winner. Sometime it takes very little time, but it can be super annoying when it takes so much

.github/workflows/ci.yml
bznein: If the PR were to up[date the version of playwright to, let's say, 1.58.0, you would need to update it below as well.

Let me know if the comment is not clear and we should update it.

thisconnect: Makes sense 👍 , but in that case I think it makes sense to change to where the playwright devDependency is defined:

```suggestion
      # Keep this in sync with frontends/web/package.json.
```

bznein:
@benma I forgot about this one honestly, but I think you are the most suited to review it :) Thanks!

bznein:
@benma bump :)

.github/workflows/ci.yml
```diff
@@ -43,8 +43,13 @@ jobs:
           OS_NAME: linux
 
   playwright:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
benma: Why not latest? Unless latest breaks, I prefer latest to not have to bump the versions every so often.

scripts/run_regtest.sh
```diff
@@ -29,27 +31,29 @@ killall() {
...
-echo -n "dbb:dbb" > "$ELECTRS_DATADIR1/rpccreds"
-echo -n "dbb:dbb" > "$ELECTRS_DATADIR2/rpccreds"
-echo "bitcoind datadir: ${BITCOIN_DATADIR}"
-echo "electrs datadir1: ${ELECTRS_DATADIR1}"
-echo "electrs datadir2: ${ELECTRS_DATADIR2}"
+docker volume create "$BITCOIND_VOLUME_NAME" >/dev/null
+docker volume create "$ELECTRS_VOLUME_NAME1" >/dev/null
+docker volume create "$ELECTRS_VOLUME_NAME2" >/dev/null
+
+docker run --rm -v "${ELECTRS_VOLUME_NAME1}:/data" alpine:3.20 \
```
benma: the below might also break use of this script under normal use locally with `make servewallet-regtest`?

codex:
---
 
Make Electrs volumes writable by its UID — scripts/run_regtest.sh:38

    (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/79723be76f9c6b6c312bf491d67798c32cd970c3/scripts/run_regtest.sh#L38-L41)

    Alpine initializes /data as root:root with mode 0755, but Electrs later runs as the invoking user via
    -u "$(id -u "$USER")" and uses /data as its database directory. Local E2E tests therefore fail for
    non-root users because Electrs cannot write its database. CI masks this because the Playwright
    container runs as root. Chown the initialized volumes to the selected UID, or stop overriding the
    Electrs container user when using named volumes.

    I reproduced the initialization locally with UID 1002; writing to the mounted volume failed with
    Permission denied.

benma (changes requested):

bznein:
@benma I left this one untouched for a while but found the time now for it; please take a look again I addressed your comments

PR #3998, opened 2026-04-09
frontend: gate swap behind connected multi.

bznein:
Note for @thisconnect: you can review this one if you want, but it is pending OK from Jad :) I just went ahead with the code while I was at it :)

We don't want to show swap to BTC only users, meaning we only show it if there is at least one non-btc account in the backend (either watch-only, or there is a multi device attached)

Add a swap-available API that returns true only when the currently connected keystore has at least one non-BTC account, including inactive accounts, and use that flag to show or hide the market swap tab and guard the swap route. Also refresh the flag when account state changes so disconnecting the device immediately removes swap access.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/market/market.tsx
+++ b/frontends/web/src/routes/market/market.tsx
@@ -7,6 +7,7 @@ import { useTranslation } from 'react-i18next';
 import { SingleValue } from 'react-select';
 import { i18n } from '@/i18n/i18n';
 import * as marketAPI from '@/api/market';
+import { getSwapStatus } from '@/api/swap';
 import { AccountCode, TAccount } from '@/api/account';
 import { GuidedContent, GuideWrapper, Header, Main } from '@/components/layout';
 import { View, ViewContent } from '@/components/view/view';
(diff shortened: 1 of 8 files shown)


frontends/web/src/routes/market/market.tsx
```diff
@@ -47,12 +48,12 @@ export const Market = ({
...
   const [supportedAccounts, setSupportedAccounts] = useState<TAccount[]>([]);
   const [activeTab, setActiveTab] = useState<marketAPI.TMarketAction>('buy');
-  const [pendingSwapNavigation, setPendingSwapNavigation] = useState(false);
 
   const regionCodes = useLoad(marketAPI.getMarketRegionCodes);
   const nativeLocale = useLoad(getNativeLocale);
   const config = useLoad(getConfig);
-  const connectedSwapAccounts = accounts.filter(account => account.keystore.connected);
+  const swapAvailable = useLoad(getSwapAvailable, [accounts]);
+  const showSwap = swapAvailable?.available || false;
```
thisconnect: nit: rename to isSwapAvailable, canSwap or similar.

frontends/web/src/routes/market/market.tsx
```diff
@@ -145,6 +157,13 @@ export const Market = ({
     buyDealsResponse, sellDealsResponse, spendDealsResponse
   ]);
 
+  useEffect(() => {
+    if (showSwap || activeTab !== 'swap') {
+      return;
+    }
+    setActiveTab('buy');
```
thisconnect: This changes the tab to "buy" when I unplug the device and have all account as watch-only.

Is this the intended behavior? 
I'd expect it to stay on swap tab.

I think you could just remove this effect completely. Or is there another purpose?

thisconnect (approved):
tested LGTM, with small nit

frontends/web/src/routes/market/market.tsx
bznein: yeah you are absolutely right!

PR #3999, opened 2026-04-12
refactor: use slices.Backward to simplify the code

chuanshanjida:
- [ ] updating the Changelog
- [ ] writing unit tests

There is a [new function](https://pkg.go.dev/slices@go1.23.0#Backward)  added in the go1.23 standard library, which can make the code more concise and easy to read.

More info can see  https://github.com/golang/go/issues/61899


benma:
No need, thanks.

PR #4000, opened 2026-04-12
frontend: show addresses in groups of 4 chars

benma:
To match BitBox02 display.


frontends/web/src/routes/account/sign-message/sign-message.test.tsx
```diff
@@ -39,8 +39,16 @@ const ethAccount: accountApi.TAccount = {
   name: 'Ethereum Account',
 };
 
+const groupAddress = (value: string): string => {
+  if (value.startsWith('0x') || value.startsWith('0X')) {
+    return `${value.slice(0, 2)} ${value.slice(2).replace(/(.{4})/g, '$1 ').trim()}`;
```
thisconnect, line 44: I wasn't aware eth addresses would be chunked too, https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1832/changes#diff-2060fe0aafeb5b2132b1235a2722c9b16d5efd4b4989e6c85f230b7707fb0eda

maybe mention in changelog or in commit message and link fw commit?

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -238,13 +238,10 @@ export const Receive = ({
...
-  if (currentAddresses) {
-    address = (currentAddresses[activeIndex] as accountApi.TReceiveAddress).address;
-    if (!verifying) {
-      address = address.substring(0, 8) + '...';
-    }
-  }
+  const currentAddress = currentAddresses?.[activeIndex];
+  const address = currentAddress?.address || '';
+  const displayAddress = currentAddress?.displayAddress ?? '';
+  const displayedMainAddress = verifying ? displayAddress : (address ? address.substring(0, 8) + '...' : '');
```
thisconnect, line 244: nit: you could use import { truncateMiddle } from '@/utils/address';

thisconnect (commented):
did light review&testing.

One small issue is if some user still are on old FW, the app shows chunked whereas the device showing one long string.

Frontend lgtm with small nit

frontends/web/src/routes/account/sign-message/sign-message.test.tsx
benma: added CHANGELOG entry :+1: 

Firmware changelog entry will come when we add the firmware.

frontends/web/src/components/wallet-connect/incoming-signing-request.tsx
```diff
@@ -37,6 +37,7 @@ export const WCSigningRequest = () => {
     setDialogContent({
       accountAddress,
       accountName,
+      displayAddress,
```
thisconnect, line 40: Just passing here doesn't do anything (and typescript doesn't care).

It would need to take this key and render in the UI in frontends/web/src/components/wallet-connect/incoming-signing-request-dialog.tsx
 ( that file was editied but only the  newline at the end was added in this PR :) )

In the WC signing request it shows the address truncated in the middle (was already like this before)

so I think currently we can just drop these changes for WC.

Image

thisconnect (approved):
partly tested LGTM with one nit

PR #3993, opened 2026-04-09
[WIP] Implement swap eth

bznein:
WIP - but can be used to test swaps from ETH.

Note: builds on top of #3988 

Before asking for reviews, here is a check list of the most common things you might need to consider:
- [ ] updating the Changelog
- [ ] writing unit tests
- [ ] checking if your changes affect other coins or tokens in unintended ways
- [ ] testing on multiple environments (Qt, Android, ...)
- [ ] having an AI review your changes

PR #4003, opened 2026-04-13
swap: implement swap from ETH

bznein:
In favour of #3993 

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/swap.go
+++ b/backend/swap.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
 	accountsTypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
-	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	btcaddresses "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/addresses"
 	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth"
@@ -40,7 +39,7 @@ type SwapSignTxInput struct {
 	PaymentRequest *paymentrequest.Slip24 `json:"paymentRequest"`
 }
 
-// SwapPreparation contains everything the frontend needs to reuse the regular BTC send flow.
+// SwapPreparation contains everything the frontend needs to reuse the regular account tx flow.
 type SwapPreparation struct {
 	ExpectedBuyAmount string          `json:"expectedBuyAmount"`
 	SwapID            string          `json:"swapId"`
@@ -247,7 +246,7 @@ func (backend *Backend) accountHasNonZeroBalance(accountCode accountsTypes.Code)
 }
 
 // PrepareSwap prepares a real SwapKit swap and returns a tx input that can be proposed and sent
-// through the existing BTC payment-request flow.
+// through the existing account tx flow.
 func (backend *Backend) PrepareSwap(
 	buyAccountCode, sellAccountCode accountsTypes.Code,
 	routeID, sellAmount string,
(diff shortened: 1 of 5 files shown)


backend/swap_destination_accounts.go
```diff
@@ -183,3 +301,149 @@ func (backend *Backend) appendERC20SwapDestinationAccounts(
...
+		}, typedAddress, nil
+	default:
+		return "", nil, nil, errp.New("Unsupported swap destination address type")
+	}
+}
+
+func swapSourceAddress(account accounts.Interface) (string, error) {
+	switch typedAccount := account.(type) {
+	case *btc.Account:
+		spendableOutputs, err := typedAccount.SpendableOutputs()
```
benma: Not all spendable outputs are in the actual swap transaction. And what is this "source address" used for anyway?

backend/swap_destination_accounts.go
```diff
@@ -183,3 +301,149 @@ func (backend *Backend) appendERC20SwapDestinationAccounts(
...
+		address, err := firstUnusedAddress(typedAccount)
+		if err != nil {
+			return "", err
+		}
+		return address.EncodeForHumans(), nil
+	case *eth.Account:
+		address, err := firstUnusedAddress(typedAccount)
+		if err != nil {
+			return "", err
+		}
```
benma: This is just `typedAccount.Address()`

benma (changes requested):
Codex

---
  1. High: ETH sell swaps are still blocked by the old BTC-only capability check.
     The new flow calls hasSwapPaymentRequest(sellAccountCode) before signSwap() in frontends/web/src/routes/market/swap/swap.tsx:253, but /has-swap-payment-request still rejects anything except *btc.Account in
     backend/coins/btc/handlers/handlers.go:823. So an ETH sell account now reaches the new backend swap path, but the UI gate fails first with "An account must be BTC based...". In practice, the feature in this
     commit is not reachable from the frontend.

bznein:
> Codex
> 
> 1. High: ETH sell swaps are still blocked by the old BTC-only capability check.
>    The new flow calls hasSwapPaymentRequest(sellAccountCode) before signSwap() in frontends/web/src/routes/market/swap/swap.tsx:253, but /has-swap-payment-request still rejects anything except *btc.Account in
>    backend/coins/btc/handlers/handlers.go:823. So an ETH sell account now reaches the new backend swap path, but the UI gate fails first with "An account must be BTC based...". In practice, the feature in this
>    commit is not reachable from the frontend.

Right, fixed!

backend/swap_destination_accounts.go
bznein: Removed this function as discussed on MM

bznein:
@benma ping :) (this is, thankfully, still testable with firmware 9.26.0)

backend/swap.go
```diff
@@ -260,37 +260,48 @@ func (backend *Backend) PrepareSwap(
...
+	switch sellAccount.(type) {
+	case *btc.Account, *eth.Account:
+	default:
+		return nil, errp.New("Only BTC/ETH/ERC20 sell accounts are currently supported")
+	}
+	switch buyAccount.(type) {
+	case *btc.Account, *eth.Account:
+	default:
+		return nil, errp.New("Only BTC/ETH/ERC20 receive accounts are currently supported")
 	}
```
benma: Should this be comparing account.Coin().Code() instead, to exclude testnets?

backend/swap.go
```diff
@@ -260,37 +260,48 @@ func (backend *Backend) PrepareSwap(
...
 
-	sourceAddress, err := swapSourceAddress(specificSellAccount)
```
benma, line 337: Is this the same as using `firstUnusedAddress`?

bznein: Yeah good gatch!

backend/swap.go
bznein: It makes sense yeah, will do

benma (approved):
utACK

PR #4004, opened 2026-04-13
Fix "x" in strings

jadzeidan:
Inter font automatically makes "x" into a
multiplication symbol when in between numbers.

This commit fixes that.

### DIFF
--- a/frontends/web/src/style/base.css
+++ b/frontends/web/src/style/base.css
@@ -15,6 +15,8 @@ body {
     background-color: var(--background);
     color: var(--color-default);
     font-family: var(--font-family);
+    font-feature-settings: "calt" 0, "clig" 0, "dlig" 0, "frac" 0, "liga" 0;
+    font-variant-ligatures: none;
     font-weight: 400;
     height: 100%;
     line-height: 1.3;
@@ -25,6 +27,11 @@ body {
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
 	  -webkit-app-region: drag;
+}
+
+button, textarea, input, select {
+    font-feature-settings: "calt" 0, "clig" 0, "dlig" 0, "frac" 0, "liga" 0;
+    font-variant-ligatures: none;
 }
 
 /*


jadzeidan:
cc @thisconnect PTAL

thisconnect:
I asked if there could be other similar "features".

Short answer: **yes — `calt` is not the only thing you should worry about**. If you need **100% literal, character-by-character rendering** (e.g. private keys), you should disable a broader set of OpenType features and also think beyond just the font.

---

## Why `calt` alone isn’t enough

In **Inter**, `calt` (contextual alternates) is the most likely cause of your `x → ×` issue — but other features can *also* change what users see:

### 1. Ligatures (`liga`, `clig`, `dlig`)

These can merge characters:

* `fi` → ﬁ
* `fl` → ﬂ

Even if harmless visually, this is **not acceptable for sensitive strings**.

---

### 2. Stylistic / contextual substitutions (`calt`, `salt`)

You already hit this — characters can change depending on neighbors.

---

### 3. Numeric features (`tnum`, `lnum`, etc.)

Usually safe, but can affect spacing/appearance of digits.

---

### 4. Case-sensitive forms (`case`)

Can subtly change punctuation positioning.

---

### 5. Fractions (`frac`)

Example:

```id="n1"
1/2 → ½
```

Definitely something you *don’t* want in a key.

---

## What you should do instead (recommended setup)

If correctness matters (like private keys), don’t just patch features — **lock the rendering down**.

### ✅ 1. Disable all ligatures and contextual behavior

```css
.secure-text {
  font-family: 'Inter', monospace;
  font-feature-settings: 
    "calt" 0,
    "liga" 0,
    "clig" 0,
    "dlig" 0,
    "frac" 0;
  font-variant-ligatures: none;
}
```

---

### ✅ 2. Even better: use a monospace font

For things like private keys, the industry standard is:

* `monospace` fonts (no ambiguity, fixed width)
* fonts designed for code

Examples:

* JetBrains Mono
* Fira Code *(but disable ligatures!)*
* Roboto Mono

⚠️ Note: some coding fonts (like Fira Code) **intentionally add ligatures**, so you still need to disable them.

---

### ✅ 3. Prevent *ambiguous characters*

This is actually more important than OpenType features.

Characters users confuse:

* `O` vs `0`
* `l` vs `1` vs `I`
* `x` vs `×`

Best practice:

* Use a **restricted alphabet** (e.g. Base32, Base58)
* Or visually distinguish characters clearly

---

### ✅ 4. Consider copying vs typing

For private keys:

* Prefer **copy/paste UI**
* Add a **copy button**
* Show warnings if manually typed

---

## Best-practice “secure text” CSS

```css
.secure-text {
  font-family: monospace;
  font-variant-ligatures: none;
  font-feature-settings: "calt" 0, "liga" 0, "clig" 0, "dlig" 0, "frac" 0;
}
```

---

## Bottom line

* Your fix (`"calt" 0`) is **good but incomplete**
* For security-critical text, assume:

  > **Fonts are allowed to change what users see unless you explicitly prevent it**

The safest route is:

👉 **Monospace font + all ligatures/features off + ambiguity-aware design**

---

If you want, I can suggest a **“zero-risk” UI pattern for private keys** (including grouping, checksum, and visual validation) — that’s usually where real-world mistakes get eliminated.

thisconnect (changes requested):
please disable the other font "features" too.

jadzeidan:
@thisconnect please take another look

frontends/web/src/style/base.css
```diff
@@ -15,6 +15,8 @@ body {
     background-color: var(--background);
     color: var(--color-default);
     font-family: var(--font-family);
+    font-feature-settings: "calt" 0, "clig" 0, "dlig" 0, "frac" 0, "liga" 0;
+    font-variant-ligatures: none;
```
thisconnect, line 19: Defining on `html, body` here is good, but doesn't cast down to all elements such as: button, textarea, input, select.

That is why some css reset have font: inherit on such elements, like this reset: https://www.joshwcomeau.com/css/custom-css-reset/

frontends/web/src/components/copy/Copy.module.css
```diff
@@ -44,19 +44,18 @@
 
 .inputField {
     border: 1px solid var(--background-quaternary);
-    caret-color: transparent;
     flex: 1;
+    font-feature-settings: "calt" 0, "clig" 0, "dlig" 0, "frac" 0, "liga" 0;
```
thisconnect: This class is not on the `<input>` element, but on the parent. But unfortunatelly input element doesn't inherit, see my other comment.

In the transaction list it's fine, but in the tx-detail where it is an `<input>` element it still formats as multiplication sign.

Image

frontends/web/src/routes/account/addresses/addresses.module.css
```diff
@@ -135,7 +135,9 @@
 
 .rowAddress {
   font-family: var(--font-family-monospace);
+  font-feature-settings: "calt" 0, "clig" 0, "dlig" 0, "frac" 0, "liga" 0;
```
thisconnect: monospace should already be fine untested but according to gpt, and .rowAddress is used on `<span>` element which already inherit from `body, html`.

so you can remove here and in other places that use monospace.

frontends/web/src/style/base.css
thisconnect: I suggest in addition to the body, html here add:

```css
button, textarea, input, select {
    font-feature-settings: "calt" 0, "clig" 0, "dlig" 0, "frac" 0, "liga" 0;
    font-variant-ligatures: none;
}
```

or can try try with

```css
button, textarea, input, select {
    font-feature-settings: inherit; /* should work but I haven't tested */
    font-variant-ligatures: inherit;
}
```

and remove from all other files.

thisconnect (changes requested):

thisconnect (approved):
untested LGTM thank you

PR #4006, opened 2026-04-14
frontend: warn user if swap sell amount is higher than available

bznein:
Similarly to what we do in send, if an user selects an amount that is higher than the available, we show a warning and disable the swap button.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/market/swapkit/helpers.go
+++ b/backend/market/swapkit/helpers.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
+	accountErrors "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/errors"
 	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
 )
@@ -53,6 +55,18 @@ func FormatAmount(coin coinpkg.Coin, amount string) (string, error) {
 		return "0", nil
 	}
 	return formattedAmount, nil
+}
+
+// ValidateSwapSellAmount checks that sell amount fits into available balance.
+func ValidateSwapSellAmount(account accounts.Interface, sellAmount coinpkg.Amount) error {
+	balance, err := account.Balance()
+	if err != nil {
+		return err
+	}
+	if sellAmount.BigInt().Cmp(balance.Available().BigInt()) > 0 {
+		return errp.WithStack(accountErrors.ErrInsufficientFunds)
+	}
+	return nil
 }
 
 func newQuoteRequestFromCoinCodes(sellCoinCode, buyCoinCode, sellAmount string, providers []string) (*QuoteRequest, *APIError) {
--- a/frontends/web/src/routes/market/swap/swap.tsx
+++ b/frontends/web/src/routes/market/swap/swap.tsx
@@ -30,6 +30,7 @@ import { SubTitle } from '@/components/title';
 import { Guide } from '@/components/guide/guide';
 import { Entry } from '@/components/guide/entry';
 import { alertUser } from '@/components/alert/Alert';
+import { Message } from '@/components/message/message';
 import { Button, Label } from '@/components/forms';
 import { BackButton } from '@/components/backbutton/backbutton';
 import { Amount } from '@/components/amount/amount';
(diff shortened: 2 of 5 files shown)


frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -140,6 +141,22 @@ export const Swap = ({
     () => routes.find(route => route.routeId === selectedRouteId),
     [routes, selectedRouteId],
   );
+  const sellDisplayUnit = useMemo(
+    () => fromAccount
+      ? getDisplayedCoinUnit(fromAccount.coinCode, fromAccount.coinUnit, btcUnit)
+      : undefined,
+    [btcUnit, fromAccount],
+  );
+  const hasInsufficientFunds = useMemo(() => {
```
benma: In send-view the backend delivers the information if there is insufficient funds. Same should be done here I think, we try to avoid doing arithmetics and computations with amounts in the frontend. Isn't the backend already involved in delivering the info to the frontend, so it could deliver an error code at the same time? Backend has `errors.ErrInsufficientFunds`, which is used by all other integrations (btcdirect, pocket, etc).

Unless there is a good reason not to do that?

benma (changes requested):

frontends/web/src/routes/market/swap/swap.tsx
bznein: Yeah you are absolutely right. Now the error comes from the backend instead

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -445,6 +459,13 @@ export const Swap = ({
                 value={sellAmount}
                 onChangeValue={setSellAmount}
               />
+              <Message
+                hidden={quoteErrorCode !== 'insufficientFunds'}
+                type="warning"
+                className={style.sellWarning}
+              >
+                {[sellDisplayUnit, t('send.error.insufficientFunds')].filter(Boolean).join(': ')}
```
thisconnect: nit: 

```suggestion
                {sellDisplayUnit}
                {': '}
                {t('send.error.insufficientFunds')}
```

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -245,7 +256,10 @@ export const Swap = ({
         clearQuoteState(
           response.success
             ? 'no route found'
-            : response.errorMessage || 'Some unexpected error occurred.',
+            : response.errorCode === 'insufficientFunds'
+              ? undefined
+              : response.errorMessage || 'Some unexpected error occurred.',
```
thisconnect: nit: Such slightly hidden strings will need to be translated. It would be good to start translating everything so we don't miss this one.

thisconnect (approved):
untested LGTM with nit

frontends/web/src/routes/market/swap/swap.tsx
bznein: I created an issue specifically for translating. I;ll start taking care of it today :)

benma: We usually assume errorMessage is populated in case of error, so no need for a fallback? LLMs always try to make fallbacks in the frontend for stuff that does not happen in practice :sweat_smile:

backend/market/swapkit/helpers.go
```diff
@@ -54,6 +56,22 @@ func FormatAmount(coin coinpkg.Coin, amount string) (string, error) {
 	return formattedAmount, nil
 }
 
+// ValidateSwapSellAmount checks that sell amount is parseable and fits into available balance.
+func ValidateSwapSellAmount(account accounts.Interface, sellAmount string) (errp.ErrorCode, error) {
```
benma: I think this should take coinpkg.Amount as input for clarity - strings always make me anxious and makes me wonder if it works in sats-mode, or what the string might contain. The callsite can do the parsing, as it already handles the string.

backend/market/swapkit/helpers.go
```diff
@@ -54,6 +56,22 @@ func FormatAmount(coin coinpkg.Coin, amount string) (string, error) {
...
+	parsedAmount, err := account.Coin().ParseAmount(sellAmount)
+	if err != nil {
+		return errp.ErrorCode(accountErrors.ErrInvalidAmount), accountErrors.ErrInvalidAmount
+	}
+	balance, err := account.Balance()
+	if err != nil {
+		return ErrInvalidRequest, err
+	}
+	if parsedAmount.BigInt().Cmp(balance.Available().BigInt()) > 0 {
+		return errp.ErrorCode(accountErrors.ErrInsufficientFunds), accountErrors.ErrInsufficientFunds
```
benma: It's very unusual to return two values for an error like this. 

 accountErrors.ErrInsufficientFunds is of type TxValidationError, so you can handle it the same way as the send screens do, and convert this to an error code in the handlers like here:

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/backend/coins/btc/handlers/handlers.go#L516

benma (changes requested):

benma (approved):
utACK

PR #4009, opened 2026-04-15
frontend: disable same-coin swap options

bznein:
On setting/chaning either buy or sell asset, disable the corresponding same coin on the other selector. This is done since it makes no sense to try and swapping BTC to BTC, for example.

@thisconnect I tried to give it a go here; I believe this is something that can be done entirely in the frontend, and it seems a fairly quick change, that to me looks ok. Let me know if you think there are better ways to achieve this; the only thing I can't really follow very clearly is the `isDisabledOption` part, to be honest :)

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/market/swap/swap.tsx
+++ b/frontends/web/src/routes/market/swap/swap.tsx
@@ -137,7 +137,12 @@ export const Swap = ({
     [routes, selectedRouteId],
   );
 
-  // initialize swap account selections once the swap accounts payload has loaded.
+  const isSameCoinAccount = (
+    candidate: TSwapAccount,
+    oppositeAccount: TSwapAccount | undefined,
+  ) => candidate.coinCode === oppositeAccount?.coinCode;
+
+  // Keeps the selected swap accounts aligned with the latest available options and redirects away if no swap is possible.
   useEffect(() => {
     if (!swapAccounts || !swapAccounts.success) {
       return;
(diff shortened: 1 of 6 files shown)


frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -137,6 +137,11 @@ export const Swap = ({
     [routes, selectedRouteId],
   );
 
+  const isSameCoinAccount = (
+    candidate: TSwapAccount,
+    oppositeAccount: TSwapAccount | undefined,
+  ) => candidate.coinCode === oppositeAccount?.coinCode;
+
   useEffect(() => {
```
thisconnect, line 146: since you are here, please add a short comment what the purpose of the effect is.

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -154,13 +159,20 @@ export const Swap = ({
           : swapAccounts.sellAccounts[0]?.code,
       );
     }
-    if (!swapAccounts.buyAccounts.some(account => account.code === buyAccountCode)) {
+    const selectedSellAccount = swapAccounts.sellAccounts.find(account => account.code === sellAccountCode);
```
thisconnect, line 163: Later down I think the code could be simplified if it first finds the selected buy account. But..

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -154,13 +159,20 @@ export const Swap = ({
...
+    if (
+      !swapAccounts.buyAccounts.some(account => account.code === buyAccountCode)
+      || swapAccounts.buyAccounts.some(
+        account => account.code === buyAccountCode && isSameCoinAccount(account, selectedSellAccount),
+      )
+    ) {
       setBuyAccountCode(
```
thisconnect, line 165: In the case the buy-account-coincode as sell-account-coincode are same it currently tries to change the buy-account-code to the first account with another coincode. 

> On setting/chaning either buy or sell asset, disable the corresponding
same coin on the other selector. 

If all options with same coincode are disabled we don't need to change the buy-account-code here, as it is not possible to pick the dropdown.

thisconnect (changes requested):

frontends/web/src/routes/market/swap/swap.tsx
bznein: If I understood it correctly, it should be fine now :)

frontends/web/src/routes/market/swap/swap.tsx
bznein: Not sure I understand this :) Does this still apply after the latest change?

frontends/web/src/routes/market/swap/swap.tsx
thisconnect: The many .find and .some in conditions confused me a bit.

frontends/web/src/routes/market/swap/swap.tsx
thisconnect: When this effect is supposed to run did confuse me. Also in the previous version. swapAccounts could in theory change at any time, if an account is discovered (maybe it took a long time)
All good, I'll resolve.

thisconnect (approved):
thanks tested LGTM

PR #4010, opened 2026-04-15
frontend: clear quote state when flipping.

bznein:
When flipping sell and buy asset, the receive amount shown was not correct. We need to wait for a proper quote to be received, and in the meantime disable the swap button so that users can't proceed to an invalid state.

- [ ] updating the Changelog
- [ ] writing unit tests


thisconnect (approved):
untested LGTM

PR #4013, opened 2026-04-16
frontend: miscellaneous improvements & minor bugfixes

shonsirsha:
Miscellaneous improvements & minor bugfixes on the frontend.

👾 = bug
⚡️ = improvement

👾 Only certain area in the total asset (in portfolio) are clickable. Need to allow the whole row to be clickable.
⚡️ Smaller unit font size can be used
⚡️ No feedback when tapping interactive items (such as in accounts, settings, tx list etc)
⚡️ On mobile, we should scroll to the tx list search bar. User needs to scroll manually atm.
👾 Cog and shield icon uses grey(?)ish colour on "More" settings. On sidebar too they're grey. Should use white/black (theme-aware) icons.
⚡️ ios: add haptic feedback on copying and on changing chart filters
👾 ios: blue highlight showing on QR-code tap-to-copy
👾 on small screens, total asset balances (portfolio) may get truncated
👾 hover color "stuck" for mobile fullscreen selector
⚡️ subtle tab navigation transition on mobile


frontends/web/src/components/token-list-item/token-list-item.tsx
```diff
@@ -31,13 +31,14 @@ export const TokenListItem = ({
         ${!active ? style.tokenInactive || '' : ''}
         ${className || ''}
       `}
-      style={lineColor ? { '--token-line-color': lineColor } : undefined}>
+      onClick={onClick}
```
sutterseba: regression: toggling token switches to off in manage accounts now also triggers navigation (and the account stays active)

sutterseba (changes requested):

frontends/web/src/components/qrcode/qrcode.tsx
```diff
@@ -60,18 +60,16 @@ const TapToCopyQRCode = ({ data, qrCodeData, size }: TTapToCopyQRCodeProps) => {
     }
   }, [success]);
 
-
   const handleCopy = () => {
-    inputRef.current?.select();
-    if (document.execCommand('copy')) {
+    if (data) {
+      navigator.clipboard.writeText(data);
```
thisconnect, line 65: If I remember correctly, navigator api doesn't always work i.e. in Qt it may be not available.

I believe we had to make the workaround with hidden input field and execCommand('copy').

If you want to change it please test that copy still works on all platforms.

frontends/web/src/routes/account/summary/asset-balance-with-unit-price.tsx
thisconnect: `style.unit` was removed in the css file but still used here.

frontends/web/src/routes/account/summary/accountssummary.module.css
```diff
@@ -161,18 +165,13 @@
...
     white-space: nowrap;
 }
 
 .unitPrice {
     color: var(--color-secondary);
 }
 
-.unit {
-    font-size: var(--size-smaller);
-}
```
thisconnect, line 174: `.unit` was removed here but still used in the component, commented on the asset-balance-with-unit-price.tsx.

As this is hard to spot in the future I think we shouldn't share .css files between components. But ofc here it is comfortable as we reuse those styles. But it is an indication to write re-useable

frontends/web/src/components/qrcode/qrcode.tsx
shonsirsha: Mh.. yeah thanks for raising this.

However, we're currently already using `navigator.clipboard.writeText` in our Copy component, also maybe this https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/frontends/qt/main.cpp#L429 answers it (?)

shonsirsha:
Thank you @sutterseba  @thisconnect!

Addressed your comments 🙏 PTAL

frontends/web/src/components/dropdown/mobile-fullscreen-selector.module.css
```diff
@@ -130,7 +130,7 @@ input.searchInput:focus {
     width: 100%;
 }
 
-.optionItem:hover {
+.optionItem:active {
     background-color: var(--background-custom-select-hover);
```
thisconnect: Q: Just looking at the code, shouldn't `--color-tap-active` variable be named `--background-tap-active`?

also should `--color-tap-active` and `--background-custom-select-hover` be same or different color?

frontends/web/src/components/qrcode/qrcode.tsx
thisconnect: Nice, I thought there was something and was searching for that commit but couldn't find it before. https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/36e315faa61505f97c38c4cc45c9fb47c6cc96ec

That commit enabled Qt and Android and I guess it works in iOS?

In that case it's all good.

frontends/web/src/routes/account/summary/asset-balance-with-line.tsx
```diff
@@ -52,7 +52,12 @@ export const AssetBalanceWithLine = ({ account, coinCode, isFirst, balances }: P
         lineColor={lineColor}
         onClick={() => navigate(`/account/${account.code}`)}
       >
-        <TokenAmounts balance={balance?.available} />
+        <div
+          className={style.clickable}
+          onClick={() => navigate(`/account/${account.code}`)}
```
thisconnect: Just looking at the code, the parent component (TokenListItem) has the same onClick, are both needed or can one be removed?

thisconnect (approved):
untested LGTM with css variable naming question

frontends/web/src/components/dropdown/mobile-fullscreen-selector.module.css
shonsirsha: Good idea. Also we can unify  the two variables --color-tap-active and --background-custom-select-hover. The former fits the latter just fine.

frontends/web/src/routes/account/summary/asset-balance-with-line.tsx
shonsirsha: Mm yeah can remove with a bit of refactor!:)

sutterseba (approved):
tested on iPhone LGTM

PR #4018, opened 2026-04-20
frontend: lightning activation wait screen

Beerosagos:
Activating a lightning account (especially a previously activated one) could take several seconds. This adds a waiting screen to clearly indicate that the wallet is being created.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/lightning/activate.tsx
+++ b/frontends/web/src/routes/lightning/activate.tsx
@@ -3,7 +3,9 @@
 import { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
+import { Logo } from '@/components/icon/logo';
 import { UseDisableBackButton } from '@/hooks/backbutton';
+import { useLightning } from '@/hooks/lightning';
 import { Header, Main } from '../../components/layout';
 import { View, ViewButtons, ViewContent, ViewHeader } from '../../components/view/view';
 import { MultilineMarkup, SimpleMarkup } from '../../utils/markup';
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/lightning/activate.tsx
```diff
@@ -41,31 +45,45 @@ export const LightningActivate = () => {
...
     if (step === 'connect' && keystores && keystores.length > 0) {
-      activateNode();
+      setStep('confirm');
     }
-  }, [keystores, step, activateNode]);
+  }, [keystores, step]);
+
+  useEffect(() => {
+    if (step === 'confirm' && !activationStarted) {
+      void activateLightning();
```
thisconnect: to me it looks like void doesn't do anything.

frontends/web/src/routes/lightning/activate.tsx
```diff
@@ -41,31 +45,45 @@ export const LightningActivate = () => {
...
-    setStep('wait');
+  const activateLightning = useCallback(async () => {
+    setActivationStarted(true);
+    setSetupError(undefined);
 
     try {
       await postActivate();
       setStep('success');
     } catch (err) {
+      setActivationStarted(false);
```
thisconnect, line 56: this only sets activation to false in case of an error. Shouldn't it always set it to falce after postActive finished?

maybe add `finally` ?

thisconnect (approved):
code LGTM with small nit (but untested so not 100% sure)

frontends/web/src/routes/lightning/activate.tsx
Beerosagos: Nice catch :pray:

frontends/web/src/routes/lightning/activate.tsx
Beerosagos: Activation needs to be started only once (unless it errors). I don't think we need to reset it to false, unless we want to rename it to e.g. `ActivationInProgress`. But I don't think it changes much. No strong opinion, tho: I can do it, if you prefer.

thisconnect: I see makes sense, keep as is. 👍

PR #4020, opened 2026-04-20
backend/bitbox02: bundle v9.26.1 firmwares

benma:
Using update.py.

See also: https://github.com/BitBoxSwiss/bitbox02-firmware/pull/1940


bznein (approved):
utACK 

(we also need to remember to add it here https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/backend/devices/bitbox02/testdata/simulators.json)

PR #4024, opened 2026-04-20
backend: support LTC as swap coin.

bznein:
LTC is handled internally in the same way as BTC, so we simply add it to the list of supported coins.

- [ ] updating the Changelog
- [ ] writing unit tests


benma (approved):
utACK

PR #4032, opened 2026-04-21
frontend: align sidebar icons

Beerosagos:
Lightning icon centering in the sidebar was wrong. This fixes it using flex centering to bring it onto the same vertical alignment as the other sidebar icons.

Before:
image

After:
image

### DIFF
--- a/frontends/web/src/components/sidebar/sidebar.module.css
+++ b/frontends/web/src/components/sidebar/sidebar.module.css
@@ -139,9 +139,13 @@
 
 .sidebarItem :global(.stacked),
 .sidebarItem .single {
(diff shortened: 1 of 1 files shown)


thisconnect:
@Beerosagos the icon is huge, could you limit the height so it doesn't look 2x bigger than the other icons?

Beerosagos:
> @Beerosagos the icon is huge, could you limit the height so it doesn't look 2x bigger than the other icons?

Thanks, now it should be better. I updated the preview in the PR description

thisconnect (approved):
utnested LGTM thanks

PR #4034, opened 2026-04-21
Frontend swap disclaimer

thisconnect:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -960,6 +960,36 @@
       }
     },
     "swapkit": {
+      "disclaimer": {
+        "dataProtection": {
+          "link": "SwapKit privacy policy",
+          "text": "The BitBoxApp does not learn or collect any personal data. SwapKit may collect some data as explained in their Privacy Policy.",
+          "title": "Data Protection"
+        },
+        "fees": {
+          "note": "Note: SwapKit's exchange rates may differ from the BitBoxApp rates.",
+          "text": "A flat rate of 1.5% is applied to each swap",
+          "title": "Fees"
+        },
+        "intro": {
+          "text": "SwapKit offers an SDK/API service for non-custodial swaps. We work with SwapKit to provide cryptocurrency swapping across different blockchains in one place. It helps power fast, simple, and non-custodial trading, so you stay in control of your assets.",
+          "title": "Swapping cryptocurrency with SwapKit"
+        },
+        "protocol": {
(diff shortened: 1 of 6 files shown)


frontends/web/src/locales/en/app.json
```diff
@@ -958,6 +958,41 @@
...
+          "text": "A flat rate of 1.5% is applied to each swap",
+          "title": "Fees"
+        },
+        "intro": {
+          "text": "SwapKit offers an SDK/API service for non-custodial swaps. We work with SwapKit to provide cryptocurrency swapping across different blockchains in one place. It helps power fast, simple, and non-custodial trading, so you stay in control of your assets.",
+          "title": "Swapping cryptocurrency with SwapKit"
+        },
+        "protocol": {
+          "link": "Learn about NEAR Intents",
+          "text": "The swapping protocol, provided by SwapKit API, currently used in the BitBoxApp is NEAR Intents.",
```
bznein: If we want to enable more providers without code changes, would that be an issue since we specifically only mention NEAR here?

frontends/web/src/locales/en/app.json
```diff
@@ -958,6 +958,41 @@
...
+          "title": "Protocol"
+        },
+        "security": {
+          "link": "BitBox security threat model",
+          "text": "SwapKit is an external service and is not covered by the BitBox security threat model. However, we closely work with SwapKit to improve security by implementing payment requests into the BitBox firmware. This means the BitBox will display the destination address as “SwapKit” when making a swap.",
```
bznein: probably not relevant, but the firmware actually show "SWAPKIT" (all caps)

bznein (commented):
overall LGTM and tested, I only have one nit and a question about adding more providers

thisconnect:
rebased

frontends/web/src/locales/en/app.json
jadzeidan: Don't think it matters. All caps is consistent in fw with POCKET. But official branding  is "SwapKit". I think it is fine to keep as is.

thisconnect: I slightly prefer the disclaimer to reflect what the FW will display in the end (uppercase).

thisconnect: We also write "Pocket" in pocket sell disclaimer. 🤦

bznein (approved):
LGTM

PR #4039, opened 2026-04-22
Address coderabbit findings on staging-swap

bznein:
4 commits: the first three are quite trivial (a markdown lint, wrapping of error, nil handling in a test_

The last one is a bit more and stems from an error we made when adding PocketOTC.

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos (approved):
utACK

PR #4046, opened 2026-04-23
frontend: fix isMounted.current

thisconnect:
useMountedRef() returns a ref. !isMounted is always false, so an in-flight conversion can still call setEstimatedFiatValue() after unmount.

- [ ] updating the Changelog
- [ ] writing unit tests


bznein (approved):
utACK

PR #4048, opened 2026-04-23
frontend: disable swap when changing amount

bznein:
When editing the amount to sell, the Swap button needs to be disabled, and we need to also clear the dropdown menu since the route is not valid anymore.

- [ ] updating the Changelog
- [ ] writing unit tests


thisconnect (approved):
tested LGTM

PR #4050, opened 2026-04-23
frontend: add support for native macos 26 tahoe app icons

sutterseba:
Ensures the macOS app icon stays consistent when dark, clear or tinted icons are used, introduced in macOS 26 Tahoe.

Declares the asset-catalog icon name in the macOS bundle plists while keeping the .icns fallback for older macOS versions.

The new icon was created and configured with [Apple Icon Composer](https://developer.apple.com/icon-composer/).


benma (approved):
tACK

PR #4064, opened 2026-04-27
swap: return better error message for no routes found

bznein:
rather than verbating showing the error message, return a structured error from the backend and show a localized error in the frontend.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/market/swapkit/helpers.go
+++ b/backend/market/swapkit/helpers.go
@@ -13,13 +13,12 @@ import (
 )
 
 var swapkitAssetByCoinCode = map[string]string{
-	"btc":    "BTC.BTC",
-	"tbtc":   "BTC.BTC",
-	"rbtc":   "BTC.BTC",
-	"ltc":    "LTC.LTC",
-	"eth":    "ETH.ETH",
-	"sepeth": "ETH.ETH",
-
+	"btc":                 "BTC.BTC",
+	"tbtc":                "BTC.BTC",
+	"rbtc":                "BTC.BTC",
+	"ltc":                 "LTC.LTC",
+	"eth":                 "ETH.ETH",
+	"sepeth":              "ETH.ETH",
 	"eth-erc20-usdt":      "ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7",
 	"eth-erc20-usdc":      "ETH.USDC-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
 	"eth-erc20-link":      "ETH.LINK-0x514910771af9ca656af840dff83e8264ecf986ca",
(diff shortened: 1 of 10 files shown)


frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -273,13 +274,30 @@ export const Swap = ({
...
+            })
+            : t('swap.noRouteFound');
+          clearQuoteState(noRouteFoundMessage, response.success ? undefined : response.errorCode);
+          return;
+        }
+        clearQuoteState(
+          response.errorCode === 'unexpectedError'
+            ? t('swap.fetchQuotesError')
+            : response.errorMessage || t('swap.fetchQuotesError'),
+          response.errorCode,
```
thisconnect: frontend lgtm, but this part is slightly on the hard to read side :)

I notice that I'd expect clearQuoteState to just clear, but it take 2 arguments and updates the state.

lets to a coderabbyai review and see what it suggests, wdyt?

bznein:
@coderabbit review

frontends/web/src/routes/market/swap/swap.tsx
bznein: I wonder if coderabbit is able to read comments and comment based on those...

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -55,6 +55,7 @@ type Props = {
 };
 
 const QUOTE_DEBOUNCE_MS = 300;
+const NO_ROUTES_FOUND_ERROR = 'NoRoutesFoundError';
```
thisconnect, line 59: Would it make sense to add `const INSUFFICIENT_FUNDS_ERROR = 'insufficientFunds';` etc here too?

frontends/web/src/routes/market/swap/swap.tsx
bznein: Tried to push a new commit that clarifies the code a bit, lmk what you think!

thisconnect (approved):
frontend LGTM

frontends/web/src/routes/market/swap/swap.tsx
bznein: done

backend/market/swapkit/helpers.go
```diff
@@ -144,6 +199,13 @@ func newSwapRequestFromCoinCodes(
 
 // NewQuoteFromCoinCode validates the provided coin codes, fetches a quote, and maps structured API errors.
 func NewQuoteFromCoinCode(ctx context.Context, httpClient *http.Client, sellCoinCode, buyCoinCode, sellAmount string) (*QuoteResponse, *APIError) {
+	quoteErrorData := &APIErrorData{}
+	if sellAssetMetadata, ok := assetMetadataFromCoinCode(sellCoinCode); ok {
```
Beerosagos: Instead of changing the info in the helper to add the Coin code, you could fetch the actual Coin object from the backend in the handler and pass it to `NewQuoteFromCoinCode`. Then you could call `buyCoin.Name()`, (or .Unit, or .Code, not sure which one is the best).

Beerosagos (approved):
backend untested LGTM, with an improvement suggestion

frontends/web/src/routes/market/swap/swap.tsx
thisconnect: yes it's much clearer now with resetQuoteStateWithError, thank you.

PR #4066, opened 2026-04-27
swap: show better errors.

bznein:
Instead of forwarding verbatim the errors syncInprogress and accountNotSynced, we pass an appropriate errorCode to the frontend and show a localized string.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -349,6 +349,14 @@ export type TTxInput = {
   }
 );
 
+export type TTxProposalErrorCode =
+  | 'accountNotSynced'
+  | 'feeTooLow'
+  | 'feesNotAvailable'
+  | 'insufficientFunds'
+  | 'invalidAddress'
+  | 'invalidAmount';
+
 export type TTxProposalResult = {
   amount: TAmountWithConversions;
   fee: TAmountWithConversions;
@@ -356,7 +364,7 @@ export type TTxProposalResult = {
   success: true;
   total: TAmountWithConversions;
 } | {
-  errorCode: string;
+  errorCode: TTxProposalErrorCode;
   success: false;
 };
 
@@ -367,6 +375,12 @@ export const proposeTx = (
   return apiPost(`account/${accountCode}/tx-proposal`, txInput);
 };
 
+export type TSendTxErrorCode =
+  | TTxProposalErrorCode
+  | 'erc20InsufficientGasFunds'
+  | 'syncInProgress'
+  | 'wrongKeystore';
+
 export type TSendTx = {
   success: true;
   txId: string;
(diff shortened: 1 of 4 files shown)


bznein:
@coderabbit review

frontends/web/src/routes/market/swap/components/swap-result.tsx
```diff
@@ -78,6 +78,27 @@ export const SwapResult = ({
       );
     }
 
+    if (result.errorCode) {
```
thisconnect, line 81: currently TSendTx.errorCode is typed as string could you update the types to add the known eror codes?

thisconnect (changes requested):

frontends/web/src/routes/market/swap/components/swap-result.tsx
bznein: done!

frontends/web/src/api/account.ts
```diff
@@ -349,14 +349,22 @@ export type TTxInput = {
   }
 );
 
+export type TTxProposalErrorCode =
+  | 'accountNotSynced'
+  | 'feeTooLow'
+  | 'feesNotAvailable'
+  | 'insufficientFunds'
+  | 'invalidAddress'
+  | 'invalidAmount';
```
thisconnect, line 358: nice, thank you

frontends/web/src/routes/market/swap/components/swap-result.tsx
```diff
@@ -78,6 +78,27 @@ export const SwapResult = ({
...
+      return (
+        <View fullscreen textCenter verticallyCentered width="520px">
+          <ViewHeader />
+          <ViewContent withIcon="error">
+            <p>
+              {t(`send.error.${result.errorCode}`)}
```
thisconnect: If I read correctly, `result.errorCode` is of type `TSendTxErrorCode` where it can have `wrongKeystore`.

But the translated errors in send.error don't have `wrongKeystore` translation, that one is under `error.wrongKeystore'`.  Could you check if I read correctly and if there are other keys missing?

I'm not sure what to suggest, should we move all errors to more generic `error.` instead of `send.error.` ?

thisconnect (changes requested):
I think there is a bug under which key the translated errors are available.

frontends/web/src/routes/market/swap/components/swap-result.tsx
bznein: That's a good point and you are right!

I'm not sure moving everything is a good idea. If the issue is contained, I think it would be better in this case just to handle wrongKeystore separately...

bznein: (I did that, but happy to revert if needed)

thisconnect (approved):
untested LGTM

bznein:
@strmci do you want to test it before I merge?

strmci:
@bznein tested, LGTM

strmci (approved):

PR #4067, opened 2026-04-27
convert some handlers to getAPIRouterNoError

benma:


frontends/web/src/utils/config.ts
```diff
@@ -32,7 +40,10 @@ export const setConfig = (object: TConfig) => {
       });
       pendingConfig = nextConfig;
       return apiPost('config', nextConfig)
-        .then(() => {
+        .then((response: TSetConfigResponse) => {
+          if (!response.success && !runningInQtWebEngine() && !runningOnMobile()) {
+            throw new Error(response.errorMessage || 'Failed to update configuration');
+          }
           pendingConfig = {};
           return nextConfig;
```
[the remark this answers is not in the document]
benma: This keeps the previous behavior - propagating the error and handling it in all callsites is a much bigger task that we won't do in this PR.

backend/handlers/handlers.go
```diff
@@ -541,12 +541,20 @@ func (handlers *Handlers) getDefaultConfig(*http.Request) interface{} {
...
 	appConfig := config.AppConfig{}
 	if err := json.NewDecoder(r.Body).Decode(&appConfig); err != nil {
-		return nil, errp.WithStack(err)
+		return response{Success: false, ErrorMessage: err.Error()}
+	}
+	if err := handlers.backend.Config().SetAppConfig(appConfig); err != nil {
+		return response{Success: false, ErrorMessage: err.Error()}
 	}
-	return nil, handlers.backend.Config().SetAppConfig(appConfig)
+	return response{Success: true}
```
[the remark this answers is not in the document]
benma: Fixed

frontends/web/src/utils/config.ts
```diff
@@ -1,12 +1,20 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { apiGet, apiPost } from '@/utils/request';
+import { runningInQtWebEngine, runningOnMobile } from './env';
```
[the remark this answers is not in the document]
benma: Fixed

frontends/web/src/utils/config.ts
```diff
@@ -32,7 +40,10 @@ export const setConfig = (object: TConfig) => {
       });
       pendingConfig = nextConfig;
       return apiPost('config', nextConfig)
-        .then(() => {
+        .then((response: TSetConfigResponse) => {
+          if (response?.success === false && !runningInQtWebEngine() && !runningOnMobile()) {
+            throw new Error(response.errorMessage || 'Failed to update configuration');
+          }
```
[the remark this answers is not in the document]
benma: it's in devmode only

frontends/web/src/api/system.ts
```diff
@@ -6,6 +6,13 @@ export const notifyUser = (text: string) => {
   return apiPost('notify-user', { text });
 };
 
-export const open = (href: string) => {
+type TOpenResponse = {
+  success: true;
+} | {
+  success: false;
+  errorMessage?: string;
```
thisconnect: Many types use SuccessResponse and FailResponse, but this has slightly different properties. Instead of "errorMessage" it has just "message". Therer is a comment that this uses `// if the backend uses maybeBB02Err`.

```ts
export type SuccessResponse = {
  success: true;
};

// if the backend uses maybeBB02Err
export type FailResponse = {
  code?: number;
  message?: string;
  success: false;
};
```

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/master/frontends/web/src/api/response.ts

I have no strong opinion on which one we should prefer, but it would be nice if we could just use those everywhere.

```ts
type TOpenResponse = SuccessResponse | FailResponse;
```

Would it make sense to refactor maybeBB02Err to use errorMessage etc?

(also some endpoints return "errMsg")

frontends/web/src/components/anchor/anchor.tsx
```diff
@@ -41,7 +45,15 @@ export const A = ({
       title={props.title || href}
       onClick={(e: SyntheticEvent) => {
         e.preventDefault();
-        open(href).catch(console.error);
+        open(href)
+          .then(response => {
+            if (!response.success) {
+              alertUser(response.errorMessage
```
thisconnect, line 51: At some point we should stop using `alertUser`, this function was always a bit a hack but very convenient. Better would be to have an error state and render the error in the component.

benma: Can there be a solution that is as easy to use as alertUser? Making ad-hoc ways to display errors everywhere is difficult (for me).

Maybe you could make a PR sometime that removes all uses of alertUser and replaces it with a streamlined way of showing such messages?

frontends/web/src/api/system.ts
benma: Almost all endpoints outside use errorMessage I think. Makes sense to me to make BitBox02 handlers use the same, but that's for another PR, as it is unrelated to this one.

backend/handlers/handlers.go
```diff
@@ -374,6 +374,14 @@ func writeJSON(w io.Writer, value interface{}) {
...
 }
 
+func logHandlerError(handler string, err error) {
+	logging.Get().
+		WithGroup("handlers").
+		WithField("handler", handler).
+		WithError(err).
+		Error("handler failed")
+}
+
```
bznein: Wouldn't it be better to have a logger inside the `Handlers` type instead? Seems more in line with what we usually do

backend/handlers/handlers.go
```diff
@@ -567,12 +585,22 @@ func (handlers *Handlers) postNotify(r *http.Request) (interface{}, error) {
 	return nil, nil
 }
 
-func (handlers *Handlers) postOpen(r *http.Request) (interface{}, error) {
+func (handlers *Handlers) postOpen(r *http.Request) interface{} {
+	type response struct {
+		Success      bool   `json:"success"`
+		ErrorMessage string `json:"errorMessage,omitempty"`
```
bznein, line 591: [nit] I'm almost thinking -but not sure- we could have a base response type like this one, and then most of the others would just either use this directly, or extend it with other fields

bznein (changes requested):

bznein (approved):

frontends/web/src/api/system.ts
thisconnect: Ok then we should change  FailResponse to use errorMessage in the future and use and extend from SuccessResponse FailResponse everywhere.

Should we rename SuccessResponse to SuccessResponseLegacy and already start with the new one?

But I'd love if in the end all use the same and we dont end up having a mix again.

frontends/web/src/api/market.ts
```diff
@@ -48,8 +48,12 @@ export const getMarketDeals = (
 };
 
 export type MoonpayBuyInfo = {
+  success: true;
   url: string;
   address: string;
+} | {
+  success: false;
+  errorMessage?: string;
 };
```
thisconnect, line 57: fine for now, but we should try to do something like

SuccessResponse & {
  url: string;
  address: string;
} | FailResponse;

frontends/web/src/components/anchor/anchor.tsx
thisconnect: This is pretty easy have a look: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4082/changes.

But I agree it would be nice to have a good reusable way.

In the future we may want to have a global notification center / error component. 🤷

frontends/web/src/utils/config.ts
```diff
@@ -1,12 +1,20 @@
...
 
 import { apiGet, apiPost } from '@/utils/request';
+import { runningInQtWebEngine, runningOnMobile } from '@/utils/env';
 
 type TConfig = {
   backend?: unknown;
   frontend?: unknown;
 };
 
+type TSetConfigResponse = null | undefined | {
```
thisconnect: Is this correct that it can return null | undefined ?

thisconnect (approved):
LGTM

'd love if we have those messages and codes  consistently in all apis' in the futurte and can do `TSuccessResponse & {custom: true} | TFailResponse`.

frontends/web/src/components/anchor/anchor.tsx
benma: The issue with that one is that it's a lot more effort - need to think where to place the error message, test if it looks good, etc. alertUser "just works" - would be nice to have something that is as easy to use.

thisconnect: it's harder yes but much nicer UI than popup. we also have popup over popup, somewhere in one case 3 popups over each other each with different dimension and visible. I'll find a screenshot once.

also from our friend gpt:

> You’re right to call this a “sin” 😄 — a global mutable function that triggers UI from anywhere (including utilities) breaks React’s core model (unidirectional data flow + explicit state). Refactoring it is less about removing alertUser and more about restoring architectural boundaries.

thisconnect: Keep as is for this PR, but if you don't mind maybe well change it for smth better and remove alertUser ok?

frontends/web/src/utils/config.ts
benma: Good catch, it can't return either of these. Removed.

PR #4068, opened 2026-04-28
frontend: misc. RC fixes

shonsirsha:
1. [frontend: use SettingsItem for items in more.tsx](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4068/changes/08d9b9c530e3c5d77a2b098e4e2963937432c198)

2. [frontend: fix Backup dialog not having bottom border radius](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4068/changes/5d2c6553d51659d22e659621e86ea8772a3f8c42)

3. [frontend: change InputWithDropdown from disabled to readOnly](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4068/changes/f3b4c7ac76562cfed3696c7d7d601fc8c1857878)

### DIFF
--- a/frontends/web/src/components/dialog/dialog.module.css
+++ b/frontends/web/src/components/dialog/dialog.module.css
@@ -27,6 +27,7 @@
     flex-direction: column;
     max-height: calc(100dvh - calc(env(safe-area-inset-top) * 2));
     max-width: 420px;
+    overflow: clip;
     text-align: left;
     touch-action: auto;
     -webkit-overflow-scrolling: touch;
@@ -189,6 +190,8 @@
     }
 
     .modal {
+        border-bottom-left-radius: 0;
+        border-bottom-right-radius: 0;
         padding: 0;
         height: auto;
         max-height: 100dvh;
--- a/frontends/web/src/components/forms/input-with-dropdown.module.css
+++ b/frontends/web/src/components/forms/input-with-dropdown.module.css
@@ -20,23 +20,12 @@
     justify-self: flex-start;
 }
 
-.align-right label {
-    flex-grow: 1;
-    text-align: right;
+.errorText {
+    color: var(--color-error) !important;
 }
 
 .align-right input {
     text-align: right;
-}
-
-.isTransparent input[disabled] {
-    border-color: var(--background);
-    color: var(--color-default) !important;
-    cursor: default;
-}
-
-.errorText {
-    color: var(--color-error) !important;
 }
 
 .inputDropdownWrapper {
@@ -69,12 +58,13 @@ input.inputField {
     will-change: border-color;
 }
 
-input.inputField:disabled {
+input.inputField[readonly] {
     border-color: var(--background-quaternary) !important;
     color: var(--color-secondary);
-    cursor: not-allowed;
-    opacity: 1; 
-    -webkit-text-fill-color: var(--color-secondary); 
+    cursor: text;
+    opacity: 1;
+    text-overflow: clip;
+    -webkit-text-fill-color: var(--color-secondary);
 }
 
 input.inputField:focus {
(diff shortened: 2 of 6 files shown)


frontends/web/src/components/dialog/dialog.module.css
```diff
@@ -27,6 +27,7 @@
     flex-direction: column;
     max-height: calc(100dvh - calc(env(safe-area-inset-top) * 2));
     max-width: 420px;
+    overflow: hidden;
```
thisconnect: I was just looking into this bug too. This will add round corners on mobile too, but I suggest to addd `border-radius: var(--radius-xl) var(--radius-xl) 0 0;` on small screen, so it is explicitly setting the bottom raidus to 0.

small nit: iirc we should use `overflow: clip` instead of hidden whenever possible.

frontends/web/src/routes/settings/more.module.css
```diff
@@ -20,4 +20,5 @@
 .iconContainer {
     display: flex;
     align-items: center;
+    gap: var(--space-quarter);
```
thisconnect: I had to checkout and test to see what the problem is and what this fixes (missing commit message).

This moves the red dot closer to the carret

Image

But the same reddot is not moved in settings here

Image

and here

Image


I think it was better before. GH PR is not great to discuss design changes lets dm

To me this reddot should be conceptually more aligned with content we have on the right hand  side.



Image

thisconnect (changes requested):

frontends/web/src/routes/settings/more.module.css
shonsirsha: The changes were made so that both reddot show at the same spot, not close to the caret. This fixes it at least on my end. Your first screenshot is how it looks on current target branch (the bug we're trying to fix).

Anyhow, it was just a quick fix. I notice on more.tsx we're not using SettingsItem yet. We should do that anyway, and the red-dot positioning bug there will be fixed as a side effect.

thisconnect: I missed that sorry. your are right.

frontends/web/src/components/dialog/dialog.module.css
thisconnect: > small nit: iirc we should use overflow: clip instead of hidden whenever possible.

This is not related to the border-radius, just a small nit.

shonsirsha:
@coderabbitai review

frontends/web/src/components/forms/input-with-dropdown.module.css
```diff
@@ -20,21 +20,6 @@
...
 }
 
-.align-right label {
-    flex-grow: 1;
-    text-align: right;
-}
-
-.align-right input {
-    text-align: right;
-}
```
thisconnect, line 30: I think those styles might be accidentally removed?

https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/98bc94553c8856575f8e435e4fd6881b0449a657/frontends/web/src/components/forms/input-with-dropdown.tsx#L43

frontends/web/src/components/forms/input-with-dropdown.tsx
```diff
@@ -7,7 +7,7 @@ import { Dropdown, TOption } from '@/components/dropdown/dropdown';
 import { ChevronDownDark } from '@/components/icon';
 import styles from './input-with-dropdown.module.css';
 
-export type TInputWithDropdownProps<T> = TBaseInputProps & {
+export type TInputWithDropdownProps<T> = Omit<TBaseInputProps, 'transparent'> & {
```
thisconnect, line 10: since transparent prop was removed in this commit you can also drop it from here.

thisconnect (approved):
partly tested LGTM but please keep the align styles

frontends/web/src/components/forms/input-with-dropdown.tsx
shonsirsha: Here I am actually omitting the transaprent prop from the TBaseInputProps here, so we don't accidentally put in transparent here - since we have removed it.

frontends/web/src/components/forms/input-with-dropdown.module.css
shonsirsha: Thx!

PR #4069, opened 2026-04-28
frontend: simplify account summary unit price

shonsirsha:
Drop the trailing `/<coin>` suffix on the unit-price line and
add an opt-in `removeTrailingZeros` flag on
`Amount`/`AmountWithUnit` (BTC only) so `1.00000000 BTC`
renders as `1 BTC`. Sat and fiat formatting are unchanged.

### DIFF
--- a/frontends/web/src/components/amount/amount-with-unit.tsx
+++ b/frontends/web/src/components/amount/amount-with-unit.tsx
@@ -13,6 +13,7 @@ type TAmountWithUnitProps = {
   sign?: string;
   alwaysShowAmounts?: boolean;
   convertToFiat?: boolean;
+  removeTrailingZeros?: boolean;
   amountClassName?: string;
   unitClassName?: string;
   maxDecimals?: number;
@@ -23,6 +24,7 @@ export const AmountWithUnit = ({
   enableRotateUnit,
   sign,
   convertToFiat,
+  removeTrailingZeros,
   alwaysShowAmounts = false,
   amountClassName = '',
   unitClassName = '',
@@ -51,6 +53,7 @@ export const AmountWithUnit = ({
   }
 
   const enableClick = enableRotateUnit && (convertToFiat || isBitcoinCoin(amount.unit));
+  const stripTrailingZeros = removeTrailingZeros && displayedUnit === 'BTC';
 
   return (
     <span className={`
@@ -67,6 +70,7 @@ export const AmountWithUnit = ({
           unit={displayedUnit}
           onMobileClick={enableClick ? onClick : undefined}
           maxDecimals={maxDecimals}
+          removeTrailingZeros={stripTrailingZeros}
         />
       ) : '---'}
       {' '}
--- a/frontends/web/src/components/amount/amount.tsx
+++ b/frontends/web/src/components/amount/amount.tsx
@@ -7,6 +7,13 @@ import { LocalizationContext } from '@/contexts/localization-context';
 import { useMediaQuery } from '@/hooks/mediaquery';
 import style from './amount.module.css';
 
+const stripTrailingZeros = (value: string): string => {
+  if (!value.includes('.')) {
+    return value;
+  }
+  return value.replace(/\.?0+$/, '');
+};
+
 const formatSats = (amount: string): JSX.Element => {
   const blocks: JSX.Element[] = [];
   const blockSize = 3;
@@ -67,7 +74,7 @@ const formatBtc = (
 ) => {
   const dot = amount.indexOf('.');
   if (dot === -1) {
-    return amount;
+    return formatLocalizedAmount(amount, group, decimal);
   }
   // localize the first part, everything up to the second decimal place, the rest is grouped by spaces
   const formattedPart = formatLocalizedAmount(amount.slice(0, dot + 3), group, decimal);
@@ -96,7 +103,7 @@ const formatEth = (
 ): string => {
   const dot = amount.indexOf('.');
   if (dot === -1) {
-    return amount;
+    return formatLocalizedAmount(amount, group, decimal);
   }
   const truncated = amount.slice(0, dot + maxDecimals + 1);
   return formatLocalizedAmount(truncated, group, decimal);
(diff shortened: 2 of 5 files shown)


frontends/web/src/routes/account/summary/asset-balance-with-unit-price.tsx
```diff
@@ -37,10 +37,9 @@ export const AssetBalanceWithUnitPrice = ({ amount, coinCode, coinName, dataTest
                 amountClassName={style.unitPrice}
                 amount={unitPrice}
                 convertToFiat
+                preferBtcOverSat
+                removeTrailingZeros
               />
-              {amount?.unit && (
-                <span className={style.pairUnit}>/{amount.unit}</span>
-              )}
```
thisconnect, line 43: 👍

frontends/web/src/components/amount/amount.tsx
```diff
@@ -7,6 +7,13 @@ import { LocalizationContext } from '@/contexts/localization-context';
 import { useMediaQuery } from '@/hooks/mediaquery';
 import style from './amount.module.css';
 
+const stripTrailingZeros = (value: string): string => {
+  if (!value.includes('.')) {
+    return value;
+  }
+  return value.replace(/\.?0+$/, '');
+};
```
thisconnect, line 15: I'm slightly worried that the low-level `<Amount>` component would render 10 BTC as 1 BTC here. (Which is more a theoretical issue and shouldn't happen as the backend always sends `'10.0000000'`).

frontends/web/src/components/amount/amount-with-unit.tsx
```diff
@@ -13,6 +13,8 @@ type TAmountWithUnitProps = {
   sign?: string;
   alwaysShowAmounts?: boolean;
   convertToFiat?: boolean;
+  preferBtcOverSat?: boolean;
```
thisconnect: 👍

frontends/web/src/routes/account/summary/asset-balance-with-unit-price.tsx
thisconnect, line 19: something is wrong here, logging unitPrice it should return TAmountWithConversions, but the amount is a unit 🤔 ?

```
{
  amount: "TLTC" // <--- this should be amount but is unit?
  ​conversions: Object { AUD: "76.89", BRL: "275.16", BTC: "0.00071928", … }
  estimated: false
  unit: "TLTC"
}
```

frontends/web/src/components/amount/amount.tsx
thisconnect, line 15: In theory the low level amount component could handle 10 btc (without dot). but I am worried that it would strip 10 btc to 1 btc. 

Instead of passing these props through AmountWithUnit -> Amount. I'd prefer if we could already format the conversion amount in the fiat-prices endpoint coming from the backend. That way we can keep the frontend simpler.

We still might need preferBtcOverSat prop.

thisconnect (changes requested):

frontends/web/src/routes/account/summary/asset-balance-with-unit-price.tsx
shonsirsha: Yup true. Thanks. We're not using `amount` anywhere tho from this endpoint, so we're "safe". But a real bug if we use it. Needs to be resolved, but isnt blocking this PR.

thisconnect (approved):
Could you add a few test cases to amount.test.tsx?

maybe large small etc?

I'll paste my test draft in a comment, but I am not so good at writing such tests.

thisconnect:
@shonsirsha quickly tried to test a few cases but I am not 100% sure if they all correct or if some important ones could be missing.

```tsx

  describe('remove trailing zeros from amounts', () => {

    it('0.00 CHF becomes 0 CHF', () => {
      const { container } = render(<Amount amount="0.00" unit="CHF" removeTrailingZeros/>);
      expect(container).toHaveTextContent('0'); // not sure how to test to NOT have 0.00
    });

    it('0.10 CHF stays 0.1 CHF', () => {
      const { container } = render(<Amount amount="0.10" unit="CHF" removeTrailingZeros/>);
      expect(container).toHaveTextContent('0.1');
    });

    it('1\'000 CHF stays 1\'000 CHF', () => {
      const { container } = render(<Amount amount="1'000" unit="CHF" removeTrailingZeros/>);
      expect(container).toHaveTextContent('1’000');
    });

    it('1\'000.00 CHF comes 1\'000 CHF', () => {
      const { container } = render(<Amount amount="1'000.00" unit="CHF" removeTrailingZeros/>);
      expect(container).toHaveTextContent('1’000');
    });

    it('10 BTC stays 10 BTC', () => {
      const { container } = render(<Amount amount="10" unit="BTC" removeTrailingZeros/>);
      expect(container).toHaveTextContent('10');
    });

    it('10\'000\'000" BTC stays to 10\'000\'000 BTC', () => {
      const { container } = render(<Amount amount="10'000'000" unit="BTC" removeTrailingZeros/>);
      expect(container).toHaveTextContent('10’000’000');
    });

    it('10\'000\'000.00000000" BTC stays to 10\'000\'000 BTC', () => {
      const { container } = render(<Amount amount="10'000'000.00000000" unit="BTC" removeTrailingZeros/>);
      expect(container).toHaveTextContent('10’000’000');
    });

    it('10\'000 BTC stays to 10\'000 BTC', () => {
      const { container } = render(<Amount amount="10'000.00" unit="BTC" removeTrailingZeros/>);
      expect(container).toHaveTextContent('10’000');
    });

    it('10.010000000 BTC trimms to 10.01 BTC', () => {
      const { container } = render(<Amount amount="10.010000000" unit="BTC" removeTrailingZeros/>);
      expect(container).toHaveTextContent('10.01');
    });

    it('10100000000 sat stays 10100000000 sat', () => {
      const { container } = render(<Amount amount="10100000000" unit="sat" removeTrailingZeros/>);
      expect(container).toHaveTextContent('10100000000');
    });
    
  });
```

shonsirsha:
Thank you! @thisconnect , added some more tests.

PR #4070, opened 2026-04-28
frontend: show decimal keyboard for amount inputs

benma:
iOS uses the inputmode hint to choose the virtual keyboard for number fields. The previous numeric hint can show a keypad without a decimal separator, which prevents entering fractional send or swap amounts such as 0.1 ETH.

Default NumberInput to decimal input mode so the existing send fiat and swap amount fields request a decimal-capable keyboard, and route the send coin amount through NumberInput as well. NumberInput still renders type=number and keeps the onInput wiring internally, so desktop parsing, validation, and physical keyboard behavior should remain unchanged; Android may show a decimal-capable virtual keyboard where the WebView honors the hint.

The runtime scope is limited to account send amount fields and the market swap amount selector. Integer-only gap limit settings use raw Input type=number and are not affected.

### DIFF
--- a/frontends/web/src/components/forms/input-number.test.tsx
+++ b/frontends/web/src/components/forms/input-number.test.tsx
@@ -11,6 +11,11 @@ describe('components/forms/input-number', () => {
     expect(container.querySelector('[type="number"')).toBeTruthy();
   });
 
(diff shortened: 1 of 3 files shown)


shonsirsha (approved):
LGTM

PR #4071, opened 2026-04-28
frontend: remove swap output to be focusable

thisconnect:
An altnernative would have been to change the :focus-within css selector to take readonly input into account, but that way the input element would technically still be focusable.

Changed to disabled to the element cannot be focesd.

- [ ] updating the Changelog
- [ ] writing unit tests


strmci (approved):
tACK

thisconnect:
rebased

PR #4077, opened 2026-04-28
swap: prevent account refresh flicker

bznein:
Keep the last successful swap accounts response while a dependency-triggered reload is pending, so the swap selectors do not briefly render skeletons.

Also skip no-op account-used updates for already visible used accounts. This avoids emitting redundant accounts reload events after regular sync-done notifications.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -1689,11 +1689,13 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 		}
 	}
 	log.Info("marking account as used")
+	var emitUpdate bool
 	err := backend.config.ModifyAccountsConfig(func(accountsConfig *config.AccountsConfig) error {
 		acct := accountsConfig.Lookup(account.Config().Config.Code)
 		if acct == nil {
 			return errp.Newf("could not find account")
 		}
+		emitUpdate = !acct.Used || acct.HiddenBecauseUnused
 		acct.Used = true
 		acct.HiddenBecauseUnused = false
 
@@ -1703,7 +1705,9 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 		log.WithError(err).Error("checkAccountUsed")
 		return
 	}
-	backend.emitAccountsStatusChanged()
+	if emitUpdate {
+		backend.emitAccountsStatusChanged()
+	}
 	backend.maybeAddHiddenUnusedAccounts()
 }
 
--- a/frontends/web/src/routes/market/swap/swap.tsx
+++ b/frontends/web/src/routes/market/swap/swap.tsx
@@ -21,6 +21,7 @@ import {
   getSwapQuote,
   signSwap,
   type TSwapAccount,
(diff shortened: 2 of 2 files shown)


backend/accounts.go
```diff
@@ -1674,8 +1674,13 @@ func (backend *Backend) checkAccountUsed(account accounts.Interface) {
 		}
 	}
 
+	accountConfig := account.Config().Config
+	if accountConfig.Used && !accountConfig.HiddenBecauseUnused {
+		return
+	}
```
benma: This does not call `backend.maybeAddHiddenUnusedAccounts()`  to keep scanning. I think the safer way to set `emitUpdate = !acct.Used || act.HiddenBecauseUnused` inside ModifyAccounts() before setting these fields, then call `emitAccountsStatusChanged()` if it is true, to make it explicit.

benma (changes requested):

backend/accounts.go
bznein: good point, changed

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -92,7 +93,14 @@ export const Swap = ({
   const navigate = useNavigate();
   const { activeCurrencies, btcUnit } = useContext(RatesContext);
   // accounts is added as a dependency, to reload swap accounts when the account list changes.
-  const swapAccounts = useLoad(getSwapAccounts, [accounts]);
+  const loadedSwapAccounts = useLoad(getSwapAccounts, [accounts]);
```
thisconnect, line 95: I am not 100% sure. Accounts is just a reference to an array, but does not do deep equality check as far as I know.

So even if the accounts stay same the dependecy list can channge and force a use load again. afaik

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -92,7 +93,14 @@ export const Swap = ({
...
+  const [retainedSwapAccounts, setRetainedSwapAccounts] = useState<TSwapAccounts>();
+  useEffect(() => {
+    if (loadedSwapAccounts !== undefined) {
+      setRetainedSwapAccounts(loadedSwapAccounts);
+    }
+  }, [loadedSwapAccounts]);
```
thisconnect, line 101: I am also not 100% sure about this. We also have  usePrevious and useDefault hooks that could maybe help here?

I'd love to defer to @shonsirsha 😇

thisconnect (commented):
@coderabbitai review

thisconnect:
@coderabbitai review

benma (commented):
backend utACK

thisconnect (approved):
tested a bit, LGTM thanks

bznein:
rebased

bznein:
> backend utACK

you requested changes so can't merge until you approve :D

benma (approved):

PR #4080, opened 2026-04-28
backend: fix amount/unit in CoinFiatPrices

benma:
Amount was not an amount, but should be.

Unit in sats mode was "sat", even though the conversions are for BTC.

Neither amount nor unit is needed, it is dead code. We keep the shape for now anyway to avoid a larger refactor in the frontend.


shonsirsha (approved):
lgtm

PR #4082, opened 2026-04-28
frontend: fix random generic error in btcdirect

thisconnect:
AlertUser is an antipaterns and goes around the react state.

Changed to just show a error message in case the response did not succeed.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/market/btcdirect.tsx
+++ b/frontends/web/src/routes/market/btcdirect.tsx
@@ -19,6 +19,7 @@ import { BTCDirectTerms } from '@/components/terms/btcdirect-terms';
 import { MarketGuide } from './guide';
 import { alertUser } from '@/components/alert/Alert';
 import { useVendorIframeResizeHeight, useVendorTerms } from '@/hooks/vendor-iframe';
+import { Message } from '@/components/message/message';
 import style from './iframe.module.css';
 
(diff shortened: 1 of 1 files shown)


thisconnect:
@smokyisthatyou please test

thisconnect:
unfortunatelly this error doesn't often popup .. maybe it happens more if the pc is very busy and the BE response very slow?

smokyisthatyou:
@thisconnect not easy to reproduce but on previous version i was able to switching very quickly between widgets. 
i have been testing in the same way for some time and not generic error has been brought app :)

frontends/web/src/routes/market/btcdirect.tsx
```diff
@@ -241,7 +242,13 @@ export const BTCDirect = ({
                     src={btcdirectInfo.url}>
                   </iframe>
                 ) : (
-                  <>{btcdirectInfo?.errorMessage ? alertUser(btcdirectInfo.errorMessage) : alertUser('genericError')}</>
```
thisconnect, line 244: Before `alertUser('genericError')` just showed 'genericError' in case errorMessage was falsy

Note: missing `t()`

benma (approved):
utack

thisconnect:
rebased

PR #4083, opened 2026-04-29
frontend: change chart default to year

thisconnect:
As the chart always tries to fit in as many datapoints as possible and adapts to different screensizes, the chart graph and the percentage difference may look different on various devices.

To have a more consistent chart and percentage differences this commit changed the default to 'Year'. With that the graph and percentage should be identical by default accross various devices.

Users that like the old behavior can still change back to 'All'.

- [x] updating the Changelog
- [ ] writing unit tests


shonsirsha (approved):
LGTM

strmci (approved):
utACK

PR #4087, opened 2026-04-29
frontend: unlock view should cover mobilefullscreen selector

thisconnect:
In the rare case where a user opens account or region selector on mobile so it goes into fullscreen AND connects a device to unlock, the unlock view was hidden by the fullscreen selector.

Reduced the mobile fullscreen selector to 1 so it is just one higher than the rest which should be sufficient.

- [ ] updating the Changelog
- [ ] writing unit tests


thisconnect:
tested:
- market region and account selector
- settings lang, currencies
- send send-to-self dropdown
- swap account selector

shonsirsha (approved):
LGTM

PR #4090, opened 2026-04-29
backend: restore OTC market errors

bznein:
Return coinNotSupported only when neither BTC Direct nor Pocket supports the coin.

For mixed OTC cases, where one vendor rejects the coin and the other rejects the region, return regionNotSupported because the selected coin is supported somewhere but not in the chosen region.

- [ ] updating the Changelog
- [ ] writing unit tests


bznein:
Added @benma for backend review but pinging @thisconnect for testing since the change is frontend-facing :)

thisconnect (commented):
tested  👍  with this PR is now shows correct "No options available for this region." message in OTC tab when there are no offers in the selected region.

benma (approved):
utACK

PR #4092, opened 2026-04-30
frontend: fix iOS decimal input

shonsirsha:
iOS can expose a comma decimal key while rejecting it in number inputs. Use an iOS text input path so users can enter comma or dot.


benma (approved):
utACK

shonsirsha:
I might still make the code a bit neater. Will convert to draft for now

PR #4093, opened 2026-04-30
UI: Implement toast component

jadzeidan:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -2,7 +2,7 @@
 
 import { useCallback, useEffect, useMemo, Fragment } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
-import { useTranslation } from 'react-i18next';
+import { getAccounts } from './api/account';
 import { useSync } from './hooks/api';
 import { useDefault } from './hooks/default';
 import { usePrevious } from './hooks/previous';
@@ -11,12 +11,9 @@ import { usePlatformClass } from './hooks/platform';
 import { useAppReady } from './hooks/appready';
 import { AppRouter } from './routes/router';
 import { Wizard as BitBox02Wizard } from './routes/device/bitbox02/wizard';
-import { getAccounts } from './api/account';
 import { syncAccountsList } from './api/accountsync';
 import { getDeviceList } from './api/devices';
 import { syncDeviceList } from './api/devicessync';
-import { syncNewTxs } from './api/transactions';
-import { notifyUser } from './api/system';
 import { ConnectedApp } from './connected';
 import { Alert } from './components/alert/Alert';
 import { Aopp } from './components/aopp/aopp';
@@ -26,15 +23,14 @@ import { Sidebar } from './components/sidebar/sidebar';
 import { RouterWatcher } from './utils/route';
 import { Darkmode } from './components/darkmode/darkmode';
 import { AuthRequired } from './components/auth/authrequired';
+import { IncomingTransactionNotifier } from './components/toast/incoming-transaction-notifier';
 import { WCSigningRequest } from './components/wallet-connect/incoming-signing-request';
 import { Providers } from './contexts/providers';
 import { BottomNavigation } from './components/bottom-navigation/bottom-navigation';
 import { getBottomNavKey } from './components/bottom-navigation/utils';
 import styles from './app.module.css';
-
 export const App = () => {
   usePlatformClass();
-  const { t } = useTranslation();
   const navigate = useNavigate();
   const { pathname } = useLocation();
   useIgnoreDrop();
@@ -46,15 +42,6 @@ export const App = () => {
 
   const deviceIDs = Object.keys(devices);
   const firstDevice = deviceIDs[0];
-
-  useEffect(() => {
-    return syncNewTxs((meta) => {
-      notifyUser(t('notification.newTxs', {
-        count: meta.count,
-        accountName: meta.accountName,
-      }));
-    });
-  }, [t]);
 
   const maybeRoute = useCallback(() => {
     const currentURL = window.location.hash.replace(/^#/, '');
@@ -173,6 +160,7 @@ export const App = () => {
     <ConnectedApp>
       <Providers>
         <Darkmode />
+        <IncomingTransactionNotifier activeAccounts={activeAccounts} />
         <div className="app">
           <AuthRequired/>
           <Sidebar
--- a/frontends/web/src/components/toast/toast-view.tsx
+++ b/frontends/web/src/components/toast/toast-view.tsx
@@ -9,17 +9,20 @@ import style from './Toast.module.css';
 
 type TToastProps = {
   type?: TMessageTypes;
+  // Deprecated prop kept for compatibility with the existing callsites.
+  theme?: TMessageTypes;
   icon?: ReactNode;
+  className?: string;
   onClose?: () => void;
   children: ReactNode;
 };
 
-export const Toast = ({ type = 'info', icon, onClose, children }: TToastProps) => {
+export const Toast = ({ type, theme, icon, className = '', onClose, children }: TToastProps) => {
+  const resolvedType = type || theme || 'info';
   const { isDarkMode } = useDarkmode();
   const iconWithSpacing = icon ? <span className={style.icon}>{icon}</span> : undefined;
-  const className = [style.toast, style.toastItem].filter(Boolean).join(' ');
   return (
-    <Message icon={iconWithSpacing} type={type} className={className}>
+    <Message icon={iconWithSpacing} type={resolvedType} className={`${style.toast || ''} ${className || ''}`.trim()}>
       <div className={style.container}>
         <div className={style.content}>{children}</div>
         <button
--- a/frontends/web/src/contexts/index.ts
+++ b/frontends/web/src/contexts/index.ts
@@ -1,0 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
+export { ToastContext, useToast } from './toast-context';
+export { ToastProvider } from './toast-provider';
+export type { TShowToast, TToastContext } from './toast-context';
--- a/frontends/web/src/contexts/providers.tsx
+++ b/frontends/web/src/contexts/providers.tsx
@@ -8,6 +8,7 @@ import { BackNavigationProvider } from './BackNavigationContext';
 import { WCWeb3WalletProvider } from './WCWeb3WalletProvider';
 import { RatesProvider } from './RatesProvider';
 import { LocalizationProvider } from './localization-provider';
+import { ToastProvider } from './toast-provider';
 
 type Props = {
   children: ReactNode;
@@ -21,9 +22,11 @@ export const Providers = ({ children }: Props) => {
           <DarkModeProvider>
             <LocalizationProvider>
               <RatesProvider>
-                <WCWeb3WalletProvider>
-                  {children}
-                </WCWeb3WalletProvider>
+                <ToastProvider>
+                  <WCWeb3WalletProvider>
+                    {children}
+                  </WCWeb3WalletProvider>
+                </ToastProvider>
               </RatesProvider>
             </LocalizationProvider>
           </DarkModeProvider>
--- a/frontends/web/src/contexts/toast-provider.tsx
+++ b/frontends/web/src/contexts/toast-provider.tsx
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { ReactNode, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { TMessageTypes } from '@/utils/types';
+import { Toast } from '@/components/toast/toast';
+import { ToastContext, TShowToast } from './toast-context';
+import style from '@/components/toast/Toast.module.css';
+
+type TToastItem = {
+  id: number;
+  duration: number;
+  icon?: ReactNode;
+  message: ReactNode;
+  persistent: boolean;
+  type: TMessageTypes;
+};
+
+type TToastProviderProps = {
+  children: ReactNode;
+};
+
+const DEFAULT_DURATION_MS = 5000;
+
+let nextToastID = 1;
+
+export const ToastProvider = ({ children }: TToastProviderProps) => {
+  const [toasts, setToasts] = useState<TToastItem[]>([]);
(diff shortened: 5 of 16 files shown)


jadzeidan:
@coderabbitai

frontends/web/src/components/toast/toast.tsx
```diff
@@ -1,27 +1,155 @@
...
+  return (
+    <Message icon={iconWithSpacing} type={resolvedType} className={`${style.toast || ''} ${className || ''}`.trim()}>
+      <div className={style.container}>
+        <div className={style.content}>{children}</div>
+        <button
+          aria-label="Close toast"
+          className={style.closeButton}
+          hidden={!onClose}
+          onClick={onClose}
+          type="button">
```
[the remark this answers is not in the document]
jadzeidan: will delete `aria-label="Close toast"` since we don't support aria yet.

jadzeidan:
@coderabbitai PTAL

jadzeidan:
@thisconnect PTAL :)

frontends/web/src/app.tsx
```diff
@@ -27,14 +27,122 @@ import { RouterWatcher } from './utils/route';
...
+import { useToast } from './components/toast/toast';
 import { Providers } from './contexts/providers';
 import { BottomNavigation } from './components/bottom-navigation/bottom-navigation';
 import { getBottomNavKey } from './components/bottom-navigation/utils';
 import styles from './app.module.css';
 
+const IncomingTxIcon = () => (
+  <svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path d="M1 17H17M9 11V1M9 11L13 7M9 11L5 7" stroke="currentColor" strokeWidth="2" strokeLinecap="round" strokeLinejoin="round"/>
+  </svg>
```
thisconnect: We should avoid inline svg if possible as react will have to needlessly scan all the attributes for dom-diffing and updating the dom.

Please move to .svg file and import from `@/components/icons`, see how the other icons are used in the app.

frontends/web/src/app.tsx
```diff
@@ -27,14 +27,122 @@ import { RouterWatcher } from './utils/route';
...
+type TIncomingTransactionNotifierProps = {
+  activeAccounts: TAccount[];
+};
+
+const IncomingTransactionNotifier = ({ activeAccounts }: TIncomingTransactionNotifierProps) => {
+  const { t } = useTranslation();
+  const { showToast } = useToast();
+  const shownIncomingTxIDsRef = useRef<Record<string, boolean>>({});
+  const initializedAccountsRef = useRef<Record<string, boolean>>({});
+  const seedingAccountsRef = useRef<Record<string, boolean>>({});
```
thisconnect: The commit message writes

> Add in-app notification using toast component for incoming TX.

So I don't understand why we need to keep a map of all transactions and fetch all transactions first, then mark them as seen (markIncomingAsSeen).

The previous notifications just subscribed to new transcations and fire and forget a notification. Which is much simpler and seems to achieve exactly the same, but I feel I am missing something.

It would help if the commit message explains what the goal is so it is clear why this code is here.

frontends/web/src/app.tsx
```diff
@@ -27,14 +27,122 @@ import { RouterWatcher } from './utils/route';
...
+        })();
+      });
+    });
+    return () => {
+      unsubscribers.forEach(unsubscribe => unsubscribe());
+    };
+  }, [activeAccounts, detectAndToastIncomingForAccount, seedIncomingTransactionsForAccount]);
+
+  return null;
+};
```
thisconnect: This component doesn't belong here, but maybe it can be dropped completely? see my other comment

frontends/web/src/routes/settings/toast-demo.tsx
```diff
@@ -0,0 +1,123 @@
...
+  <svg width="16" height="18" viewBox="0 0 8 9" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path d="M6.61718 3.4375V0.34375C6.61718 0.153614 6.46357 0 6.27343 0H1.63281C1.44267 0 1.28906 0.153614 1.28906 0.34375V3.4375C0.577934 3.4375 0 3.99717 0 4.6836V8.50782C0 8.55508 0.038672 8.59375 0.085938 8.59375H0.687504C0.734764 8.59375 0.773434 8.55508 0.773434 8.50782V4.6836C0.773434 4.42256 1.00547 4.21094 1.29013 4.21094H6.61611C6.90078 4.21094 7.13281 4.42256 7.13281 4.6836V8.50782C7.13281 8.55508 7.17148 8.59375 7.21875 8.59375H7.82031C7.86757 8.59375 7.90625 8.55508 7.90625 8.50782V4.6836C7.90625 3.99717 7.32832 3.4375 6.61718 3.4375ZM2.0625 3.4375V0.773438H5.84375V3.4375H2.0625ZM3.35156 1.46094H2.83593C2.78867 1.46094 2.75 1.49961 2.75 1.54688V2.0625C2.75 2.10977 2.78867 2.14844 2.83593 2.14844H3.35156C3.39882 2.14844 3.4375 2.10977 3.4375 2.0625V1.54688C3.4375 1.49961 3.39882 1.46094 3.35156 1.46094ZM5.07031 1.46094H4.55468C4.50742 1.46094 4.46875 1.49961 4.46875 1.54688V2.0625C4.46875 2.10977 4.50742 2.14844 4.55468 2.14844H5.07031C5.11757 2.14844 5.15625 2.10977 5.15625 2.0625V1.54688C5.15625 1.49961 5.11757 1.46094 5.07031 1.46094Z" fill="#5E94C0"/>
+  </svg>
+);
```
thisconnect, line 44: I was going to ask to move all inlinve SVG to `@/components/icon` components, but if since these are only used in the demo file I guess it's ok to keep them here.

It would be nice if these dont land in the production built, no need to do anything rightnow, but I want to check if we can exclude from our vite production build.

frontends/web/src/contexts/index.ts
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
+export { ToastContext, useToast } from './toast-context';
+export { ToastProvider } from './toast-provider';
+export type { TShowToast, TToastContext } from './toast-context';
```
thisconnect, line 5: Either add all contexts here in this file or remove it.
 
But since we don't need it, please remove this file.

frontends/web/src/components/toast/toast.tsx
```diff
@@ -1,27 +1,5 @@
...
-    setTimeout(() => setActive(true), 5);
-  }, []);
-
-  return (
-    <div
-      className={[style.toast, style[theme], active ? style.active : '', withGuide ? style.shifted : ''].join(' ')}>
-      <p>{children}</p>
-    </div>
-  );
-};
```
thisconnect, line 25: We unfortunatelly never removed the old toast, it is used in backups.tsx which is used in settings and setup. Did you test that it still works?

frontends/web/src/components/toast/toast.tsx
```diff
@@ -1,27 +1,5 @@
...
-
-
+export { Toast } from './toast-view';
+export { ToastContext, ToastProvider, useToast } from '@/contexts';
+export type { TShowToast, TToastContext } from '@/contexts';
```
thisconnect: We try to not have circular imports/dependencies, so importing from context into components is a bit a red flag.

Components that need to import any of these can directly import from context/toast. No need to import and re-export here.

frontends/web/src/components/toast/toast-view.tsx
```diff
@@ -0,0 +1,39 @@
...
+          className={style.closeButton}
+          hidden={!onClose}
+          onClick={onClose}
+          type="button">
+          {isDarkMode ? <CloseXWhite /> : <CloseXDark />}
+        </button>
+      </div>
+    </Message>
+  );
+};
```
thisconnect, line 39: Please move this to components/toast/toast.tsx

frontends/web/src/components/toast/toast-view.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { CloseXDark, CloseXWhite } from '@/components/icon';
+import { Message } from '@/components/message/message';
+import { useDarkmode } from '@/hooks/darkmode';
+import { TMessageTypes } from '@/utils/types';
+import style from './Toast.module.css';
+
+type TToastProps = {
+  type?: TMessageTypes;
+  // Deprecated prop kept for compatibility with the existing callsites.
+  theme?: TMessageTypes;
```
thisconnect, line 13: the old Toast is only used in backups.tsx.

Please remove the deprecated theme prop here and update backups.tsx to use the new toast with type prop.

frontends/web/src/contexts/toast-provider.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return (
+    <ToastContext.Provider value={value}>
+      {children}
+      {toasts.length > 0 && (
+        <div className={style.viewport} aria-live="polite" role="status">
+          {toasts.map(({ id, icon, message, persistent, type }) => (
+            <Toast
+              key={id}
+              className={style.toastItem}
```
thisconnect: As toast item should always have this class, remove className prop and this style here and just use the class from toast module.css

It doesn't make sense if the context/provider imports styles from the toast component to apply that style to the toast component.

frontends/web/src/locales/en/app.json
```diff
@@ -2055,6 +2053,32 @@
...
+      },
+      "error": "Something went wrong. Please try again.",
+      "info": "Heads up! This is an informational toast.",
+      "incomingBtc": "Incoming 0.01 BTC",
+      "persistent": "This toast will stay here until you dismiss it.",
+      "success": "Success! The action completed.",
+      "warning": "Warning: Please double-check before continuing."
+    },
+    "title": "Toast demo"
+  },
```
thisconnect: I don't think demo views should have translated text as all those need to be maintained and ad extra work for translators.


A component library can just use hardcoded text. In cases where we want to test how a component behaves with  longer text it would be better to add more demo cases with hardcoded longer text. For example to visualize how the text in the component break in cases.

frontends/web/src/routes/router.tsx
```diff
@@ -329,6 +337,7 @@ export const AppRouter = ({ devices, devicesKey, accounts, activeAccounts }: TAp
           <Route path="device-settings/recovery-words/:deviceID" element={RecoveryWordsEl} />
           <Route path="device-settings/bip85/:deviceID" element={Bip85El} />
           <Route path="advanced-settings" element={AdvancedSettingsEl} />
+          <Route path="toast-demo" element={ToastDemoEl} />
```
thisconnect: I can't find that this route is linked from anywhere.

I am not sure 100% if we should start adding demo views  in to the production build. 

This route could maybe be conditionally added in case isDevServers? 

We could also add a link somewhere for example in sidebar conditionally if isDevServers

frontends/web/src/locales/en/app.json
```diff
@@ -1635,6 +1634,7 @@
     "title": "Note"
   },
   "notification": {
+    "incomingTxToast": "Incoming {{amount}} {{unit}}",
     "newTxs_one": "New transaction in: {{accountName}}",
     "newTxs_other": "{{count}} new transactions in: {{accountName}}"
```
thisconnect, line 1641: if these 2 are not used anymore in the codebase please remove.

frontends/web/src/contexts/providers.tsx
```diff
@@ -22,7 +23,9 @@ export const Providers = ({ children }: Props) => {
             <LocalizationProvider>
               <RatesProvider>
                 <WCWeb3WalletProvider>
-                  {children}
+                  <ToastProvider>
+                    {children}
+                  </ToastProvider>
                 </WCWeb3WalletProvider>
```
thisconnect: A provider only makes its context available to components inside it in the tree.

So in case we ever want to create a toast message from WCWeb3WalletProvider the ToastProvider would have to be nested as parent.

I haven't checked, but it looks like all others should not need toast, but for WCWeb3WalletProvider it could make sense.

frontends/web/src/app.tsx
```diff
@@ -27,14 +27,122 @@ import { RouterWatcher } from './utils/route';
...
+        }
+        markIncomingAsSeen(account.code, tx.internalID);
+        showIncomingTxToast(tx.amount.amount, tx.amount.unit);
+      });
+  }, [markIncomingAsSeen, showIncomingTxToast]);
+
+  useEffect(() => {
+    const unsubscribers = activeAccounts.map((account) => {
+      return syncdone(account.code, () => {
+        void (async () => {
```
thisconnect: nit: void doesn't do anything here afaik. remove?

frontends/web/src/app.tsx
```diff
@@ -27,14 +27,122 @@ import { RouterWatcher } from './utils/route';
...
+        .forEach(tx => markIncomingAsSeen(account.code, tx.internalID));
+      initializedAccountsRef.current[account.code] = true;
+    } finally {
+      seedingAccountsRef.current[account.code] = false;
+    }
+  }, [markIncomingAsSeen]);
+
+  // Seed known incoming transactions for active accounts so we only toast truly new ones.
+  useEffect(() => {
+    void Promise.all(activeAccounts.map(account => seedIncomingTransactionsForAccount(account)));
```
thisconnect: nit: void doesn't do anything here afaik. remove?

thisconnect (changes requested):
light code review with  a few questions.

frontends/web/src/app.tsx
jadzeidan: I think we still need this extra state for the following:
- So old TXs don't trigger a toast on sync
- to distinguish between outgoing and incoming txs (we only want incoming to trigger a toast)
- So we can have the amount and units in the toast as well (the current notification doesn't know the amount)

thisconnect: At the moment I think we should keep the logic in backend and the frontend only subscribes to `/new-txs` endpoint.

frontends/web/src/components/toast/toast.tsx
jadzeidan: I think old toast should happen when aborting restore from backup, but I don't see it when testing. Just "Restoring backup was aborted" dialog.

PR #4095, opened 2026-04-30
swap: show receive amount even with no funds.

bznein:
Instead of just erroring, we still get a quote and show the user how much they would get from the swap.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/market/swap/swap.test.tsx
+++ b/frontends/web/src/routes/market/swap/swap.test.tsx
@@ -302,6 +302,49 @@ describe('routes/market/swap', () => {
     expect(screen.queryByText(/ETH\.USDC/)).not.toBeInTheDocument();
   });
 
+  it('keeps quote output visible for insufficient funds', async () => {
+    const user = userEvent.setup();
+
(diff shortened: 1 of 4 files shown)


frontends/web/src/routes/market/swap/swap.test.tsx
```diff
@@ -302,6 +302,49 @@ describe('routes/market/swap', () => {
...
+      </RatesContext.Provider>,
+    );
+
+    await user.click(await screen.findByTestId('agree-swap-terms'));
+    await user.type(await screen.findByLabelText('swapSendAmount'), '2');
+
+    expect(await screen.findByText('THORChain + Mayachain')).toBeInTheDocument();
+    expect(await screen.findByTestId('swapGetAmount')).toHaveTextContent('1.23');
+    expect(screen.getByText(/insufficient funds/i)).toBeInTheDocument();
+    expect(screen.getByRole('button', { name: 'Swap' })).toBeDisabled();
```
thisconnect, line 345: 👍

thisconnect (approved):
makes sense to me untested LGTM

benma (approved):
utACK

PR #4096, opened 2026-05-01
ios: fix keyboard pushing content under status bar safe area

shonsirsha:
add native overlay that covers the status bar area while keyboard is visible, preventing Webview content from showing under the notch/ safearea.

Done by an invisible coloured bar on top of the screen that only appears when the keyboard opens.

Also guard updateUIView to load index.html only once.


sutterseba (approved):
tested on iPhone

PR #4097, opened 2026-05-04
frontend: use mobile viewport bugfix on width <=768px

shonsirsha:
When used on all sizes, this breaks ipadOS

Before (see whitespace on the bottom):

image

After:

https://github.com/user-attachments/assets/60a36cd0-ec02-4d53-bcfa-67ad9891a1d6

### DIFF
--- a/frontends/web/src/app.module.css
+++ b/frontends/web/src/app.module.css
@@ -4,8 +4,6 @@
     flex-direction: column;
     height: 100dvh;
     min-width: 0;
-    /* mobile viewport bug fix */
-    max-height: -webkit-fill-available;
     /* iOS safe area handling */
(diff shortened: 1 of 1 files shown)


frontends/web/src/globals.d.ts
```diff
@@ -18,6 +18,7 @@ export declare global {
     onMobileCallResponse?: (queryID: number, response: unknown) => void;
     onMobilePushNotification?: (msg: TPayload) => void;
     runningOnIOS?: boolean;
+    runningOnIPadOS?: boolean;
```
sutterseba: see comment in Linear

sutterseba (changes requested):

sutterseba (approved):
tested on iPad

PR #4099, opened 2026-05-04
frontend: add back button to all market providers 

thisconnect:


frontends/web/src/routes/market/bitrefill.tsx
```diff
@@ -195,12 +196,19 @@ export const Bitrefill = ({
...
       <div className="container">
         <div className="innerContainer">
           <div className={style.header}>
-            <Header title={<h2>{t('generic.spend', { context: translationContext })}</h2>} />
+            <Header title={
+              <>
+                <h2 className="hide-on-small">{title}</h2>
+                <MobileHeader withGuide title={title} />
+              </>
+            } />
```
thisconnect, line 211: In the future we should refactor `<Header>` & `<MobileHeader>` maybe we can make it one nice component

shonsirsha (approved):
LGTM

PR #4104, opened 2026-05-04
frontend: fix partly visible cursor in swap send

thisconnect:
The cursor was only partly visible in case input field was empty, it was cut-off by border-radius.

- [ ] updating the Changelog
- [ ] writing unit tests


sutterseba (approved):
utACK

PR #4110, opened 2026-05-06
frontend: carry input attrs in NumberInput iOS change event

shonsirsha:
Ensures id/name/type/disabled/checked reading to not be undefined on the iOS-only event path.


shonsirsha:
Closing in favour of: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4112

PR #4111, opened 2026-05-06
frontend: keep insufficient funds error for large amount

bznein:
When the user enters an amount that is both too much to get a quote, and mroe than they have, instead of only showing the "no routes found" error, also show the insufficientFundsError.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/api/swap.ts
+++ b/frontends/web/src/api/swap.ts
@@ -42,6 +42,14 @@ export type TSwapQuoteRoute = {
   routeId: string;
 };
 
+export type TSwapQuoteErrorCode =
+  | 'insufficientFunds'
+  | 'invalidRequest'
+  | 'noRoutesFound'
+  | 'unexpectedError';
+
+export type TSwapQuoteValidationErrorCode = Extract<TSwapQuoteErrorCode, 'insufficientFunds'>;
+
 export type TSwapQuoteResponse = {
   success: true;
   quote: {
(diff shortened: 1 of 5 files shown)


thisconnect:
@coderabbitai review

thisconnect (approved):
frontend tested LGTM

nit: I guess in the frontend  validationErrorCode would be sufficient and we don't really need validationErrorMessage

bznein:
> nit: I guess in the frontend validationErrorCode would be sufficient and we don't really need validationErrorMessage

Yeah you are right, no need to pollute the api with more fields than needed, removed

frontends/web/src/api/swap.ts
```diff
@@ -56,6 +56,7 @@ export type TSwapQuoteResponse = {
     sellCoin?: string;
   };
   errorMessage: string;
+  validationErrorCode?: string;
```
benma: Could/should this be typed like ` : `insufficientFunds' | 'NoRoutesFoundError' 'unexpectedError';? for clarity and type safety?

Also

```
const INSUFFICIENT_FUNDS_ERROR = 'insufficientFunds';
const NO_ROUTES_FOUND_ERROR = 'NoRoutesFoundError';
const UNEXPECTED_ERROR = 'unexpectedError';
```

Why are all camelCase but one is PascalCase? 😄

bznein: Both good points :) Swtiched vall to camelCase for consistency and added stronger typing for the error strings

bznein:
@benma ping (if it is too late to make it into release, I can change the base branch)

benma (approved):
utACK

PR #4115, opened 2026-05-07
Frontend update localize for node24

thisconnect:


shonsirsha (approved):
LGTM. Tested on node 24

PR #4116, opened 2026-05-07
frontend: add BitBox screen when checking backup

bznein:
When checking the validity of a backup, add the graphics that points the user towards checking the BitBox screen so they know something is happening on the device.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/device/bitbox02/checkbackup.tsx
+++ b/frontends/web/src/routes/device/bitbox02/checkbackup.tsx
@@ -7,6 +7,7 @@ import { Backup } from '@/api/backup';
 import { alertUser } from '@/components/alert/Alert';
 import { Dialog, DialogButtons } from '@/components/dialog/dialog';
 import { Button } from '@/components/forms';
+import { PointToBitBox02 } from '@/components/icon';
 import { useTranslation } from 'react-i18next';
 
 type TProps = {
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/device/bitbox02/checkbackup.tsx
```diff
@@ -74,6 +77,12 @@ export const Check = ({ deviceID, backups, disabled }: TProps) => {
           {message && (
             <p>{message}</p>
           )}
+          {isWaitingForDevice && (
+            <>
+              <p className="text-center">{t('confirm.info')}</p>
+              <PointToBitBox02 />
+            </>
+          )}
```
thisconnect, line 90: Small design nit: almost everywhere (with the exception on send-confirm) the order is title->some-info->point-to-device. To force the user to read first and last see the point-to-device graphic as at this point the user should focus on the device (with all the info in mind).

Please move the confirm.info + PointToBitBox02 after the found backup list item a few lines below.

then it will look like this:

Image

thisconnect (approved):
tested LGTM with suggestion

Also since you are here you could add the point-2-device also to create backup:

Image

frontends/web/src/routes/device/bitbox02/checkbackup.tsx
bznein: done!

PR #4123, opened 2026-05-11
ipados: fix tap highlight on hamburger (menu) btn

shonsirsha:
On some iPads like the 11 inch ones, the sidebar are collapsed and user can toggle it using the hamburger menu button.
However, its tap highlight color looks rather crooked. In this commit, we remove the tap highlight color, make the icon react to it instead.


thisconnect (approved):
untested LGTM

PR #4135, opened 2026-05-12
backend: add btc-only software test keystore

bznein:
Allow the software test keystore to be started as either Multi or BTC-only.

The test wallet unlock dialog now exposes a BTC-only checkbox and sends the selected edition to the backend. The software keystore defaults to Multi for existing callers, while BTC-only mode only supports BTC/TBTC/RBTC and rejects ETH signing paths.

This is useful for testing BTC-only features without having to on on mainnet or starting regtest.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -970,9 +970,17 @@ func (backend *Backend) CheckElectrumServer(serverInfo *config.ServerInfo) error
 
 // RegisterTestKeystore adds a keystore derived deterministically from a PIN, for convenience in
 // devmode.
-func (backend *Backend) RegisterTestKeystore(pin string) {
-	softwareBasedKeystore := software.NewKeystoreFromPIN(pin)
+func (backend *Backend) RegisterTestKeystore(pin string, edition software.Edition) error {
+	switch edition {
+	case "":
+		edition = software.EditionMulti
+	case software.EditionMulti, software.EditionBTCOnly:
+	default:
+		return errp.Newf("Unsupported test keystore edition: %s", edition)
+	}
+	softwareBasedKeystore := software.NewKeystoreFromPINWithEdition(pin, edition)
 	backend.registerKeystore(softwareBasedKeystore)
+	return nil
 }
 
 // NotifyUser creates a desktop notification.
--- a/backend/keystore/software/software.go
+++ b/backend/keystore/software/software.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/btc/types"
-	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
+	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/eth"
 	keystorePkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/keystore"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/signing"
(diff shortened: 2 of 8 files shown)


bznein:
@coderabbit review

frontends/web/src/locales/en/app.json
```diff
@@ -2038,6 +2038,7 @@
       "title": "Disconnect test wallet"
     },
     "prompt": {
+      "btcOnlyLabel": "BTC-only",
```
thisconnect: nit: we don't need to translate this string, but as it is user facing I would change to  "Bitcoin-only" instead of "BTC-only". So it aligns with the editions we sell.

thisconnect (approved):
frontend tested in webdev and QT LGTM
with small nit

backend/keystore/software/software.go
```diff
@@ -294,6 +325,9 @@ func (keystore *Keystore) SignTransaction(
 	case *btc.ProposedTransaction:
 		return keystore.signBTCTransaction(specificProposedTx)
 	case *eth.TxProposal:
+		if keystore.edition == EditionBTCOnly {
+			return errp.Newf("coin not supported: %s", specificProposedTx.Coin.Code())
+		}
 		return keystore.signETHTransaction(specificProposedTx)
```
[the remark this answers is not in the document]
bznein: Fixed

backend/keystore/software/software.go
```diff
@@ -40,30 +40,58 @@ type Keystore struct {
 	// The master extended private key from which all keys are derived.
 	master     *hdkeychain.ExtendedKey
 	identifier string
+	edition    Edition
 	log        *logrus.Entry
 }
 
+// Edition models the BitBox02 firmware edition emulated by the software keystore.
```
benma: I would not say that the software keystore emulates BitBox02. If we wanted that it could be powered by a simulator.

It may still be useful to have a BTC-only software keystore, but I would not mention BitBox02 in this context.

backend/keystore/software/software.go
```diff
@@ -40,30 +40,58 @@ type Keystore struct {
...
 	return &Keystore{
 		master:     master,
 		identifier: hex.EncodeToString(hash[:]),
+		edition:    edition,
 		log:        logging.Get().WithGroup("software"),
 	}
 }
 
 // NewKeystoreFromPIN creates a new unique keystore derived from the PIN.
 func NewKeystoreFromPIN(pin string) *Keystore {
```
benma: Unused? If so, please remove.

backend/keystore/software/software.go
```diff
@@ -103,9 +131,12 @@ func (keystore *Keystore) Configuration() *signing.Configuration {
 }
 
 // SupportsCoin implements keystore.Keystore.
-func (keystore *Keystore) SupportsCoin(coin coin.Coin) bool {
-	switch coin.(type) {
+func (keystore *Keystore) SupportsCoin(coinInstance coin.Coin) bool {
```
benma: we usually do `coin coinpkg.Coin`, not sure which is better.

backend/keystore/software/software.go
```diff
@@ -103,9 +131,12 @@ func (keystore *Keystore) Configuration() *signing.Configuration {
...
+	switch coinInstance.(type) {
 	case *btc.Coin, *eth.Coin:
+		if keystore.edition == EditionBTCOnly {
```
benma: Shouldn't this be outside the switch statement? If we added a third coin type it would continue to work.

backend/backend.go
```diff
@@ -968,11 +968,31 @@ func (backend *Backend) CheckElectrumServer(serverInfo *config.ServerInfo) error
...
+// TestKeystoreEdition models the software test keystore edition.
+type TestKeystoreEdition string
+
+const (
+	// TestKeystoreEditionMulti emulates a BitBox02 Multi.
+	TestKeystoreEditionMulti TestKeystoreEdition = "multi"
+	// TestKeystoreEditionBTCOnly emulates a BitBox02 BTC-only.
+	TestKeystoreEditionBTCOnly TestKeystoreEdition = "btc-only"
+)
+
```
benma: Redundant with software.Edition, remove?

frontends/web/src/locales/en/app.json
```diff
@@ -2038,6 +2038,7 @@
       "title": "Disconnect test wallet"
     },
     "prompt": {
+      "btcOnlyLabel": "Bitcoin-only",
```
benma: I'd put this under `generic`, sounds re-usable 😄

frontends/web/src/routes/device/components/skipfortesting.tsx
```diff
@@ -48,6 +50,12 @@ export const SkipForTesting = ({
             autoFocus
             label={t('testWallet.prompt.passwordLabel')}
             onValidPassword={(pw) => setTestPIN(pw ? pw : '')}/>
+          <Checkbox
+            id="test-wallet-btc-only"
```
benma: Is the ID needed?

frontends/web/src/routes/device/components/skipfortesting.tsx
```diff
@@ -48,6 +50,12 @@ export const SkipForTesting = ({
...
+            checked={btcOnly}
+            onChange={e => setBTCOnly((e.target as HTMLInputElement).checked)}
+            label={t('testWallet.prompt.btcOnlyLabel', { defaultValue: 'BTC-only' })}
```
benma: No need for defaultValue, it falls back to English automatically.

benma (changes requested):

frontends/web/src/routes/device/components/skipfortesting.tsx
bznein: Not really, we might need "data-testid" if we want to write playwright tests for it, but probably overkill (and id would not be the best field, as mentioned)

bznein (commented):
Thanks @benma, addressed your comments!

frontends/web/src/routes/device/components/skipfortesting.tsx
bznein: Actually it is needed by the Checkbox type, so I'd rather leave this change in, instead of having to modify checkbox.tsx

frontends/web/src/locales/en/app.json
thisconnect: I think we don't need to translate the term "Bitcoin-only" and it can just be a hardcoded string in the react components

frontends/web/src/routes/device/components/skipfortesting.tsx
thisconnect: having an id attribute is great as this is also used on the label element in the checkbox component.

With this the user can also click on the label and it will check the checkbox with the same id.

benma (approved):
utACK

PR #4137, opened 2026-05-12
backend: encode empty balances as array

benma:
Unplugging a BitBox can briefly leave the frontend with the previous account list while /accounts/balance-summary is computed after account teardown.

A nil Go slice encoded coinsTotalBalance as null, but the frontend treats it as an array and dereferences .length, blanking the screen.

Initialize the slice so the JSON response is [] during that empty-account race window.


bznein (approved):
utACK

PR #4145, opened 2026-05-14
frontend: script check and fix sorting translation keys

thisconnect:
expected to fail as some keys are not sorted in en/app.json


thisconnect:
@coderabbitai review

frontends/web/scripts/sort-json.mjs
```diff
@@ -15,6 +18,7 @@
  * - Pretty prints with 2-space indentation
  * - Adds trailing newline
  * - Deterministic sorting
+ * - --check mode exits with code 1 instead of modifying files
```
bznein: Can we make this clearer? I assume it exits 0/1 depending on whether it is ordered correctly or not :)

frontends/web/scripts/sort-json.mjs
```diff
@@ -109,14 +131,26 @@ const main = () => {
...
   }
 
-  files.forEach(processFile);
+  let hasChanges = false;
+
+  files.forEach((file) => {
+    const changed = processFile(file, checkMode);
+
+    if (changed) {
+      hasChanges = true;
```
bznein, line 172: Not that performance really matters here, but we could already exit here if checkMode is true.

However, I think it would be even better if we could log all the files that are not sorted, what do you think?

bznein (commented):
overall looks good to me, just a couple of small comments

frontends/web/scripts/sort-json.mjs
thisconnect: Yes exactly this script only cares about "Sort all keys in JSON files alphabetically (recursively)". It should have same behavior as we had before locize was removed in https://github.com/thisconnect/bitbox-wallet-app/commit/b6931de65a45996ef83a4a292e50750ee809b5f5 

I'll update the docstring

frontends/web/scripts/sort-json.mjs
thisconnect: >  I think it would be even better if we could log all the files that are not sorted, what do you think?

It should already list all files that are not sorted, same behavior as with locize-cli.

This scripts could be slightly faster (maybe) by not using sync file api, but I think the perf gain is negligible

bznein: Ah right, sorry, I missed that the log line is inside the function that processes a single file, rather than here !

bznein (approved):
utACK

PR #4146, opened 2026-05-14
lightning: encrypt stored seed

Beerosagos:
Add platform hooks for storing a Lightning seed encryption key outside the backend config and use them on Android via the keystore.

Activation now derives the mnemonic as before, generates a random 32-byte key in the backend, encrypts the mnemonic with AES-GCM, stores only the ciphertext in the Lightning config, and hands the base64-encoded key to the platform environment.

On Android, the environment wraps that key with an account-specific AES key from AndroidKeyStore and stores the wrapped blob in SharedPreferences. Unlocking loads and unwraps the stored key, decrypts the mnemonic in the backend, and then connects Breez. Deactivation removes both the wrapped key and the encrypted mnemonic.

Android intentionally treats missing wrapped keys as requiring Lightning reactivation. This can affect existing dev wallets created before this change, but the feature has no production users yet and the migration cost is acceptable at this stage.

Other platforms keep returning false or stub implementations for now. This is intentional preparatory work so iOS, Qt, and web can adopt secure storage later without another backend API change.

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos:
@coderabbit review

backend/bridgecommon/bridgecommon.go
```diff
@@ -265,6 +269,38 @@ func (env *BackendEnvironment) OnAuthSettingChanged(enabled bool) {
...
+	return nil
+}
+
+// LoadLightningEncryptionKey implements backend.Environment.
+func (env *BackendEnvironment) LoadLightningEncryptionKey(accountCode string) (string, error) {
+	if env.LoadLightningEncryptionKeyFunc != nil {
+		return env.LoadLightningEncryptionKeyFunc(accountCode)
+	}
+	return "", nil
+}
```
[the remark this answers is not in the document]
Beerosagos: We need this to allow using the seed without encryption in webdev mode. Lightning activation will be hidden in the UI for all the other platforms that don't support encryption in a future PR. I'll keep it as is for now.

backend/lightning/lightning.go
```diff
@@ -252,6 +294,96 @@ func (lightning *Lightning) connect(_ bool) error {
...
+		return "", errp.New(lightningReactivationRequiredMessage)
+	}
+
+	mnemonic, err := decryptMnemonic(account.Mnemonic, encryptionKey)
+	if err != nil {
+		return "", errp.New(lightningReactivationRequiredMessage)
+	}
+
+	return mnemonic, nil
+}
```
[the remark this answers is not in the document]
Beerosagos: Added in the error handling or seal/unseal callers

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/LightningEncryptionHelper.java
```diff
@@ -0,0 +1,103 @@
...
+            throw new Exception(ERR_KEY_MISSING);
+        }
+
+        SecretKey wrappingKey = getOrCreateWrappingKey(accountCode);
+        byte[] wrappedKey = Base64.decode(wrappedKeyBase64, Base64.DEFAULT);
+        Cipher cipher = Cipher.getInstance("AES/GCM/NoPadding");
+        byte[] iv = new byte[12];
+        System.arraycopy(wrappedKey, 0, iv, 0, iv.length);
+        byte[] ciphertext = new byte[wrappedKey.length - iv.length];
+        System.arraycopy(wrappedKey, iv.length, ciphertext, 0, ciphertext.length);
```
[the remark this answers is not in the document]
Beerosagos: Added

backend/lightning/lightning.go
```diff
@@ -149,6 +178,12 @@ func (lightning *Lightning) Deactivate() error {
 		return err
 	}
 
+	if lightning.environment.CanEncryptLightningMnemonic() {
+		if err := lightning.environment.DeleteLightningEncryptionKey(string(account.Code)); err != nil {
+			lightning.log.WithError(err).Warn("Error deleting lightning encryption key")
```
benma, line 186: Why not return the error? Seems like a serious enough error.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/LightningEncryptionHelper.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private static String getKeyAlias(String accountCode) {
+        return "bitboxapp.lightning.seed." + accountCode + ".v1";
+    }
+
+    private static SharedPreferences preferences(Context context) {
+        return context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE);
+    }
+
+    private static SecretKey getOrCreateWrappingKey(String accountCode) throws Exception {
```
benma: Codex:

Restore path can strand Lightning accounts.
     frontends/android/BitBoxApp/app/src/main/AndroidManifest.xml:32 enables backup, and Android backup includes app data/shared prefs by default. If lightning.json and lightning-encryption.xml restore without
     the AndroidKeyStore entry, loadKey() still calls getOrCreateWrappingKey() at frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/LightningEncryptionHelper.java:82, creating a fresh wrong
     key instead of reporting “missing key”. Decryption then fails generically. I’d split create/load paths: load should error if the alias is absent, and backup rules should likely exclude the Lightning prefs
     and/or lightning.json.

backend/lightning/lightning.go
Beerosagos: Well, even if the encryption key cannot be deleted for any reason, the deactivation basically completes anyway. And I don't think it should bring any issue on a following new activation, unless the Android keystore is completely broken. So I think that returning an error may add more confusion than value

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/LightningEncryptionHelper.java
Beerosagos: Great catch! Should be fixed now 🙏

Beerosagos:
@benma PTAL :pray:

frontends/android/BitBoxApp/app/src/main/res/xml/backup_rules.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <exclude domain="file" path="lightning.json" />
```
benma, line 4: Codex follow-up:
---
Android backup rules still allow Breez storage backup. Android starts the backend in
    getFilesDir() at frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/
    MainActivity.java:196, and the backend creates cache/ under that directory at backend/arguments/
    arguments.go:73. Breez then uses cache/breez-* as StorageDir at backend/lightning/lightning.go:231,
    but the new backup rules exclude only lightning.json and the wrapped-key prefs at frontends/android/
    BitBoxApp/app/src/main/res/xml/backup_rules.xml:3. If Breez storage contains wallet state or
    sensitive metadata, it remains eligible for cloud backup/device transfer. Exclude the Breez/cache
    path or place it under Android no-backup/cache storage.

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/LightningEncryptionHelper.java
```diff
@@ -0,0 +1,128 @@
...
+import android.util.Base64;
+
+import java.security.KeyStore;
+
+import javax.crypto.Cipher;
+import javax.crypto.KeyGenerator;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+
+public class LightningEncryptionHelper {
```
benma, line 24: Docstrings on the class or on the methods would be helpful, it is not trivial to see what the functions do and which keys they handle without reading the function bodies.

backend/lightning/lightning.go
```diff
@@ -91,13 +108,25 @@ func (lightning *Lightning) Activate() error {
...
+	accountCode := types.Code(strings.Join([]string{"v0-", hex.EncodeToString(fingerprint), "-ln-0"}, ""))
+	sealedMnemonic, err := lightning.sealMnemonic(string(accountCode), entropyMnemonic)
+	if err != nil {
+		lightning.log.WithError(err).Warn("Error configuring Lightning secure storage")
+		return errp.New("Could not configure Lightning secure storage on this device")
+	}
+
 	lightningAccount := config.LightningAccountConfig{
-		Mnemonic:        entropyMnemonic,
+		Mnemonic:        sealedMnemonic,
```
benma: i would rename the `Mnemonic` field name too, as it sounds like it would contain a mnemonic. Also please add a docstring  to the struct field about what is stored in it.

It also seems that the memonic field value can contain the sealed mnemonic or the plaintext memonic depending on CanEncryptLightningMnemonic. If this split will remain for a while, I'd use two different fields, or a union-like struct for clarity.

backend/lightning/lightning.go
```diff
@@ -252,6 +294,96 @@ func (lightning *Lightning) connect(_ bool) error {
...
+	gcm, err := cipher.NewGCM(block)
+	if err != nil {
+		return "", err
+	}
+	nonce := make([]byte, gcm.NonceSize())
+	if _, err := io.ReadFull(rand.Reader, nonce); err != nil {
+		return "", err
+	}
+	ciphertext := gcm.Seal(nil, nonce, []byte(mnemonic), nil)
+	return base64.StdEncoding.EncodeToString(append(nonce, ciphertext...)), nil
```
benma: How about prepending a `0x00` version byte in front of the bytes, or a `v1:` version string prefix to to the base64 encoding? Then potential migrations will be simpler/clearer.

backend/lightning/lightning.go
Beerosagos: The plaintext mnemonic is just for webdev mode: I think it's useful to be able to read it in the config if you are developing. Ack to just renaming it to `Seed`?

Beerosagos:
@coderabbit review

Beerosagos:
@benma PTAL :pray:

benma (approved):
utACK

PR #4151, opened 2026-05-17
frontend: settings search

shonsirsha:
Adds search to the settings area so users can find settings without manually scanning each tab.

Behaviour:
- Typing in the settings search field filters matches across General, Manage device, Advanced settings, and About.
- Results are grouped by destination page and navigate directly to the matching settings page with the item highlighted (border flashes 3x).
- Device-only settings are shown only when a device is available. Same with bluetooth, their entries are also gated by device Bluetooth support.

Technical notes:
- Introduces shared settings search descriptors, grouping/filtering helpers, and unit coverage for matching and availability rules.
- Adds the reusable clearable SearchInput variant and keeps the search term in the URL query params.

### DIFF
--- a/frontends/web/src/components/forms/search-input.tsx
+++ b/frontends/web/src/components/forms/search-input.tsx
@@ -1,13 +1,61 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { forwardRef } from 'react';
-import { Loupe } from '@/components/icon';
+import { useTranslation } from 'react-i18next';
+import { CloseXDark, CloseXWhite, Loupe } from '@/components/icon';
 import { Input } from './input';
 import type { TInputProps } from './input';
 import styles from './search-input.module.css';
 
-export const SearchInput = forwardRef<HTMLInputElement, TInputProps>((props, ref) => (
-  <Input ref={ref} type="text" {...props}>
-    <Loupe className={styles.searchIcon} />
-  </Input>
-));
+type TBaseProps = Omit<TInputProps, 'children' | 'type'>;
+
+type TClearProps = TBaseProps & {
+  onClear: () => void;
+  variant: 'clear';
+};
+
+type TSearchProps = TBaseProps & {
+  onClear?: never;
+  variant?: 'search';
+};
+
+type TProps = TClearProps | TSearchProps;
+
+export const SearchInput = forwardRef<HTMLInputElement, TProps>((props, ref) => {
+  const { t } = useTranslation();
+
+  if (props.variant === 'clear') {
+    const {
+      onClear,
+      variant: _variant,
+      ...inputProps
+    } = props;
+
+    return (
+      <Input ref={ref} type="text" {...inputProps}>
+        {String(inputProps.value ?? '').trim().length > 0 ? (
+          <button
+            aria-label={t('generic.close')}
+            className={styles.clearButton}
+            onClick={onClear}
+            type="button"
+          >
+            <CloseXDark className="show-in-lightmode" />
+            <CloseXWhite className="show-in-darkmode" />
+          </button>
+        ) : null}
+      </Input>
+    );
+  }
+
+  const {
+    variant: _variant,
+    ...inputProps
+  } = props;
+
+  return (
+    <Input ref={ref} type="text" {...inputProps}>
+      <Loupe className={styles.searchIcon} />
+    </Input>
+  );
+});
--- a/frontends/web/src/routes/settings/components/settings-content.tsx
+++ b/frontends/web/src/routes/settings/components/settings-content.tsx
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { Fragment, ReactNode, useEffect, useRef } from 'react';
+import { useSettingsHighlight } from '../use-settings-highlight';
+import styles from './settings-content.module.css';
+
+export type TSettingsContentItem = {
+  content: ReactNode;
+  id: string;
+};
+
+export type TSettingsContentSection = {
+  id: string;
+  items: TSettingsContentItem[];
+  title?: ReactNode;
+};
+
+type TProps = {
+  sections: TSettingsContentSection[];
+};
+
+export const SettingsContent = ({
+  sections,
+}: TProps) => {
+  const highlightedItemID = useSettingsHighlight();
+  const highlightedItemRef = useRef<HTMLDivElement | null>(null);
+  const scrolledItemID = useRef<string>();
+
+  useEffect(() => {
+    if (!highlightedItemID) {
+      scrolledItemID.current = undefined;
+      return;
+    }
+
+    if (!highlightedItemRef.current || scrolledItemID.current === highlightedItemID) {
+      return;
+    }
+
+    const scrollToHighlightedItem = () => {
+      highlightedItemRef.current?.scrollIntoView({
+        behavior: 'smooth',
+        block: 'center',
+      });
+    };
+
+    // waits for the page layout to settle
+    // after navigation before centering the item.
+    window.setTimeout(scrollToHighlightedItem, 150);
+    scrolledItemID.current = highlightedItemID;
+  });
+
+  return (
+    sections.map(section => {
+      if (section.items.length === 0) {
+        return null;
+      }
+
+      return (
+        <Fragment key={section.id}>
+          {section.title ? section.title : null}
+          {section.items.map(settingItem => {
+            const isHighlighted = settingItem.id === highlightedItemID;
+
+            return (
+              <div
+                className={isHighlighted ? styles.highlightedItem : undefined}
+                key={settingItem.id}
+                ref={isHighlighted ? highlightedItemRef : undefined}
+              >
+                {settingItem.content}
+              </div>
+            );
+          })}
+        </Fragment>
+      );
+    })
+  );
+};
--- a/frontends/web/src/routes/settings/components/settings-search-content.tsx
+++ b/frontends/web/src/routes/settings/components/settings-search-content.tsx
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { useTranslation } from 'react-i18next';
+import { useNavigate } from 'react-router-dom';
+import type { TDevices } from '@/api/devices';
+import { SubTitle } from '@/components/title';
+import {
+  SETTINGS_HIGHLIGHT_QUERY_PARAM,
+  SETTINGS_SEARCH_PAGE_ORDER,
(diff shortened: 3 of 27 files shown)


shonsirsha:
@coderabbitai review!

frontends/web/src/routes/settings/components/settings-search-content.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export const SettingsSearchContent = ({
+  devices,
+  hasAccounts,
+  searchResults,
+}: TProps) => {
+  const { t } = useTranslation();
+  const firstDeviceID = Object.keys(devices)[0];
+
```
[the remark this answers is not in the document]
shonsirsha: Invalid. We're only targetting first device for now.

frontends/web/src/routes/settings/settings-search.ts
```diff
@@ -0,0 +1,273 @@
...
+  {
+    id: 'device-name',
+    isAvailable: whenDeviceExists,
+    getTitle: ({ t }) => t('bitbox02Settings.deviceName.input'),
+    page: 'device',
+  },
+  {
+    id: 'bluetooth',
+    isAvailable: whenDeviceExists,
+    getTitle: ({ t }) => t('bitbox02Settings.bluetoothToggleEnabled.titleEnabled'),
```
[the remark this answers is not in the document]
shonsirsha: Skipped as invalid.

settings-search.ts only gets TDevices, which is just { deviceID: platformName }. Bluetooth support lives in DeviceInfo.bluetooth, loaded later in bb02-settings.tsx.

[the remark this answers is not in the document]
shonsirsha: Actually this is valid but needed a way so getSettingsSearchItems can accept the deviceInfo. Fixed now.

shonsirsha:
@coderabbitai review!

frontends/web/src/routes/settings/components/settings-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const scrollToHighlightedItem = () => {
+      highlightedItemRef.current?.scrollIntoView({
+        behavior: 'smooth',
+        block: 'center',
+      });
+    };
+
+    [0, 100, 300].forEach(delay => {
+      window.setTimeout(scrollToHighlightedItem, delay);
```
thisconnect: please add a comment why the 0, 100 and 300 delay.

Also I am not sure where when and how the scrolling works, could you mention in the commit message how to test or what the intention is?

frontends/web/src/routes/settings/components/settings-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type TProps = {
+  sections: TSettingsContentSection[];
+};
+
+export const SettingsContent = ({
+  sections,
+}: TProps) => {
+  const highlightedItemID = useSettingsHighlight();
+  const highlightedItemRef = useRef<HTMLDivElement | null>(null);
```
thisconnect, line 26: not sure what highlightedItemRef is used for, it seems to be forever `null` maybe leftover code, remove?

frontends/web/src/routes/settings/components/settings-content.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    [0, 100, 300].forEach(delay => {
+      window.setTimeout(scrollToHighlightedItem, delay);
+    });
+    scrolledItemID.current = highlightedItemID;
+  });
+
+  return (
+    <>
+      {sections.map(section => {
```
thisconnect: nit: outer fragment has no purpose.
```suggestion
      return sections.map(section => {
```

frontends/web/src/routes/settings/use-settings-search.ts
```diff
@@ -0,0 +1,72 @@
...
+  const showSearchResults = !!searchTerm.trim();
+
+  useEffect(() => {
+    setSearchTerm(searchTermFromParams);
+  }, [searchTermFromParams]);
+
+  const updateSearchTerm = (value: string) => {
+    setSearchTerm(value);
+    const nextSearchParams = new URLSearchParams(searchParams);
+    if (value.toLowerCase().trim()) {
```
thisconnect:
```suggestion
    if (value.trim()) {
```

frontends/web/src/routes/settings/use-settings-search.ts
```diff
@@ -0,0 +1,72 @@
...
+  }), [deviceInfo, devices, hasAccounts, isTesting, t]);
+  const searchResults = useMemo(
+    () => filterSettingsSearchItems(searchItems, searchTerm),
+    [searchItems, searchTerm],
+  );
+  const showSearchResults = !!searchTerm.trim();
+
+  useEffect(() => {
+    setSearchTerm(searchTermFromParams);
+  }, [searchTermFromParams]);
```
thisconnect: The next function takes a value, sets setSearchTerm, but then uses searchParams for nextSearchParams.  I find it a bit hard mentally to follow.

GPT comment on this effect:
> That’s usually fine, but it creates dual sources of truth: React state / URL params
>
> You could simplify by making URL params the sole source of truth and removing local state entirely.

I think with that this hook would become a bit simpler to follow.

frontends/web/src/components/forms/search-input.tsx
```diff
@@ -1,13 +1,61 @@
...
+      onClear,
+      variant: _variant,
+      ...inputProps
+    } = props;
+
+    return (
+      <Input ref={ref} type="text" {...inputProps}>
+        {String(inputProps.value ?? '').trim().length > 0 ? (
+          <button
+            aria-label={clearButtonLabel}
```
thisconnect: nit: just hardcode `aria-label={t('generic.close')}` and remove the prop.

frontends/web/src/components/forms/search-input.tsx
```diff
@@ -1,13 +1,61 @@
...
+
+type TSearchProps = TBaseProps & {
+  clearButtonLabel?: never;
+  onClear?: never;
+  variant?: 'search';
+};
+
+type TProps = TClearProps | TSearchProps;
+
+export const SearchInput = forwardRef<HTMLInputElement, TProps>((props, ref) => {
```
thisconnect, line 24: With this PR there are now different "variant's" so I checked the other 2 places where `<SearchInput>` is used.

Not directly related to this PR, but in the other 2 places 'ref' is used to scroll to the results, but in both places the original list is replaced by the actually found results. Sometimes it quickly starts scrolling but then immediately updates the list with the found results stopping the scrolling a spit second later.

The search in settings has scrolling but doesn't need this ref to do so.

So it looks to me that we could drop ref here and the scrolling in address list and account overview.

frontends/web/src/routes/settings/components/tabs.tsx
```diff
@@ -50,17 +54,48 @@ export const WithSettingsTabs = ({
...
-          hideMobileMenu={hideMobileMenu}
-          devices={devices}
-          hasAccounts={hasAccounts}
+      <div className={styles.searchContainer}>
+        <SearchInput
+          autoFocus={!isMobile}
+          clearButtonLabel={t('generic.close')}
+          onChange={event => updateSearchTerm(event.currentTarget.value)}
+          onClear={() => updateSearchTerm('')}
+          placeholder={t('generic.search')}
```
thisconnect: The  placeholder of this input field is "Search…" (`t('generic.search')`), in account it is hardcoded to "Search transactions..." (no translation) and the one in address-list is "Search address" (`t('addresses.searchPlaceholder')`).

Not a blocker, but would be nice to translate the account search.

frontends/web/src/routes/settings/settings-search.ts
```diff
@@ -0,0 +1,287 @@
...
+  id: string;
+  page: TSettingsSearchPage;
+  getTitle: (context: TSettingsSearchContext) => string;
+  isAvailable?: (context: TSettingsSearchContext) => boolean;
+};
+
+
+const whenAccountsExist = ({ hasAccounts }: TSettingsSearchContext) => hasAccounts;
+const whenDeviceExists = ({ deviceIDs }: TSettingsSearchContext) => deviceIDs.length > 0;
+const whenDeviceSupportsBluetooth = ({ deviceInfo }: TSettingsSearchContext) => !!deviceInfo?.bluetooth;
```
thisconnect: It is not ideal to have to maintain the same logic for showing setting-items in two times, in search results (here) and for the actual settings.

Would it make sense to import these functions to use the code to conditionally render the settings items i.e.  in bb02-settings etc.?

thisconnect (changes requested):

frontends/web/src/routes/settings/components/settings-content.tsx
shonsirsha: Was just for redundancy (retries). One 150ms should do the trick instead of retry. Thanks.

frontends/web/src/routes/settings/components/settings-content.tsx
shonsirsha: It’s used to call scrollIntoView() for the clicked search result. So we'd need to keep it.

frontends/web/src/routes/settings/use-settings-search.ts
shonsirsha: Thank you

frontends/web/src/routes/settings/settings-search.ts
shonsirsha: Good point. Will clean this up by moving the shared "visibility checks" into a separate file.

frontends/web/src/components/forms/search-input.tsx
thisconnect: still needed for focusing

frontends/web/src/routes/settings/settings-search.ts
```diff
@@ -0,0 +1,282 @@
...
+export const filterSettingsSearchItems = (
+  searchItems: TSettingsSearchItem[],
+  searchTerm: string,
+): TSettingsSearchItem[] => {
+  const normalizedSearchTerm = searchTerm.trim().toLowerCase();
+  if (!normalizedSearchTerm) {
+    return [];
+  }
+  return searchItems.filter(searchItem => (
+    searchItem.title.trim().toLowerCase().includes(normalizedSearchTerm)
```
thisconnect, line 270: at the moment it only searches the term in titles, but could we also search in description string? 

but not a blocker, just a nice to have imo

thisconnect (approved):
untested LGTM

PR #4156, opened 2026-05-19
ios: refresh widget prices more often

shonsirsha:
The widget now asks iOS to refresh about every 10 minutes.

When iOS grants a refresh, the widget reuses cached price data only if it is less than 10 minutes old, instead of the previous 1 hour.

### DIFF
--- a/frontends/ios/BitBoxApp/BitBoxAppWidget/WidgetTimelineProvider.swift
+++ b/frontends/ios/BitBoxApp/BitBoxAppWidget/WidgetTimelineProvider.swift
@@ -2,6 +2,9 @@ import Foundation
 import WidgetKit
 
 struct Provider: TimelineProvider {
+    private static let refreshInterval: TimeInterval = 10 * 60
+    private static let retryInterval: TimeInterval = 60
+
     private let dataService = WidgetDataService()
 
     func placeholder(in context: Context) -> SimpleEntry {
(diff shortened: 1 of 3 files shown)


frontends/ios/BitBoxApp/BitBoxAppWidget/WidgetTimelineProvider.swift
```diff
@@ -48,15 +50,14 @@ struct Provider: TimelineProvider {
...
             let entry = resolvedEntry(for: coinCode, currency: currency, data: exactCacheHit)
-            let nextUpdate = Date().addingTimeInterval(15 * 60)
+            let nextUpdate = Date().addingTimeInterval(Self.refreshInterval)
             completion(Timeline(entries: [entry], policy: .after(nextUpdate)))
         } else {
             Task {
                 let fetched = await dataService.fetchChartData(coinCode: coinCode, currency: currency)
                 let data = fetched ?? dataService.cachedFallback(for: coinCode, currency: currency)
                 let entry = resolvedEntry(for: coinCode, currency: currency, data: data)
-                let retryMinutes = fetched == nil ? 1 : 15
```
benma, line 58: If it fails it now waits 10 min instead of 1 to retry. Unintended change?

frontends/ios/WIDGET.md
```diff
@@ -100,8 +100,9 @@ If the network fetch fails:
 
 ## 5. Refresh schedule
 
-- iOS calls the widget's timeline roughly **every 15 minutes**. This is not exact - iOS controls the actual timing based on widget budget, battery, and usage patterns.
-- If a network fetch fails, the widget asks iOS to retry in **1 minute**.
```
benma, line 104: Why not retry in 1min?

shonsirsha:
Good catch, thanks. PTAL @benma .

benma (approved):
utACK

PR #4158, opened 2026-05-19
backend: extract account derivation specs

benma:


strmci (approved):
nit: consider renaming `BIP44CoinType` to `HardenedBIP44CoinType`. It returns the hardened keypath element rather than the SLIP-44 coin type. 

Besides that LGTM.

benma:
> nit: consider renaming `BIP44CoinType` to `HardenedBIP44CoinType`. It returns the hardened keypath element rather than the SLIP-44 coin type.
> 
> Besides that LGTM.

Good point. I changed the map to define the unhardened numbers instead. Please check.

strmci (approved):
LGTM

PR #4160, opened 2026-05-20
ios: widget price fix

shonsirsha:


frontends/ios/BitBoxApp/BitBoxAppWidget/WidgetTimelineProvider.swift
```diff
@@ -45,8 +45,9 @@ struct Provider: TimelineProvider {
     func getTimeline(in context: Context, completion: @escaping (Timeline<SimpleEntry>) -> Void) {
         let coinCode = WidgetAppGroupStore.selectedCoinCode()
         let currency = WidgetAppGroupStore.userCurrency()
+        let forceFreshPriceReload = WidgetAppGroupStore.consumeForceFreshPriceReload()
```
benma: Codex:
---
consumeForceFreshPriceReload() clears the force flag before the fresh fetch succeeds. If a lifecycle reload has a valid cache but the
    network fetch fails, line 62 schedules a 1-minute retry, but the retry will no longer bypass the cache and will immediately take the exact-cache path, pushing the next refresh out by 10 minutes. That makes the
    retry policy ineffective for the new forced-refresh path. Either keep the force flag pending until a fetch succeeds, or schedule the normal interval when falling back to cache after the single forced attempt.

shonsirsha: Mmh good catch. Thanks

benma (approved):
utACK

PR #4168, opened 2026-05-27
eth: speed up active account transaction discovery

Tomasvrba:
Fix to an old issue: https://github.com/BitBoxSwiss/bitbox-wallet-app/issues/837

Incoming ETH transactions are currently only discovered during the regular Etherscan history sync, which runs every five minutes to stay conservative with API usage. That means a mined incoming transaction can remain invisible for minutes even though Ethereum blocks are produced much faster.

Add a short-lived foreground activity lease so the backend can identify ETH accounts the user is actively viewing. While such a lease is active, cheaply probe the remote balance and trigger a targeted full account update only when the balance changes. This improves mined incoming transaction discovery without polling full transaction history for every account more often, and the existing five-minute full sync remains the fallback.


backend/backend.go
```diff
@@ -639,6 +639,28 @@ func (backend *Backend) ManualReconnect(reconnectETH bool) {
...
+func (backend *Backend) SetAccountActivity(accountCode accountsTypes.Code, active bool) error {
+	for _, account := range backend.Accounts() {
+		if account.Config().Config.Code != accountCode {
+			continue
+		}
+		ethAccount, ok := account.(*eth.Account)
+		if !ok {
+			return nil
+		}
+		if active && backend.environment != nil && backend.environment.UsingMobileData() {
```
Beerosagos: Is the updater fetching a lot of data? Otherwise, I'd skip the mobile data check. We possibly do the whole initial blockchain sync using mobile data, I don't see this as a big issue

frontends/web/src/routes/account/receive/receive.tsx
```diff
@@ -139,7 +140,9 @@ export const Receive = ({
   const [currentAddressIndex, setCurrentAddressIndex] = useState<number>(0);
 
   const account = accounts.find(({ code: accountCode }) => accountCode === code);
+  const isEthereumAccount = account !== undefined && isEthereumBased(account.coinCode);
   const insured = account?.bitsuranceStatus === 'active';
+  useAccountActivity(code, isEthereumAccount);
```
Beerosagos: Could it make sense to move the eth check inside the hook to avoid duplicating it around? Also, a comment about why it's only needed for eth accounts could help

backend/backend.go
```diff
@@ -639,6 +639,28 @@ func (backend *Backend) ManualReconnect(reconnectETH bool) {
...
 
+// SetAccountActivity marks an account as foreground-active for short-lived background refreshes.
+func (backend *Backend) SetAccountActivity(accountCode accountsTypes.Code, active bool) error {
+	for _, account := range backend.Accounts() {
+		if account.Config().Config.Code != accountCode {
+			continue
+		}
+		ethAccount, ok := account.(*eth.Account)
+		if !ok {
+			return nil
```
Beerosagos, line 693: Should not happen, but maybe a log to trace that this has been called for a non-eth account?

backend/coins/eth/updater.go
```diff
@@ -92,13 +105,85 @@ func (u *Updater) EnqueueUpdateForAllAccounts() {
...
+		etherScanClient := etherscan.NewEtherScan(chainID, u.etherscanClient, u.etherscanRateLimiter)
+		u.ProbeActiveAccountBalances(ethAccounts, etherScanClient)
+	}
+}
+
 // PollBalances updates the balances of all ETH accounts.
 // It does that in three different cases:
 // - When a timer triggers the update.
 // - When the signanl to update all accounts is sent through UpdateETHAccountsCh.
 // - When a specific account is updated through EnqueueUpdateForAccount.
```
bznein: We should probably update the comment here a bit to reflect what the method does now

frontends/web/src/api/account.ts
```diff
@@ -113,6 +113,13 @@ export const getEthAccountCodeAndNameByAddress = (address: string): Promise<TEth
   return apiPost('accounts/eth-account-code', { address });
 };
 
+export const setAccountActivity = (
```
bznein: I'll let someone from frontend confirm, but I think we tend to call api methods starting with either "get" or "post"

bznein (commented):
Just a couple of fly-by comments, I am reviewing a bit the eth updater in particular and will take some more time to do so :)

frontends/web/src/api/account.ts
thisconnect: Correct we usually prefix get request with `getApiendpointname` and often use the verb of the action for the post request. 

Reading what `account/${code}/activity` does...

> Add a short-lived foreground activity lease so the backend can identify ETH
accounts the user is actively viewing. While such a lease is active, cheaply
probe the remote balance and trigger a targeted full account update only when
the balance changes. This improves mined incoming transaction discovery without
polling full transaction history for every account more often, and the existing
five-minute full sync remains the fallback.

maybe `activateAccout` but this name is a bit confusing. `focusAccount`, `postAccountActivity` ... hm naming is hard :)

if it is ETH specific maybe move to `api/eth.ts` OR add ETH in the function name?

Sorry I dont have a good name, just some pointers 🙏

Tomasvrba: renamed to `postEthAccountActivity`

Tomasvrba:
Thanks for the review, comments addressed in the fixup commit.

One thing I did to make the ownership of `isEthereumBased` cleaner from @Beerosagos comment

> Could it make sense to move the eth check inside the hook to avoid duplicating it around? Also, a comment about why it's only needed for eth accounts could help

was to refactor the coin helpers into a shared coin utility module. See top commit (@thisconnect @bznein) if this is acceptable or you'd like me to revert it and do that differently

Beerosagos:
Didn't look into the updater logic (I'll leave it to @bznein), but everything else LGTM :+1:

thisconnect (approved):
frontend untested LGTM

bznein (commented):
The overall updater logic LGTM but I am a bit worried about the added calls to Etherscan. @Tomasvrba how many more calls would this make on average? @benma how close are we to hitting our limits?

Tomasvrba:
> The overall updater logic LGTM but I am a bit worried about the added calls to Etherscan. @Tomasvrba how many more calls would this make on average? @benma how close are we to hitting our limits?

It changes the logic for the active account(s) that the user is currently viewing (there's an expiry for when the user switches to a different account, so more than one can be active at a time, but not for more than 5 minutes), so that for the active account the etherscan balance (balance only) gets probed every minute instead of every 5 minutes

So for the active account, the balance would get fetched ~5x as often

benma:
Haven't forgotten this, just postponed it a few times because it seems difficult to review :innocent: But @Beerosagos and @bznein are in here, should I still check it?

PR #4169, opened 2026-05-27
tests: remove Android e2e tests.

bznein:
These tests were never really useful as we only implemented extremely basic ones, and they put burden on maintenaince that is not worth the effort.

- [ ] updating the Changelog
- [ ] writing unit tests


thisconnect (approved):
thank you. untested LGTM

PR #4175, opened 2026-06-01
frontend/backend: use macos number format system settings

thisconnect:
In case the user configured a custom 'Number format' in macOS system settings, prefer this over formatting by locale.

- [x] updating the Changelog
- [x] writing unit tests

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -164,6 +164,12 @@ type authEventObject struct {
 	Result *AuthResultType `json:"result,omitempty"`
 }
 
+// NumberFormat can contain different decimal and group separators specified by the user to format numbers can be empty in case the user did not overwrite.
+type NumberFormat struct {
+	DecimalSeparator string `json:"decimal"`
+	GroupSeparator   string `json:"group"`
+}
+
 // Environment represents functionality where the implementation depends on the environment the app
 // runs in, e.g. Qt5/Mobile/webdev.
 type Environment interface {
@@ -186,6 +192,7 @@ type Environment interface {
 	// to return empty string or unsupported value, in which case the app will use
 	// English by default.
 	NativeLocale() string
+	NumberFormat() *NumberFormat
 	// Invoke a file picker dialog for the user to select a destination to store a file.
 	// `suggestedFilename` is the proposed/default destination.
 	// The function should return the empty string if the user aborted the process.
--- a/backend/mobileserver/mobileserver.go
+++ b/backend/mobileserver/mobileserver.go
@@ -185,9 +185,12 @@ func Serve(dataDir string, testnet bool, environment GoEnvironmentInterface, goA
 				}
 				return []usb.DeviceInfo{deviceInfo{i}}
 			},
-			SystemOpenFunc:           environment.SystemOpen,
-			UsingMobileDataFunc:      environment.UsingMobileData,
-			NativeLocaleFunc:         environment.NativeLocale,
+			SystemOpenFunc:      environment.SystemOpen,
+			UsingMobileDataFunc: environment.UsingMobileData,
+			NativeLocaleFunc:    environment.NativeLocale,
+			NumberFormatFunc: func() *backend.NumberFormat {
+				return nil
+			},
 			GetSaveFilenameFunc:      environment.GetSaveFilename,
 			SetDarkThemeFunc:         environment.SetDarkTheme,
 			DetectDarkThemeFunc:      environment.DetectDarkTheme,
--- a/cmd/servewallet/main.go
+++ b/cmd/servewallet/main.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -118,6 +119,49 @@ func (webdevEnvironment) NativeLocale() string {
 	// used in the frontend and breaks UI in unexpected ways.
 	// We are always UTF-8 anyway.
 	return strings.Split(v, ".")[0]
+}
+
+func normalizeAppleSeparator(s string) string {
+	// macOS defaults encodes narrow no-break space as \U202f
+	return strings.ReplaceAll(s, `\\U202f`, "\u202f")
+}
+
+func parseAppleICUNumberSymbols(out []byte) *backendPkg.NumberFormat {
+	result := &backendPkg.NumberFormat{}
+
+	re := regexp.MustCompile(`(\d+)\s*=\s*"([^"]*)"`)
+	for _, m := range re.FindAllStringSubmatch(string(out), -1) {
+		switch m[1] {
+		case "0":
+			result.DecimalSeparator = normalizeAppleSeparator(m[2])
+		case "1":
+			result.GroupSeparator = normalizeAppleSeparator(m[2])
+		}
+	}
+
+	if result.DecimalSeparator == "" &&
+		result.GroupSeparator == "" {
+		return nil
+	}
+
+	return result
+}
+
+// NumberFormatting in webdev
+func (webdevEnvironment) NumberFormat() *backendPkg.NumberFormat {
+	log := logging.Get().WithGroup("servewallet")
+	if runtime.GOOS != "darwin" {
+		return nil
+	}
+
+	out, err := exec.Command("defaults", "read", "-g", "AppleICUNumberSymbols").Output()
+
+	if err != nil {
+		log.Warnf("failed to read AppleICUNumberSymbols via defaults: %v", err)
+		return nil
+	}
+
+	return parseAppleICUNumberSymbols(out)
 }
 
 // GetSaveFilename implements backend.Environment.
--- a/frontends/qt/server/server.go
+++ b/frontends/qt/server/server.go
@@ -54,6 +54,7 @@ import (
 	"strings"
 	"unsafe"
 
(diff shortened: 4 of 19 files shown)


thisconnect:
tested in webdev and on macOS

thisconnect:
@coderabbitai review

frontends/web/src/contexts/localization-provider.tsx
```diff
@@ -9,9 +12,28 @@ type TProps = {
...
+  const {
+    decimal: decimalFromLocale,
+    group: groupFromLocale,
+  } = useLocalizedPunctuation(nativeLocale);
+
+  const systemNumberFormat = useDefault(useLoad(getNumberFormat), null);
+
+  const decimal = systemNumberFormat?.decimal ?? decimalFromLocale;
+
+  const group = normalizeSeparator(systemNumberFormat?.group) ?? groupFromLocale;
```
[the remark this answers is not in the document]
thisconnect: > As per coding guidelines Use undefined to represent "loading" state in data hooks. Never use a separate isLoading boolean. Render a skeleton or nothing while the value is undefined.

in components this is good advice, but here in the context we dont ahve loading spinner or skeleton. this formatting options are used later in the app which needs to fetch transactions amounts etc and will have therefore have loading spinner, but I dont think we need to wait for getting the os number formatting. Ofc this could in unexpected rare cases also take long, but shouldn't.

frontends/web/src/contexts/localization-provider.tsx
```diff
@@ -9,9 +12,28 @@ type TProps = {
...
 
+const normalizeSeparator = (separator?: string) => {
+  if (
+    separator === '\\\\U202f'
+  ) {
+    return '\u202f';
+  }
+
+  return separator;
+};
```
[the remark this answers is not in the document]
thisconnect: the other ai suggest to normaliuze in the go backend 🤔

frontends/web/src/contexts/localization-provider.tsx
```diff
@@ -9,9 +12,28 @@ type TProps = {
   children: ReactNode;
 };
 
+const normalizeSeparator = (separator?: string) => {
+  if (
+    separator === '\\\\U202f'
```
thisconnect: Actually this seems not necessary with a recommendation from gpt, I'll update the PR

```diff
@@ -363,8 +363,8 @@ int main(int argc, char *argv[])
     QString decimalSeparator(locale.decimalPoint());
     QString groupSeparator(locale.groupSeparator());
 
-    std::string decimalSeparatorStr = decimalSeparator.toStdString();
-    std::string groupSeparatorStr = groupSeparator.toStdString();
+    std::string decimalSeparatorStr = decimalSeparator.toUtf8().toStdString();
+    std::string groupSeparatorStr = groupSeparator.toUtf8().toStdString();
 ```

Comment (thisconnect) [reply frontends/web/src/contexts/localization-provider.tsx]: \u202f is a narrow white space when configuring this option in macOS:

Image

File: frontends/web/src/contexts/localization-provider.tsx (hunk 2 above)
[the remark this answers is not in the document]
Comment (thisconnect) [reply frontends/web/src/contexts/localization-provider.tsx]: thanks added, so it also works correctly in webdev with servewallet

File: backend/mobileserver/mobileserver.go [hunk 4]
```diff
@@ -69,6 +69,7 @@ type GoEnvironmentInterface interface {
 	SystemOpen(string) error
 	UsingMobileData() bool
 	NativeLocale() string
+	// NumberFormat() *backend.NumberFormat
```
bznein: Leftover comment?

backend/backend.go
```diff
@@ -164,6 +164,12 @@ type authEventObject struct {
 	Result *AuthResultType `json:"result,omitempty"`
 }
 
+// NumberFormat can contain different decimal and group separators specified by the user to format numbers can be empty in case the user did not overwritex
```
bznein:
```suggestion
// NumberFormat can contain different decimal and group separators specified by the user to format numbers can be empty in case the user did not overwrite.
```

cmd/servewallet/main.go
```diff
@@ -120,6 +121,47 @@ func (webdevEnvironment) NativeLocale() string {
 	return strings.Split(v, ".")[0]
 }
 
+func normalizeAppleSeparator(s string) string {
+	// macOS defaults encodes narrow no-break space as \U202f
+	return strings.ReplaceAll(s, `\\U202f`, "\u202f")
```
bznein, line 126: codex comment:

```
  - [P3] Decode single-backslash Apple Unicode escapes — /home/bznein-coding-agent/
    agent-work/bitbox-wallet-app/cmd/servewallet/main.go:126-126
    When macOS defaults read -g AppleICUNumberSymbols emits a narrow no-break-space
    separator, it uses \U202f with one backslash in the output. This replacement is
    looking for the two-backslash string from the Go literal, so
    parseAppleICUNumberSymbols returns the literal escape and webdev amounts render as
    1\U202f234 instead of using the real separator.
```

Honestly, wouldn't know how to test it, so up to you to decide whether this is a valid comment :)

cmd/servewallet/main.go
```diff
@@ -120,6 +121,47 @@ func (webdevEnvironment) NativeLocale() string {
...
+// NumberFormatting in webdev
+func (webdevEnvironment) NumberFormat() *backendPkg.NumberFormat {
+	if runtime.GOOS != "darwin" {
+		return nil
+	}
+
+	out, err := exec.Command("defaults", "read", "-g", "AppleICUNumberSymbols").Output()
+
+	if err != nil {
+		return nil
```
bznein, line 161: [nit] maybe log the error?

frontends/qt/server/server.go
```diff
@@ -153,6 +156,18 @@ func serve(
 	// from the stack.
 	nativeLocale := C.GoString(preferredLocale)
 
+	decimalSeparator := C.GoString(preferredDecimalSeparator)
+	groupSeparator := C.GoString(preferredGroupSeparator)
+
+	var numberFormat *backend.NumberFormat
+
+	if decimalSeparator != "" && groupSeparator != "" {
```
bznein, line 164: [q] is it never possible to have one set but not the other?

backend/mobileserver/mobileserver.go
thisconnect: If we want to have support for this on mobile (iOS / Android) this would need to be added somehow, but maybe not with *backend.NumberFormat.

I'll just remove it for now I guess.

frontends/qt/server/server.go
thisconnect: If only one of the 2 are defined ever happens I think we should ignore it. 

The frontend should either get decimalSeparator&groupSeparator or no info about this.

cmd/servewallet/main.go
thisconnect: In my test the frontend displayed something like 10\\U202f000.00 in case I set the macOS settings to use space+dot. 

Screenshot 2026-06-03 at 09 09 54

but if I only replace single slash the frontend shows this in webdev/servewallet:
Screenshot 2026-06-03 at 10 54 06

I am not 100% sure where that comes from but by replacing double slash it was fixed..

thisconnect: In an earlier version (here is the force push diff https://github.com/BitBoxSwiss/bitbox-wallet-app/compare/215500a0ff0e0e36427fa50d24a1bf952b2ee872..66783fedd7eca69032212d3cb110343fc92861e6) I replaced 4 backslashes to make it work 😅

cmd/servewallet/main.go
thisconnect: added as warning (or should it be info?)

In case the user has not configured any number format `exec.Command("defaults", "read", "-g", "AppleICUNumberSymbols").Output()` will not return anything.

so the endpoint will return null is expected. See type in api/nativelocale.ts
```ts
export type TNumberFormat = {
  decimal: string;
  group: string;
} | null;
```

cmd/servewallet/main.go
thisconnect: probably still bit a hack but this is supposed to only be running in webdevEnvironment

thisconnect:
@bznein PTAL

bznein (approved):
backend untested LGTM. I'm not that comfortable reviewing the frontend :) though what I see looks good to me too

thisconnect:
@shonsirsha please check the frontend 🙏

frontends/web/src/utils/localize.ts
```diff
@@ -39,13 +45,32 @@ export const localizePercentage = (
       style: 'percent',
     });
   } catch (error) {
+    // in case invalid locale / Intl formatting failures and fall back below
+    formatter = Intl.NumberFormat('de-CH', {
```
shonsirsha: I may have misunderstood but..

nit: the fallback docs above says `(amount * 100).toFixed(2)` for invalid locales, but here it falls back to a `de-CH` `Intl.NumberFormat`.

Can update the jsdoc?

shonsirsha (approved):
LGTM

frontends/web/src/utils/localize.ts
thisconnect: updated docstring, but also added another try/catch so the fallback would apply on very old browser engines. this could happen on old android's with very old webviews

shonsirsha (approved):
LGTM

PR #4176, opened 2026-06-02
frontend: fix orientation change on all-accounts

thisconnect:
The app currently redirects to settings/manage-accounts when a orientation change happens on all-accounts. Ideally the view does not change, but all-accounts is not available on large screens.

This fixes by changing the view to account-summary.

- [ ] updating the Changelog
- [ ] writing unit tests


shonsirsha (approved):
LGTM

PR #4179, opened 2026-06-03
swap: allow UTXOs selection for BTC swaps.

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -1054,6 +1054,29 @@ func (account *Account) SpendableOutputs() ([]*SpendableOutput, error) {
 	return account.makeSpendableOutputs(utxos), nil
 }
 
+// SelectedUTXOsAmount returns the total amount of the selected spendable outputs.
+func (account *Account) SelectedUTXOsAmount(selectedUTXOs map[wire.OutPoint]struct{}) (coin.Amount, error) {
+	amount := coin.NewAmountFromInt64(0)
+	if len(selectedUTXOs) == 0 {
+		return amount, nil
+	}
+	if !account.Synced() {
+		return amount, accounts.ErrSyncInProgress
+	}
+	outputs, err := account.transactions.SpendableOutputs()
+	if err != nil {
+		return amount, err
+	}
+	for outPoint := range selectedUTXOs {
+		output, ok := outputs[outPoint]
+		if !ok {
+			return amount, errp.New("selected UTXO not found")
+		}
+		amount = coin.SumAmounts(amount, coin.NewAmountFromInt64(output.TxOut.Value))
+	}
+	return amount, nil
+}
+
 // ReusedAddressesForOutputs returns the subset of the provided outputs whose addresses are reused
 // across all indexed wallet outputs.
 func (account *Account) ReusedAddressesForOutputs(
--- a/frontends/web/src/api/account.ts
+++ b/frontends/web/src/api/account.ts
@@ -437,6 +437,21 @@ export const getUTXOs = (code: AccountCode): Promise<TUTXO[]> => {
   return apiGet(`account/${code}/utxos`);
 };
 
+export type TUTXOsAmount = {
+  amount: TAmountWithConversions;
+  success: true;
+} | {
+  errorMessage?: string;
+  success: false;
+};
+
+export const getUTXOsAmount = (
+  code: AccountCode,
+  selectedUTXOs: string[],
+): Promise<TUTXOsAmount> => {
+  return apiPost(`account/${code}/utxos/amount`, { selectedUTXOs });
+};
+
 type TSecureOutput = {
   hasSecureOutput: boolean;
   optional: boolean;
--- a/frontends/web/src/routes/market/swap/swap.tsx
+++ b/frontends/web/src/routes/market/swap/swap.tsx
@@ -5,6 +5,7 @@ import { useNavigate } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import {
   getBalance,
+  getUTXOsAmount,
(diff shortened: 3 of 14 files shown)


bznein:
@coderabbit review

backend/coins/btc/account.go
```diff
@@ -1054,6 +1054,30 @@ func (account *Account) SpendableOutputs() ([]*SpendableOutput, error) {
...
+			continue
+		}
+		amount = coin.SumAmounts(amount, coin.NewAmountFromInt64(output.TxOut.Value))
+		matches++
+	}
+	if matches != len(selectedUTXOs) {
+		return amount, errp.New("selected UTXO not found")
+	}
+	return amount, nil
+}
```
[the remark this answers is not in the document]
bznein: fixed

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -161,6 +178,69 @@ export const Swap = ({
...
+      return;
+    }
+    if (useSelectedUTXOs) {
+      fetchSelectedUTXOsAmount(selectedUTXOs).then(amount => {
+        if (amount !== undefined && mounted.current) {
+          setSelectedUTXOsAmount(amount);
+        }
+      });
+    }
+  }, [fetchSelectedUTXOsAmount, mounted, useSelectedUTXOs]);
```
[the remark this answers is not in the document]
bznein: fixed

bznein:
@coderabbit review

bznein:
Note: AI code review points out the following:

>  - [P1] Subtract fees before spending selected UTXOs — /home/bznein-coding-agent/agent-    work/bitbox-wallet-app/frontends/web/src/routes/market/swap/swap.tsx:517-518
>     When useSelectedUTXOs is true, effectiveSellAmount is the raw sum of selected UTXOs,    and the same selected UTXOs are then passed into the tx proposal. For BTC/LTC swaps
>     this creates a normal non-send-all transaction that must pay sellAmount + fee from
>     only those inputs, so selecting coins worth exactly the displayed amount always
>     fails with insufficient funds after the swap is signed. The selected-coin amount
>     needs to reserve/estimate fees before requesting/signing the swap, or avoid treating    the gross UTXO total as the output amount.


That is something we can't really control as we do in send (where we subtract the fee from the send amount) for the same reason we don't have a "max" button. This creates a circular dependency as in order to calculate the fee wee need all the tx details, but for them we need the quote, for which we need the amount...

frontends/web/src/api/account.ts
```diff
@@ -437,6 +437,21 @@ export const getUTXOs = (code: AccountCode): Promise<TUTXO[]> => {
...
+} | {
+  errorMessage?: string;
+  success: false;
+};
+
+export const getUTXOsAmount = (
+  code: AccountCode,
+  selectedUTXOs: string[],
+): Promise<TUTXOsAmount> => {
+  return apiPost(`account/${code}/utxos/amount`, { selectedUTXOs });
```
thisconnect, line 452: nit: other api functions dont use get prefix for post calls, but I am not sure what naming would be better.

is this endpoint swap specific or would we use the same in the future for example for pocket sell?

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -161,6 +178,71 @@ export const Swap = ({
...
+      setSelectedUTXOsAmount('');
+      fetchSelectedUTXOsAmount(selectedUTXOs).then(amount => {
+        if (amount !== undefined && mounted.current) {
+          setSelectedUTXOsAmount(amount);
+        }
+      });
+    }
+  }, [fetchSelectedUTXOsAmount, mounted, useSelectedUTXOs]);
+
+  const handleUseSelectedUTXOsChange = useCallback(async (useSelectedUTXOs: boolean) => {
```
thisconnect, line 231: was slightly confused about the similar function name
handleSelectedUTXOsChange and handleUseSelectedUTXOsChange

But then reading your comment
> That is something we can't really control as we do in send (where we subtract the fee from the send amount) for the same reason we don't have a "max" button. This creates a circular dependency as in order to calculate the fee wee need all the tx details, but for them we need the quote, for which we need the amount...

How about just dropping the max checkbox for now?

frontends/web/src/api/account.ts
bznein: Good point about the naming, I will think of a better one

> is this endpoint swap specific or would we use the same in the future for example for pocket sell?

I think it could be easibly reusable outside swap

frontends/web/src/routes/market/swap/swap.tsx
bznein: > was slightly confused about the similar function name
handleSelectedUTXOsChange and handleUseSelectedUTXOsChange

Fair point, it is indeed confusing, will try to rename it

> How about just dropping the max checkbox for now?

The checkbox I added is not a "max" checkbox but a checkbox that says "use the selected UTXOs" as the amount. This is the same thing we do in the BTC send flow, where an user can select the UTXOs but they are not actually used until they press the button.

We could change it so that selecting UTXOs automatically make them "used" (which I think would be maybe nicer) but that would create a discrepancy with how send is managed

thisconnect: > This is the same thing we do in the BTC send flow, where an user can select the UTXOs but they are not actually used until they press the button.

In normal BTC/LTC send they are used for the selection. but the checkbox is to send max/all selected UTXO's then the amount field is set to the max.

bznein:
Closing as the proper implementation requires change on the swapkit side

PR #4182, opened 2026-06-08
swap: add periodic test for slip24 verification

bznein:
Add a test that runs periodically (twice a day) on github CI only (or when SWAPKIT_LIVE_TEST env var is set to 1) for all the sellAsset we support, and verifies that the provided slip24 payload can be verified.

Right now the live test fails if the proxied provider is offline because quote returns no routes. Provider availability handling will be updated separately.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/market/swapkit/methods.go
+++ b/backend/market/swapkit/methods.go
@@ -1,8 +1,6 @@
 package swapkit
 
-import (
-	"context"
-)
+import "context"
 
 // Quote performs a SwapKit V3 quote request.
 func (c *Client) Quote(ctx context.Context, req *QuoteRequest) (*QuoteResponse, error) {
--- a/.github/workflows/swapkit-slip24.yml
+++ b/.github/workflows/swapkit-slip24.yml
@@ -1,0 +1,27 @@
+# See reference docs at
+# https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax
+name: swapkit-slip24
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "17 9,21 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  live-test:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+    steps:
+      - name: Clone the repo
+        uses: actions/checkout@v6
+      - name: Install Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+      - name: Run live SwapKit SLIP-24 test
+        run: |
+          go version
+          SWAPKIT_LIVE_TEST=1 go test -mod=vendor ./backend/market/swapkit -run TestSwapKitLiveSlip24 -count=1 -v
--- a/backend/market/swapkit/live_test.go
+++ b/backend/market/swapkit/live_test.go
@@ -1,0 +1,281 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package swapkit
+
+import (
+	"context"
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"net/http"
+	"os"
+	"strings"
+	"testing"
+
+	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/paymentrequest"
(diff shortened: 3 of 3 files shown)


backend/market/swapkit/types.go
```diff
@@ -35,6 +35,13 @@ type SwapRequest struct {
 	DisableBuildTx      bool     `json:"disableBuildTx"`
 }
 
+// Provider represents a SwapKit provider returned by the /providers endpoint.
+// See https://docs.swapkit.dev/swapkit-api/v3-get-supported-providers
```
benma: Dead link

backend/market/swapkit/methods.go
```diff
@@ -2,8 +2,18 @@ package swapkit
 
 import (
 	"context"
+	"strings"
 )
 
+// Providers returns the providers supported by the SwapKit API.
+func (c *Client) Providers(ctx context.Context) ([]Provider, error) {
+	var resp []Provider
+	if err := c.get(ctx, strings.TrimSuffix(c.baseURL, "/v3")+"/providers", &resp); err != nil {
```
benma: Stripping hardcoded `v3` is going to break once the version changes. Maybe there is an url method to get the base instead, or simply define a new var?

backend/market/swapkit/client.go
```diff
@@ -70,3 +71,42 @@ func (c *Client) post(ctx context.Context, path string, body any, out any) error
 
 	return nil
 }
+
+func (c *Client) get(ctx context.Context, path string, out any) error {
```
benma: You could delete this and use util.APIGet instead (or call it from here)?

backend/market/swapkit/client.go
```diff
@@ -70,3 +71,42 @@ func (c *Client) post(ctx context.Context, path string, body any, out any) error
...
+	ctx, cancel := context.WithTimeout(ctx, 20*time.Second)
+	defer cancel()
+
+	url := path
+	if strings.HasPrefix(path, "/") {
+		url = c.baseURL + path
+	}
```
benma: `post` does baseURL+path in every case. Please make it consistent if you keep this function.

benma (changes requested):
Asked codex review to focus on reusing prod code instead of replicating code in the tests, to avoid adding production code that is only used by tests, and to scrutinize the workflow file. 
Result:

---
• Findings

  - Medium: the live test bypasses production request construction at backend/market/swapkit/live_test.go:146 and backend/market/swapkit/live_test.go:172. Production goes through NewQuoteFromCoinCode / NewSwap at
    backend/market/swapkit/helpers.go:150 and backend/market/swapkit/helpers.go:192. This means the scheduled test can pass even if coin-code mapping or swap request construction breaks.

  - Medium: slip24PaymentRequest() duplicates production conversion logic at backend/market/swapkit/live_test.go:206, while Slip24.ToRequest() already validates/converts the payload at backend/paymentrequest/
    convert.go:14. Reuse slip24.ToRequest() first so the test covers the same JSON-to-internal path production uses.

  - Low: Provider, Client.Providers, and Client.get appear to be production code added only for the live test: backend/market/swapkit/types.go:38, backend/market/swapkit/methods.go:8, backend/market/swapkit/
    client.go:75. I’d move the provider probe into _test.go unless the app is going to use it.

  Workflow
  The workflow is valid, but not quite best-practice hardened:

  - Add permissions: contents: read.
  - Add timeout-minutes, e.g. 15, so a live dependency hang does not use the default 360-minute job timeout.
  - Change cron away from minute 0, e.g. 17 9,21 * * *; GitHub documents start-of-hour schedule load as more likely to delay/drop runs.
  - Use the repo convention: go test -mod=vendor ....
  - If strict supply-chain hardening is desired, pin actions by full commit SHA; GitHub documents SHA pinning as the immutable option.

  Suggested Shape
  Yes, the live test can reuse production code without adding test-only prod code. Make test cases use coin codes, then call:

  quote, apiError := NewQuoteFromCoinCode(ctx, http.DefaultClient, sellCoin, buyCoin, testCase.sellAmount)
  swap, apiError := NewSwap(ctx, http.DefaultClient, testCase.sellCoinCode, testCase.buyCoinCode, testCase.sellAmount, route.RouteID, ...)

  Keep any coin stubs and /providers probing helpers in _test.go. Do not export production helpers just for this test; only move code into production if production itself also uses the moved helper.

benma (approved):

PR #4183, opened 2026-06-08
frontend: keep mobile nav for lightning wallet

bznein:
Show the bottom navigation when a Lightning wallet is remembered, even without a connected keystore or regular active accounts.

Treat Lightning routes as account-tab destinations so mobile users can navigate back to Settings after disconnecting the keystore.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/app.tsx
+++ b/frontends/web/src/app.tsx
@@ -15,6 +15,7 @@ import { getAccounts } from './api/account';
 import { syncAccountsList } from './api/accountsync';
 import { getDeviceList } from './api/devices';
 import { syncDeviceList } from './api/devicessync';
+import { getLightningAccount, subscribeLightningAccount } from './api/lightning';
 import { syncNewTxs } from './api/transactions';
 import { notifyUser } from './api/system';
 import { ConnectedApp } from './connected';
@@ -42,10 +43,12 @@ export const App = () => {
 
   const accounts = useDefault(useSync(getAccounts, syncAccountsList), []);
   const devices = useDefault(useSync(getDeviceList, syncDeviceList), {});
+  const lightningAccount = useSync(getLightningAccount, subscribeLightningAccount);
   const prevDevices = usePrevious(devices);
 
   const deviceIDs = Object.keys(devices);
   const firstDevice = deviceIDs[0];
+  const hasLightningAccount = lightningAccount !== undefined && lightningAccount !== null;
 
   useEffect(() => {
     return syncNewTxs((meta) => {
@@ -68,15 +71,17 @@ export const App = () => {
       return;
     }
     // if no accounts are registered on specified views route to /
-    if (accounts.length === 0 && (
+    const canNavigateWithLightningAccount = currentURL === '/settings/more' || currentURL === '/accounts/all';
+    const requiresRegularAccount =
       currentURL.startsWith('/account-summary')
       || currentURL.startsWith('/add-account')
       || currentURL.startsWith('/settings/manage-accounts')
       || currentURL.startsWith('/accounts/')
-      // Workaround on mobile where the bottom menu is not shown when there are no devices/accounts.
-      // If one is on "More" and the bottom menu disappears, one is stuck.
-      || currentURL === '/settings/more'
-    )) {
+      || currentURL === '/settings/more';
+    const shouldRedirectNoRegularAccount =
+      !canNavigateWithLightningAccount
+      || lightningAccount === null;
+    if (accounts.length === 0 && requiresRegularAccount && shouldRedirectNoRegularAccount) {
       navigate('/');
       return;
     }
@@ -127,7 +132,7 @@ export const App = () => {
       return;
     }
 
-  }, [accounts, deviceIDs, firstDevice, navigate]);
+  }, [accounts, deviceIDs, firstDevice, lightningAccount, navigate]);
 
   useEffect(() => {
     const oldDeviceIDList = Object.keys(prevDevices || {});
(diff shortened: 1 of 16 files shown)


bznein:
@coderabbit review

frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
```diff
@@ -14,11 +14,13 @@ import styles from './bottom-navigation.module.css';
 type Props = {
   activeAccounts: TAccount[];
   devices: TDevices;
+  hasLightningAccount: boolean;
 };
```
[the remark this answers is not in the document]
bznein: Done

frontends/web/src/routes/lightning/lightning.tsx
```diff
@@ -30,39 +30,81 @@ import styles from './lightning.module.css';
...
+    if (lightningAccount === null) {
+      return;
+    }
+
+    const subscriptions = [
+      subscribeListPayments(onStateChange),
+    ];
     return () => unsubscribe(subscriptions);
-  }, [onStateChange, btcUnit]);
+  }, [btcUnit, isLightningAccountLoaded, lightningAccount, onStateChange]);
```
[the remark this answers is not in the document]
bznein: Fixed

frontends/web/src/routes/lightning/lightning.tsx
```diff
@@ -30,39 +30,81 @@ import styles from './lightning.module.css';
...
 
-    const subscriptions = [subscribeListPayments(onStateChange)];
+    const subscriptions = [
+      subscribeListPayments(onStateChange),
+    ];
     return () => unsubscribe(subscriptions);
-  }, [onStateChange, btcUnit]);
+  }, [btcUnit, isLightningAccountLoaded, lightningAccount, onStateChange]);
+
+  useEffect(() => {
```
Beerosagos: I think that it would be better to emit an event from the backend when Lightning is ready, instead of polling in the frontend.

Beerosagos (changes requested):

frontends/web/src/routes/lightning/lightning.tsx
bznein: good idea, done!

frontends/web/src/hooks/lightning.ts
```diff
@@ -4,6 +4,6 @@ import { useContext } from 'react';
 import { LightningContext } from '../contexts/LightningContext';
 
 export const useLightning = () => {
```
Beerosagos, line 6: useLightning is basically a wrapper of the LightningContext now, right? Maybe we can remove it and keep only the context?

frontends/web/src/app.tsx
```diff
@@ -32,11 +33,20 @@ import { BottomNavigation } from './components/bottom-navigation/bottom-navigati
...
+    </Providers>
+  </ConnectedApp>
+);
+
+const AppContent = () => {
   usePlatformClass();
   const { t } = useTranslation();
   const navigate = useNavigate();
   const { pathname } = useLocation();
+  const { lightningAccount } = useLightning();
```
Beerosagos: I'd suggest skipping the hook (which requires the context) and going with `const lightningAccount = useSync(getLightningAccount, subscribeLightningAccount);` so that we don't need to split the App component.

Beerosagos (commented):
Backend LGTM 👍 

I left a couple of suggestions for the frontend, but I think it's better if @thisconnect or @shonsirsha have a look at that

frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
```diff
@@ -45,36 +54,34 @@ export const BottomNavigation = ({
...
-          <NewBadge
-            className={styles.marketplaceNudgeDot}
-            configKey="hasSeenMarketplaceNudge"
-            hideOnPathPrefix="/market/"
-            pathname={pathname}
-            type="dot"
-          />
-        </span>
-      </Link>
+      {!onlyHasLightningAccount && (
```
thisconnect, line 64: Please mention the that this changes + reason in the commit message.

I guess Marketplace is intentionally hidden if there is only a lightning acccount for the reason that none of the offers in marketplace currently work with lighting.

frontends/web/src/hooks/lightning.ts
thisconnect: In a different context I also wondered about why using a hook to get context, but (if I remember correctly) @shonsirsha mentioned it was kind of best practice now a days? cc @shonsirsha

frontends/web/src/routes/lightning/lightning.tsx
```diff
@@ -30,39 +30,63 @@ import styles from './lightning.module.css';
...
+      return;
+    }
+
     onStateChange();
 
-    const subscriptions = [subscribeListPayments(onStateChange)];
+    const subscriptions = [
+      subscribeListPayments(onStateChange),
+    ];
     return () => unsubscribe(subscriptions);
```
thisconnect: nit (was already like this beofre): all unsubscribe does is itereate over the array, call each function and remove all those functions from the array. So it is suitable if there is multiple subscriptions.

feel free to change to 

```suggestion
    const unsubscribeCallback = subscribeListPayments(onStateChange);
    return () => unsubscribeCallback();
```

or just

```tsx
    return subscribeListPayments(onStateChange);
```

frontends/web/src/app.tsx
```diff
@@ -32,11 +33,20 @@ import { BottomNavigation } from './components/bottom-navigation/bottom-navigati
...
 import styles from './app.module.css';
 
-export const App = () => {
+export const App = () => (
+  <ConnectedApp>
+    <Providers>
+      <AppContent />
+    </Providers>
+  </ConnectedApp>
+);
```
thisconnect: nit: this component feels  a bit useless to me, but no strong opinion, keep it if you find it more readable.

However if you keep it, I'd find it more logical if this comes after `const AppContent`

thisconnect (commented):
nice 👍 
added a few minor nits.

frontends/web/src/routes/lightning/lightning.tsx
```diff
@@ -30,39 +29,60 @@ import styles from './lightning.module.css';
...
+  }, [mounted]);
 
   useEffect(() => {
+    if (lightningAccount === undefined || isLightningReady === undefined) {
+      return;
+    }
+
+    if (lightningAccount === null || !isLightningReady) {
+      return;
+    }
```
thisconnect, line 76: nit: if it is not important if lightningAccount is falsy (null, undefined or empty string :) ) an alternative for both conditions maybe just, also later for the `<Spinner>`
```suggestion
    if (!lightningAccount || !isLightningReady) {
      return;
    }
```

frontends/web/src/app.tsx
```diff
@@ -68,15 +72,17 @@ export const App = () => {
...
       || currentURL.startsWith('/settings/manage-accounts')
       || currentURL.startsWith('/accounts/')
-      // Workaround on mobile where the bottom menu is not shown when there are no devices/accounts.
-      // If one is on "More" and the bottom menu disappears, one is stuck.
-      || currentURL === '/settings/more'
-    )) {
+      || currentURL === '/settings/more';
+    const shouldRedirectNoRegularAccount =
+      !canNavigateWithLightningAccount
+      || hasLoadedLightningAccount && !hasLightningAccount;
```
thisconnect: this condition is my least favorit part of the PR, maybe it could simpler? 😅 

hasLightningAccount already contains that `lightningAccount !== undefined`, so I guess this can be removed here.

thisconnect (approved):
untested LGTM with small nit

Beerosagos (approved):
backend LGTM 🙏

PR #4185, opened 2026-06-09
lightning: synchronize sdk service access

bznein:
Guard Breez SDK service reads and reconnects with a mutex, and route payment operations through the active SDK helper so disconnects cannot race direct service access.

- [ ] updating the Changelog
- [ ] writing unit tests


bznein:
@coderabbit review

PR #4186, opened 2026-06-09
backend/btc: sanity check txID

benma:
Same as in `TransactionBroadcast`, sanity check the server response.

### DIFF
--- a/backend/coins/btc/electrum/client.go
+++ b/backend/coins/btc/electrum/client.go
@@ -130,6 +130,9 @@ func (c *client) TransactionGet(txHash chainhash.Hash) (*wire.MsgTx, error) {
 	if err := tx.BtcDecode(bytes.NewReader(rawTx), 0, wire.WitnessEncoding); err != nil {
 		return nil, err
 	}
+	if txHash != tx.TxHash() {
+		return nil, errp.New("Response is unexpected (transaction hash mismatch)")
+	}
 	return tx, nil
 }
 
--- a/backend/coins/btc/transactions/verification.go
+++ b/backend/coins/btc/transactions/verification.go
@@ -31,6 +31,14 @@ func (transactions *Transactions) unverifiedTransactions() (map[chainhash.Hash]i
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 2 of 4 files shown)


backend/coins/btc/transactions/verification.go
```diff
@@ -48,6 +56,10 @@ func hashMerkleRoot(merkle []blockchain.TXHash, start chainhash.Hash, pos int) c
 	return start
 }
 
+func (transactions *Transactions) handleTxInclusionVerificationFailed(txHash chainhash.Hash, reason string) {
+	transactions.log.WithField("txHash", txHash.String()).Warning(reason)
+}
+
```
bznein, line 62: [nit] I feel like this kind of wrappers is very codex-like and not necessarily useful. Up to you whether to keep it or not, but imho we can live without it

backend/coins/btc/electrum/client.go
```diff
@@ -130,6 +130,9 @@ func (c *client) TransactionGet(txHash chainhash.Hash) (*wire.MsgTx, error) {
 	if err := tx.BtcDecode(bytes.NewReader(rawTx), 0, wire.WitnessEncoding); err != nil {
 		return nil, err
 	}
+	if txHash != tx.TxHash() {
+		return nil, errp.New("Response is unexpected (transaction hash mismatch)")
```
bznein, line 134: [nit] down in transactions.go the error also contains the expected and actual hash, maybe for consistency we can log it here as well?

bznein (approved):
utACK

With just a couple of nits

backend/coins/btc/transactions/verification.go
benma: I specifically asked it to make it, so if we ever want to handle the inclusion error and surface it, there is one location to update and there is no risk of forgetting the second location :smile:

backend/coins/btc/electrum/client.go
benma: I went for the opposite, txIDs should not be in log files: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4189

PR #4192, opened 2026-06-17
frontend: add optional onion block explorer URLs

smokyisthatyou:
Add an expert setting to use mempool's .onion address for BTC/TBTC explorer links, whitelist the URLs in SystemOpen, and route existing explorer link call sites through a shared BlockExplorerLink component.

- [x] updating the Changelog
- [x] writing unit tests

### DIFF
--- a/frontends/web/src/api/config.ts
+++ b/frontends/web/src/api/config.ts
@@ -57,6 +57,7 @@ export type TConfigFrontend = Readonly<{
   darkmode?: boolean;
   expertFee?: boolean;
   coinControl?: boolean;
+  useOnionExplorerUrls?: boolean;
   selectedExchangeRegion?: string;
   hideEnableRememberWalletDialog?: boolean;
   hasUsedWalletConnect?: boolean;
--- a/frontends/web/src/components/block-explorer-link/block-explorer-link.tsx
+++ b/frontends/web/src/components/block-explorer-link/block-explorer-link.tsx
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { ReactNode } from 'react';
+import { useTranslation } from 'react-i18next';
+import { A } from '@/components/anchor/anchor';
+import { useConfig } from '@/contexts/ConfigProvider';
+import { getMempoolExplorerUrl } from '@/utils/explorer-url';
+
+type TProps = {
+  children: ReactNode;
+  className?: string;
+  explorerId: string;
+  prefix: string;
+  title?: string;
+};
+
+export const BlockExplorerLink = ({
+  children,
+  className,
+  explorerId,
+  prefix,
+  title,
+}: TProps) => {
+  const { t } = useTranslation();
+  const { config } = useConfig();
+  const useOnion = config?.frontend.useOnionExplorerUrls ?? false;
+  const href = getMempoolExplorerUrl(prefix, explorerId, useOnion);
+  const linkTitle = title ?? `${t('transaction.explorerTitle')}\n${href}`;
+
+  return (
+    <A className={className} href={href} title={linkTitle}>
+      {children}
+    </A>
+  );
+};
--- a/frontends/web/src/utils/explorer-url.ts
+++ b/frontends/web/src/utils/explorer-url.ts
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+
+const MEMPOOL_ONION_HOST = 'mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad.onion';
+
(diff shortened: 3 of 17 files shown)


frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/Util.java
```diff
@@ -47,6 +47,20 @@ public static void systemOpenExternal(Application application, String url) throw
         }
     }
 
+    public static void systemOpenExplorerChooser(Application application, String onionUrl, String clearnetUrl) throws Exception {
```
benma: Why not always give the choice when Tor proxy is enabled?

Does Android make the URL visible in the chooser? Can one pick the browser too app too? Screenshots would be nice :)

frontends/web/src/api/config.ts
```diff
@@ -57,6 +57,7 @@ export type TConfigFrontend = Readonly<{
   darkmode?: boolean;
   expertFee?: boolean;
   coinControl?: boolean;
+  useOnionExplorerUrls?: boolean;
```
benma, line 60: If it is a setting, why is there a chooser? If toggled on, shouldn't it do the onion url?

smokyisthatyou: yes in theory, and i was considering it actually. but my thinking was that on mobile, i suppose not many users to have tor browser as default, so having to chose on mobile seemed reasonable.

i am not 100% sure yet though, that's why i left the pr on draft 😁
wdyt? would you think that opening the onion url right away also on mobile could be better?

frontends/android/BitBoxApp/app/src/main/java/ch/shiftcrypto/bitboxapp/Util.java
smokyisthatyou: onion explorer URLs are separate from tor proxy because the proxy only affects in-app traffic, not the external browser. i supposed that maybe some users want to open the mempool onion address even if they do not have the tor proxy enabled. 

posting android screenshots asap :)

smokyisthatyou: after some testing, I can say the Android chooser does not appear reliably: once a default browser is set, it is not shown, even when multiple browsers are installed (including Tor).
i still want to try on a physical device, since Tor Browser keeps crashing on the emulator, and that may be affecting the behavior.

if the chooser does not appear as soon as a default browser is selected though, I don’t think it makes sense to keep this option. we should probably open the onion URL directly when the setting is enabled instead, as per other operating systems :)

frontends/web/src/utils/explorer-url.ts
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+
+const MEMPOOL_ONION_HOST = 'mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad.onion';
```
thisconnect, line 3: In current master prefixes are passed from the backend to the frontend so they dont need to be maintained in the frontend.

But with this approach, we'd need to update both backend and frontend in case something changes...

also if we ever had another frontend these would need to be dupicated and maintained there as well.

What would it take if the backend is aware of this new config and already passes the correct prefix (clear or onion depending on the setting?

I slightly prefer this comes from the backend, but cc @benma

frontends/web/src/components/block-explorer-link/block-explorer-link.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { ReactNode } from 'react';
+import { useTranslation } from 'react-i18next';
+import { A } from '@/components/anchor/anchor';
+import { useConfig } from '@/contexts/ConfigProvider';
+import { getMempoolExplorerUrl } from '@/utils/explorer-url';
+
+type TProps = {
+  children: ReactNode;
+  className?: string;
+  id: string;
```
thisconnect: naming-nit: seeing an `id` prop on a component I would expect it to be HTML id attribute, but here it is used to pass the address-id or tx-id (consider renaming to addressOrTxId or something similar?)

thisconnect (changes requested):

frontends/web/src/utils/explorer-url.ts
thisconnect: >  they dont need to be maintained in the frontend.

By "they" I meant the prefixes `https://mempool.space/tx/` etc

frontends/web/src/components/block-explorer-link/block-explorer-link.tsx
smokyisthatyou: makes total sense, will update

PR #4194, opened 2026-06-17
transactions: show fee rate in details

bznein:
Expose a backend-formatted BTC/LTC fee rate string on transaction details. Show it in the advanced transaction details section and cover the behavior in the existing send Playwright test.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/coins/btc/feetarget.go
+++ b/backend/coins/btc/feetarget.go
@@ -29,10 +29,15 @@ func (feeTarget *FeeTarget) Code() accounts.FeeTargetCode {
 
 // FormattedFeeRate returns a string showing the fee rate.
 func (feeTarget *FeeTarget) FormattedFeeRate() string {
-	if feeTarget.feeRatePerKb == nil {
+	return FormatFeeRate(feeTarget.feeRatePerKb)
+}
+
+// FormatFeeRate returns a string showing the fee rate.
+func FormatFeeRate(feeRatePerKb *btcutil.Amount) string {
+	if feeRatePerKb == nil {
 		return ""
 	}
-	feePerByte := fmt.Sprintf("%.03f", float64(*feeTarget.feeRatePerKb)/1000.0)
+	feePerByte := fmt.Sprintf("%.03f", float64(*feeRatePerKb)/1000.0)
 	// Truncate trailing zeroes, and final '.' if the number has no decimal places.
 	feePerByte = strings.TrimRight(strings.TrimRight(feePerByte, "0"), ".")
 	return feePerByte + " sat/vB"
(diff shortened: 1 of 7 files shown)


bznein:
@coderabbit review

backend/coins/btc/handlers/handlers.go
```diff
@@ -173,6 +174,7 @@ func (handlers *Handlers) getTxInfoJSON(txInfo *accounts.TransactionData, detail
 			txInfoJSON.Size = txInfo.Size
 			txInfoJSON.Weight = txInfo.Weight
 			feeRatePerKb := txInfo.FeeRatePerKb
+			txInfoJSON.FeeRateInfo = btc.FormatFeeRatePerVByte(feeRatePerKb)
 			if feeRatePerKb != nil {
 				txInfoJSON.FeeRatePerKb = coin.ConvertBTCAmount(handlers.account.Coin(), *feeRatePerKb, true, accountConfig.RateUpdater)
```
Beerosagos: It seems that FeeRatePerKb is not actually used anywhere. Could we get rid of it?

backend/coins/btc/feetarget.go
```diff
@@ -29,13 +29,22 @@ func (feeTarget *FeeTarget) Code() accounts.FeeTargetCode {
 
 // FormattedFeeRate returns a string showing the fee rate.
 func (feeTarget *FeeTarget) FormattedFeeRate() string {
-	if feeTarget.feeRatePerKb == nil {
+	return formatFeeRate(feeTarget.feeRatePerKb, "sat/vB")
+}
+
+// FormatFeeRatePerVByte returns a string showing the transaction fee rate.
+func FormatFeeRatePerVByte(feeRatePerKb *btcutil.Amount) string {
```
Beerosagos: Looks to me that this is mostly a duplication of feeTarget.FormattedFeeRate, no? It just formats 'sat/vByte' instead of 'sat/vB'.

backend/coins/btc/handlers/handlers.go
benma: This makes me a bit sad lol.

We used to show the fee rate in the past, I think. Trying to figure out why/how it got dropped.

  The effective removal happened in:

  846c84990c392bcc3ffb5f1a5dbc1b0ebe85b07a
  2024-06-21 frontend: create TxDetail component

  That commit changed the fee row to pass title into <TxDetail>, but TxDetail only rendered label and
  children, so the title never reached the DOM. So it got dropped in a refactoring bug 😢 

@thisconnect noticed this https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/2776#discussion_r1677533442 but somehow there it shows that the fee rate was not populated by the backend? Seems like @thisconnect assumed it never was or sth and him and PR author decided to remove it completely.

Anyway, good it's coming back 😄

shonsirsha:
frontend looks fine (as-is) :)

Beerosagos (approved):
utACK

PR #4196, opened 2026-06-20
backend: add user agent and version to update checks

jstrnbrg:


backend/update.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/BitBoxSwiss/bitbox02-api-go/util/semver"
 )
 
-const updateFileURL = "https://bitboxapp.shiftcrypto.io/desktop.json"
+const updateFileURL = "https://bitboxapp-analytics.shiftcrypto.io/desktop.json"
```
benma: this sounds very wrong? fetching update banner is not an analytics feature

benma (changes requested):

backend/bridgecommon/bridgecommon.go
```diff
@@ -188,6 +188,7 @@ type BackendEnvironment struct {
 	AuthFunc                 func()
 	OnAuthSettingChangedFunc func(bool)
 	BluetoothConnectFunc     func(string)
+	UserAgentHostFunc        func() string
```
benma: Renamed Host to Platform, seems like a better name. Please amend https://github.com/jstrnbrg/bitbox-wallet-app/pull/1 into your commit. (It also provides a fallback for non-ipad/iphone ios platforms)

benma (approved):
utACK

PR #4197, opened 2026-06-22
frontend: use mobile header for back navigation on mobile

shonsirsha:
**Chevron Back On Mobile**
- `Send`
- `Receive`
- `AddAccount`
- `ConnectScreenWalletConnect` / `WCConnectForm`
- `ElectrumSettings` (Connect full node)
- `Swap`
- `ManageBackups` / `BackupsList`
- `SDCardCheck` covered by surrounding `ManageBackups` header
- `RecoveryWords`
- `Bip85`
- `Passphrase` / `EnableInfo` / `DisableInfo`
- `SetupOptions` advanced screen
- `SetDeviceName`
- `SetDeviceNameWithSDCard` via `SetDeviceName`
- `RestoreFromSDCardBackup`

**titleOnly Usage**
- `MobileHeader` gets the new `variant="titleOnly"` support.
- `ConnectScreenWalletConnect` uses it while WalletConnect is loading:
  - `variant={loading ? 'titleOnly' : 'back'}`. This shows the mobile title without a chevron during connect/pairing loading.

### DIFF
--- a/frontends/web/src/routes/account/add/add-account.tsx
+++ b/frontends/web/src/routes/account/add/add-account.tsx
@@ -16,7 +16,8 @@ import { CoinDropDown } from '@/components/dropdown/coin-dropdown';
 import { SubTitle } from '@/components/title';
 import { useMediaQuery } from '@/hooks/mediaquery';
 import { UseBackButton } from '@/hooks/backbutton';
-import { BackButton } from '@/components/backbutton/backbutton';
+import { DesktopBackButton } from '@/components/backbutton/backbutton';
+import { MobileHeader } from '@/routes/settings/components/mobile-header';
 import { AddAccountGuide } from './add-account-guide';
 import { Skeleton } from '@/components/skeleton/skeleton';
 import styles from './add-account.module.css';
@@ -241,7 +242,13 @@ export const AddAccount = ({ accounts }: TAddAccountProps) => {
     <Main>
       <GuideWrapper>
         <GuidedContent>
-          <Header title={<h2>{t('manageAccounts.title')}</h2>} />
+          <Header
+            title={
+              <>
+                <h2 className="hide-on-small">{t('manageAccounts.title')}</h2>
+                <MobileHeader onClick={back} withGuide title={t('manageAccounts.title')} />
+              </>
+            } />
           <View
             fitContent
             textCenter
@@ -309,11 +316,11 @@ export const AddAccount = ({ accounts }: TAddAccountProps) => {
                     {t('addAccount.success.addAnotherAccount')}
                   </Button>
                 ) : (
-                  <BackButton
+                  <DesktopBackButton
                     onClick={back}
                   >
                     {t('button.back')}
-                  </BackButton>
+                  </DesktopBackButton>
                 )}
               </ViewButtons>
             </form>
--- a/frontends/web/src/routes/account/walletconnect/connect.tsx
+++ b/frontends/web/src/routes/account/walletconnect/connect.tsx
@@ -11,6 +11,7 @@ import { TConnectStatus } from './types';
 import { GuideWrapper, GuidedContent, Header, Main } from '@/components/layout';
 import { alertUser } from '@/components/alert/Alert';
 import { View, ViewContent, ViewHeader } from '@/components/view/view';
+import { MobileHeader } from '@/routes/settings/components/mobile-header';
 import { WCHeader } from './components/header/header';
 import { WCConnectForm } from './components/connect-form/connect-form';
 import { WCIncomingPairing } from './components/incoming-pairing/incoming-pairing';
@@ -92,7 +93,15 @@ export const ConnectScreenWalletConnect = ({
     <GuideWrapper>
       <GuidedContent>
         <Main>
-          <Header />
+          <Header
+            title={
+              status === 'connect' ? (
+                <MobileHeader
+                  variant={loading ? 'titleOnly' : 'back'}
+                  withGuide
+                  title={t('walletConnect.walletConnect')} />
+              ) : undefined
+            } />
           <View
             fitContent
             verticallyCentered
--- a/frontends/web/src/routes/device/bitbox02/bip85.tsx
+++ b/frontends/web/src/routes/device/bitbox02/bip85.tsx
@@ -11,8 +11,10 @@ import { SimpleMarkup } from '@/utils/markup';
 import { A } from '@/components/anchor/anchor';
 import { Column, ResponsiveGrid } from '@/components/layout';
 import { useDarkmode } from '@/hooks/darkmode';
+import { useMediaQuery } from '@/hooks/mediaquery';
 import { UseDisableBackButton } from '@/hooks/backbutton';
-import { BackButton } from '@/components/backbutton/backbutton';
+import { DesktopBackButton } from '@/components/backbutton/backbutton';
+import { MobileHeader } from '@/routes/settings/components/mobile-header';
 import bip85Graphic from './assets/bip85-graphic.svg';
 import bip85GraphicLight from './assets/bip85-graphic-light.svg';
 
@@ -28,8 +30,10 @@ export const Bip85 = ({
   const navigate = useNavigate();
   const { t } = useTranslation();
   const { isDarkMode } = useDarkmode();
+  const isMobile = useMediaQuery('(max-width: 768px)');
   const [status, setStatus] = useState<Status>('info-what');
   const [disclaimer, setDisclaimer] = useState(false);
+  const handleClose = () => navigate(-1);
 
   switch (status) {
   case 'info-what':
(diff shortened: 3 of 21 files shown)


frontends/web/src/routes/account/walletconnect/connect.tsx
```diff
@@ -92,7 +93,15 @@ export const ConnectScreenWalletConnect = ({
     <GuideWrapper>
       <GuidedContent>
         <Main>
-          <Header />
+          <Header
+            title={
+              status === 'connect' ? (
+                <MobileHeader
+                  variant={loading ? 'titleOnly' : 'back'}
```
thisconnect, line 100: The "loading" state confused me quickly, it is not about loading the UI but gets set to true when doing a new connection while pairing. 

tested on Android 👍

frontends/web/src/routes/device/bitbox02/setup/name.tsx
```diff
@@ -41,6 +42,7 @@ export const SetDeviceName = ({
         withBottomBar
         verticallyCentered
         width="600px">
+        <MobileHeader onClick={onBack} title={t('bitbox02Wizard.stepCreate.title')} />
```
thisconnect: on mobile the Chevron icon is not aligned with the content and there is missing space so it is just aligned with the edge of the screen.

frontends/web/src/routes/device/bitbox02/setup/restore.tsx
```diff
@@ -27,6 +28,7 @@ export const RestoreFromSDCardBackup = ({
       verticallyCentered
       withBottomBar
       width="700px">
+      <MobileHeader onClick={onBack} title={t('backup.restore.confirmTitle')} />
```
thisconnect: On mobile there is now 2 titles "Restore backup", also the chevron is not aligned with the content

frontends/web/src/routes/device/bitbox02/bip85.tsx
```diff
@@ -38,6 +39,7 @@ export const Bip85 = ({
         key="bip85-info-what"
         fullscreen
         verticallyCentered>
+        <MobileHeader title={t('deviceSettings.expert.bip85.title')} />
```
thisconnect: on mobile the title breaks a bit weirdly and becomes left aligned. icon is not aligned with content.

frontends/web/src/routes/device/bitbox02/bip85.tsx
```diff
@@ -120,10 +128,10 @@ export const Bip85 = ({
             }}>
             {t('button.continue')}
           </Button>
-          <BackButton
+          <DesktopBackButton
             onClick={() => setStatus('info-how')}>
             {t('button.back')}
-          </BackButton>
+          </DesktopBackButton>
```
thisconnect, line 147: small nit, I believe before it was possible to back & exit the workflow.

now back is going 1 step back in the workflow.

I think we could use the (x) to exit workflow or "<" back to do 1 step back, but I forgot when to use which. maybe it would be good to quickly check with @jadzeidan when to use what.

frontends/web/src/routes/device/bitbox02/passphrase.tsx
```diff
@@ -187,9 +190,12 @@ const EnableInfo = ({ handleAbort, setPassphrase }: TInfoProps) => {
       key={`step-${infoStep}`}
       fullscreen
       minHeight={CONTENT_MIN_HEIGHT}
-      onClose={handleAbort}
+      onClose={isMobile ? undefined : handleAbort}
       verticallyCentered
     >
+      <MobileHeader
+        onClick={handleBack}
+        title={t('deviceSettings.expert.passphrase.title')} />
```
thisconnect: space between icon and screen missing here and probably also in DisableInfo

frontends/web/src/routes/device/bitbox02/recovery-words.tsx
```diff
@@ -69,10 +70,7 @@ export const RecoveryWords = ({ deviceID }: TProps) => {
       fullscreen
       minHeight={CONTENT_MIN_HEIGHT}
       verticallyCentered>
-      <UseBackButton handler={() => {
-        handleAbort();
-        return false;
-      }} />
+      <MobileHeader onClick={handleAbort} title={t('backup.showMnemonic.title')} />
```
thisconnect: space between icon screen missing

frontends/web/src/routes/account/add/add-account.tsx
```diff
@@ -241,7 +242,13 @@ export const AddAccount = ({ accounts }: TAddAccountProps) => {
...
       <GuideWrapper>
         <GuidedContent>
-          <Header title={<h2>{t('manageAccounts.title')}</h2>} />
+          <Header
+            title={
+              <>
+                <h2 className="hide-on-small">{t('manageAccounts.title')}</h2>
+                <MobileHeader onClick={back} withGuide title={t('addAccount.title')} />
+              </>
+            } />
```
thisconnect, line 251: Maybe not now, but this screams to me that we should merge Header + MobileHeader into one component.

frontends/web/src/routes/settings/components/mobile-header.tsx
```diff
@@ -2,37 +2,47 @@
...
-export const MobileHeader = ({ title, withGuide = false, onClick }: TProps) => {
+export const MobileHeader = ({ title, variant = 'back', withGuide = false, onClick }: TProps) => {
   const navigate = useNavigate();
+  const { goBack } = useBackNavigation();
   const handleClick = () => {
     // goes to the previous page if no onClick function is provided
     if (!onClick) {
-      navigate(-1);
+      if (!goBack()) {
+        navigate(-1);
```
thisconnect, line 30: `goBack()` already tries to do `navigate()` so this looks a bit wrong to me but I didn't fully look into it.

One bug I found (but didn't check if already on master or if it is introduced with this PR) is:

Swiping back eventually asks to close the App on Android.

But in "More" sometimes it doesn't ask.

1) goto More
2) swipe back
3) "Do you really want to exit?" (expected)

But this doesn't work:

1) goto More
2) goto Settings
3) swipe back goes back to More (expected)
4) swipe back nothing happens (expected is "Do you really want to exit?")

thisconnect (changes requested):

frontends/web/src/routes/device/bitbox02/bip85.tsx
shonsirsha: > small nit, I believe before it was possible to back & exit the workflow.

It was possible only on desktop :) on mobile, user can only go back one step. In this PR we'll preserve this behaviour.

frontends/web/src/routes/settings/components/mobile-header.tsx
shonsirsha: > goBack() already tries to do navigate() so this looks a bit wrong to me but I didn't fully look into it.

Mmh.. well, `goBack()` only navigates if the internal navigation stack (`BackNavigationContext`) has a previous entry. If there is no previous entry, it returns `false` without navigating.

The `navigate(-1)` is a "fallback" -  only runs when `goBack()` did nothing and it keeps the old `MobileHeader` behavior.

> One bug I found (but did...

I think this bug is already on master

shonsirsha:
@thisconnect , thanks a lot for the reviews! PTAL 🙏

frontends/web/src/routes/settings/components/mobile-header.tsx
thisconnect: > I think this bug is already on master

confirmed. It's already a bug in 4.51

frontends/web/src/routes/device/bitbox02/bip85.tsx
```diff
@@ -28,16 +30,27 @@ export const Bip85 = ({
...
 
   switch (status) {
   case 'info-what':
     return (
       <View
         key="bip85-info-what"
         fullscreen
+        onClose={isMobile ? undefined : handleClose}
         verticallyCentered>
+        {mobileHeader()}
```
thisconnect: nit:

```suggestion
    <MobileHeader
      title={t('deviceSettings.expert.bip85.title')}
      withViewPadding
    />
```

frontends/web/src/routes/device/bitbox02/bip85.tsx
```diff
@@ -79,7 +92,9 @@ export const Bip85 = ({
       <View
         key="bip85-info-how"
         fullscreen
+        onClose={isMobile ? undefined : handleClose}
         verticallyCentered>
+        {mobileHeader(() => setStatus('info-what'))}
```
thisconnect: nit: simpler and less mental overhead here and all the other places below
```suggestion
    <MobileHeader
      onClick={() => setStatus('info-what')}
      title={t('deviceSettings.expert.bip85.title')}
      withViewPadding
    />
```

thisconnect (approved):
tested Qt + Android LGTM with one small last nit

PR #4200, opened 2026-06-23
frontend: use isdarkmode from darkmode hook

thisconnect:
As per comment getDarkmode function was a workaround for class based components. Functional components should use the hook directly.

- [ ] updating the Changelog
- [ ] writing unit tests


strmci (approved):
LGTM

PR #4201, opened 2026-06-24
frontend: fix strip trailing zeros also in cased of tbtc rbtc ltc

thisconnect:
The removeTrailingZeros prop only had an effect if the unit was 'BTC' but not in case of tbtc, ltc etc.

'sat' is also bitcoin based but is handled correclty by internal amount component, zeros should not be removed for sats.


shonsirsha (approved):
LGTM. Thanks!!

PR #4204, opened 2026-06-29
Release v4.51.2

benma:
This is a bugfix / patch release, based on v4.51.1.

Review notes:

- The first two commits and the last commitare freshly authored - they include new firmwares and an upgrade UX improvement (avoiding of blank screen while the upgrade is being prepared) and the version bump
- All the other commits are **selected** commits cherry picked from `v4.51.1..7be5e9f39` (changes since last release until `master` at time of writing)

Here is a helpful command to verify the cherry picks. Each cherry pick commit contains the commit hash of the original commit in the commit message. Use the following command to compare the patch ID of each such commit and the original - if it is the same, it applied cleanly and the patch is the same. If there is a difference, it will output which commits differ:

```sh
for c in $(git log --format=%H "$(git merge-base HEAD v4.51.1)"..HEAD); do src=$(git show -s --format=%B "$c" | sed -n "s/.*cherry picked from commit \([0-9a-f]*\)).*/\1/p"); [ -z "$src" ] && continue; src_pid=$(git show "$src" --pretty=format:%H --no-ext-diff --no-textconv | git patch-id --stable | cut -d" " -f1); c_pid=$(git show "$c" --pretty=format:%H --no-ext-diff --no-textconv | git patch-id --stable | cut -d" " -f1); [ "$src_pid" = "$c_pid" ] || printf "patch-id mismatch:\n  source:   %s %s\n  backport: %s %s\n" "$src" "$(git show -s --format=%s "$src")" "$c" "$(git show -s --format=%s "$c")"; done
```

Expected output, only one commit needed a conflict resolution, but it's a trivial commit that drops a CI check:

```
patch-id mismatch:
  source:   0e0c3b25eb5c0e4652a639f2e04f66096aec61b0 tests: remove Android e2e tests.
  backport: 83a0b2ddc4cf69c3af5b8810163c248009c45e06 tests: remove Android e2e tests.
```


backend/devices/bitbox02bootloader/firmware.go
```diff
@@ -141,16 +179,61 @@ func bundledFirmware(product bitbox02common.Product) (*firmwareInfo, error) {
...
+) (*firmwareInfo, error) {
 	firmwares, ok := bundledFirmwares[product]
 	if !ok {
 		return nil, errp.New("unrecognized product")
 	}
+	return nextFirmwareFromList(firmwares, currentFirmwareVersion), nil
+}
 
-	for _, fwInfo := range firmwares {
+func nextFirmwareFromList(
```
bznein: just to make sure I understand, what is the point of extracting this function? It doesn't seem to be used anywhere else and it doesn't change the current logic

bznein (approved):
Overall LGTM I looked mostly at the three newly authored commits.

backend/devices/bitbox02bootloader/firmware.go
benma: Good catch, I used it in a unit test that I since removed. I undid this.

PR #4210, opened 2026-07-05
backend: restrict config file permissions

benma:


backend/arguments/arguments.go
```diff
@@ -62,30 +62,43 @@ func NewArguments(
 		panic("Cannot use -regtest with -mainnet.")
 	}
 
+	if err := utilconfig.EnsurePrivateDir(mainDirectoryPath); err != nil {
+		panic("Cannot create the main directory.")
+	}
+
+	appConfigFilename := path.Join(mainDirectoryPath, "config.json")
+	accountsConfigFilename := path.Join(mainDirectoryPath, "accounts.json")
```
bznein: [nit] can we move these down below right before they are used?

util/config/appdir.go
```diff
@@ -69,6 +76,57 @@ func AppDir() string {
...
+func ensurePrivateFile(filename string, ignoreMissing bool) error {
+	info, err := os.Lstat(filename)
+	if ignoreMissing && os.IsNotExist(err) {
+		return nil
+	}
+	if err != nil {
+		return errp.WithStack(err)
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		return nil
```
bznein, line 123: can we add a comment to this line? Is not immediately clear what it does

bznein (commented):
Codex:

  - [P2] Tighten existing rotated log files too — /home/bznein-coding-agent/agent-work/
    bitbox-wallet-app/util/logging/logger.go:79-79
    When a user already has a pre-existing log.txt.1 with broad permissions and the
    current log.txt is below maxLogFileSizeBytes, startup only calls
    EnsurePrivateFile(name) for the active log. The rotated log is still included by
    WriteCombinedLog and can contain the same sensitive diagnostics, so it remains
    world-readable until some future rotation/truncation happens. Please also restrict
    name + rotatedSuffix when it exists.

backend/arguments/arguments.go
benma: Done

util/config/appdir.go
benma: :+1: added

bznein:
Sorry @benma I havent's seen a reply to https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4210#pullrequestreview-4634164228 so I don't know if you saw that/what you think about it

benma:
> Sorry @benma I havent's seen a reply to [#4210 (review)](https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4210#pullrequestreview-4634164228) so I don't know if you saw that/what you think about it

I did miss it - addressed now, thanks!

bznein (approved):
utACK

PR #4211, opened 2026-07-05
dev: bind servers to loopback by default

benma:
servewallet and Vite should not listen on every host interface during normal local development. Dev mode exposes unauthenticated local services.

Docker still needs BITBOX_DEV_BIND_HOST=0.0.0.0 inside the container because published ports are reached through the container's external interface. The published host ports are bound to 127.0.0.1 so they remain accessible from the host without exposing them to the LAN.

### DIFF
--- a/cmd/servewallet/main.go
+++ b/cmd/servewallet/main.go
@@ -26,12 +26,20 @@ import (
 )
 
 const (
-	port    = 8082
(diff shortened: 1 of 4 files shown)


thisconnect (approved):
LGTM thanks

tested normal webdev + servewallet

PR #4213, opened 2026-07-06
frontend: harden AOPP VASP matching

benma:
AOPP requester branding matched known VASPs with a raw string suffix. A callback host like evilbitcoinsuisse.com could therefore render Bitcoin Suisse branding even though it is not that domain or one of its subdomains.

Require an exact host match or a dot-boundary subdomain match, and keep the real callback host visible when branding a subdomain. Add regression tests for exact, subdomain, and suffix-spoof hosts.

### DIFF
--- a/frontends/web/src/components/aopp/vasp.tsx
+++ b/frontends/web/src/components/aopp/vasp.tsx
@@ -36,14 +36,18 @@ const VASPHostnameMap: TVASPMap = {
   'testing.aopp.group': 'AOPP.group',
 };
 
+const matchesVaspHostname = (hostname: string, vasp: string): boolean => (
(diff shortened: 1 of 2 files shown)


frontends/web/src/components/aopp/vasp.tsx
```diff
@@ -36,14 +36,18 @@ const VASPHostnameMap: TVASPMap = {
   'testing.aopp.group': 'AOPP.group',
 };
 
+const matchesVaspHostname = (hostname: string, vasp: string): boolean => (
+  hostname === vasp || hostname.endsWith(`.${vasp}`)
```
thisconnect, line 40: 👍

thisconnect (approved):
untested LGTM thanks

PR #4214, opened 2026-07-06
lightning: add lnurl pay send flow

Beerosagos:
Add LNURL Pay support to the Lightning send flow while keeping rail selection inside the backend lightning package. Parse payment inputs into explicit BOLT11 and LNURL Pay variants, prepare and send each rail through the shared payment endpoints, and return localized error codes for invalid input and amounts.

Refactor the frontend review step into rail-specific BOLT11 and LNURL Pay components. Pass each component only the payload it needs, keep BOLT11 behavior aligned with the existing review flow, and keep LNURL Pay as an amount-entry flow within the server-provided min and max bounds.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/lightning/payments.go
+++ b/backend/lightning/payments.go
@@ -3,6 +3,7 @@
 package lightning
 
 import (
+	"encoding/json"
 	"errors"
 	"math/big"
 	"strconv"
@@ -16,20 +17,32 @@ import (
 )
 
 const (
-	errPaymentApprovalRequired     errp.ErrorCode = "paymentApprovalRequired"
-	errLightningInsufficientFunds  errp.ErrorCode = "lightningInsufficientFunds"
-	errLightningInvoiceAlreadyUsed errp.ErrorCode = "lightningInvoiceAlreadyUsed"
+	errPaymentApprovalRequired      errp.ErrorCode = "paymentApprovalRequired"
+	errLightningInvalidAmount       errp.ErrorCode = "lightningInvalidAmount"
+	errLightningInvalidPaymentInput errp.ErrorCode = "lightningInvalidPaymentInput"
+	errLightningInsufficientFunds   errp.ErrorCode = "lightningInsufficientFunds"
+	errLightningInvoiceAlreadyUsed  errp.ErrorCode = "lightningInvoiceAlreadyUsed"
 )
 
-type lightningInvoice struct {
-	Bolt11      string  `json:"bolt11"`
+type lightningBolt11Invoice struct {
+	Invoice     string  `json:"invoice"`
 	Description *string `json:"description,omitempty"`
 	AmountSat   *uint64 `json:"amountSat,omitempty"`
 }
 
-type parsePaymentInputResponse struct {
-	Type    string           `json:"type"`
-	Invoice lightningInvoice `json:"invoice"`
+type lightningLNURLPay struct {
+	Input        string  `json:"input"`
+	Address      *string `json:"address,omitempty"`
+	Domain       string  `json:"domain"`
+	Description  *string `json:"description,omitempty"`
+	MinAmountSat uint64  `json:"minAmountSat"`
+	MaxAmountSat uint64  `json:"maxAmountSat"`
+}
+
+type paymentInput struct {
+	Type     string                  `json:"type"`
+	Bolt11   *lightningBolt11Invoice `json:"invoice,omitempty"`
+	LNURLPay *lightningLNURLPay      `json:"lnurlPay,omitempty"`
 }
 
 type lightningPayment struct {
@@ -50,12 +63,14 @@ type receivePaymentResponse struct {
 }
 
 type preparePaymentRequest struct {
-	Bolt11    string  `json:"bolt11"`
-	AmountSat *uint64 `json:"amountSat"`
+	Type         string  `json:"type"`
+	PaymentInput string  `json:"paymentInput"`
+	AmountSat    *uint64 `json:"amountSat,omitempty"`
 }
 
 type sendPaymentRequest struct {
-	Bolt11         string  `json:"bolt11"`
+	Type           string  `json:"type"`
+	PaymentInput   string  `json:"paymentInput"`
 	AmountSat      *uint64 `json:"amountSat"`
 	ApprovedFeeSat uint64  `json:"approvedFeeSat"`
 }
@@ -66,14 +81,73 @@ type paymentFee struct {
 	TotalDebitSat uint64 `json:"totalDebitSat"`
 }
 
+const (
+	paymentInputTypeBolt11   = "bolt11"
+	paymentInputTypeLNURLPay = "lnurlPay"
+)
+
+type msatToSatRounding int
+
+const (
+	roundToFloor msatToSatRounding = iota
+	roundToCeil
+)
+
+func msatToSat(msat uint64, rounding msatToSatRounding) uint64 {
+	if rounding == roundToCeil {
+		return (msat + 999) / 1000
+	}
+	return msat / 1000
+}
+
+func validateLNURLPayAmount(payRequest breez_sdk_spark.LnurlPayRequestDetails, amountSat uint64) error {
+	if amountSat == 0 {
+		return errLightningInvalidAmount
+	}
+	if amountSat < msatToSat(payRequest.MinSendable, roundToCeil) ||
+		amountSat > msatToSat(payRequest.MaxSendable, roundToFloor) {
+		return errLightningInvalidAmount
+	}
+	return nil
+}
+
+func lnurlPayDescription(metadataStr string) *string {
+	var metadata [][]string
+	if err := json.Unmarshal([]byte(metadataStr), &metadata); err != nil {
+		return nil
+	}
+	for _, entry := range metadata {
+		// LUD-06 requires a text/plain entry:
+		// https://github.com/lnurl/luds/blob/luds/06.md
+		// Use the first valid one as the payment description and ignore malformed entries.
+		if len(entry) >= 2 && entry[0] == "text/plain" {
+			description := entry[1]
+			return &description
+		}
+	}
+	return nil
+}
+
+func toLightningLNURLPay(inputStr string, payRequest breez_sdk_spark.LnurlPayRequestDetails) lightningLNURLPay {
+	return lightningLNURLPay{
+		Input:        inputStr,
+		Address:      payRequest.Address,
+		Domain:       payRequest.Domain,
+		Description:  lnurlPayDescription(payRequest.MetadataStr),
+		MinAmountSat: msatToSat(payRequest.MinSendable, roundToCeil),
+		MaxAmountSat: msatToSat(payRequest.MaxSendable, roundToFloor),
+	}
+}
+
 // ParsePaymentInput validates and classifies a lightning input string.
-func (lightning *Lightning) ParsePaymentInput(inputStr string) (*parsePaymentInputResponse, error) {
+func (lightning *Lightning) ParsePaymentInput(inputStr string) (*paymentInput, error) {
 	if err := lightning.CheckActive(); err != nil {
 		return nil, err
 	}
 	input, err := lightning.sdkService.Parse(inputStr)
 	if err != nil {
-		return nil, err
+		lightning.log.WithError(err).Error("Parse lightning payment input failed")
+		return nil, errLightningInvalidPaymentInput
 	}
 
 	switch inputType := input.(type) {
@@ -85,14 +159,14 @@ func (lightning *Lightning) ParsePaymentInput(inputStr string) (*parsePaymentInp
 		var amountSat *uint64
 		if inputType.Field0.AmountMsat != nil {
 			amount = strconv.FormatUint(*inputType.Field0.AmountMsat, 10)
-			value := *inputType.Field0.AmountMsat / 1000
+			value := msatToSat(*inputType.Field0.AmountMsat, roundToCeil)
 			amountSat = &value
 		}
 		lightning.log.Printf("Input is BOLT11 invoice for %s msats", amount)
-		return &parsePaymentInputResponse{
-			Type: "bolt11",
-			Invoice: lightningInvoice{
-				Bolt11:      inputType.Field0.Invoice.Bolt11,
+		return &paymentInput{
+			Type: paymentInputTypeBolt11,
+			Bolt11: &lightningBolt11Invoice{
+				Invoice:     inputType.Field0.Invoice.Bolt11,
 				Description: inputType.Field0.Description,
 				AmountSat:   amountSat,
 			},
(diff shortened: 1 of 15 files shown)


Beerosagos:
@coderabbit review

frontends/web/src/routes/lightning/send/components/lnurl-pay-review-step.tsx
```diff
@@ -0,0 +1,217 @@
...
+}: TProps) => {
+  const { t } = useTranslation();
+  const mounted = useMountedRef();
+  const customAmountRef = useRef<number>();
+  const [customAmount, setCustomAmount] = useState<number>();
+  const debouncedCustomAmount = useDebounce(customAmount, 300);
+  const [preparedPayment, setPreparedPayment] = useState<TPreparedPayment>(emptyPreparedPayment);
+  const [isSending, setIsSending] = useState(false);
+  const [sendError, setSendError] = useState<string>();
+
```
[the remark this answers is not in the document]
Beerosagos: I don't think we need this: the min/max sendable amounts are defined at lnurl server level and there are no reasonable use cases for making them the same amount. I'd consider it an edge case and avoid adding code clutter to improve the related ui.

frontends/web/src/routes/lightning/send/components/bolt11-review-step.tsx
```diff
@@ -0,0 +1,247 @@
...
+          disabled={!canSend}>
+          {t('generic.send')}
+        </Button>
+        <Button secondary onClick={() => backToPaymentInput()}>
+          {t('button.back')}
+        </Button>
+      </ViewButtons>
+    </View>
+  );
+};
```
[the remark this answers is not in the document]
Beerosagos: fixed

frontends/web/src/routes/lightning/send/components/bolt11-review-step.tsx
```diff
@@ -0,0 +1,247 @@
...
+        return;
+      }
+      if (error instanceof TSdkError && error.code === 'paymentApprovalRequired') {
+        setSendError(errorMessage);
+        await preparePayment(needsCustomAmount ? amountSat : undefined);
+        return;
+      }
+
+      setSendError(errorMessage);
+    }
```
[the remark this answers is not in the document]
Beerosagos: fixed

frontends/web/src/routes/lightning/send/components/payment-review.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { type TFunction } from 'i18next';
+import { type TPreparePaymentResponse, TSdkError } from '@/api/lightning';
+
+export type TPreparedPayment = {
+  amountSat?: number;
+  fees?: TPreparePaymentResponse;
+  error?: string;
+  isPreparing: boolean;
+};
```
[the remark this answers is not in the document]
Beerosagos: fixed

backend/lightning/payments.go
```diff
@@ -66,14 +81,66 @@ type paymentFee struct {
...
+	paymentInputTypeBolt11   = "bolt11"
+	paymentInputTypeLNURLPay = "lnurlPay"
+)
+
+func msatToCeilSat(msat uint64) uint64 {
+	if msat%1000 == 0 {
+		return msat / 1000
+	}
+	return msat/1000 + 1
+}
```
bznein: [nit] not even sure if it's nicer to read or not, but 

```go
func msatToCeilSat(msat uint64) uint64 {
	return (msat + 999) / 1000
}
```
would be functionally the same (provided msat + 999 doesn't overflow)

backend/lightning/payments.go
```diff
@@ -66,14 +81,66 @@ type paymentFee struct {
...
+
+func msatToFloorSat(msat uint64) uint64 {
+	return msat / 1000
+}
```
bznein: overall, I think these two methods could be combined into a single one with an additional parameter, wdyt?

backend/lightning/payments.go
```diff
@@ -66,14 +81,66 @@ type paymentFee struct {
...
+	if err := json.Unmarshal([]byte(metadataStr), &metadata); err != nil {
+		return nil
+	}
+	for _, entry := range metadata {
+		if len(entry) >= 2 && entry[0] == "text/plain" {
+			description := entry[1]
+			return &description
+		}
+	}
+	return nil
```
bznein, line 128: can we add some comments here please? It's not super clear what this does :)

bznein (commented):
I did a first quick pass on the backend, but I'll take some time to review more thoroughly

backend/lightning/payments.go
Beerosagos: done 👍

Beerosagos:
@bznein all done :+1:

bznein (approved):
backend untested LGTM

Beerosagos:
@thisconnect can you have a look at the frontend? :pray:

frontends/web/src/api/lightning.ts
```diff
@@ -51,14 +60,25 @@ export type TReceivePaymentResponse = {
...
 export type TSendPaymentRequest = {
-  bolt11: string;
+  type: TPaymentInputType.BOLT11;
+  paymentInput: string;
   amountSat?: number;
   approvedFeeSat: number;
+} | {
+  type: TPaymentInputType.LNURL_PAY;
+  paymentInput: string;
+  amountSat: number;
```
thisconnect, line 71: Question: Is it correct that the only difference between BOLT11 and LNURL_PAY SendPaymentRequest is that amountSat is optional for BOLT11, but mandatory for LNURL_PAY?

Same question for PreparePaymentRequest

frontends/web/src/locales/en/app.json
```diff
@@ -918,6 +918,8 @@
     "aoppVersion": "Unknown version.",
     "keystoreTimeout": "Wallet request expired. Please try again.",
     "lightningInsufficientFunds": "Insufficient funds in your Lightning wallet.",
+    "lightningInvalidAmount": "Invalid amount.",
+    "lightningInvalidPaymentInput": "Invalid invoice or payment request.",
```
thisconnect, line 922: Could you point to the code where these 2 keys are used?

frontends/web/src/routes/lightning/send/components/lnurl-pay-review-step.tsx
```diff
@@ -0,0 +1,95 @@
...
+              min={lnurlPay.minAmountSat}
+              max={lnurlPay.maxAmountSat}
+              label={t('lightning.receive.amountSats.label')}
+              placeholder={t('lightning.receive.amountSats.placeholder')}
+              id="amountSatsInput"
+              onInput={(event) => {
+                const amount = event.currentTarget.valueAsNumber;
+                setCustomAmount(Number.isNaN(amount) ? undefined : amount);
+              }}
+              value={customAmount === undefined ? '' : `${customAmount}`}
```
thisconnect: nit


```suggestion
              value={customAmount}
```

frontends/web/src/routes/lightning/send/components/review-step.tsx
```diff
@@ -1,316 +1,35 @@
...
-  }, [customAmount, debouncedCustomAmount, invoice, preparePayment]);
-
-
-  if (!invoice || !paymentDetails || paymentDetails.type !== TPaymentInputTypeVariant.BOLT11) {
-    return null;
-  }
-
-  if (isSending) {
-    return <SendingSpinner />;
+  if (paymentInput.type === TPaymentInputType.BOLT11) {
```
thisconnect: nit: fine for now, but in the future if we add other types (bolt12?) this would be nicer as switch / case imo
```
switch(type)
  case: TPaymentInputType.BOLT11
  return (
    <Bolt11ReviewStep>
  );
  case: TPaymentInputType.LNURL
  return (
    <Bolt11ReviewStep>
  )
```

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+    return error.message;
+  }
+  return String(error);
+};
+
+type TPaymentReviewDetails = TLightningBolt11Invoice | TLightningLNURLPay;
+
+const isBolt11Invoice = (paymentDetails: TPaymentReviewDetails): paymentDetails is TLightningBolt11Invoice => (
+  'invoice' in paymentDetails
```
thisconnect: usePaymentReview is used in `<Bolt11ReviewStep>` and `<LNURLPayReviewStep>` so we should already know what type it is.

small nit: instead checking for `'invoice' in paymentDetails` key, wouldn't it make sense to add type (`type: TPaymentInputType.BOLT11 | TPaymentInputType.LNURL_PAY`) to `TLightningBolt11Invoice` and `TLightningLNURLPay` or similar?

no blocker it's also ok as is.

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
```diff
@@ -0,0 +1,259 @@
...
+        type: TPaymentInputType.BOLT11 as const,
+        paymentInput: paymentDetails.invoice,
+        amountSat,
+      };
+    } else {
+      if (!isValidAmount(paymentDetails, amountSat)) {
+        return;
+      }
+      preparePaymentRequest = {
+        type: TPaymentInputType.LNURL_PAY as const,
```
thisconnect: here and before in the other assignment

```diff
@@ -79,13 +79,13 @@ export const usePaymentReview = ({
     : undefined;
 
   const preparePayment = useCallback(async (amountSat?: number) => {
-    let preparePaymentRequest;
+    let preparePaymentRequest: TPreparePaymentRequest;
     if (isBolt11Invoice(paymentDetails)) {
       if (needsCustomAmount && !isValidAmount(paymentDetails, amountSat)) {
         return;
       }
       preparePaymentRequest = {
-        type: TPaymentInputType.BOLT11 as const,
+        type: TPaymentInputType.BOLT11,
         paymentInput: paymentDetails.invoice,
         amountSat,
       };
@@ -94,7 +94,7 @@ export const usePaymentReview = ({
         return;
       }
       preparePaymentRequest = {
-        type: TPaymentInputType.LNURL_PAY as const,
+        type: TPaymentInputType.LNURL_PAY,
         paymentInput: paymentDetails.input,
         amountSat,
       };
```

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
```diff
@@ -0,0 +1,259 @@
...
+    try {
+      const sendPaymentRequest = isBolt11Invoice(paymentDetails)
+        ? {
+          type: TPaymentInputType.BOLT11 as const,
+          paymentInput: paymentDetails.invoice,
+          amountSat: paymentDetails.amountSat === undefined ? currentAmountSat : undefined,
+          approvedFeeSat: fees.feeSat,
+        }
+        : {
+          type: TPaymentInputType.LNURL_PAY as const,
```
thisconnect: similar to the previous comment

```diff
@@ -166,15 +166,15 @@ export const usePaymentReview = ({
     setSendError(undefined);
 
     try {
-      const sendPaymentRequest = isBolt11Invoice(paymentDetails)
+      const sendPaymentRequest: TSendPaymentRequest = isBolt11Invoice(paymentDetails)
         ? {
-          type: TPaymentInputType.BOLT11 as const,
+          type: TPaymentInputType.BOLT11,
           paymentInput: paymentDetails.invoice,
           amountSat: paymentDetails.amountSat === undefined ? currentAmountSat : undefined,
           approvedFeeSat: fees.feeSat,
         }
         : {
-          type: TPaymentInputType.LNURL_PAY as const,
+          type: TPaymentInputType.LNURL_PAY,
           paymentInput: paymentDetails.input,
           amountSat: currentAmountSat,
           approvedFeeSat: fees.feeSat,

```

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
```diff
@@ -0,0 +1,259 @@
...
+  ]);
+
+  useEffect(() => {
+    setCustomAmount(undefined);
+    setPreparedPayment(undefined);
+    setIsSending(false);
+    setSendError(undefined);
+
+    if (!needsCustomAmount) {
+      void preparePayment();
```
thisconnect: here and further down in this hook

```suggestion
     preparePayment();
```

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
```diff
@@ -0,0 +1,259 @@
...
+      if (error instanceof TSdkError && error.code === 'lightningInvoiceAlreadyUsed') {
+        backToPaymentInput(errorMessage);
+        return;
+      }
+      if (error instanceof TSdkError && error.code === 'paymentApprovalRequired') {
+        setSendError(errorMessage);
+        const amountSat = isBolt11Invoice(paymentDetails) && paymentDetails.amountSat !== undefined
+          ? undefined
+          : currentAmountSat;
+        await preparePayment(amountSat);
```
thisconnect, line 206: I am reading something like "in case of error: set amount to undefined in case of Bolt11Invoice, but to currentAmountSat in case of lnurl."

but I am not 100% why that is.

Could you add a comment why this is needed?

thisconnect (approved):
LGTM with some questions and small nit

frontends/web/src/api/lightning.ts
Beerosagos: Correct. The `paymentInput` value will also have different formats, but it's a string in both cases.

frontends/web/src/locales/en/app.json
Beerosagos: `use-payment-review.ts`: `toErrorMessage()` takes the backend error codes defined in `backend/lightning/payments.go` as params to get the localized messages.

thisconnect: hm that is a bit unfortunate. 

all other errors can be found by searching for the key in the codebase within frontends/web

other translatable errors we dont use the key from the backend but pass errorCode. see for example. e104

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
Beerosagos: Bolt11Invoice can already define an amount: we don't need to pass it to the prepare call in that case. So we use `currentAmountSat`, which is the user-defined amount,  for amountless Bolt11 invoices and LNURLPay (which is amountless by definition). Will add a comment 🙏

thisconnect: That all makes sense, I'm more wondering why this has to be done in the catch in case of postSendPayment error.

frontends/web/src/locales/en/app.json
Beerosagos: Added a frontend map to translate backend error codes into i18n keys

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
Beerosagos: Good point, added a type field inside `TPaymentReviewDetails` to improve it.

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
Beerosagos: This is an edge case: it is possible that the payment fee retrieved during the prepare phase is no longer valid when trying to send. In this case, the send call returns a PAYMENT_APPROVAL_REQUIRED error and we need to re-prepare the payment to display the new fee and confirm sending again.

Beerosagos:
rebased

Beerosagos:
@thisconnect PTAL :pray:

frontends/web/src/locales/en/app.json
thisconnect: thank you

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
thisconnect: Makes sense. Maybe small comment about this edge case that the fee may be expired could make this line more clear.

thisconnect (approved):
very nice untested LGTM

PR #4218, opened 2026-07-06
notes: track transaction note metadata

benma:
Empty notes become tombstones, and we add modified timestamps. This is necessary so BitBoxSync can merge conflicts and sync deletions.

### DIFF
--- a/backend/accounts/notes/notes.go
+++ b/backend/accounts/notes/notes.go
@@ -5,8 +5,10 @@ package notes
 
 import (
 	"encoding/json"
+	"errors"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
 )
@@ -20,6 +22,22 @@ type Data struct {
 
 	// a map of transaction ID to transaction note.
 	TransactionNotes map[string]string `json:"transactions"`
+	// TransactionsMetadata tracks metadata for entries in TransactionNotes. It
+	// is stored separately so the longstanding notes file shape remains a
+	// simple txID -> note map for backwards compatibility.
+	TransactionsMetadata map[string]TransactionMetadata `json:"transactionsMetadata,omitempty"`
+}
+
+// TransactionMetadata stores metadata for a transaction note.
+type TransactionMetadata struct {
+	// ModifiedAt is when the transaction note was last edited.
+	ModifiedAt time.Time `json:"modifiedAt,omitempty"`
+}
+
+// TransactionNoteEntry is a transaction note plus sync metadata.
+type TransactionNoteEntry struct {
+	Note     string
+	Metadata TransactionMetadata
 }
 
 // read deserializes the json files into notes. If the file does not exist yet, no error is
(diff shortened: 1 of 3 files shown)


backend/accounts/notes/notes.go
```diff
@@ -124,11 +166,68 @@ func (notes *Notes) MergeLegacy(legacy *Notes) error {
...
+		}
+	}
+}
+
+func (metadata TransactionMetadata) isZero() bool {
+	return metadata.ModifiedAt.IsZero()
+}
+
+func (notes *Notes) setTransactionMetadata(txID string, metadata TransactionMetadata) {
+	if !metadata.ModifiedAt.IsZero() {
```
bznein, line 222: [nit] you can use the isZero method defined above here if I'm not mistaken

backend/accounts/notes/notes.go
```diff
@@ -124,11 +166,68 @@ func (notes *Notes) MergeLegacy(legacy *Notes) error {
...
+	}
+	notes.data.TransactionsMetadata[txID] = metadata
+}
+
+func validateTransactionNoteEntries(entries map[string]TransactionNoteEntry) error {
+	for _, entry := range entries {
+		if len(entry.Note) > MaxNoteLen {
+			return errp.Newf("Length of note must be smaller than %d. Got %d", MaxNoteLen, len(entry.Note))
+		}
+	}
```
bznein, line 237: [nit] what if instead of early return, we log all errors? 

and do something similar to

```golang
func validateTransactionNoteEntries(entries map[string]TransactionNoteEntry) error {
	var errs []error

	for _, entry := range entries {
		if len(entry.Note) > MaxNoteLen {
			errs = append(errs, errp.Newf(
				"Length of note must be smaller than %d. Got %d",
				MaxNoteLen,
				len(entry.Note),
			))
		}
	}

	if len(errs) > 0 {
		return errors.Join(errs...)
	}
	return nil
}
```

I don't feel particularly strong about it tbh, just an idea

bznein (commented):
Codex pointed out this:

The metadata/tombstone model mostly works, but blank notes for transactions that never  had notes are now recorded as deletion tombstones through existing send paths. That
  can corrupt sync semantics and should be fixed before merging.

  Review comment:

  - [P2] Avoid tombstones for missing blank notes — /home/bznein-coding-agent/agent-
    work/bitbox-wallet-app/backend/accounts/notes/notes.go:108-113
    When note == "" and the txID is absent, this now marks the note as changed and
    persists an empty entry with fresh ModifiedAt. Current BTC/ETH send paths call
    SetTxNote(..., txNote) unconditionally, so sending without a note creates a
    tombstone for a note the user never deleted; callers of TransactionNoteEntries()
    then cannot distinguish this from an actual deletion and may propagate false
    deletes. Keep missing+empty as a no-op, or skip blank send notes before calling
    SetTxNote.

backend/accounts/notes/notes.go
benma: Yeah why not. I am also considering truncating to the max length instead of failing, wdyt?

bznein: Yeah I think that makes sense!

benma:
> Codex pointed out this:
> 
> The metadata/tombstone model mostly works, but blank notes for transactions that never had notes are now recorded as deletion tombstones through existing send paths. That can corrupt sync semantics and should be fixed before merging.
> 
> Review comment:
> 
>     * [P2] Avoid tombstones for missing blank notes — /home/bznein-coding-agent/agent-
>       work/bitbox-wallet-app/backend/accounts/notes/notes.go:108-113
>       When note == "" and the txID is absent, this now marks the note as changed and
>       persists an empty entry with fresh ModifiedAt. Current BTC/ETH send paths call
>       SetTxNote(..., txNote) unconditionally, so sending without a note creates a
>       tombstone for a note the user never deleted; callers of TransactionNoteEntries()
>       then cannot distinguish this from an actual deletion and may propagate false
>       deletes. Keep missing+empty as a no-op, or skip blank send notes before calling
>       SetTxNote.

Very good catch! Fixed.

backend/accounts/notes/notes.go
benma: SetTxNote also has been failing on too large notes, so I'll leave it like this for consistency.

I applied your suggestion.

benma:
@bznein bump :smile:

bznein (approved):
utACK! Sorry for the delay I actually had the approval drafted and forgot to hit "send"!

PR #4221, opened 2026-07-07
add option to change portfolio percentage to performance

jadzeidan:
- Adds option for user to toggle percentage between portfolio value and portfolio performance. Performance percentage is calculated using money-weighted return (MWR) in order to take into account deposits and withdrawals.

- Adds playwright test for this feature


jadzeidan:
@coderabbitai

jadzeidan:
@coderabbitai pls review this PR, ty.

frontends/web/src/components/guide/guide.module.css
```diff
@@ -75,6 +75,16 @@
...
 }
 
+.list {
+    margin: 0.4em 0;
+    padding-left: var(--space-default);
+}
+
+.list li {
+    font-size: var(--size-default);
+    margin: 0.4em 0;
```
thisconnect: @jadzeidan and @coderabbitat Use px if the space should always be fixed. If the space should grow in with responsive mode with increased font-sizes then user `rem` and not `em`.

We use the `font-size: 62.5%;` trick, so 1rem = 10px.

frontends/web/src/routes/account/summary/percentage-diff.tsx
```diff
@@ -1,47 +1,106 @@
...
+  </svg>
+);
+
+const ValueBadgeIcon = ({ className }: TBadgeIconProps) => (
+  <svg aria-hidden="true" className={className || ''} fill="none" viewBox="0 0 14 14" xmlns="http://www.w3.org/2000/svg">
+    <path d="M0.3677 14H2.94161C3.14384 14 3.30931 13.8788 3.30931 13.7308V2.96154C3.30931 2.81346 3.14384 2.69231 2.94161 2.69231H0.3677C0.165464 2.69231 -1.66893e-06 2.81346 -1.66893e-06 2.96154V13.7308C-1.66893e-06 13.8788 0.165464 14 0.3677 14ZM5.67702 14H8.25093C8.45317 14 8.61863 13.8788 8.61863 13.7308V5.58654C8.61863 5.43846 8.45317 5.31731 8.25093 5.31731H5.67702C5.47479 5.31731 5.30932 5.43846 5.30932 5.58654V13.7308C5.30932 13.8788 5.47479 14 5.67702 14ZM10.8677 14H13.4416C13.6439 14 13.8093 13.8788 13.8093 13.7308V0.269231C13.8093 0.121154 13.6439 0 13.4416 0H10.8677C10.6655 0 10.5 0.121154 10.5 0.269231V13.7308C10.5 13.8788 10.6655 14 10.8677 14Z" fill="currentColor" />
+  </svg>
```
thisconnect: move to Icon component and use .svg files, if necessary one for normal and one for dark mode. see how other icons are done frontends/web/src/components/icon/icon.tsx

reason is so vitejs can decide if the svg should be inlined in the JS bundle or an external file and that the JS bundle doesn't get MB's of SVG's ... which then also react has to diff in the virtual DOM. cc @coderabbitai please learn this.

jadzeidan:
@coderabbitai please review again

frontends/web/src/contexts/AppProvider.tsx
```diff
@@ -92,8 +106,10 @@ export const AppProvider = ({ children }: TProps) => {
         setGuideExists,
         setHideAmounts,
         setChartDisplay,
+        setPortfolioPercentageType,
```
thisconnect: Remove as `setPortfolioPercentageType` is internal and sets the state internal in the provider.

Outside of the provider `updatePortfolioPercentageType` should be used .

frontends/web/src/contexts/AppContext.tsx
```diff
@@ -23,10 +25,12 @@ type AppContextProps = {
   setGuideExists: Dispatch<SetStateAction<boolean>>;
   setHideAmounts: Dispatch<SetStateAction<boolean>>;
   setChartDisplay: Dispatch<SetStateAction<TChartDisplay>>;
+  setPortfolioPercentageType: Dispatch<SetStateAction<TPortfolioPercentageType>>;
```
thisconnect: Remove as `setPortfolioPercentageType` is internal and sets the state internal in the provider.

Outside of the provider `updatePortfolioPercentageType` should be used .

frontends/web/src/locales/en/app.json
```diff
@@ -1132,9 +1136,13 @@
       "title": "How is the total amount calculated?"
     },
     "accountSummaryDescription": {
-      "text": "Here you can see the performance of your portfolio over time. A summary of your individual crypto accounts is displayed under the chart.",
+      "text": "Here you can see the value and performance of your portfolio over time. A summary of your individual crypto accounts is displayed under the chart.",
```
thisconnect, line 1141: After we merged I'd like to double check that weblate did update this key..

frontends/web/src/routes/account/summary/percentage-diff.tsx
```diff
@@ -1,47 +1,106 @@
...
-  const className = hasDifference ? styles[style] : '';
+  const className = hasDifference ? (styles[style] || '') : '';
+  const badgeClassName = [
+    styles.badge,
+    badgeVisible ? styles.badgeVisible : '',
+  ].filter(Boolean).join(' ');
+  const valueBadgeIconClassName = [
+    styles.badgeIcon,
+    styles.valueBadgeIcon,
+  ].filter(Boolean).join(' ');
```
thisconnect: style-nit: In older components we still have `[].join(' ')`. 

But in newer components we use template literals to contact multiple classnames, see for example  frontends/web/src/components/badge/badge.tsx

frontends/web/src/components/guide/entry.tsx
```diff
@@ -21,6 +21,47 @@ type TEntryProps = {
...
+
+    flushList();
+    content.push(<p key={`paragraph-${content.length}`}>{trimmedLine}</p>);
+  });
+
+  flushList();
+
+  return content;
+};
+
```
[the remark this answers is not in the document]
thisconnect: Entry component was intentionally kept very simple with only newline support.

As we want to render a list it will need to be extend to support this.

But it should already be possible without all this logic, as Entry component accepts children props we can just used that to pass the remaining lists and paragraph.

I'll add simpler a diff with all required changes, see comment below in app.json

frontends/web/src/locales/en/app.json
```diff
@@ -1132,9 +1136,13 @@
...
       "title": "What does this page show me?"
     },
+    "accountSummaryPercentageMeaning": {
+      "text": "The percentage on your portfolio chart shows one of two options for the selected time range:\n\n- Value over time: This is the change of your portfolio value over time, including deposits and withdrawals.\n- Performance: This is the money-weighted return (MWR). It takes into account deposits and withdrawals, so it reflects the performance of portfolio.\n\nYou can switch between performance and value over time by clicking on the percentage, or in Settings under Appearance. By default, the app shows value over time.",
+      "title": "What does the percentage mean?"
+    },
```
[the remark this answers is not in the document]
thisconnect: Instead of adding a new markup looking like non-standard "syntax" (one long string with newlines and dashes).

We already do use lists, see how other components did it. For example `passphrase.summary.understandList`  in passphrase.tsx

I'll add a diff to show how to do this in another comment.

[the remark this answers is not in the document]
thisconnect: Please use this approach, see diff below.

Note: Luckily we never used children with entry component, as it needs a small change

```diff
diff --git a/frontends/web/src/components/guide/entry.tsx b/frontends/web/src/components/guide/entry.tsx
index 6833b98fb..cb805637b 100644
--- a/frontends/web/src/components/guide/entry.tsx
+++ b/frontends/web/src/components/guide/entry.tsx
@@ -83,7 +83,8 @@ export const Entry = (props: TProps) => {
       <div className={[style.entryContent, shown ? style.expanded : ''].join(' ')}>
         {shown ? (
           <div className="flex-1">
             {entry.text.trim().split('\n').map((p, idx) => <p key={idx}>{p}</p>)}
+            {props.children}
             {entry.link && (
               <p>
                 {entry.link.url ? (
@@ -100,7 +101,6 @@ export const Entry = (props: TProps) => {
                 )}
               </p>
             )}
-            {props.children}
           </div>
         ) : null}
diff --git a/frontends/web/src/locales/en/app.json b/frontends/web/src/locales/en/app.json
index 2dcb1632d..b331c93ff 100644
--- a/frontends/web/src/locales/en/app.json
+++ b/frontends/web/src/locales/en/app.json
@@ -1140,7 +1140,12 @@
       "title": "What does this page show me?"
     },
     "accountSummaryPercentageMeaning": {
-      "text": "The percentage on your portfolio chart shows one of two options for the selected time range:\n\n- Value over time: This is the change of your portfolio value over time, including deposits and withdrawals.\n- Performance: This is the money-weighted return (MWR). It takes into account deposits and withdrawals, so it reflects the performance of your portfolio.\n\nYou can switch between performance and value over time by clicking on the percentage, or in Settings under Appearance. By default, the app shows value over time.",
+      "listOptions": [
+        "Value over time: This is the change of your portfolio value over time, including deposits and withdrawals.",
+        "Performance: This is the money-weighted return (MWR). It takes into account deposits and withdrawals, so it reflects the performance of your portfolio."
+      ],
+      "outro": "You can switch between performance and value over time by clicking on the percentage, or in Settings under Appearance. By default, the app shows value over time.",
+      "text": "The percentage on your portfolio chart shows one of two options for the selected time range:",
       "title": "What does the percentage mean?"
     },
     "accountTransactionAttributesBTC": {
diff --git a/frontends/web/src/routes/account/summary/accountssummary.tsx b/frontends/web/src/routes/account/summary/accountssummary.tsx
index 3d0551c36..ab819f966 100644
--- a/frontends/web/src/routes/account/summary/accountssummary.tsx
+++ b/frontends/web/src/routes/account/summary/accountssummary.tsx
@@ -7,6 +7,7 @@ import { TDevices } from '@/api/devices';
 import { statusChanged, syncdone } from '@/api/accountsync';
 import { unsubscribe } from '@/utils/subscriptions';
 import { TUnsubscribe } from '@/utils/transport-common';
+import { SimpleMarkup } from '@/utils/markup';
 import { useMountedRef } from '@/hooks/mount';
 import { GuideWrapper, GuidedContent, Header, Main } from '@/components/layout';
 import { View } from '@/components/view/view';
@@ -227,7 +228,22 @@ export const AccountsSummary = ({
         <Entry key="accountSummaryPercentageMeaning" entry={{
           text: t('guide.accountSummaryPercentageMeaning.text'),
           title: t('guide.accountSummaryPercentageMeaning.title'),
-        }} />
+        }}>
+          <ul>
+            <SimpleMarkup
+              tagName="li"
+              markup={t('guide.accountSummaryPercentageMeaning.listOptions.0')}
+            />
+            <SimpleMarkup
+              tagName="li"
+              markup={t('guide.accountSummaryPercentageMeaning.listOptions.1')}
+            />
+          </ul>
+          <SimpleMarkup
+            tagName="p"
+            markup={t('guide.accountSummaryPercentageMeaning.outro')}
+          />
+        </Entry>
         <Entry key="accountSummaryAmount" entry={{
           link: {
             text: 'www.coingecko.com',
```

frontends/web/src/locales/en/app.json
```diff
@@ -1196,6 +1204,10 @@
         "title": "Why is this useful?"
       }
     },
+    "appearancePortfolioPercentageOptions": {
+      "text": "- Value over time: This is the change of your portfolio value over time, including deposits and withdrawals.\n- Performance: This is the money-weighted return (MWR). It takes into account deposits and withdrawals, so it reflects the performance of your portfolio.",
```
thisconnect: Same as comment before

thisconnect (changes requested):
reviewed frontend, but it probably makes sense to review the backend part before continuing. cc @bznein

frontends/web/src/routes/account/summary/percentage-diff.tsx
thisconnect: could you formulate an AGENT.md change to take this into account from now on for new PRs or do you remember this in the future in reviews?

jadzeidan:
@thisconnect thanks for the review, addressed the comments. PTAL.

thisconnect (approved):
thank you frontend tested LGTM

I also tested with a fresh wallet and new tx, but the history is too short the % diff it takes a week I guess :)

backend/chart.go
```diff
@@ -95,6 +112,391 @@ func (backend *Backend) addChartData(
...
+type chartCashFlow struct {
+	Time           time.Time
+	Value          float64
+	ValueAvailable bool
+}
+
+type chartWeightedCashFlow struct {
+	Value           float64
+	RemainingWeight float64
+}
```
bznein: [nit] I'd be happy with a couple of comments here :) It is quite obscure what this types (and some of their fields, like `RemainingWeight`) are supposed to do

backend/chart.go
```diff
@@ -95,6 +112,391 @@ func (backend *Backend) addChartData(
...
+
+func findPerformanceStartEntry(entries []ChartEntry, from time.Time, cashFlows []chartCashFlow) *ChartEntry {
+	if len(entries) == 0 {
+		return nil
+	}
+
+	startIndex := 0
+	if !from.IsZero() {
+		startIndex = sort.Search(len(entries), func(i int) bool {
```
bznein: I assume here that we now that `entries` is sorted? can we add a comment?

backend/chart.go
```diff
@@ -95,6 +112,391 @@ func (backend *Backend) addChartData(
...
+	return false
+}
+
+func findPerformanceStartEntry(entries []ChartEntry, from time.Time, cashFlows []chartCashFlow) *ChartEntry {
+	if len(entries) == 0 {
+		return nil
+	}
+
+	startIndex := 0
+	if !from.IsZero() {
```
bznein: [nit] we technically don't need this; when `isZero` is true, its value is `-62135596800`; so we will return `startIndex = 0` in the code below unless we have entries that have a timestamp lower than that (so before the 1st of January of year 0001 :D)

backend/chart.go
```diff
@@ -95,6 +112,391 @@ func (backend *Backend) addChartData(
...
+		if entries[i].Value > 0 {
+			if i > startIndex && entries[i-1].Value == 0 {
+				previousEntry := &entries[i-1]
+				if hasCashFlowBetween(cashFlows, time.Unix(previousEntry.Time, 0), time.Unix(entries[i].Time, 0)) {
+					return previousEntry
+				}
+			}
+			return &entries[i]
+		}
+	}
```
bznein: I think this would be slightly easier to read as 

```go
for i := startIndex; i < len(entries); i++ {
	if entries[i].Value <= 0 {
		continue
	}

	if i > startIndex && entries[i-1].Value == 0 {
		previousEntry := &entries[i-1]
		if hasCashFlowBetween(
			cashFlows,
			time.Unix(previousEntry.Time, 0),
			time.Unix(entries[i].Time, 0),
		) {
			return previousEntry
		}
	}

	return &entries[i]
}
```

bznein (commented):
I did a fly-by check with some readability comments, but I will need some more time as the backend changes are quite complex

benma:
Codex 5.6 Max review:

---

  ### Findings

  1. High — month/year performance uses a different period than the visible chart.
     The backend starts month/year at UTC midnight, while the frontend starts at the current UTC hour. With daily data, lightweight-charts selects the next daily point, so MWR includes one additional day compared
     with both the visible chart and “value over time.” This can materially change the percentage when that day contains volatility or cash flows. Align both boundaries, ideally with one canonical range
     definition.
     backend/chart.go:465 (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/ac8353a1ab6f1e0d9329299e6ea784d83cd5ee68/backend/chart.go#L465-L487)
     chart.tsx:51 (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/ac8353a1ab6f1e0d9329299e6ea784d83cd5ee68/frontends/web/src/routes/account/summary/chart.tsx#L51-L99)

  2. High — cash-flow-only investment periods are silently omitted.
     findPerformanceStartEntry() requires a positive sampled chart balance. If funds are received and fully sent again between two daily samples, every chart entry can remain zero and performance becomes null,
     even though the cash flows produce a valid return. If an older closed investment period precedes a later positive balance, the older cash flows are also discarded. Start selection needs to consider the first
     relevant cash flow, not only the first positive chart sample. Add tests for a same-day receive/send round trip and an earlier closed investment cycle.
     backend/chart.go:139 (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/ac8353a1ab6f1e0d9329299e6ea784d83cd5ee68/backend/chart.go#L139-L164)

  3. Medium — zero or unavailable performance makes the toggle disappear.
     Zero and null both make hasDifference false. PercentageDiff then renders an empty, zero-padding button, so selecting performance can remove the only chart-level way to switch back. Zero is a valid result and
     should display 0.00%; unavailable performance needs a visible placeholder/control with an accessible label. The E2E test only covers nonzero values.
     chart.tsx:671 (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/ac8353a1ab6f1e0d9329299e6ea784d83cd5ee68/frontends/web/src/routes/account/summary/chart.tsx#L671-L739)
     percentage-diff.tsx:56 (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/ac8353a1ab6f1e0d9329299e6ea784d83cd5ee68/frontends/web/src/routes/account/summary/percentage-diff.tsx#L56-L83)

  ### Simplicity/refactor assessment

  A targeted preliminary refactor would materially improve this change:

  - Consolidate the two existing frontend range-setting switches into one pure range helper, then make backend performance use the identical boundary contract—preferably returning the actual start timestamp with
    each metric.

  - Move the numerical and cash-flow code into chart_performance.go; adding 441 lines to chart.go mixes chart assembly, transaction interpretation, and root solving.
  - Put “latest confirmed balance” on OrderedTransactions or the account domain instead of inferring it in chart code.

  The root solver itself is necessarily nontrivial for exact MWR; replacing it with Modified Dietz merely to shorten the code would change the metric.

jadzeidan:
> The root solver itself is necessarily nontrivial for exact MWR; replacing it with Modified Dietz merely to shorten the code would change the metric.

fyi Modified Dietz was originally used for simplicity but is actually not a good approximation for when assets that are volatile and where in/out flows can be large (both of which apply to crypto). Therefore XIRR is used since it is much more accurate.

thisconnect:
ping @jadzeidan PTAL

jadzeidan:
Addressed comments, PTAL

backend/chart_performance.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+	return scale
+}
+
+func chartHasSignChange(a, b float64) bool {
+	return (a < 0 && b > 0) || (a > 0 && b < 0)
+}
+
+// Solve EV = BV*(1+r) + sum(CF_i*(1+r)^remaining_i) in log-return space.
+func solveChartMoneyWeightedReturn(
```
bznein, line 118: This method is really hard to follow; I think it would be easier to rid with some comments at the very least, and I think there are parts that can be extracted; the flow for "positive" and "negative" are pretty much identical.

For example, codex simplified it like this: https://github.com/bznein/bitbox-wallet-app/commit/95549c4b815314b0947ee8561b7fab152c13c312

bznein (commented):
Overall it looks of to me but the backend code is really hard to follow... I left one comment and I also have a followup commit from codex make things easier to read: https://github.com/bznein/bitbox-wallet-app/commit/26f2d29c7ed128fa2ff7a14a262bdf199bb78473#diff-88a8d3bb3500d3c98b0b11f9f058a5f59e3b500eb49d10105e578d9a16b9bc22

I don't necessarily want those changes to be made as they are in my branch but maybe you can go through some refinement with your LLM to see if we can make the backend diff a bit easier to review?

thisconnect:
There is some merge conflicts, please rebase

backend/chart_performance.go
jadzeidan: ack, PTAL when you get a chance to see if it is more clear now.

thisconnect:
IMO it would make sense to change the default to the new MWR calculation and keep the old one as legacy option.

backend/chart.go
```diff
@@ -357,17 +385,41 @@ func (backend *Backend) ChartData() (*Chart, error) {
...
+		chartPerformanceTotal = &performanceTot
+	}
+
+	chartDataDailyForPerformance := toSortedSlice(chartEntriesDaily, fiat)
+	chartDataHourlyForPerformance := toSortedSlice(chartEntriesHourly, fiat)
+	chartDataDaily := trimLeadingZeroes(chartDataDailyForPerformance, len(chartEntriesDaily) > 0)
+	chartDataHourly := trimLeadingZeroes(chartDataHourlyForPerformance, len(chartEntriesHourly) > 0)
+	sort.Slice(chartCashFlows, func(i, j int) bool {
+		return chartCashFlows[i].Time.Before(chartCashFlows[j].Time)
+	})
```
bznein: [nit] https://pkg.go.dev/slices#SortFunc this is the preferred method since Go 1.21 to sort slices with a custom function

backend/chart_performance.go
```diff
@@ -0,0 +1,448 @@
...
+		}
+
+		if currentStep == maxAbsLogReturn {
+			return chartLogReturnBracket{}, false
+		}
+	}
+}
+
+// bisectChartMoneyWeightedReturn returns an exactly sampled root from a
+// degenerate bracket, or narrows a sign-changing bracket to a root.
```
bznein: can we make it clearer what the boolean return value means?

bznein (commented):
backend overall LGTM (though I won't pretend I fully understood the math behind it!)

Attaching codex finding and a couple of personal comments in-line

Codex:

  Full review comments:

  - [P2] Align confirmed cash flows with the ending-balance cutoff — /home/bznein-
    coding-agent/agent-work/bitbox-wallet-app/backend/chart_performance.go:283-285
    When a confirmed Bitcoin transaction has a timestamp ahead of the local clock,
    LatestConfirmedBalance() includes its amount, but this filter excludes its cash
    flow. With unchanged prices, a deposit doubling the balance therefore reports 100%
    performance instead of 0%. Clamp future-dated confirmed cash flows to endTime, or
    apply the same cutoff to the ending balance.

  - [P2] Calculate full withdrawals before displaying zero percentages — /home/bznein-
    coding-agent/agent-work/bitbox-wallet-app/frontends/web/src/routes/account/summary/
    chart.tsx:341-343
    When a portfolio drops from a positive value to zero, the valueTo truthiness check
    produces valueDiff = 0. Previously that sentinel was hidden; the updated
    PercentageDiff now displays it as 0.00%, incorrectly reporting no change instead of
    -100.00%. Distinguish a zero ending balance from an unavailable balance before
    calculating the difference.

  - [P2] Preserve a clickable toggle when performance is unavailable — /home/bznein-
    coding-agent/agent-work/bitbox-wallet-app/frontends/web/src/routes/account/summary/
    percentage-diff.tsx:79-81
    When the selected range returns moneyWeightedReturn: null, the toggle has no
    content. Combined with .button having zero padding, no border, and no minimum
    dimensions, this removes its pointer/touch target, preventing users from switching
    back through the chart. Keep a visible, nonzero-size control even without a numeric
    placeholder; fireEvent.click() on an empty element does not verify real
    clickability.

  - [P3] Remove the unused Bitcoin conversion wrapper — /home/bznein-coding-agent/agent-    work/bitbox-wallet-app/backend/chart.go:282-282
    This replaces the last production call to convertBtcAmountToFiat; only
    TestConvertBtcAmountToFiat still references it. Remove the wrapper and obsolete
    test, or move that coverage to convertToFiat, rather than retaining test-only
    production code. This follows the removed-callee review guidance in AGENTS.md:205
    (AGENTS.md#L205).

jadzeidan:
@bznein addressed comments, PTAL

bznein (approved):
overall tested LGTM but I don't get why the option shows like this in my case:

Image

jadzeidan:
> overall tested LGTM but I don't get why the option shows like this in my case:

@bznein do you have no TXs in your wallet?

bznein:
> > overall tested LGTM but I don't get why the option shows like this in my case:
> 
> @bznein do you have no TXs in your wallet?

I do, and it also allows me to change that option, it seems just like a visual glitch.

But I also noticed a weird behaviour: when using a narrow screen, or mobile view, it shows regularly; the screenshot is only for wide screen :/

jadzeidan:
> > > overall tested LGTM but I don't get why the option shows like this in my case:
> > 
> > 
> > @bznein do you have no TXs in your wallet?
> 
> I do, and it also allows me to change that option, it seems just like a visual glitch.
> 
> But I also noticed a weird behaviour: when using a narrow screen, or mobile view, it shows regularly; the screenshot is only for wide screen :/

Ok I will test some more and try to fix the issue.

jadzeidan:
Re settings disabled UI bug

Seems it was a bug introduced in the last commit. Fixed now. @bznein PTAL, thanks!

bznein (approved):
tACK

PR #4222, opened 2026-07-08
frontend: lightning settings UI

shonsirsha:
Lightning settings UI. UI only (values hardcoded or no-op).

Screenshot 2026-07-08 at 06 37 55

Searchable:
Screenshot 2026-07-08 at 06 37 48

Lightning settings:
Screenshot 2026-07-08 at 06 38 00


bznein:
@coderabbit review

bznein (approved):
overall LGTM, just a question which we might want need some discussion.

Currently, when a LN account is present, the same option to disable the wallet appears twice. Once in advanced settings -> Disable lightning wallet and once in the settings you added -> Shut down

Ideally I think we should only have it once, and remove it from the generic "advanced settings", but we'd still need the "enable lightning wallet" when it is not active. Maybe we could always show the lightning wallet settings but only have an "enable" button there when LN is not active? Or at least have the enable and the "lightning settings" in the same spot on the list of settings so that either one of the two shows up but it doesn't "move around"?

I hope I was clear, I fear not :D

shonsirsha:
@bznein , Agree 100% - only one should show up. Not both.  Thank you.

Made it so that we only have "Lightning Settings". Updated screenshots:

Always visible:

Screenshot 2026-07-08 at 11 39 30


When LN is inactive:
Screenshot 2026-07-08 at 11 39 44

 🙏

PR #4230, opened 2026-07-08
show transaction indicators and tooltips in chart

sutterseba:


frontends/web/src/routes/account/summary/accountssummary.tsx
```diff
@@ -69,22 +72,58 @@ export const AccountsSummary = ({
     balance.coinCode !== 'lightning' || isLightningFeatureAvailable()
   ));
 
-  const getChartData = useCallback(async () => {
+  const getChartData = useCallback(async (includeTransactionMarkers = false) => {
+    if (chartRequestInFlight.current) {
+      chartMarkerRefreshQueued.current ||= includeTransactionMarkers;
```
thisconnect: on https://caniuse.com/?search=%7C%7C%3D logical or assignment only works for newer browser we sometimes have older engines on some Androids.

Could you rewrite without using logical or assignment?

frontends/web/src/routes/account/summary/chart.tsx
```diff
@@ -178,104 +200,335 @@ export const Chart = ({
...
-  const prevChartDataDaily = usePrevious(data.chartDataDaily);
-  const prevChartDataHourly = usePrevious(data.chartDataHourly);
-  const prevChartFiat = usePrevious(data.chartFiat);
-  const prevHideAmounts = usePrevious(hideAmounts);
-  const hasChartAnimationParam = searchParams.get('with-chart-animation');
+    const times = markerTimes.current;
+    let low = 0;
+    let high = times.length;
+    while (low < high) {
+      const middle = (low + high) >>> 1;
```
thisconnect: The unsigned right-shift operator `>>> 1` adds unnecessary mental overhead here. Since this is calculating the midpoint of an array range, could we use Math.floor((low + high) / 2) instead? It makes the intent clearer to readers.

thisconnect (changes requested):
please also rebase and run `make webfix`

thisconnect:
@coderabbitai review

backend/chart.go
```diff
@@ -71,6 +92,162 @@ type Chart struct {
...
+			if !includeChartTransaction(tx) {
+				continue
+			}
+
+			amount := tx.Amount
+			if tx.Type == accounts.TxTypeSend {
+				amount = tx.DeductedAmount
+			}
+			historicalRateMissing := latestHistoryTime.IsZero() || latestHistoryTime.Before(*tx.Timestamp)
+			estimated := historicalRateMissing && now.Sub(*tx.Timestamp) < 2*time.Hour
```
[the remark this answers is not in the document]
sutterseba: no change needed

sutterseba:
@thisconnect thank you, addressed comments

backend/chart.go
```diff
@@ -71,6 +92,162 @@ type Chart struct {
...
+}
+
+func includeChartTransaction(tx *accounts.TransactionData) bool {
+	if tx.Timestamp == nil {
+		return false
+	}
+	if tx.Status == accounts.TxStatusFailed {
+		return false
+	}
+	return tx.Type == accounts.TxTypeSend || tx.Type == accounts.TxTypeReceive
```
bznein, line 107: are we explicitly ignoring send-to-self txs? Is there a specific reason? Since they impact balance (via the fee) I think we should include them

backend/chart.go
```diff
@@ -71,6 +92,162 @@ type Chart struct {
...
+) []ChartTransactionMarker {
+	markers := make([]ChartTransactionMarker, 0, len(buckets))
+	for bucketTime, bucket := range buckets {
+		markers = append(markers, ChartTransactionMarker{
+			Time:    bucketTime,
+			Receive: bucket.receive.formatted(fiat),
+			Send:    bucket.send.formatted(fiat),
+		})
+	}
+	sort.Slice(markers, func(i, j int) bool { return markers[i].Time < markers[j].Time })
```
bznein, line 184: not a big deal, but since go 1.21 we have slices.SortFunc which is (marginally) better (https://pkg.go.dev/sort#Slice:~:text=type%27s%20Less%20method.-,Note%3A%20in%20many%20situations%2C%20the%20newer%20slices.SortFunc%20function%20is%20more%20ergonomic%20and%20runs%20faster.,-Example%20%C2%B6):

```go
slices.SortFunc(markers, func(a, b ChartTransactionMarker) int {
	return cmp.Compare(a.Time, b.Time)
})
```

bznein (commented):
overall backend LGTM (and tested); just one nit a comment about send-to-self

backend/chart.go
sutterseba: yes it was discussed to ignore regardless of the fee, main reasoning is that the dots are supposed to add context to chart movement, and there would be no visible chart movement for self transfers i.e. it would add more noise and clutter up the chart than add information

bznein (approved):
backend LGTM

backend/chart.go
sutterseba: updated

sutterseba:
rebased

PR #4235, opened 2026-07-10
lightning: add lightning address setup flow

Beerosagos:
Lightning wallets need a recoverable way to create and change the LNURL address that users share to receive payments. Previously the app only created a random address lazily when the address was read, which gave users no control over the username.

Add backend helpers and endpoints to expose the configured LNURL domain, check username availability, generate an available suggestion and register a chosen username. Keep automatic initial address creation on SDK connect so existing receive flows still get an address without an explicit settings step, but make regular address reads side-effect free.

Persist the time of a manual address change and reject further manual changes during the local 24 hour cooldown. Add the settings screen implementation that loads the current address, validates and debounces availability checks, offers generated usernames and saves the selected username through the new registration endpoint.

### DIFF
--- a/backend/lightning/address.go
+++ b/backend/lightning/address.go
@@ -7,22 +7,64 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"regexp"
+	"strings"
+	"time"
 
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/config"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/util/observable"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/util/observable/action"
 	"github.com/breez/breez-sdk-spark-go/breez_sdk_spark"
 )
 
 const (
-	lightningAddressDescription = "BitBoxApp Lightning wallet"
-	lightningAddressNumberMax   = 10000
-	maxLightningAddressAttempts = 5
+	lightningAddressDescription       = "BitBoxApp Lightning wallet"
+	lightningAddressNumberMax         = 10000
+	maxLightningAddressAttempts       = 5
+	maxLightningAddressUsernameLength = 64
+	lightningAddressChangeCooldown    = 24 * time.Hour
 )
+
+const (
+	errLightningAddressInvalidUsername     errp.ErrorCode = "lightningAddressInvalidUsername"
+	errLightningAddressUsernameUnavailable errp.ErrorCode = "lightningAddressUsernameUnavailable"
+	errLightningAddressChangeCooldown      errp.ErrorCode = "lightningAddressChangeCooldown"
+)
+
+var lightningAddressUsernameRegexp = regexp.MustCompile(`^[a-z0-9]+$`)
+var lightningAddressNow = time.Now
+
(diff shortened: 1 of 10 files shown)


Beerosagos:
@coderabbit review

backend/lightning/address.go
```diff
@@ -66,49 +163,167 @@ func (lightning *Lightning) ensureLightningAddress() (*string, error) {
...
+		if len(cfg.Accounts) == 0 {
+			return errp.New("Lightning not initialized")
+		}
+		cfg.Accounts[0].LightningAddressLastChangedAt = &changedAt
+		return nil
+	}); err != nil {
+		return errp.New("Error updating lightning address change timestamp")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Beerosagos: Fixed

frontends/web/src/routes/lightning/set-lnurl-address.tsx
```diff
@@ -1,63 +1,175 @@
...
+      setAddress(address ?? '');
+      setUsername(currentUsername);
+      setRegisteredUsername(currentUsername);
+      setAvailability('unchecked');
+    }).catch((err) => {
+      if (mounted.current) {
+        setError(toLightningErrorMessage(t, err));
+      }
+    });
+  }, [mounted, t]);
```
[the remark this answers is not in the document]
Beerosagos: @shonsirsha do you agree on this?

backend/lightning/address.go
```diff
@@ -7,24 +7,68 @@ import (
...
+	errLightningAddressChangeCooldown      errp.ErrorCode = "lightningAddressChangeCooldown"
+)
+
+var lightningAddressUsernameRegexp = regexp.MustCompile(`^[a-z0-9]+$`)
+var lightningAddressNow = time.Now
+
+// AddressAvailability describes whether a Lightning address username can be registered.
+type AddressAvailability struct {
+	Username  string `json:"username"`
+	Address   string `json:"address"`
```
bznein: Any reason to not simply use "GeneratedAddress" here? Do you see them drifting apart in the future?

backend/lightning/address.go
```diff
@@ -7,24 +7,68 @@ import (
...
 	return &info.LightningAddress
 }
 
+func (lightning *Lightning) lightningAddressForUsername(username string) string {
+	return username + "@" + lightning.lnurlDomain()
+}
+
+func normalizeLightningAddressUsername(username string) (string, error) {
+	normalized := strings.ToLower(strings.TrimSpace(username))
+	if normalized == "" ||
```
bznein: [nit] technically the first check is not needed because the regex won't match an empty string, so if you want you can drop it

even more nit, you could also embed the len check in the regex with smth like

```go
regexp.MustCompile(
    fmt.Sprintf(`^[a-z0-9]{1,%d}$`, maxLightningAddressUsernameLength),
)
```

But it's probably not worth it.

Up to you for both though, if you want to be explicit I think this is fine

bznein (approved):
backend untested LGTM with a couple nits

backend/lightning/address.go
Beerosagos: Good point, done! :pray:

backend/lightning/address.go
Beerosagos: removed the first check, but kept the lenght check explicit :pray:

PR #4237, opened 2026-07-10
Fix native building on Fedora and add a target to build only the executable

fabio-porcedda:
This pull request introduces two improvements:
- Fix native builds on Fedora
- Add a make target to build only the executable, without generating the DEB, RPM, or AppImage packages

### DIFF
--- a/frontends/qt/BitBox.pro
+++ b/frontends/qt/BitBox.pro
@@ -19,7 +19,6 @@ TEMPLATE = app
 CFORTIFY = -O2 -D_FORTIFY_SOURCE=2
 CSTACK = -fstack-protector-all -fstack-check
 CMISC = -fwrapv
-CASLR = -fPIE -fPIC
 
 # The following define makes your compiler emit warnings if you use
 # any feature of Qt which has been marked as deprecated (the exact warnings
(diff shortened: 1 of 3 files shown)


frontends/qt/BitBox.pro
```diff
@@ -47,6 +47,9 @@ win32 {
     QMAKE_CXXFLAGS += $$CFORTIFY
     QMAKE_CXXFLAGS += $$CSTACK
     QMAKE_CXXFLAGS += $$CMISC
+    # On Debian/Ubuntu (docker) GCC adds -fPIE by default, but other
+    # toolchains (e.g. Fedora/GCC) do not.
+    QMAKE_CXXFLAGS += $$CASLR
```
benma: Unless I am missing something, `$$CASLR` expands to `-fPIE -fPIC`, but the latter is not needed and turns on PIC mode instead of PIE mode. Please change to `+= -fPIE`, and also remove the otherwise unused `CASLR` definition on line 22.

benma (changes requested):
Thanks a lot. Please see my comment.

frontends/qt/BitBox.pro
fabio-porcedda: I've done the suggested changes.

benma (approved):

PR #4242, opened 2026-07-13
frontend: fix account nr display on send flow

shonsirsha:
Fix account number display on send screens


thisconnect (approved):
thank you LGTM

PR #4245, opened 2026-07-14
backend: schedule update checks

benma-agent:
Update checks were being triggered by frontend navigation, a regression from the previous behavior
of checking only once at launch. The backend now repeats the check daily so users who leave the app
open for a long time still receive update notifications. This is especially useful on Android and
iOS, where apps may keep running in the background.

Move remote update checks out of frontend-mounted handlers. Run one asynchronous check when the
backend starts, then another 24 hours after each attempt. Cache successful results and push them to
subscribed frontend consumers.

- Add a cancellable update-check loop in `backend/update.go`, started from `Backend.Start()` and
  stopped from `Backend.Close()`.
- Make update requests context-aware so shutdown can cancel an in-flight request.
- Protect the cached `*UpdateFile` with a read/write lock.
- On every successful check, including a successful “no update” result, replace the cache
  and emit an `update` event with `action.Replace`.
- On failure, log the error, retain the last successful value, emit no replacement, and wait until
  the next daily check.
- Replace `CheckForUpdateIgnoringErrors()` in the handler-facing backend interface with a
  side-effect-free `GetUpdate()` cache getter. Keep `GET /api/update` and its nullable response
  shape unchanged.
- Add `subscribeUpdate()` to the frontend version API. Convert both the global update banner and
  About version setting from `useLoad(getUpdate)` to `useSync(getUpdate, subscribeUpdate)`, so late
  mounts receive cached state and remain subscribed without initiating a remote check.

- `GET /api/update`: unchanged wire format, but now returns cached state only.
- WebSocket subject `update`: publishes nullable `TUpdateFile` replacement events.
- Backend handler interface: expose `GetUpdate() *UpdateFile`.
- Frontend API: expose `subscribeUpdate(callback)` alongside `getUpdate()`.

- Add backend tests with an injected check function and short interval to verify the initial
  immediate check, daily recurrence, and cancellation.
- Verify successful results are cached and emitted, successful `nil` results clear an older update,
  and failed checks preserve cached state without emitting.
- Add focused frontend tests confirming both consumers render the cached response and react to
  subscription updates.
- Run targeted Go tests with the race detector, `make webtest`, and `make weblint`.

- “Once a day” means 24 hours after the previous attempt finishes, preventing
  overlapping checks.
- Startup checks remain asynchronous so network latency cannot delay application startup.
- Failed checks receive no early retry; the next attempt occurs after the normal 24-hour interval.

### DIFF
--- a/frontends/web/src/components/banners/update.tsx
+++ b/frontends/web/src/components/banners/update.tsx
@@ -2,15 +2,15 @@
 
 import { useTranslation } from 'react-i18next';
 import { runningInAndroid } from '@/utils/env';
-import { getUpdate } from '@/api/version';
+import { getUpdate, subscribeUpdate } from '@/api/version';
 import { Status } from '@/components/status/status';
 import { AppDownloadLink } from '@/components/appdownloadlink/appdownloadlink';
-import { useLoad } from '@/hooks/api';
+import { useSync } from '@/hooks/api';
 import style from './update.module.css';
 
 export const Update = () => {
   const { t } = useTranslation();
-  const file = useLoad(getUpdate);
+  const file = useSync(getUpdate, subscribeUpdate);
   if (!file) {
     return null;
   }
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -270,6 +270,7 @@ type Backend struct {
 	etherScanRateLimiter *rate.Limiter
 	ratesUpdater         *rates.RateUpdater
 	banners              *banners.Banners
+	updateChecker        *updateChecker
 	started              bool
 
 	// For unit tests, called when `backend.checkAccountUsed()` is called.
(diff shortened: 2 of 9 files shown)


frontends/web/src/components/banners/update.tsx
```diff
@@ -2,15 +2,15 @@
...
 import { Status } from '@/components/status/status';
 import { AppDownloadLink } from '@/components/appdownloadlink/appdownloadlink';
-import { useLoad } from '@/hooks/api';
+import { useSync } from '@/hooks/api';
 import style from './update.module.css';
 
 export const Update = () => {
   const { t } = useTranslation();
-  const file = useLoad(getUpdate);
+  const file = useSync(getUpdate, subscribeUpdate);
```
shonsirsha, line 13: `useSync` calls `getUpdate()` then subscribes, right?

Q: Could there be a race here? 

1.) `useSync` starts `getUpdate()` and subscribes to `update` events at the same time. 

2.) If the subscription receives a newer update before `getUpdate()` resolves, the older `getUpdate()` response could overwrite it and hide the available update.

shonsirsha (approved):
1 question

bznein (commented):
backend utACK

bznein (approved):

PR #4247, opened 2026-07-15
update: prevent stale synchronized state

benma-agent:
The initial HTTP response and update events can arrive in either order. Without ordering
information, a stale HTTP response can overwrite a newer event, or a stale event can
overwrite a newer HTTP response.

Add revisions to update snapshots and make the Update consumers retain the newest revision.
This is a quick fix scoped to Update only.

A general solution would need to define ordering for every useSync endpoint and would
require a larger refactor of the shared synchronization API and its backend event contracts.

### DIFF
--- a/backend/update.go
+++ b/backend/update.go
@@ -31,6 +31,7 @@ type updateChecker struct {
 	check updateCheckFunc
 
 	latest     *UpdateFile
+	revision   uint64
 	latestLock locker.Locker
 	cancel     context.CancelFunc
 }
@@ -45,6 +46,12 @@ type UpdateFile struct {
 
 	// Description gives additional information on the release.
 	Description string `json:"description"`
+}
+
+// UpdateState is the revisioned result of the latest successful update check.
+type UpdateState struct {
+	Revision uint64      `json:"revision"`
+	Update   *UpdateFile `json:"update"`
 }
 
 func newUpdateChecker(proxy *socksproxy.SocksProxy) *updateChecker {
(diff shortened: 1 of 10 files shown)


backend/update.go
```diff
@@ -31,6 +31,7 @@ type updateChecker struct {
 	check updateCheckFunc
 
 	latest     *UpdateFile
+	revision   uint64
 	latestLock locker.Locker
```
bznein, line 35: [mini-nit] I guess revision is technically also "locked" by this locker, so maybe rename it to make it clearer? No big deal anyway

bznein (approved):
backend LGTM untested

frontends/web/src/hooks/api.ts
```diff
@@ -82,12 +82,22 @@ export const useLoad = <T>(
...
   const [response, setResponse] = useState<T>();
   const mounted = useMountedRef();
   const onData = (data: T) => {
     if (mounted.current) {
-      setResponse(data);
+      setResponse((current) => {
+        if (
+          current !== undefined
+          && getRevision !== undefined
+          && getRevision(current) >= getRevision(data)
```
thisconnect, line 95: It's not 100% clear to me why handle it in the frontend. Could the backend keep track of "revision" and not send the event to the frontend?

benma: The race is end-to-end. The backend knows in which order it is created (of course, it produces the revision), but not in which order it will arrive in the frontend. 

With something like https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4248, it is ordered and delivered through one channel, so there the revision is not needed at all.

PR #4248, opened 2026-07-15
events: initialize subscriptions with snapshots

benma-agent:
Request initial values through the subscription event stream instead of a separate GET request.
This prevents a delayed GET response from overwriting a newer event.

Serialize snapshot requests with live updates in the backend event queue and convert reload
notifications to complete replacement events before delivery. Remove GET handlers that only fed
subscription initial values.

Potential alternative to https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4248, that removes the ordering problem in general


thisconnect:
I think this is really nice approach and simplifies a lot of things.

From a frontend perspective I'd even propose to drop all get api calls and ONLY use subscriptions, but maybe that is too much for this initial change.

PR #4258, opened 2026-07-21
android: hardcod devservers true for testing

bznein:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/lightning/address_test.go
+++ b/backend/lightning/address_test.go
@@ -339,7 +339,7 @@ func TestAddressAvailability(t *testing.T) {
 	require.Equal(t, &AddressAvailability{
 		GeneratedAddress: GeneratedAddress{
 			Username: "user123",
-			Address:  "user123@bitbox.pay",
+			Address:  "user123@" + lnurlDomainProd,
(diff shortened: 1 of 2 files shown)


backend/bridgecommon/bridgecommon.go
```diff
@@ -350,7 +350,7 @@ func Serve(
 			config.AppDir(),
 			testnet,
 			false,
-			false,
+			true,
```
benma: This is too wide imo - just overwrite the actual prod URL? if this is forgotten and lands in master the blast radius is would be too big 😅

benma (changes requested):

backend/bridgecommon/bridgecommon.go
bznein: makes sense, will push in a bit

benma (approved):
nit: commit msg says android, but it's not only android

bznein:
> nit: commit msg says android, but it's not only android

It's also not devserves anymore :D will amend and push :)

PR #4262, opened 2026-07-21
frontend: fix actionButtons to be more compact

shonsirsha:


frontends/web/src/components/layout/main.module.css
```diff
@@ -9,3 +9,7 @@
     overflow-y: auto;
     position: relative;
 }
+
+.noHorizontalScroll {
+    overflow-x: hidden;
```
thisconnect: what does this fix?

please update commit message.

also would it make sense to always have it `overflow-x: hidden;` without a prop, as we never want noHorizontalScroll

shonsirsha: Answered in DM: dont need it anymroe:)

frontends/web/src/routes/account/components/account-action-buttons.tsx
```diff
@@ -5,15 +5,21 @@ import style from '@/routes/account/account.module.css';
 
 type TProps = {
   children: ReactNode;
+  // compact makes the buttons hug their labels (instead of stretching to equal
+  // widths) and scales the label size and horizontal padding fluidly with the
+  // available width. Used on the Lightning screen.
+  compact?: boolean;
```
shonsirsha: Draft for now. Once tested and looks good, will remove this `compact` prop and use it for all actionbuttons by default.

thisconnect (approved):
tested webdev & Android 14 LGTM

PR #4264, opened 2026-07-22
Transaction list filters

jstrnbrg:


NickeZ:
We want as little business logic as possible in the frontend. So I think the filtering and ordering should be done in the backend.

PR #4266, opened 2026-07-22
lightning: refresh lightning balance across views

strmci:
Lightning balances were fetched only when a view loaded,
so the view could keep showing a stale balance after a
payment was received.

Publish balance notifications from the backend whenever
the balance or its displayed value changes, and consume
them with useSync wherever the frontend displays a
Lightning balance. This preserves the initial load
while keeping the current view updated as the balance
changes.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/api/lightning.ts
+++ b/frontends/web/src/api/lightning.ts
@@ -302,3 +302,9 @@ export const subscribeLightningReady = (cb: TSubscriptionCallback<boolean>): TUn
 export const subscribeListPayments = (cb: TSubscriptionCallback<TLightningPayment[]>) => {
   return subscribeEndpoint('lightning/list-payments', cb);
 };
+
+export const subscribeLightningBalance = (
+  cb: TSubscriptionCallback<TBalance>,
+): TUnsubscribe => {
+  return subscribeEndpoint('lightning/balance', cb);
+};
--- a/frontends/web/src/api/lightning.ts
+++ b/frontends/web/src/api/lightning.ts
@@ -411,3 +411,9 @@ export const subscribeLightningReady = (cb: TSubscriptionCallback<boolean>): TUn
 export const subscribeListPayments = (cb: TSubscriptionCallback<TLightningPayment[]>) => {
   return subscribeEndpoint('lightning/list-payments', cb);
 };
+
+export const subscribeLightningBalance = (
+  cb: TSubscriptionCallback<TLightningBalance>,
+): TUnsubscribe => {
+  return subscribeEndpoint('lightning/balance', cb);
+};
--- a/frontends/web/src/routes/lightning/lightning.tsx
+++ b/frontends/web/src/routes/lightning/lightning.tsx
@@ -9,6 +9,7 @@ import {
   getBlockExplorerTxPrefix,
   getLightningBalance,
   getListPayments,
+  subscribeLightningBalance,
   subscribeListPayments,
   getSparkStatus,
   TSparkStatus,
@@ -25,7 +26,7 @@ import { Status } from '../../components/status/status';
 import { HideAmountsButton } from '../../components/hideamountsbutton/hideamountsbutton';
 import { PaymentDetails } from './components/payment-details';
 import { RatesContext } from '@/contexts/RatesContext';
-import { useLoad } from '@/hooks/api';
+import { useLoad, useSync } from '@/hooks/api';
 import { useMountedRef } from '@/hooks/mount';
 import { useLightning } from '@/hooks/lightning';
 import { TransactionList } from '@/routes/account/components/transaction-list';
@@ -283,7 +284,7 @@ export const Lightning = () => {
   const { t } = useTranslation();
   const { btcUnit } = useContext(RatesContext);
   const { isLightningReady, lightningAccount } = useLightning();
-  const [balance, setBalance] = useState<accountApi.TBalance>();
+  const balance = useSync(getLightningBalance, subscribeLightningBalance);
   const [syncedAddressesCount] = useState<number>();
   const [payments, setPayments] = useState<TLightningPayment[]>();
   const [sparkStatus, setSparkStatus] = useState<TSparkStatus>();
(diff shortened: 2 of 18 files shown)


strmci:
@coderabbit review

frontends/web/src/hooks/lightning.ts
```diff
@@ -1,9 +1,19 @@
...
 export const useLightning = () => {
   const { isLightningReady, lightningAccount } = useContext(LightningContext);
   return { isLightningReady, lightningAccount };
 };
+
+export const useLightningBalance = (): TBalance | undefined => {
+  return useSync<TBalance | undefined>(
+    getLightningBalance,
+    subscribeLightningBalance,
+  );
```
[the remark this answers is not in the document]
thisconnect: Please note that BitBoxApp is not a tranditional web app with a backend on a random server on the internet, but a local app where the backend runs on the client's machine. That said race condition could happen regardless.

However fixing this is a bit out of scope of this PR. Here is a PR that conceptually changes from useSync to only useSubscription https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4248

So please for now keep using useSync, so it will be easier to migrate with https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4248

One small note: useLightningBalance doesn't do anything else than wrapping useSync with getLightningBalance/subscribeLightningBalance .. please just use the useSync directly without the wrapper hook.

frontends/web/src/hooks/api.ts
```diff
@@ -94,7 +94,7 @@ export const useSync = <T>(
   };
   useEffect(
     () => {
-      apiCall().then(onData);
+      apiCall().then(onData).catch(console.error);
       return subscription(onData);
     }, // we pass no dependencies because it's only queried once
     []); // eslint-disable-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
thisconnect: In case of error the backend (which runs on the clients machine will log an error no need to console.error here)

frontends/web/src/api/lightning.ts
```diff
@@ -266,3 +267,23 @@ export const subscribeLightningReady = (cb: TSubscriptionCallback<boolean>): TUn
...
   return subscribeEndpoint('lightning/list-payments', cb);
 };
+
+export const subscribeLightningBalance = (
+  cb: TSubscriptionCallback<TBalance | undefined>,
+): TUnsubscribe => {
+  let latest = 0;
+  return apiSubscribe('lightning/balance', () => {
+    const current = ++latest;
+    getLightningBalance()
```
thisconnect: Looks wrong to call get here, the balance should just be passed along the subscription event. see how other subscriptions work.

strmci:
@thisconnect thanks

I addressed your comments, could you please check again?

frontends/web/src/api/lightning.ts
```diff
@@ -367,3 +367,9 @@ export const subscribeLightningReady = (cb: TSubscriptionCallback<boolean>): TUn
 export const subscribeListPayments = (cb: TSubscriptionCallback<TLightningPayment[]>) => {
   return subscribeEndpoint('lightning/list-payments', cb);
 };
+
+export const subscribeLightningBalance = (
```
shonsirsha, line 415: Nice. Just an improvement, could we switch `LightningSettings` to this new subscription too?

frontends/web/src/routes/lightning/lightning.tsx
```diff
@@ -283,7 +284,7 @@ export const Lightning = () => {
   const { t } = useTranslation();
   const { btcUnit } = useContext(RatesContext);
   const { isLightningReady, lightningAccount } = useLightning();
-  const [balance, setBalance] = useState<accountApi.TBalance>();
+  const balance = useSync(getLightningBalance, subscribeLightningBalance);
```
shonsirsha, line 307: Hm.. `useSync` doesnt return any error. So should we handle `getLightningBalance()` errors here too? 

If the request fails, the useSync hook returns undefined `balance`  while the page keeps waiting for `balance !== undefined`, as far as I can understand it. So it may keep showing skeleton instead of the errors.

bznein (approved):
backend untested LGTM

shonsirsha (changes requested):

frontends/web/src/routes/lightning/lightning.tsx
strmci: yeah, this is a limitation of `useSync`. If we wanted to add error handling, we would have to wrap `getLightningBalance` and `subscribeLightningBalance`. I’m not sure if we want to do that or keep it as is before the `useSync` refactor. please see also https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4266#discussion_r3630721964

cc @thisconnect

frontends/web/src/api/lightning.ts
strmci: updated. The only  effect is that the Close and withdraw funds option becomes enabled when an empty wallet receives a payment.

frontends/web/src/routes/lightning/lightning.tsx
thisconnect: If I read it correctly we didn't show an error before this PR either.

If there is a problem with lightning it would be good to surface a useful error to the user, but IMO can do outside of this PR.. 

but @Beerosagos is a bit more into lighting maybe  we do already have some error?

Beerosagos: I think @thisconnect is right. I'd keep that for a separate PR

frontends/web/src/api/lightning.ts
strmci: I'm checking it again, not sure if it's worth changing it there, wdyt

shonsirsha (approved):

frontends/web/src/api/lightning.ts
shonsirsha: > not sure if it's worth changing it there

IMO would be good for consistency

>  the Close and withdraw funds option becomes enabled when an empty wallet receives a payment.

 That isn't technically wrong.

PR #4267, opened 2026-07-22
Wallet connect update

Tomasvrba:
- [ ] updating the Changelog
- [x] writing unit tests

•  PR Summary: makes WalletConnect signing safer and more reliable:
  - Updates WalletConnect dependencies while keeping Apache-licensed versions.
  - Ensures requests use the network and account the user actually approved.
  - Prevents signing typed data or transactions for a different network.
  - Rejects malformed, unauthorized, or unsupported requests consistently.
  - Prevents duplicate responses and overlapping signing dialogs.
  - Improves transaction sender and destination checks.
  - Fixes WalletConnect pairing-state typos.
  - Adds tests for these cases.

Expected behavior can be tested with Reown's test kit at: https://lab.reown.com/appkit/?name=wagmi

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -5,6 +5,7 @@ package eth
 import (
 	"context"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"math/big"
 	"net/http"
@@ -31,6 +32,7 @@ import (
 	ethereum "github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/rlp"
@@ -880,6 +882,9 @@ func (account *Account) SignTypedMsg(
 	chainId uint64,
 	data string,
 ) (string, error) {
+	if chainId == 0 {
+		return "", errp.New("WalletConnect chain ID must not be zero")
+	}
 	keystore, err := account.Config().ConnectKeystore()
 	if err != nil {
 		return "", err
@@ -923,16 +928,17 @@ func (account *Account) SignETHMessage(message string) (string, string, error) {
 	return account.address.Address.Hex(), "0x" + hex.EncodeToString(signature), nil
 }
 
-// WalletConnectArgs are the tx proposal arguments received from Wallet Connect with Gas, GasPrice,
-// Value and Nonce being optional.
+// WalletConnectArgs are the transaction proposal arguments received from WalletConnect.
 type WalletConnectArgs struct {
-	From     string `json:"from"`
-	To       string `json:"to"`
-	Data     string `json:"data"`
-	Gas      string `json:"gas,omitempty"`
-	GasPrice string `json:"gasPrice,omitempty"`
-	Value    string `json:"value,omitempty"`
-	Nonce    string `json:"nonce,omitempty"`
+	From     string          `json:"from"`
+	To       string          `json:"to"`
+	Data     string          `json:"data"`
+	Input    json.RawMessage `json:"input,omitempty"`
+	Gas      string          `json:"gas,omitempty"`
+	GasPrice string          `json:"gasPrice,omitempty"`
+	Value    string          `json:"value,omitempty"`
+	Nonce    string          `json:"nonce,omitempty"`
+	ChainId  json.RawMessage `json:"chainId,omitempty"`
 }
 
 // EthSignWalletConnectTx signs an Ethereum Tx received from WalletConnect.
@@ -948,6 +954,27 @@ func (account *Account) EthSignWalletConnectTx(
 	var message ethereum.CallMsg
 	var gasPrice *big.Int
 	var value *big.Int
+
+	matches, err := account.MatchesAddress(proposedTx.From)
+	if err != nil {
+		return "", "", err
+	}
+	if !matches {
+		return "", "", errp.New("transaction from address does not match account")
+	}
+
+	if proposedTx.Input != nil {
+		return "", "", errp.New("transaction input field is unsupported; use data")
+	}
+	if proposedTx.ChainId != nil {
+		var proposedChainID hexutil.Uint64
+		if err := proposedChainID.UnmarshalJSON(proposedTx.ChainId); err != nil {
+			return "", "", errp.WithStack(err)
+		}
+		if uint64(proposedChainID) != chainId {
+			return "", "", errp.New("transaction chain ID does not match the WalletConnect request chain")
+		}
+	}
 
 	// Error if chaindId != account.coin.ChainID() (i.e. 1) until L2 RPCs and proper support are added
 	if chainId != account.coin.ChainID() {
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -3,8 +3,10 @@
 package eth
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"math/big"
 	"net/http"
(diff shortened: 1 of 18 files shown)


bznein:
Sorry, I haven't forgotten about this one I swear; I am focused on other tasks but I will try to get to this as soon as I can :)

frontends/web/package.json
```diff
@@ -44,17 +45,23 @@
...
       "react-app/jest"
     ]
   },
   "scripts": {
+    "check-walletconnect-licenses": "node ./scripts/check-walletconnect-licenses.mjs",
     "dev": "npm run start",
     "start": "vite --cors",
     "build": "npm run typescript && vite build",
-    "lint": "npm run typescript && eslint ./src ./tests",
+    "lint": "npm run check-walletconnect-licenses && npm run typescript && eslint ./src ./tests",
```
thisconnect: I think this check could be more generic, and test all our prod deps.

Also it feels wrong to me to do it on lint. Maybe just having a script and then mention in the README to run this after updating frontend deps would be enough?

Tomasvrba: 1 .I don't really see the need to test other deps generically in this way? This is a special case where we specifically care about the version because of the license change.
2. Agree that maybe the lint step is not the right place to do this, but a manual script mentioned in the readme is a worse solution imo. The point is to catch and flag it automatically, not rely on someone remembering to do it

thisconnect: > .I don't really see the need to test other deps generically in this way? 

Well we care about licenses of all our production dependencies not just about the WC ones. A more generic script would check that all prod deps use open source licenses. But doesn't need to be in this PR.

 >  Agree that maybe the lint step is not the right place to do this, but a manual script mentioned in the readme is a worse solution imo. 
 
 ACK. We could use `postinstall` npm task, but that is not great either. How about just having the npm task, and then calling it from a make file task? i.e. during/after `make buildweb` ?

thisconnect: an alternative could be `prebuild` so the lincense check runs always before building and would prevent building an app that has no opensource WC dep?

frontends/web/package.json
```diff
@@ -44,17 +45,23 @@
     "react-select": "5.10.2",
     "request-address": "0.0.11"
   },
+  "overrides": {
+    "viem@2.31.0": {
```
thisconnect, line 49: Is that needed?

I checking out this branch and installing new deps (buildeweb) I can't find any module called viem (`npm list viem`). Can you remove this "overrides" section?

frontends/web/package-lock.json
```diff
@@ -8,9 +8,10 @@
       "name": "bitboxapp",
       "version": "0.0.0",
       "dependencies": {
-        "@reown/walletkit": "1.2.0",
-        "@walletconnect/jsonrpc-types": "1.0.4",
-        "@walletconnect/types": "2.17.2",
+        "@reown/walletkit": "1.2.10",
```
thisconnect, line 11: How did you update the dependency versions?

Reason I am asking is that I still get the old 1.2.0  (see `npm list`) after doing `make buildweb`. I see there are changes in package-lock but there it still says 1.2.0 and not 1.2.10. 🤔

thisconnect: And looking at the check-walletconnect-licenses.mjs file it should allow up to version  1.2.11

frontends/web/scripts/check-walletconnect-licenses.mjs
```diff
@@ -0,0 +1,47 @@
...
+
+  if (policy && isAtLeast(version, policy[1])) {
+    console.error(
+      `WalletConnect Community License version found: ${policy[0]}@${version}`
+    );
+    process.exit(1);
+  }
+}
+
+console.log('WalletConnect dependencies remain on Apache-2.0 versions.');
```
thisconnect: add `process.exit(0);` at the end in case of success.

thisconnect:
also there seems to be a small conflict in app.tsx, please rebase.

frontends/web/package.json
Tomasvrba: that sounds good

frontends/web/package.json
Tomasvrba: Yes, i suggest keeping this
- viem@2.31.0 is present through @walletconnect/utils, and the override fixes two documented `ws` advisories.

frontends/web/package-lock.json
Tomasvrba: > And looking at the check-walletconnect-licenses.mjs file it should allow up to version 1.2.11

Where are you seeing that? It lists 1.2.11 as the first community license version and errors if found

Tomasvrba: > Reason I am asking is that I still get the old 1.2.0 (see npm list) after doing make buildweb. I see there are changes in package-lock but there it still says 1.2.0 and not 1.2.10. 

It says `1.2.10. ` in my lockfile and `make buildweb` installs `1.2.10`, I can't reproduce the issue

frontends/web/package.json
Tomasvrba: split the license check into a separate PR as discussed: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4358

backend/coins/eth/account_test.go
```diff
@@ -513,3 +518,138 @@ func TestSignMsgUsesAccountChainID(t *testing.T) {
...
+
+func TestEthSignWalletConnectTxRejectsInvalidChainID(t *testing.T) {
+	acct := newAccount(t)
+	defer acct.Close()
+
+	tx := validWalletConnectTx(acct)
+	tx.ChainId = json.RawMessage(`null`)
+	_, _, err := acct.EthSignWalletConnectTx(false, acct.ETHCoin().ChainID(), tx)
+	require.EqualError(t, err, "json: cannot unmarshal non-string into Go value of type hexutil.Uint64")
+}
```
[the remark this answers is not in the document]
Tomasvrba: done

backend/coins/eth/account.go
```diff
@@ -949,6 +955,27 @@ func (account *Account) EthSignWalletConnectTx(
...
+	if proposedTx.ChainId != nil {
+		var proposedChainID hexutil.Uint64
+		if err := proposedChainID.UnmarshalJSON(proposedTx.ChainId); err != nil {
+			return "", "", errp.WithStack(err)
+		}
+		if uint64(proposedChainID) != chainId {
+			return "", "", errp.New("transaction chain ID does not match the WalletConnect request chain")
+		}
+	}
+
```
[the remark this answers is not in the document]
Tomasvrba: don

frontends/web/scripts/check-walletconnect-licenses.mjs
```diff
@@ -0,0 +1,47 @@
...
+  return match[0].split('.').map(Number);
+};
+
+const isAtLeast = (version, minimum) => {
+  const actual = versionParts(version);
+  const boundary = versionParts(minimum);
+  const difference = actual.findIndex(
+    (part, index) => part !== boundary[index]
+  );
+  return difference === -1 || actual[difference] > boundary[difference];
```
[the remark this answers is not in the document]
Tomasvrba: done

frontends/web/src/utils/walletconnect-eth-sign-handlers.ts
```diff
@@ -32,187 +55,448 @@ export type TRequestDialogContent = {
...
+
+const getApprovedChains = (namespace: SessionTypes.Namespace) => {
+  if (namespace.chains) {
+    return namespace.chains;
+  }
+  return namespace.accounts.flatMap(account => {
+    const match = /^(eip155:(?:0|[1-9][0-9]*)):0x[0-9a-fA-F]{40}$/.exec(account);
+    return match?.[1] ? [match[1]] : [];
+  });
+};
```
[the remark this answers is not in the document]
Tomasvrba: done

Tomasvrba:
rebased, split license check into https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4358, and addressed rabbit review in fixup commits

frontends/web/package-lock.json
thisconnect: Oups, I'm not sure what happended, I tried multiple times. `npm list @reown/walletkit` looks good resolving.

thisconnect:
LGTM please squash

thisconnect (approved):
LGTM thank you 🙏

bznein (approved):
backend utACK

PR #4268, opened 2026-07-23
android: target API level 36

benma-agent:
Deadline: end of August

Google Play requires app updates to target Android 16 (API level 36) or
higher starting August 31, 2026. Raise compileSdk and targetSdkVersion to
36 and install the Android 36 SDK platform in the development container.

Use Android Gradle Plugin 8.10.1, the first stable AGP line with explicit
API 36 support, with its required Gradle 8.11.1 runtime. These versions
differ because AGP is a plugin and Gradle is the runtime that executes it.

Keep SDK Build Tools 35.0.0 because it is AGP 8.10's supported default.
The SDK platform, compileSdk, and targetSdkVersion determine the Android
API level, so Build Tools 36 is not required and no explicit version pin
needs to be restored.

Android 16's behavior changes were reviewed against the app. It already
uses OnBackPressedDispatcher, handles edge-to-edge insets, declares no
orientation or aspect-ratio restrictions, and produces 16 KB-aligned
native libraries. It does not use the affected scheduling, health,
MediaStore, companion-device, or Android Bluetooth APIs, so no app source
changes or compatibility opt-outs are needed.

References:
https://developer.android.com/about/versions/16/behavior-changes-all
https://developer.android.com/about/versions/16/behavior-changes-16
https://developer.android.com/build/releases/agp-8-10-0-release-notes


Beerosagos (approved):
utACK

benma:
> utACK

@Beerosagos I was hoping you could do a `tACK` by running this in the simulators :smile: Could you do that?

Beerosagos:
> > utACK
> 
> @Beerosagos I was hoping you could do a `tACK` by running this in the simulators 😄 Could you do that?

tACK!

PR #4270, opened 2026-07-23
lightning: prefer connected wallet for top-ups

bznein:
Top-up currently preselects the first funded Bitcoin account, even when that account belongs to a disconnected wallet.

Select the first funded Bitcoin account belonging to the connected keystore, falling back to the first funded account overall. Expose this choice through a backend endpoint and wait for it before rendering the top-up form.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/lightning.go
+++ b/backend/lightning.go
@@ -2,7 +2,14 @@
 
 package backend
 
-import coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
+import (
+	"bytes"
+
+	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts"
+	accountsTypes "github.com/BitBoxSwiss/bitbox-wallet-app/backend/accounts/types"
+	coinpkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend/coins/coin"
+	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
+)
 
 // coinCodeLightning is a portfolio-only pseudo coin code. Use BTC for formatting and conversions;
 // backend.Coin(coinCodeLightning) does not resolve.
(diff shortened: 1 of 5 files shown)


bznein:
@coderabbit review

backend/lightning.go
```diff
@@ -33,6 +40,52 @@ func (backend *Backend) lightningFormattedBalance() (*coinFormattedAmount, error
 	return &formattedBalance, nil
 }
 
+// DefaultLightningTopUpAccountCode returns the first funded or still-syncing Bitcoin account of
```
benma: Why return still-syncing? Seems weird to have different results based on if it is still syncing. Shouldn't it propagate that error?

bznein: I'm not sure I understand what you mean but could you please take a look now? (unfortunately I rebased and amended in the same push so the diff from GH does not help... but it's a self-contained change in that file)

benma:
@bznein 

I asked codex if this was as simple and clear as it can be, and it found that a backend endpoint is not needed in the first place, so it made this PR: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4280. See the PR body, it behaves slightly differently in error cases, but is a lot simpler.

Original codex below:

```

• Findings

  1. P2: Syncing can permanently select the wrong wallet.
     DefaultLightningTopUpAccountCode() (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/6bcf76b6b2ecdd12d1222aa78e7d4600325e0df6/backend/lightning.go#L62-L66) returns ErrSyncInProgress for any encountered
     account. The frontend then converts that error to null
     (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/6bcf76b6b2ecdd12d1222aa78e7d4600325e0df6/frontends/web/src/routes/lightning/topup/topup.tsx#L84-L87) and falls back to the first account. Because this
     useLoad runs only once, it never retries after syncing completes. A disconnected syncing account can therefore reproduce the original bug for the lifetime of the page.

  2. The backend endpoint is unnecessary and makes the change much more complex.
     Every TAccount already has keystore.connected (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/6bcf76b6b2ecdd12d1222aa78e7d4600325e0df6/frontends/web/src/api/account.ts#L34-L40), and topUpAccounts is
     already the funded-account subset. The repository already uses precisely this selection pattern in getFallbackMarketAccountCode
     (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/6bcf76b6b2ecdd12d1222aa78e7d4600325e0df6/frontends/web/src/routes/market/utils.ts#L6-L10).

     The selection can remain entirely in topup.tsx:

     setSourceAccountCode(
       topUpAccounts.find(account => account.keystore.connected)?.code
         || topUpAccounts[0]?.code
         || ''
     );

     That removes the new backend method, handler, API wrapper, duplicate balance queries, error handling, and roughly 170 backend/test lines.
```

bznein:
Closing in favour of https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4280

PR #4272, opened 2026-07-23
lightning: support on-chain payments

bznein:
Recognize Bitcoin addresses in the standard Lightning send flow and use Spark fee quotes to prepare and send payments to them.

Add the on-chain review UI and turn Spark minimum-amount failures into translated errors that show the number of sats required by the address.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -103,6 +103,14 @@ Functions in `src/api/` should stay as thin wrappers around the underlying reque
 should only type the function arguments and the returned promise, and should not add extra client
 side control flow or business logic on top of the request.
 
+### UUID Generation
+Do not call `crypto.randomUUID()` directly in bundled frontend code. Some Qt WebEngine builds may
+not expose it. Use `randomUUID()` from `src/utils/uuid.ts`, which falls back to
+`crypto.getRandomValues()`.
+
+Standalone files under `frontends/web/public/` that are not bundled with the React app must provide
+the same fallback locally.
+
 ### Data Loading Hooks
 Key custom hooks in `src/hooks/` handle all async data:
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -103,6 +103,9 @@ Functions in `src/api/` should stay as thin wrappers around the underlying reque
 should only type the function arguments and the returned promise, and should not add extra client
 side control flow or business logic on top of the request.
 
+Standalone files under `frontends/web/public/` that are not bundled with the React app must provide
+the same fallback locally.
+
 ### Data Loading Hooks
 Key custom hooks in `src/hooks/` handle all async data:
 
--- a/backend/lightning/payments.go
+++ b/backend/lightning/payments.go
@@ -17,15 +17,31 @@ import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/signing"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
 	"github.com/breez/breez-sdk-spark-go/breez_sdk_spark"
+	"github.com/btcsuite/btcd/mempool"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/google/uuid"
 )
 
 const (
 	errPaymentApprovalRequired      errp.ErrorCode = "paymentApprovalRequired"
+	errLightningAmountBelowMinimum  errp.ErrorCode = "lightningAmountBelowMinimum"
 	errLightningInvalidAmount       errp.ErrorCode = "lightningInvalidAmount"
 	errLightningInvalidPaymentInput errp.ErrorCode = "lightningInvalidPaymentInput"
 	errLightningInsufficientFunds   errp.ErrorCode = "lightningInsufficientFunds"
 	errLightningInvoiceAlreadyUsed  errp.ErrorCode = "lightningInvoiceAlreadyUsed"
 )
+
+type lightningAmountBelowMinimumError struct {
+	minAmountSat uint64
+}
+
+func (err *lightningAmountBelowMinimumError) Error() string {
+	return errLightningAmountBelowMinimum.Error()
+}
+
+func (err *lightningAmountBelowMinimumError) Cause() error {
+	return errLightningAmountBelowMinimum
+}
 
 type lightningBolt11Invoice struct {
 	Invoice     string  `json:"invoice"`
@@ -42,10 +58,16 @@ type lightningLNURLPay struct {
 	MaxAmountSat uint64  `json:"maxAmountSat"`
 }
 
+type lightningBitcoinPaymentInput struct {
+	Address   string  `json:"address"`
+	AmountSat *uint64 `json:"amountSat,omitempty"`
+}
+
 type paymentInput struct {
-	Type     string                  `json:"type"`
-	Bolt11   *lightningBolt11Invoice `json:"invoice,omitempty"`
-	LNURLPay *lightningLNURLPay      `json:"lnurlPay,omitempty"`
+	Type           string                        `json:"type"`
+	Bolt11         *lightningBolt11Invoice       `json:"invoice,omitempty"`
+	LNURLPay       *lightningLNURLPay            `json:"lnurlPay,omitempty"`
+	BitcoinAddress *lightningBitcoinPaymentInput `json:"bitcoinAddress,omitempty"`
 }
 
 type bitcoinDepositState string
@@ -76,6 +98,7 @@ type lightningPayment struct {
 	DeductedAmountAtTime coin.FormattedAmountWithConversions `json:"deductedAmountAtTime"`
 	Fee                  coin.FormattedAmountWithConversions `json:"fee"`
 	Invoice              string                              `json:"invoice,omitempty"`
+	TxID                 string                              `json:"txId,omitempty"`
 	BitcoinDeposit       *bitcoinDeposit                     `json:"bitcoinDeposit,omitempty"`
 }
 
@@ -84,9 +107,10 @@ type receivePaymentResponse struct {
 }
 
 type preparePaymentRequest struct {
-	Type         string  `json:"type"`
-	PaymentInput string  `json:"paymentInput"`
-	AmountSat    *uint64 `json:"amountSat,omitempty"`
+	Type           string  `json:"type"`
+	PaymentInput   string  `json:"paymentInput"`
+	AmountSat      *uint64 `json:"amountSat,omitempty"`
+	IdempotencyKey string  `json:"idempotencyKey,omitempty"`
 }
 
 type sendPaymentRequest struct {
@@ -94,12 +118,14 @@ type sendPaymentRequest struct {
 	PaymentInput   string  `json:"paymentInput"`
 	AmountSat      *uint64 `json:"amountSat"`
 	ApprovedFeeSat uint64  `json:"approvedFeeSat"`
+	IdempotencyKey string  `json:"idempotencyKey,omitempty"`
 }
 
 type paymentFee struct {
-	AmountSat     uint64 `json:"amountSat"`
-	FeeSat        uint64 `json:"feeSat"`
-	TotalDebitSat uint64 `json:"totalDebitSat"`
+	AmountSat      uint64 `json:"amountSat"`
+	FeeSat         uint64 `json:"feeSat"`
+	TotalDebitSat  uint64 `json:"totalDebitSat"`
+	IdempotencyKey string `json:"idempotencyKey,omitempty"`
 }
 
 type closeWithdrawQuote struct {
@@ -115,8 +141,9 @@ type closeWithdrawResult struct {
 }
 
 const (
-	paymentInputTypeBolt11   = "bolt11"
-	paymentInputTypeLNURLPay = "lnurlPay"
+	paymentInputTypeBitcoinAddress = "bitcoinAddress"
+	paymentInputTypeBolt11         = "bolt11"
+	paymentInputTypeLNURLPay       = "lnurlPay"
 )
 
 type msatToSatRounding int
@@ -186,6 +213,12 @@ func (lightning *Lightning) ParsePaymentInput(inputStr string) (*paymentInput, e
 	switch inputType := input.(type) {
 	case breez_sdk_spark.InputTypeBitcoinAddress:
 		lightning.log.Printf("Input is Bitcoin address %s", inputType.Field0.Address)
+		return &paymentInput{
+			Type: paymentInputTypeBitcoinAddress,
+			BitcoinAddress: &lightningBitcoinPaymentInput{
+				Address: inputType.Field0.Address,
+			},
+		}, nil
 
 	case breez_sdk_spark.InputTypeBolt11Invoice:
 		amount := "unknown"
@@ -335,6 +368,8 @@ func (lightning *Lightning) toLightningPayment(payment breez_sdk_spark.Payment) 
 		if result.BitcoinDeposit != nil {
 			result.BitcoinDeposit.TxID = details.TxId
 		}
+	case breez_sdk_spark.PaymentDetailsWithdraw:
+		result.TxID = details.TxId
 	}
 
 	return result
@@ -529,11 +564,15 @@ func checkApprovedPaymentFee(fee uint64, approvedFee uint64) error {
 	return nil
 }
 
-func checkPaymentBalance(fee *paymentFee, availableBalance coin.Amount) error {
-	if new(big.Int).SetUint64(fee.TotalDebitSat).Cmp(availableBalance.BigInt()) > 0 {
+func checkAvailableBalance(amountSat uint64, availableBalance coin.Amount) error {
+	if new(big.Int).SetUint64(amountSat).Cmp(availableBalance.BigInt()) > 0 {
 		return errLightningInsufficientFunds
 	}
 	return nil
+}
+
+func checkPaymentBalance(fee *paymentFee, availableBalance coin.Amount) error {
+	return checkAvailableBalance(fee.TotalDebitSat, availableBalance)
 }
 
 func lightningPaymentError(err error) error {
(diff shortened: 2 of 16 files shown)


bznein:
@coderabbit review

frontends/web/src/routes/lightning/send/components/payment-input-details.tsx
```diff
@@ -89,6 +90,26 @@ type TLNURLPayRecipientDetailsProps = {
...
+      <h2 className={`${styles.label || ''} ${styles.receiverLabel || ''}`}>
+        {t('lightning.send.confirm.receiver')}
+        <Badge className={styles.receiverBadge} type="info">
+          {t('lightning.send.onChain')}
+        </Badge>
+      </h2>
+      <div className={styles.address}>{bitcoinAddress.address}</div>
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
bznein: fixed

frontends/web/src/routes/lightning/send/send.module.css
```diff
@@ -20,6 +20,20 @@
...
 
+.receiverLabel {
+    align-items: center;
+    display: flex;
+    gap: var(--space-eight);
+}
+
+.receiverBadge {
+    padding: 3px 6px;
+}
```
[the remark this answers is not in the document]
bznein: fixed

bznein:
@coderabbit review

backend/lightning/payments.go
```diff
@@ -532,6 +575,19 @@ func lightningPaymentError(err error) error {
...
+		const suffix = " sats required for this address"
+		if minimum, found := strings.CutPrefix(invalidInput.Field0, prefix); found {
+			if minimum, found = strings.CutSuffix(minimum, suffix); found {
+				minAmountSat, parseErr := strconv.ParseUint(minimum, 10, 64)
+				if parseErr == nil {
+					return &lightningAmountBelowMinimumError{minAmountSat: minAmountSat}
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
bznein: fixed

frontends/web/src/routes/lightning/lightning.tsx
```diff
@@ -150,7 +150,7 @@ const paymentToTransaction = (
       : undefined,
     time: payment.time,
     type: payment.type,
-    txID: payment.id,
+    txID: payment.txId || payment.id,
```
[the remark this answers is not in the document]
bznein: fixed

benma (changes requested):
---
• Findings

  1. High: on-chain retries can send the payment twice.
     sendBitcoinPayment (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/bf45a62c6ca0ced0a8bbb04de51491594dea389d/backend/lightning/payments.go#L757-L763) leaves IdempotencyKey unset, despite the SDK
     contract (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/bf45a62c6ca0ced0a8bbb04de51491594dea389d/vendor/github.com/breez/breez-sdk-spark-go/breez_sdk_spark/breez_sdk_spark.go#L18546-L18554)
     explicitly supporting deduplication. If the withdrawal succeeds but its response is lost, the frontend presents Send again; retrying creates another withdrawal. Generate one UUID per reviewed payment, carry
     it through sendPaymentRequest, reuse it across retries, and pass it to the SDK.

  2. Medium: hybrid BIP21 requests are always downgraded to on-chain.
     ParsePaymentInput (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/bf45a62c6ca0ced0a8bbb04de51491594dea389d/backend/lightning/payments.go#L217-L229) only looks for InputTypeBitcoinAddress. Breez places
     the fallback address before query-provided methods, so bitcoin:<address>?lightning=<bolt11> always selects the address and ignores the supported Lightning invoice. Conversely, bitcoin:?lightning=<bolt11> is
     rejected. Prefer supported Lightning methods before the on-chain fallback, or explicitly let the user choose. Add mixed-method and Lightning-only BIP21 cases.

frontends/web/src/routes/lightning/send/hooks/use-payment-review.ts
```diff
@@ -35,34 +38,38 @@ const isValidAmount = (paymentDetails: TPaymentReviewDetails, amount?: number):
...
 }: TUsePaymentReviewProps) => {
   const { t } = useTranslation();
   const fixedAmountSat = paymentDetails.type === TPaymentInputType.BOLT11
+    || paymentDetails.type === TPaymentInputType.BITCOIN_ADDRESS
     ? paymentDetails.details.amountSat
     : undefined;
   const needsCustomAmount = fixedAmountSat === undefined;
   const mounted = useMountedRef();
   const customAmountRef = useRef<number>();
+  const bitcoinPaymentIdempotencyKey = useRef(newIdempotencyKey());
```
shonsirsha: Is it safe to call `crypto.randomUUID()` directly here?

Some QT builds may not have it? Asking because we have some polyfill and there it's mentioned that some QT web engine builds don't have it. 

see `bitrefill.html`, `coin-to-fiat.html`, `fiat-to-coin.html`

bznein: Good question, I'll verify :)

bznein: That was a good point, I added the same fallback as in other places. I will also make sure to update AGENTS.md so that the agent doesn't make the same mistake in the future

benma: Does this belong in the frontend or can it easily be moved to the backend? As ususal, we don't want business logic in the frontend, frontend should concern itself with presentation only if possible.

bznein: That's a good point; I moved it to the backend (only for this occurrence, not for other places where we use crypto.randomUUID)

Beerosagos: Maybe that's another good addition for agents.md: "keep logic in backend"

bznein:
@benma @shonsirsha ping :)

shonsirsha (approved):
F.E LGTM

AGENTS.md
```diff
@@ -103,6 +103,14 @@ Functions in `src/api/` should stay as thin wrappers around the underlying reque
 should only type the function arguments and the returned promise, and should not add extra client
 side control flow or business logic on top of the request.
 
+### UUID Generation
+Do not call `crypto.randomUUID()` directly in bundled frontend code. Some Qt WebEngine builds may
+not expose it. Use `randomUUID()` from `src/utils/uuid.ts`, which falls back to
```
benma: Doesn't exist anymore - remove whole section?

benma (changes requested):

AGENTS.md
bznein: yeah good point

bznein: done

frontends/web/src/api/lightning.ts
```diff
@@ -279,8 +301,12 @@ export const postPreparePayment = async (data: TPreparePaymentRequest): Promise<
   );
 };
 
-export const postSendPayment = async (data: TSendPaymentRequest): Promise<void> => {
-  return postApiResponse<void, TSendPaymentRequest>('lightning/send-payment', data, 'Error calling postSendPayment');
+export const postSendPayment = async (data: TSendPaymentRequest): Promise<TLightningPayment> => {
```
Beerosagos: Why adding the TLightningPayment return? It doesn't seem used in the FE and adds quite a few changes in the BE

backend/lightning/payments.go
```diff
@@ -42,10 +57,16 @@ type lightningLNURLPay struct {
 	MaxAmountSat uint64  `json:"maxAmountSat"`
 }
 
+type lightningBitcoinAddress struct {
```
Beerosagos: nit: maybe rename to `lightningBitcoinPaymentInput` or similar, since it's not an address? We could also consider removing the `lightning` prefix from all of these

backend/lightning/payments.go
```diff
@@ -185,25 +210,35 @@ func (lightning *Lightning) ParsePaymentInput(inputStr string) (*paymentInput, e
...
-				Invoice:     inputType.Field0.Invoice.Bolt11,
-				Description: inputType.Field0.Description,
-				AmountSat:   amountSat,
+			Type: paymentInputTypeBitcoinAddress,
+			BitcoinAddress: &lightningBitcoinAddress{
+				Address: inputType.Field0.Address,
 			},
 		}, nil
 
+	case breez_sdk_spark.InputTypeBip21:
```
Beerosagos: Are we sure we want to support bip21 input as well, here?

backend/lightning/payments.go
```diff
@@ -525,13 +581,60 @@ func checkPaymentBalance(fee *paymentFee, availableBalance coin.Amount) error {
 	return nil
 }
 
+// Breez SDK v0.13.1 emits on-chain minimum-amount failures as SdkErrorInvalidInput
```
Beerosagos: The min send amount handling looks a bit brittle to me: it adds quite a lot of code and depends heavily on the sdk error text. 
 
Codex says that it should be quite easy to compute and doesn't depend on the current fee rates:
```
  It depends on:                                                                                                                                                                                                                              

  - The destination’s scriptPubKey type and serialized size.                                                      
  - The estimated size of a future input spending that output.
  - The dust relay fee, conventionally Bitcoin Core’s default of 3 sat/vByte.

  Approximately:                                                                                                       
                                                           
  minimum = (output size + estimated spending-input size) × 3 sat/vByte                                                                                                                                                                       
                                                                                                                                                                                                                                              
  Examples:
                                                                                                                                                                                                                                              
  - P2WPKH: (31 + 67) × 3 = 294 sats
  - P2TR: (43 + 67) × 3 = 330 sats
  - P2SH: (32 + 148) × 3 = 540 sats                                                                                                                                                                                                           
  - P2PKH: (34 + 148) × 3 = 546 sats
```

I guess we could check it directly and avoid having to deal with the sdk error. Maybe we could even just set a lower bitcoin send threshold of 550/600 sats. wdyt?

Beerosagos (changes requested):

backend/lightning/payments.go
bznein: I don't think it's "that" brittle since the error string is vendored so it shouldn't break unless we manually bump the version of the sdk (and the test would fail). I'm ok with computing directly anyway! I will push a commit soon

> Maybe we could even just set a lower bitcoin send threshold of 550/600 sats. wdyt?

I think this is a product decision that needs a bit more consideration; ok to leave it to a separate PR/issue?

backend/lightning/payments.go
bznein: Good point; I'll do it, however I'd rather not change all prefixes on this PR, wdyt?

backend/lightning/payments.go
bznein: Yes, I think we should. Without this case standard on-chain payment URIs/QR codes would be rejected while bare addresses worked. I think that would be an unpleasant UX

frontends/web/src/api/lightning.ts
bznein: absolutely that's a leftover from some previous iteration

backend/lightning/payments.go
Beerosagos: Not sure how standard it is to support this kind of payment input, bth. Anyway, if we want to offer this, I think it should be slightly more complete, e.g. not discarding label/message and using it to precompile the payment form. wdyt about doing it in a separate PR? Looks a bit OT here imo.

backend/lightning/payments.go
Beerosagos: Good point about vendoring :+1: Anyway, if you are ok with computing it, then we don't need to set up a generic lower threshold. I suggested that only to simplify :)

backend/lightning/payments.go
bznein: Yeah that makes sense. I will remove it from here and we can put it back in a separate PR

bznein: pushed

Beerosagos: Thanks :pray:

backend/lightning/payments.go
```diff
@@ -734,6 +823,33 @@ func (lightning *Lightning) sendLNURLPay(request sendPaymentRequest) error {
 	return nil
 }
 
+type bitcoinAddressToPkScripter interface {
+	AddressToPkScript(string) ([]byte, error)
+}
+
+func (lightning *Lightning) minimumNonDustAmountSat(destinationAddress string) (uint64, error) {
```
Beerosagos: nit: this is very nice! But maybe we can merge iy with validateBitcoinPaymentAmount

backend/lightning/payments.go
```diff
@@ -745,6 +861,9 @@ func (lightning *Lightning) prepareBitcoinPayment(
 	if amountSat == 0 {
 		return breez_sdk_spark.PrepareSendPaymentResponse{}, nil, errLightningInvalidAmount
 	}
+	if err := lightning.validateBitcoinPaymentAmount(destinationAddress, amountSat); err != nil {
```
Beerosagos, line 902: Here we should check the fee policy before passing the amount to be validated: if the policy is to subtract the fees from the amount, the validation might incorrectly pass.

Also, I feel like validating against dust amounts in the prepare phase (i.e. while the user is still typing) could be a bit aggressive. In the normal onchain send flow we don't even explicitly check for that and just show a (quite generic, honestly) error after confirming the TX on the bitbox.

I'd suggest to move this check in the send phase. Not sure if the amount in the send request takes into account the fee policy already.

frontends/web/src/locales/en/app.json
```diff
@@ -922,6 +921,7 @@
     "lightningAddressChangeCooldown": "Please wait before changing your Lightning address again.",
     "lightningAddressInvalidUsername": "Use only lowercase letters and numbers.",
     "lightningAddressUsernameUnavailable": "This name is not available.",
+    "lightningAmountBelowMinimum": "Enter at least {{minAmountSat}} sats for this address.",
```
Beerosagos: nit: I'd suggest something like: `The amount must be at least {{minAmountSat}}`

frontends/web/src/api/lightning-errors.ts
```diff
@@ -25,12 +27,18 @@ const lightningErrorTranslationKeys: Partial<Record<string, string>> = {
   [TLightningErrorCode.ADDRESS_USERNAME_UNAVAILABLE]: 'error.lightningAddressUsernameUnavailable',
 };
 
+export type TLightningErrorData = {
```
Beerosagos: nit: I don't like much the specific minAmountSat field in a generic structure like this. 
Codex suggests smth like 

```suggestion
  type TLightningErrorDataByCode = {                                                                                                                                                                                                          
    [TLightningErrorCode.AMOUNT_BELOW_MINIMUM]: {          
      minAmountSat: number;                                                                                                                                                                                                                   
    };                                                                                                                 
  }; 
```

But I'm not sure if that's much better, actually.

Beerosagos: Another option could be to add the minAmount in the prepare response, so that we don't have to complicate TSdkError

backend/lightning/payments_test.go
```diff
@@ -637,6 +664,43 @@ func TestToLightningLNURLPay(t *testing.T) {
...
+	input, err := lightning.ParsePaymentInput("bc1qraw")
+
+	require.NoError(t, err)
+	require.Equal(t, paymentInputTypeBitcoinAddress, input.Type)
+	require.NotNil(t, input.BitcoinAddress)
+	require.Equal(t, "bc1qraw", input.BitcoinAddress.Address)
+	require.Nil(t, input.BitcoinAddress.AmountSat)
+}
+
+func TestParseBIP21PaymentInputNotSupported(t *testing.T) {
```
Beerosagos: nit: I think we can remove it. We'll add bip21 support soon anyway

Beerosagos (changes requested):

backend/lightning/payments.go
Beerosagos, line 546: Not strictly related to this PR: but please add a comment here explaining that we automatically pick the fastest fee because we can't RBF (it would require the Spark Entity to co-sign and that's not supported by the SDK, afaik) and we don't want users' TXs to get stuck in the mempool.

Beerosagos (changes requested):
Testing the new code, I realized that the prepare SDK call already fails on dust amounts 😓 so my suggestion to move the dust check in the send phase doesn't work here :( 

Also, new Codex finding which looks pretty edgy but probably worth fixing anyway: we should reuse the idempotency key when re-approving a payment after a fee spike.

```
  1. The first SDK send succeeds, but the app receives an ambiguous error—for example, the HTTP/bridge response is lost or the SDK times out after submitting it.                                                                             
  2. The UI still has the original quote and idempotency key A.
  3. The user retries.
  4. The backend re-prepares before every send. If the fee has increased, it returns paymentApprovalRequired without calling SendPayment with key A.
  5. The frontend obtains a refreshed quote, which currently receives a new key B.
  6. The user approves it, and the SDK treats key B as a new payment.

  Because an on-chain address is reusable, there is no invoice/payment-hash uniqueness to prevent that second transaction. If sufficient balance remains, two payments can be made.

  It requires both an ambiguous first result and a fee increase before retrying, so it is probably rare. If the fee stays equal or decreases, the retry uses key A and the SDK deduplicates it correctly. But payment systems generally treat
  even rare duplicate-send paths seriously, especially since preserving the key across fee refreshes is a small change. 
```

backend/lightning/payments.go
bznein: added

backend/lightning/payments.go
```diff
@@ -745,6 +851,41 @@ func (lightning *Lightning) sendLNURLPay(request sendPaymentRequest) error {
 	return nil
 }
 
+type bitcoinAddressToPkScripter interface {
+	AddressToPkScript(string) ([]byte, error)
+}
+
+func (lightning *Lightning) validateBitcoinPaymentAmount(
```
Beerosagos, line 858: nit: please rename to smth that mentions the check against the dust limit, since this is not a general validation

backend/lightning/payments.go
```diff
@@ -756,6 +897,18 @@ func (lightning *Lightning) prepareBitcoinPayment(
 	if amountSat == 0 {
 		return breez_sdk_spark.PrepareSendPaymentResponse{}, nil, errLightningInvalidAmount
 	}
+	// The SDK rejects dust outputs during preparation with a generic error. Validate first so
+	// callers receive the actionable minimum amount instead.
```
Beerosagos, line 901: nit: worth mentioning that with the included fees policy, we'll still return a generic error since the exact amount depends on the prepare result, which will fail instead. Not much that we can do about it.

Beerosagos (approved):
tACK 👍 🙏 

Just two last nits

bznein:
Superseeded by https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4324

PR #4280, opened 2026-07-28
lightning: prefer connected wallet for top-ups

benma-agent:
Top-up already filters the available Bitcoin accounts to those with a
balance. Prefer the first such account from the connected wallet, falling
back to the first funded account.

Codex note comparing this to the original PR:

---
The subagent found that the simplified implementation matches PR #4270 in the normal synced, error-free case.

It is not identical around balance errors:

  - Our version preserves existing getTopUpAccounts behavior, which treats balance errors as eligible. It could therefore prefer a connected account whose funding status is unknown.
  - PR #4270 skips non-sync balance errors.
  - Conversely, PR #4270 mishandles sync errors: the endpoint fails, the frontend silently falls back to the first account, and never retries. Our version behaves better here.


bznein (approved):
Yeah that's definitely so much better :)

Thank you!
LGTM!

PR #4281, opened 2026-07-28
backend: restrict export file permissions

benma-agent:
Create notes, transaction, and log exports with private file permissions. Tighten existing destinations before writing and document the policy.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,6 +158,8 @@ component state — no form library is used.
 
 ### Go
 - Must pass `gofmt` and `goimports`. Linting via `golangci-lint run` (config in `.golangci.yml`).
+- Create files containing sensitive data with owner-only (`0600`) permissions, and tighten the
+  permissions of existing destination files when overwriting them.
 
 ### TypeScript / React
 - ESLint config in `eslint.config.js`.
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -1299,7 +1299,7 @@ func (backend *Backend) ExportLogs() error {
 
 	exportFile, err := os.OpenFile(
 		path,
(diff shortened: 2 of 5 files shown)


bznein (changes requested):
Codex finding:
  
• The normal owner-owned export case is improved, but the new permission check
  introduces a destructive failure path by truncating existing destinations before
  confirming the file can be made private.

  Review comment:

  - [P2] Check permissions before truncating exports — /home/bznein-coding-agent/agent-
    work/bitbox-wallet-app/backend/backend.go:1300-1302
    When the selected destination is an existing file that this process can write but
    cannot chmod, such as a group-writable file owned by another user or a chmod-
    restricted mount, O_TRUNC erases it before EnsurePrivateFile returns an error; the
    same open-then-check pattern is used for notes and CSV exports. Tighten or validate
    the existing path before truncating, or write to a private temp file and replace it
    only after permissions are secured.

benma:
> Codex finding:
> 
> • The normal owner-owned export case is improved, but the new permission check introduces a destructive failure path by truncating existing destinations before confirming the file can be made private.
> 
> Review comment:
> 
>     * [P2] Check permissions before truncating exports — /home/bznein-coding-agent/agent-
>       work/bitbox-wallet-app/backend/backend.go:1300-1302
>       When the selected destination is an existing file that this process can write but
>       cannot chmod, such as a group-writable file owned by another user or a chmod-
>       restricted mount, O_TRUNC erases it before EnsurePrivateFile returns an error; the
>       same open-then-check pattern is used for notes and CSV exports. Tighten or validate
>       the existing path before truncating, or write to a private temp file and replace it
>       only after permissions are secured.

Fixed in 2nd commit.

bznein (approved):
utACK

PR #4283, opened 2026-07-29
lightning: add default top-up transaction note

bznein:
Pre-fill the top-up transaction note with "Lightning top-up" so the outgoing Bitcoin transaction is labeled before it is sent.

The note remains editable in the existing send note field, and the top-up flow continues to send it through the normal Bitcoin account send path.

- [ ] updating the Changelog
- [ ] writing unit tests


bznein:
@jadzeidan note that this is slightly different from swap. There, we do not give the option to set the note while creating the tx; in this case we do so we simply pre-fill the field:

image

jadzeidan:
@bznein looks good to me.

Beerosagos (approved):
tACK 👍

PR #4287, opened 2026-07-30
lightning: add account selector in send

bznein:
Just quickly put together by codex, code not reviewed yet. Based on  #4272 so it also contains its commit

https://github.com/user-attachments/assets/397f11ea-ee19-4cab-885a-6a8423514bdf

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Lightning: select a connected Bitcoin account as the send recipient
 - Bitcoin: show account details for the persisted receive address type by default
 - Add external block explorer links to used addresses
 - Settings search
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -103,6 +103,14 @@ Functions in `src/api/` should stay as thin wrappers around the underlying reque
 should only type the function arguments and the returned promise, and should not add extra client
 side control flow or business logic on top of the request.
 
+### UUID Generation
+Do not call `crypto.randomUUID()` directly in bundled frontend code. Some Qt WebEngine builds may
+not expose it. Use `randomUUID()` from `src/utils/uuid.ts`, which falls back to
+`crypto.getRandomValues()`.
+
+Standalone files under `frontends/web/public/` that are not bundled with the React app must provide
+the same fallback locally.
+
 ### Data Loading Hooks
 Key custom hooks in `src/hooks/` handle all async data:
 
--- a/backend/lightning/handlers.go
+++ b/backend/lightning/handlers.go
@@ -4,6 +4,7 @@ package lightning
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"strconv"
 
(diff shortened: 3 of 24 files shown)


jadzeidan:
@bznein after testing I think we should just make it like on regular send page (drop down icon in invoice/address field).  Don't think it would be an issue since it already works like that on send, and just pasting in your BTC address (instead of selecting the account) still works anyways.

bznein:
@jadzeidan what commit did you test? admittedly I haven't updated the video in the description, but it is now like the regular send:


https://github.com/user-attachments/assets/ba9d71cd-30cc-4811-a501-707c86e561cd

jadzeidan:
@bznein I meant I tested how on-chain looks to compare. Looks good now!

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Lightning: select a connected Bitcoin account as the send recipient
```
Beerosagos: I don't think we need a changelog about this, since LN is not release yet.

Beerosagos (changes requested):
I noticed a few small bugs/nits:

1. The input text caption gets cut if the account selector is enabled. Maybe remove 'here'? The label has a left padding, which I think is where the onchain send flow shows the small QR icon. Maybe we can hide that too..

Image

2. After picking an account, I don't seem to be able to cancel/modify the input field to e.g. paste an invoice instead: I need to go back to the account view and start the send flow again. In the onchain send flow we have a small 'reset' button that clears the input field.

3. Remembered accounts are not shown in the account selector. Which brings me to a question: should we verify the address on the bitbox before proceeding? If so, imo it's fine not to show remembered accounts (even if the flow in this case differs from the onchain send flow, which shows the addresses and asks the user to connect the keystore to proceed after the account selection). If we are fine with proceeding without verifying the address on the bb, then it would make sense to allow remembered addresses as well, right?

CHANGELOG.md
bznein: whoops my bad, codex added it and it slipped my review

bznein:
> Member



> I noticed a few small bugs/nits:
> 
> 1. The input text caption gets cut if the account selector is enabled. Maybe remove 'here'? The label has a left padding, which I think is where the onchain send flow shows the small QR icon. Maybe we can hide that too..
> 
> Image
> 2. After picking an account, I don't seem to be able to cancel/modify the input field to e.g. paste an invoice instead: I need to go back to the account view and start the send flow again. In the onchain send flow we have a small 'reset' button that clears the input field.
> 3. Remembered accounts are not shown in the account selector. Which brings me to a question: should we verify the address on the bitbox before proceeding? If so, imo it's fine not to show remembered accounts (even if the flow in this case differs from the onchain send flow, which shows the addresses and asks the user to connect the keystore to proceed after the account selection). If we are fine with proceeding without verifying the address on the bb, then it would make sense to allow remembered addresses as well, right?

Addressed 1 and 2.

For 3, I don't have a strong opinion. I think we can show remembered wallets anyway, even if requiring verification. So we just need to decide if we want that :)

Beerosagos:
> For 3, I don't have a strong opinion. I think we can show remembered wallets anyway, even if requiring verification. So we just need to decide if we want that :)

I think we can go without address verification (an evil app would skip that anyway, so no additional security). Given that, listing remembered accounts as well would be nice

Beerosagos (approved):
Didn't check the code, but tested and LGTM

Beerosagos:
@shonsirsha could you review the code? :pray:

shonsirsha (approved):

PR #4299, opened 2026-08-04
frontend: bind messages to vendor iframes

benma-agent:
Validate postMessage sources against the active iframe and retain the
validated source and origin for asynchronous replies.

Apply the same protection to Pocket, Bitsurance, BTC Direct, and Bitrefill,
and add regression tests for source binding and exact reply origins.


Beerosagos (approved):
tACK, tested pocket and bitrefill integrations

PR #4300, opened 2026-08-04
swap: support large EVM payment request amounts

bznein:
SwapKit returns token amounts in the asset's smallest unit. Values for 18-decimal tokens can exceed uint64 and previously failed while decoding the signed payment request.

Preserve output amounts as decimal strings across JSON and frontend round trips. Parse them as arbitrary-precision integers for swap formatting and payment-request verification. Keep the legacy uint64 protobuf amount for BTC-family signing and reject values that do not fit.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/devices/bitbox02/keystore.go
+++ b/backend/devices/bitbox02/keystore.go
@@ -340,6 +340,9 @@ func (keystore *keystore) signBTCTransaction(btcProposedTx *btc.ProposedTransact
 	paymentRequest := btcProposedTx.TXProposal.PaymentRequest
 	var paymentRequestIndex *uint32
 	if paymentRequest != nil {
+		if paymentRequest.TotalAmount == nil || !paymentRequest.TotalAmount.IsUint64() {
+			return errp.New("BTC payment request amount overflows uint64")
+		}
 		btcPaymentRequests = []*messages.BTCPaymentRequestRequest{
 			newBTCPaymentRequest(paymentRequest),
 		}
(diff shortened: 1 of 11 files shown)


frontends/web/src/api/account.ts
```diff
@@ -321,7 +321,7 @@ export type Slip24 = {
     };
   }>;
   outputs: Array<{
-    amount: number;
+    amount: number | string;
```
thisconnect, line 325: Ideally this was always a string, also in slip24

thisconnect (approved):
tested frontend LGTM

frontends/web/src/api/account.ts
Beerosagos: We need `number` for compatibility with the `request-address` library inside Pocket integration. I guess we could remove it by using a function that converts the `request-address` slip24 into our own 🤷‍♂️ not sure what would be the best.

backend/devices/bitbox02/keystore.go
```diff
@@ -484,7 +456,10 @@ func (keystore *keystore) signETHTransaction(txProposal *eth.TxProposal) error {
 	return nil
 }
 
-func newBTCPaymentRequest(txPaymentRequest *paymentrequest.Request) *messages.BTCPaymentRequestRequest {
+func newBTCPaymentRequest(
+	txPaymentRequest *paymentrequest.Request,
+	totalAmount uint64,
+) *messages.BTCPaymentRequestRequest {
```
Beerosagos: Nit: I think you could use something like this and avoid the additional parameter

```
  totalAmount := uint64(0)
  if amount := txPaymentRequest.TotalAmount; amount != nil && amount.IsUint64() {
      totalAmount = amount.Uint64()
  }

  return &messages.BTCPaymentRequestRequest{
      // existing fields...
      TotalAmount: totalAmount,
  }
```

Beerosagos (approved):
utACK with a small nit

PR #4301, opened 2026-08-04
Merge master staging spark

Beerosagos:
Resolve the four overlapping changes while preserving the behavior from
both branches.

frontends/qt/Makefile keeps the linux-bin split from master and the
Spark library layout from staging. The server and Breez libraries remain
in linux-tmp/lib before Linux packaging runs.

frontends/web/src/app.tsx keeps the iframe-aware AppFrame from master.
It passes Lightning account state into the shared navigation visibility
logic and into BottomNavigation.

frontends/web/src/components/bottom-navigation/bottom-navigation.tsx
keeps the floating glass bar, sliding indicator and Android keyboard
handling. It also preserves Lightning account counting, direct Lightning
routing and the Lightning-only marketplace exclusion.

frontends/web/src/components/bottom-navigation/utils.ts combines typed
tab indexes and the visibility helper with Lightning route
classification. A Lightning account can show navigation while
bootloader devices still suppress it.

Update the navigation utility tests for the merged visibility signature
and cover the Lightning-only case.


shonsirsha (approved):
Frontend LGTM

PR #4304, opened 2026-08-05
gate lightning on desktop

shonsirsha:
When unavailable (not mobile / not webdev):

- Settings: hides the Lightning settings option.

- Settings search: removes Lightning results.

- Direct links are blocked: `/lightning/*` goes to home,  and `/settings/lightning-settings`
 goes back to Advanced Settings.
 
- App/Lightning context: ignores Lightning account state. 
This means sidebar, bottom nav, and account lists do not show Lightning.

- Portfolio summary: filters out the Lightning balance row.

### DIFF
--- a/cmd/servewallet/main.go
+++ b/cmd/servewallet/main.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 
 	backendPkg "github.com/BitBoxSwiss/bitbox-wallet-app/backend"
 	"github.com/BitBoxSwiss/bitbox-wallet-app/backend/arguments"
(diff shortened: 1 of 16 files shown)


cmd/servewallet/main.go
```diff
@@ -104,22 +116,60 @@ func (webdevEnvironment) OnAuthSettingChanged(enabled bool) {
...
-	return nil
+func (env webdevEnvironment) DeleteLightningEncryptionKey(accountCode string) error {
+	env.lightningEncryptionKeysMu.Lock()
+	defer env.lightningEncryptionKeysMu.Unlock()
+
+	keys, err := env.loadLightningEncryptionKeys()
+	if err != nil {
+		return err
+	}
+	delete(keys, accountCode)
```
bznein, line 160: [nit] do we want to consider an error the case in which we are trying to delete smth that is not present?

`delete` is a no-op in that case, so no big deal, but maybe worth treating that as an error case?

bznein (approved):
backend LGTM with a small comment

cmd/servewallet/main.go
Beerosagos: If we return an error here, it will be treated by the caller as if we could not delete the keys. I don't think this is worth changing the caller code to tell apart the two cases (keys exist but can't be deleted vs missing keys)

PR #4305, opened 2026-08-05
frontend: warn before restoring an external wallet

cedwies:
This adds an info box to the "restore from recovery words" workflow, to inform users that an imported seed which was not created on the BitBox might not be secure.

Open points: 

- I used `Message type="info"` instead of `Message type="warning"`. I find the warning box too alerting.
- No regression test was added. IMO the state transition is simple enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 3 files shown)


thisconnect (approved):
untested LGTM, thanks

PR #4307, opened 2026-08-06
lightning: recover stuck top-ups

bznein:
Let users recover an unclaimed Bitcoin top-up from its payment details. They can review Spark's claim fee, claim the deposit into Lightning, or refund it to a selected Bitcoin account using the recommended fee rate.

Add backend quote and recovery endpoints with fee approval checks, typed errors and refunded-deposit filtering. Wire the flow through targeted routing, confirmation, success and failure states, including transaction links and actionable SDK error details.

- [ ] updating the Changelog
- [ ] writing unit tests


backend/lightning/topup_recovery.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	maxFee := breez_sdk_spark.MaxFee(breez_sdk_spark.MaxFeeFixed{Amount: feeSat})
+	response, err := lightning.sdkService.ClaimDeposit(breez_sdk_spark.ClaimDepositRequest{
+		Txid:   deposit.Txid,
+		Vout:   deposit.Vout,
+		MaxFee: &maxFee,
+	})
+	if err != nil {
+		lightning.log.WithError(err).Error("Claim Bitcoin deposit failed")
+		return nil, err
```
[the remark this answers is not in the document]
bznein: fixed

frontends/web/src/api/lightning.ts
```diff
@@ -44,7 +44,9 @@ export type TBitcoinDeposit = {
   txid: string;
   vout: number;
   state: TBitcoinDepositState;
-  claimError?: string;
+  claimFee?: TAmountWithConversions;
+  claimFeeSat?: number;
+  refundFeeRateSatPerVbyte?: number;
```
Beerosagos, line 48: We should get user ack on the actual fee, not the fee rate

backend/lightning/topup_recovery.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (lightning *Lightning) recommendedRefundFeeRate() (uint64, error) {
+	recommendedFees, err := lightning.sdkService.RecommendedFees()
+	if err != nil {
+		return 0, errp.Wrap(err, "breez: recommended fees")
+	}
+	if recommendedFees.MinimumFee == 0 {
+		return 0, errp.New("no recommended fee rate available")
+	}
+	return recommendedFees.MinimumFee, nil
```
Beerosagos: Sounds risky to use the MinimumFee. The refund tx might get stuck as well and RBF would be.. complicated. I'd use the highest possible. 

A couple of caveats I noticed in my tests: 
1. using the minimum fee I got this error: `generic error: fee must be at least 194 sats`. Looks like a breez sdk error, we could check if bumping breez fixes that.
2. Using the highest fee would be better, but we should check that the remaing amount is above dust limit, or the refund will fail with smth like `Refund amount (104 sats) is below the minimum of 294 sats required for this address`.  Not sure how to deal with that, but it sounds wise to put a lower limit to top-up amounts.

backend/lightning/handlers.go
```diff
@@ -218,6 +220,52 @@ func (lightning *Lightning) GetBoardingAddress(_ *http.Request) interface{} {
 	return responseDto{Success: true, Data: address}
 }
 
+// PostClaimTopUp handles the POST request to manually claim a Bitcoin top-up.
+func (lightning *Lightning) PostClaimTopUp(r *http.Request) interface{} {
+	var jsonBody struct {
+		TxID           string `json:"txid"`
```
Beerosagos: Could we use the LightningPayment ID field instead to identify the deposit? Could be a bit simpler and doesn't require the FE to deal with Vout

Beerosagos (changes requested):

frontends/web/src/api/lightning.ts
bznein: we don't get the actual fee until we explicitly try to refund unfortunately, because that is not finalized until the tx is built, which requires the destination address. And if the user has multiple accounts saved, with different address recv type, the fee will change based on what they select

bznein: I guess we can show the actual fee again, just before the final confirmation. Will need to take some liberty with the UI but wec hange it anyway if we don't like it

backend/lightning/handlers.go
bznein: Yep absolutely, will do!

frontends/web/src/api/lightning.ts
bznein: Actually, we can't :) Unless we implement the whole math to compute the fee. Spark does not provide the fee as a separate step from the confirmation, so we never really know the fee until it is too late. Not sure what the best thing to do would be here, wdyt?

backend/lightning/topup_recovery.go
bznein: Agreed, MinimumFee is too risky. I (well, codex) checked the SDK and the 194-sat minimum is still present in newer versions, so bumping Breez would not fix this. I’d use max(FastestFee, 2 sat/vB) and check that the remaining amount stays above the selected destination’s dust limit before attempting the refund.

bznein: Implemented

frontends/web/src/api/lightning.ts
Beerosagos: Ah, you are right. I thought that the `fee := breez_sdk_spark.Fee(breez_sdk_spark.FeeRate{SatPerVbyte: recommendedFees.HalfHourFee})` provided the number but ofc it doesn't have all the info to do that. 

Since the sdk doesn't allow to check for the fee (and the user doesn't have any other option between claim and refund), I think it's ok to just show the needed claim fee and allow the user to choose between accepting it and claim, or refunding (with an unknown fee).

cc @jadzeidan wdyt?

backend/lightning/topup_recovery.go
Beerosagos: Not sure if I already wrote it here, but I think it would make sense to set a lower limit in the top up flow, to avoid edge cases where the topup can't be either claimed or refunded. But it can also be a separate PR.

Beerosagos: Makes sense to use `max(FastestFee, 2 sat/vB)`, but I'd honestly wouldn't complicate the code for the dust amount case: the stuck top-up in this very edgy case is basically lost anyway, I think. Better to prevent it with a lower top-up amount threshold as I suggested above, imo. wdyt?

frontends/web/src/routes/lightning/claim-top-up/failure-step.tsx
```diff
@@ -0,0 +1,53 @@
...
+  errorMessage,
+  onDone,
+  onRefund,
+}: TProps) => {
+  const { t } = useTranslation();
+  const canTryRefund = action === 'claim' && !!onRefund;
+  const messageKey = action === 'claim'
+    ? 'lightning.claimTopUp.failure.claimFailedMessage'
+    : 'lightning.claimTopUp.failure.refundFailedMessage';
+  const note = action === 'claim'
```
shonsirsha: Mm ok. Why not do

```suggestion
const note = canTryRefund
  ? t('lightning.claimTopUp.failure.claimFailedTryRefund')
  : errorMessage || t('lightning.claimTopUp.failure.refundFailedTryAgain');
```
When a claim fails but refund isn't available somehow (no btc account / no refund fee rate), this still says "Try refunding it..." - which feels incorrect to me.

frontends/web/src/api/lightning.ts
```diff
@@ -271,6 +276,33 @@ export const postCloseWithdraw = async (
...
+  approvedFeeRateSatPerVbyte: number,
+): Promise<TTopUpRecoveryResult> => {
+  return postApiResponse<TTopUpRecoveryResult, {
+    paymentId: string;
+    destinationAccountCode: AccountCode;
+    approvedFeeRateSatPerVbyte: number;
+  }>(
+    'lightning/refund-top-up',
+    { paymentId, destinationAccountCode, approvedFeeRateSatPerVbyte },
+    'Error calling postRefundTopUp'
```
shonsirsha: We shoul add lcalisation for these new top-up errors, right? 

If refunding fails, the UI can show `Error calling postRefundTopUp` - which isn't ideal imo.

frontends/web/src/api/lightning.ts
jadzeidan: makes sense to me.

backend/lightning/topup_recovery.go
bznein: > Makes sense to use `max(FastestFee, 2 sat/vB)`, but I'd honestly wouldn't complicate the code for the dust amount case: the stuck top-up in this very edgy case is basically lost anyway, I think. Better to prevent it with a lower top-up amount threshold as I suggested above, imo. wdyt?

I'm not super sure I follow, can you clarify what you mean with "lower top-up threshold?" (we currently do not have any except for the wallet limit) and how it would help? thanks!

Beerosagos: > I'm not super sure I follow, can you clarify what you mean with "lower top-up threshold?" (we currently do not have any except for the wallet limit) and how it would help? thanks!

I am referring to the `validateRefundAmount` part: that basically checks if the amount of the stuck top-up is enough to be refunded, right? What I mean is that if a user ends up in a situation where the amount is not enough, that top-up is basically lost anyway (assuming it can't be claimed either). Imo we should avoid the user ending up in this situation by adding a minimum amount threshold in the top-up creation. If we do that, we can avoid adding the code that checks for the refundability, as top-ups should always be claimable/refundable.
And even if we end up in that situation, the code would only help in showing a nicer error compared to the sdk native error, so not a big deal, imo.

bznein: aaah by "lower" you mean like, the minimum amount allowed to top-up? 

Yeah I agree then! Maybe let's create an issue to decide on that value? I don't want this PR to get bigger than it already is :)

Beerosagos: Ack! Do you agree on removing `validateRefundAmount` from this PR since that case will be prevented in the top-up creation flow?

bznein: It's already removed: https://github.com/BitBoxSwiss/bitbox-wallet-app/compare/9be84af1111672b2937fbbe95101d5cd0ef8959a..0cf66227d38332af2794e2222daf3183c56e293e You might have been seeing an older commit :)

Beerosagos: Sorry, must have looked at a previous version

Beerosagos (approved):
backend LGTM 🙏

PR #4308, opened 2026-08-06
bitbox02: require antiklepto for signing

benma-agent:
Reject transaction and message signing on firmware versions that predate
the relevant antiklepto support.

Check required signing features at each workflow's existing keystore connection,
so firmware upgrades are prompted immediately after connecting.

  - [ ] Normal BTC/LTC transaction send
  - [ ] Normal ETH/ERC20 transaction send
  - [ ] Bitrefill checkout payment
  - [ ] BTC Direct sell/payment
  - [ ] Pocket sell payment
  - [ ] Swap sell transaction
  - [ ] Account message signing, BTC or ETH
  - [ ] AOPP ownership proof, BTC or ETH
  - [ ] Pocket address ownership proof
  - [ ] Bitsurance address ownership proof
  - [ ] WalletConnect plain message signing, eth_sign or personal_sign
  - [ ] WalletConnect transaction signing, eth_signTransaction or eth_sendTransaction
  - [ ] WalletConnect typed-data signing, using any one typed-data variant

### DIFF
--- a/frontends/web/src/components/wallet-connect/incoming-signing-request.tsx
+++ b/frontends/web/src/components/wallet-connect/incoming-signing-request.tsx
@@ -15,6 +15,7 @@ import { alertUser } from '@/components/alert/Alert';
 import type { TAccount } from '@/api/account';
 import { connectKeystore } from '@/api/keystores';
 import { TStage, WCIncomingSignRequestDialog } from './incoming-signing-request-dialog';
+import { FirmwareUpgradeRequiredDialog } from '@/components/dialog/firmware-upgrade-required-dialog';
 
 type TActiveSigningRequest = {
   accountCode: TLaunchSignDialog['accountCode'];
@@ -32,6 +33,7 @@ export const WCSigningRequest = ({ accounts }: TProps) => {
   const [dialogOpen, setDialogOpen] = useState(false);
   const [dialogContent, setDialogContent] = useState<TRequestDialogContent>();
   const [stage, setStage] = useState<TStage>('initial');
+  const [firmwareUpgradeRequired, setFirmwareUpgradeRequired] = useState(false);
   const activeRequestRef = useRef<TActiveSigningRequest>();
   const successTimerRef = useRef<ReturnType<typeof setTimeout>>();
 
@@ -128,22 +130,32 @@ export const WCSigningRequest = ({ accounts }: TProps) => {
 
     setStage('initial');
     setDialogOpen(false);
+    if (result.errorCode === 'firmwareUpgradeRequired') {
+      setFirmwareUpgradeRequired(true);
+      return;
+    }
     if (!result.aborted) {
       alertUser(result.errorMessage || t('pairing.error.text'));
     }
   };
 
-  if (!dialogContent || !dialogOpen) {
-    return null;
-  }
-
   return (
-    <WCIncomingSignRequestDialog
-      content={dialogContent}
-      open={dialogOpen}
-      stage={stage}
-      onAccept={handleAcceptBtn}
-      onReject={handleRejectBtn}
-    />
+    <>
+      {firmwareUpgradeRequired && (
+        <FirmwareUpgradeRequiredDialog
+          open
+          onClose={() => setFirmwareUpgradeRequired(false)}
+        />
+      )}
+      {dialogContent && dialogOpen && (
+        <WCIncomingSignRequestDialog
+          content={dialogContent}
+          open={dialogOpen}
+          stage={stage}
+          onAccept={handleAcceptBtn}
+          onReject={handleRejectBtn}
+        />
+      )}
+    </>
   );
 };
--- a/frontends/web/src/hooks/keystore.ts
+++ b/frontends/web/src/hooks/keystore.ts
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { useCallback, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import {
+  connectAnyKeystore,
+  connectKeystore,
+  TKeystoreFeature,
+} from '@/api/keystores';
+import { alertUser } from '@/components/alert/Alert';
+
+export const useFeatureConnect = () => {
+  const { t } = useTranslation();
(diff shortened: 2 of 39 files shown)


bznein (approved):
backend code LGTM haven't had time to test it yet

frontends/web/src/hooks/keystore.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const useFeatureConnect = () => {
+  const { t } = useTranslation();
+  const [firmwareUpgradeRequired, setFirmwareUpgradeRequired] = useState(false);
+
+  const handleResult = useCallback((result: Awaited<ReturnType<typeof connectKeystore>>) => {
+    if (result.errorCode === 'firmwareUpgradeRequired') {
+      setFirmwareUpgradeRequired(true);
+    } else if (result.errorCode === 'unsupportedFeature') {
+      alertUser(t('device.unsupportedFeature'));
```
thisconnect, line 20: question: In the BE

> // ErrUnsupportedFeature is returned when a certain feature is unsupported by the keystore.

but I wonder when this should happen or how to test... or why doesn't it just say firmwareUpgradeRequired ?

The type only allows to use this api with known TKeystoreFeature's so how can it be unknown?

frontends/web/src/routes/account/sign-message/sign-message.tsx
```diff
@@ -42,6 +43,12 @@ export const SignMessage = ({
 
   return (
     <Main>
+      {controller.firmwareUpgradeRequired && (
```
thisconnect, line 46: testing "sing message" with an old FW 9.10.0 in testnet/webdev

firmwareUpgradeRequired is false here

and connect-keystore is only called after I click "Sign on device", it submits requiredFeature	"messageSigning" and rootFingerprint and returns success	true

ending up in "An unexpected error occurred."

Image

frontends/web/src/routes/account/actionButtons.tsx
```diff
@@ -23,18 +24,25 @@ type TProps = {
...
+  // connected to decide which ETH transaction proposals to construct (legacy vs EIP1559).
   const sendLink = `/account/${code}/send`;
-  const maybeRouteSend = async (e: MouseEvent<HTMLAnchorElement>) => {
+  const routeSend = async (e: MouseEvent<HTMLAnchorElement>) => {
     e.preventDefault();
-    const connectResult = await connectKeystore(account.keystore.rootFingerprint);
-    if (connectResult.success) {
+    const requiredFeature = isEthereumBased(coinCode)
+      ? 'ethTransactionSigning'
+      : 'btcTransactionSigning';
```
thisconnect, line 45: Wouldn't all fw versions support ethTransactionSigning and btcTransactionSigning ?

Reason I was lookin at this is that Send button is a bit early to show a fw-upgrade, it would be better to allow the user to go to send and then show the dialog.
but since this never comes up we might as well remove completely or just keep as is 🤷 ?

frontends/web/src/components/wallet-connect/incoming-signing-request.tsx
```diff
@@ -102,6 +104,11 @@ export const WCSigningRequest = () => {
         setStage('initial');
         setDialogOpen(false);
         await web3wallet?.respondSessionRequest({ topic, response: rejectMessage(id) });
+      } else if (error.errorCode === 'firmwareUpgradeRequired') {
```
thisconnect: I could sign a message with eth & FW 9.10.0

and in testnet it allows me to sign but says "warnging unknown network with chain ID..."
then "UNKNOWN unusal keypath warning"..

I guess that is expected, but couldn't test this code path with fw 9.10.0.

frontends/web/src/routes/market/swap/swap.tsx
```diff
@@ -410,18 +406,6 @@ export const Swap = ({
...
-      if (!paymentRequestSupport.success) {
-        if (paymentRequestSupport.errorCode === 'firmwareUpgradeRequired') {
-          setFwRequiredDialog(true);
-        } else if (paymentRequestSupport.errorCode) {
-          alertUser(t(`device.${paymentRequestSupport.errorCode}`));
-        } else {
-          alertUser(paymentRequestSupport.errorMessage || t('genericError'));
-        }
-        return;
-      }
```
thisconnect, line 423: nice 👍 

tested swap with fw 9.10 shows firmware upgrade 👍

frontends/web/src/routes/market/market.tsx
```diff
@@ -182,6 +197,12 @@ export const Market = ({
 
   return (
     <GuideWrapper>
+      {firmwareUpgradeRequired && (
+        <FirmwareUpgradeRequiredDialog
+          open
+          onClose={dismissFirmwareUpgrade}
+        />
+      )}
```
thisconnect, line 205: partly tested 👍

thisconnect (approved):
very nice, partly tested LGTM with 2 questions and one unexpected error tested with really old FW 9.10

frontends/web/src/routes/account/actionButtons.tsx
benma: The linear issue discusses that it should prompt early, not when filling out everything and clicking review, as it would be annoying to do all this and fail after.

Not all fw support signing it with antiklepto. This PR is about checking this and enforcing it.

frontends/web/src/hooks/keystore.ts
benma-agent: `TKeystoreFeature` constrains which capability can be requested; it does not mean that every keystore supports every listed capability.

`firmwareUpgradeRequired` means the connected keystore supports the feature in principle, but its current firmware is too old. `unsupportedFeature` means that this keystore implementation cannot provide the feature at all, so prompting for a firmware upgrade would be misleading. A concrete example is the software/test keystore: `paymentRequests` is a known feature, but `software.Keystore.SupportsFeature()` returns `ErrUnsupportedFeature` for it. `TestConnectKeystoreRequiredFeature/unsupported feature` covers that case.

The default branch in the BitBox02 implementation also protects the runtime HTTP boundary against arbitrary strings or a future frontend/backend enum mismatch, since the TypeScript type is not present at runtime.

PR #4309, opened 2026-08-07
backend: stabilize websocket test

benma-agent:
Wait for the two expected messages with a one-second deadline instead of assuming they arrive
within 10ms. This avoids scheduler-dependent failures when the test suite runs under the race
detector.


bznein (approved):
utACK

PR #4310, opened 2026-08-08
Use gap limit 20 for change address discovery

fabio-porcedda:
Use the commonly adopted gap limit of 20 when discovering UTXOs on change addresses generated by other wallets.

Hardware wallets used with other apps commonly use a gap limit of 20. As a result, those apps may generate change addresses beyond the BitBoxApp's previous gap limit of 6, preventing the BitBoxApp from discovering the corresponding addresses and UTXOs.

Since not all users will recognize this issue or know how to adjust the gap limit manually, use the more common value of 20 by default.

Reviewed with AI.

Tested on Linux and Android.

### DIFF
--- a/backend/coins/btc/account.go
+++ b/backend/coins/btc/account.go
@@ -174,7 +174,7 @@ func (account *Account) defaultGapLimits(signingConfiguration *signing.Configura
 		// Usually 20, but BWS used to not have any limit. We put it fairly high to cover most
 		// outliers.
 		limits.Receive = 60
-		account.log.Warning("increased change gap limit to 20 and gap limit to 60 for BWS compatibility")
+		account.log.Warning("increased change gap limit to 25 and gap limit to 60 for BWS compatibility")
 	}
 
 	return limits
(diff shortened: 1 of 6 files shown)


backend/coins/btc/account.go
cedwies: This comment is stale.

frontends/web/src/routes/settings/components/advanced-settings/custom-gap-limit-setting.tsx
cedwies, line 31: This loads any positive value from the config. In advanced settings -> Custom gap limit settings: Users who saved a previously legal value (between 6 and 19) will see the value plus an Error (manually tested it), even though backend silently turns the now illegal value to 20. This value should be migrated properly.

cedwies (changes requested):
Thank you, left some comments. This PR should have a regression tests, testing proper migration and the new gap limit.

frontends/web/src/routes/settings/components/advanced-settings/custom-gap-limit-setting.tsx
fabio-porcedda: I've added migration logic.

backend/coins/btc/account.go
fabio-porcedda: Removed the stale comment.

thisconnect:
hi, we added the configurtation so merchants could increase to 20.

I am slightly against changing the default for everyone. Why is it needed?

PR #4312, opened 2026-08-10
frontend: LN menu

shonsirsha:
- Add a Lightning shortcut to the mobile bottom navigation when enabled
- Hide "Accounts" when Lightning is the only active account


frontends/web/src/components/bottom-navigation/menu-icons.tsx
```diff
@@ -13,6 +13,12 @@ export const AccountIconSVG = () => (
   </svg>
 );
 
+export const LightningIconSVG = () => (
+  <svg width="40" height="70" viewBox="0 0 40 70" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path d="M23.6011 2L2 43.2503H16.3989V68L38 26.7497H23.6011V2Z" stroke="currentColor" strokeWidth="4" strokeLinejoin="round"/>
+  </svg>
+);
```
thisconnect, line 20: nit We shouldn't need any of these big SVG's in .tsx files.

Please move to .svg files and import via icon component.

thisconnect (approved):
partly tested LGTM with nit

PR #4314, opened 2026-08-11
lightning: enforce minimum top-up amount

bznein:
Small top-ups can become impossible to claim or refund. Enforce a 1,000 sat minimum using backend-owned validation and expose the result through the existing balance and amount responses.

Hide Bitcoin accounts whose available balance does not exceed the threshold, since an exact-minimum balance cannot also cover network fees. Leave final fee sufficiency to the normal transaction proposal.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/backend/lightning/handlers.go
+++ b/backend/lightning/handlers.go
@@ -61,6 +61,7 @@ func (lightning *Lightning) PostPrepareTopUp(r *http.Request) interface{} {
 		Success      bool          `json:"success"`
 		ErrorCode    string        `json:"errorCode,omitempty"`
 		FundingLimit *fundingLimit `json:"fundingLimit,omitempty"`
+		MinAmountSat uint64        `json:"minAmountSat,omitempty"`
 		*topUpProposal
 	}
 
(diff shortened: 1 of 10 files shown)


frontends/web/src/routes/lightning/topup/topup.tsx
```diff
@@ -102,6 +102,34 @@ export const LightningTopUp = ({ activeAccounts, hasAccounts }: TProps) => {
...
+      });
+      return {
+        amount,
+        unit: sourceAmountUnit,
+        valid: response.success ? response.validLightningTopUpAmount : null,
+      };
+    } catch {
+      return { amount, unit: sourceAmountUnit, valid: null };
+    }
+  }, [amount, sourceAmountUnit]);
```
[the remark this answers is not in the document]
bznein: fixed

frontends/web/src/locales/en/app.json
```diff
@@ -1852,8 +1852,9 @@
         "total": "Total"
       },
       "from": "From",
+      "minimumAmount": "Enter at least 1000 sats.",
```
bznein: @Beerosagos @jadzeidan suggestions on how to phrase this? I put a temporary string but we should probably use something more explicit such as "to minimize the risk on unclaimable deposits, deposit at least 1000 sats"

Beerosagos: I would avoid tech details. I think it's fine just to state that the min amount is 1K sats

backend/lightning/topup.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/BitBoxSwiss/bitbox-wallet-app/util/errp"
 )
 
+const minimumTopUpAmountSat = 1000
+
+var minimumTopUpAmount = big.NewInt(minimumTopUpAmountSat)
```
Beerosagos: nit: this is only used in the validate function, no need to be defined outside.

backend/lightning/handlers.go
Beerosagos, line 59: nit: error handling in this function is a bit weird: we mix three different ways for catching specific errors. Maybe we can use something like:


```suggestion
var limitErr *topUpFundingLimitError
if errors.As(err, &limitErr) {
    // ...
}
var minimumErr *lightningAmountBelowMinimumError
if errors.As(err, &minimumErr) {
      // ...
}

var validationErr accountErrors.TxValidationError
if errors.As(err, &validationErr) {
    // ...
}
```

Beerosagos (approved):
tACK with a couple of nits 🙏

backend/lightning/handlers.go
bznein: done

backend/lightning/topup.go
bznein: moved

PR #4318, opened 2026-08-11
lightning: warn about incoming funds before shutdown

bznein:
Incoming and unclaimed funds cannot be included in a close-and-withdraw transaction. Show a warning in the confirmation step and link back to the Lightning account, where users can claim or refund those funds.

Keep shutdown available for users who choose to continue without withdrawing every fund.

screenshot_11-Aug-2026_16-01-48

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos (approved):
utACK! Nice

PR #4319, opened 2026-08-11
lightning: warn about lightning not using tor.

bznein:
Lightning's connections are handled by the breeze sdk and can not be configured to use a specific proxy. Warn the user that lightning will still work but it will not be using Tor.

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos (approved):
tACK

PR #4321, opened 2026-08-12
build: verify vendored Go dependencies

benma-agent:
Release builds compile vendor/ directly, but go.sum authenticates downloaded
module archives, not the copied vendor tree. A source-only vendor change can
therefore bypass go mod verify.

Regenerate vendor/ from a fresh, read-only module cache and compare it before
platform builds in CI and release workflows. This prevents Qt, Android, and iOS
artifacts from using vendored sources that do not match the checksummed modules.

Verified with a clean vendor tree and a temporary source-only modification,
which the check rejected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ WEBROOT  := frontends/web
 
 include version.mk.inc
 
+.PHONY: check-go-vendor
+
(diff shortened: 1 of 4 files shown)


bznein (approved):
utACK

PR #4322, opened 2026-08-12
feat: adjust unit price

shonsirsha:
- Hide the redundant BTC unit price in account summaries when the selected currency is BTC or sat.
- Normalize native BTC/sat exchange rates in the backend so CoinGecko BTC/BTC drift cannot affect totals,

### DIFF
--- a/backend/rates/rates.go
+++ b/backend/rates/rates.go
@@ -40,6 +40,31 @@ const interval = time.Minute
 // unitSatoshi is 1 BTC (default unit) in Satoshi.
 const unitSatoshi = 1e8
 
+func normalizeBitcoinUnitRates(rates map[string]map[string]float64) {
+	btcRates := rates[BTC.String()]
+	if btcRates == nil {
+		return
+	}
+	// BTC/sat conversions are unit identities, not market prices.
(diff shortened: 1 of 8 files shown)


backend/rates/rates.go
```diff
@@ -40,6 +40,31 @@ const interval = time.Minute
...
+		case SAT.String():
+			return unitSatoshi
+		}
+	}
+	if fiat == SAT.String() {
+		return value * unitSatoshi
+	}
+	return value
+}
+
```
bznein, line 67: [nit] I think this nesting of ifs and switch is a bit complicated to follow, imho something like this would be clearer to follow:

```go
func normalizeHistoricalRate(geckoCoinID, fiat string, value float64) float64 {
	isBitcoin := geckoCoinID == geckoCoin["btc"]

	switch {
	case isBitcoin && fiat == BTC.String():
		return 1
	case isBitcoin && fiat == SAT.String():
		return unitSatoshi
	case fiat == SAT.String():
		return value * unitSatoshi
	default:
		return value
	}
}

bznein (approved):
backend tested LGTM

PR #4325, opened 2026-08-13
frontend: remove Lightning from accounts list

Beerosagos:
The mobile bottom navigation now provides a dedicated Lightning shortcut. Keeping another Lightning entry in the accounts list duplicates navigation and leaves a regular-account-only view available when empty.

Remove the Lightning row and its balance request. Redirect an empty accounts list to the portfolio when Lightning is available, or home when no wallet accounts exist. Cover both outcomes with route tests.


shonsirsha (approved):
Lgtm

PR #4326, opened 2026-08-14
frontend: revamp QR scanner

shonsirsha:
- revamp QR scanner
- use this QR scanner in 3 places: onchain send, LN send, and WalletConnect scan


frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -1,29 +1,42 @@
...
 };
 
 export const useQRScanner = (
   videoRef: RefObject<HTMLVideoElement>, {
     onStart,
     onResult,
     onError,
+    highlightScanRegion = true,
+    highlightCodeOutline = true,
+    stopOnResult = true,
```
thisconnect: If  any of these are always disabled you can remove from the props and just hardcode.

frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -1,29 +1,42 @@
...
   }: TUseQRScannerOptions
 ) => {
   const { t } = useTranslation();
-  const [initErrorMessage, setInitErrorMessage] = useState();
+  const [initErrorMessage, setInitErrorMessage] = useState<string | undefined>(undefined);
+  const [hasFlash, setHasFlash] = useState(false);
+  const [isFlashOn, setIsFlashOn] = useState(false);
   const scanner = useRef<QrScanner | null>(null);
-  // loading is set to true while the scanner is being created/started/stopped/destroyed,
-  // this allows to sync across re-renders.
```
thisconnect, line 24: keep comment?

frontends/web/src/hooks/qrcodescanner.ts
```diff
@@ -63,21 +78,23 @@ export const useQRScanner = (
             }
           }
         );
-        // Somehow, the new QrScanner may return before it is ready to be started.
-        // We don't have a way to know when it is ready, but this 300ms wait seems
-        // to work well enough.
```
thisconnect, line 68: I believe this comment is still relevant? keep comment?

frontends/web/src/locales/en/app.json
```diff
@@ -2247,7 +2249,10 @@
     "noFeeTargets": "Fee rate estimations are currently unavailable. Please try again later or enter a custom fee.",
     "priority": "Priority",
     "scanQR": "Scan QR code",
+    "scanQRInstruction": "Point your camera at an address QR code",
     "scanQRNoCameraMessage": "Camera not found. Please ensure that your device supports a camera and permissions are correctly set.",
+    "scanQRPasteAddress": "Paste an address",
+    "scanQRToggleFlash": "Toggle flashlight",
```
thisconnect, line 2265: tiny-nit: We don't really support aria (yet) and this is **only** used  for aria label, imo  it is good enough for now to hardcode the aria-label to english or completely remove .

at some point we could do a separate PR to support aria everywhere

frontends/web/src/routes/account/send/components/inputs/scan-qr.tsx
```diff
@@ -0,0 +1,235 @@
...
+    }
+    if (!fromPaste && text === processedRef.current) {
+      return;
+    }
+    processedRef.current = text;
+    busyRef.current = true;
+    setProcessing(true);
+    try {
+      const accepted = await onResult(text);
+      if (accepted === false) {
```
thisconnect, line 80: I wondered why the onResult has type

`(result: string) => void | boolean | Promise<void | boolean>`

Could you add a docstring to the ScanQR to document what the intention of this prop is especially the return value (boolean)?

frontends/web/src/routes/account/send/components/inputs/scan-qr.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const handleToggleFlash = () => {
+    triggerHapticFeedback();
+    toggleFlash();
+  };
+
+  const handleSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    const value = manualValue.trim();
+    if (value) {
+      void handleValue(value, true);
```
thisconnect: void does nothing here right?

thisconnect (dismissed):

frontends/web/src/routes/lightning/lightning.test.tsx
```diff
@@ -67,18 +68,20 @@ const balance: lightningApi.TLightningBalance = {
...
+        btcUnit: 'sat',
+        rotateDefaultCurrency: vi.fn(),
+        rotateBtcUnit: vi.fn(),
+        addToActiveCurrencies: vi.fn(),
+        updateDefaultCurrency: vi.fn(),
+        removeFromActiveCurrencies: vi.fn(),
+      }}>
+        <Lightning />
+      </RatesContext.Provider>
+    </ConfigContext.Provider>
```
shonsirsha, line 88: Fixes failing test (exists on the target branch).

bznein (commented):
I tested and it looks great (haven't look much at the code).

One nit that I noticed: if the clipboard is empty or it contains something other than text (for example images), the "paste from clipboard" (both the button with text in the camera view, and the icon near the text input) is clickable but has no effect, making it a bit confusing.

I'm not sure what's the best way to fix it, is it possible to detect whether the clipboard contains non-text content (or is empty)?

bznein (approved):
tested LGTM!

(haven't looked much at the code)

Beerosagos (approved):
Didn't check the code, but tACK on qt/linux

PR #4329, opened 2026-08-18
frontend: limit eth token decimals

thisconnect:
- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/components/amount/amount.tsx
+++ b/frontends/web/src/components/amount/amount.tsx
@@ -2,6 +2,7 @@
 
 import { useContext } from 'react';
 import type { CoinUnit, ConversionUnit } from '@/api/account';
+import { isSupportedERC20Unit } from '@/api/erc20';
 import { AppContext } from '@/contexts/AppContext';
 import { LocalizationContext } from '@/contexts/localization-context';
 import { useMediaQuery } from '@/hooks/mediaquery';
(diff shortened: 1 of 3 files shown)


frontends/web/src/components/amount/amount.tsx
```diff
@@ -183,6 +207,16 @@ export const FormattedAmount = ({
...
     return formatEth(displayedAmount, group, decimal, maxDecimals);
+  case 'USDT':
+  case 'USDC':
+  case 'LINK':
+  case 'BAT':
+  case 'MKR':
+  case 'ZRX':
+  case 'WBTC':
+  case 'PAXG':
+  case 'DAI':
```
strmci: Could `supportedERC20Tokens` be used here instead of listing every token?

thisconnect: yes but it would make the code a bit more complicated as it needs to check against each item in the `supportedERC20Tokens` array

so it would be slightly less performant, some views like account overview have many instances of the amount component... I'm not sure if we should but probably it wont have a big performance impact 🤔

strmci (approved):
LGTM

frontends/web/src/components/amount/amount.tsx
thisconnect: We could pre-map the supportedERC20Tokens into a flat array and always check against that.. but still couldn't use the switch case directly but need some function to invoke it.. I think for now lets just keep as is.

thisconnect: @strmci I added one more commit, let me know if you prefer this then I squash into the last commit

strmci: thanks, I checked it, and I think it looks better. I also asked GPT about the performance impact, and it thinks it's negligible.

PR #4331, opened 2026-08-18
frontend: move mobile settings into bottom navigation

sutterseba:
After moving Insurance to the Marketplace, "Settings" was now the only item inside "More", causing unnecessary friction to enter the actual settings on mobile.

This moves settings directly into the menu bar, removing the redundant intermediate step.


frontends/web/src/components/bottom-navigation/menu-icons.tsx
```diff
@@ -19,9 +19,15 @@ export const MarketIconSVG = () => (
   </svg>
 );
 
-export const MoreIconSVG = () => (
+export const SettingsIconSVG = () => (
```
thisconnect: Nit: could you just use CogBlue from Icon componet?

thisconnect (approved):
tested Android & macOS LGTM with 1 nit.

frontends/web/src/components/bottom-navigation/menu-icons.tsx
sutterseba: the cog components have baked-in color, so we would need to explicitly switch between them (blue, dark, light). the other bottom nav icons are also inline and simply inherit `currentColor`, so I thought it would be more consistent this way. 

if this is about visual consistency: we could just use the existing cog svg inline as well

thisconnect: correct we cannot use currentColor but it's still better to have 2 .svg files than having all icons in javscript bundle. That is the reason we add -light.svg and -dark.svg for all icons that need it.

sutterseba: updated

note the other three menu icons are still inline JS

thisconnect: ack. lets merge as is,.. one day I'll go over again and move them to svg files.

PR #4332, opened 2026-08-19
ios: disable web view inspection

Beerosagos:
iOS 16.4 allows apps to opt embedded web views into Safari Web Inspector access. Keeping this enabled in production lets an attached inspector execute JavaScript in the app context and reach the native bridge, including the Lightning payment endpoint.

Explicitly disable inspection so release builds do not expose the bridge through Safari developer tooling.


shonsirsha (approved):
LGTM

PR #4336, opened 2026-08-19
frontend: fix back navigation not working in lightning menus

sutterseba:
BackButton component should be used for back buttons to enable back navigation gestures on iOS and Android.

Also enables back navigation for the QR scanner.


bznein (approved):
utACK

PR #4342, opened 2026-08-20
backend: require HTTPS AOPP callbacks

benma-agent:
Reject AOPP requests whose callback is not an absolute HTTPS URL
with a non-empty host. Allow redirects only while HTTPS is preserved
so signed addresses and extended public keys cannot be downgraded to
cleartext transport.

Make AOPP error rendering tolerate missing or malformed callback URLs.
Cover callback validation, HTTPS redirects, HTTP downgrade rejection,
and missing callbacks in backend and frontend tests.

### DIFF
--- a/backend/aopp.go
+++ b/backend/aopp.go
@@ -248,9 +248,15 @@ func (backend *Backend) handleAOPP(uri url.URL) {
 		backend.aoppSetError(errAOPPInvalidRequest)
 		return
 	}
-	_, err := url.Parse(callback)
+	callbackURL, err := url.Parse(callback)
 	if err != nil {
 		log.WithError(err).Error("Invalid callback")
+		backend.aoppSetError(errAOPPInvalidRequest)
+		return
+	}
+	if !strings.EqualFold(callbackURL.Scheme, "https") ||
+		callbackURL.Host == "" {
+		log.Error("Invalid callback")
 		backend.aoppSetError(errAOPPInvalidRequest)
 		return
 	}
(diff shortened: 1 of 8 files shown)


backend/aopp.go
```diff
@@ -248,12 +248,19 @@ func (backend *Backend) handleAOPP(uri url.URL) {
...
 		return
 	}
-	_, err := url.Parse(callback)
+	callbackURL, err := url.Parse(callback)
 	if err != nil {
 		log.WithError(err).Error("Invalid callback")
 		backend.aoppSetError(errAOPPInvalidRequest)
 		return
 	}
+	if !callbackURL.IsAbs() ||
```
bznein: [nit] I think we can avoid this check, because this simply does:

```go
func (u *URL) IsAbs() bool {
	return u.Scheme != ""
}
```
And if the scheme is empty it would fail anyway the `EqualFold` check blow.

bznein (commented):
Codex findings:

```
• The HTTPS restriction can still downgrade through redirects, and its newly rejected
  HTTP path sends frontend state that crashes error rendering. Both defects undermine
  the intended secure failure behavior.

  Full review comments:

  - [P1] Preserve rejected callbacks for error rendering — /home/bznein-coding-agent/
    agent-work/bitbox-wallet-app/backend/aopp.go:260-262
    When an AOPP URI contains a parseable HTTP callback, this branch sets the error
    before backend.aopp.Callback is assigned, so the frontend receives an empty
    callback. The error view calls new URL(aopp.callback).host in frontends/web/src/
    components/aopp/aopp.tsx, which throws and blanks the flow instead of displaying
    aoppInvalidRequest; preserve the parseable callback or make error rendering tolerate    absent callbacks.

  - [P1] Reject cleartext redirect targets — /home/bznein-coding-agent/agent-work/
    bitbox-wallet-app/backend/aopp.go:257-259
    When an accepted HTTPS callback responds with a 307 or 308 redirect to an HTTP URL,
    Go's default http.Client.Post follows it and resends the signed address and optional    xpub over cleartext. Validating only the initial URL therefore does not enforce the
    new transport guarantee; disable redirects for this callback or reject every
    redirect whose target is not HTTPS.
```

backend/aopp.go
benma: Fixed

bznein (approved):
utACK

PR #4343, opened 2026-08-20
frontend: reduce page load flicker

shonsirsha:
Fast page loads and tab navigation caused visible flickering. This is because the global banners were mounted inside each route, so navigating remounted them and changed the page layout after rendering. 

Also loaded balances, charts, and transaction lists replaced the skeletons abruptly.

How we fix it in this PR:

- Mount global banners once above the router so their state survives navigation.
- Fade loaded balances, charts, and transaction lists into place.

### DIFF
--- a/frontends/web/src/components/balance/balance.tsx
+++ b/frontends/web/src/components/balance/balance.tsx
@@ -9,10 +9,12 @@ import style from './balance.module.css';
 
 type TProps = {
   balance?: TBalance;
+  className?: string;
 };
 
 export const Balance = ({
   balance,
+  className = '',
 }: TProps) => {
   const { t } = useTranslation();
   if (!balance) {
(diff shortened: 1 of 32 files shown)


frontends/web/src/components/balance/balance.tsx
```diff
@@ -9,10 +9,12 @@ import style from './balance.module.css';
 
 type TProps = {
   balance?: TBalance;
+  loadedClassName?: string;
```
thisconnect: naming nit: from the Balance component's perspective this should be just regular className prop

frontends/web/src/contexts/global-banners-provider.tsx
```diff
@@ -0,0 +1,34 @@
...
+      {children}
+      {createPortal(
+        <ContentWrapper className={style.content}>
+          <GlobalBanners devices={devices} />
+        </ContentWrapper>,
+        container,
+      )}
+    </GlobalBannersContainerContext.Provider>
+  );
+};
```
thisconnect, line 49: nit: could you move the provider and context into `./contexts` ?

thisconnect (changes requested):

thisconnect (approved):
LGTM with few small nits, thank you

PR #4349, opened 2026-08-24
frontend: rename shutdown to disable wallet (LN) and rearrange

shonsirsha:
Screenshot 2026-08-24 at 13 55 19


sutterseba (approved):
utACK

PR #4350, opened 2026-08-24
docs: clarify backend and frontend roles

benma-agent:
Keep business logic in the backend and the frontend focused on
presentation. Document UUID generation, amount parsing, and arithmetic
as backend responsibilities.


thisconnect (approved):
untested LGTM

(I don't know how to test this)

PR #4358, opened 2026-08-25
frontend: check production dependency licenses 

Tomasvrba:
Add a generic license check. Record expected licenses to flag changes, including between allowed licenses, and support package-specific exceptions.  Run the check before builds and in CI. Add regression tests and document how to review and resolve failures.

### DIFF
--- a/frontends/web/scripts/check-walletconnect-licenses.mjs
+++ b/frontends/web/scripts/check-walletconnect-licenses.mjs
@@ -1,0 +1,32 @@
+#!/usr/bin/env node
+
+import { readFileSync } from 'node:fs';
+import semver from 'semver';
+
+const FIRST_COMMUNITY_LICENSED_VERSION = {
+  '@reown/walletkit': '1.2.11',
+  '@walletconnect/core': '2.21.9',
+  '@walletconnect/sign-client': '2.21.9',
+  '@walletconnect/types': '2.21.9',
+  '@walletconnect/utils': '2.21.9',
+};
+
+const { packages } = JSON.parse(
+  readFileSync(new URL('../package-lock.json', import.meta.url), 'utf8')
+);
+
+for (const [packagePath, { version }] of Object.entries(packages)) {
+  const policy = Object.entries(FIRST_COMMUNITY_LICENSED_VERSION).find(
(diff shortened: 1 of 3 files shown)


frontends/web/scripts/check-walletconnect-licenses.mjs
```diff
@@ -0,0 +1,32 @@
...
+import { readFileSync } from 'node:fs';
+import semver from 'semver';
+
+const FIRST_COMMUNITY_LICENSED_VERSION = {
+  '@reown/walletkit': '1.2.11',
+  '@walletconnect/core': '2.21.9',
+  '@walletconnect/sign-client': '2.21.9',
+  '@walletconnect/types': '2.21.9',
+  '@walletconnect/utils': '2.21.9',
+};
```
thisconnect, line 12: Instead of hardcoded list here, I currently slightly prefer to:

1) get production dependencies `npm list --omit=dev --json --package-lock-only`
2) check each dep license field in (package.json or the LICENSE file?)  if the license is something like MIT/Apache

but I would like to get more feedback from others.. ping @bznein @Beerosagos 

and we probably should do something similar in the future for the go deps.

Tomasvrba: I see where you're coming from and it makes sense for a generic solution if we want to extend this check to all other dependencies.
But when it comes to wallet connect, we don't need to check the license, because we already now exactly when they changed it.

If we already had a generic license check in place, I'd obviously agree that we shouldn't hardcode this. But since WC is already in production and it's important we don't pull in the community license, it seems more important to add this rather than postpone until we have a generic check. If you'd like me to make it completely generic, we can do that, but imo that's out of scope for this particular issue, can be added in a separate PR later and this WC check can be discarded

thisconnect: > But when it comes to wallet connect, we don't need to check the license, because we already now exactly when they changed it.

well hardcoding those deps/versions wouldn't work in the case if they even release a version that is again opensource.

> If you'd like me to make it completely generic, we can do that, but imo that's out of scope for this particular issue, can be added in a separate PR later and this WC check can be discarded

I'd 100% prefer a generic check and not introduce it for WC only, but I'd like to get feedback from others maybe cc @NickeZ ?

PR #4359, opened 2026-08-25
Wallet connect multi chain support

Tomasvrba:
This PR adds WalletConnect support for multiple EVM networks:

- Ethereum
- optimism
- BNB Smart Chain
- Gnosis
- Polygon
- HyperEVM
- Base
- Arbitrum
- Ethereum Sepolia

The BitBoxApp now:

- Handles required and optional permissions requested by dApps.
- Shows the requested networks with their names and icons.
- Supports read-only WalletConnect sessions.
- Rejects connection requests that require unsupported features.
- Uses the correct network when preparing, signing, and broadcasting
transactions.
- Keeps pending transactions from different networks separate.

This depends on https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4267 but would be nice to get into the next release. Only the top 3 commits are for this PR.


backend/coins/eth/account.go
```diff
@@ -500,16 +497,14 @@ func (account *Account) Balance() (*accounts.Balance, error) {
 	return accounts.NewBalance(account.balance, coin.NewAmountFromInt64(0)), nil
 }
 
-// TxProposal holds all info needed to create and sign a transacstion.
+// TxProposal holds all information needed to create and sign a transaction.
 type TxProposal struct {
-	Coin *Coin
-	Tx   *types.Transaction
-	Fee  *big.Int
+	ChainID uint64
```
benma, line 502: It seems conceptually off to use a coin/account instance tied to a specific chainID to handle transactions of walletconnect by arbitrary other chains. Now account.go has its own chainID and is supplied an external one, which is probably confusing and brittle in the long run, one has to always check to not mix anything up.

Is there a reason for mixing all of this here, or could walletconnect not simply call shared helper functions to get the next nonce, sign the tx, etc?

backend/coins/eth/account.go
```diff
@@ -339,31 +334,33 @@ func (account *Account) outgoingTransactions(allTxs []*accounts.TransactionData)
 	return transactions, nil
 }
 
-func (account *Account) nextNonce() (uint64, error) {
-	var nextNonce uint64
-
-	// Nonce to be used for the next tx, fetched from the ETH node. It might be out of date due to
-	// latency, which is addressed below by using the locally stored nonce.
-	nodeNonce, err := account.coin.client.PendingNonceAt(context.TODO(), account.address.Address)
+func (account *Account) nextNonceForChain(
```
benma, line 337: Does the tx coming from WC not supply the nonce?

benma (commented):
Codex finds below.

P3 is tricky, as it's annoying to deal with - does it need to be dealt with?

---


  1. [P1] Native-chain contract transactions can crash the backend.
     broadcastTransaction() (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/13e12d20a54b6e5ec5eb40ae4420b7142c149fb1/backend/coins/eth/account.go#L719) now stores WalletConnect transactions and schedules
     an account update. However, TransactionData() (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/13e12d20a54b6e5ec5eb40ae4420b7142c149fb1/backend/coins/eth/types/types.go#L109) panics when an ETH
     account’s pending transaction contains calldata. A pending approval or swap on Ethereum/Sepolia can hit this path. Pending-transaction presentation must support general contract calls before these
     transactions enter that storage. The new tests only exercise empty calldata.

  2. [P2] "input": null is incorrectly rejected.
     The new parser (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/13e12d20a54b6e5ec5eb40ae4420b7142c149fb1/backend/coins/btc/handlers/walletconnect.go#L35) rejects every non-nil json.RawMessage,
     including JSON null. The previous implementation explicitly accepted it, and its corresponding test was removed during the refactor. This breaks previously accepted requests. Preserve the distinction between
     an absent/null field and a populated unsupported alias.

  3. [P2] Other chains lack pending-transaction protection.
     Non-native broadcasts skip storage (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/13e12d20a54b6e5ec5eb40ae4420b7142c149fb1/backend/coins/eth/account.go#L716), and nonce selection trusts the remote
     nonce directly (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/13e12d20a54b6e5ec5eb40ae4420b7142c149fb1/backend/coins/eth/account.go#L346). If the provider returns a stale pending nonce, consecutive
     requests can reuse it, causing replacement or rejection. Transactions lost after an apparently successful broadcast also cannot be rebroadcast. The native path already accounts for these provider behaviors.
     Tracking by chain and address would preserve those protections without mixing networks.

backend/coins/eth/chains.go
```diff
@@ -0,0 +1,23 @@
...
+	1:        {SupportsType2Transactions: true},
+	10:       {SupportsType2Transactions: true},
+	56:       {SupportsType2Transactions: true},
+	100:      {SupportsType2Transactions: true},
+	137:      {SupportsType2Transactions: true},
+	146:      {SupportsType2Transactions: true},
+	999:      {SupportsType2Transactions: true},
+	8453:     {SupportsType2Transactions: true},
+	42161:    {SupportsType2Transactions: true},
+	11155111: {SupportsType2Transactions: true},
```
benma, line 22: If all are the same, what is the var useful for?

backend/coins/eth/chains_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+package eth
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEVMChainCapabilities(t *testing.T) {
```
benma, line 11: This is basically a copypaste of the actual var, don't think this test is valuable and could be removed.

backend/coins/eth/chains.go
Tomasvrba: ah, my bad, that was a leftover from my initial spec, I assumed some chains might not support type 2 txs, but turns out all the ones we have support it

backend/coins/eth/account.go
Tomasvrba: not necessarily, wc docs mark it as optional
nextNonceForChain is already only called when wc doesn't provide one:
https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4359/changes#diff-1804b4cc9303dd2c4aa6234b8b2e82dc45d6606ac1cf430aba6f52793ec8411fR1012-R1016

leaving this unchanged (sort of, the nonce handling was changed with the account ownership change above)

backend/coins/eth/account.go
Tomasvrba: addressed in https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4359/changes/7095079547fc96a5369a3c262ccfeb91a2921a7c

All other issues addressed in the fixup commits

frontends/web/src/utils/walletconnect.ts
```diff
@@ -1,36 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { ReactNode, createElement } from 'react';
 import { IWalletKit } from '@reown/walletkit';
-import { ArbitrumLogo, OptimismLogo, BaseLogo, ETHLogo } from '@/components/icon';
```
benma-agent, line 5: This removes the last uses of `ArbitrumLogo`, `OptimismLogo`, `BaseLogo`, and `ETHLogo`. Please also remove the four now-unused wrappers and their SVG imports from `frontends/web/src/components/icon/icon.tsx`.

backend/coins/btc/handlers/handlers.go
```diff
@@ -862,25 +861,38 @@ func (handlers *Handlers) postEthSignWalletConnectTx(r *http.Request) (interface
...
 	if !ok {
 		return signingResponse{Success: false, ErrorMessage: "Must be an ETH based account"}, nil
 	}
-	txHash, rawTx, err := ethAccount.EthSignWalletConnectTx(args.Send, *args.ChainId, args.Tx)
+	transaction, err := parseWalletConnectTransactionRequest(*args.ChainID, args.Tx)
+	if err != nil {
+		return signingResponse{Success: false, ErrorMessage: err.Error()}, nil
+	}
+	signedTx, err := ethAccount.SignTransaction(eth.SignTransactionArgs{
+		ChainID:     *args.ChainID,
```
benma, line 876: My comment about confusing ownership still seems to apply?

https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4359#discussion_r3966279279

chainID is not in the account struct anymore, but passed to functions on the account instance instead, which still means the account tied to one chain is handling other chains.

What do you think?

Codex suggestion:

---
 I would make these changes:                                                                                                                                                                                        
                                                                                                                                                                                                                     
  1. Move Account.SignTransaction() into a package-level eth.SignTransaction() function.                                                                                                                             
     Give it the request, target-chain client, signing configuration, the existing ConnectKeystore callback, and access to pending transactions for that chain/address. It should perform sender validation, nonce   
     selection, construction, signing, broadcasting, and tracking.                                                                                                                                                   
                                                                                                                                                                                                                     
     It should receive no *Account. Derive the expected sender and keypath from the supplied signing configuration. Keeping ConnectKeystore as a callback preserves the current ordering: validate and prepare       
     before connecting the device.                                                                                                                                                                                   
                                                                                                                                                                                                                     
  2. Move client selection out of Account.                                                                                                                                                                           
     Remove Account.chainClient() and the account’s chainClientProvider field. Backend wiring resolves the client from the requested chain ID and passes it to the function. It can reuse the existing native-chain  
     client when appropriate; that is dependency selection during setup.

Tomasvrba: Makes sense, addressed in top commit

benma (approved):
utACK, nice!

please address the comment about dead code here: https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4359#pullrequestreview-5185772953

PR #4362, opened 2026-08-26
backend: refresh update checks hourly

benma-agent:
Run automatic update checks hourly and append an incrementing c query
parameter, starting with c=0 at startup.

Refresh the update state when the About page opens and tag that check
with about=1. Serialize overlapping checks so an older response cannot
overwrite a newer result.


bznein (approved):
utACK

PR #4363, opened 2026-08-27
Fix syncInProgress error

jstrnbrg:
This PR resolves the syncInProgress error pop-up that appears when you try to swap to a destination account that has not been activated yet.

To reproduce try swapping BTC to a non-activated account on mainnet, you should get the syncInProgress error. Then disable the account again via settings, switch to this branch, restart webdev and servewaller-mainnet and try again. Now it should show that the account is being activated and then the swap proceeds.

### DIFF
--- a/backend/handlers/handlers.go
+++ b/backend/handlers/handlers.go
@@ -69,7 +69,7 @@ type Backend interface {
 	Coin(coinpkg.Code) (coinpkg.Coin, error)
 	Testing() bool
 	Accounts() backend.AccountsList
-	PrepareSwap(buyAccountCode, sellAccountCode accountsTypes.Code, routeID, sellAmount string) (*backend.SwapPreparation, error)
+	PrepareSwap(ctx context.Context, buyAccountCode, sellAccountCode accountsTypes.Code, routeID, sellAmount string) (*backend.SwapPreparation, error)
 	SwapAccounts() (backend.SwapAccounts, error)
 	SwapStatus() backend.SwapStatus
 	AccountsByKeystore() (backend.KeystoresAccountsListMap, error)
@@ -2029,6 +2029,7 @@ func (handlers *Handlers) postConnectKeystore(r *http.Request) interface{} {
 func (handlers *Handlers) postSwapSign(r *http.Request) interface{} {
 	type result struct {
 		Success           bool                     `json:"success"`
+		ErrorCode         string                   `json:"errorCode,omitempty"`
 		ErrorMessage      string                   `json:"errorMessage,omitempty"`
 		ExpectedBuyAmount string                   `json:"expectedBuyAmount,omitempty"`
 		SwapID            string                   `json:"swapId,omitempty"`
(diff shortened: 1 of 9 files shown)


backend/handlers/handlers.go
```diff
@@ -69,7 +69,8 @@ type Backend interface {
 	Coin(coinpkg.Code) (coinpkg.Coin, error)
 	Testing() bool
 	Accounts() backend.AccountsList
-	PrepareSwap(buyAccountCode, sellAccountCode accountsTypes.Code, routeID, sellAmount string) (*backend.SwapPreparation, error)
+	EnsureSwapAccountsReady(ctx context.Context, buyAccountCode, sellAccountCode accountsTypes.Code) error
```
benma: Looks overengineered on first glance - would not expect a new handler to deal with syncInProgress.


Codex comment:

---
The first commit (https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/9e3bb323220eae61b4363fe7fc9ef141f31d6f5c) was essentially the right
  architecture; the second commit introduced most of the complexity.

  Key findings:

  - The new /swap/ready endpoint is redundant. PrepareSwap must repeat swapAccountsReady() anyway, so readiness is not a durable API boundary.
  - A quote older than 45 seconds causes the click handler to refresh and return without calling signSwap (code
    (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/b7b3e4dcce4139e22a5e04f6ae1d2485cd69042f/frontends/web/src/routes/market/swap/swap.tsx#L438-L459)). The test explicitly expects this
    (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/b7b3e4dcce4139e22a5e04f6ae1d2485cd69042f/frontends/web/src/routes/market/swap/swap.test.tsx#L575-L580). The user therefore has to click Swap again,
    silently.

  - Quote expiry does not justify the preflight endpoint. SwapKit says /swap automatically refreshes quotes older than 60 seconds, while retaining the route for five minutes. The refreshed response includes the
    current price. SwapKit documentation (https://docs.swapkit.dev/swapkit-api/v3-swap-obtain-swap-transaction-details)

  - The confirmation currently still displays the old expectedOutput, despite calculating fiat conversions from response.expectedBuyAmount (code
    (https://github.com/BitBoxSwiss/bitbox-wallet-app/blob/b7b3e4dcce4139e22a5e04f6ae1d2485cd69042f/frontends/web/src/routes/market/swap/swap.tsx#L477-L500)). That is the real quote-refresh correctness gap.

  I’d simplify it to:

  1. Keep the bounded, cancellable account-sync wait inside PrepareSwap.
  2. Keep passing r.Context() through the existing /swap/sign handler and into SwapKit.
  3. Remove /swap/ready, EnsureSwapAccountsReady, the frontend API wrapper, quote timestamps, and refresh counter.
  4. Call signSwap once under the existing loading state.
  5. Display response.expectedBuyAmount, converted through getSwapDisplayAmount(), in the confirmation.
  6. Optionally keep the form locking—it is small and independently sensible.

  That keeps activation, synchronization, address derivation, and swap preparation as one atomic backend operation while reducing the PR substantially. No PR comments or local changes were made.

benma (changes requested):

PR #4367, opened 2026-08-31
fix(lightning): handle initialization failures

shonsirsha:
1. Expose connection status and retry support so failed startup cannot leave Lightning and portfolio views loading forever.

2. Return partial portfolio data when Lightning is unavailable and

3. Keep the mobile bottom navigation visible during Lightning initialization failures.

### DIFF
--- a/backend/accounts.go
+++ b/backend/accounts.go
@@ -317,8 +317,9 @@ func (backend *Backend) formattedCoinBalance(
 }
 
 // getCoinsTotalBalance returns the total balances grouped by coins.
-func (backend *Backend) coinsTotalBalance() ([]coinFormattedAmount, error) {
+func (backend *Backend) coinsTotalBalance() ([]coinFormattedAmount, []coinpkg.Code, error) {
 	coinFormattedAmounts := []coinFormattedAmount{}
+	unavailableCoinCodes := []coinpkg.Code{}
 	var sortedCoins []coinpkg.Code
 	totalCoinsBalances := make(map[coinpkg.Code]*big.Int)
 
@@ -331,12 +332,12 @@ func (backend *Backend) coinsTotalBalance() ([]coinFormattedAmount, error) {
 		}
 		err := account.Initialize()
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 		coinCode := account.Coin().Code()
 		b, err := account.Balance()
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 		amount := b.Available()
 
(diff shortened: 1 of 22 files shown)


backend/lightning/connection.go
```diff
@@ -0,0 +1,74 @@
...
+	// ConnectionConnecting means that the SDK connection is being initialized.
+	ConnectionConnecting ConnectionState = "connecting"
+	// ConnectionReady means that the SDK connection is ready for requests.
+	ConnectionReady ConnectionState = "ready"
+	// ConnectionFailed means that SDK connection initialization failed.
+	ConnectionFailed ConnectionState = "failed"
+)
+
+// ConnectionErrorInitializationFailed is returned when the Lightning SDK could not initialize.
+const ConnectionErrorInitializationFailed = "initializationFailed"
```
bznein, line 26: From what I can see this is the same conceptually as any of the error above, so it should be with them and typed as ConnectionState

backend/lightning/connection.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (lightning *Lightning) setConnectionStatus(state ConnectionState, errorCode string) {
+	status := ConnectionStatus{
+		State:     state,
+		ErrorCode: errorCode,
+	}
+
+	lightning.connectionStatusLock.Lock()
+	if lightning.connectionStatus == status {
+		lightning.connectionStatusLock.Unlock()
```
bznein, line 49: unless I am mistaken, there is no reason to just not do `defer lighting.connectionStatusLock.Unlock()` after line 47 instead of manually adding it here

backend/lightning/lightning.go
```diff
@@ -349,10 +360,28 @@ func accountBreezFolder(accountCode types.Code) string {
 }
 
 // connect initializes the connection configuration and calls connect to create a Breez SDK instance.
-func (lightning *Lightning) connect() error {
+func (lightning *Lightning) connect() (connectErr error) {
```
bznein, line 363: Is the addition of a named return needed because of the deferred method at line 378? If so, it's fine but it feels a bit out of place as we never do it in other parts of the codebase. Maybe you can see if codex has a cleaner way to do it? Not a blocker though

backend/accounts.go
```diff
@@ -317,8 +317,9 @@ func (backend *Backend) formattedCoinBalance(
 }
 
 // getCoinsTotalBalance returns the total balances grouped by coins.
```
bznein, line 319: we should update the docstring to document the new returned field

backend/accounts.go
```diff
@@ -473,8 +483,9 @@ func (backend *Backend) convertBtcAmountToFiat(amount coinpkg.Amount, fiat strin
 
 // AccountsBalanceSummary holds the total balance for each coin and of each keystore.
```
bznein, line 484: Please update docstring here

bznein (commented):
Codex findings:

```
# Review comments                                                                                                                                                                                         
                                                                                                                                                                                                            
Connection retry can race teardown and leave the Lightning SDK active after 
deactivation. Status synchronization can also lose the terminal connection
state,  recreating the indefinite loading behavior  this patch intends to fix.                                                                                                                                                                                

## P1: Serialize reconnect with account teardown                                                                                                                                                                                                                                                                                                                                                                   
`backend/lightning/connection.go:72`                                                                                                                                                                      
When a user retries and then deactivates the wallet or closes the backend before 
`Connect` finishes, this goroutine can continue using the captured account while 
`Deactivate` removes its config, key, and cache. Because `Disconnect` does not share
 `connectionLock`, it may observe a nil `sdkService`; the connection can then publish
 `ready` and leave a live SDK after deactivation, or race with SDK destruction. Serialize
 or cancel connection attempts during teardown.      

## P2: Prevent stale status snapshots from overwriting events      
`frontends/web/src/contexts/LightningProvider.tsx:19-22`                  
 When the SDK changes from `connecting` to `ready` or `failed` while the initial status
 request is in flight, `useSync` can apply the WebSocket event first and then overwrite it
with the older `connecting` response. No later status event is guaranteed, so the
 Lightning page can remain on the initialization spinner instead of showing the wallet or
 Retry action. Subscribe before reconciling the initial snapshot, or add revision ordering.

PR #4347, opened 2026-08-24
eth: parse ERC-681 payment requests

Tomasvrba:
Correctly parse native ETH and ERC20 payment requests scanned from QR codes. Use the request recipient and replace the entered amount only when the request includes one.

Before asking for reviews, here is a check list of the most common things you might need to consider:
- [x] updating the Changelog
- [x] writing unit tests
- [x] checking if your changes affect other coins or tokens in unintended ways
- [x] testing on multiple environments (Qt, Android, ...)
- [x] having an AI review your changes

PR #4370, opened 2026-08-31
eth: simplify and harden ERC-681 parsing

benma-agent:
Follow-up to #4347.

## Summary

- replace manual scientific-notation expansion with the shared exact amount parser
- reject impossible exponents and mantissas before invoking the generic parser
- reuse the canonical coin formatter and make it honor each token decimal count
- normalize address prefixes before mixed-case checksum validation
- clarify the supported ERC-681 subset and network-or-asset mismatch message
- add coverage for amount boundaries, unusual decimal counts, and address prefixes

The formatter generalization is kept in a separate commit.

## Validation

- go test -mod=vendor ./...
- make weblint
- git diff --check

### DIFF
--- a/backend/coins/eth/account.go
+++ b/backend/coins/eth/account.go
@@ -41,7 +41,6 @@ import (
 )
 
 func isMixedCase(s string) bool {
-	s = strings.TrimPrefix(strings.TrimPrefix(s, "0x"), "0X")
 	return strings.ToLower(s) != s && strings.ToUpper(s) != s
 }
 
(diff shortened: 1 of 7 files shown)


Tomasvrba (changes requested):
Good improvements imo, touches some stuff that wasn't part of the previous PR, but the improvements are still good.

Only one nit, here the original string was passed instead of the bounded amount, opened a PR against your branch: https://github.com/benma-agent/bitbox-wallet-app/pull/1

After that it's good to merge

Tomasvrba (approved):
ack

PR #4381, opened 2026-09-01
lightning: move files outside of cache and handle clearing cache 

bznein:
This PR contains 2 commits: the first one moves the Lightning files outside of cache so that clearing the cache does not delete them; codex also found an issue such that clearing cache will recreate the rates updates and the btc coin, but Lightning still had pointers to the old ones; the second commit fixes it.

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos (approved):
tACK! 👍

PR #4382, opened 2026-09-01
frontend: fix spacing eth receive screen

shonsirsha:
On systems that don't support pasting programatically, the paste button wont be displayed on several inputs such as the input on the send screen for Ethereum.

However, the padding didn't get adjusted. This commit adjusts the padding so it doesn't leave whitespace.

Screenshot 2026-09-01 at 16 30 06


thisconnect (approved):
untested LGTM

PR #4383, opened 2026-09-01
frontend: slightly push down settings search icon

shonsirsha:
Slightly push down (2px) settings search icon


sutterseba (approved):
utACK

PR #4388, opened 2026-09-02
frontend: do not override window app icon on macos

sutterseba:
https://github.com/BitBoxSwiss/bitbox-wallet-app/pull/4050 introduced native macOS Tahoe icons, but they are overridden when the app launches i.e. the icon changes whenever you close/open the app. Fixed by skipping this override on macOS.


shonsirsha (approved):
LGTM

PR #4389, opened 2026-09-02
frontend: show connect keystore prompt on top-up

bznein:
If the user tries to top-up but no keystore is connected, instead of redirecting them to the home page, show the connectAnyKeystore prompt.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/lightning/topup/topup-result.tsx
+++ b/frontends/web/src/routes/lightning/topup/topup-result.tsx
@@ -2,6 +2,7 @@
 
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
+import { connectAnyKeystore } from '@/api/keystores';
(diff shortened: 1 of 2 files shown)


frontends/web/src/routes/lightning/topup/topup-result.tsx
```diff
@@ -51,17 +52,14 @@ export const TopUpSuccess = () => {
...
-        route: '/settings/manage-accounts',
-      }
-      : {
-        label: t('welcome.connect'),
-        route: '/',
-      }
-  );
+
+  const handlePrimaryAction = async () => {
+    if (hasAccounts) {
```
thisconnect, line 58: tiny nit: Could you add a short comment that this is only in the edge-case when there is accounts but no BTC-accounts?

thisconnect (approved):
tested Android and webdev LGTM with small nit to add a comment

PR #4401, opened 2026-09-08
frontend: use name attribute to supress autocomplete

thisconnect:
Fixed home address autofill suggestion in iOS.

- [ ] updating the Changelog
- [ ] writing unit tests


frontends/web/src/routes/account/send/components/inputs/receiver-address-input.tsx
```diff
@@ -88,6 +88,7 @@ export const ReceiverAddressInput = ({
             Send to self
           </span>
         ) : undefined}
+        name="onchain-destination"
```
thisconnect, line 91: I am not 100% that it is fixed by adding a name attribute only..

but this is the only difference I could find compared to onchain address.

we already set autoComplete=off and suggestions to false if I read the code correctly

PR #4403, opened 2026-09-08
frontend: remove expert section in lighthing settings

thisconnect:
Lightning is already an advanced feature so it does not need to have its own expert setting for closing the wallet.

Closing the wallet should not be an expert feature.

- [ ] updating the Changelog
- [ ] writing unit tests


Beerosagos (approved):
tACK

PR #4407, opened 2026-09-08
frontend: always use 256px for lightning receive qr code

thisconnect:
This could be larger, but the backend currently sends an image that is 256px so anything larger would currently look blurry.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/routes/lightning/receive/receive.tsx
+++ b/frontends/web/src/routes/lightning/receive/receive.tsx
@@ -169,7 +169,10 @@ export function Receive() {
             <div className={styles.addressContent}>
               <p className={styles.qrInstruction}>{t('lightning.receive.address.scanQRCode')}</p>
               <div className={styles.addressQRCode}>
-                <QRCode data={lightningAddress || undefined} size={168} tapToCopy={false} />
+                <QRCode
+                  data={lightningAddress || undefined}
+                  size={216}
+                  tapToCopy={false} />
               </div>
               {lightningAddress && (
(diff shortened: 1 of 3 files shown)


frontends/web/src/routes/lightning/receive/receive.tsx
```diff
@@ -169,7 +169,10 @@ export function Receive() {
             <div className={styles.addressContent}>
               <p className={styles.qrInstruction}>{t('lightning.receive.address.scanQRCode')}</p>
               <div className={styles.addressQRCode}>
-                <QRCode data={lightningAddress || undefined} size={168} tapToCopy={false} />
+                <QRCode
+                  data={lightningAddress || undefined}
+                  size={216}
```
Beerosagos, line 174: says 216 here, while the PR title says 256. Is this on purpose?

Beerosagos (approved):
tACK with a question :)

frontends/web/src/routes/lightning/receive/receive.tsx
Beerosagos: Tested: it looks good, anyway

thisconnect: thanks. This is the small one it was 168 before. This never had a problem but still can be slightly bigger. I'll keep as is for now. 

We might want to increase and make the bolt11 invoice larger than 256 but that needs to come from the BE

PR #4408, opened 2026-09-08
frontend: update waiting view

thisconnect:
Slightly decrease horizontal space for the get started text. Updated .withMobileSafetyMargin so it works in fullscreen and normal mode.

Added custom footer on mobile with DBB logo. Dropped langauge switch as it can be accessed from the settings. Dropped old swiss-made-opensource graphic.

- [ ] updating the Changelog
- [ ] writing unit tests


sutterseba (approved):
tested LGTM

PR #4420, opened 2026-09-10
feat: disable LN on testnet

shonsirsha:


bznein (approved):
backend tACK

frontends/web/src/app.tsx
```diff
@@ -272,18 +267,22 @@ export const App = () => {
...
   );
 };
+
+export const App = () => (
+  <ConnectedApp>
+    <Providers>
+      <AppContent />
+    </Providers>
+  </ConnectedApp>
+);
```
thisconnect, line 288: Everything else looks good to me, but this restructuring seems not related/needed?

thisconnect (approved):
untested LGTM with nit

thisconnect (commented):
In the meantime @bznein added an explicit  warning  https://github.com/BitBoxSwiss/bitbox-wallet-app/commit/537075f38b707eb09043436395e6ae47bd8e3ec0

thisconnect (changes requested):
It's not clear to me if and how we should disable. Lets keep as is for as long as Lightning is in beta.

PR #4423, opened 2026-09-10
backend: add accounts database boundary

benma-agent:
Account and keystore records are persisted state, but backend callers previously read
and wrote them directly through the JSON config. This made read consistency and write
ownership difficult to reason about and coupled account code to the current storage format.

Put coherent reads behind Snapshot and route every backend write through Update. The JSON
adapter initially preserves existing update semantics, while the boundary provides
caller-owned snapshots and one persistence path that can later be replaced by structured
SQLite tables.


backend/notes.go
```diff
@@ -221,6 +220,10 @@ func (backend *Backend) ImportNotes(jsonLines []byte) (*ImportNotesResult, error
 	if err := sanityCheck(); err != nil {
 		return nil, err
 	}
+	accountsConfigSnapshot, err := backend.accountsDB.Snapshot()
+	if err != nil {
+		return nil, err
+	}
```
bznein: [mini-nit]: can we move this a bit closer to when it is actually needed?

bznein (approved):
utACK very clean diff thanks :)

PR #4424, opened 2026-09-10
frontend: restore scrolling above bottom navigation on iOS

Beerosagos:
WebKit 26.4 can omit flex container padding from scrollable overflow, leaving the LNURL send button behind the bottom navigation with no room to scroll. Replace the mobile bottom padding with a non-shrinking flex spacer to preserve that scroll space.

- [ ] updating the Changelog
- [ ] writing unit tests

### DIFF
--- a/frontends/web/src/app.module.css
+++ b/frontends/web/src/app.module.css
@@ -40,8 +40,11 @@
        max-height: 100dvh;
     }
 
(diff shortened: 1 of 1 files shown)


thisconnect (approved):
tested in webdev LGTM

PR #4426, opened 2026-09-12
eth: handle missing fees in pending transactions

benma-agent:
Internal ETH transfers have no separate fee, so Etherscan represents their fee as nil.
Successful transfers remain pending until 12 confirmations. Account updates include these
transfers in pendingTxsAmount(), which dereferences the fee unconditionally for ETH accounts.

Refreshing an account with a recent internal transfer can therefore panic in a background
goroutine and terminate the app. This can happen after sending, during startup, or during
periodic synchronization, including for incoming internal transfers. The amount is irrelevant.

Restarting can encounter the same transfer until it reaches 12 confirmations. Clearing the
cache does not directly resolve this condition; time spent clearing it and rebooting may
allow the transfer to reach the confirmation threshold.

Skip absent fees when computing the pending amount.


bznein (approved):
utACK

PR #4428, opened 2026-09-14
backend: centralize account lifecycle

benma-agent:
Account membership and lifecycle were spread across Backend helpers, with sorting,
observation, initialization, and teardown all mutating the same runtime slice.
That made ownership unclear and made targeted changes hard to implement without
rebuilding every account.

Move live-account membership, initialization, observation, and teardown into an
unordered registry keyed by stable account codes. Keep presentation ordering in
backend snapshots. This gives each account one lifecycle path, preserves final
close events, and lets reconciliation add or remove accounts without coupling
runtime ownership to display order.

Part of #4276.


backend/account_registry.go
```diff
@@ -0,0 +1,104 @@
...
+func (registry *accountRegistry) all() AccountsList {
+	return slices.Clone(registry.accounts)
+}
+
+func (registry *accountRegistry) lookup(code accountsTypes.Code) accounts.Interface {
+	return registry.accounts.lookup(code)
+}
+
+// add registers and initializes an account unless its code is already present. Initialization
+// failures leave the account registered, matching the previous backend behavior.
```
bznein, line 51: The last part "matching the previous backend behavior" doesn't really make sense in a comment imho; I know codex tends to mention things like this when it has been part of its prompt :)

bznein (approved):
changes LGTM! 

utACK

backend/account_registry.go
benma: Ah yeah, very true. I'll open a separate PR to add this to AGENTS.md.

PR #4435, opened 2026-09-16
frontend: hide redundant lightning unit price

shonsirsha:
Hide LN's unit price when the display currency is BTC or sat. Other than that, keeps existing behaviour.

Screenshot 2026-09-16 at 12 02 19


sutterseba (approved):
tACK

PR #4438, opened 2026-09-16
backend: synchronize route registration and matching

benma-agent:
Account and device initialization adds routes while HTTP requests match
against the same mux, allowing concurrent reads and writes of its route list.

Use the existing handlers map lock for route matching as well as registration.
Release the read lock before executing the handler so requests can register
accounts or devices without deadlocking.

Addresses the pre-existing router race reported by CI in #4428.


bznein (approved):
utACK

PR #4439, opened 2026-09-17
frontend: reset recipient account on scan or paste

bznein:
Scanning or pasting could replace a selected account's address while leaving its name in the selector.

Reuse the reset action before opening the scanner or applying nonempty clipboard text. Clear the selected account, restore address editing, and ignore pending account lookups after reset. Keep the selection when the clipboard is empty or cannot be read.

Cover scanning, pasting, clipboard failures, and late address responses with regression tests.

- [ ] updating the Changelog
- [ ] writing unit tests


bznein:
@coderabbit review

bznein:
ping @thisconnect  :)

PR #4446, opened 2026-09-20
backend: separate account metadata from runtime

benma-agent:
Changing an account's active status rebuilds loaded accounts even though their runtime membership
is unchanged. Join runtime accounts with snapshots of persisted metadata, including ERC20 metadata
derived from its ETH parent, so status changes can save the new value and notify listeners without
rebuilding accounts or triggering another initial sync.

Extracts the first two commits from #4276: the account-view refactor and the status-update change
that depends on it.


bznein (approved):
utACK

PR #4451, opened 2026-09-21
backend: centralize account load eligibility

benma-agent:
Initial loading and targeted reconciliation must make the same decisions
about which persisted accounts can own live resources. Keeping those rules in
the batch loader would invite the two paths to drift.

Centralize coin policy, watch-only, fingerprint, and keystore capability checks
in one helper backed by a coherent accounts database snapshot. This preserves
the existing support rules while making them reusable by incremental
reconciliation.

Extracted from #4276 for independent review.


bznein (approved):
utACK (there are merge conflicts)

